PR #6, opened 2018-09-11
a few package list changes

dustymabe:


fedora-coreos-base.yaml
```diff
@@ -77,7 +77,7 @@ packages:
  - podman skopeo runc
  # Networking
  - bridge-utils nfs-utils biosdevname iptables-services
- - NetworkManager dnsmasq
+ - NetworkManager dnsmasq hostname
  # Storage
  - container-storage-setup cloud-utils-growpart
```
dustymabe: while we're in here I'd like to ask.. do we need container-storage-setup ?

fedora-coreos-base.yaml
```diff
@@ -77,7 +77,7 @@ packages:
  - podman skopeo runc
```
dustymabe: i wonder if we should start pulling in moby-engine ?

fedora-coreos-base.yaml
```diff
@@ -90,6 +90,8 @@ packages:
  - authconfig sssd shadow-utils
  - logrotate
  # Used by admins interactively
- - sudo coreutils less tar xz gzip bzip2 rsync tmux
+ - sudo coreutils less tar xz gzip bzip2 tmux
  - nmap-ncat net-tools bind-utils
  - bash-completion
+ # Moving files around
+ - rsync fuse-sshfs
```
dustymabe: i've always wanted fuse-sshfs in so I could easily use vagrant-sshfs

fedora-coreos-base.yaml
cgwalters: https://pagure.io/atomic-wg/issue/384

PR #9, opened 2018-09-17
image: Use reflinks for / by default

cgwalters:
This is transparently used by overlayfs for copyups, rpm-ostree
can use it for the rpmdb, and ostree uses it for `/etc`.

It's The Future.  (Also probably the default for `mkfs.xfs` at
some point I imagine...if that happens we'll need to explicitly
turn it off for `/boot` or perhaps suck it up and use `ext4`
or something)

https://lwn.net/Articles/659677/
https://lwn.net/Articles/747633/


dustymabe:
LGTM

when we add a /var/ we'll want to also do it there too, right?

dustymabe:
leaving open for more input from others. will merge end of day.

PR #11, opened 2018-09-17
base: Fix the `Before=sshd` in coreos-growpart.service

cgwalters:
Apparently systemd just ignores it as is.


dustymabe:
this was done in https://github.com/coreos/fedora-coreos-config/pull/10

PR #12, opened 2018-09-18
image.ks: Drop touch /boot/coreos-firstboot

cgwalters:
See https://github.com/coreos/coreos-assembler/pull/96


jlebon:
I think it's been long enough now. Merging this!

PR #13, opened 2018-09-18
image.ks: Add some doc text

cgwalters:
See discussion in https://github.com/coreos/coreos-assembler/pull/96


jlebon:
While we're here, how about also something near that `%post` to mention that `coreos-assembler` may have additional `%post` scripts? I can imagine people scratching their heads trying to figure out where some behaviour comes from (e.g. `touch firstboot`) and a simple pointer like this saving lots of time.

cgwalters:
I added that.

PR #14, opened 2018-09-21
RPM Overlay to Reduce Python Dependencies

Promaethius:
Purpose of this repository: provide modified packages which do not require Python.

Currently in Repo:
- [container-selinux](https://copr-be.cloud.fedoraproject.org/results/promaethius/rpm-overlay/fedora-28-x86_64/00801076-container-selinux/container-selinux.spec)

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -14,6 +14,7 @@ repos:
   - fedora
   - fedora-updates
   - dustymabe-ignition
+  - promaethius-rpm-overlay
 
 ignore-removed-users:
   - root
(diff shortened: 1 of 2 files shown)


jlebon:
It'd be good if the copr dist-git included the patch that drop the python dependency as a separate commit to make it easier to review.

That said... it feels like this is something which we should tackle with the packagers themselves. E.g. proposing patches to split out python bindings into separate binary RPMs.

dustymabe:
thanks @Promaethius! 



> It'd be good if the copr dist-git included the patch that drop the python dependency as a separate commit to make it easier to review.
> 
> That said... it feels like this is something which we should tackle with the packagers themselves. E.g. proposing patches to split out python bindings into separate binary RPMs.

agree on both points. @Promaethius do you think you could submit patches for the changes.. For example a PR to https://src.fedoraproject.org/rpms/container-selinux

Promaethius:
Based on https://src.fedoraproject.org/rpms/container-selinux/pull-request/1 getting merged, I'm closing this. I'll use the copr repo for my own testing and defer to @cgwalters cool rpm-overlay.

PR #15, opened 2018-09-24
image: install default PATH file

arithx:
This drops a file in `/etc/profile.d` which configures the default
value of `PATH` for FCOS. This is mostly only useful for testing
purposes.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -49,6 +49,16 @@ postprocess:
     echo '%sudo        ALL=(ALL)       NOPASSWD: ALL' > /etc/sudoers.d/coreos-sudo-group
     # We're not using resolved yet
     rm -f /usr/lib/systemd/system/systemd-resolved.service
+    # https://github.com/openshift/os/issues/191
+    # install default PATH file to expand non-login shells PATH for kola
+    cat > /etc/profile.d/path.sh << 'EOF'
+    pathmunge /bin
+    pathmunge /sbin
+    pathmunge /usr/bin
+    pathmunge /usr/sbin
+    pathmunge /usr/local/bin
+    pathmunge /usr/local/sbin
+    EOF
     # https://github.com/coreos/fedora-coreos-tracker/issues/18
     # See also image.ks.
     # Growpart /, until we can fix Ignition for separate /var


cgwalters:
Hmm.  When you say testing purposes...like for kola?  

Today we have
```
# Path manipulation
if [ "$EUID" = "0" ]; then
    pathmunge /usr/sbin
    pathmunge /usr/local/sbin
else
    pathmunge /usr/local/sbin after
    pathmunge /usr/sbin after
fi
```

arithx:
@cgwalters yes, without this patch no `sbin` utilities were found.

cgwalters:
Related: https://github.com/openshift/os/pull/205

cgwalters:
OK it'd be useful to have a link to the original discussion, I am not finding it offhand.

The thing that confuses me still is - what is kola doing differently about its ssh logins that it fails to have `sbin` in path?

arithx:
Similar RHCOS Change PR: https://github.com/openshift/redhat-release-coreos/pull/8
Original RHCOS bug report: https://github.com/openshift/os/issues/191

arithx:
Mantle SSH Agent Creation: https://github.com/coreos/mantle/blob/master/network/ssh.go#L54
Base SSH function for clusters: https://github.com/coreos/mantle/blob/master/platform/base.go#L107

cgwalters:
OK so...I think this has something to do with login shells and the real fix should probably live in kola.  But I am also OK to merge this as is in the interest of velocity/momentum.

dustymabe:
> Hmm, should this conceptually be in `fedora-coreos-base.yaml`'s `postprocess` instead? We want to keep the `%post` in the cloud kickstart limited to cloud-related changes, right?

I agree, but I also think it's worth noting that over time we'll probably think the image output of the kickstart+anaconda run as less cloud related and more as the "global image" we want to get propagated everywhere that also get's put on bare metal with a simple install script. Eventually we may remove anaconda even from that first step.

dustymabe:
@arithx - do you mind upating this PR to put the change in the postprocess part of the manifest? https://github.com/coreos/fedora-coreos-config/blob/master/fedora-coreos-base.yaml#L34

arithx:
Moved to `postprocess` ⬆️

cgwalters:
> I agree, but I also think it's worth noting that over time we'll probably think the image output of the kickstart+anaconda run as less cloud related and more as the "global image" we want to get propagated everywhere that also get's put on bare metal with a simple install script. 

Yes, but there's still a difference between "in ostree" vs "in image" - previously `ostree admin config-diff` would have shown `/etc/profile.d/path` as a user-created file - and hence a "factory reset" would eliminate it.

We also want to drain `image.ks` of content in preparation for using a non-Anaconda path.

cgwalters:
LGTM, will give a bit for other comments.

dustymabe:
> Yes, but there's still a difference between "in ostree" vs "in image" - previously `ostree admin config-diff` would have shown `/etc/profile.d/path` as a user-created file - and hence a "factory reset" would eliminate it.
> 
> We also want to drain `image.ks` of content in preparation for using a non-Anaconda path.

100% agree with both statements

dustymabe:
merging.

PR #16, opened 2018-09-25
base: Ensure version numbers are strings

cgwalters:
Newer versions of serde that we may use for rpm-ostree are more
strict about this.


jlebon:
Ahh heh, good ol' YAML typing gotcha.

dustymabe:
wow, nice

PR #17, opened 2018-09-26
base.yaml: Strict mode postprocessing and add spacing

jlebon:
Use the regular bash strict mode header. Also add some spaces around
unrelated chunks of code to make reading easier.


dustymabe:
LGTM

PR #18, opened 2018-09-26
base: Use /var/home for home directories by default

cgwalters:
This came up in `#silverblue` discussion; let's do it here
first under the general concept that most changes to the "core"
underlying both Silverblue and FCOS should be made in FCOS first.

Tested locally with a build+run.  It doesn't affect the `core`
user today, because anaconda+libuser aren't reading the defaults
from the target OS.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -48,6 +48,9 @@ postprocess:
     # Persistent journal by default
     echo 'Storage=persistent' >> /etc/systemd/journald.conf
 
+    # Fix /home to be /var/home - software wants to canonicalize this.
+    sed -i -e 's,^HOME=.*,HOME=/var/home,' /etc/default/useradd
+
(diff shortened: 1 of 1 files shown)


miabbott:
LGTM

jlebon:
Can you expand on this? Shouldn't it Just Work :tm: both ways (canonicalized or not)?

cgwalters:
Further discussion about this in https://discussion.fedoraproject.org/t/adapting-user-home-in-etc-passwd/487/6

cgwalters:
> Can you expand on this? Shouldn't it Just Work tm both ways (canonicalized or not)?

I think this is basically about improving the UX in edge cases.  Going to go ahead and merge since it is an improvement.

cgwalters:
And after reflecting on this later on in the evening, I decided instead to do:

https://github.com/projectatomic/rpm-ostree/pull/1726

Pushed a revert of this one to master.

PR #21, opened 2018-10-04
base: Drop grub2 auto-hide and boot counting bits

cgwalters:
The design where "boot success" is "interactive user login"
doesn't match well with our expected use case.  Plus, I think
the boot menu hiding clashes with the ostree value proposition.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -41,6 +41,12 @@ postprocess:
       sed -i -e 's,enabled=[01],enabled=0,' ${x}
     done
 
+    # The grub bits are mainly designed for desktops, and IMO haven't seen
+    # enough testing in concert with ostree.  At some point we'll flesh out
+    # the full plan in https://github.com/coreos/fedora-coreos-tracker/issues/47
+    rm -v /etc/grub.d/01_{fallback_counting,menu_auto_hide}
+    rm -v /usr/lib/systemd/user/grub-boot-success*
+
     # See machineid-compat in host-base.yaml.
     # Since that makes presets run on boot, we need to have our defaults in /usr
     ln -sfr /usr/lib/systemd/system/{multi-user,default}.target


vathpela:
There's a typo here, you've got 01_ both inside and outside the brackets.

cgwalters:
> There's a typo here, you've got 01_ both inside and outside the brackets.

Thanks for catching that!  I updated this to use `-v` instead of `-f`, so things fail if the files aren't found.

While you're here @vathpela - we're basically going into the guts of grub here and removing part of it, as best I can tell the functionality is encapsulated in those files, but please keep this in mind going forward.

There's a lot of interacting things here - the idea seems nice for desktops, but for servers we really want to see the console.

I also just want to wait to have some time to validate all of the interactions with ostree myself, or have confidence that someone else has done it.  We're almost certain to look at the boot counting stuff, just not right now.

dustymabe:
LGTM

LorbusChris:
@cgwalters with current rpm build grub2-2.02-62.fc29 this will fail as the boot counting part has been (temporarily) removed from the package: https://src.fedoraproject.org/rpms/grub2/c/7c415d82878e13fcd626a9cbd3a336e208f15d33.

dustymabe:
@LorbusChris was that part of the work you did during GSOC? 

you're saying we don't need this PR now?

LorbusChris:
@dustymabe Yes, it was. It was merged prematurely, though, and has not yet worked the way intended.
See https://bugzilla.redhat.com/show_bug.cgi?id=1614637 and fix rhboot/grub2#42

LorbusChris:
@dustymabe and to answer the question: If we want to remove these two scripts, we will need this PR as is as soon as https://github.com/rhboot/grub2/pull/42 unblocks it.

LorbusChris:
Update: The code got merged again, but hasn't yet made it into the rpm. I'm tracking this.

dustymabe:
cc @LorbusChris - what should we do with this PR. Still waiting on something? Should we close it?

LorbusChris:
@dustymabe grub2 in F29 does contain the menu hide script:
```
$ rpm -qlp grub2-tools-2.02-62.fc29.x86_64.rpm | grep /etc/grub.d/01_*
/etc/grub.d/01_menu_auto_hide
/etc/grub.d/01_users
```
we probably still want to drop that.

LorbusChris:
Also, we will need this when we begin composing F30 packages.

fedora-coreos-base.yaml
```diff
@@ -41,6 +41,12 @@ postprocess:
       sed -i -e 's,enabled=[01],enabled=0,' ${x}
     done
 
+    # The grub bits are mainly designed for desktops, and IMO haven't seen
+    # enough testing in concert with ostree.  At some point we'll flesh out
+    # the full plan in https://github.com/coreos/fedora-coreos-tracker/issues/47
+    rm -v /etc/grub.d/01_{fallback_counting,menu_auto_hide}
```
LorbusChris: `rm -v /etc/grub.d/01_menu_auto_hide`

should suffice for now

LorbusChris:
fixed in #36, also added a TODO to add back removal of fallback script in f30.

PR #25, opened 2018-10-31
manifest: add fedora-updates-testing

jlebon:
So that we get the latest OSTree version which is needed to match the
recent Ignition/OSTree initrd services re-ordering.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -13,6 +13,7 @@ automatic_version_prefix: "29"
 repos:
   - fedora
   - fedora-updates
+  - fedora-updates-testing
   - dustymabe-ignition
 
 ignore-removed-users:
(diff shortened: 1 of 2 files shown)


dustymabe:
> So that we get the latest OSTree version which is needed to match the
> recent Ignition/OSTree initrd services re-ordering.

I want to make sure I understand. This is just to get the new ostree and in the future we can do something like use the url from the koji build as proposed in https://github.com/projectatomic/rpm-ostree/issues/1572 ?

dustymabe (approved):

jlebon:
Yeah, this is just to unbreak CentOS CI for now.

OK, this works and fixes it though I had to also switch off of mirrors since the update was just pushed. I don't mind dropping that last commit, though ISTM like it's fine to do this at this stage.

PR #26, opened 2018-11-05
base: add moby-engine

mike-nguyen:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/64

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -133,7 +133,7 @@ packages:
   # SSH
   - openssh-server openssh-clients
   # Containers
-  - podman skopeo runc
+  - podman skopeo runc moby-engine
   # Networking
   - bridge-utils nfs-utils biosdevname iptables-services
   - NetworkManager dnsmasq hostname


mike-nguyen:
I tested a compose using coreos-assembler with moby-engine added
- moby-engine is installed and docker daemon is not started by default
- podman works as expected
- docker pull and run ngninx works as expected

mskarbek:
I assume that this means default configuration for moby, no `journald` logging, `cgroupfs` as cgroup driver.

dustymabe:
@dm0- or @lsm5 do you know the answer to the question above ^^

mike-nguyen:
> does this mean docker is "socket activated" ? I think maybe we should not have it be socket activated but rather require it to be explicitly enabled in order to be used (that way someone couldn't accidentally start it by running a random docker command).

> If it is socket activated let's start an issue on the coreos-tracker repo to discuss.

The docker daemon is not started or enabled by default so the socket is not activated

lsm5:
@runcom @rhatdan ^

runcom:
> I assume that this means default configuration for moby, no `journald` logging, `cgroupfs` as cgroup driver. Can we have something more civilized than that? Or do we expect that everyone will have to reconfigure moby each time a new instance of FCOS will be created?

I don't really know what to assume, it depends on how we'll ship moby-engine. We could certainly provide better options for logging and cgroupfs so I guess it depends on us. Shipping moby-engine means shipping the docker/moby every people gets from download.docker.com so it's still up to people and distribution on how to best configure it on a given distro.

dustymabe:
> @mike-nguyen 
> The docker daemon is not started or enabled by default so the socket is not activated

cool, thanks mike for confirming.

dustymabe:
@mskarbek can you open an issue in the issue tracker for the "moby configuration" question and we can pick up the discussion there?

rhatdan:
The moby-engine daemon should not be configured or started by default.  We don't want to force people to use it, when better tools exist.  :^)

Moby-engine should be configured with what people consider the best way to run it.  If the packager wants to default to systemd that is fine with me.

PR #27, opened 2018-11-06
manifest: Use ${basearch}

cgwalters:
Like we do elsewhere.  It turns out multiple architectures may be
important...

Depends https://github.com/coreos/coreos-assembler/pull/192


jlebon:
LGTM!

dustymabe:
+1

PR #28, opened 2018-11-08
manifest: Remove biosdevname from package list

stellirin:
This package is only for x86_64 (also i686 if this architecture is
ever suported in the future) which causes non-x86_64 architectures
to fail the rpm-ostree compose.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -135,7 +135,7 @@ packages:
   # Containers
   - podman skopeo runc moby-engine
   # Networking
-  - bridge-utils nfs-utils biosdevname iptables-services
+  - bridge-utils nfs-utils iptables-services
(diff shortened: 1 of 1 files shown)


cgwalters:
Thanks for the patch!  Actually...I am not sure we should even be shipping `biosdevname` at all.  We don't in Fedora Atomic Host 29 today. 

I suspect this is simply a confusion/oversight on my part when creating this config initially in the very first commit 8e276a1fed80a77caf66750efd86e8f312764c64 - I probably adapted it from some other work which was oriented around RHEL7.

cgwalters:
Want to rework the PR to drop it?

stellirin:
Update to simply remove the package.

cgwalters:
Thanks!

dustymabe:
thanks @stellirin !

PR #31, opened 2018-11-20
Use metalink instead of baseurl in fedora repo

sinnykumari:
baseurl differs for primary and secondary arches
in Fedora

### DIFF
--- a/fedora.repo
+++ b/fedora.repo
@@ -1,8 +1,8 @@
 [fedora]
 name=Fedora $releasever - $basearch
 failovermethod=priority
-baseurl=https://dl.fedoraproject.org/pub/fedora/linux/releases/$releasever/Everything/$basearch/os/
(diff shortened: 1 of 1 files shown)


cgwalters:
This basically reverts  5a3cd6f98c711ab0f18eb2f036c85fe4f4be8061 right?

Which, I'm OK with that; probably we should use a COPR or whatever for moving faster instead.

sinnykumari:
> This basically reverts [5a3cd6f](https://github.com/coreos/fedora-coreos-config/commit/5a3cd6f98c711ab0f18eb2f036c85fe4f4be8061) right?

yeah, this time main reason is to allow downloading packages for arches like ppc64le as well whose location is https://dl.fedoraproject.org/pub/fedora-secondary instead of https://dl.fedoraproject.org/pub/fedora/linux/

> Which, I'm OK with that; probably we should use a COPR or whatever for moving faster instead.

Using COPR will have implication on aarch64 unless aarch64 gets added to COPR

dustymabe:
:+1:

PR #32, opened 2018-11-30
fedora-coreos-base: Add console-login-helper-messages package

rfairley:
Adds the rfairley-console-login-helper-messages COPR repo to download
the console-login-helper-messages package from.

---

Right now, this gets the console-login-helper-messages package and the required PAM version (`pam-1.3.1-12.fc29`) into the build when using coreos-assembler with the config on this branch. These two packages are sourced from COPR https://copr.fedorainfracloud.org/coprs/rfairley/console-login-helper-messages/monitor/.

There are 3 issues left to get motdgen fully working:

1) The main issue left to sort out is making sure the `motdgen` and `issuegen` units are enabled and started without needing to enter the commands manually. As of now, to use them requires the following when first booting in after `coreos-assembler run`:

```
# systemctl enable motdgen.service motdgen.path issuegen.service issuegen.path
# systemctl start motdgen.service issuegen.service
```
- [x] solved

2) The other blocker is getting `selinux-policy >= 3.14.3-11` in so that the generated `/run/motd.d/console-login-helper-messages.motd` can be displayed by `sshd`. For now the following is required to do the labeling:

```
echo "placeholder" > /run/motd
mkdir -p /run/motd.d
chcon -t etc_t /run/motd
chcon -t etc_t /run/motd.d
```

Looking into building `selinux-policy` in my COPR repo to use here, or using the rawhide version. Is it possible to specify the rawhide  `selinux-policy` in the `fedora-coreos-base.yml`? The rawhide version has the needed changes.

- [x] solved

3) One last issue is when using `coreos-assembler run`, `sshd` is not used so the generated motd under `/run` will not be displayed. To have `login` display it,  a config in `/etc/login.defs`  `MOTD_FILE=/etc/motd:/run/motd.d/console-login-helper-messages.motd` could be added so that the generated message gets displayed (http://man7.org/linux/man-pages/man1/login.1.html under `CONFIG FILE ITEMS`).

- [ ] solved (handling this in https://github.com/coreos/coreos-assembler/issues/244)

Other than that, `issuegen` works well to display the SSH keys and IP address from `eth0`, with `agetty` following a symlink in `/etc/issue.d` to read the generated file. The profile script showing the systemd failed units is also working. Just minor formatting edits are needed to the issue and profile scripts.

One possibility to immediately get the profile and issue messages in is to install the `console-login-helper-messages`, `console-login-helper-messages-issuegen`, `console-login-helper-messages-profile` packages (not `motdgen`) while the remaining issues are sorted out. Possibly do that, and follow up to add the `console-login-helper-messages-motdgen` subpackage in.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -15,6 +15,7 @@ repos:
   - fedora-updates
   - fedora-updates-testing
   - dustymabe-ignition
+  - rfairley-console-login-helper-messages
 
 ignore-removed-users:
   - root
@@ -119,6 +120,8 @@ postprocess:
     enable ignition-firstboot-complete.service
     enable coreos-growpart.service
     enable coreos-useradd-core.service
+    enable console-login-helper-messages-*.service
+    enable console-login-helper-messages-*.path
     EOF
 
     # Let's have a non-boring motd, just like CL (although theirs is more subdued
@@ -168,3 +171,8 @@ packages:
   - bash-completion
   # Moving files around
   - rsync fuse-sshfs
+  # User experience
+  - console-login-helper-messages
+  - console-login-helper-messages-motdgen
+  - console-login-helper-messages-issuegen
+  - console-login-helper-messages-profile
--- a/rfairley-console-login-helper-messages.repo
+++ b/rfairley-console-login-helper-messages.repo
@@ -1,0 +1,10 @@
+[rfairley-console-login-helper-messages]
+name=Copr repo for console-login-helper-messages owned by rfairley
+baseurl=https://copr-be.cloud.fedoraproject.org/results/rfairley/console-login-helper-messages/fedora-$releasever-$basearch/
+type=rpm-md
+skip_if_unavailable=True
+gpgcheck=1
+gpgkey=https://copr-be.cloud.fedoraproject.org/results/rfairley/console-login-helper-messages/pubkey.gpg
+repo_gpgcheck=0
+enabled=1
+enabled_metadata=1


dustymabe:
I can try to help get some answers to your questions here in a bit. Thanks robert for this!

dustymabe:
> There are 3 issues left to get motdgen fully working:

> The main issue left to sort out is making sure the `motdgen` and `issuegen` units are enabled and started without needing to enter the commands manually. As of now, to use them requires the following when first booting in after `coreos-assembler run`:


for now you can add them to [the list of presets we add in the manifest](https://github.com/coreos/fedora-coreos-config/blob/a733e2cfb39cfc55981ad1d37c359665aece7dd3/fedora-coreos-base.yaml#L116-L122). Longer term could you create an issue for us to get those into a proper place?



> The other blocker is getting `selinux-policy >= 3.14.3-11` in so that the generated `/run/motd.d/console-login-helper-messages.motd` can be displayed by `sshd`. For now the following is required to do the labeling:

It looks like the patches [were backported to f28/f29](https://github.com/fedora-selinux/selinux-policy/pull/232#issuecomment-443591593) but no rpm was built yet. Can you just create an srpm from the f29 branch and upload the srpm to your copr for now? 


>     1. One last issue is when using `coreos-assembler run`, `sshd` is not used so the generated motd under `/run` will not be displayed.

we can probably open a separate issue for `coreos-assembler run` and work on a fix after this PR is merged. WDYT?

rfairley:
Thanks Dusty for the advice! Will add the units to the list of presets, and will create the srpm in copr shortly.

> Longer term could you create an issue for us to get those into a proper place?

will write this up, should this issue be filed in `fedora-coreos-tracker`?

> open a separate issue for coreos-assembler run and work on a fix after this PR is merged. WDYT?

sounds good to me, will open this later today

dustymabe:
> will write this up, should this issue be filed in `fedora-coreos-tracker`?

I'd say either there or against fedora-coreos-config itself. If we think we'd like to get input from people on where it should go then maybe against tracker. If it's straightforward then maybe against config.

rfairley:
To note what the output will look like in FCOS upon login (using `coreos-assembler run`):

```
...
[  OK  ] Reached target Login Prompts.
[  OK  ] Reached target Multi-User System.
         Starting Update UTMP about System Runlevel Changes...
[  OK  ] Started Update UTMP about System Runlevel Changes.

Fedora 29 (CoreOS preview)
Kernel 4.19.6-300.fc29.x86_64 on an x86_64 (ttyS0)

This is coreos (Linux x86_64 4.19.6-300.fc29.x86_64) 00:21:28
SSH host key: SHA256:SUkswm9XKBpC756or5BUjyT4hinOMwqdakI8ei5v9nQ (ECDSA)
SSH host key: SHA256:Oc66LDhBNf9sK+jSQb4e7IdD7ylXAIcXGAj2WDToGOA (ED25519)
SSH host key: SHA256:JgLkrfHJdVUl2aUGNIrQhtCZZ1Ek4DLk7f//2BUut+4 (RSA)
eth0: 10.0.2.15 fec0::5054:ff:fe12:3456

coreos login: core (automatic login)

Fedora CoreOS (preview)
https://github.com/coreos/fedora-coreos-tracker
WARNING: All aspects subject to change, highly experimental
---

[core@coreos ~]$ 
```

The motd gets generated at the following location:
```
[core@coreos ~]$ cat /run/motd.d/console-login-helper-messages.motd
Fedora (29 (CoreOS preview))


```

The fields in the generated motd are sourced from `/etc/os-release` (like how CL did it). Once the edits to the `login` config to show the generated motd are made, we can look at changing where the static motd information at https://github.com/coreos/fedora-coreos-config/blob/master/fedora-coreos-base.yaml#L128 will go. For now this message will be generated but silent. It is safe to pull in, and displaying it on login can be dealt with in another issue.

One more note: there is an `[ostree]` section with information on package updates to show as a login. However, formatting is a bit complicated and breaking at the moment. I took it out of `console-login-helper-messages-motdgen` for now, the issue to handle it is https://github.com/rfairley/console-login-helper-messages/issues/16.

dustymabe:
will review today

dustymabe:
looks like it's working!

on serial console:

```
Fedora 29 (CoreOS preview)                                                  
Kernel 4.19.6-300.fc29.x86_64 on an x86_64 (ttyS0)                          
                                                                            
This is localhost.localdomain (Linux x86_64 4.19.6-300.fc29.x86_64) 21:46:07
SSH host key: SHA256:qWpm8OfAiQXgVmBjOv2a0jNOcFZpWpy1bFslkuII8Uc (ED25519)  
SSH host key: SHA256:m7sTSlhJFXvQj9R/dtr4jc5xLjLlpnqg8+OWH9cOdjY (ECDSA)    
SSH host key: SHA256:X5k62CzK5RIlMVWUdqSjTea1ovfSr1ikzOQXnTPV9n0 (RSA)      
eth0: 192.168.122.213 fe80::5054:ff:fef2:db47

```

on ssh login:
```
Fedora CoreOS (preview)                                    
https://github.com/coreos/fedora-coreos-tracker            
WARNING: All aspects subject to change, highly experimental
---                                                        
                                                           
[systemd]                                                  
Failed Units: 1                                            
  example.service                                          

```

dustymabe:
i'll merge this.. one thing that is interesting is that I can't `systemctl status` the service files after boot. I think we need to add `RemainAfterExit=yes` in the unit files. See https://github.com/coreos/ignition-dracut/pull/15 where we did this somewhere else.

rfairley:
Thanks for linking to that PR, looks like that's it. I'll add that in, https://github.com/rfairley/console-login-helper-messages/issues/17.

Thank you Dusty for reviewing!

cgwalters:
I consistently see:

```
Running pre scripts... console-login-helper-messages
console-login-helper-messages.prein: /usr/console-login-helper-messages.prein: line 9:     3 Aborted                 (core dumped) systemd-tmpfiles --replace=/usr/lib/tmpfiles.d/console-login-helper-messages-isF
console-login-helper-messages.prein:  
console-login-helper-messages.prein: SYSTEMD_INLINE_EOF 
console-login-helper-messages.prein: 
console-login-helper-messages.prein: 
console-login-helper-messages.prein: systemd-tmpfiles --replace=/usr/lib/tmpfiles.d/console-login-helper-messages-profile.conf --create - <<SYSTEMD_INLINE_EOF &>/dev/null || : 
console-login-helper-messages.prein:  
console-login-helper-messages.prein: SYSTEMD_INLINE_EOF
```

rfairley:
Reproduced - it looks like it is something to do with the tmpfile symlink desination not existing - I had renamed the script and didn't update this file https://github.com/rfairley/console-login-helper-messages/commit/6083589d8cdc78f0482e9bba3266c8a6f72c1f76. The symlink still got created successfully despite the tmpfile being wrong as the specfile has it installed https://github.com/rfairley/console-login-helper-messages/blob/master/console-login-helper-messages.spec#L88.

Just gave the console-login-helper-messages RPM another build - hopefully fixes it.

rfairley:
The [%tmpfiles_create_package macro](https://github.com/systemd/systemd/blob/master/src/core/macros.systemd.in#L133) was where this was coming from. I have removed it from the specfile as it is not needed (has the needed effect of recreating the symlink if broken on `systemd-tmpfiles --create`) and now follows the [example](https://fedoraproject.org/wiki/Packaging:Tmpfiles.d#Example_spec_file) (https://github.com/rfairley/console-login-helper-messages/commit/704f915ea8af79051269f600682467ffd6f1e98f).

rfairley:
@dustymabe @cgwalters do you think it'd be worth filing a bug report for the core dump? If it's a segmentation fault, that seems like something that shouldn't happen even if the macro was incorrectly used.

I'm not sure of the best way to approach getting more information on the core dump before submitting a bug report. I couldn't get the same `prein` error to show in a `fedora/29-cloud-base` vagrant VM after installing as follows:

```
dnf install -v --rpmverbosity DEBUG console-login-helper-messages
```

Will leave it at that for now as the issue we were seeing is fixed, but would be good to report this as well if it's a potential bug on the `systemd-tmpfiles` side.

dustymabe:
> @dustymabe @cgwalters do you think it'd be worth filing a bug report for the core dump? If it's a segmentation fault, that seems like something that shouldn't happen even if the macro was incorrectly used.

agreed. probably would be good to file an issue. 

> 
> I'm not sure of the best way to approach getting more information on the core dump 

can you grab the core dump file and attach it to the bug report? Check to see if there is one in `/var/lib/systemd/coredump/` on the system where the abort happened.

rfairley:
Thanks Dusty! Got the coredump file in that location and am writing up the issue.

rfairley:
Issue opened: https://github.com/systemd/systemd/issues/11124

PR #34, opened 2018-12-14
manifest: add --no-hostonly and VMWare storage drivers to initramfs

imcleod:


jlebon:
Looks sane to me. Were you able to build with this fine?

fedora-coreos-base.yaml
```diff
@@ -5,6 +5,12 @@ include: minimal.yaml
 boot_location: new
 tmp-is-dir: true
 
+# Add a richer set of early boot drivers, including vmware storage
+initramfs-args:
+  - "--no-hostonly"
```
dustymabe: weird - were we not doing `--no-hostonly` before in FCOS?

dustymabe:
one question I asked in another place: should we try to get this change into dracut itself upstream rather than just carrying it here?

fedora-coreos-base.yaml
imcleod: It's in Fedora Atomic Host I believe:

https://pagure.io/fedora-atomic/blob/master/f/fedora-atomic-host.json#_14

May have gotten lost in the shuffle.

imcleod:
> one question I asked in another place: should we try to get this change into dracut itself upstream rather than just carrying it here?

Happy to pursue it there but I expect that will be a bit slower of a change to trickle out.  I appreciate "let's just patch it here" is not a great argument....

dustymabe:
> Happy to pursue it there but I expect that will be a bit slower of a change to trickle out. I appreciate "let's just patch it here" is not a great argument....

How about a compromise? We can merge this here (quick fix), but can we open a PR there to see if we can make progress on the "long term" solution too?

fedora-coreos-base.yaml
jlebon: It's the default in rpm-ostree, so we were still using it in FCOS: https://github.com/projectatomic/rpm-ostree/blob/b889a510560e0866fac94f7de06c258546c23ce7/src/libpriv/rpmostree-postprocess.c#L419

imcleod:
> > Happy to pursue it there but I expect that will be a bit slower of a change to trickle out. I appreciate "let's just patch it here" is not a great argument....
> 
> How about a compromise? We can merge this here (quick fix), but can we open a PR there to see if we can make progress on the "long term" solution too?

SGTM - I'm perusing the dracut code at the moment

cgwalters:
LGTM

dustymabe:
> SGTM - I'm perusing the dracut code at the moment

+1 - running this locally now to make sure things don't blow up.

PR #35, opened 2018-12-17
manifest: Remove redundant module includes

imcleod:
Further testing reveals that, for F28/F29 at least, the mptspi
and VMWare PV SCSI driver are already included by dracut when
you specify --no-hostonly.  Removing them here as they are not
actually needed.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -5,12 +5,6 @@ include: minimal.yaml
 boot_location: new
 tmp-is-dir: true
 
-# Add a richer set of early boot drivers, including vmware storage
-initramfs-args:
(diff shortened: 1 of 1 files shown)


fedora-coreos-base.yaml
```diff
@@ -8,8 +8,6 @@ tmp-is-dir: true
 # Add a richer set of early boot drivers, including vmware storage
 initramfs-args:
   - "--no-hostonly"
```
cgwalters: OK, but then we can just drop this too, as rpm-ostree already does it by default.

imcleod: Dropped

PR #36, opened 2018-12-17
base: Drop grub2 auto-hide bits

LorbusChris:
Obsoletes #21


LorbusChris:
I just noticed, we might be ok with #21 if we switch to using the f30 version of grub2 rebuilt for f29.
Created a COPR for that today: https://copr.fedorainfracloud.org/coprs/lorbus/grub2/builds/

dustymabe:
> I just noticed, we might be ok with #21 if we switch to using the f30 version of grub2 rebuilt for f29.
> Created a COPR for that today: https://copr.fedorainfracloud.org/coprs/lorbus/grub2/builds/

do you just want to add that copr to the input? We've done that for a couple of things already. You can just add a comment that says this can be removed when we move to f30

LorbusChris:
Yes, that sounds good. I'll rework this PR to do that.

LorbusChris:
@dustymabe done.

LorbusChris:
forgot to check in the repo file. ready now. please take another look.

dustymabe:
i think you need to rebase - robert also added a copr recently and your change is conflicting with his.

LorbusChris:
resolved :)

dustymabe:
LGTM - @cgwalters - do you want to give it a look over?

cgwalters:
Seems sane to me!

dustymabe:
looks like this broke some things:

```
Executing `postprocess` inline script '0'
+ for x in /etc/yum.repos.d/*modular.repo
+ sed -i -e 's,enabled=[01],enabled=0,' /etc/yum.repos.d/fedora-modular.repo
+ for x in /etc/yum.repos.d/*modular.repo
+ sed -i -e 's,enabled=[01],enabled=0,' /etc/yum.repos.d/fedora-updates-modular.repo
+ for x in /etc/yum.repos.d/*modular.repo
+ sed -i -e 's,enabled=[01],enabled=0,' /etc/yum.repos.d/fedora-updates-testing-modular.repo
+ rm -v /etc/grub.d/01_fallback_counting /etc/grub.d/01_menu_auto_hide
rm: cannot remove '/etc/grub.d/01_fallback_counting': No such file or directory
removed '/etc/grub.d/01_menu_auto_hide'
[31m[1merror: [22m[0mPostprocessing: While executing inline postprocessing script '0': Executing bwrap(/usr/bin/rpmostree-postprocess-inline-0): Child process killed by signal 1
```

When I do the compose locally for some reason the rpms aren't getting pulled in from the copr. It's choosing to pull in `grub2-common-2.02-62.fc29.noarch` from the fedora repos rather than the version from the copr. There must be a dependency that is missing from the copr that is causing the older version from the fedora repos to get chosen. One thing I notice in the copr is that grub2-pc-modules is not one of the subpackages that is built. Might be somewhere to look. 

cc @LorbusChris

LorbusChris:
Ok. I see no easy fix for grub2 on COPR. let's remove the fallback script from rm here, and use go back to using the f29 version from the main repos.

PR #39, opened 2018-12-26
WIP: image: Enable read-only /boot and /sysroot

cgwalters:
Part of the implementation of https://github.com/ostreedev/ostree/issues/1265

Convert to systemd mount units so we can easily specify the `ro` flag,
and also to use the `LABEL={root,var}` rather than UUIDs.

Enable the sysroot/readonly flag.

### DIFF
--- a/image.ks
+++ b/image.ks
@@ -70,4 +70,35 @@ rm -f /var/lib/random-seed
 
 echo "Removing /root/anaconda-ks.cfg"
 rm -f /root/anaconda-ks.cfg
+
+# Replace /etc/fstab with mount units; mostly so we can easily enable the "ro"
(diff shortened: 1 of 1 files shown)


dustymabe:
> Convert to systemd mount units so we can easily specify the `ro` flag,
> and also to use the `LABEL={root,var}` rather than UUIDs.

I think you mean `LABEL={root,boot}` ?

cgwalters:
This is still useful/relevant no?

jlebon:
Yeah, so I think what happened here is that when I switched over the default branch to `testing-devel` and deleted `master`, it closed off all the pending PRs against `master`. Can you reopen it against `testing-devel`?

PR #40, opened 2019-01-10
fcos-base: add utils for static firewalling

lucab:
This adds utils and service units for loading static firewalling rules,
supporting both iptables and nftables flavors.
Ref: https://github.com/coreos/fedora-coreos-tracker/pull/103

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -159,8 +159,10 @@ packages:
   # Containers
   - podman skopeo runc moby-engine
   # Networking
-  - bridge-utils nfs-utils iptables-services
+  - bridge-utils nfs-utils
   - NetworkManager dnsmasq hostname
+  # Static firewalling
+  - iptables nftables iptables-nft iptables-services
   # Storage
   - cloud-utils-growpart
   - lvm2 iscsi-initiator-utils sg3_utils


lucab:
I cross-checked packages name/content on F29, but I didn't run this through a composer run.

/cc @dustymabe

dustymabe:
i'll run through an assembler run :) - thanks luca

dustymabe:
ok so with this change to the configs the change that was applied to the ostree was this:

```
[dustymabe@media repo]$ rpm-ostree --repo=./ db diff fedora/29/x86_64/coreos^ fedora/29/x86_64/coreos
ostree diff commit old: fedora/29/x86_64/coreos^ (95c547351440987a8c28e7671f35ca41dae0639853f7d39c243dd25418134918)
ostree diff commit new: fedora/29/x86_64/coreos (4f4db64f70730c870460328350e75821738ca2bb2680d9f241ce33f25cdda578)
Removed:
  iptables-services-1.8.0-3.fc29.x86_64
Added:
  iptables-nft-1.8.0-3.fc29.x86_64
```

fedora-coreos-base.yaml
```diff
@@ -159,8 +159,10 @@ packages:
   # Containers
   - podman skopeo runc moby-engine
   # Networking
-  - bridge-utils nfs-utils iptables-services
```
dustymabe: just checking: did we intend to remove `iptables-services` ?

dustymabe: here are the files provided by that rpm:

```
[root@coreos ~]# rpm -ql iptables-services
/etc/sysconfig/ip6tables
/etc/sysconfig/iptables
/usr/lib/systemd/system/ip6tables.service
/usr/lib/systemd/system/iptables.service
/usr/libexec/initscripts/legacy-actions/ip6tables
/usr/libexec/initscripts/legacy-actions/ip6tables/panic
/usr/libexec/initscripts/legacy-actions/ip6tables/save
/usr/libexec/initscripts/legacy-actions/iptables
/usr/libexec/initscripts/legacy-actions/iptables/panic
/usr/libexec/initscripts/legacy-actions/iptables/save
/usr/libexec/iptables
/usr/libexec/iptables/ip6tables.init
/usr/libexec/iptables/iptables.init
```

lucab: Yes, the removal was on purpose. If I'm not mistaken those are initrc compatibility bits. My cross-reference was https://fedora.pkgs.org/29/fedora-x86_64/iptables-services-1.8.0-3.fc29.x86_64.rpm.html

dustymabe: - user story: as a user of FCOS I want to allow port 8000/tcp. 
- implementation: write an ignition config that writes an iptables configuration to  `/etc/sysconfig/iptables` and enabled the `iptables.service`. The firewall gets applied by the `iptables.service` systemd unit on boot. 

without the `iptables.service` systemd unit, how does that user story get executed?

dustymabe: > If I'm not mistaken those are initrc compatibility bits

meant to say this in the last response, the rpm also includes `iptables.service` systemd unit, so not just initrc compat??

dustymabe:
is it safe to say we still need the `iptables-services` rpm?

fedora-coreos-base.yaml
lucab: We are speaking about https://src.fedoraproject.org/rpms/iptables/blob/f29/f/iptables.service and https://src.fedoraproject.org/rpms/iptables/blob/f29/f/iptables.init here. It is a systemd unit calling the initrc script.

lucab:
I've added back the `iptables-services` package. Even though I still believe it is (two layers) of initrc compat, I don't have enough knowledge about compat layers in fedora. For reference, ContainerLinux has dedicated restoring units which are straightforward:  https://gitweb.gentoo.org/repo/gentoo.git/tree/net-firewall/iptables/files/systemd/iptables-restore.service. However, I fear fedora does not currently provide those.

dustymabe:
> However, I fear fedora does not currently provide those.

yes, I have the same concern.. Could you open a BZ against iptables-services rpm and start a discussion ?

lucab:
@dustymabe ack, reported https://bugzilla.redhat.com/show_bug.cgi?id=1667875 for clarification/RFE.

dustymabe:
thanks.. finally got around to testing this.. LGTM

PR #46, opened 2019-01-27
Add image.yaml

cgwalters:
Let's just support configuring the base disk size for now so that we
say we *do* have some configuration.  We need to abstract this for
multiple reasons (we want to remove Anaconda, and even without removing
Anaconda we want to do dual bios/UEFI and also potentially arch-specific
code).

Most people will probably be OK for now configuring the base ostree and the
outer image size.

This commit doesn't yet remove image.ks in order to support version skew,
but we intend to shortly drop it.

### DIFF
--- a/image.ks
+++ b/image.ks
@@ -1,3 +1,7 @@
+# THIS FILE IS DEPRECATED, use image.yaml
+#
+# original contents below:
+#
 # Currently the coreos-assembler tool is generating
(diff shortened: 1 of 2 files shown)


cgwalters:
See also https://github.com/coreos/coreos-assembler/pull/298

dustymabe:
tested in conjuction with coreos/coreos-assembler#298 - seems to work.

PR #47, opened 2019-02-05
add isolinux files for installer work

dustymabe:


dustymabe:
this is one part of https://github.com/coreos/coreos-assembler/pull/306

dustymabe:
ok I removed the "hack for fast development" commits from this and am moving from WIP. Let's go ahead and get this reviewed/merged and we'll be one step closer to delivering on this.

fedora-coreos-base.yaml
```diff
@@ -209,3 +209,6 @@ packages:
   - console-login-helper-messages-motdgen
   - console-login-helper-messages-issuegen
   - console-login-helper-messages-profile
+  # needed for installer for now
+  - bc
+  - dialog
```
jlebon: What actually requires these?

dustymabe: [the installer](https://github.com/coreos/coreos-installer/blob/master/dracut-module/module-setup.sh#L18-L31) - I guess theoretically this can be handled by rpm deps when we have an rpm for the installer (which should be soon: https://github.com/coreos/coreos-installer/issues/1)

jlebon: Gotcha. How about a comment like "needed by installer dracut module for now" ?

dustymabe: since these will be handled by the rpm I'll just remove them and add the rpm as a dep and we'll wait until the rpm is available to merge this.

dustymabe:
moving back to WIP until we have the rpm available

dustymabe:
ok back out of WIP - we have a copr repo for now with the installer rpm while it goes through code review

fedora-coreos-base.yaml
```diff
@@ -14,6 +14,7 @@ repos:
   - fedora
   - fedora-updates
   - dustymabe-ignition
+  - dustymabe-coreos-installer
```
cgwalters: (Not sure it's really worth having multiple COPRs, easier to have a shared one?)

isolinux/boot.msg
```diff
@@ -0,0 +1,5 @@
+
```
cgwalters: What reads these files?

jlebon:
Looks good to me!
Will let @yrobla take a look as well.

fedora-coreos-base.yaml
dustymabe: yeah i thought about that. we'll rip these both out pretty soon (#52) and if we need to add a copr back we'll make a combined one

isolinux/boot.msg
dustymabe: they are picked up by `cosa buildextend-installer` and placed into the target ISO.

cgwalters: Maybe put them in an `installer/` subdirectory with a `README.md` that just says that?

dustymabe: created an installer dir and a readme in a follow up commit

cgwalters:
Looks great thanks!

dustymabe:
In the final commit I renamed `isolinux` to `installer`. Have opened a corresponding PR against COSA: https://github.com/coreos/coreos-assembler/pull/352

PR #48, opened 2019-02-06
base: Switch console-login-helper-messages to come from updates-testing

rfairley:
Now that console-login-helper-messages is built as an update for Fedora
29, and is available in fedora-updates-testing, let's remove the COPR
repo it came from before.

---

Opening this to drop the COPR repo - I have run through a `coreos-assembler fetch` and console-login-helper-messages is sourced from fedora-updates-testing.

Wondering if there's any reason not to do this right now? I think the COPR repo won't be needed for quick testing as any quick adjustments can be tested by installing console-login-helper-messages from the command line (after downloading the RPM from a Koji build).


dustymabe:
> Wondering if there's any reason not to do this right now?

i don't think we have the updates-testing repo enabled by default do we? If not then we can just wait and merge this when it hits stable.

jlebon:
We do enable fedora-updates-testing today: https://github.com/coreos/fedora-coreos-config/blob/bb5d120b9ee3a82d97c1fcd6b223c5375a7bdbdb/fedora-coreos-base.yaml#L16. So this could go in now. Though I just tested it and was hitting some issues (or I'm likely not using it properly :))

https://bodhi.fedoraproject.org/updates/FEDORA-2019-a19feded91#comment-892031

dustymabe:
> We do enable fedora-updates-testing today

yeah we should probably disable that

rfairley:
+1 on waiting and merging once in stable, if the long term goal is to disable fedora-updates-testing

dustymabe:
> +1 on waiting and merging once in stable, if the long term goal is to disable fedora-updates-testing

https://github.com/coreos/fedora-coreos-config/pull/49

rfairley:
Looks to be in updates: https://dl.fedoraproject.org/pub/fedora/linux/updates/29/Everything/x86_64/Packages/c/

PR #50, opened 2019-02-12
base: add gnupg2 package

dustymabe:
Provides /usr/bin/gpg binary for verifying/signing/decrypting/encrypting
secret data.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -202,8 +202,9 @@ packages:
   - sudo coreutils less tar xz gzip bzip2 tmux
   - nmap-ncat net-tools bind-utils
   - bash-completion
-  # Moving files around
+  # Moving files around and verifying them
   - rsync fuse-sshfs
+  - gnupg2
   # User experience
   - console-login-helper-messages
   - console-login-helper-messages-motdgen


fedora-coreos-base.yaml
```diff
@@ -171,6 +171,7 @@ postprocess:
     EOF
 
 packages:
+  - gnupg
```
dustymabe: what "category" should I put this under?

jlebon: Q: as someone who's not intimate with the details, should this be gnupg2, or alternatively should we ship both?

> what "category" should I put this under?

Maybe just in "Used by admins interactively" ?

dustymabe:
> Q: as someone who's not intimate with the details, should this be gnupg2, or alternatively should we ship both?

It should be gnugp2. I thought I had copied atomic host but I didn't do it right:

https://pagure.io/fedora-atomic/blob/master/f/fedora-atomic-host.json#_81

fixed. 

> 
> > what "category" should I put this under?
> 
> Maybe just in "Used by admins interactively" ?

Yeah. I didn't want it to be looked at as a minimization candidate, though. I think this is useful for downloading and verifying things in automation too. One example might be if we drop an ignition config that drops down a unit that says to download a file and then verify it was signed by XYZ key. I added it to the moving files around section.

jlebon:
Commit message needs to be updated, otherwise LGTM!

dustymabe:
> Commit message needs to be updated, otherwise LGTM!

👍

PR #51, opened 2019-02-13
Drop image.ks

cgwalters:
Let's assume most people have an updated cosa now.  Having
the `image.ks` in the repo can lead to confusion as to why
it's not being used.


dustymabe:
👍

PR #52, opened 2019-02-14
Remove x86-only repo

r4f4:
It fails on aarch64 since the repo only contains x86_64 packages. Use
the default fedora repos instead.

Signed-off-by: Rafael Fonseca <r4f4rfs@gmail.com>


dustymabe:
LGTM. discussed with @r4f4. Will merge next week. 

For right now I'm going to keep the copr because I want to fast track a change I'm making that will allow us to read ignition configs from `/boot/ignition/config.ign`. I'll get this in shortly after that.

dustymabe (approved):

dustymabe:
@jlebon switched over the default branch to `testing-devel` and deleted `master`. This action closed off all the pending PRs against master. If still relevant can you re-open and edit your PR to target `testing-devel`?

PR #54, opened 2019-02-20
WIP: Drop /etc/sysconfig and NetworkManager ifcfg-rh plugin

cgwalters:
This is Red Hat Linux era legacy.  For now...people can write
network configs using the keyfile plugin.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -37,6 +37,11 @@ postprocess:
     #!/usr/bin/env bash
     set -xeuo pipefail
 
+    # Nuke this Red Hat Linux-era legacy
+    rm /etc/sysconfig -rf
+    # And drop support for the initscripts network format that lives in there
+    rm -v /usr/lib64/NetworkManager/*/libnm-settings-plugin-ifcfg-rh.so
+
     # https://github.com/projectatomic/rpm-ostree/issues/1542#issuecomment-419684977
     for x in /etc/yum.repos.d/*modular.repo; do
       sed -i -e 's,enabled=[01],enabled=0,' ${x}


cgwalters:
I tested this both builds and boots, but not much beyond that.

ashcrow:
Deferring to @jlebon and @dustymabe

ajeddeloh:
Any reason to not use the treefile's `remove-files` or `remove-from-packages`?

fedora-coreos-base.yaml
```diff
@@ -37,6 +37,11 @@ postprocess:
     #!/usr/bin/env bash
     set -xeuo pipefail
 
+    # Nuke this Red Hat Linux-era legacy
+    rm /etc/sysconfig -rf
```
jlebon: That's one way to solve it... :)

Though it seems like we should carefully go through the list of files in there and work with packagers/upstreams as needed, no? (E.g. if there's a file in there that's not a Red Hat-ism, but straight from upstream, then we should keep it, right? Not sure if that's presently the case though.)

Oh, or did you mean `/etc/sysconfig/network*` here?

ajeddeloh: Oh man I misread that, thought it was `/etc/sysconfig/network`. If we _do_ want to remove `/etc/sysconfig` I think we ought to do it via a list of files, so if packagers/upstreams add them we see it before removing it.

LorbusChris:
```
[chris@chrishost api]$ rpm -q --whatprovides /etc/sysconfig/*
file /etc/sysconfig/anaconda is not owned by any package
at-3.1.23-1.fc29.x86_64
chrony-3.4-1.fc29.x86_64
initscripts-10.01-1.fc29.x86_64
cronie-1.5.2-3.fc29.x86_64
ebtables-2.0.10-28.fc29.x86_64
firewalld-0.6.3-1.fc29.noarch
grub2-tools-extra-2.02-62.fc29.x86_64
httpd-2.4.38-2.fc29.x86_64
iptables-1.8.0-3.fc29.x86_64
iptables-1.8.0-3.fc29.x86_64
libvirt-daemon-4.7.0-1.fc29.x86_64
libvirt-client-4.7.0-1.fc29.x86_64
man-db-2.8.4-3.fc29.x86_64
initscripts-10.01-1.fc29.x86_64
file /etc/sysconfig/network is not owned by any package
NetworkManager-1.12.6-5.fc29.x86_64
nfs-utils-2.3.3-1.rc2.fc29.x86_64
nftables-0.9.0-2.fc29.x86_64
qemu-guest-agent-3.0.0-3.fc29.x86_64
radvd-2.17-17.fc29.x86_64
mdadm-4.1-rc2.0.2.fc29.x86_64
rpcbind-1.2.5-0.fc29.x86_64
crontabs-1.11-17.20150630git.fc29.noarch
samba-common-4.9.4-1.fc29.noarch
cyrus-sasl-2.1.27-0.3rc7.fc29.x86_64
selinux-policy-3.14.2-49.fc29.noarch
sheepdog-1.0.1-7.fc29.x86_64
openssh-server-7.9p1-4.fc29.x86_64
subversion-1.11.1-1.fc29.x86_64
libvirt-daemon-4.7.0-1.fc29.x86_64
libvirt-daemon-4.7.0-1.fc29.x86_64
wpa_supplicant-2.7-1.fc29.x86_64
zfs-fuse-0.7.2.2-6.fc27.x86_64
```

ajeddeloh:
Is there a BZ ticket or similar to get this split out from the NM packages anywhere?

mrguitar:
careful this change might diverge from RHEL/RHCOS.

bgilbert:
@mrguitar Yep, it's good to be aware of when we're doing that.  In this case, I think FCOS has a great opportunity to leave behind a painful legacy format, even if RHEL and RHCOS don't have that flexibility.  If the plugin's absence turns out to be a huge problem, we can always re-enable it later.

dustymabe:
> Deferring to @jlebon and @dustymabe

I agree with @jlebon  on this point: https://github.com/coreos/fedora-coreos-config/pull/54#discussion_r258685329

will follow the discussion there

dustymabe:
one thing to note is that it looks like `/etc/sysconfig/network*` ended up getting removed in https://github.com/coreos/coreos-assembler/pull/497

bgilbert:
https://github.com/coreos/fedora-coreos-config/pull/98 drops the ifcfg-rh plugin.  Should we close in favor of that PR?

bgilbert:
Closing in favor of #98.

PR #56, opened 2019-02-21
manifest: Add static sysusers entries

cgwalters:
For use with https://github.com/projectatomic/rpm-ostree/pull/1679


fedora-coreos-base.yaml
```diff
@@ -7,6 +7,16 @@ tmp-is-dir: true
 
 # Required by Ignition, and makes the system not compatible with Anaconda
 machineid-compat: false
+# New model
+sysusers: true
```
jlebon: Hmm, how this intersects with Ignition is interesting. Ignition allows you to specify users/groups by name for each file/dir/link. I think this means that we'd have to run sysusers too in the `initrd` on first boot. That would also allow us to just do a blanket `systemd-tmpfiles --create --boot --root=/sysroot --prefix=/var`.

cgwalters: Yep see https://github.com/coreos/fedora-coreos-tracker/issues/155

cgwalters: Where's the code that runs `systemd-tmpfiles` for Ignition today?

jlebon: Today, nowhere. :)
Soon: https://github.com/coreos/ignition-dracut/pull/47

dustymabe:
@jlebon switched over the default branch to `testing-devel` and deleted `master`. This action closed off all the pending PRs against master. If still relevant can you re-open and edit your PR to target `testing-devel`?

PR #57, opened 2019-02-22
manifest: Use oneshot for custom services

jlebon:
That way we make sure those invocations completely finish before
starting sshd. I haven't seen this race happen in practice. Just noticed
it.


dustymabe:
LGTM - +1 for merge if you've tested

bgilbert:
LGTM

PR #59, opened 2019-03-01
installer: add grub.cfg for EFI boot from ISO

dustymabe:
- installer: now using an images and isolinux sub dir
- installer: add grub.cfg for EFI boot from ISO


installer/EFI/fedora/grub.cfg
```diff
@@ -0,0 +1,29 @@
...
+set gfxpayload=keep
+insmod gzio
+insmod part_gpt
+insmod ext2
+
+set timeout=60
+### END /etc/grub.d/00_header ###
+
+### BEGIN /etc/grub.d/10_linux ###
+menuentry 'Install Fedora CoreOS' --class fedora --class gnu-linux --class gnu --class os {
```
cgwalters: Wondering whether we should instead ship this with cosa and just substitute the name from the manifest.

cgwalters: (This applies to the other stuff already here in `installer/` of course...not a blocking issue but let's at least consider it?)

dustymabe: I personally don't think so, but I think it's reasonable to have that conversation in a ticket (or meeting) dedicated for that discussion to get input from others.

cgwalters: OK we can change it later.

PR #60, opened 2019-03-04
manifest: Omit nfs dracut module

jlebon:
The nfs dracut module is needed to support root-on-NFS. But this is
something we don't support for FCOS, so let's just disable it. This also
works around issues with supporting separate `/var` filesystems and the
`rpc_pipefs` mount that the module creates.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -5,6 +5,13 @@ include: minimal.yaml
 boot_location: new
 tmp-is-dir: true
 
+initramfs-args:
+  - --no-hostonly
+  # We don't support root on NFS, so we don't need it in the initramfs. It also
+  # conflicts with /var mount support.
+  - --omit
+  - nfs
+
 # Required by Ignition, and makes the system not compatible with Anaconda
 machineid-compat: false
 


dustymabe:
I would move the comment down to right above the nfs part since there will probably be more of these and we should comment them individually

jlebon:
Done!

dustymabe:
looks good to me - i don't really know the implications so if we hit problems we can always revert. One thing that would be good is to maybe explain more about `issues with supporting separate /var filesystems` so that if we do revert in the future we have more context. Is there an open issue or a discussion where someone could pick up some context that we should link?

dustymabe (approved):

ajeddeloh (approved):
:100: LGTM
We actually ought to go through and also purge all the other modules we're not using, but that can be another PR.

jlebon:
> One thing that would be good is to maybe explain more about issues with supporting separate /var filesystems so that if we do revert in the future we have more context.

Sorry, meant to get back to this. Will just write it out here for posterity: the specific issue is that Ignition wants to be able to mount stuff and then unmount them on first boot. NFS conflicts with that by also wanting to mount stuff in `/var`, so then Ignition can't cleanly unmount.

dustymabe:
> NFS conflicts with that by also wanting to mount stuff in `/var`

what causes NFS to do that? what does it mount? Is it user specified (fstab)?

jlebon:
A random thought I had while working on this was that it'd be cool if there was a clean easy way to share a private mount namespace just for the Ignition stuff across multiple units. There's `JoinsNamespaceOf`, though actually that only allows sharing the same private `/var/tmp` and `/tmp`.

But anyway, we should strive to only have `/sysroot` mounted at switchroot and nothing under that, so that the full specification of how everything is mounted is subject to be configured through Ignition.

jlebon:
> what causes NFS to do that? what does it mount? Is it user specified (fstab)?

It always mounts it if the kernel supports it. See https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfs-start-rpc.sh and https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfsroot-cleanup.sh.

PR #61, opened 2019-03-05
manifest: add comments for removing nfs from dracut

dustymabe:
This will help us choose to keep or remove it in the future.


dustymabe:
cc @jlebon to check for accuracy of statements

jlebon:
Sure. Could also just have a one-liner that links to https://github.com/coreos/fedora-coreos-config/pull/60?

dustymabe:
> Sure. Could also just have a one-liner that links to #60?

I often prefer to summarize the discussion because often linked to discussions can be lengthy, but you are right that we should also link to the original discussion. I added that.

jlebon:
SGTM!

PR #62, opened 2019-03-05
dracut: omit modules we don't use

ajeddeloh:
This should help build a smaller initramfs and clean up the logs a bit

Still testing, will make non-draft once tested

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -12,8 +12,11 @@ initramfs-args:
   # in /var/ and then ignition can't cleanly unmount it. For example:
   # https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfs-start-rpc.sh#L7
   # See also discussion in https://github.com/coreos/fedora-coreos-config/pull/60
-  - --omit
-  - nfs
+  - --omit=nfs
+  # Omit these since we don't use them
+  - --omit=lvm
+  - --omit=multipath
+  - --omit=iscsi
 
 # Required by Ignition, and makes the system not compatible with Anaconda
 machineid-compat: false


ajeddeloh:
Ready for review

fedora-coreos-base.yaml
```diff
@@ -13,7 +13,7 @@ initramfs-args:
   # https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfs-start-rpc.sh#L7
   # See also discussion in https://github.com/coreos/fedora-coreos-config/pull/60
   - --omit
-  - nfs
```
dustymabe: would be nice to have a separate `--omit arg` for each of these with comments above each one I think. I know you have multiple `--omit` args passed to dracut. I wonder if `--omit=arg` works. If it does then we can just have 4 lines.

ajeddeloh:
turns out dracut does support `--omit=<modname>`. Pushed an update.

dustymabe:
> turns out dracut does support `--omit=<modname>`. Pushed an update.

perfect.. now we have a good message for nfs and why we don't need it. anything for each of the others? Doesn't need to be that verbose, of course.

ajeddeloh:
Added message

cgwalters:
I'm OK with this for now, but we do want to support e.g. `/var` being LVM (or Stratis) in the future right?  And probably iSCSI?

dustymabe:
> I'm OK with this for now, 

same here. my knee jerk on this was to not remove them

> but we do want to support e.g. `/var` being LVM (or Stratis) in the future right? And probably iSCSI?

at least when we add support back in the future (if we do) it will be for a specific reason so I'm +1 for removing and then adding back when we need it

ajeddeloh:
> at least when we add support back in the future (if we do) it will be for a specific reason so I'm +1 for removing and then adding back when we need it

Yup.

We gotta be fearless in ripping things out now, because it will get way harder once we ship. 

> I'm OK with this for now, but we do want to support e.g. /var being LVM (or Stratis) in the future right? And probably iSCSI?

`/var` doesn't need to be mounted in the initramfs and if it does, then Ignition needs to know how to mount it in the mount stage, meaning we'll have to roll our own dracut module. We'll cross the iscsi bridge when we hit it, but until we have a use case, we should keep it out. Even when we do have a use case we shouldn't be running in unconditionally.

dustymabe:
> meaning we'll have to roll our own dracut module

implying we'd need our own dracut module that pulls in the lvm tools?

we'll cross this bridge when we get there.. merging for now

PR #64, opened 2019-03-12
installer: make grub.cfg usable on aarch64

### LINKED ISSUE #63: Use linux/initrd in grub.cfg for aarch64
Because `aarch64` has no non-UEFI boot, the grub commands are just called `linux` and `initrd` instead of `linuxefi` and `initrdefi` like in `x86_64`.

dustymabe:
From some brief testing it looks like `linux` and `initrd` work
on x86_64 EFI boot so we'll just use that so we can use the same
grub.cfg for aarch64 or for x86_64.

Fixes #63

### DIFF
--- a/installer/EFI/fedora/grub.cfg
+++ b/installer/EFI/fedora/grub.cfg
@@ -2,6 +2,10 @@
 # efiboot.img on the Fedora Server DVD iso. Diff this file with that
 # file in the future to pick up changes.
 #
+# One diff to note is we use linux and initrd instead of linuxefi and
+# initrdefi. We do this because it works and allows us to use this same
+# file on other architecutres. https://github.com/coreos/fedora-coreos-config/issues/63
+#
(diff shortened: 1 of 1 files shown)


dustymabe:
cc @r4f4

r4f4:
LGTM

dustymabe:
> LGTM

do you want to confirm this works on aarch64 first or should we go ahead and merge?

r4f4:
I'm facing a problem currently where `virt-make-fs` is getting stuck inside the `cosa` container, so the ISO generation never passes the creation of the `ESP` partition phase. This worked before with a different machine.
So if you want to wait confirmation on `aarch64` it might take quite a while.

dustymabe:
> So if you want to wait confirmation on `aarch64` it might take quite a while.

no worries. just open PRs/Issues when you get to it.

PR #66, opened 2019-03-15
Add overlay/ directory and empty out manifest

jlebon:
Right now the manifest's postprocess is a combination of temporary hacks
and glue code. Split those out in a separate `overlay/` directory
instead. This will be picked up by coreos-assembler and automatically
transformed into an RPM.

Uses: https://github.com/coreos/coreos-assembler/pull/414

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -46,145 +46,39 @@ check-groups:
   type: "file"
   filename: "group"
 
+default_target: multi-user.target
+
+remove-from-packages:
+  # We're not using resolved yet
+  - [systemd, /usr/lib/systemd/system/systemd-resolved.service]
+
+  # The grub bits are mainly designed for desktops, and IMO haven't seen
+  # enough testing in concert with ostree. At some point we'll flesh out
+  # the full plan in https://github.com/coreos/fedora-coreos-tracker/issues/47
+  # NOTE: Also remove 01_fallback_counting once we move to f30
+  - [grub2-tools, /etc/grub.d/01_menu_auto_hide,
+                  /usr/lib/systemd/.*]
+
+
+# ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
 postprocess:
+  # This will be dropped once rpm-ostree because module-aware.
+  # https://github.com/projectatomic/rpm-ostree/issues/1542#issuecomment-419684977
+  # https://github.com/projectatomic/rpm-ostree/issues/1435
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
-
-    # https://github.com/projectatomic/rpm-ostree/issues/1542#issuecomment-419684977
     for x in /etc/yum.repos.d/*modular.repo; do
       sed -i -e 's,enabled=[01],enabled=0,' ${x}
     done
-
-    # The grub bits are mainly designed for desktops, and IMO haven't seen
-    # enough testing in concert with ostree. At some point we'll flesh out
-    # the full plan in https://github.com/coreos/fedora-coreos-tracker/issues/47
-    # NOTE: Also remove 01_fallback_counting once we move to f30
-    rm -v /etc/grub.d/01_menu_auto_hide
-    rm -v /usr/lib/systemd/user/grub-boot-success*
-
-    # See machineid-compat in host-base.yaml.
-    # Since that makes presets run on boot, we need to have our defaults in /usr
-    ln -sfr /usr/lib/systemd/system/{multi-user,default}.target
-
-    # Persistent journal by default
-    echo 'Storage=persistent' >> /etc/systemd/journald.conf
-
-    # https://github.com/openshift/os/issues/96
-    # sudo group https://github.com/openshift/os/issues/96
-    echo '%sudo        ALL=(ALL)       NOPASSWD: ALL' > /etc/sudoers.d/coreos-sudo-group
-
-    # We're not using resolved yet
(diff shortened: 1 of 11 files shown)


fedora-coreos-base.yaml
```diff
@@ -45,106 +45,42 @@ check-groups:
...
+
+  # The grub bits are mainly designed for desktops, and IMO haven't seen
+  # enough testing in concert with ostree. At some point we'll flesh out
+  # the full plan in https://github.com/coreos/fedora-coreos-tracker/issues/47
+  # NOTE: Also remove 01_fallback_counting once we move to f30
+  - [grub2-tools, /etc/grub.d/01_menu_auto_hide,
+                  /usr/lib/systemd/.*]
+
+
+# ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
```
jlebon: Not to say that any postprocessing = bad. I'm sure we'll have examples of postprocessing that really does belong here. But the bar for those should be high.

fedora-coreos-base.yaml
```diff
@@ -45,106 +45,42 @@ check-groups:
...
 postprocess:
+  # This will be dropped once rpm-ostree because module-aware.
+  # https://github.com/projectatomic/rpm-ostree/issues/1542#issuecomment-419684977
```
dustymabe: What's the value in breaking this out into multiple postprocess entries vs a single file ?

dustymabe (commented):
I think this looks nice but I worry that it's easier to forget about some of the things in the overlay since they will be "out of sight, out of mind".

fedora-coreos-base.yaml
jlebon: Just to be more explicit about the fact that they're separate/independently-tracked hacks. `rpm-ostree compose tree` also prints stuff like ``Executing `postprocess` inline script '2'`` which is nice.

jlebon:
> I think this looks nice but I worry that it's easier to forget about some of the things in the overlay since they will be "out of sight, out of mind".

Yeah, that's a concern. I think the right thing to do there too is to have tracker issues to make sure it doesn't fall off the radar?

cgwalters:
> but I worry that it's easier to forget about some of the things in the overlay since they will be "out of sight, out of mind".

Isn't it the opposite?  My eyes glaze over a bit reading shell-in-yaml but I think a directory `overlay` with the files directly is very clear and hard to miss.

dustymabe:
> Isn't it the opposite? My eyes glaze over a bit reading shell-in-yaml (edit: in some places shell-generating-shell-in-YAML!)

It's definitely harder to read, but I don't think harder to miss.

> but I think a directory `overlay` with the files directly is very clear and hard to miss.

yeah. I think it depends. The issues can mostly be helped by good use of grep, but it was worth pointing out

jlebon:
OK, this is ready to go now! I've added tracking links for some of the hacks.

overlay/usr/lib/systemd/system/coreos-useradd-core.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
dustymabe: lost comments here - can you add them back?

dustymabe (commented):
Since various files are used to contribute to the same "fix" for certain things (i.e. a systemd unit and a executable file work together to achieve the same goal) I think it would be nice if we supported a way to group things together. One way to do this would be to support sub directories under `overlay`. Each sub directory could contain a readme that gets ignored but provides some more context for the group of files.

We could also still support a standalone or  `general` directory as well for single file standalone goals.

WDYT?

jlebon:
> One way to do this would be to support sub directories under `overlay`.

Hmm, I like the idea of surfacing the fact that some of the files are related somehow, but having multiple rootfs'es feels like overkill. How about just `# see also ...` comments? Most of the files I think are stand-alone, so it wouldn't be that many.

fedora-coreos-base.yaml
```diff
@@ -46,145 +46,39 @@ check-groups:
   type: "file"
   filename: "group"
 
+default_target: multi-user.target
```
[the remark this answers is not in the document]
jlebon: This replaces `ln -sfr /usr/lib/systemd/system/{multi-user,default}.target`. AFAIK, it was inherited from FAH from before https://github.com/projectatomic/rpm-ostree/pull/1427.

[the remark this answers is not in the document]
cgwalters: I don't think it's needed anymore, but it's not going to hurt anything either.

overlay/usr/lib/systemd/journald.conf.d/10-coreos-persistent.conf
```diff
@@ -0,0 +1,5 @@
+# Hardcode persistent journal by default. journald has this "auto" behaviour
+# that only makes logs persistent if `/var/log/journal` exists, which it won't
+# on first boot because `/var` isn't fully populated.
```
jlebon: Though actually... I think we'll be able to drop this one once we move to sysusers and run it from the initrd.

dustymabe:
> Hmm, I like the idea of surfacing the fact that some of the files are related somehow, but having multiple rootfs'es feels like overkill. How about just `# see also ...` comments? Most of the files I think are stand-alone, so it wouldn't be that many.

Yeah. It's not something i would block you on, but I do think that it's easier to set up a structure where people do the right thing vs relying on them to do the right thing. I'll approve the PR and let you make the judgement call.

dustymabe (approved):

jlebon:
> I'll approve the PR and let you make the judgement call.

OK cool, let's go with this for now, and if this we have more instances, we'll reconsider!

PR #70, opened 2019-03-21
Add 40coreos dracut module

jlebon:
This module is the integration point between Ignition and Fedora CoreOS.
Notably, we add two new systemd services in the initrd which mesh with
Ignition's own systemd services:

1. `coreos-mount-var.service`, which finds the `/var` stateroot bind
   mount and mounts it under `/sysroot/var` before `ignition-mount`
   potentially adds more mounts underneath (or even shadowing `/var`
   itself).
2. `coreos-populate-var.service`, which sets up `/var` on first boot
   using `systemd-tmpfiles` before `ignition-files` runs.

### DIFF
--- a/overlay/usr/lib/dracut/modules.d/40coreos-var/coreos-mount-var.service
+++ b/overlay/usr/lib/dracut/modules.d/40coreos-var/coreos-mount-var.service
@@ -1,0 +1,24 @@
+[Unit]
+Description=Mount OSTree /var
+DefaultDependencies=false
+
+# Make sure ExecStop= runs before we switch root
+Before=initrd-switch-root.target
+
+# Make sure if ExecStop= fails, the boot fails
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+# Make sure /sysroot is mounted first, since we're mounting under there
+Requires=initrd-root-fs.target
+After=initrd-root-fs.target
+
+# Need to do this before Ignition mounts any other filesystems (potentially
+# shadowing our own bind mount).
+Before=ignition-mount.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/coreos-mount-var mount
+ExecStop=/usr/sbin/coreos-mount-var umount
--- a/overlay/usr/lib/dracut/modules.d/40coreos-var/coreos-mount-var.sh
+++ b/overlay/usr/lib/dracut/modules.d/40coreos-var/coreos-mount-var.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+set -euo pipefail
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if [ $# -ne 1 ] || { [[ $1 != mount ]] && [[ $1 != umount ]]; }; then
+    fatal "Usage: $0 <mount|umount>"
+fi
+
+get_ostree_arg() {
+    # yes, this doesn't account for spaces within args, e.g. myarg="my val", but
+    # it still works for our purposes
+    (
+    IFS=$' '
+    # shellcheck disable=SC2013
+    for arg in $(cat /proc/cmdline); do
+        if [[ $arg == ostree=* ]]; then
+            echo "${arg#ostree=}"
+        fi
+    done
+    )
+}
+
+do_mount() {
+    ostree=$(get_ostree_arg)
+    if [ -z "${ostree}" ]; then
+        fatal "No ostree= kernel argument in /proc/cmdline"
+    fi
+
+    deployment_path=/sysroot/${ostree}
+    if [ ! -L "${deployment_path}" ]; then
+        fatal "${deployment_path} is not a symlink"
+    fi
+
+    stateroot_var_path=$(realpath "${deployment_path}/../../var")
+    if [ ! -d "${stateroot_var_path}" ]; then
+        fatal "${stateroot_var_path} is not a directory"
+    fi
+
+    echo "Mounting $stateroot_var_path"
+    mount --bind "$stateroot_var_path" /sysroot/var
+}
+
+do_umount() {
+    echo "Unmounting /sysroot/var"
+    umount /sysroot/var
+}
+
+"do_$1"
--- a/overlay/usr/lib/dracut/modules.d/40coreos-var/coreos-populate-var.service
+++ b/overlay/usr/lib/dracut/modules.d/40coreos-var/coreos-populate-var.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Populate OSTree /var
+DefaultDependencies=false
+
+# Need to do this with all mount points active
+After=ignition-mount.service
+
+# But *before* we start dumping files in there
+Before=ignition-files.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/coreos-populate-var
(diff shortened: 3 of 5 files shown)


cgwalters:
How would we handle `/var` as a real partition?

jlebon:
> How would we handle `/var` as a real partition?

The `mount` stage will mount it just like any other. The bind-mount we do here is for the case where only e.g. `/var/home` is specified, in which case, we still want to populate all the other stuff under `/var` that would go in the stateroot. Instead of doing it conditionally, we just always bind-mount and always unmount to keep it simple (that's the "shadowing" part in the commit message).

Re. `/var` in FCOS specifically, I think the current idea is to ship that as a base config that users can override/tweak as they want.

dustymabe:
so I know we are wanting `ignition-dracut` repo to be mostly distro agnostic but would there be harm in having code in there that activates on one distro but is dormant on another? i.e. it would be kind of nice to just put this module in ignition-dracut and then non-ostree based distros could leave it out.

cc @ajeddeloh @bgilbert for input. It's also possible there was another conversations that I missed or forgot that led us here to begin with.

jlebon:
OK, this is working pretty well. Two things that bit me were https://src.fedoraproject.org/rpms/ignition/pull-request/31 and https://github.com/coreos/ignition/pull/771. I played with various mounts, e.g. `/var`, `/var/lib/containers`, `/var/home`, as well as all three at the same time!

```
[root@localhost ~]# findmnt | grep /var
├─/var                    /dev/vdc1  xfs  rw,relatime,seclabel,attr2,inode64,noquota
│ ├─/var/home             /dev/vdb1  xfs  rw,relatime,seclabel,attr2,inode64,noquota
│ └─/var/lib/containers   /dev/vdd1  xfs  rw,relatime,seclabel,attr2,inode64,noquota
```

For anyone who wants to follow along, how I'm testing this right now is with:

```diff
diff --git a/fedora-coreos-base.yaml b/fedora-coreos-base.yaml
index fef318a..95a732a 100644
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -58,6 +58,11 @@ remove-from-packages:
   # NOTE: Also remove 01_fallback_counting once we move to f30
   - [grub2-tools, /etc/grub.d/01_menu_auto_hide,
                   /usr/lib/systemd/.*]
+  # XXX
+  - [ignition, /usr/lib/dracut/modules.d/30ignition]
+
+add-files:
+  - [ignition, /usr/lib/dracut/modules.d/31ignition/ignition]


 # ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
```

And `rsync -al` of `30ignition` in https://github.com/coreos/ignition-dracut/pull/47 at `overlay/usr/lib/dracut/modules.d/31ignition`.

(I was hitting issues packaging ignition git master with the recent git module rework, so just went with the nuclear option.)

And then a VM with an extra disk added and a config like:

```json
    "storage": {
        "disks": [
            {
                "device": "/dev/vdb",
                "wipeTable": true,
                "partitions": [
                    {
                        "label": "VAR",
                        "number": 0,
                        "sizeMiB": 0,
                        "wipePartitionEntry": true
                    }
                ]
            }
        ],
        "filesystems": [
            {
                "device": "/dev/vdb1",
                "path": "/var",
                "format": "xfs",
                "wipeFilesystem": true,
                "label": "VAR"
            }
        ],
        "files": [
            {
                "path": "/etc/fstab",
                "append": [
                    {
                        "source": "data:,%0A/dev/vdb1%20/var%20xfs%20defaults%200%200%0A"
                    }
                ]
            },
            {
                "path": "/var/lib/hello.txt",
                "contents": {
                    "source": "data:,hello%20world"
                }
            }
        ]
    }
```

jlebon:
I dropped WIP on this, but do note that this requires https://github.com/coreos/ignition-dracut/pull/47, which is pending on FCOS switching to the new Ignition spec. So we need to hold this for now. That said, it's ready for review, and it'd be good if someone else tested this as well!

ajeddeloh (approved):
LGTM

jlebon:
>  i.e. it would be kind of nice to just put this module in ignition-dracut and then non-ostree based distros could leave it out

Here's a question related to this: let's say we keep them separate, then couldn't we just fold the Ignition dracut module into https://github.com/coreos/ignition? Other than `ignition-setup`, all the services there are solely about integrating Ignition into an initrd using stable systemd targets. We could probably move off the few other stuff remaining and drop the ignition-dracut repo entirely.

dustymabe:
> > i.e. it would be kind of nice to just put this module in ignition-dracut and then non-ostree based distros could leave it out
> 
> Here's a question related to this: let's say we keep them separate, then couldn't we just fold the Ignition dracut module into https://github.com/coreos/ignition? Other than `ignition-setup`, all the services there are solely about integrating Ignition into an initrd using stable systemd targets. We could probably move off the few other stuff remaining and drop the ignition-dracut repo entirely.

Perhaps we could move ignition-dracut into ignition but probably not worth it right now. I'm cool with this as is and we can re-org in the future if it's an attractive option.

bgilbert:
I think we were talking about putting FCOS-specific modules into fedora-coreos-config, but that seems kind of awkward.  If we're thinking of restructuring later on anyway, I'm okay adding a separate module here for now.

bgilbert:
Aaaand I thought this PR was in ignition-dracut, not fedora-coreos-config.  Yes, here seems fine.

overlay/usr/lib/dracut/modules.d/40coreos/coreos-populate-var.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+# See the similar code block in Anaconda, which handles this today for Atomic
+# Host and Silverblue:
+# https://github.com/rhinstaller/anaconda/blob/b9ea8ce4e68196b30a524c1cc5680dcdc4b89371/pyanaconda/payload/rpmostreepayload.py#L332
+
+# Simply manually mkdir /var/lib; the tmpfiles.d entries otherwise reference
+# users/groups which we don't have access to from here (though... we *could*
+# import them from the sysroot, and have nss-altfiles in the initrd, but meh...
+# let's just wait for systemd-sysusers which will make this way easier:
```
bgilbert: As a starting point, won't we want a `systemd-sysusers` invocation in this script?

jlebon:
OK, this is now ready for review again!
Needs to be merged together with https://github.com/coreos/ignition-dracut/pull/76.

Also requires: https://github.com/coreos/fedora-coreos-config/pull/85

overlay/usr/lib/dracut/modules.d/40coreos/coreos-populate-var.sh
jlebon: I was playing with this, but I think for now we should wait for the full sysusers switchover. The issue is that it'd require fooling systemd-sysusers into somehow being aware of the entries in `/sysroot/usr/lib/{passwd,group}` so that it notices entries that already exist. Alternatively, we can drop from group/passwd in this repo the entries that we expect to be created by sysusers, but that seems error-prone.

The other issue is that we'll then have some sysusers in `/etc/passwd` and some others in `/usr/lib/passwd`, and I'm not sure how well rpm-ostree layering will play with that. See e.g. https://github.com/projectatomic/rpm-ostree/blob/6b2ac5885fd1c4bdff2de90a2234d03c839f76ef/src/libpriv/rpmostree-passwd-util.c#L1183.

I think we *could* get this to work, but I'm not sure it's worth the investment if we'll just switch to sysusers anyway.

jlebon:
Rebased! :arrow_up:

ajeddeloh (approved):
LGTM

PR #71, opened 2019-03-22
Update console-login-helper-messages manifest and enabled presets

rfairley:
These changes should be added once the update to the latest source lands in stable https://bodhi.fedoraproject.org/updates/console-login-helper-messages-0.16-2.fc29. This update will remove the `.path` units from the source.

This PR can be merged anytime before the update lands as well, as the `.path` units are not involved in the MOTD or issue generation today, and they do not need enabling.

Also lists only the following in the manifest:

```
console-login-helper-messages-issuegen
console-login-helper-messages-motdgen
console-login-helper-messages-profile
```

as the `console-login-helper-messages` package on its own provides common dependencies for the 3 subpackages, and need not be requested explicitly.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -128,9 +128,8 @@ packages:
   - rsync fuse-sshfs
   - gnupg2
   # User experience
-  - console-login-helper-messages
-  - console-login-helper-messages-motdgen
   - console-login-helper-messages-issuegen
+  - console-login-helper-messages-motdgen
   - console-login-helper-messages-profile
   # CoreOS Installer
   - coreos-installer coreos-installer-dracut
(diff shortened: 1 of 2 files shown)


fedora-coreos-base.yaml
```diff
@@ -116,9 +116,6 @@ packages:
   - rsync fuse-sshfs
   - gnupg2
   # User experience
-  - console-login-helper-messages
-  - console-login-helper-messages-motdgen
-  - console-login-helper-messages-issuegen
-  - console-login-helper-messages-profile
+  - console-login-helper-messages-issuegen console-login-helper-messages-motdgen console-login-helper-messages-profile
```
dustymabe: i kind of prefer these on different lines than having them all on one line - any reason to have them all on one line?

rfairley: No real reason - thought I'd group them on one line with these being related - but can see it is easier to distinguish each being on separate lines. Fixed!

rfairley:
Rebased and tested this with `cosa build` - good to merge.

jlebon:
This looks good to me! Just one typo in the commit message: `s/explict/explicit/` ?

rfairley:
@jlebon thanks for catching! Updated

PR #73, opened 2019-03-27
postprocess: Nuke everything in /etc/systemd/system

cgwalters:
This should really be done in rpm-ostree if `machineid-compat: false`
but for now let's do the quick hack here.

The core problem is today the RPM scriptlets are enabling units,
but for us Ignition+presets should be the canonical definition of
things.

This avoids having another source of truth.


jlebon:
> This should really be done in rpm-ostree if `machineid-compat: false`
but for now let's do the quick hack here.

Hmm, yeah that makes sense. I filed https://github.com/projectatomic/rpm-ostree/issues/1803. Can you add a link to that just above?

cgwalters:
>  Can you add a link to that just above?

Done!

PR #74, opened 2019-03-27
Use Ignition to create core user

### LINKED ISSUE #41: core user isn't added to groups when created via Ignition
The `core` user is only added to groups if the user doesn't exist on first boot:

https://github.com/coreos/fedora-coreos-config/blob/f08f79e43ca20772d3b622e105c6a6030fc627f4/fedora-coreos-base.yaml#L118

Thus, if Ignition is used to add an SSH key to the `core` user, but the Ignition config doesn't explicitly add the user to groups, it will not have sudo access:

```yaml
passwd:
  users:
    - name: core
      ssh_authorized_keys:
        - "ssh-rsa ..."
```

The above config works on Container Linux.

cc @ajeddeloh

bgilbert:
Create the core user in the Ignition base config and allow users to modify it via config merging.  This fixes the user not receiving default groups if anything about the user (e.g. SSH keys) is specified via Ignition.

Fixes #41.


arithx:
Note: didn't test it on my end yet

arithx:
Tested & working for me! 🎉

PR #76, opened 2019-04-02
aarch64: Remove unnecessary packages for aarch64

andymcc:
aarch64 is currently using UEFI and not u-boot, additionally the
bcm283x-firmware package is only needed for Raspberry Pi devices which
have the Broadcom BCM283x devices, and can also be safely removed.

### DIFF
--- a/minimal.yaml
+++ b/minimal.yaml
@@ -9,7 +9,6 @@ packages:
 # bootloader
 packages-aarch64:
   - grub2-efi ostree-grub2 efibootmgr shim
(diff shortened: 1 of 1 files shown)


cgwalters:
What's up with `bcm283x-firmware` though?   Looks like that's wireless firmware, not related to the bootloader.

andymcc:
It's the driver for the broadcom device for Raspberry Pi - we don't need it, and would need to enable additional repos to install it. The commit message should be more specific about that though so let me update that!

cgwalters:
LGTM, giving to bgilbert for another review/merge.

PR #83, opened 2019-04-23
overlay: Seed /root with default bash profile

jlebon:
This has a long history in Atomic Host:
https://bugzilla.redhat.com/show_bug.cgi?id=1193590

The TL;DR is that since OSTree commits ship with empty `/var`, and
`/root` is really `/var/roothome`, then the default `.bash*` files for
the root don't actually exist. The RHBZ above was to try to teach bash
to fallback to the default files automatically, but it was deemed too
intrusive in the end and backed out.

In Atomic Host, we've worked around this by simply seeding the files
from the kickstart:
https://pagure.io/fedora-kickstarts/c/740a9b4969ca30da36458faf6809908f4da904c4?branch=master

This patch does something similar but at Ignition time. We do this
before the files stage so that for all purposes, this is part of the
distro (e.g. an Ignition config can append to `/root/.bashrc` for
example, or can add a file to `/etc/profile.d/` and expect it to be
sourced).

Concretely, this fixes non-login shells not sourcing `/etc/bashrc` and
`/etc/profile.d/` and thus having a prompt like `bash-5.0# `, no `ls`
colouring, non-default umask, etc...

### DIFF
--- a/overlay/usr/lib/dracut/modules.d/15coreos-root/coreos-root-bash.service
+++ b/overlay/usr/lib/dracut/modules.d/15coreos-root/coreos-root-bash.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=CoreOS Root Bash
+DefaultDependencies=false
+Before=ignition-complete.target
+
+# We need all the filesystems already mounted.
+Requires=ignition-mount.service
+After=ignition-mount.service
+
+# Run before ignition-files.service so configs can further modify.
+Before=ignition-files.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/sh -c "cp /sysroot/etc/skel/.bash* /sysroot/root"
--- a/overlay/usr/lib/dracut/modules.d/15coreos-root/module-setup.sh
+++ b/overlay/usr/lib/dracut/modules.d/15coreos-root/module-setup.sh
@@ -1,0 +1,17 @@
+# Populate /root with default bash configs. See:
+# https://bugzilla.redhat.com/show_bug.cgi?id=1193590
+
+depends() {
+    echo ignition
+}
+
+check() {
+    return 0
+}
+
+install() {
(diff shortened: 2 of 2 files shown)


overlay/usr/lib/dracut/modules.d/15coreos-root/coreos-root-bash.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=CoreOS Root Bash
+DefaultDependencies=false
+Before=ignition-complete.target
```
cgwalters: Hmm...can't this be a `ConditionFirstBoot` unit that runs just in early boot?

jlebon: Hmm, I mentioned this in the commit message:

> We do this
before the files stage so that for all purposes, this is part of the
distro (e.g. an Ignition config can append to `/root/.bashrc` for
example, or can add a file to `/etc/profile.d/` and expect it to be
sourced).

Does that help?

Another example: an Ignition config can that way empty `/root/.bashrc` completely if one doesn't want the default behaviour (instead of e.g. masking a specific systemd unit, which doesn't feel as natural).

Note that this is only executed on first boot anyway, since it gets pulled in by `ignition-complete.target` (which is only activated when Ignition runs, see https://github.com/coreos/ignition-dracut/pull/55).

cgwalters: Ah yep makes sense.

I did spent an hour at one point discover Debian doesn't need this and ended up in a rabbit hole of bash patches and config options versus the PAM setups etc.

cgwalters:
LGTM

jlebon:
(Will merge this soonish if there's no objection!)

dustymabe:
thanks for the detailed commit message - that helps a lot!

PR #84, opened 2019-04-23
fedora-coreos-base: rm things with python deps

ajeddeloh:
This doesn't remove all things with python dependencies, just authconfig
and nfs-utils (replaced by nfs-utils-coreos).

Tested, works.

Addresses https://github.com/coreos/fedora-coreos-tracker/issues/123 and https://github.com/coreos/fedora-coreos-tracker/issues/121


ajeddeloh:
Tested, works.

PR #85, opened 2019-04-25
15coreos-root: run after coreos-populate-var.service

jlebon:
Otherwise `/var/roothome` might not have been created yet.

I didn't notice this before because we weren't fully cleaning out `/var`
in the image: coreos/coreos-assembler#494.


jlebon:
Ping? This should be an easy one to review.

arithx (approved):

PR #88, opened 2019-05-01
base: rm setools-console, policycoreutils-python

ajeddeloh:
Remove these as they depend on python

closes https://github.com/coreos/fedora-coreos-tracker/issues/122
closes https://github.com/coreos/fedora-coreos-tracker/issues/126


jlebon:
:no_entry_sign: :snake:

PR #89, opened 2019-05-06
overlay: add 40coreos-network dracut module

yuqi-zhang:
Add coreos-teardown-initramfs-network.service to run after ignition
has finished using initramfs networking, so NetworkManager properly
brings up ignition-configured networking in the real root. Otherwise
the initramfs network persists into the real root.

Note: I'm not sure what the correct priority is, so I made it 40 to match other services.
It shouldn't really matter since this service actually runs during `ExecStop` and
before `Before=initrd-switch-root.target`.

### DIFF
--- a/overlay/usr/lib/dracut/modules.d/40coreos-network/coreos-teardown-initramfs-network.service
+++ b/overlay/usr/lib/dracut/modules.d/40coreos-network/coreos-teardown-initramfs-network.service
@@ -1,0 +1,20 @@
+# Clean up the initramfs networking on first boot
+# so the real network is being brought up
+
+[Unit]
+Description=Tear down initramfs networking
+DefaultDependencies=false
+After=ignition-files.service
+
+# Make sure ExecStop= runs before we switch root
+Conflicts=initrd-switch-root.target umount.target
+Before=initrd-switch-root.target
+
+# Make sure if ExecStop= fails, the boot fails
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStop=/usr/sbin/coreos-teardown-initramfs-network
--- a/overlay/usr/lib/dracut/modules.d/40coreos-network/coreos-teardown-initramfs-network.sh
+++ b/overlay/usr/lib/dracut/modules.d/40coreos-network/coreos-teardown-initramfs-network.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -euo pipefail
+
+# Clean up the interfaces set up in the initramfs
+# This mimics the behaviour of dracut's ifdown() in net-lib.sh
+# This script should be considered temporary. We eventually
+# want to move to NetworkManager based dracut modules. See:
+# https://github.com/dracutdevs/dracut/tree/master/modules.d/35network-manager
+if ! [ -z "$(ls /sys/class/net)" ]; then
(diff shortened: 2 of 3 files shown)


dustymabe:
so for RHCOS we have this in the ignition-dracut repo and for FCOS we have it here? should we merge those two strategies in the future?

overlay/usr/lib/dracut/modules.d/40coreos-network/coreos-teardown-initramfs-network.service
```diff
@@ -0,0 +1,20 @@
...
+[Unit]
+Description=Tear down initramfs networking
+DefaultDependencies=false
+After=ignition-files.service
+
+# Make sure ExecStop= runs before we switch root
+Conflicts=initrd-switch-root.target umount.target
+Before=initrd-switch-root.target
+
+# Make sure if ExecStart= fails, the boot fails
```
jlebon: This should read `ExecStop=`.

jlebon:
> should we merge those two strategies in the future?

Once RHCOS rebases to spec3, this will have to happen (i.e. it'll have to be part of the RHCOS `overlay/`). So it'll converge eventually. But yeah, could definitely do this now too.

yuqi-zhang:
Fixed up, thanks!

> should we merge those two strategies in the future?

And I would vote to do this after GA at the earliest, but I agree we should match the two eventually.

overlay/usr/lib/dracut/modules.d/40coreos-network/coreos-teardown-initramfs-network.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euo pipefail
+
+# Clean up the interfaces set up in the initramfs
+# This mimics the behaviour of dracut's ifdown() in net-lib.sh
```
dustymabe: hmm - if it mimics that code then why not source it and run it directly rather than implement our own?

bgilbert:
For FCOS, at least, this is a temporary fix until we can move away from dracut's networking scripts.  In the meantime, I'm fine with merging it.

overlay/usr/lib/dracut/modules.d/40coreos-network/coreos-teardown-initramfs-network.sh
dustymabe: bump

yuqi-zhang: Sorry, this is the original function: https://github.com/dracutdevs/dracut/blob/master/modules.d/40network/net-lib.sh#L97
It does some things we don't necessarily need/want, and I recall having issues running it directly, so I opted to take a snippet of it.

yuqi-zhang: If you want I can look into making that work instead? I didn't think it was necessary. We are likely going to try to move for NM managed dracut networking. See: https://github.com/dracutdevs/dracut/tree/master/modules.d/35network-manager

dustymabe: ehh - if it's temporary then it's fine IMHO - should we mention that NM thing in a comment?

yuqi-zhang:
Rebased and added comment, feel free to merge

PR #91, opened 2019-05-14
repos: move to coreos-pool repo

dustymabe:
This is an intermediate step towards our longer term goal of consuming
everything for our production refs from a distRepo generated from the
coreos-pool koji tag. Right now we'll just consume from the coreos-pool
as well as the traditional fedora repos.

This replaces the fedora-coreos-continuous repo as that will be used for
CI streams that rebuild quickly from upstream sources. We were currently
using it to bypass bodhi but we can use coreos-pool for that too.

### DIFF
--- a/fedora-coreos-pool.repo
+++ b/fedora-coreos-pool.repo
@@ -1,0 +1,8 @@
+[fedora-coreos-pool]
+name=Fedora coreos pool repository - $basearch
+baseurl=https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/latest/$basearch/
+enabled=1
+repo_gpgcheck=0
+type=rpm-md
+gpgcheck=1
+skip_if_unavailable=True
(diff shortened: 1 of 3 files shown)


dustymabe:
FYI I tagged all the latest items from the continuous tag into coreos-pool as well so we shouldn't see any regressions. Also I tagged in the latest xfsprogs so now we **no longer pull in python** (cc @ajeddeloh @sinnykumari ).

fedora-coreos-pool.repo
```diff
@@ -0,0 +1,8 @@
+[fedora-coreos-continuous]
```
sinnykumari: rename the repo to fedora-coreos-pool for consistency ?

sinnykumari (commented):
One minor comment, other than that LGTM

jlebon (approved):
Looks sane to me other than @sinnykumari's nit. The repo name also ends up in the commit metadata.

fedora-coreos-pool.repo
dustymabe: fixed!

PR #92, opened 2019-05-14
Clean-up unused repos

sinnykumari:
With coreos-pool and f*-coreos-continuous tags in place, copr repo shouldn't be needed


cgwalters:
LGTM!

dustymabe:
> copr repo shouldn't be needed

agree!

Issue #33, opened 2018-12-05
Add a preset file for Fedora CoreOS in the release rpm


rfairley:
This issue tracks moving the systemd presets FCOS needs to a proper place. Currently, FCOS writes a preset file with the needed additional presets during [post-processing of the build](https://github.com/coreos/fedora-coreos-config/blob/a733e2cfb39cfc55981ad1d37c359665aece7dd3/fedora-coreos-base.yaml#L116-L122).

The preset file would fit cleanly into `fedora-release-coreos`, named `80-coreos.preset`, analogous to how `fedora-release-workstation` contains `80-workstation.preset`.

Once the list of needed presets in FCOS stabilizes, we could open a PR to https://src.fedoraproject.org/rpms/fedora-release with the new preset file to go in `fedora-release-coreos` in Rawhide and backport to F29 (to use in testing).

cc @dustymabe @sgallagher

sgallagher:
:+1:, this definitely belongs in `fedora-release-coreos`.

dustymabe:
Thanks @rfairley for the well defined problem statement and eventual solution. When the list stabilizes, let's have someone go add them to the `fedora-release-coreos` rpm.

rfairley:
Thanks @sgallagher, @dustymabe !

rfairley:
PR to get this into Rawhide was merged: https://src.fedoraproject.org/rpms/fedora-release/pull-request/74

Should we backport this to F30 as well? Then we can use presets from `fedora-release-coreos` in the F30 preview.

jlebon:
> Should we backport this to F30 as well?

SGTM!

rfairley:
PR to backport to F30: https://src.fedoraproject.org/rpms/fedora-release/pull-request/79

dustymabe:
bumped that PR since F30 is GA now

rfairley:
Since the presets might still change (as they did in #93), we are deciding to keep the [overlay](https://github.com/coreos/fedora-coreos-config/blob/master/overlay/usr/lib/systemd/system-preset/42-coreos.preset) as a source of truth for the presets, at least until the FCOS is stable. Adding this to the FCOS stable milestone.

The preset in the fedora-release-coreos RPM will be updated to include a comment referring to the overlay as the place to configure presets.

rfairley:
> The preset in the fedora-release-coreos RPM will be updated to include a comment referring to the overlay as the place to configure presets.

PRs for this in f31 and f30:

[...]

PR #93, opened 2019-05-14
Add afterburn package

rfairley:
Add Afterburn to the package manifest, and enable the provider checkin
services through the coreos preset file.

Part of coreos/fedora-coreos-tracker#4.

---

Now gets sourced from `fedora-coreos-pool`:

```
afterburn-4.1.0-1.module_f30+4209+68fe9bdd.x86_64 (fedora-coreos-pool)
```

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -97,6 +97,7 @@ packages:
   - selinux-policy-targeted
   # System setup
   - ignition
+  - afterburn
   - dracut-network
   - passwd
   # SSH
--- a/overlay/usr/lib/systemd/system-preset/42-coreos.preset
+++ b/overlay/usr/lib/systemd/system-preset/42-coreos.preset
@@ -4,3 +4,6 @@ enable console-login-helper-messages-issuegen.service
 enable console-login-helper-messages-motdgen.service
 # This one is from https://github.com/coreos/ignition-dracut
 enable ignition-firstboot-complete.service
+# Boot checkin services for cloud providers.
+enable afterburn-checkin.service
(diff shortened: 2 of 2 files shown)


rfairley:
Note we will need to update the presets [upstream](https://src.fedoraproject.org/rpms/fedora-release/blob/master/f/80-coreos.preset), if enabling the `afterburn` presets here.

dustymabe:
> Note we will need to update the presets [upstream](https://src.fedoraproject.org/rpms/fedora-release/blob/master/f/80-coreos.preset), if enabling the `afterburn` presets here.

Noted.. if we keep the overlay of files in this repo maybe we should just stop shipping presets in the `fedora-release-coreos` rpm. I'd be interested in others thoughts here.

rfairley:
I think `overlay/usr/lib/systemd/system-preset/42-coreos.preset` is worth keeping for faster iteration (like adding the unit here. We shouldn't introduce too much churn in `fedora-release-coreos`.

Thinking about it again, #33 was opened when we didn't have the overlay. The overlay probably does make sense for now as the main source of truth for presets with things still changing. Maybe we could come back to #33 as an FCOS stable goal (to reconser shipping presets in `fedora-release-coreos`)?

I'd be +1 to removing the presets from the `fedora-release-coreos` RPM for now (maybe just keep an empty file with a comment there so we can add to it easily if we do come back to it).

dustymabe:
> I'd be +1 to removing the presets from the `fedora-release-coreos` RPM for now (maybe just keep an empty file with a comment there so we can add to it easily if we do come back to it).

agree

PR #94, opened 2019-05-15
manifest: disable systemd-firstboot.service

jlebon:
We don't need its functionality. We should put emphasis on Ignition for
setting things up on first boot.

/cc @ajeddeloh @bgilbert

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -92,6 +92,14 @@ postprocess:
     set -xeuo pipefail
     rm -rf /etc/systemd/system/*
     systemctl preset-all
+  # We don't want systemd-firstboot.service. It conceptually conflicts with
+  # Ignition.
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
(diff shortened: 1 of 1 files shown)


fedora-coreos-base.yaml
```diff
@@ -92,6 +92,12 @@ postprocess:
     set -xeuo pipefail
     rm -rf /etc/systemd/system/*
     systemctl preset-all
+  # We don't want systemd-firstboot.service. It conceptually conflicts with
+  # Ignition.
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    rm /usr/lib/systemd/system/sysinit.target.wants/systemd-firstboot.service
```
cgwalters: I'd also remove `/usr/bin/systemd-firstboot`.

jlebon: I left it in (as well as `systemd-firstboot.service`) so that one could still enable/run it manually if they wished. Not against completely removing it until there's a valid request though. Done below! :arrow_down:

dustymabe (approved):

ajeddeloh (approved):
LGTM

bgilbert (approved):

PR #95, opened 2019-05-17
Have the kernel conditionally disable SMT if vulnerable

bgilbert:
Requires https://github.com/coreos/coreos-assembler/pull/525.  Implements https://github.com/coreos/fedora-coreos-tracker/issues/181.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -1,7 +1,10 @@
 # This replaces image.ks
 # size is the target disk size in GB.
-# Currently, it is the only supported key.
 size: 8
+
+extra-kargs:
+    # Disable SMT on systems vulnerable to MDS or any similar future issue.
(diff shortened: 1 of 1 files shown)


image.yaml
```diff
@@ -1,8 +1,10 @@
 # This replaces image.ks
 # size is the target disk size in GB.
-# Currently, it is the only supported key.
 size: 8
 
+# Disable SMT on systems vulnerable to MDS or any similar future issue.
+extra-kargs: mitigations=auto,nosmt
```
jlebon: Might be nicer to have this be a list instead to make it cleaner to e.g. update and maintain context comments.

bgilbert:
Updated.

PR #96, opened 2019-05-17
Disable SSH password logins by default

bgilbert:
For https://github.com/coreos/fedora-coreos-tracker/issues/138.  We don't support `sshd_config` fragments yet, so the only way to override this with Ignition is by writing a `ConditionFirstBoot=true` unit `Before=sshd.service` that runs `sed` over `sshd_config`.  But it is _possible_ to override it, so let's start testing with this configuration.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -78,6 +78,14 @@ postprocess:
     set -xeuo pipefail
     sed -i 's/^AuthorizedKeysFile[[:blank:]]/#&/' /etc/ssh/sshd_config
     echo -e '\n# Read authorized_keys fragments written by Ignition and Afterburn\nAuthorizedKeysFile .ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn' >> /etc/ssh/sshd_config
+  # Disable SSH password logins by default
+  # Move to overlay once sshd_config fragments are supported
+  # https://github.com/coreos/fedora-coreos-tracker/issues/138
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    sed -Ei 's/^(PasswordAuthentication|PermitRootLogin)[[:blank:]]/#&/' /etc/ssh/sshd_config
+    echo -e '\n# Disable password logins by default\nPasswordAuthentication no\nPermitRootLogin prohibit-password' >> /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


fedora-coreos-base.yaml
```diff
@@ -78,6 +78,14 @@ postprocess:
     set -xeuo pipefail
     sed -i 's/^AuthorizedKeysFile[[:blank:]]/#&/' /etc/ssh/sshd_config
     echo -e '\n# Read authorized_keys fragments written by Ignition and Afterburn\nAuthorizedKeysFile .ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn' >> /etc/ssh/sshd_config
+  # Disable SSH password logins by default
+  # Move to overlay once sshd_config fragments are supported
```
ajeddeloh: Is this something that is coming to sshd?

bgilbert: https://github.com/coreos/fedora-coreos-tracker/issues/138#issuecomment-459188872 is also an option.

dustymabe:
LGTM.. but this does make my head hurt when thinking about upgrading to f31 where the default config will change and also updating to sshd_config fragments when that is available. i.e. are our existing systems going to get the changes or are they going to keep this config as shipped?

jlebon:
> i.e. are our existing systems going to get the changes or are they going to keep this config as shipped?

Hmm not sure I follow. It should be just like any other config changes, right? IOW, once fragments are supported, it should be safe to drop this snippet in favour of a fragment drop-in. The drop-in will be added on system updates, and unless users modified their `sshd_config` locally, the main config itself as well will be updated.

bgilbert:
Considering the cases:

- If we're still carrying this postprocess fragment by F31 we'll update it for the Fedora changes.  The net result should be:
    - no behavior change for users with unmodified configs (who will then get the new one), and
    - no behavior change for users with modified configs (who will keep the modified old one).
- The key point about config fragments is that the user can use them to override our defaults, so
    - we might choose to continue putting our changes in `sshd_config` even when fragments become available.
    - If we choose to move our changes to a fragment,
        - we're fine if the user has not customized their `sshd_config`.
        - If the user removes our changes and we then move them to a fragment, we could end up overriding their customizations, which would be bad.

So if we want to move our changes to a fragment, we should probably do it before the stable release.

PR #98, opened 2019-05-29
Remove some stuff from the image

bgilbert:
Pursuant to coreos/fedora-coreos-tracker#186.  Subsumes, I think, the relevant part of https://github.com/coreos/fedora-coreos-config/pull/54.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -73,6 +73,8 @@ remove-files:
   # We don't ship man(1) or info(1)
   - usr/share/info
   - usr/share/man
+  # Drop text docs too
+  - usr/share/doc
 
 
 # ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -65,6 +65,9 @@ remove-from-packages:
                   /etc/grub.d/12_menu_auto_hide,
                   /usr/lib/systemd/.*]
 
+  # Drop NetworkManager support for ifcfg files.
+  - [NetworkManager, /usr/lib64/NetworkManager/.*/libnm-settings-plugin-ifcfg-rh.so]
+
 
 # ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
 postprocess:
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -28,6 +28,8 @@ repos:
   - fedora
   - fedora-updates
   - fedora-coreos-pool
+# Be minimal
+recommends: false
 
 ignore-removed-users:
   - root
@@ -119,8 +121,9 @@ postprocess:
     systemctl preset-all
 
 packages:
-  # SELinux
+  # Security
   - selinux-policy-targeted
+  - polkit
   # System setup
   - ignition
   - afterburn
@@ -133,6 +136,8 @@ packages:
   # Networking
   - bridge-utils nfs-utils-coreos
   - NetworkManager hostname
+  - iproute-tc
+  - adcli
   # Static firewalling
   - iptables nftables iptables-nft iptables-services
   # Storage
@@ -142,7 +147,7 @@ packages:
   - xfsprogs e2fsprogs mdadm
   - cryptsetup
   # Time sync
-  - chrony
+  - chrony timedatex
   # Extra runtime
   - sssd shadow-utils
   - logrotate
@@ -150,6 +155,7 @@ packages:
   - sudo coreutils less tar xz gzip bzip2
   - nmap-ncat net-tools bind-utils
   - bash-completion
+  - vim-minimal
   # Moving files around and verifying them
   - rsync fuse-sshfs
   - gnupg2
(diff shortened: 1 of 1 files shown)


fedora-coreos-base.yaml
```diff
@@ -49,7 +49,8 @@ default-target: multi-user.target
 
 remove-from-packages:
   # We're not using resolved yet
-  - [systemd, /usr/lib/systemd/system/systemd-resolved.service]
+  - [systemd, /usr/lib/systemd/systemd-resolved,
```
ajeddeloh: we should see about getting these split out from the systemd package, but this is fine for now

ajeddeloh: Should we add networkd?

dustymabe:
Doesn't `NetworkManager` use `dnsmasq` ? 

```
$ rpm -qf /etc/NetworkManager/dnsmasq.d/
NetworkManager-1.12.6-5.fc29.x86_64
```

bgilbert:
@dustymabe For connection sharing, sure.  It's not a hard dependency (or even a recommends) of the package, though.

bgilbert:
Updated to remove networkd.

dustymabe:
> @dustymabe For connection sharing, sure. It's not a hard dependency (or even a recommends) of the package, though.

I know in the past it was [used by openshift-ansible](https://github.com/openshift/openshift-ansible/issues/9351) when installing/configuring openshift. If we need it we can add it back in the future?

dustymabe:
>   9815016         cracklib-dicts
>   1340771         criu

not sure if any of our container runtimes need criu 

>   108423          fuse-overlayfs

I think fuse-overlayfs is needed for rootless podman containers 

>   61391018        geolite2-city
>   3772462         geolite2-country
>   901443          gnupg2-smime
>   65039           grubby
>   42366           hardlink
>   409724          libbsd
>   36062           libmaxminddb
>   256502          libnet
>   554285          libsecret
>   69838           libsss_autofs
>   64075           libsss_sudo
>   337118          libxkbcommon
>   108859          mkpasswd
>   1256348         openssl

I feel like openssl is a swiss army knife and useful for a lot of different low level things.

>   262138          openssl-pkcs11
>   214262          pinentry
>   124595          podman-manpages
>   2535150         shared-mime-info
>   170814          slirp4netns

needed for networking in rootless podman

>   53300           sssd-nfs-idmap
>   157915          sssd-proxy
>   202650          systemd-bootchart
>   552654          trousers
>   730381          trousers-lib
>   5909991         xkeyboard-config

ajeddeloh:
> not sure if any of our container runtimes need criu

I don't see it as a dep on gentoo

> I feel like openssl is a swiss army knife and useful for a lot of different low level things.

Any reason those things can't happen in a container?

> I think fuse-overlayfs is needed for rootless podman containers
> (slirp4netns) needed for networking in rootless podman

Can confirm

dustymabe:
> > not sure if any of our container runtimes need criu
> 
> I don't see it as a dep on gentoo

yeah we can add it back if someone needs it

> 
> > I feel like openssl is a swiss army knife and useful for a lot of different low level things.
> 
> Any reason those things can't happen in a container?

It depends - it's kind of like trying to debug your dns resolution by running `dig` in a container. You need DNS in order to get the container that has `dig` in it. The only reason I bring this up is in a previous lifetime I worked for a company that grabbed encrypted artifacts (i.e. a provisioning script) from s3 on first boot and decrypted them using `openssl`. It's also useful for debugging when a machine can't grab content from a remote server and you need to verify the validity of the cert in the environment you ran it from. 

> 
> > I think fuse-overlayfs is needed for rootless podman containers
> > (slirp4netns) needed for networking in rootless podman
> 
> Can confirm

ajeddeloh:
> It depends - it's kind of like trying to debug your dns resolution by running dig in a container. You need DNS in order to get the container that has dig in it. The only reason I bring this up is in a previous lifetime I worked for a company that grabbed encrypted artifacts (i.e. a provisioning script) from s3 on first boot and decrypted them using openssl. It's also useful for debugging when a machine can't grab content from a remote server and you need to verify the validity of the cert in the environment you ran it from.

Hmm fair. We also ship it in CL. I'll defer to others on this one.

fedora-coreos-base.yaml
```diff
@@ -59,6 +74,17 @@ remove-from-packages:
...
                   /usr/lib/systemd/.*]
 
+  # Drop NetworkManager support for ifcfg files.
+  - [NetworkManager, /usr/lib64/NetworkManager/.*/libnm-settings-plugin-ifcfg-rh.so]
+
+
+remove-files:
+  # We don't ship man(1) or info(1)
+  - usr/share/info
+  - usr/share/man
```
jlebon: This is probably going to be unpopular, but any thoughts about keeping man pages? (But not `info/` or `doc/` though). You're essentially getting versioned documentation for your operating system for what... 10M? That's pretty good usefulness/size ratio. This has come up many times in RHELAH/FAH. Looking at man pages is pretty standard, and it's always a pain to have to switch away and lookup online (again, which might not even match the installed version).

jlebon: (Not pushing super strongly on this, but I think it's worth discussing at least.)

bgilbert: It's caused friction in CL too, and I agree that it's not that much space.  I'm inclined to drop them anyway for now to avoid creating compatibility constraints, and maybe re-add later.

bgilbert:
Okay, I've re-added fuse-overlayfs, slirp4netns, and openssl.  They add no additional dependencies of their own.

bgilbert:
This is a net 120 MB improvement in the size of `/usr`.

dustymabe:
> Okay, I've re-added fuse-overlayfs, slirp4netns, and openssl. They add no additional dependencies of their own.

+1

PR #101, opened 2019-06-11
Split out ref into `manifest.yaml`

jlebon:
Make `manifest.yaml` be the "stream identity". I.e. this is the thing
that will *not* change from merges across branches. Crucially, it
includes the `ref` and the `repos`.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -24,10 +24,6 @@ machineid-compat: false
 releasever: "30"
 automatic-version-prefix: "30"
 mutate-os-release: "30"
-repos:
-  - fedora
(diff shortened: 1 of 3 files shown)


dustymabe:
> I.e. this is the thing that will _not_ change from merges across branches


i.e. you're trying to put the one thing that will be different across branches into a separate file so that changes we make to other files can propagate without merge conflicts ?

jlebon:
> i.e. you're trying to put the one thing that will be different across branches into a separate file so that changes we make to other files can propagate without merge conflicts ?

Yup, exactly!

dustymabe (approved):

PR #102, opened 2019-06-13
README.md: rework with more information

jlebon:
1. link to main Fedora CoreOS website
2. no longer reference fedora-atomic repo
3. link to discussions away from repo
4. add note about repo structure

PR #104, opened 2019-06-13
manifest: set ref to bodhi-updates

jlebon:
This is the `bodhi-updates` stream. Right now, this is identical
otherwise to `testing-devel`, though in the future, the latter should
only be sourcing from `coreos-pool`, as opposed to the stable repos as
well.

### DIFF
--- a/fedora.repo
+++ b/fedora.repo
@@ -34,3 +34,28 @@ gpgcheck=1
 metadata_expire=6h
 gpgkey=file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-$releasever-primary
 skip_if_unavailable=False
+
+[fedora-modular]
+name=Fedora Modular $releasever - $basearch
+#baseurl=http://download.fedoraproject.org/pub/fedora/linux/releases/$releasever/Modular/$basearch/os/
+metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-modular-$releasever&arch=$basearch
+enabled=1
+#metadata_expire=7d
+repo_gpgcheck=0
+type=rpm
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
+skip_if_unavailable=False
+
+[fedora-updates-modular]
+name=Fedora Modular $releasever - $basearch - Updates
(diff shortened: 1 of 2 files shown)


dustymabe:
LGTM - should we wait for the #102 to merge first and rebase?

sinnykumari:
unless we are relying something from coreos-pool, should we remove `fedora-coreos-pool` repo from manifest.yaml? That way repos list will be equivalent to what should be for bodhi-updates.

dustymabe:
> unless we are relying something from coreos-pool, should we remove `fedora-coreos-pool` repo from manifest.yaml? That way repos list will be equivalent to what should be for bodhi-updates.

+1 - good catch sinny. This is one of our mechanical refs so it should just be pulling from fedora release and bodhi-updates repo.

jlebon:
Yeah, I left it in there (and thought I had mentioned it in the commit message, but only did for the opposite case of `testing-devel` and Bodhi repos), but basically: right now we're abusing the `coreos-pool` tag a bit by manually tagging things in to short-circuit Bodhi. I agree long-term this is what we want, but until then, we don't want to be in a situation where `bodhi-updates` pulls in older packages than `testing-devel`.

dustymabe:
> we don't want to be in a situation where `bodhi-updates` pulls in older packages than `testing-devel`.

i think that's going to be a reality for us for a while for a subset of packages (at least until we can automate some interactions and gating in bodhi a bit). My thoughts here are that we'd take the lockfile from the passive stream (`bodhi-updates`) and we have an overrides file in `testing-devel` where we add in our rpms that override those from the passive stream. These overrides rpms would be ones that we wanted to fast track (new features or backported fixes in rpms that were never going to bodhi).

jlebon:
Had a chat with @dustymabe, @sinnykumari, and @bgilbert about this. To summarize: we're keeping `bodhi-updates` away from `coreos-pool` so that the overrides in `testing-devel` truly represents the "diff" from Fedora stable. The downside (upside?) is that we need to keep a closer eye tracking our overrides, but we don't expect that many so it shouldn't be too much maintenance.

I've updated the PR to drop `coreos-pool` and add the modular repos! :arrow_up:

dustymabe (approved):

dustymabe:
LGTM

PR #105, opened 2019-06-19
overlay/systemd: add boot.mount unit

ajeddeloh:
Add boot.mount unit so the systemd gpt mount generator doesn't think the
efi system partition should be /boot.

### DIFF
--- a/overlay/usr/lib/systemd/system/boot-efi.mount
+++ b/overlay/usr/lib/systemd/system/boot-efi.mount
@@ -1,0 +1,11 @@
+[Unit]
+Description=EFI System Partition
+Documentation=https://github.com/coreos/fedora-coreos-config
+ConditionPathExists=/boot/efi
+
+[Mount]
+What=/dev/disk/by-label/EFI-SYSTEM
+Where=/boot/efi
+
+[Install]
+RequiredBy=local-fs.target
--- a/overlay/usr/lib/systemd/system/boot.mount
+++ b/overlay/usr/lib/systemd/system/boot.mount
@@ -1,0 +1,9 @@
+[Unit]
+Description=Boot partition
+Documentation=https://github.com/coreos/fedora-coreos-config
+# This prevents the systemd-gpt-auto-generator from generating a mount unit for
+# /boot using the ESP.
+
+[Mount]
+What=/dev/disk/by-label/boot
+Where=/boot
--- a/overlay/usr/lib/systemd/system-preset/42-coreos.preset
+++ b/overlay/usr/lib/systemd/system-preset/42-coreos.preset
@@ -9,3 +9,5 @@ enable afterburn-checkin.service
 enable afterburn-firstboot-checkin.service
 # Service to write SSH key snippets from cloud providers.
 enable afterburn-sshkeys@.service
+# mount /boot/efi
+enable boot-efi.mount


cgwalters:
Related: https://github.com/coreos/fedora-coreos-config/pull/39

cgwalters:
LGTM!  Just one minor thing, probably worth a comment at the top of the file saying it came from fedora-coreos-config and probably basically the same text as the commit message?

ajeddeloh:
Fixed.

jlebon:
LGTM!
Do we also need a mount unit for the EFI partition?

overlay/usr/lib/systemd/system/boot.mount
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Boot partition
+# From fedora-coreos-config
+# This prevents the systemd-gpt-auto-generator from generating a mount unit for
+# /boot using the ESP.
```
jlebon, line 5: (Just comparing to the auto-generated `boot.mount` on my system). I think we may want `Before=local-fs.target` here, though it may be implied.

Also, instead of `# From fedora-coreos-config`, WDYT about `Documentation=https://github.com/coreos/fedora-coreos-config` ?

ajeddeloh: No need for the `Before=`
https://www.freedesktop.org/software/systemd/man/systemd.mount.html#Default%20Dependencies

+1 to the doc

jlebon: Yeah, that's what I thought. I guess the fstab generator just plops it in to be explicit. Cool with omitting it here I suppose.

cgwalters: I like the `Documentation=` idea.

ajeddeloh:
Ok, added `Documentation=` lines, `boot-efi.mount`, pulled in the correct target, added the preset to enable it, and made the EFI one conditional on `/boot/efi` existing to hopefully not break ppc64le.

Tested on x86_64 with the rm-anaconda PR.

overlay/usr/lib/systemd/system/boot-efi.mount
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=EFI System Partition
+Documentation=https://github.com/coreos/fedora-coreos-config
+ConditionPathExists=/boot/efi
```
cgwalters, line 4: How about `ConditionPathExists=/sys/firmware/efi/efivars`  as used by `/usr/lib/systemd/system/dbxtool.service` ?

ajeddeloh: We still want to mount it on combined UEFI+BIOS image for consistency, yes? We want to key off what image we're using instead of what mode we booted in.

cgwalters: Hmm.  What value would it be to mount the ESP on BIOS?

ajeddeloh: In practice I can't think of any. But I was to preserve consistency. One of the things I like from CL is things largely happen the same on different platforms. That's one of the reasons why we're shipping one image to all the clouds. Ultimately this should be invisible to users.

I don't feel super strongly on this but would be curious what other historically CL/Atomic folks thought. cc @bgilbert @jlebon ?

jlebon: No strong feelings either. I'm not sure of the ramifications either way. (Though related, I was just this morning catching up on the latest posts on the Silverblue board, and I saw [this post about trying to install both BIOS and UEFI](https://discussion.fedoraproject.org/t/install-grub2-to-support-bios-and-uefi-at-the-same-time), which I see you've already replied to. :) -- I guess in such a case, there is some value in having it mounted so you can e.g. [update ESP files](https://github.com/ostreedev/ostree/pull/1873) even if the current machine you're on is BIOS. That use case doesn't translate quite as well for FCOS though...)

jlebon: Hmm, in that case though, why do we need this conditional at all? systemd already knows to order `boot.mount` before `boot-efi.mount`, right? Ahh, or this is to account for the fact that https://github.com/coreos/coreos-assembler/pull/556 doesn't yet handle metal BIOS images as well?

ajeddeloh: Well that, but that should be fixed soon. I was more thinking of arches other than x86_64 and arm64 where there is no EFI.

overlay/usr/lib/systemd/system/boot-efi.mount
```diff
@@ -0,0 +1,11 @@
...
+
+[Mount]
+What=/dev/disk/by-label/EFI-SYSTEM
+Where=/boot/efi
+
+[Install]
+RequiredBy=local-fs.target
```
jlebon, line 11: Do we need this and the preset blurb for `boot.mount` as well?
Could also statically enable those by dropping a symlink in the overlay.

ajeddeloh: I'm not sure why we don't need it in `boot.mount` (which kind of scares me), but we don't in CL either. Maybe something about the mounts under `/` are special?

:+1: to static enablement.

ajeddeloh: Actually I'm not sure how I feel about static enablement anymore. It would either need to go in `/usr` which would not be disable-able aside from masking or in /etc which isn't where vendor-provided stuff should live. Plus don't we have https://github.com/coreos/fedora-coreos-config/blob/testing-devel/fedora-coreos-base.yaml#L124 ?

jlebon: Not to rehash the whole convo in #77, but yup presets are fine too. I guess I see those mounts as part of the OS (i.e. can't update without it), so you wouldn't want to disable it, unless you *really* know what you're doing (in which case having to resort to `mask` seems acceptable to me).

ajeddeloh: just a note: we can update without it, even on uefi. The only thing in `/boot/efi` is grub itself. Grub is configured to use `/boot` regardless of whether it's bios or uefi.

jlebon:
OK, let's merge this to unblock the Anaconda work. We can do tweaks as follow-ups!

overlay/usr/lib/systemd/system/boot-efi.mount
jlebon: > we can update without it, even on uefi

Right, speaking more to `/boot` in general. Well, we still could without a unit, but we'd have to teach OSTree to mount and and then unmount or something.

jlebon:
So oddly, the metal BIOS image for me right now hangs on trying to mount the ESP, which of course fails because Anaconda didn't create one. It's like the condition isn't inhibiting it. Also tried #106 with the same result, but haven't really debugged further. I guess this'll be fixed once we stop using Anaconda for the metal images as well, though we'll probably need to get to the bottom of this if we do decide we don't want to mount the ESP on BIOS systems but systemd disregards our conditionals.

PR #106, opened 2019-06-20
DNM: overlay/systemd/boot-efi.mount: use efivars

ajeddeloh:
Only mount /boot/efi if we're running on EFI

see discussion here:
https://github.com/coreos/fedora-coreos-config/pull/105#pullrequestreview-252017906

Marking DNM since we don't have consensus on it yet

### DIFF
--- a/overlay/usr/lib/systemd/system/boot-efi.mount
+++ b/overlay/usr/lib/systemd/system/boot-efi.mount
@@ -1,7 +1,7 @@
 [Unit]
 Description=EFI System Partition
 Documentation=https://github.com/coreos/fedora-coreos-config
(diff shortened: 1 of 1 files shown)


cgwalters:
(Rather than DNM you can now on GH mark PRs as drafts btw)

Anyways LGTM!

ajeddeloh:
Thinking about this more, once https://github.com/coreos/coreos-assembler/pull/563 merges there shouldn't be any reason to mount the ESP at all. How do you feel about just not mounting it ever?

cgwalters:
This will be obsoleted by https://github.com/coreos/fedora-coreos-config/pull/155

PR #109, opened 2019-07-05
fedora.repo: use dl.fp.o directly and add modular repos

jlebon:
See individual commit messages for details.

### DIFF
--- a/fedora.repo
+++ b/fedora.repo
@@ -1,8 +1,8 @@
 [fedora]
 name=Fedora $releasever - $basearch
 failovermethod=priority
-#baseurl=https://dl.fedoraproject.org/pub/fedora/linux/releases/$releasever/Everything/$basearch/os/
-metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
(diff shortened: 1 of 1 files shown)


rfairley (approved):
LGTM. Makes sense adding the repos here, but only enabling them in `manifest.yaml` in `bodhi-updates`, so `testing-devel` only includes modular content that was tagged in `coreos-pool`.

sinnykumari:
This change will affect arches like ppc64le because baseurl differs for arches which are considered secondary arches in Fedora. baseurl for secondary arch is https://dl.fedoraproject.org/pub/fedora-secondary while mainstream is https://dl.fedoraproject.org/pub/fedora/ . That's the main reason we switched to using metalink in https://github.com/coreos/fedora-coreos-config/commit/6c24e2eb24bee7e65d84310c11a4c0e363bfb9c1

jlebon:
> This change will affect arches like ppc64le because baseurl differs for arches which are considered secondary arches in Fedora

Ouch, thanks for catching this!
OK, let's revert this for now: https://github.com/coreos/fedora-coreos-config/pull/111

PR #110, opened 2019-07-08
manifest: inject stream name into commit metadata

### LINKED ISSUE #163: Stream introspection and forced upgrades (stream-switching)
This is a spin-off of one of the topics from https://github.com/coreos/fedora-coreos-tracker/issues/83#issuecomment-467016083. See bottom of https://meetbot.fedoraproject.org/teams/fedora_coreos_meeting/fedora_coreos_meeting.2019-03-13-16.32.log.html for context.

We should decide, implement and document the following things:
 * how to inject stream label into OS content via coreos-assembler
 * how to retrieve the above information (read-only) on a running node
 * how users would perform a manual operation to switch stream

This is because the update story for FCOS is slightly different than it used to be on CL. In particular, for a node to update from current OS version to a new release, there must be an edge in the cincinnati graph going `current->new`. However, cross-stream edges are problematic as they would bridge different upgrades path, and nodes could end up erratically jumping across streams.

As such, we will likely start with nodes introspecting their own stream and sticking to that for upgrades. For a node to switch stream, a manual `rpm-ostree upgrade` plus reboot would be required.

jlebon:
This is required for node introspection by client apps like Zincati.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/163

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -5,3 +5,6 @@ repos:
   - fedora
   - fedora-updates
   - fedora-coreos-pool
+
+add-commit-metadata:
+  fedora-coreos.stream: testing-devel


jlebon:
Requires: https://github.com/projectatomic/rpm-ostree/pull/1865

bgilbert (approved):
LGTM

jlebon:
OK, latest rpm-ostree is in continuous pool now. Merging!
I triggered a new cosa build in quay to pick it up. Or for local dev, you can of course `dnf update` too.

lucab:
The consuming side in Zincati is up at https://github.com/coreos/zincati/pull/67.

PR #111, opened 2019-07-09
fedora.repo: go back to using metalink

jlebon:
The baseurl doesn't apply equally across all basearchs, so let's go back
to using metalink for now. I think we could get what we want using
custom mirrorlists, though want to explore this space a bit more first.

### DIFF
--- a/fedora.repo
+++ b/fedora.repo
@@ -1,8 +1,11 @@
+# Note we use metalink= here to abstract over the base URLs being different
+# between primary and secondary architectures.
+
 [fedora]
 name=Fedora $releasever - $basearch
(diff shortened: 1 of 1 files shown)


rfairley (approved):
LGTM

fedora.repo
```diff
@@ -1,8 +1,11 @@
+# Note we use metalink= here to abstract over the base URLs being different
+# between primary and secondary architectures.
```
sinnykumari, line 2: Thanks for adding the comment!

sinnykumari (approved):
LGTM

PR #116, opened 2019-07-12
fedora.repo: go back again to using dl.fp.o directly

jlebon:
This is take two on #109. What's different this time is that we just
include *both* the primary and secondary `dl.fp.o` baseurls. The stack
knows how to handle this and just tries one before going to the next.

### DIFF
--- a/fedora.repo
+++ b/fedora.repo
@@ -1,11 +1,12 @@
-# Note we use metalink= here to abstract over the base URLs being different
-# between primary and secondary architectures.
+# Note we use baseurl= here because using auto-selected mirrors conflicts with
+# change detection: https://github.com/coreos/fedora-coreos-pipeline/issues/85.
 
 [fedora]
 name=Fedora $releasever - $basearch
 failovermethod=priority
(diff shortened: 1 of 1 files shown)


jlebon:
I tested this worked on my x86_64 machine by swapping the order of the baseurls and checking that rpm-ostree correctly falls back to the second baseurl.

I sanity-checked the URLs themselves with these commands:

```
grep -o 'https://dl.fedoraproject.org/.*$' fedora.repo | grep fedora/linux | sed 's/$releasever/30/; s/$basearch/x86_64/; s,$,repodata/repomd.xml,' | xargs curl -fI
grep -o 'https://dl.fedoraproject.org/.*$' fedora.repo | grep secondary | sed 's/$releasever/30/; s/$basearch/ppc64le/; s,$,repodata/repomd.xml,' | xargs curl -fI
```

sinnykumari (approved):
Nice! LGTM

sinnykumari:
Tested it on ppc64le machine by running `dnf update` with baseurl from this patch /etc/yum.repos.d/fedora-updates.repo and commented out metalink. Update went as expected.

PR #118, opened 2019-07-16
overlay: tweak experimental motd for preview

jlebon:
Let's convey a message that inspires slightly more confidence than
'highly experimental', while still making it clear things may break.


mike-nguyen (approved):
LGTM

sinnykumari (approved):
LGTM

PR #119, opened 2019-07-17
fedora-coreos-base.yaml: nuke systemd-gpt-auto-generator

### LINKED ISSUE #215: OSTree finalization on shutdown errors out
The latest FCOS preview release (30.20190716.0) seems to have an issue with OSTree finalization on shutdown:

```
Jul 16 18:19:02 localhost ostree[1914]: error: Unexpected state: /run/ostree-booted found and in / sysroot but not in a booted deployment
Jul 16 18:19:02 localhost systemd[1]: ostree-finalize-staged.service: Control process exited, code=exited, status=1/FAILURE
Jul 16 18:19:02 localhost systemd[1]: ostree-finalize-staged.service: Failed with result 'exit-code'.
Jul 16 18:19:02 localhost systemd[1]: Stopped OSTree Finalize Staged Deployment.
```

Still investigating this, but offhand it seems like `/boot` is being unmounted *before* `ostree-finalize-staged.service` "shuts down", which doesn't make sense since we have `After=local-fs.target`.

The workaround is manually calling `ostree admin finalize-staged` before restarting.

jlebon:
Otherwise, `systemd-gpt-auto-generator`'s generated `boot.automount`
will cause a funky mount structure:

```
├─/boot           systemd-1    autofs
│ └─/boot         /dev/vda1    ext4
│   └─/boot/efi   /dev/vda2    vfat
```

The nature of `autofs` then causes issues for us around shutdown
ordering. See also the linked systemd issue.

Really, we don't need the GPT generator at all for now, so just nuke it
from the tree.

Closes: coreos/fedora-coreos-tracker#215

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -63,6 +63,10 @@ remove-from-packages:
               /usr/lib/systemd/system/systemd-resolved.service,
               /usr/lib/systemd/system/sysinit.target.wants/systemd-firstboot.service]
 
+  # We don't want auto-generated mount units. See also
+  # https://github.com/systemd/systemd/issues/13099
+  - [systemd-udev, /usr/lib/systemd/system-generators/systemd-gpt-auto-generator]
+
   # The grub bits are mainly designed for desktops, and IMO haven't seen
   # enough testing in concert with ostree. At some point we'll flesh out
   # the full plan in https://github.com/coreos/fedora-coreos-tracker/issues/47
--- a/overlay/usr/lib/systemd/system/boot.mount
+++ b/overlay/usr/lib/systemd/system/boot.mount
@@ -1,8 +1,6 @@
 [Unit]
 Description=Boot partition
 Documentation=https://github.com/coreos/fedora-coreos-config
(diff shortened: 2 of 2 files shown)


fedora-coreos-base.yaml
```diff
@@ -63,6 +63,10 @@ remove-from-packages:
               /usr/lib/systemd/system/systemd-resolved.service,
               /usr/lib/systemd/system/sysinit.target.wants/systemd-firstboot.service]
 
+  # We don't want auto-generated mount units. See also
+  # https://github.com/systemd/systemd/issues/13099
+  - [systemd-udev, /usr/lib/systemd/system-generators/systemd-gpt-auto-generator]
```
ajeddeloh, line 68: LGTM, but why are generators in the udev package?

ajeddeloh (approved):
LGTM

fedora-coreos-base.yaml
jlebon: Not sure. Not all of them are in there, just a few.

PR #120, opened 2019-07-18
Add coreos-update-ca-trust.service

cgwalters:
This blends well with Ignition, allowing people to just drop
in certs without having to worry about injecting their own
systemd service to update them.

We also want this for RHCOS.
See: https://github.com/openshift/machine-config-operator/issues/528
For FCOS, in theory this could be a "run at most once" service.
But for RHCOS, since we support dynamic changes, we need it to
run on each boot too.  The cost of doing so is very small.

### DIFF
--- a/overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
+++ b/overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
@@ -1,0 +1,21 @@
+# This service is currently specific to Fedora CoreOS,
+# but we may want to add it to the base OS in the future.
+# The idea here is to allow users to just drop in CA roots
+# via Ignition without having to know to run the special
+# update command.
+[Unit]
+Description=Run update-ca-trust
+ConditionFirstBoot=true
+ConditionDirectoryNotEmpty=/etc/pki/ca-trust/source/anchors/
+# We want to run quite early, in particular before anything
+# that may speak TLS to external services.  In the future,
+# it may make sense to do this in the initramfs too.
+DefaultDependencies=no
+
+[Service]
+ExecStart=/usr/bin/update-ca-trust extract
+Type=oneshot
+RemainAfterExit=yes
+
+[Install]
+WantedBy=basic.target
--- a/overlay/usr/lib/systemd/system-preset/42-coreos.preset
+++ b/overlay/usr/lib/systemd/system-preset/42-coreos.preset
@@ -2,6 +2,8 @@
 enable coreos-growpart.service
 enable console-login-helper-messages-issuegen.service
 enable console-login-helper-messages-motdgen.service
+# CA certs (probably to add to base fedora eventually)
+enable coreos-update-ca-trust.service
 # This one is from https://github.com/coreos/ignition-dracut
 enable ignition-firstboot-complete.service
 # Boot checkin services for cloud providers.


overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
```diff
@@ -0,0 +1,20 @@
...
+# but we may want to add it to the base OS in the future.
+# The idea here is to allow users to just drop in CA roots
+# via Ignition without having to know to run the special
+# update command.
+[Unit]
+Description=Run update-ca-trust
+ConditionDirectoryNotEmpty=/etc/pki/ca-trust/source/anchors/
+# We want to run quite early, in particular before anything
+# that may speak TLS to external services.  In the future,
+# it may make sense to do this in the initramfs too.
```
ajeddeloh, line 12: WRT the initramfs, I'm not sure it's needed. The only things in there I can think of that might need TLS are Ignition (which supports custom certs in the `ignition` section) and afterburn, which is talking to cloud providers.

ajeddeloh:
Half of me wants this as FCCT sugar instead since that would prevent any future compatibility constraints. If we ship it in the OS we need to be _very_ careful about how we update it to not break systems. On the other hand this is pretty simple.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
cgwalters: Right https://github.com/coreos/ignition/pull/636 is about this, but I can imagine people wanting to have the ones provided to Ignition persist for the host too.

But, this doesn't conflict with that; if Ignition writes it optionally, this service could read it.

cgwalters:
What updates are you thinking of that we might do that could break something?  Is there anything similar in CL that was changed and broke something?

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
```diff
@@ -0,0 +1,20 @@
...
+DefaultDependencies=no
+
+[Service]
+ExecStart=/usr/bin/update-ca-trust
```
jlebon: I was perusing `update-ca-trust(8)`, and found:

> COMMANDS
>   (absent/empty command)
>   Same as the extract command described below. (However, the command may print fewer warnings, as this command is being run during rpm package installation, where non-fatal status output is undesired.)

So let's use `extract`?

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
```diff
@@ -0,0 +1,20 @@
...
+# The idea here is to allow users to just drop in CA roots
+# via Ignition without having to know to run the special
+# update command.
+[Unit]
+Description=Run update-ca-trust
+ConditionDirectoryNotEmpty=/etc/pki/ca-trust/source/anchors/
+# We want to run quite early, in particular before anything
+# that may speak TLS to external services.  In the future,
+# it may make sense to do this in the initramfs too.
+DefaultDependencies=no
```
jlebon, line 13: This needs `ConditionFirstBoot=yes` too, right? I don't think we want users to rely on reboots magically having any new certs they added be extracted.

jlebon (commented):
This looks reasonable to me. The mechanic here is pretty clean. Just two comments.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
cgwalters: If we did this, we wouldn't be able to use this service for the MCD.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
cgwalters: I'm not seeing an obvious difference playing with it, but sure; done.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
ajeddeloh: MCD doesn't run on FCOS right now. MCD could also add it's own drop in for turning `ConditionFirstBoot=no` if needed.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
ajeddeloh: I'm not sure I 100% follow? Ignition does _not_ write them to the initramfs root, it adds them to its HTTP client directly from the config. I also think it would be a little misleading for Ignition to write them out to the initramfs root since the `ignition` section is just for configuring Ignition, not the initramfs in general. I still don't see a use case for this either.

If people want them to persist to the host they should add `storage.files` entries. We could also add sugar to FCCT for this that generates both.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
cgwalters: Fine.  Done.

ajeddeloh:
Actually the `ConditionFirstBoot=` is a good example of something that might change and break people on upgrade. User's shouldn't have that behavior change out from under them.

cgwalters:
> Actually the ConditionFirstBoot= is a good example of something that might change and break people on upgrade. User's shouldn't have that behavior change out from under them.

If...we decided to drop the condition?  Spell this out more?

jlebon (approved):
LGTM! Will let @ajeddeloh do the final merge.

ajeddeloh:
> If...we decided to drop the condition? Spell this out more?

Yeah, or add other conditions, dependencies, etc. Basically I'm hesitant to add configuration for things that users could write just as easily, or that can be generated by FCCT. In that case we basically make FCCs always generate the same config given that same config version and if we want to make changes we bump the config version. Basically I'm not sure if this should be our responsibility to manage/maintain.

Curious if @bgilbert has thoughts. I feel like we've had a similar discussion for something in CL before but I don't remember what.

ajeddeloh:
In fact, if we added it to FCCT, we could make it autogenerate this unit when they want CAs in the host as well as ignition. e.g. something like:

```yaml
certs:
  - name: foo
  - contents: <foo>
  - ignition: true
  - host: true
```

could generate that unit, enable it, add the CAs to the `ignition` section, and add the files entries. Still easy for the end user, but the behavior is locked to that FCCT spec version.

cgwalters:
That's presuming the only thing generating Ignition for the node is FCCT.  Since part of the plan is to have e.g. OKD target FCOS, and the installer today generates Ignition too, we would need to merge them.  Literally in fact this same time we're discussing this someone added a duplicate of this service to the installer, which would race.  It seems more like a fundamental base OS feature - the concerns you have about compatibility seem too vague to me to outweigh the utility of having it in the base OS.

ajeddeloh:
> That's presuming the only thing generating Ignition for the node is FCCT.

It's presuming there is only one tool generating the config, not necessarily FCCT, just not FCCT in tandem with another tool.

> which would race

I don't think that race situation is limited to just this case; any other programs a user might want to only run once on first boot would also suffer this problem. Any time you have two tools that don't know about each other both generating Ignition configs then you could hit similar problems. I think it'd be better to ensure that either there's only one thing generating the Ignition config or that all the pieces that generate configs are aware of each other. I guess this gets at the larger question of "how will OKD configuration work". 

side note: if they use the same unit name then the units will be merged (child's config's contents replacing the parent's).

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
crawford: People have asked that Ignition persist the CA to disk, but I'm very much against this. The `ignition` section of the config directs Ignition; nothing more.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
crawford: Have we verified that systemd honors conditions in drop-ins? I seem to recall this not working in the past.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
cgwalters: Right, no one is talking about changing that default, just adding a new mechanism to do both.

bgilbert:
Putting the unit file in FCCT wouldn't remove compatibility constraints, just move them around.  For example, `/usr/bin/update-ca-trust extract` would need to exist forever to support older versions of FCCT, and that constraint would be invisible because it'd only be in the FCCT Git history.  I also think @cgwalters is correct that we can't assume that all configuration comes through FCCT.  So I think I'm generally in favor of keeping non-trivial code in the distro rather than in FCCT.

ajeddeloh:
Hrm that's a good point, though in general things like binary paths are also potential compat issues. I'm fine with this.

> So I think I'm generally in favor of keeping non-trivial code in the distro rather than in FCCT.

I think it's fine to have non-trivial code in FCCT as long as it's using things we know have strong compat guarantees. For example, writing out an `etcd.service` unit would be complex, but it's just using docker/podman which we need to do a good job maintaining compatibility of anyway since pretty much everyone uses it. If we want to discuss that more we should move the conversation over to the FCCT repo instead.

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
cgwalters: In playing with this, it seems to work to do `ConditionFirstBoot=` but *not* `ConditionFirstBoot=no`.  Dunno exactly why, maybe something about "reset semantics" vs override?

lucab:
For reference:
 * the corresponding CL service: https://github.com/coreos/coreos-overlay/blob/master/app-misc/ca-certificates/files/update-ca-certificates.service
 * some friction with it: https://github.com/coreos/bugs/issues/2301
 * some further nuances on paths and OpenSSL: https://github.com/coreos/bugs/issues/2262

cgwalters:
> the corresponding CL service: https://github.com/coreos/coreos-overlay/blob/master/app-misc/ca-certificates/files/update-ca-certificates.service

Thanks, that's useful!

> some friction with it: coreos/bugs#2301

Hm...without diving in a lot there, I'm guessing what happened is that at some point the bundle started to be generated in the base image, which defeated the point of the unit.  Here we're using `ConditionFirstBoot=`.

(Though, as an aside...`ConditionFirstBoot` has deep issues around the semantics if the machine crashes on first boot...we're often using it for "run exactly once" semantics but in that scenario it'll be violated.)

> some further nuances on paths and OpenSSL: coreos/bugs#2262

Again offhand...that looks likely to be fallout from some distribution misconfiguration.  The OpenSSL is definitely configured to read the bundle `update-ca-trust` generates.  (BTW, the path here ends up being the "Fedora/RH" branch of `/etc/pki` versus the Debian-ish `/etc/ssl`.)

cgwalters:
Rebased :surfing_man: 

Though as of this moment we merged into RHCOS a variant of this unit without `ConditionFirstBoot` - but as soon as we make some forward progress on having RHCOS use this repo as a `git submodule` we can reconcile things by having RHCOS ship a drop-in.

lucab:
I don't want to derail this PR, but I do wonder if longer term we would/should have something akin to `systemd-sysctl` for CAs.
That is, fragments (here: certificates) that are re-assembled and applied (here: bundled/hashed) on each boot, and optionally the same refresh task can be manually triggered by a human OP at any point at runtime (we would discourage this, in our immutable case).

cgwalters:
> That is, fragments (here: certificates) that are re-assembled and applied (here: bundled/hashed) on each boot, and optionally the same refresh task can be manually triggered by a human OP at any point at runtime (we would discourage this, in our immutable case).

How would that be different from dropping the `ConditionFirstBoot` here, and "manually trigger" = `systemctl start coreos-update-ca-trust.service`?

lucab:
@cgwalters yes, that was implicit point in favor of dropping the first-boot condition, I guess.
However:
 * I'm not very familiar with fedora PKI tooling, so I don't know what are the semantics/side-effects of the command we run in this unit
 * in its current form, I think this has some expectations on using /etc has a target for the generated bundle

JAORMX (approved):

overlay/usr/lib/systemd/system/coreos-update-ca-trust.service
JAORMX: When a user adds a new cert via machineconfig, wouldn't that restart the hosts anyway? So, in that case, we would actually want this to run on firstboot only.

cgwalters:
> I'm not very familiar with fedora PKI tooling, so I don't know what are the semantics/side-effects of the command we run in this unit

It updates the pre-built bundle that TLS libraries like OpenSSL, gnutls, golang's, etc. know how to read.

> in its current form, I think this has some expectations on using /etc has a target for the generated bundle

Right, that's where the bundle lives.  What issue are you thinking of?

cgwalters:
How do we drive this to consensus?  Options seem to be:

 - Do nothing
 - Retain as is
 - Drop `ConditionFirstBoot` (identical with RHCOS)

Based on discussion so far...I don't have a strong opinion between current and RHCOS-identical; there are arguments both ways that just leave me feeling "eh".  (Argument for retain as is is that if someone wants it to run manually they can just invoke `update-ca-trust` manually too)  I do think we should ship this though in some form per above.

sferich888:
Which method is tested with more layered product offering expectations? 
IE what method is going to be what customers expect, to see us doing, so that they can leverage what we have done and extended upon this (model) the simplest?

ajeddeloh:
I'm in favor of shipping with the conditional. It matches with the "Set everything up on first boot" scheme. If users want to mask/disable it, they can. If they want to run every boot, they can disable the condition by setting it to the default or write their own unit. In fact, if they're writing files there after first boot, whatever is writing those files should probably own running update-ca-trust or setting it up to run.

jlebon:
> I don't want to derail this PR, but I do wonder if longer term we would/should have something akin to `systemd-sysctl` for CAs.

Yeah, I think that could make sense, though it seems safer to me to drive a change like this in the rest of Fedora too (and in the process shake out any other repercussions this may have for crypto libraries). For now, it's much easier to commit to this API with `ConditionFirstBoot=true` than without, as a natural complement to Ignition.

> I'm in favor of shipping with the conditional.

Also in favour. Let's give it a bit more time and if no new objections are raised, let's merge it!

PR #121, opened 2019-07-19
Add udev rule to ignore Azure SR-IOV interface

cgwalters:
Pulled from a downstream RHCOS change originally authored
by Jerry Zhang.  I would like RHCOS to directly inherit
from FCOS's config, and having this upstream helps that.

Slightly tweaked original commit message follows:

Add udev rule to mark the SR-IOV interface on Azure to be unmanaged;
otherwise it will bond to the synthetic interface by default which
causes networking issues.

### DIFF
--- a/overlay/usr/lib/udev/rules.d/68-azure-sriov-nm-unmanaged.rules
+++ b/overlay/usr/lib/udev/rules.d/68-azure-sriov-nm-unmanaged.rules
@@ -1,0 +1,4 @@
+# Accelerated Networking on Azure exposes a new SRIOV interface to the VM.
+# This interface is transparently bonded to the synthetic interface,
+# so NetworkManager should just ignore any SRIOV interfaces.
+SUBSYSTEM=="net", DRIVERS=="hv_pci", ACTION=="add", ENV{NM_UNMANAGED}="1"


ajeddeloh (approved):
LGTM

overlay/usr/lib/udev/rules.d/68-azure-sriov-nm-unmanaged.rules
```diff
@@ -0,0 +1,4 @@
+# Accelerated Networking on Azure exposes a new SRIOV interface to the VM.
+# This interface is transparently bonded to the synthetic interface,
```
jlebon, line 2: Does `synthetic interface` here essentially mean the regular virtual network adapter? From reading [this](https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli#handle-dynamic-binding-and-revocation-of-virtual-function), I *think* that's what it means, but I hadn't heard it referred to as "synthetic" before.

jlebon: Ahh OK, it's a Hyper-V term: https://blogs.msdn.microsoft.com/virtual_pc_guy/2008/02/25/hyper-v-terminology/.

PR #122, opened 2019-07-20
Also systemctl --user preset-all

cgwalters:
Not doing so means that we miss enabling `dbus.service` for the user
session, which breaks GDM for Silverblue, among other things.  And
`systemctl --user` is also generally useful even on CoreOS systems.


ajeddeloh (approved):
> which breaks GDM for Silverblue

This isn't a concern for us here, is it? LGTM regardless, just curious.

jlebon (approved):

PR #126, opened 2019-07-30
manifest: Add systemd-container (nspawn)

### LINKED ISSUE #230: Ship systemd-nspawn
## Version
```
State: idle
AutomaticUpdates: disabled
Deployments:
● ostree://fedora:fedora/x86_64/coreos/testing
                   Version: 30.20190725.0 (2019-07-25T18:54:22Z)
                    Commit: 8b79877efa7ac06becd8637d95f8ca83aa385f89f383288bf3c2c31ca53216c7
              GPGSignature: Valid signature by F1D8EC98F241AAF20DF69420EF3C111FCFC659B9
```
## Enhancement

We don't ship the `systemd-container` package, which includes `systemd-nspawn`.  Should we add it?

cc @mrguitar

cgwalters:
Consensus was this should join our suite of container engines.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/230

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -154,7 +154,7 @@ packages:
   # SSH
   - openssh-server openssh-clients
   # Containers
-  - podman skopeo runc moby-engine
+  - podman skopeo runc moby-engine systemd-container
   - fuse-overlayfs slirp4netns
   # Networking
   - bridge-utils nfs-utils-coreos


jlebon (approved):
LGTM! Will give it some time for others to review.

ajeddeloh:
skimming through the package contents we ought to remove `/etc/systemd/network/*` (networkd units). Otherwise LGTM

bgilbert (approved):
LGTM

lucab:
@ajeddeloh do you mean `/usr/lib/systemd/network/*`? Otherwise, I think latest version of this package is not shipping anything in `/etc` anymore.

jlebon:
Ahh yup, I do see:

```
$ rpm -ql systemd-container | grep '.network$'
/usr/lib/systemd/network/80-container-ve.network
/usr/lib/systemd/network/80-container-vz.network
```

Hmm, we actually have a couple of other units in there too:

```
# ls -l /usr/lib/systemd/network
total 8
-rw-r--r--. 2 root root 645 Jan  1  1970 80-container-host0.network
-rw-r--r--. 2 root root 417 Jan  1  1970 99-default.link
# rpm -qf /usr/lib/systemd/network/*
systemd-241-8.git9ef65cb.fc30.x86_64
systemd-udev-241-8.git9ef65cb.fc30.x86_64
```

We should add those to `remove-from-packages`. Let's do that in a follow-up!

jlebon:
> We should add those to `remove-from-packages`. Let's do that in a follow-up!

https://github.com/coreos/fedora-coreos-config/pull/129

PR #128, opened 2019-07-31
Disable default NetworkManager plugins

bgilbert:
NetworkManager tries to load `ifcfg-rh` by default unless we tell it not to.

Fixes journal warning:

    settings: could not load plugin 'ifcfg-rh' from file '/usr/lib64/NetworkManager/1.16.2-1.fc30/libnm-settings-plugin-ifcfg-rh.so': No such file or directory


lucab:
@bgilbert assuming NM supports sourcing fragment in lexicographic order, can we put some kind of low numerical prefix in this (e.g. `10-disable-default-plugins.conf`) so that we can selectively enable some of them later on?

bgilbert:
@lucab Done.

ajeddeloh (approved):
LGTM

PR #129, opened 2019-07-31
manifest: drop more networkd units

jlebon:
Remove all networkd unit files that come from systemd. It'd be easier to
just use a postprocess script here to `rm -rf` everything, though being
able to keep track of which packages bring them in like this is nice.


ajeddeloh (approved):
LGTM

bgilbert (approved):
LGTM

PR #130, opened 2019-08-01
tree: promote changes from testing-devel

arithx:


jlebon (approved):

dustymabe:
in the future would it be useful to add info to this commit message that points back to the PRs or commits that were part of it ?

PR #131, opened 2019-08-02
ci: add Jenkins pipeline

jlebon:
Run all contributions through a `cosa build` and `cosa kola run` so we
know `git master` is always somewhat sane.


jlebon:
Ready for review!
(Of course, this will be even nicer once we actually use lockfiles).

cgwalters:
Let's try it!

cgwalters (approved):

PR #132, opened 2019-08-06
Split overlay into overlay.d

cgwalters:
Make use of the new `overlay.d` model:
https://github.com/ostreedev/ostree/pull/1894

I'm looking at rebasing RHCOS (and Silveblue) on this
repo.  Of the stuff in `overlay/` today, the thing that immediately
stands out as "really FCOS specific" is the experimental motd bit.
Move that into `overlay.d/80experimental`; everything else goes
into `overlay.d/05core`.


cgwalters:
Maybe in the future it'd make sense to split up the overlays into like `05mounts/`,  `08NetworkManager/`, ... or something but eh this is fine for now I think.

ajeddeloh:
LGTM though I think you meant to link to https://github.com/coreos/coreos-assembler/pull/639 instead.

ajeddeloh (approved):

cgwalters:
Fixed the link, thanks!

PR #134, opened 2019-08-06
overlay.d: Split up preset file

cgwalters:
RHCOS also has the name `42-coreos.preset` in a different place.
Trying to avoid having them conflict, let's both work around
the conflict and make things clearer by renaming/splitting the
preset files.

The FCOS-specific bit goes into `15fcos`.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -15,5 +15,3 @@ enable afterburn-sshkeys@.service
 enable boot-efi.mount
 # Update agent
 enable zincati.service
-# User metrics client
-enable fedora-coreos-pinger.service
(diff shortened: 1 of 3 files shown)


ajeddeloh:
This renames without modifying, so pinger is enabled in both. Was that intended?

cgwalters:
Good catch, fixed!

ajeddeloh (approved):
LGTM other than lets add a comment to each explaining what should go where so we don't add things to the wrong place in the future by accident.

cgwalters:
Good idea, prep in https://github.com/coreos/coreos-assembler/pull/691

cgwalters:
OK now with a `README.md`!

dustymabe (approved):
LGTM

dustymabe:
> RHCOS also has the name `42-coreos.preset` in a different place.

maybe in the future we could name the RHCOS version 42-rhcos.preset and put it in its own subdir so we prevent having another problem like this down the line.

PR #135, opened 2019-08-06
overlay.d: Move core user into 10coreuser

cgwalters:
RHCOS can't use this fragment because the Ignition version is
`3.0.0` (actually in practice it's compatible with 2x...in
theory we could translate but...)

With this RHCOS can inherit from the `05core`, and I did
a duplicate of `10coreuser` with the Ignition spec version changed.


dustymabe:
tested this and it works.. Now that #134 is merged could we add it to the readme and also maybe make it `10fcos-coreuser` so we know it's for fcos only.

cgwalters:
Rebased and added to `README.md` but 

>  and also maybe make it 10fcos-coreuser so we know it's for fcos only.

I kept it as `10coreuser` because it's not necessarily FCOS specific, e.g. Silverblue may use it?  Or may not.  Dunno.

dustymabe (approved):

Issue #136, opened 2019-08-08
/etc/skel files in /root should be copied by tmpfiles


ajeddeloh:
Having a systemd service to copy a file on first boot seems hacky. We ought to convert it to a tmpfiles rule and let tmpfiles handle it.

We could even just use symlinks since we're not modifying the files.

jlebon:
Oh neat, just saw the `C` entry type. Hadn't realized systemd-tmpfiles could do this. Yeah, I think we can and should do this.

jlebon:
> We could even just use symlinks since we're not modifying the files.

I don't think we want to do this though. First, it'd mean that all other user's homes are different from `/root` in that respect, but also a naive `vi /root/.bashrc` will actually end up modifying the template instead of just for the root user.

jlebon:
OK, now I see this is addressed in: https://github.com/coreos/fedora-coreos-config/pull/137.

jlebon:
Ahh OK, the symlinks as proposed in #137 are actually to `/usr/etc`, which makes this better. Though the UX of having to unlink and copy in order to edit one of those files is a bit odd. Having to hardcode the list of skeleton files is also unfortunate.

Hmm, actually another UX issue too is that one can't easily just append to the default config anymore. (See the commit message in https://github.com/coreos/fedora-coreos-config/pull/83#issue-272861713).

PR #137, opened 2019-08-08
DNM: overlay: drop coreos-root in favor of tmpfiles

ajeddeloh:
Instead of using a systemd unit to populate /var/roothome use tmpfiles
entries.

Untested


cgwalters:
LGTM if tested

ajeddeloh:
Aside from needing `/usr/etc/skel` instead of `/etc/skel`, I'm hitting https://github.com/systemd/systemd/issues/12467 which is fixed by https://github.com/systemd/systemd/pull/12750 but that hasn't made it's way into fcos yet. The relative path solution there also doesn't work because tmpfiles disallows relative sources and if we hardcode `/sysroot/<path>` then we'll break when we _do_ pick up the fix.

ajeddeloh:
retest this please

Ok, switched to symlinks. I intentionally used `/usr/etc/skel` instead of `/etc/skel` so people can't mistakenly edit the version the /etc by following the symlinks. Tested locally and it works fine.

arithx (approved):
Change LGTM (didn't test), once merged we should probably file a tracking issue for systemd 243 lands (to pick up systemd/systemd#12750) to update to `C!`

dustymabe:
> once merged we should probably file a tracking issue for systemd 243 lands (to pick up [systemd/systemd#12750](https://github.com/systemd/systemd/pull/12750)) to update to `C!`

@ajeddeloh can you file that tracking bug and then merge this PR?

jlebon:
Needs a `Closes: #136`, though I think we should hold this until we can use `C` instead of symlinks: https://github.com/coreos/fedora-coreos-config/issues/136#issuecomment-520606962

ajeddeloh:
Hmm I'm on the fence on holding off until we can use `C!`. If we don't, we ought to roll this into `coreos-populate-var.service` since it's part of populating `/var`.

ajeddeloh:
Opened https://github.com/coreos/fedora-coreos-config/pull/143 as a stopgap. Will rebase this once that merges, use C!, open a tracker ticket for it, and mark DNM until we can use C!.

travier:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/2556

PR #139, opened 2019-08-09
Split off removal of the NetworkManager ifcfg plugin

cgwalters:
It turned out to be a bit too hard for us to drop (15+?) years of
legacy for RHEL CoreOS.  Split this off so we can more easily
keep it.


dustymabe:
LGTM - testing now

cgwalters:
I tested this locally too FWIW.

dustymabe (approved):

PR #141, opened 2019-08-12
overlay.d/05core: Fix boot-efi.mount unit dependency

### LINKED ISSUE #115: "enable boot-efi.mount" make ppc64le platform fail
cosa run failed on ppc64le platform.
ppc64le qcow2 generated image does not end its boot and get stuck with no access to login.

The problem is due to the "enable boot-efi.mount" added in 
overlay/usr/lib/systemd/system-preset/42-coreos.preset  for all platforms.
see commit cb3f240
and also #105 and #106 pull requests

I didn't find a fair way to avoid it, please @ajeddeloh could you have a look ?

jcajka:
Fix #115

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/boot-efi.mount
+++ b/overlay.d/05core/usr/lib/systemd/system/boot-efi.mount
@@ -1,11 +1,11 @@
 [Unit]
 Description=EFI System Partition
 Documentation=https://github.com/coreos/fedora-coreos-config
-ConditionPathExists=/boot/efi
+ConditionPathExists=/sys/firmware/efi/
 
 [Mount]
 What=/dev/disk/by-label/EFI-SYSTEM
 Where=/boot/efi
 
 [Install]
-RequiredBy=local-fs.target
+WantedBy=local-fs.target


jcajka:
For the record tested only on ppc64le, working on other arches atm.

ajeddeloh (approved):
LGTM

dustymabe:
@jcajka - can you report back once you've tested on other arches?

jcajka:
> @jcajka - can you report back once you've tested on other arches?

Sure.

jlebon:
https://github.com/coreos/fedora-coreos-config/issues/115#issuecomment-520616849

bgilbert:
Won't this still try to pull in the device node, but just make the failure non-fatal?  Conditionalized units still pull in their dependencies even if the condition fails.  Avoiding that generally requires a systemd generator.  https://github.com/coreos/fedora-coreos-config/pull/155 moves `boot-efi.mount` to a generator and enables it only on architectures that support UEFI.

cgwalters:
Yeah, this will be obsoleted by https://github.com/coreos/fedora-coreos-config/pull/155

PR #142, opened 2019-08-13
overlay: re-enable 40coreos-network

yuqi-zhang:
Reanble 40coreos-network to run after ignition. When this service
was ported, it was no longer linked to ignition-complete.

(Was there a reason we disabled this? If so maybe we should remove
the service entirely. I will re-test behaviour on some platforms first)

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40coreos-network/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40coreos-network/module-setup.sh
@@ -11,4 +11,6 @@ install() {
     inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
     inst_script "$moddir/coreos-teardown-initramfs-network.sh" \
         "/usr/sbin/coreos-teardown-initramfs-network"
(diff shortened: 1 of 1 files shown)


jlebon:
> Was there a reason we disabled this?

AFAICT, I think it just was never enabled. In the spec2x branch on ignition-dracut, it was manually linked from the generator.

Have you been able to test this?

dustymabe:
@yuqi-zhang - I think you added this service in 24e952e - Is it needed? Can we move to the networkmanager based modules?

ajeddeloh:
I'm not sure if this is worth pursuing on fcos since we'll be moving to networkmanager soon. Thoughts?

yuqi-zhang:
In that case, thoughts on removing the module entirely?

ajeddeloh:
I assume you mean from master only since I'm pretty sure rhcos uses it. I'm fine with that. @bgilbert thoughts?

yuqi-zhang:
Opened https://github.com/coreos/fedora-coreos-config/pull/153 instead

PR #143, opened 2019-08-13
overlay/dracut: populate /root in coreos-pop-var

ajeddeloh:
Currently we have two units that populate var: coreos-populate-var and
coreos-root. The latter is just a cp /usr/etc/skel/.*bash*
/var/roothome. Merge that with coreos-populate-var since it's populating
var.

Ultimately this should be handled with systemd-tmpfiles but there's
currently a bug where the source for C! lines doesn't respect the --root
argument. See https://github.com/coreos/fedora-coreos-config/pull/137.


jlebon:
Yeah, having it in `coreos-populate-var.sh` makes sense to me. Thoughts about the other concerns raised in https://github.com/coreos/fedora-coreos-config/issues/136#issuecomment-520614094?

ajeddeloh:
We're doing a real copy not a link here. What concerns aren't related to the symlink?

jlebon:
OK right, sorry for the confusion here. I thought somehow your end goal was for real root `systemd-tmpfiles-setup.service` to handle the `C` entries. But right, it'll get picked up by the `systemd-tmpfiles` run we do here in the initrd, so it'd still get set up before ignition-files.

jlebon (approved):

PR #144, opened 2019-08-15
minimal.yaml: tweak GRUB2 EFI package name for aarch64

jlebon:
So that we match https://github.com/coreos/coreos-assembler/pull/710.

See also:
https://github.com/coreos/coreos-assembler/issues/660#issuecomment-520768829


ajeddeloh (approved):
LGTM

PR #146, opened 2019-08-19
s390x: add zipl parameter

tuan-hoang1:
This will be read by s390x zipl bootloader during
cosa buildextend-installer

### DIFF
--- a/installer/zipl.prm
+++ b/installer/zipl.prm
@@ -1,0 +1,1 @@
+rd.neednet=1 coreos.inst=yes


ajeddeloh:
Are we still building the installer with anaconda? The installer should just be the initrd and kernel from the normal build, which should have these args as a result of that, right?

tuan-hoang1:
This is for FCOS thus no anaconda.
Basically I just copy these 2 from isolinux files: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/installer/isolinux/isolinux.cfg#L68
`rd.neednet=1` might be already there as a result of a normal build, but `coreos.inst=yes` seems to be required here ?

cgwalters:
> Are we still building the installer with anaconda? 

We never did that (on any arch).

ajeddeloh:
> We never did that (on any arch).

Oh right, I was conflating the BM images and the installer.

ajeddeloh (approved):
LGTM

PR #149, opened 2019-08-20
manifest: Split grub2-tools removals into arch-specific manifest

cgwalters:
The package isn't present on s390x and previously we'd we'd error out.

This is really a hack; I think the better fix would be to move
that stuff out of the low-level `grub2-tools` package.

Requires: projectatomic/rpm-ostree#1886

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -4,6 +4,10 @@
 # https://github.com/coreos/coreos-assembler/pull/639
 
 include: minimal.yaml
+arch-include:
+  x86_64: grub2-removals.yaml
+  aarch64: grub2-removals.yaml
+  ppc64le: grub2-removals.yaml
 
 # Modern defaults we want
 boot-location: new
@@ -74,13 +78,6 @@ remove-from-packages:
   # https://github.com/systemd/systemd/issues/13099
   - [systemd-udev, /usr/lib/systemd/system-generators/systemd-gpt-auto-generator]
 
-  # The grub bits are mainly designed for desktops, and IMO haven't seen
-  # enough testing in concert with ostree. At some point we'll flesh out
-  # the full plan in https://github.com/coreos/fedora-coreos-tracker/issues/47
-  - [grub2-tools, /etc/grub.d/08_fallback_counting,
-                  /etc/grub.d/10_reset_boot_success,
-                  /etc/grub.d/12_menu_auto_hide,
-                  /usr/lib/systemd/.*]
 
 
 remove-files:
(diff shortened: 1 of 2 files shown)


ajeddeloh (approved):
LGTM once the dependent PR merges

cgwalters:
Lifting draft state, current cosa has what we need:

https://github.com/coreos/fedora-coreos-tracker/issues/259

```
root@quicksilver ~# podman run --rm -ti quay.io/coreos-assembler/coreos-assembler:latest shell
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
[coreos-assembler]$ rpm -q ostree
ostree-2019.3-3.fc30.x86_64
[coreos-assembler]$ rpm -q rpm-ostree
rpm-ostree-2019.5.7.gcac5aa41-3.fc30.x86_64
[coreos-assembler]$ exit
root@quicksilver ~# 
```

fedora-coreos-base.yaml
```diff
@@ -4,6 +4,9 @@
 # https://github.com/coreos/coreos-assembler/pull/639
 
 include: minimal.yaml
+arch_include:
+  x86_64: grub2-removals.yaml
+  aarch64: grub2-removals.yaml
```
cgwalters, line 9: Hm, this should have `ppc64le` too right?

cgwalters: OK, yes clearly.  Fixed.

cgwalters:
Will merge soon if no further comments are raised.

cgwalters:
bot, retest this please

cgwalters:
bot, retest this please

cgwalters:
I dunno how I had it with `_` here, it's clearly `-`.  Fixed.

fedora-coreos-base.yaml
```diff
@@ -4,6 +4,10 @@
 # https://github.com/coreos/coreos-assembler/pull/639
 
 include: minimal.yaml
+arch-include:
```
jlebon, line 7: Not new, though WDYT about making this part of `minimal.yaml` instead since that's where the bootloader stuff is defined (so it's trivial to cross-check which arches use GRUB)?

jlebon (approved):
One optional comment, otherwise LGTM!

fedora-coreos-base.yaml
cgwalters: I thought about that, but decided it was better to unblock s390x first - let's land this, then I think I agree on rolling it into minimal.

PR #151, opened 2019-08-20
README.md: add instructions on adding new packages

jlebon:
Now that we've completely switched over to lockfiles, we need to
describe the procedure for adding new packages to the OS.

Also add back the Bodhi repos to make the task easier. Otherwise, we'd
get a bootstrapping problem: we can't add packages not tagged in
coreos-pool, but the tagger won't tag a package that's not added.

But also, we really want those repos there by default so that the local
dev case of pulling in all new packages can just be
`cosa fetch --update-lockfile`.

---

The fundamental issue here is that right now we're relying on `bodhi-updates` to sync up our lockfile with our manifests. But this means that adding a package without a corresponding lockfile entry will instantly break `testing-devel` until the next `bodhi-updates` sync. This is what happened right now with `jq`: because `bodhi-updates` is currently failing (due to https://github.com/coreos/fedora-coreos-releng-automation/pull/35), the lockfile in `testing-devel` doesn't yet have `jq`. If we want to remain decoupled from `bodhi-updates` builds, we need to enforce lockfiles at the stage where package get added.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -3,6 +3,13 @@ include: fedora-coreos.yaml
 
 repos:
   - fedora-coreos-pool
+  # these repos are there to make it easier to add new packages to the OS and to
+  # use `cosa fetch --update-lockfile`; but note that all package versions are
+  # still pinned
+  - fedora
+  - fedora-updates
+  - fedora-modular
+  - fedora-updates-modular
 
 add-commit-metadata:
   fedora-coreos.stream: testing-devel
--- a/README.md
+++ b/README.md
@@ -28,3 +28,23 @@ following exceptions:
   imported from `bodhi-updates` to `testing-devel`.
   Overrides (`manifest-lock.overrides.*`) are manually
   curated.
+
+## Adding packages to the OS
+
+Since `testing-devel` is directly promoted to `testing`, it
+must always be in a known state. The way we enforce this is
+by requiring all packages to have a corresponding entry in
+the lockfile.
+
+Therefore, to add new packages to the OS, one must also add
+the corresponding entries in the lockfiles:
+- for packages which should follow Bodhi updates, place them
+  in `manifest-lock.$basearch.json`
+- for packages which should remain pinned, place them
+  in `manifest-lock.overrides.$basearch.json`
+
+There will be better tooling to come to enable this, though
+one easy way to do this is for now:
+- add packages to the correct YAML manifest
+- run `cosa fetch --update-lockfile`
+- commit only the new package entries
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -532,6 +532,10 @@
       "evra": "2.12-2.fc30.x86_64",
       "digest": "sha256:f4270a5ace04775e27156a29424c98ddbf7d3f76ecef4d865fbc0c8634a67b56"
     },
+    "jq": {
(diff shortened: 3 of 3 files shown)


dustymabe:
I almost wonder if we shouldn't just have a staging branch (call it `master`) where we collect changes before we sync them up to testing-devel. So we can make several small changes and then it automerges to testing-devel once a compose/test passes.  So the new development target is now `master` and `testing-devel` just represents the latest commit in master that builds and passes tests. `coreos-koji-tagger` now watches the `master` branch etc..

dustymabe:
What I just described is basically PR testing, but allows us to use a stable branch name to trigger other operations off of.

I don't know if that is really better or not, just spitballing. Either way it seems like we have a few chicken/egg problems that can happen :).

manifest.yaml
```diff
@@ -3,6 +3,13 @@ include: fedora-coreos.yaml
 
 repos:
   - fedora-coreos-pool
+  # these repos are there to make it easier to add new packages to the OS and to
+  # use `cosa fetch --update-lockfile`; but note that all package versions are
+  # still pinned
+  - fedora
+  - fedora-updates
+  - fedora-modular
+  - fedora-updates-modular
```
dustymabe, line 12: I'm not sure if I'd prefer to optimize for the 'adding a new package' case. What we are doing here is trading off "knowing that testing will build when merged" vs "ease of adding a new package". Would it be better to have testing-devel fail if a package isn't in the `fedora-coreos-pool` or optimize for adding a new package?

jlebon:
On Tue, Aug 20, 2019, 19:30 Dusty Mabe, <notifications@github.com> wrote:

> *@dustymabe* commented on this pull request.
> ------------------------------
>
> In manifest.yaml
> <https://github.com/coreos/fedora-coreos-config/pull/151#discussion_r315945862>
> :
>
> > @@ -3,6 +3,13 @@ include: fedora-coreos.yaml
>
>  repos:
>    - fedora-coreos-pool
> +  # these repos are there to make it easier to add new packages to the OS and to
> +  # use `cosa fetch --update-lockfile`; but note that all package versions are
> +  # still pinned
> +  - fedora
> +  - fedora-updates
> +  - fedora-modular
> +  - fedora-updates-modular
>
> I'm not sure if I'd prefer to optimize for the 'adding a new package'
> case. What we are doing here is trading off "knowing that testing will
> build when merged" vs "ease of adding a new package".
>
Not just adding a new package, but also to enable the `cosa fetch` and
`cosa fetch --update-lockfile` workflow. I'm hoping we can avoid devs
having to manually add the repos back to pull newer packages.

But I agree it's not ideal either. I think there is a tension here due to
testing-devel being both the development target and the staging area for
testing. I'm not a fan of adding another branch though.

> Would it be better to have testing-devel fail if a package isn't in the
> fedora-coreos-pool or optimize for adding a new package?
>
Once we implement lockfile updates via PR, then we can ensure at merge time
that all packages are in the pool. This is something we want anyway even
for regular bodhi-updates lockfile promotions so we close various races. So
testing-devel should always be buildable (and passing tests).

In the end, regardless of the repos enabled, the lockfiles are the source
of truth here. We want good tooling around managing them for both
development and production use cases.

dustymabe:
we'll merge this for now and have a bigger discussion in a separate ticket.

PR #153, opened 2019-08-21
overlay: remove 40coreos-network dracut module

yuqi-zhang:
Remove the 40coreos-network, which was intended to bring down
initramfs networking. Currently this module wasn't enabled anyways,
and will become obsolete once we move to NetworkManager based
initramfs networking.

For context, this is currently used in RHCOS to teardown networking
before the switch root, to account for static use cases. This
effectively allowed users to dhcp in the initramfs to fetch ignition
configs, and then have ignition set up static networking on the real
root. Without this teardown it will instead just inherit initramfs
networking.


cgwalters:
So...for FCOS it doesn't quite work right today then to do "DHCP for provisioning, static configs in Ignition" anyways, and this PR won't affect RHCOS as we aren't directly inheriting FCOS yet - does that all sound right?

yuqi-zhang:
Yes. Once the inheritance is in place we can overlay the one for RHCOS in still.

cgwalters:
LGTM

ajeddeloh (approved):
LGTM

PR #154, opened 2019-08-22
manifest: add libvarlink-util

### LINKED ISSUE #231: Ship libvarlink-utils
We don't have `libvarlink-utils` as part of default ostree, any plan to add it. This is required to make podman remotely accessible as pointed out https://podman.io/blogs/2019/01/16/podman-varlink.html post.

dustymabe:
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/231

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -135,6 +135,8 @@ packages:
   # Containers
   - podman skopeo runc moby-engine systemd-container
   - fuse-overlayfs slirp4netns
+  # Remote IPC for podman
+  - libvarlink-util
   # Networking
   - nfs-utils-coreos
   - NetworkManager hostname


bgilbert (approved):
LGTM

jlebon:
FWIW, this should've included a corresponding lockfile :lock: entry as described in https://github.com/coreos/fedora-coreos-config#adding-packages-to-the-os.

This will also be a hard requirement if we want to switch to strict lockfile mode (https://github.com/projectatomic/rpm-ostree/pull/1858).

jlebon:
(To be clear, this still worked fine by virtue of the manifest being propagated to bodhi-updates, then an updated lockfile being promoted back down. The issue is that before that happens, `testing-devel` isn't completely defined, and so we can't promote it to `testing`. `testing` only uses `coreos-pool` now, which won't have the package until `coreos-koji-tagger` picks it up from the lockfile).

PR #155, opened 2019-09-05
Initial support for live PXE and ISO images

### LINKED ISSUE #115: "enable boot-efi.mount" make ppc64le platform fail

bgilbert:
Deferred to follow-up PRs:
- Functioning with SELinux enabled
- Allowing users to provide an Ignition config via a custom initrd image
- Automatic login in the ISO image

Other known issues:
- dracut passes the ostree path to ostree-prepare-root by bind-mounting over `/proc/cmdline`
- rpm-ostreed refuses to start
- ISO kargs are not automatically synced from the BLS configs

Fixes https://github.com/coreos/fedora-coreos-config/issues/115.


overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
```diff
@@ -0,0 +1,4 @@
+install() {
+    inst_script "$moddir/is-live-image.sh" \
+        "/usr/bin/is-live-image"
```
cgwalters, line 3: I know this is just the initramfs but I am not a fan of "/usr/bin pollution".  `/usr/libexec/coreos-is-live-image` ?

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-populate-writable.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+DefaultDependencies=false
+RequiresMountsFor=/writable
```
cgwalters, line 3: Don't we preserve `/run` from the initramfs?  If so, let's put stuff in `/run/coreos-live-writable` or so?

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-populate-writable.service
```diff
@@ -0,0 +1,9 @@
...
+Before=sysroot-etc.mount sysroot-var.mount
+
+[Service]
+Type=oneshot
```
cgwalters, line 7: I need to write a bot that detects pull requests that have systemd units with `Type=oneshot` and without `RemainAfterExit=yes`.... https://github.com/ostreedev/ostree/pull/1697

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -0,0 +1,91 @@
...
+    ln -sf "../${name}" "${requires_dir}/${name}"
+}
+
+if [ -e /root.squashfs ]; then
+    add_requires sysroot.mount
+    add_requires sysroot-etc.mount
+    add_requires sysroot-var.mount
+
+    mkdir -p "${UNIT_DIR}/ostree-prepare-root.service.d"
+    cat > "${UNIT_DIR}/ostree-prepare-root.service.d/10-live.conf" <<EOF
```
cgwalters, line 25: I'm OK with this, though I'd also like to consider standardizing support for this in ostree itself at some point.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -0,0 +1,45 @@
...
+Before=local-fs.target
+
+[Mount]
+What=/dev/disk/by-label/boot
+Where=/boot
+EOF
+
+    # Don't try to mount /boot/efi on arches that don't have it
+    arch="$(uname -m)"
+    if [ "$arch" = "x86_64" -o "$arch" = "aarch64" ]; then
```
cgwalters: `test -d /sys/firmware/efi` ?

live/EFI/fedora/grub.cfg
```diff
@@ -28,6 +28,6 @@ set timeout=1
 
 ### BEGIN /etc/grub.d/10_linux ###
 menuentry 'Fedora CoreOS (Live)' --class fedora --class gnu-linux --class gnu --class os {
-	linux /images/vmlinuz rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
+	linux /images/vmlinuz mitigations=auto,nosmt rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
```
cgwalters, line 31: Hmm.  It feels cleaner if we pulled this stuff out of an existing disk image.  But OK for now.

cgwalters (commented):
Overall looks good!  I'd like to think a bit harder about the "API" around which live mounts are detected.  I feel like e.g. `/run/ostree-booted` has worked well; why not have `/run/ostree-live` or so and use `ConditionPathExists=/run/ostree-live` rather than `ConditionPathIsReadWrite=/`?

It's funny because just last night I was working on https://github.com/ostreedev/ostree/issues/1265
and while that *happens* to leave `/` as technically `rw`...I am not sure I really want to encode that in many systemd units permanently.

BTW...there is also https://www.freedesktop.org/software/systemd/man/systemd-volatile-root.service.html which was clearly intended for something like this, but it doesn't deal with `/etc` for example.

overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
bgilbert: This is an API with ignition-dracut, which is distro-independent, so it shouldn't be `coreos-`.  We could use `/usr/libexec` but nothing else in the initrd does (except coreos-installer), and `/usr/bin` pollution is rampant in the initrd, so I'm inclined to leave it.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
bgilbert: Yeah, that'd be good.  This is meant as a starting point.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
bgilbert: That would omit the mount on BIOS-booting x86 systems.  They don't use the ESP, but they still _have_ it, so it feels more consistent to mount it.

cgwalters: `if [ "$arch" = "x86_64" -o -d /sys/firmware/efi ]` ?

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-populate-writable.service
bgilbert: We do.  I thought it'd be cleaner to have a dedicated tmpfs for user-created state.  Also, this avoids exposing the overlayfs upperdir and workdir within the directory structure.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
bgilbert: Sure.

bgilbert:
> I feel like e.g. `/run/ostree-booted` has worked well; why not have `/run/ostree-live` or so and use `ConditionPathExists=/run/ostree-live` rather than `ConditionPathIsReadWrite=/`?

I'm okay with that.  `ConditionPathIsReadWrite` doesn't feel :100: to me.

bgilbert:
Updated.

lucab:
Just for clarity, which component is going to create `/run/ostree-live` and at what point of the build (or boot) flow?

bgilbert:
@lucab It's currently being done at boot time, in `live-generator` in the initramfs.

cgwalters:
This all LGTM!

bgilbert:
To test:

- Pull in this PR and https://github.com/coreos/coreos-assembler/pull/738.
- Copy the contents of https://github.com/coreos/ignition-dracut/pull/102 into `overrides/rootfs`.
- `coreos-assembler build && coreos-assembler buildextend-live`
- Test:
```
# PXE
qemu-system-x86_64 -machine accel=kvm -m 2048 \
    -netdev user,id=eth0,hostfwd=tcp::2222-:22,hostname="live" -device virtio-net-pci,netdev=eth0 \
    -fw_cfg name=opt/com.coreos/config,file=$HOME/ssh-key.ign \
    -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0 \
    -kernel fedora-coreos-30.*-live-kernel \
    -initrd fedora-coreos-30.*-live-initramfs.img \
    -append 'rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=qemu enforcing=0'`
# ISO
qemu-system-x86_64 -machine accel=kvm -m 2048 -cdrom fedora-coreos-30.*.iso \
    -netdev user,id=eth0,hostfwd=tcp::2222-:22,hostname="live" -device virtio-net-pci,netdev=eth0 \
    -fw_cfg name=opt/com.coreos/config,file=$HOME/ssh-key.ign \
    -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0
```
For ISO boot, manually edit the kargs to specify `ignition.platform.id=qemu enforcing=0`.

cgwalters:
I'm OK merging this as is; do you think someone else needs to test before merging to master?

overlay.d/05core/usr/lib/dracut/modules.d/20live/writable.mount
```diff
@@ -0,0 +1,11 @@
...
+DefaultDependencies=false
+Requires=coreos-populate-writable.service
+Before=coreos-populate-writable.service
+
+[Mount]
+What=tmpfs
+Where=/writable
+Type=tmpfs
+Options=mode=0700
+EOF
```
SerialVelocity: Probably want to remove this line before merging. Pretty sure it is leftover from when you extracted it out from a `cat <<EOF`

bgilbert: Great catch, thanks!  Updated.

dustymabe:
> I'm OK merging this as is; do you think someone else needs to test before merging to master?

i'm running through reviews/tests today

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -0,0 +1,94 @@
...
+DefaultDependencies=false
+Before=initrd-root-fs.target
+
+[Mount]
+What=/root.squashfs
+Where=/sysroot
+Type=squashfs
+EOF
+
+cat >"${UNIT_DIR}/sysroot-etc.mount" <<EOF
```
dustymabe: indent

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -0,0 +1,94 @@
...
+Before=ignition-mount.service
+
+[Mount]
+What=overlay
+Where=/sysroot/etc
+Type=overlay
+Options=lowerdir=/sysroot/etc,upperdir=/writable/etc/upper,workdir=/writable/etc/work,redirect_dir=on,index=on,xino=on
+EOF
+
+cat >"${UNIT_DIR}/sysroot-var.mount" <<EOF
```
dustymabe: indent

overlay.d/05core/usr/lib/dracut/modules.d/20live/ostree-cmdline.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -euo pipefail
+
+case "${1:-unset}" in
+start)
```
dustymabe: indent these cases?

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
dustymabe: It's definitely hacky to bind mount over `/proc/cmdline` but is a creative way to get this working now. Can you open an RFE against ostree to support this natively and copy the URL to the issue in this dropin unit as well as a comment in the ostree-cmdline.sh script?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -0,0 +1,44 @@
...
+Description=Boot partition
+Documentation=https://github.com/coreos/fedora-coreos-config
+Before=local-fs.target
+
+[Mount]
+What=/dev/disk/by-label/boot
+Where=/boot
+EOF
+
+    # Mount /boot/efi only on UEFI platforms, as well as x86_64 BIOS
```
dustymabe: can we add a comment here on **why** we want this behavior on these specific cases

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -0,0 +1,44 @@
...
+
+add_wants() {
+    local name="$1"; shift
+    local wants_dir="${UNIT_DIR}/local-fs.target.wants"
+    mkdir -p "${wants_dir}"
+    ln -sf "../${name}" "${wants_dir}/${name}"
+}
+
+# Don't create mount units for /boot or /boot/efi on live systems
+if [ ! -f /run/ostree-live ]; then
```
dustymabe, line 17: dumb question. What guarantees the `usr/lib/dracut/modules.d/20live/live-generator` runs before this generator so that `/run/ostree-live` is guranteed to exist on a live system?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -0,0 +1,44 @@
...
+UNIT_DIR="${1:-/tmp}"
+
+add_wants() {
+    local name="$1"; shift
+    local wants_dir="${UNIT_DIR}/local-fs.target.wants"
+    mkdir -p "${wants_dir}"
+    ln -sf "../${name}" "${wants_dir}/${name}"
+}
+
+# Don't create mount units for /boot or /boot/efi on live systems
```
dustymabe: it would be nice if we add a comment here about not being able to use `ConditionPathExists /run/ostree-live`. You have this comment in your commit message, but would be nice to have it here.

live/zipl.prm
```diff
@@ -0,0 +1 @@
+mitigations=auto,nosmt rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
```
dustymabe, line 1: regarding the `mitigations=auto,nosmt` (the last commit in this series) it seems like this is going to be a bit of a maintenance burden because we store that information currently in `image.yaml` can we pull those args from there?

dustymabe (commented):
some initial comments. I'll probably have some more tomorrow when I finally test the artifacts. Good news is they did build locally for me!

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
bgilbert: This generator runs in the real root and `live-generator` runs in the initramfs.

live/zipl.prm
bgilbert: You mean just the `extra-kargs`?  We can, but there are also args coming from other places (`ignition.platform.id`, `console`, the temporary networking args, `rootflags`, etc).  We could read them out of the BLS config and filter out the undesired ones, but I thought that might be more trouble than it's worth.  Thoughts?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
dustymabe: ahh missed that point 👍

live/zipl.prm
dustymabe: > We could read them out of the BLS config and filter out the undesired ones, but I thought that might be more trouble than it's worth. Thoughts?

I like that, depending on how much trouble it really is. Basically we'd maintain a blacklist in a single place vs duplicated kernel args in various places.

bgilbert: OK.  Can I queue that for a followup PR?

dustymabe:
it looks like the selinux issues are related to files getting labeled with the `tmpfs_t` type probably because they are getting fiddled with under the `/writable` tmpfs at some point?

bgilbert:
Updated.

dustymabe:
OK. I ran through and tested the live PXE, live ISO, and installer ISO with these changes. We have some outstanding comments but they are mostly small fixup items and some of them will lead to followup work (rather than work to do in this PR). We should be able to address the remaining small comments on Monday and open a few issues for followup work (maybe others can pick up some of the tasks) and then get this in!

Nice work @bgilbert!

bgilbert:
@dustymabe Thanks for the reviews and testing!

You're right about SELinux.  I have a followup commit that fixes that via `relabel-extra.d`, but it depends on both [selinux-policy-3.14.3-46.fc30](https://koji.fedoraproject.org/koji/buildinfo?buildID=1377250) and https://github.com/systemd/systemd/pull/13524.

As far as I know, all of the feedback on this PR set has now been addressed, other than missing functionality that can be handled in a followup.  @dustymabe, is there something I've missed?  I'd like to get this set of PRs merged to clear the way for the next set.

dustymabe (approved):
LGTM - I've enumerated some followup items below. Should we open up a few issues to track anything that doesn't already have one? 

- https://github.com/coreos/fedora-coreos-config/pull/155#discussion_r324017202
    - pull kernel args for live images from common location as other disk images, possibly blacklist ones we know we don't want for lives

- https://github.com/coreos/fedora-coreos-config/pull/155#issuecomment-531588558
    - enable selinux for live images 


- rpm-ostreed refuses to start
    - https://github.com/ostreedev/ostree/issues/1921


- https://github.com/coreos/fedora-coreos-config/pull/155#discussion_r324011650
    - https://github.com/ostreedev/ostree/issues/1920
    - dracut passes the ostree path to ostree-prepare-root by bind-mounting over /proc/cmdline rpm-ostreed refuses to start

dustymabe:
One other thing that I thought of this weekend: It would probably be really nice to have a high level design doc for how this all works since it can be hard to stitch all of the pieces together from the various repos. It's optional of course, but a design doc in the fcos tracker repo (or somewhere) might be helpful to us all in the future.

bgilbert:
> pull kernel args for live images from common location as other disk images, possibly blacklist ones we know we don't want for lives

Should be done this week.

> enable selinux for live images

Coming in followup PR.

PR #157, opened 2019-09-06
overlay.d: Add 08nouveau to blacklist the nouveau driver

cgwalters:
Pulled from RHCOS, following the upstreaming principle.  The
nouveau driver apparently has bugs with some NVidia hardware in EC2.
Since (unlike a Fedora desktop system) having a GPU enabled isn't
critical path, blacklist the driver.  Anyone who wants to enable
nouveau can do so of course, or they can install the proprietary
driver.


bgilbert (approved):
Confirmed seeing this in 30.20190905.0.

jlebon (approved):

PR #159, opened 2019-09-06
overlay: add Azure udev rules

miabbott:
The downstream RHBZ#1747575[0] requests the addition of these udev
rules to allow nodes in Azure to handle the dynamic provisioning of
storage devices.  The WALinuxAgent repo[1] provides the two udev rules
required for this functionality, which haven't changed in a year or more.
This commit is just a copy of the rules from the repo.

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1747575
[1] https://github.com/Azure/WALinuxAgent/tree/master/config/

PR #160, opened 2019-09-07
overlay: add GCP udev rules

miabbott:
The downstream BZ#1748638[0] requests the addition of udev rules for
nodes in GCP to support the dynamic provisioning of storage.  The
rules live in the `GoogleCloudPlatform/compute-image-packages` repo[1]
and were most recently updated in Dec 2018.  This commit is a copy of
the rules from the repo.

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1748638
[1] https://github.com/GoogleCloudPlatform/compute-image-packages/tree/master/packages/google-compute-engine/src/lib/udev/rules.d

### DIFF
--- a/overlay.d/05core/usr/lib/udev/rules.d/64-gce-disk-removal.rules
+++ b/overlay.d/05core/usr/lib/udev/rules.d/64-gce-disk-removal.rules
@@ -1,0 +1,17 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+#     distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# When a disk is removed, unmount any remaining attached volumes.
+
+ACTION=="remove", SUBSYSTEM=="block", KERNEL=="sd*|vd*", ENV{ID_VENDOR}=="Google", RUN+="/bin/sh -c '/bin/umount -fl /dev/$name && /usr/bin/logger -p daemon.warn -s WARNING: hot-removed /dev/$name that was still mounted, data may have been corrupted'"
--- a/overlay.d/05core/usr/lib/udev/rules.d/65-gce-disk-naming.rules
+++ b/overlay.d/05core/usr/lib/udev/rules.d/65-gce-disk-naming.rules
@@ -1,0 +1,38 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+#     distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Name the attached disks as the specified by deviceName.
+
+ACTION!="add|change", GOTO="gce_disk_naming_end"
+SUBSYSTEM!="block", GOTO="gce_disk_naming_end"
+
+# SCSI naming
+KERNEL=="sd*|vd*", ENV{ID_VENDOR}=="Google", IMPORT{program}="scsi_id --export --whitelisted -d $tempnode"
+
+# NVME naming
+KERNEL=="nvme0n1*", ENV{ID_VENDOR}=="Google", ENV{ID_SERIAL_SHORT}="local-nvme-ssd-0"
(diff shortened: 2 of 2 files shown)


miabbott:
Worth noting is that the BZ references one other udev rule (`99-gce.rules`) that does not appear in the repo referenced in the commit.  I'm not sure where that rule comes from and/or how we could include it.

bgilbert:
These rules aren't conditionalized for GCP, so they'd run on every platform.  We should limit them to GCP.  [Here's](https://github.com/coreos/init/blob/master/udev/rules.d/90-cloud-storage.rules) how Container Linux does it.

lucab:
Additionally, this would hijack filepaths that belong to other packages (In Fedora, that would be [google-compute-engine-tools](https://src.fedoraproject.org/rpms/google-compute-engine-tools)).

Have we considered splitting the binary package out instead of duplicating them? That would also help with maintenance/updates/ownership.

miabbott:
> These rules aren't conditionalized for GCP, so they'd run on every platform. We should limit them to GCP.

Doh!  I didn't bother to check; just assumed that they would behave like the Azure rules in #159.  I'll rework this.

> Have we considered splitting the binary package out instead of duplicating them? That would also help with maintenance/updates/ownership.

Similar to what you've proposed for the [Azure rules](https://bugzilla.redhat.com/show_bug.cgi?id=1748432)?

cgwalters:
> Additionally, this would hijack filepaths that belong to other packages (In Fedora, that would be google-compute-engine-tools).

That doesn't seem like a real problem to me; we're not supporting installing the agent on FCOS right?  That said, it *would* be cleaner to have them come from a `google-compute-engine-agent-core` or something I guess.

lucab:
Oh, definitely not something to stall this on, but more like whether a plan was in place for longer term (similar to Azure, yes).

miabbott:
I'd like to keep the existing rules as intact as possible, so would it be reasonable to scope them with `ENV{ID_VENDOR}=="Google"`?  I haven't been able to test all of them with this change, but a simple add/remove of a disk seemed to trigger the correct rules.

Related:  if the rules are changed as such, how does the copyright stanza have to change, if at all?

cgwalters:
> Oh, definitely not something to stall this on, but more like whether a plan was in place for longer term (similar to Azure, yes).

Yep, understood.  We're basically making the rules up as we go along, so to speak.

bgilbert:
@miabbott I'm not super-comfortable with the way the rules assign to `ENV{ID_SERIAL}` and `ENV{ID_SERIAL_SHORT}`, but yeah, I think that'd be okay.

We could add a 2019 Red Hat copyright statement, but AIUI there's not a strong need to do so.

cgwalters:
> These rules aren't conditionalized for GCP, so they'd run on every platform. We should limit them to GCP. Here's how Container Linux does it.

It'd be safe to propose a patch upstream for this too, right?

miabbott:
Pushed a commit with additional scoping :arrow_up: 

Thanks for the feedback so far!

miabbott:
Rebased to see if CI will be happy :crossed_fingers:

cgwalters:
One thing I wanted to quickly verify is the vendor, and yep, LGTM:

```
walters@toolbox ~/s/g/o/installer> oc debug node/ci-ln--v4m6v-w-b-rmbmt.c.openshift-gce-devel-ci.internal
...
sh-4.2# chroot /host
...
sh-4.4# cat /sys/block/sda/device/vendor 
Google  
sh-4.4# 
```

overlay.d/05core/usr/lib/udev/rules.d/64-gce-disk-removal.rules
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+#     distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# When a disk is removed, unmount any remaining attached volumes.
+
+ACTION=="remove", SUBSYSTEM=="block", KERNEL=="sd*|vd*", ENV{ID_VENDOR}=="Google", RUN+="/bin/sh -c '/bin/umount -fl /dev/$name && /usr/bin/logger -p daemon.warn -s WARNING: hot-removed /dev/$name that was still mounted, data may have been corrupted'"
```
bgilbert, line 17: I'm not sure about this rule.  It means that we're behaving differently on GCP than on other platforms when devices are yanked from the system.

cgwalters: Yeah, in the end I don't think it's *really* specific to GCE, on at least AWS too one can force detach with [the API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DetachVolume.html).

And AIUI this occurs when e.g. higher level code like Kubernetes Persistent Volume management cleans up resources before it unmounts it on the OS.

bgilbert: Could we just drop this file?

cgwalters: The [Bugzilla that spawned this](https://bugzilla.redhat.com/show_bug.cgi?id=1748638) doesn't say *exactly* what went wrong...but presumably it was related to the naming, and not this hack to do the unmount.  So we probably could drop it...

miabbott: I pinged the reporter in the BZ about the removal rules.  We can carry them in RHCOS if needed, but I agree we could probably remove them for FCOS.

bgilbert: https://github.com/coreos/fedora-coreos-config/pull/162

PR #162, opened 2019-09-12
05core: drop 64-gce-disk-removal.rules

bgilbert:
When a disk is yanked out of the machine, this rule provides different behavior on GCP than on other platforms.  It doesn't meaningfully affect functionality, so let's remove the rule for consistency.

See discussion in the [original PR](https://github.com/coreos/fedora-coreos-config/pull/160/files#r323279618).


cgwalters (approved):
The reporter didn't really answer the question, I haven't gotten around to running their test suite, but...if it relies on this behavior somehow it would have to be broken.

So, LGTM.

miabbott (approved):
We got a reply from the reporter on the downstream BZ; he confirmed their product [does not need the removal rule](https://bugzilla.redhat.com/show_bug.cgi?id=1748638#c12).

Additionally, I was able to [run the suggested test suite](https://bugzilla.redhat.com/show_bug.cgi?id=1748638#c13) on an RHCOS GCP image that has the rule included and it passed, so the rule seems inconsequential. 

All to to say, LGTM.  :)

I'll also mention I filed a BZ with the `google-compute-engine-tools` package in Fedora to get the [udev rules in a separate package](https://bugzilla.redhat.com/show_bug.cgi?id=1751310).

PR #163, opened 2019-09-12
ci: use Jenkins checkout instead of recloning

jlebon:
Just do the checkout in a subdirectory in the the workspace, then point
`cosa` at it. That way we don't clone the repo twice.

This should fix CI errors when PRs are not fully rebased.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/747

miabbott (approved):
Makes sense!  :+1:

PR #164, opened 2019-09-12
manifest: override rojig.summary for testing-devel

bgilbert:
We're currently releasing AMIs with a description field of the form `Fedora CoreOS base image <version>`.  Let's include the stream name in the rojig summary for each stream.

Copy over the license and name fields because that appears to be required.

I'll PR the other branches once this lands.


cgwalters (approved):
LGTM

PR #165, opened 2019-09-16
 Introduce ignition-and-ostree.yaml

cgwalters:
Currently RHCOS includes `minimal.yaml`.  I would like
to increase the amount of sharing.  There's some stuff in
`fedora-coreos-base.yaml` which we don't want (such as pinger, etc.)

Introduce a new `ignition-and-ostree.yaml` which adds the core
Ignition stuff plus some default tweaks.


jlebon (approved):
LGTM.

I have a slight worry that all this splitting is going to make things harder to maintain. Probably deserves a blurb in `README.md` at this point describing the inheritance. No need to block on this though.

cgwalters:
> I have a slight worry that all this splitting is going to make things harder to maintain

It does a bit - but it increases sharing with *derivatives* of FCOS which is the idea!

> Probably deserves a blurb in README.md at this point describing the inheritance.

I tried to describe things at the tops of the files - and the `include:` stanzas are right there in the code.

jlebon:
> It does a bit - but it increases sharing with derivatives of FCOS which is the idea!

Right, the motivation is clear, just calling out that there is a cost! :)

> I tried to describe things at the tops of the files - and the `include:` stanzas are right there in the code.

Yeah, I think that helps. Though I think having them all documented in inheritance order in the same file would be easier to digest. Not asking for this here, we/I can do that as a follow-up.

Will leave open for a bit.

dustymabe:
> It does a bit - but it increases sharing with _derivatives_ of FCOS which is the idea!

I honestly think we're going to need some re-working of the architecture if we want derivatives of FCOS. You've mentioned several times Silverblue inheriting from FCOS, but I just don't know if it makes sense. We should set aside some time to talk about it.

cgwalters:
> I honestly think we're going to need some re-working of the architecture if we want derivatives of FCOS. 

Well, we're shipping one today in RHCOS...or at least, that's how I like to think of it...it's "truthy" anyways!  This PR is about making that increasingly true.

> You've mentioned several times Silverblue inheriting from FCOS, but I just don't know if it makes sense. We should set aside some time to talk about it.

As a tracker issue?  It's not just Silverblue of course but also e.g. Fedora IoT.  

(BTW in this discussion it's actually more like a "Fedora CoreOS CoreOS"...we don't want `moby-engine` in RHCOS for example)

dustymabe:
> As a tracker issue?

Yep. A tracker issue that would lead into a meeting discussion (realtime with various interested parties and high visibility) would be best I think.

PR #166, opened 2019-09-16
manifest: Explicitly pull in kernel-core, not kernel-debug-core

cgwalters:
We were just adding `kernel` which is a metapackage.  Recently
it looks like the kernel package grew a separate `kernel-debug-core`
(I'm guessing to make it easy to switch between the two)

We need to explicitly pull in `kernel-core` and `kernel-modules`
to ensure the depsolver doesn't pull in the debug kernel.

### DIFF
--- a/minimal.yaml
+++ b/minimal.yaml
@@ -5,8 +5,9 @@
 # We expect most people though using coreos-assembler to inherit from
 # fedora-coreos-base.yaml.
 packages:
- # Kernel + systemd
- - kernel systemd
+ # Kernel + systemd.  Note we explicitly specify kernel-{core,modules}
+ # because otherwise depsolving could bring in kernel-debug.
+ - kernel kernel-core kernel-modules systemd
  # rpm-ostree
  - rpm-ostree nss-altfiles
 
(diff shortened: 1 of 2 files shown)


cgwalters:
(This bumps the lockfiles to ensure the tagger re-pulls `kernel-core` into the pool)

minimal.yaml
```diff
@@ -5,8 +5,9 @@
 # We expect most people though using coreos-assembler to inherit from
 # fedora-coreos-base.yaml.
 packages:
- # Kernel + systemd
- - kernel systemd
+ # Kernel + systemd.  Note we explicitly specify kernel-{core,modules}
+ # because otherwise depsolving could bring in kernel-debug.
+ - kernel-core kernel-modules systemd
```
jlebon: Can we do `kernel kernel-core kernel-modules` so that we still install the `kernel` metapackage? (So then, e.g. `rpm -q kernel` still works at least).

jlebon (commented):
Good catch!

minimal.yaml
cgwalters: Done!

jlebon (approved):

jlebon:
Note to self: this is a really good example of why strict lockfiles would be more appropriate. Rather than interpret Provides from the manifest, we directly use the package list from the lockfiles for no misinterpretations.

dustymabe:
> (This bumps the lockfiles to ensure the tagger re-pulls `kernel-core` into the pool)

I think they all come from the same "build" so this was probably unnecessary

PR #167, opened 2019-09-16
manifest: Use boot-location: modules

cgwalters:
This is the new default; we end up with the kernel data in
only once place.  See the treefile docs as well as
https://github.com/ostreedev/ostree/pull/1079
and discussion in
https://github.com/ostreedev/ostree/pull/1873

### DIFF
--- a/ignition-and-ostree.yaml
+++ b/ignition-and-ostree.yaml
@@ -9,7 +9,7 @@
 include: bootable-rpm-ostree.yaml
 
 # Modern defaults we want
-boot-location: new
+boot-location: modules
 tmp-is-dir: true
 
 # Required by Ignition, and makes the system not compatible with Anaconda


dustymabe:
since I need to look it up docs are here: https://github.com/coreos/rpm-ostree/blob/7db30fec1324bb92b98e491f035f98cdf4094e6f/docs/manual/treefile.md

*"modules": Kernel data goes just in /usr/lib/modules. Use this for new systems, and systems that don't need to be upgraded from very old libostree versions.*

I'm wondering why we haven't switched before now?

dustymabe (approved):

jlebon (approved):

jlebon:
> I'm wondering why we haven't switched before now?

We probably just inherited it from FAH.

dustymabe:
> We probably just inherited it from FAH.

I was thinking it would have been a technical reason and was interested in the reason :)

cgwalters:
> We probably just inherited it from FAH.

Yeah, I'm not aware of another reason.

PR #171, opened 2019-09-16
More live PXE/ISO work

bgilbert:
In this PR:
- Allow running with SELinux enabled (requires https://github.com/systemd/systemd/pull/13524 and https://koji.fedoraproject.org/koji/buildinfo?buildID=1377250)
- Autologin on console when ISO booted and no Ignition config is supplied
- Fail the boot early if the kernel and initrd don't match
- Avoid copying a development README into ISO images
- Unrelated unit fixes

Deferred to follow-up PRs:

- PXE with persistent `/var`
- ISO kargs are not automatically synced from the BLS configs

Other known issues:

- rpm-ostreed refuses to start
- Unit failures in `ignition-diskful.target` don't fail the boot


overlay.d/05core/usr/lib/dracut/modules.d/50coreos-kernel/coreos-check-kernel.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
jlebon, line 1: Heh, this is a clever service unit.

overlay.d/05core/usr/lib/dracut/modules.d/50coreos-kernel/coreos-check-kernel.service
```diff
@@ -0,0 +1,10 @@
...
+Description=Check that initrd matches kernel
+DefaultDependencies=false
+Before=sysinit.target systemd-modules-load.service
+ConditionPathIsDirectory=!/lib/modules/%v
```
jlebon: Minor: thoughts on `/usr/lib` here for consistency and greppability?

live/EFI/fedora/grub.cfg
```diff
@@ -28,6 +28,6 @@ set timeout=1
 
 ### BEGIN /etc/grub.d/10_linux ###
 menuentry 'Fedora CoreOS (Live)' --class fedora --class gnu-linux --class gnu --class os {
-	linux /images/vmlinuz mitigations=auto,nosmt rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
+	linux /images/vmlinuz iso mitigations=auto,nosmt rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
```
jlebon: How generic is this karg? Do we want something more specific here in case other things we don't want key off of it down the line?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
```diff
@@ -0,0 +1,39 @@
...
+    local i
+    for i in "${cmdline[@]}"; do
+        if [ "$arg" = "$i" ]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
+# Only allow autoautologin on live systems
```
jlebon: autologin

jlebon (approved):
Just some minor comments, LGTM overall.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
bgilbert: I meant what I said.  :grin:   Users can still provision autologin by writing a drop-in from their Ignition config.  We just won't do it for them.

live/EFI/fedora/grub.cfg
bgilbert: The name intentionally describes the environment rather than the functionality.  I'm trying to avoid creating something like an `autologin` karg that people might try to use in other contexts.

dustymabe: I'd still like to make this arg something that is specific to us. `iso` is very generic.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
dustymabe: it does **look** like a typo: maybe put a space or hyphen `auto autologin` or `auto-autologin`

dustymabe:
Seems like we could scrap the coreos-autologin-generator and do all the conditions with systemd unit conditions. It's just queuing off of files existing and/or kernel command line params.

bgilbert:
@dustymabe The conditions could be implemented in a unit, but the output is a systemd drop-in, which must be written from a generator.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-populate-writable.service
```diff
@@ -7,5 +7,6 @@ Before=sysroot-etc.mount sysroot-var.mount
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/usr/bin/mkdir -m 755 -p /writable/etc/upper /writable/var
+ExecStart=/usr/bin/mkdir -m 755 -p /writable/etc/upper /writable/var /run/systemd/relabel-extra.d
 ExecStart=/usr/bin/mkdir -m 700 -p /writable/etc/work
+ExecStart=/bin/bash -c 'echo -e "/etc\n/var" > /run/systemd/relabel-extra.d/coreos-writable.relabel'
```
cgwalters, line 12: It's surprising to me this is needed.

cgwalters (commented):
Overall LGTM.  Maybe consider in the future splitting of separate PRs for the obvious unrelated fixes, but not a big deal.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-populate-writable.service
bgilbert: How so?  `/var`, and the root of `/etc`, default to `tmpfs_t`.

bgilbert:
Updated.

overlay.d/05core/usr/lib/dracut/modules.d/50coreos-kernel/coreos-check-kernel.service
dustymabe: @jlebon @cgwalters - is there any reason that the assertion in this unit would ever fail but the system also be in a valid state? if someone overrides the kernel does client side initramfs generation automatically kick in?

jlebon: > is there any reason that the assertion in this unit would ever fail but the system also be in a valid state?

Not AFAICT.

> if someone overrides the kernel does client side initramfs generation automatically kick in?

Yup!

jlebon (approved):
LGTM!

dustymabe:
LGTM - testing now

dustymabe:
I pulled in the new version of the selinux policy and did a [scratch build for systemd](https://koji.fedoraproject.org/koji/taskinfo?taskID=37721110).
I see one final selinux denial that occurs a lot in the logs:

```
Sep 18 19:02:18 localhost audit[904]: AVC avc:  denied  { write } for  pid=904 comm="systemd-journal" name="var" dev="tmpfs" ino=14308 scontext=system_u:system_r:syslogd_t:s0 tcontext=system_u:object_r:var_t:s0 tclass=dir permissive=0
```

bgilbert:
AIUI there's a window between policy load and the directory labels being fixed, during which journald can't write to the journal.

dustymabe:
> AIUI there's a window between policy load and the directory labels being fixed, during which journald can't write to the journal.

it spews quite a bit.. any ideas on how to fix it just for the journal case?

```
[root@localhost ~]# journalctl | grep avc | grep denied | wc -l
34993
```

jlebon:
> AIUI there's a window between policy load and the directory labels being fixed, during which journald can't write to the journal.

Note we already delay some services in order to not hit this race: https://github.com/coreos/fedora-coreos-config/blob/cad5acaf1f5b4490e3678650fef4be54cb8f2ec5/overlay.d/05core/usr/lib/dracut/modules.d/40coreos-var/coreos-populate-var.sh#L38

> ```
> Sep 18 19:02:18 localhost audit[904]: AVC avc:  denied  { write } for  pid=904 comm="systemd-journal" name="var" dev="tmpfs" ino=14308 scontext=system_u:system_r:syslogd_t:s0 tcontext=system_u:object_r:var_t:s0 tclass=dir permissive=0
> ```

Hmm, that doesn't look like an unlabeled issue though. It looks like `syslogd_t` can't write to `var_t`?  Actually, this might also be related to using tmpfs.

dustymabe:
>     * Autologin on console when ISO booted and no Ignition config is supplied

this works on tty0. I'd really like to make ttyS0 a first class citizen too. Can we make it work on both?

ajeddeloh (approved):
Haven't tested but LGTM

bgilbert:
> this works on tty0. I'd really like to make ttyS0 a first class citizen too. Can we make it work on both?

ttyS0 is intentionally not enabled in the live ISO kernel parameters.  Do we want it when booted from CD?

dustymabe:
> ttyS0 is intentionally not enabled in the live ISO kernel parameters. Do we want it when booted from CD?

I always explicitly enable it (add the kernel arg) because serial consoles are so much better IMO. So I guess weirdos like me would want it.

dustymabe:
>* Allow running with SELinux enabled (requires [systemd/systemd#13524](https://github.com/systemd/systemd/pull/13524) and https://koji.fedoraproject.org/koji/buildinfo?buildID=1377250)

Trying to move this along and get it built sooner:
- https://src.fedoraproject.org/rpms/systemd/pull-request/17
- https://src.fedoraproject.org/rpms/systemd/pull-request/18

bgilbert:
Thanks for the push on the SELinux denials.  Turns out we weren't creating `/var/log` from `coreos-populate-var.sh`, so journald was doing it itself.  The directory was thus inheriting a `var_t` label until `systemd-tmpfiles` came along and fixed it afterward.

bgilbert:
Updated to fix SELinux AVCs and to enable autologin on serial consoles.  I didn't enable a serial console by default on ISO boots, though.

dustymabe:
> Updated to fix SELinux AVCs and to enable autologin on serial consoles. 

Woot! It works!

```
[root@localhost ~]# journalctl | grep avc | grep denied | wc -l
0
```

> I didn't enable a serial console by default on ISO boots, though.

That's reasonable. I didn't expect that :)

dustymabe (approved):
LGTM

cgwalters:
>  journalctl | grep avc | grep denied 

A lot more efficient to do `journalctl --grep=avc.*denied` BTW

PR #176, opened 2019-09-20
ci: buildprep from latest testing-devel

jlebon:
That way if a PR adds a package, we'll print the pkg diff.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -10,7 +10,14 @@ cosaPod {
         cosa init ${env.WORKSPACE}
         curl -LO https://raw.githubusercontent.com/coreos/fedora-coreos-releng-automation/master/scripts/download-overrides.py
         python3 download-overrides.py
+        # prep from the latest builds so that we generate a diff on PRs that add packages
+        cosa buildprep https://builds.coreos.fedoraproject.org/prod/streams/testing-devel/builds
     """)
 
     fcosBuild(skipInit: true)
+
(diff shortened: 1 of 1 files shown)


jlebon:
> ostree diff commit from: eb9c3da15432707382294fca7e1a2c80aa81768ec5ce14aedcc181491a00b897
> ostree diff commit to:   c12b04a3eff9495d52f65a40cdc8581028988e1170fce08160f41080061fb3fc

Understandably, there's no diff in this PR, but it's comparing to the previous commit, which is the important part.

cgwalters (approved):

ajeddeloh:
This LGTM but are there plans to make this more visible from GH? It's somewhat buried atm, which I suppose is fine since it only applies to a certain set of PRs. OTOH more visibility is better.

jlebon:
> This LGTM but are there plans to make this more visible from GH?

There aren't plans *not to* ;)

jlebon:
I made it print the pkgdiff as a separate stage at least so it's easier to access. :arrow_up:

ajeddeloh (approved):
:+1: to separate stage for visibility.

ajeddeloh:
CI is failing for unrelated reasons. Is there anything blocking this?

jlebon:
No I think the CI failure is related, I'm just not sure how yet.

jlebon:
OK, let's try this again on top of the new CI!

PR #178, opened 2019-09-23
manifest: Move moby-engine and zincati to fedora-coreos.yaml

cgwalters:
These two aren't used by RHCOS; move them to the toplevel.
Currently RHCOS inherits from `ignition-and-ostree.yaml`; this
is part of an effort to have RHCOS use something closer to
`fedora-coreos-base.yaml`.

### DIFF
--- a/fedora-coreos.yaml
+++ b/fedora-coreos.yaml
@@ -11,8 +11,19 @@ mutate-os-release: "${releasever}"
 packages:
   - fedora-release-coreos
   - fedora-repos-ostree
+  # CL ships this.
+  - moby-engine
   # User metrics
   - fedora-coreos-pinger
+  # Updates
+  - zincati
+
+etc-group-members:
+  # Add the docker group to /etc/group
+  # https://github.com/coreos/fedora-coreos-tracker/issues/2
+  # This will be no longer needed when systemd-sysusers has been implemented:
+  # https://github.com/projectatomic/rpm-ostree/issues/49
+  - docker
 
 # XXX: this is used by coreos-assembler for artifact naming...
 rojig:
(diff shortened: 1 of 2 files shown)


jlebon:
LGTM though requires https://github.com/coreos/rpm-ostree/pull/1908.

jlebon:
Latest rpm-ostree tagged in continuous tag. COSA is rebuilding, then we should see CI be happy.

fedora-coreos.yaml
```diff
@@ -11,8 +11,19 @@ mutate-os-release: "${releasever}"
 packages:
   - fedora-release-coreos
   - fedora-repos-ostree
+  # CL ships this.
```
dustymabe, line 14: We can probably drop this comment?

dustymabe (approved):
LGTM, but let's merge once @jlebon has done the FCOS release (i.e. syncing of changes from testing-devel to testing branch).

fedora-coreos.yaml
cgwalters: Why?

cgwalters:
BTW in practice some other nontrivial work is going to be necessary to have RHCOS use `fedora-coreos-base.yaml`; for example, in Fedora 30+ there's `nfs-utils-coreos`, but not in RHEL 8 yet.

fedora-coreos.yaml
dustymabe: Is that the only reason we're shipping it? I know it is a big one.

cgwalters: If there's a tracker issue for this I can't find it.  I think it was just part of the original design scoped out in person or something?

jlebon (approved):

PR #179, opened 2019-09-24
manifest:  enable SELinux booleans

miabbott:
These two commits enable the following SELinux booleans that are needed for OpenShift operation:

  - `container_use_cephfs`
  - `virt_use_samba`

The first boolean in the list is a backport of something that is currently enabled in RHCOS.  It allows for Ceph volumes to be mounted into containers.

The second boolean in the list is a new request via [RHBZ#1754825](https://bugzilla.redhat.com/show_bug.cgi?id=1754825) that would enable privileged containers to mount in Azure files.

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -89,10 +89,12 @@ postprocess:
     echo -e '\n# Read authorized_keys fragments written by Ignition and Afterburn\nAuthorizedKeysFile .ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn' >> /etc/ssh/sshd_config
 
   # Enable SELinux booleans used by OpenShift
+  # https://github.com/coreos/fedora-coreos-tracker/issues/284
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
     setsebool -P -N container_use_cephfs on  # RHBZ#1692369
+    setsebool -P -N virt_use_samba on  # RHBZ#1754825
 
 packages:
   # Security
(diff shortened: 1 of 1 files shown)


ashcrow (approved):

fedora-coreos-base.yaml
```diff
@@ -93,6 +93,7 @@ postprocess:
     #!/usr/bin/env bash
     set -xeuo pipefail
     setsebool -P -N container_use_cephfs on  # RHBZ#1692369
+    setsebool -P -N virt_use_samba on  # RHBZ#1754825
```
jlebon, line 97: Do we also want `virt_use_nfs` here?

jlebon (approved):
LGTM, just one comment!

(Though it seems like the origin package would be a better place for these.)

fedora-coreos-base.yaml
miabbott: The default SELinux policy had it already enabled (and `virt_sandbox_use_all_caps` too):

```
[core@coreos ~]$ rpm-ostree status
State: idle
AutomaticUpdates: disabled
Deployments:
* ostree://fedora:fedora/x86_64/coreos/testing-devel
                   Version: 30.20190924.dev.2 (2019-09-24T14:07:33Z)
                    Commit: cd46b5ca68e7c897758a29e7e642d8f1dba19539f1edae1d034ddf1d06b22c11
              GPGSignature: (unsigned)
[core@coreos ~]$ getsebool virt_use_nfs
virt_use_nfs --> on
[core@coreos ~]$ getsebool virt_sandbox_use_all_caps
virt_sandbox_use_all_caps --> on
```

jlebon: Ahh gotcha, thanks for checking. Will leave open for a bit in case anyone has any concerns.

miabbott:
Pushed a small change to include a link to a tracker issue (coreos/fedora-coreos-tracker#284)

PR #180, opened 2019-09-24
Prep patches for setting up branched/next-devel/next

jlebon:
Move the `releasever` key to `manifest.yaml`, and add branched development yum repo definitions.

### DIFF
--- a/fedora-coreos.yaml
+++ b/fedora-coreos.yaml
@@ -4,7 +4,6 @@
 
 include: fedora-coreos-base.yaml
 
-releasever: "30"
 automatic-version-prefix: "${releasever}.<date:%Y%m%d>.dev"
 mutate-os-release: "${releasever}"
 
--- a/fedora-next.repo
+++ b/fedora-next.repo
@@ -1,0 +1,81 @@
+# Note we use baseurl= here because using auto-selected mirrors conflicts with
+# change detection: https://github.com/coreos/fedora-coreos-pipeline/issues/85.
+
+[fedora-next]
+name=Fedora $releasever - $basearch
+failovermethod=priority
+baseurl=https://dl.fedoraproject.org/pub/fedora/linux/development/$releasever/Everything/$basearch/os/
+        https://dl.fedoraproject.org/pub/fedora-secondary/development/$releasever/Everything/$basearch/os/
+#metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
+enabled=1
+#metadata_expire=7d
+repo_gpgcheck=0
+type=rpm
+gpgcheck=1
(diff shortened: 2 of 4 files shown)


jlebon:
So basically, there's two ways to approach the "config" problem here: we could keep `testing-devel` as the "source branch" from which we propagate files to the next streams, or we could have `next-devel` be a separate source branch. The latter provides more flexibility, at the expense of more overhead. E.g. changing something that should apply to *both* `testing` and `next` will have to be opened twice against each respective source branch.

I'm leaning more towards the first approach for now (with `testing-devel` being the only source branch), given that IMO we will want the majority of changes to take effect on both `testing` and `next`. And as usual, anything specific to either can be added to their respective `manifest.yaml`.

jlebon:
One hiccup where I see difficulties is if we want to carry something in the overlay only on one of the two streams. Right now, `overlay.d/` is purely a cosa concept, so there's no way to make it conditional if we have a single source branch.

I think it'd be cleaner if we folded the `overlay.d/` concept into the treefile (see related discussions in https://github.com/coreos/coreos-assembler/pull/639). Then we'd have the `manifest.yaml` be able to fully describe stream-level deltas.

Another cleanup from doing this is that intermediate manifests would be able to reference the overlays to which they "correspond". (E.g. `05core` would be in `fedora-coreos-base.yaml`, `15fcos` in `fedora-coreos.yaml`, etc... things that we're currently documenting in the [README.md](https://github.com/coreos/fedora-coreos-config/blob/2ee95eba30d38a7ef53c50ff52590ad1a27b92b6/overlay.d/README.md)).

jlebon:
:telephone_receiver: ?
/cc @dustymabe @bgilbert

dustymabe:
What is fedora-next.repo? Is it rawhide? Is it f31? If it's f31 then why not just let the releasever control it and use fedora.repo ?

jlebon:
> What is fedora-next.repo?

It's the set of repos that `next` targets. So right now, f31.

> Is it f31? If it's f31 then why not just let the releasever control it and use fedora.repo ?

The paths are not the same yet.

dustymabe:
> The paths are not the same yet.

right.. if we were using the metalink URLs it wouldn't matter but since we are using stable URLs (i.e. no mirrors) it does. Got ya.

dustymabe:
this LGTM. Regarding the "managing the config" I'm afraid I don't have a good answer right now :(. In Fedora we use a branch for each release and then manually pull back stuff that we need. In this repo it's not always clear where the most development will be happening (i.e. not one clear upstream to everything like rawhide is in Fedora proper) so we'll have to organically come up with something that works for us.

dustymabe (approved):

jlebon:
Will merge this after promoting for a release!

PR #181, opened 2019-09-24
manifest: supplement rpm-ostreed.conf with info about zincati

miabbott:
Users who are previously familiar with `rpm-ostree` based systems may
want to configure automatic updates in the `rpm-ostreed.conf` file.
This post-process script replaces the default conf file with one that
has some more information about using `zincati`.

### DIFF
--- a/fedora-coreos.yaml
+++ b/fedora-coreos.yaml
@@ -49,6 +49,22 @@ postprocess:
     #!/usr/bin/env bash
     set -xeuo pipefail
     sed -i -e 's/CoreOS/CoreOS preview/' $(realpath /etc/os-release)
+  # Users shouldn't be configuring `rpm-ostreed.conf`
+  # https://github.com/coreos/fedora-coreos-tracker/issues/271
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    cat > /tmp/rpm-ostreed.conf << 'EOF'
+    # By default, this system has its OS updates managed by
+    # `zincati.service`.  Changes made to this file may
+    # conflict with the configuation of `zincati.service`.
+    # See https://github.com/coreos/zincati for additional
+    # information.
+
+    EOF
+    cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
+    cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
+    rm /tmp/rpm-ostreed.conf
 
 remove-from-packages:
   # Drop NetworkManager support for ifcfg files, see also corresponding


miabbott:
Err...this worked locally.  Will debug.

cgwalters:
No opposition to this but I think effort here is better spent on https://github.com/coreos/rpm-ostree/issues/1747

miabbott:
Manged to muck up my `git` history somehow, but the latest commit should be right :arrow_up:

fedora-coreos.yaml
```diff
@@ -49,6 +49,22 @@ postprocess:
...
+  # Users shouldn't be configuring `rpm-ostreed.conf`
+  # https://github.com/coreos/fedora-coreos-tracker/issues/271
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    cat > /tmp/rpm-ostreed.conf << 'EOF'
+    # By default, this system has its OS updates managed by
+    # the `zincati` service.  Changes made to this file may
+    # conflict with the configuation of `zincati`.
+    # See https://github.com/coreos/zincati for additional
```
cgwalters, line 61: I'd say at least once here `zincati.service`.   The systemd unit name *usually* matches upstream, and does in this case, but not always - it's good to be unambiguous, since the systemd unit is the main "touch point" for administrators, not the binary (or RPM package name).

cgwalters (approved):

lucab:
LGTM. For the future, I'd like to leave the /etc path for user-customizations and move our distro-defaults to /usr.
@cgwalters @jlebon do you mind if I move that part of this conversation to an rpm-ostree ticket?

miabbott:
Incorporated feedback in new commit :arrow_up:

Thanks folks!

fedora-coreos.yaml
```diff
@@ -49,6 +49,22 @@ postprocess:
...
+    # By default, this system has its OS updates managed by
+    # `zincati.service`.  Changes made to this file may
+    # conflict with the configuation of `zincati.service`.
+    # See https://github.com/coreos/zincati for additional
+    # information.
+
+    EOF
+    cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
+    cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
+    rm /tmp/rpm-ostreed.conf
```
jlebon, line 67: Super minor: why not just `mv /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf` ?

jlebon (approved):
LGTM, just a tiny optional nit.

jlebon:
>  @cgwalters @jlebon do you mind if I move that part of this conversation to an rpm-ostree ticket?

Sure, WFM.

fedora-coreos.yaml
miabbott: My thought was `cp` would be more reliable than `mv`?  But I don't know where I got that idea from...

lucab:
Followup at https://github.com/coreos/rpm-ostree/issues/1913.

PR #182, opened 2019-09-25
overlay: Drop dracut/40coreos-var

cgwalters:
Will move to https://github.com/coreos/ignition-dracut/pull/112
because it's cleaner to have the ignition+ostree in one place.


jlebon:
The reason this was moved here was to try to keep `ignition-dracut` as distro-independent as possible. Not an easy task of course, given that it's all about distro integration in the first place. But I think it's worked OK so far.

See also discussions in https://github.com/coreos/fedora-coreos-config/pull/70 around this. Notably, one semi-realistic idea I had was to fold ignition-dracut back into the ignition repo.

cgwalters:
> The reason this was moved here was to try to keep ignition-dracut as distro-independent as possible. Not an easy task of course, given that it's all about distro integration in the first place. But I think it's worked OK so far.

Did you see the details in
https://github.com/coreos/ignition-dracut/pull/112#issue-320970559
?  Let's discuss there?

cgwalters:
Obsoleted by https://github.com/coreos/fedora-coreos-config/pull/187

PR #184, opened 2019-09-26
Add general ignition-ostree module with rootfs replacement

cgwalters:
Previously the `40coreos-var` module was "special /var
handling for Ignition + OSTree".

This new module takes over that, renaming it to `ignition-ostree`,
and extends it with support for replacing the root filesystem.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+set -euo pipefail
+
+rootdisk=/dev/disk/by-label/root
+rootmnt=/sysroot
+tmproot=/run/ignition-ostree-rootfs
+saved_rootmnt=${tmproot}/orig-sysroot
+memdev=/dev/ram0
+
+case "${1:-}" in
+    detect)
+        # This is obviously crude; perhaps in the future we could change ignition's `fetch`
+        # stage to write out a file if the rootfs is being replaced or so.  But eh, it
+        # works for now.
+        has_rootfs=$(jq '.storage?.filesystems? // [] | map(select(.label == "root")) | length' < /run/ignition.json)
+        if [ "${has_rootfs}" = "0" ]; then
+            exit 0
+        fi
+        echo "Detected rootfs replacement in fetched Ignition config: /run/ignition.json"
+        mkdir "${tmproot}"
+        ;;
+    save)
+        size=$(lsblk -bn -o SIZE "${rootdisk}")
+        sizekbs="$(($size / 1024 + 1))"
+        if lsmod | grep -q brd; then
+            echo 'error: brd module is already loaded' 1>&2; exit 1
+        fi
+        modprobe brd rd_nr=1 rd_size="$sizekbs" max_part=1
+        echo "Moving rootfs to RAM..."
+        dd "if=${rootdisk}" "of=${memdev}" bs=8M
+        echo "Moved rootfs to RAM, pending redeployment: ${memdev}"
+        ;;
+    restore)
+        # This one is in a private mount namespace since we're not "offically" mounting
+        mount "$rootdisk" $rootmnt
+        # This can occur when specifying `wipeFilesystem: false`; TODO detect that above
+        if [ -d "${rootmnt}/boot" ]; then
+            echo "NOTE: Detected Ignition rootfs replacement, but filesystem is not empty"
+            exit 0
+        fi
+        echo "Restoring rootfs from RAM..."
+        mkdir "${saved_rootmnt}"
+        mount "${memdev}" "${saved_rootmnt}"
+        # Remove the immutable bits so we can use `mv`
+        chattr -i ${saved_rootmnt} ${saved_rootmnt}/ostree/deploy/*/deploy/*.0
+        for x in .coreos-aleph-version.json boot ostree; do
+            mv -Tn ${saved_rootmnt}/${x} ${rootmnt}/${x}
+        done
+        # And restore the immutable bits
+        chattr +i ${rootmnt}/ostree/deploy/*/deploy/*.0 ${rootmnt}
+        echo "...done"
+        umount $rootmnt
+        umount $saved_rootmnt
+        rmmod brd
+        rm -rf "${tmproot}"
+        ;;
+    *)
+        echo "Unsupported operation: ${1:-}" 1>&2; exit 1
+        ;;
+esac
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
@@ -7,13 +7,16 @@ tmproot=/run/ignition-ostree-rootfs
 saved_rootmnt=${tmproot}/orig-sysroot
 memdev=/dev/ram0
 
+# Note that save & restore run in private mount namespaces, hence why we don't
+# bother unmounting things here.
+
 case "${1:-}" in
     detect)
         # This is obviously crude; perhaps in the future we could change ignition's `fetch`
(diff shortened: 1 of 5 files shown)


cgwalters:
This is work for https://github.com/coreos/fedora-coreos-tracker/issues/94

ajeddeloh:
This is blocked by the rootfs mount options issue, right?

cgwalters:
> This is blocked by the rootfs mount options issue, right?

Working on that!

overlay.d/05core/usr/lib/dracut/modules.d/31ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -euo pipefail
+
```
darkmuggle: perhaps `.  /lib/dracut-crypt-lib.sh`

And then use `info` over `echo`? That will make sure that the dracut messages are emitted.

darkmuggle (approved):
I like the direction. My only question is whether reading it to an in-ram location (tar -cf ...) might be faster since you avoid the overhead of the delete process. For some cloud-based network attach storage, there's going to be a cost and avoiding that could make it marginally faster.

cgwalters:
Will rebase on https://github.com/coreos/fedora-coreos-config/pull/187

cgwalters:
> My only question is whether reading it to an in-ram location (tar -cf ...) might be faster since you avoid the overhead of the delete process. 

Yeah I started with that, but hit tar+SELinux issues.  My plan is to put this in rpm-ostree, see https://github.com/coreos/rpm-ostree/pull/1911

overlay.d/05core/usr/lib/dracut/modules.d/31ignition-ostree/ignition-ostree-dracut-rootfs.sh
ajeddeloh: Why `dracut-crypt-lib.sh` over `dracut-lib.sh`?

overlay.d/05core/usr/lib/dracut/modules.d/31ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -0,0 +1,54 @@
...
+        # This one is in a private mount namespace since we're not "offically" mounting
+        mount /dev/disk/by-label/root $rootmnt
+        echo "Moving rootfs to RAM..."
+        # OSTree added the immutable bit on the deployment root, and
+        # cosa's create_disk added it to the rootfs
+        chattr -i ${rootmnt} ${rootmnt}/ostree/deploy/*/deploy/*.0
+        for x in boot ostree; do
+            # TODO; copy instead of mv to avoid writes, since we're just
+            # about to blow away the whole FS anyways?
+            mv -Tn ${rootmnt}/${x} ${tmproot}/${x}
```
ajeddeloh: We should copy in case we don't actually blow away root.

overlay.d/05core/usr/lib/dracut/modules.d/31ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -0,0 +1,54 @@
...
+        # This one is in a private mount namespace since we're not "offically" mounting
+        mount /dev/disk/by-label/root $rootmnt
+        # This can occur when specifying `wipeFilesystem: false`; TODO detect that above
+        if [ -d "${rootmnt}/boot" ]; then
+            echo "NOTE: Detected Ignition rootfs replacement, but filesystem is not empty"
+            exit 0
+        fi
+        echo "Restoring rootfs from RAM..."
+        for x in boot ostree; do
+            mv -Tn ${tmproot}/${x} ${rootmnt}/${x}
```
ajeddeloh: What's the impact of this and hardlinks? Should we be doing ostree operations instead of just mv/cp?

overlay.d/05core/usr/lib/dracut/modules.d/31ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -0,0 +1,54 @@
...
+        done
+        # And restore the immutable bits
+        chattr +i ${rootmnt}/ostree/deploy/*/deploy/*.0 ${rootmnt}
+        echo "...done"
+        umount $rootmnt
+        ;;
+    *)
+        echo "Unsupported operation: ${1:-}"
```
ajeddeloh: we should die hard here

overlay.d/05core/usr/lib/dracut/modules.d/31ignition-ostree/ignition-ostree-dracut-rootfs.sh
cgwalters: Yeah, I plan to move this portion into rpm-ostree.

jlebon:
As mentioned in https://github.com/coreos/fedora-coreos-tracker/issues/94#issuecomment-633145651, I've now sent the patch upstream to be able to read labels from the initrd. In order to be confident it worked, I rebased this on top of the latest `testing-devel`. With the kernel patch + this PR, I can boot up FCOS on ext4 with SELinux on and the following Ignition config:

```json
"storage": {
    "filesystems": [
        {
            "device": "/dev/disk/by-partlabel/root",
            "format": "ext4",
            "wipeFilesystem": true,
            "label": "root"
        }
    ]
}
```

Anyway, took the liberty to force-push the rebase to this PR. I also added a follow-up commit with some tweaks. We can squash that in if you approve.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -0,0 +1,46 @@
...
+    save)
+        size=$(lsblk -bn -o SIZE "${rootdisk}")
+        sizekbs="$(($size / 1024 + 1))"
+        modprobe --first-time brd rd_nr=1 rd_size="$sizekbs" max_part=1
+        echo "Moving rootfs to RAM..."
+        dd "if=${rootdisk}" "of=${memdev}" bs=8M
+        echo "Moved rootfs to RAM, pending redeployment: ${memdev}"
+        ;;
+    restore)
+        mount "$rootdisk" $rootmnt
```
cgwalters, line 34: Aside: I am still leaning towards "rootfs reprovisioning *must* add a root= kernel argument" (UUID is a sane default, but it doesn't have to be that).  That way we avoid all races with changing the meaning of `/dev/disk/by-label/root`.

jlebon: Yeah, I think I agree with this. I feel like we should spend some time to write up some of the complex root device options we want to support (e.g. at least starting with multipath, RAID, and LUKS) and see how it fits into one cohesive picture.

jlebon:
So circling back, once we unblock the SELinux patch (I have [an RHBZ](https://bugzilla.redhat.com/show_bug.cgi?id=1845210) pending for that), this is the next crucial piece in the puzzle for https://github.com/coreos/fedora-coreos-tracker/issues/94. On its own, it will allow e.g. reconfiguring `/dev/sda4` from `xfs` to `ext4` or moving it to another disk entirely. https://github.com/coreos/fedora-coreos-config/pull/503 builds on it to support targeting complex devices.

It's been pretty solid for me while hacking on https://github.com/coreos/fedora-coreos-tracker/issues/94. Though we should get some fresh eyes on it too.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
jlebon: This is done in https://github.com/coreos/fedora-coreos-config/pull/503.

jlebon:
This is now part of https://github.com/coreos/fedora-coreos-config/pull/503.

PR #185, opened 2019-09-26
overrides: add ignition-2.0.1-4.git641ec6a.fc30

jlebon:
For the new `fetch` stage for the rootfs redeploy work.

### DIFF
--- a/manifest-lock.overrides.x86_64.json
+++ b/manifest-lock.overrides.x86_64.json
@@ -1,5 +1,8 @@
 {
   "packages": {
+    "ignition": {
+      "evra": "2.0.1-4.git641ec6a.fc30.x86_64"
+    },
     "ostree": {
       "evra": "2019.4-1.fc30.x86_64",
       "digest": "sha256:96cdbdd3417be58c2c8399adf30c7efd1b74c4571460fcac34d2ea2dd5442b51"


cgwalters (approved):

manifest-lock.overrides.x86_64.json
```diff
@@ -1,5 +1,8 @@
 {
   "packages": {
+    "ignition": {
+      "evra": "2.0.1-4.git641ec6a.fc30.x86_64"
```
dustymabe, line 4: add digest?

jlebon: If we rely on coreos-koji-tagger, then we can't have a digest yet, because it's not signed yet, right? This is kinda part of https://github.com/coreos/fedora-coreos-releng-automation/issues/32 (second point)...

dustymabe: very good point. I forgot about that.

I almost wonder if we should just trim the digest from these lockfiles.

dustymabe (approved):

dustymabe:
and it's tagged:

```
2019-09-26 13:30:22,927 INFO coreos_koji_tagger - Tagging the following (tag, nvr): 
	('f30-coreos-signing-pending', 'ignition-2.0.1-4.git641ec6a.fc30')
Watching tasks (this may be safely interrupted)...
37876217 tagBuild (noarch): free
37876217 tagBuild (noarch): free -> closed
  0 free  0 open  1 done  0 failed

37876217 tagBuild (noarch) completed successfully
```

PR #186, opened 2019-09-26
overrides add zincati-0.0.5-1.module_f30+6699+1eaa40c6.x86_64

dustymabe:
Fast track the new zincati version for our next release of FCOS.


dustymabe:
cc @lucab since you requested this I believe

dustymabe:
the test failure is going to fail with `error: Packages not found: zincati` until this PR is merged so we should ignore that.

lucab (approved):
LGTM
Double-check: this matches the one available in koji https://koji.fedoraproject.org/koji/buildinfo?buildID=1391362

PR #187, opened 2019-09-27
 dracut: Add ignition-ostree-mount-sysroot.service

cgwalters:
This is inert by default because coreos-assembler adds the `root=`
kernel argument today.  But the intention is to drop those and
allow integration with Ignition rootfs replacement.

We only want to pass the `prjquota` flag for our default XFS filesystem,
so this can't be a static `[Mount]` unit type, it has to be determined
after `ignition-fetch.service`.


cgwalters:
Tested both with and without https://github.com/coreos/coreos-assembler/pull/786

darkmuggle (approved):
LGTM to me.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+# Today this is trivial, but in the future this will require integration
+# with Ignition.
+mountflags=prjquota
```
ajeddeloh, line 6: I don't know if it was ever needed, but we used to have `defaults`. Was it just redundant before?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
```diff
@@ -0,0 +1,7 @@
...
+mount -o "${mountflags}" /dev/disk/by-partlabel/root /sysroot
```
ajeddeloh: Any reason to prefer `by-partlabel` over `by-label`?

darkmuggle: In checking, I think we need to use `by-label`. For the LUKS work in RHCOS land, we'll be using, this will fail since the `partlabel` is still root. With `by-label` we should get the root defice. 

But perhaps another tweak here, @cgwalters might be to do something like: 
```
. /lib/dracut-lib.sh
root=$(getarg root)
mount -o "${mountflags}" ${ROOT} /sysroot
``` 
Then we can let `mount` figure out mount via `UUID` or `LABEL`.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
cgwalters: Just redundant.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
cgwalters: Thanks, fixed to use `by-label`.  

> root=$(getarg root)

That's exactly what the `systemd-fstab-generator` is doing though and what we're avoiding here right?

We are aiming to support reprovisioning the rootfs via Ignition (potentially fetched over network in initramfs, long after generators have run), so we can't depend on or use anything in the kernel cmdline.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-var.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Mount OSTree /var
 DefaultDependencies=false
+ConditionKernelCommandLine=ostree
```
jlebon, line 4: I'm guessing this dates from when you had that PR against ignition-dracut, right? Now that it's here... feels like we can drop this.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Mount OSTree /sysroot
+DefaultDependencies=false
+# If root is specified, then systemd's generator will win
+ConditionKernelCommandLine=!root
+ConditionKernelCommandLine=ostree
+ConditionPathExists=!/run/ostree-live
+
+Before=initrd-root-fs.target
+After=ignition-disks.service
```
jlebon, line 10: I think we want `After=initrd-root-device.target` here instead? (Which `ignition-disks.service` is already ordered against too.)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-var.service
cgwalters: See https://github.com/coreos/fedora-coreos-config/pull/187/commits/0359cc61564d4faf25d0fdecb8d49852ea9ed2ca

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.service
cgwalters: Hum...as far as I can see, `initrd-root-device.target` is only referenced by the systemd fstab generator's handling of the `root=` kernel argument.

I don't see any references to `initrd-root-device.target` in the ignition-dracut git.  How is `ignition-disks.service` ordered against it?

cgwalters: Also a tricky thing in this is that since we want to support reprovisioning the rootfs...I don't think we could e.g. say that we've reached `initrd-root-device.target` when `/dev/disk/by-label/root` appears - because we may need to tear it down and make a new one.

Oh but hm I see https://github.com/coreos/ignition-dracut/commit/4cdd02487aea163fe2541861ea0eab9383287114 is what you were talking about, I missed that before.

cgwalters: OK yeah, I think we need to take over `initrd-root-device.target` as part of supporting reprovisioning.   But we can't do it in *this* PR because we don't do reprovisioning yet.

I think...let's keep this PR as is, and try to implement the target in the later reprovisioning PR, and change this dependency then?

jlebon: Will `ignition-ostree-mount-sysroot.service` play a different role once reprovisioning is supported? Or will it always be the thing that mounts the final possibly-reprovisioned root? If so, seems like we can settle on this ordering right now and just change the "backing" for `initrd-root-device.target` independently, no?

cgwalters: So with reprovisioning, there's ~two~ three mounts of the root device.  The way I have things currently in [the PR](https://github.com/coreos/fedora-coreos-config/pull/184/files) is that the reprovisioner uses a private mount namespace, to help ensure that nothing else sees it, i.e. it "owns" that mount until this unit comes through.

jlebon: OK right, so systemd generates this:

```
# cat /run/systemd/generator/initrd-root-device.target.d/50-root-device.conf

# Automatically generated by systemd-fstab-generator

[Unit]
Requires=dev-disk-by\x2dlabel-root.device
After=dev-disk-by\x2dlabel-root.device
```

Which yeah, we'll want to do something better here.
Anyway, let's get this one in!

jlebon (approved):

PR #188, opened 2019-10-02
tree: promote changes from testing-devel at 3e78f9f

jlebon:
…015e845eda4027d


dustymabe:
Should we do this first: https://github.com/coreos/fedora-coreos-config/pull/190 ? 

I notice in my local testing that the version of ostree is the "older version" that is in the override file (I think).

jlebon:
OK, it looks like `ostree-2019.4-2` wasn't tagged into the pool: https://koji.fedoraproject.org/koji/buildinfo?buildID=1391500, which is why CI is failing.

dustymabe:
> OK, it looks like `ostree-2019.4-2` wasn't tagged into the pool: https://koji.fedoraproject.org/koji/buildinfo?buildID=1391500, which is why CI is failing.

The issue has been resolved.. Changes LGTM

dustymabe (approved):

PR #190, opened 2019-10-02
lockfiles: drop overrides

dustymabe:
The versions in Fedora stable have matched or moved past the
ostree/rpm-ostree versions in the overrides file.


jlebon:
That's the same as https://github.com/coreos/fedora-coreos-config/pull/189, right?

dustymabe:
yeah sorry.

PR #191, opened 2019-10-02
overlay.d: teach coreos-growpart about growing LUKS rootfs

darkmuggle:
In order to support the rootfs in LUKS containers, `coreos-growpart`
needs awareness of mapping the device mapper target to the underlying
device. This change only support LUKS for now.

### DIFF
--- a/overlay.d/05core/usr/libexec/coreos-growpart
+++ b/overlay.d/05core/usr/libexec/coreos-growpart
@@ -12,6 +12,16 @@ path=$1
 shift
 
 majmin=$(findmnt -nvr -o MAJ:MIN "$path")
+
+# Detect if the rootfs is on a luks container and map
+# it to the underlying partition. This assumes that the
+# LUKS volumes is housed in a partition.
+src=$(findmnt -nvr -o SOURCE "$path")
+is_luks=0
+if [[ "${src}" =~ /dev/mapper ]]; then
+    majmin=$(dmsetup table ${src} | cut -d " " -f7)
+fi
+
 devpath=$(realpath "/sys/dev/block/$majmin")
 partition=$(cat "$devpath/partition")
 parent_path=$(dirname "$devpath")
@@ -20,6 +30,12 @@ parent_device=/dev/$(basename "${parent_path}")
 # TODO: make this idempotent, and don't error out if
 # we can't resize.
 growpart "${parent_device}" "${partition}" || true
+
+eval $(blkid -o export "${parent_device}${partition}")
+if [ "${TYPE}" == "crypto_LUKS" ]; then
+    luks_name=$(dmsetup info ${src} -C -o name --noheadings)
+    cryptsetup resize ${luks_name}
+fi
 
 # this part is already idempotent
 xfs_growfs /sysroot


darkmuggle:
This is needed as part of the work with https://github.com/coreos/fedora-coreos-tracker/issues/287.

ajeddeloh:
@darkmuggle which part in particular?

ajeddeloh:
Looks like this script in general needs a revamp/to be replaced with something in the initramfs. What if we grew the default root partition (on FCOS, would include the LUKS container too on RHCOS) and FS before the disks stage?

Could be part of the `coreos-gpt-setup` service

overlay.d/05core/usr/libexec/coreos-growpart
```diff
@@ -12,6 +12,20 @@ path=$1
...
 majmin=$(findmnt -nvr -o MAJ:MIN "$path")
+
+# Detect if the rootfs is on a luks container and map
+# it to the underlying partition. This assumes that the
+# LUKS volumes is housed in a partition.
+src=$(findmnt -nvr -o SOURCE "$path")
+is_luks=0
+if [[ "${src}" =~ /dev/mapper ]]; then
+    majmin=$(dmsetup table ${src} | cut -d " " -f7)
+    dm_major=$(dmsetup info ${src} -C -o major --noheadings)
```
cgwalters: I think it'd be a bit cleaner to `source $(blkid -o export ${src})` and conditionalize on `TYPE=crypto_LUKS`.

cgwalters (approved):

cgwalters:
> What if we grew the default root partition (on FCOS, would include the LUKS container too on RHCOS) and FS before the disks stage?

Can't do that, it would break [creating a /var partition in unused space](https://github.com/cgwalters/playground/blob/master/fcct/var.fcct).

ajeddeloh:
Arg, you're right. Bleh.

cgwalters:
(But, growing in the initramfs would probably be better indeed)

ajeddeloh:
Yeah we could do several things:
1) What we currently do. Kinda hacky, but works
2) Resize in the initramfs pre-disks but then users would have to resize root and var if they wanted to use a separate var after root. Would only be useful for the partitions, since you can't shrink xfs. 
3) Ship a base config with that partition with size/start 0 and let Ignition resize it. Has all the same problems as 2
4) Resize _after_ disks. This would eliminate any potential problems with running out of space, but also gives the user no agency to _not_ resize.

ajeddeloh:
(Accident, didn't mean to close)

overlay.d/05core/usr/libexec/coreos-growpart
darkmuggle: Open to the idea, can you clarify? 
```
bash-4.4# findmnt -nvr -o SOURCE /
/dev/mapper/luks-00000000-0000-4000-a000-000000000002

bash-4.4# blkid -o export /dev/mapper/luks-00000000-0000-4000-a000-000000000002
DEVNAME=/dev/mapper/luks-00000000-0000-4000-a000-000000000002
LABEL=root
UUID=79b7d2d1-905c-4340-893a-c65aff1f4618
TYPE=xfs
```
The reason for the dmsetup was so that we could map the underly block device for `luks-...`.  

That said, I think using `blkid -o export  $devpath` to determine if the underlying device is a LUKS container, is a wise course.

darkmuggle:
@ajeddeloh what do you think about going with option 1 in the short term, but re-work the handling in light of the OSTree work @cgwalters is driving? Once we have support for re-writing the root partition via the OSTree dracut module, I wonder if we want to have something 

Pushed a change that reduces the code and pulls in @cgwalters suggestion.

cgwalters:
One thing Ben and I talked about in chat is that in fact systemd has `systemd-growfs@` - it supports LUKS but quite impressively omits XFS support (it'd be a trivial patch).  Anyways see e.g. https://github.com/systemd/systemd/commit/c34b75a124c6a23ed2dbf6d6b6d56d1385a2d191

Longer term I think it'd be better to get fixes into systemd, but this looks simple and easy for now.

cgwalters (approved):

PR #195, opened 2019-10-09
Pin podman due to `systemctl preset-all` regression

jlebon:
https://github.com/coreos/fedora-coreos-tracker/issues/290

The fix for this was merged, but it'll be some time before it makes it
to the stable repos. This is a temporary hack to get updates going again
into FCOS. Working on a cleaner approach to this in:

https://github.com/coreos/fedora-coreos-tracker/issues/293


cgwalters (approved):

dustymabe (approved):
LGTM

PR #196, opened 2019-10-11
overrides: add systemd-241-12.git323cdf4.fc30

bgilbert:
Fix `relabel-extra.d` for live PXE support.


jlebon:
LGTM! (Though note https://bodhi.fedoraproject.org/updates/FEDORA-2019-66b24a298e#comment-1091192).

CI failure is from https://github.com/coreos/fedora-coreos-releng-automation/issues/32#issuecomment-535626797.

PR #200, opened 2019-10-15
Move to Fedora 31!

jlebon:
See individual commit messages.

Requires: https://github.com/coreos/rpm-ostree/pull/1927

Marking as draft until the f30 release is completely done (and I'd like to investigate the `efivar-libs` bit a little more).

<details>

<summary>Diff from latest `testing` release</summary>

```
Upgraded:
  NetworkManager 1:1.16.4-1.fc30 -> 1:1.20.4-1.fc31
  NetworkManager-libnm 1:1.16.4-1.fc30 -> 1:1.20.4-1.fc31
  acl 2.2.53-3.fc30 -> 2.2.53-4.fc31
  adcli 0.8.2-3.fc30 -> 0.8.2-7.fc31
  afterburn 4.1.3-1.module_f30+6561+89082b01 -> 4.1.3-1.module_f31+6563+49f1a537
  afterburn-dracut 4.1.3-1.module_f30+6561+89082b01 -> 4.1.3-1.module_f31+6563+49f1a537
  alternatives 1.11-4.fc30 -> 1.11-5.fc31
  audit-libs 3.0-0.9.20190507gitf58ec40.fc30 -> 3.0-0.12.20190507gitf58ec40.fc31
  avahi-libs 0.7-18.fc30 -> 0.7-20.fc31
  basesystem 11-7.fc30 -> 11-8.fc31
  bash 5.0.7-1.fc30 -> 5.0.7-3.fc31
  bash-completion 1:2.8-6.fc30 -> 1:2.8-7.fc31
  bc 1.07.1-8.fc30 -> 1.07.1-9.fc31
  bind-libs 32:9.11.10-1.fc30 -> 32:9.11.11-1.fc31
  bind-libs-lite 32:9.11.10-1.fc30 -> 32:9.11.11-1.fc31
  bind-license 32:9.11.10-1.fc30 -> 32:9.11.11-1.fc31
  bind-utils 32:9.11.10-1.fc30 -> 32:9.11.11-1.fc31
  brotli 1.0.7-3.fc30 -> 1.0.7-6.fc31
  btrfs-progs 5.2.1-1.fc30 -> 5.2.1-1.fc31
  bubblewrap 0.3.3-2.fc30 -> 0.3.3-3.fc31
  bzip2 1.0.6-29.fc30 -> 1.0.8-1.fc31
  bzip2-libs 1.0.6-29.fc30 -> 1.0.8-1.fc31
[...]

PR #202, opened 2019-10-18
[bodhi-updates] manifest: bump to Fedora 31

jlebon:
Following https://github.com/coreos/fedora-coreos-config/pull/200.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,7 +1,7 @@
 ref: fedora/${basearch}/coreos/bodhi-updates
 include: fedora-coreos.yaml
 
-releasever: "30"
+releasever: "31"
 
 rojig:
   license: MIT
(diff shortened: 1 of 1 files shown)


dustymabe:
hmm - there's technically no bodhi runs happening in f31 yet, right? I guess we're just consuming branched content for now?

manifest-lock.x86_64.json
```diff
@@ -1,7 +1,7 @@
 {
   "packages": {
     "podman": {
-      "evra": "2:1.5.1-3.fc30.x86_64"
+      "evra": "2:1.5.1-3.git0005792.fc31.x86_64"
```
dustymabe: There is a newer podman in stable?

```
$ koji latest-build f31 podman | tail -n 1
podman-1.6.1-2.fc31
```

dustymabe: ahh - actually this was a "freeze" because of a regression and I assume a fix hasn't landed yet.

dustymabe (approved):
LGTM

jlebon:
We should hold this until #200 goes in. Tests won't pass until the changes there are propagated to `bodhi-updates` anyway.

> hmm - there's technically no bodhi runs happening in f31 yet, right?

Oh, I thought it was. Isn't that how we get zero day updates?

dustymabe:
The nightly pungi run is what is controlling the stable repo for F31. There is an updates-testing compose that is running, but no updates composes - you can see it in the list of composes here: https://kojipkgs.fedoraproject.org/compose/updates/ .

Either way it mostly doesn't matter as long as the repo files are pulling in the right content anyway.

jlebon:
OK, this one is ready to go now!

dustymabe (approved):
LGTM

PR #204, opened 2019-10-22
[bodhi-updates] manifest: bump podman pin for f31

jlebon:
This matches the NEVRA we pin to in `testing-devel`:
https://github.com/coreos/fedora-coreos-config/commit/19bd54c9d007f49a36af141d9af02defced12e2a


jlebon:
Fun stuff: the config-bot propagated the f31 bump in `testing-devel` to `bodhi-updates`, but not before promoting the last `bodhi-updates` f30 lockfile down to `testing-devel`. This will fix itself of course once an f31 bodhi-updates is built, which needs this patch.

This won't be an issue anymore once we fully uncouple from bodhi-updates for lockfile bumps (https://github.com/coreos/fedora-coreos-tracker/issues/293).

miabbott (approved):

PR #205, opened 2019-10-22
overlay: Add a README to /etc/sysconfig

cgwalters:
This directory is a legacy of the Red Hat Linux days; it
was an attempt to basically declare `/etc` as broken and
start a new "/etc in your /etc".  It didn't succeed.

The only reason we carry it forward to this day is inertia
and legacy.  Other distributions (all of the Debian derivatives
for example) don't ship it.

Let's add this file here starting with FCOS; we can bring
it to the other Fedora distributions later.

### DIFF
--- a/overlay.d/09misc/etc/sysconfig/README
+++ b/overlay.d/09misc/etc/sysconfig/README
@@ -1,0 +1,9 @@
+This directory is a legacy of Red Hat Linux days.
+Do not write new software that uses configuration
+files here.  Instead your software should use a regular
+config file in `/etc/foo.conf`, a configuration directory
+such as `/etc/foo/`.
+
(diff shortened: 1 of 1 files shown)


jlebon (approved):
LGTM! Though slight bikeshed for just calling it `README` to match `/etc/rc.d/init.d/README` and since we're not really extensively using Markdown here.

cgwalters:
> Though slight bikeshed for just calling it README to match /etc/rc.d/init.d/README and since we're not really extensively using Markdown here.

Yeah...but we *should* be using Markdown in the OS :wink: 

Anyways for now, renamed to `README`.

jlebon:
Merging!

PR #206, opened 2019-10-27
overlay/ignition-ostree-mount-sysroot: Only use prjquota on XFS

cgwalters:
This helps complete the move in
https://github.com/coreos/coreos-assembler/pull/786
to have our rootflags be fully dynamic.

It's also necessary for ext4+fsverity by default.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
@@ -1,7 +1,16 @@
 #!/bin/bash
 set -euo pipefail
 
-# Today this is trivial, but in the future this will require integration
-# with Ignition.
-mountflags=prjquota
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
```diff
@@ -1,7 +1,16 @@
 #!/bin/bash
 set -euo pipefail
 
-# Today this is trivial, but in the future this will require integration
-# with Ignition.
-mountflags=prjquota
-mount -o "${mountflags}" /dev/disk/by-label/root /sysroot
+# We use prjquota on XFS by default to aid multi-tenant
+# Kubernetes (and other container) clusters.  Se
```
jlebon: Minor: See

jlebon (commented):
Hmm, would a better title for this be "Only use prjquota by default on XFS" ? That's essentially what changed here, right?

cgwalters:
Reworded!

jlebon (approved):

PR #207, opened 2019-10-29
overlay: Bind ostree-prepare-root.service to ignition-complete.target

cgwalters:
This unit comes from ostree upstream, but forms part of our
Ignition sequence.  Make it required by ignition-complete.target
so that if it fails we get the same OnFailure= behavior.

Making this change as during development `ignition-ostree-mount-sysroot.service`
wasn't running which caused this unit to fail; the root cause would have been
a bit more obvious with this change.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -30,4 +30,9 @@ install() {
     install_ignition_unit ignition-ostree-mount-sysroot.service
     inst_script "$moddir/ignition-ostree-mount-sysroot.sh" \
         "/usr/sbin/ignition-ostree-mount-sysroot"
+
+    # This unit comes from ostree upstream, but forms part of our
+    # Ignition sequence.  Make it required by ignition-complete.target
+    # so that if it fails we get the same OnFailure= behavior.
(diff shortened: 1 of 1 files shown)


jlebon:
Not totally against, though `ignition-complete.target` is meant to be a master on/off switch for whether to run Ignition (see https://github.com/coreos/ignition-dracut/pull/55). So it's only really useful on first boot. Adding non-firstboot-only services under it means there will be different behaviours wrt failures between first and subsequent boots for that unit, and can also give off the impression that `ostree-prepare-root.service` is a firstboot-only service too.

Hmm actually, `ostree-prepare-root.service` already has `OnFailure=emergency.target`. So maybe we just need:

```diff
diff --git a/src/boot/ostree-prepare-root.service b/src/boot/ostree-prepare-root.service
index 63357581..1017c2fd 100644
--- a/src/boot/ostree-prepare-root.service
+++ b/src/boot/ostree-prepare-root.service
@@ -22,6 +22,7 @@ DefaultDependencies=no
 ConditionKernelCommandLine=ostree
 ConditionPathExists=/etc/initrd-release
 OnFailure=emergency.target
+OnFailureJobMode=isolate
 After=sysroot.mount
 Before=initrd-root-fs.target

```

for [the same reason as `ignition-complete.target`](https://github.com/coreos/ignition-dracut/blob/9a25a5e8eb7061f59ed0d50baa487e3095b5be19/dracut/30ignition/ignition-complete.target#L10-L13)?

cgwalters:
Yep, you're totally right.  Will do.

PR #211, opened 2019-10-31
[testing] manifest.yaml: adapt for new path to fedora-coreos.yaml 

jlebon:
This was changed on `testing-devel`.

While we're here, I also bumped the releasever to 31 in prep for the next release.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,7 +1,7 @@
 ref: fedora/${basearch}/coreos/testing
 include: manifests/fedora-coreos.yaml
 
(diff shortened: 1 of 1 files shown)


dustymabe (approved):

dustymabe:
thanks @jlebon

jlebon:
Tests are naturally going to fail until we do a promotion with the new paths. I guess we can just hold this until then.

dustymabe:
I forgot about #211. Rebasing this PR on top of that one.

PR #217, opened 2019-11-01
manifest: Add Ignition override

cgwalters:
To help us merge
https://github.com/coreos/coreos-assembler/pull/883


cgwalters:
> error: Packages not found: ignition

Right...that's what I thought would block this, we have a deadlock.

jlebon (approved):
LGTM. CI failure is due to https://github.com/coreos/fedora-coreos-releng-automation/issues/32#issuecomment-535626797.

PR #218, opened 2019-11-05
manifest: use socat instead of nmap-ncat

vrutkovs:
This binary is required for e2e tests for OKD-on-FCOS project. It 
probably barely useful for enduser, so it might be worth keeping this 
package in `testing-devel` and not backport to release branches

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -795,9 +795,6 @@
     "nftables": {
       "evra": "1:0.9.1-3.fc31.x86_64"
     },
-    "nmap-ncat": {
-      "evra": "2:7.80-2.fc31.x86_64"
-    },
     "npth": {
       "evra": "1.6-3.fc31.x86_64"
     },
@@ -974,6 +971,9 @@
     },
     "slirp4netns": {
       "evra": "0.4.0-20.1.dev.gitbbd6f25.fc31.x86_64"
+    },
+    "socat": {
+      "evra": "1.7.3.3-2.fc31.x86_64"
     },
     "sqlite-libs": {
       "evra": "3.30.0-1.fc31.x86_64"
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -132,7 +132,7 @@ packages:
   - logrotate
   # Used by admins interactively
   - sudo coreutils less tar xz gzip bzip2
-  - nmap-ncat net-tools bind-utils
+  - socat net-tools bind-utils
   - bash-completion
   - openssl
   - vim-minimal


jlebon:
Is this the only dep missing for e2e tests to run?

manifests/file-transfer.yaml
```diff
@@ -3,3 +3,4 @@ packages:
   - git-core
   - gnupg2
   - rsync
+  - socat
```
jlebon: Hmm, maybe in the "Networking" section of `fedora-coreos-base.yaml` instead? Probably worth a comment too about the OpenShift e2e tests.

jlebon (commented):
I'm good with this assuming this is the only dep.
/cc @bgilbert @dustymabe 

> so it might be worth keeping this
package in `testing-devel` and not backport to release branches

The way releases work, we can't really do that: `testing-devel` is promoted as is to `testing`. Using the exact same package set is a big part of the process.

vrutkovs:
>Is this the only dep missing for e2e tests to run?

Yes. There is also a fix for `nfs-utils-coreos` coming.

Other deps might be mixed in machine-os-content for OKD, but this package might be useful for file transmission (and maybe running kubelet too)

manifests/file-transfer.yaml
vrutkovs: Good idea, fixed

cgwalters:
blast from the past xref

 - https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2015-March/msg00066.html
 - https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2018-August/msg00001.html

lucab:
Related datapoint: we are already shipping `nmap-ncat` https://github.com/coreos/fedora-coreos-config/blob/22fc6e48096a1457a0d25a1672ba6ff303de5c1c/manifest-lock.x86_64.json#L798-L800

Can we consolidate on a single socket tool?

vrutkovs:
Note, that crio requires socat as well - see https://github.com/cri-o/cri-o/blob/55ccb1e9bd8508ced18198444a56fd51daa76611/internal/oci/runtime_oci.go#L809 and https://github.com/cri-o/cri-o/blob/5779b74e9e4ca11487b9c823bd4dc0dc183c65d6/contrib/test/integration/system-packages.yml#L117. Fedora spec is about to be updated by @lsm5

ajeddeloh:
I'm normally very hesitant about adding things to the distro, but in addition to being needed, socat is a wonderful swiss army knife for debugging and not very big. :+1:

bgilbert:
We've historically kept socat out of Container Linux (https://github.com/coreos/bugs/issues/1114), and https://github.com/coreos/fedora-coreos-config/pull/218#issuecomment-549868823 notes a similar history for AH.  I'm okayish with adding it, but let's make sure that's really what we want to do.

LorbusChris:
I'm +1 for adding this. We need this on the host for OKD and we don't have a way to properly overlay RPMs in a containerized fashion (i.e. without a VM).

If we want to consolidate all sockets tools, can we use socat for all? Otherwise I vote to get this in now, and try to optimize later.

lsm5:
> Note, that crio requires socat as well - see https://github.com/cri-o/cri-o/blob/55ccb1e9bd8508ced18198444a56fd51daa76611/internal/oci/runtime_oci.go#L809 and https://github.com/cri-o/cri-o/blob/5779b74e9e4ca11487b9c823bd4dc0dc183c65d6/contrib/test/integration/system-packages.yml#L117. Fedora spec is about to be updated by @lsm5

Hopefully these should be handled by:
https://bodhi.fedoraproject.org/updates/FEDORA-MODULAR-2019-1c7afa1aa5
https://bodhi.fedoraproject.org/updates/FEDORA-MODULAR-2019-94a453d89b

lucab:
It turns out nothing in FCOS seems to depend on `nmap-ncat` (from a quick `rpm -q --whatrequires`) and we are bringing that in manually via manifest.
For this specific case, I'm +1 on using this PR to just `s/nmap-ncat/socat/` on our config as a zero-sum replacement.

vrutkovs:
Sounds good to me, replaced `nmap-ncat` with `socat`

cgwalters (approved):
Yeah, seems fine to me.

dustymabe (approved):

PR #219, opened 2019-11-06
overlay: Tweak coreos-boot-mount-generator to deal with older RHCOS

cgwalters:
I was debugging a customer's test system that was installed as
OpenShift 4.2 and then upgraded, and noticed rpm-ostreed wasn't
starting on reboot, which was because `/boot` wasn't mounted,
which was because the `systemd-fstab-generator` and this code
had actually raced and generated a corrupted `/run/systemd/boot.mount`.

Tweak this code so it cleanly exits and defers to systemd-fstab-generator
for RHCOS-4.2 or older systems.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -11,6 +11,14 @@ add_wants() {
     ln -sf "../${name}" "${wants_dir}/${name}"
 }
 
+# If there's no aleph version from https://github.com/coreos/coreos-assembler/pull/768
+# then this is a non-FCOS system, likely RHCOS pre-4.3.  In
+# that case, we don't want to overwrite what the systemd-fstab-generator
+# will do.
+if ! [ -f /sysroot/.coreos-aleph-version.json ]; then
+    exit 0
+fi
+
 # Don't create mount units for /boot or /boot/efi on live systems.
 # ConditionPathExists won't work here because conditions don't affect
 # the dependency on the underlying device unit.


cgwalters:
I tested this didn't break FCOS, but didn't test RHCOS with it yet because doing so on latest pulls in https://github.com/coreos/fedora-coreos-config/commit/52dd9db13dd83400356fb0fad3c4812530304a93 which in turn requires https://github.com/coreos/ignition-dracut/pull/132

cgwalters:
Update: tested a 4.2 :arrow_right: 4.3 upgrade (by rebasing this commit on one before https://github.com/coreos/fedora-coreos-config/commit/52dd9db13dd83400356fb0fad3c4812530304a93) and confirmed this fixes the bug.

cgwalters:
And yes, this is some RHCOS specifics in the FCOS config, but it's hard to get a cleaner solution - we're going to get *some* of these types of bits when the projects are sharing code.  This one is pretty minimal.

(Perhaps at some point we could teach the MCO or RHCOS to migrate `/boot` out of `/etc/fstab` and be able to drop this, but...not trivial)

cgwalters:
Tweaked this to just test for the aleph version, since it's the cleanest and most accurate test.

miabbott (approved):
Makes sense to me!

jlebon:
Likely regression in https://github.com/coreos/fedora-coreos-tracker/issues/322.

jlebon:
Any reason we're not just checking `/etc/fstab` here?

jlebon:
Follow-up to this here: https://github.com/coreos/fedora-coreos-config/pull/247

PR #220, opened 2019-11-07
overlay/populate-var: Add missing realpath dependency

cgwalters:
This was hitting RHCOS, something else must be pulling it in on FCOS.


ajeddeloh (approved):
LGTM

You mean mount-var, not populate-var, right?

cgwalters:
Yep, tweaked the commit message.

cgwalters:
OK, counting the earlier review as sufficient!

PR #221, opened 2019-11-07
05core: disable pwquality dictionary checking

### LINKED ISSUE #308: Can't change password for a user complains about missing dictionary
So I add a user to the wheel group, and sudo requires a password.
Attempting sudo passwd usename fails due to missing dictionary.

bgilbert:
We don't ship a cracklib dictionary, and without one, `pam_pwquality` is preventing password changes.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/308.

### DIFF
--- a/overlay.d/05core/etc/security/pwquality.conf.d/20-disable-dict.conf
+++ b/overlay.d/05core/etc/security/pwquality.conf.d/20-disable-dict.conf
@@ -1,0 +1,3 @@
+# We don't ship cracklib dicts, so don't try to use them to validate
+# password changes.
+dictcheck = 0


cgwalters (approved):

lucab:
Self-note: I did a quick double-check and confirmed that `libpwquality` does not (currently) support fragments in `/usr/lib/security/`.

lucab (commented):
Minor: I'd add a `xy-` prefix to the filename for consistent sorting with any other user customizations in the same directory.

dustymabe (approved):
LGTM

bgilbert:
Updated.

PR #226, opened 2019-11-11
overlay/mount-generator: Test directly for /root/.squashfs

cgwalters:
This should close a race I hit while trying out a live system
(Silverblue-based-on-FCOS).

The generators here run in parallel; so we can't test for
`/run/ostree-live` in `coreos-boot-mount-generator` while
writing it in `live-generator.`

(Probably we should fold these generators together, but
 this is a small simple patch)

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -13,7 +13,7 @@ add_requires() {
     ln -sf "../${name}" "${requires_dir}/${name}"
 }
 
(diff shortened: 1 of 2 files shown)


cgwalters:
(Not tested locally yet since my toolbox devenv blew up for some reason)

cgwalters:
Tested now and works with a FCOS live build - at least doesn't break anything.

ajeddeloh:
cc @bgilbert

cgwalters:
Nevermind, the bug is something else.  The live generator lives in the initramfs, the other in the real rootfs, so they can't race.

PR #227, opened 2019-11-11
overrides: switch to YAML

jlebon:
For those sweet sweet comments.

### DIFF
--- a/manifest-lock.overrides.x86_64.yaml
+++ b/manifest-lock.overrides.x86_64.yaml
@@ -1,0 +1,5 @@
+packages:
+  # https://github.com/coreos/ignition/pull/846
+  # https://bodhi.fedoraproject.org/updates/FEDORA-2019-8f331fb834
+  ignition:
+    evra: 2.0.1-9.gita8f91fa.fc31.x86_64
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ the corresponding entries in the lockfiles:
 - for packages which should follow Bodhi updates, place them
   in `manifest-lock.$basearch.json`
 - for packages which should remain pinned, place them
-  in `manifest-lock.overrides.$basearch.json`
+  in `manifest-lock.overrides.$basearch.yaml`
 
 There will be better tooling to come to enable this, though
 one easy way to do this is for now:
--- a/manifest-lock.overrides.x86_64.json
+++ b/manifest-lock.overrides.x86_64.json
@@ -1,7 +1,0 @@
-{
-  "packages": {
-    "ignition": {
-      "evra": "2.0.1-9.gita8f91fa.fc31.x86_64"
-    }
-  }
-}


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/914

miabbott (approved):
Nice!

manifest-lock.overrides.x86_64.yaml
```diff
@@ -0,0 +1,9 @@
+packages:
+  # https://github.com/coreos/ignition-dracut/pull/131
+  # https://bodhi.fedoraproject.org/updates/FEDORA-2019-dfc1bf4727
+  ignition:
+    evra: 2.0.1-5.git641ec6a.fc31.x86_64
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1768897
+  # not in Bodhi yet
+  nfs-utils-coreos:
+    evra: 2.4.1-2.rc1.fc31.x86_64
```
dustymabe: add in epoch here

dustymabe:
I'm so happy about switching to yaml so we can get comments. Thanks @jlebon!

A few things:

- what about the other lockfile? Should we delete it or convert it to yaml as part of this PR? I know the config bot syncs things so maybe you are planning on it getting deleted/updated that way?
- Before we merge this we need to teach coreos-koji-tagger to look for/read yaml files instead of json

dustymabe:
>* what about the other lockfile? Should we delete it or convert it to yaml as part of this PR? I know the config bot syncs things so maybe you are planning on it getting deleted/updated that way?

It looks like from https://github.com/coreos/coreos-assembler/pull/914 you're only allowing this for the overrides lockfile. It seems like it would be better if we had our lockfiles in a consistent format. I realize that the non-overrides files should probably only be updated by machines, but it would be nice if they were in a consistent format for all lockfiles.

jlebon:
> Before we merge this we need to teach coreos-koji-tagger to look for/read yaml files instead of json

Ahh yes, good catch!

> I realize that the non-overrides files should probably only be updated by machines

Yeah, that's basically why I went this way. I think having them be separate formats highlights nicely the difference in purpose: one is meant for humans, the other is meant for machines.

> but it would be nice if they were in a consistent format for all lockfiles.

This would require changing rpm-ostree to output the lockfile in YAML since it's taken as-is (or alternatively, convert from JSON to YAML in config-bot). I agree consistency is nice, though. (Does it help that JSON is technically a subset of YAML? ;) ).

So... not sure, leaning more towards this approach, though we could think on it some more and see what others think? I don't think it's worth blocking on though. This is still a nice quality of life improvement overall.

dustymabe:
> > Before we merge this we need to teach coreos-koji-tagger to look for/read yaml files instead of json
> 
> Ahh yes, good catch!

I opened https://github.com/coreos/fedora-coreos-releng-automation/issues/53. Since I have a batch of work to do there I might as well make the switch and test it as well.

> 
> > I realize that the non-overrides files should probably only be updated by machines
> 
> Yeah, that's basically why I went this way. I think having them be separate formats highlights nicely the difference in purpose: one is meant for humans, the other is meant for machines.
> 
> > but it would be nice if they were in a consistent format for all lockfiles.
> 
> This would require changing rpm-ostree to output the lockfile in YAML since it's taken as-is (or alternatively, convert from JSON to YAML in config-bot). I agree consistency is nice, though. (Does it help that JSON is technically a subset of YAML? ;) ).

:)

> 
> So... not sure, leaning more towards this approach, though we could think on it some more and see what others think? I don't think it's worth blocking on though. This is still a nice quality of life improvement overall.

Yeah, don't need to block (other than https://github.com/coreos/fedora-coreos-releng-automation/issues/53) . My personal preference/vote is for consistency and for it to be yaml since that allows us comments.

cgwalters:
> It seems like it would be better if we had our lockfiles in a consistent format.

It's a valid point, but on the other hand, that seems more like a "nice to have" than a blocker.  IOW I tend to lean towards merging PRs like this and circle back to "nice to have"s later.

jlebon:
OK, updated this! Still blocks on https://github.com/coreos/fedora-coreos-releng-automation/issues/53 though.

dustymabe:
> OK, updated this! Still blocks on [coreos/fedora-coreos-releng-automation#53](https://github.com/coreos/fedora-coreos-releng-automation/issues/53) though.

now fixed.. so unblocked!

dustymabe:
maybe let's rebase this on top of #250 or the other way around?

jlebon:
Rebased and dropped `hold` label!

manifest-lock.overrides.x86_64.yaml
```diff
@@ -0,0 +1,5 @@
+packages:
+  # https://github.com/coreos/ignition/pull/846
+  # https://bodhi.fedoraproject.org/updates/FEDORA-2019-c2e727aa40
```
dustymabe: maybe https://bodhi.fedoraproject.org/updates/FEDORA-2019-8f331fb834 instead?

jlebon: Oh whoops! Fixed, thanks.

dustymabe (approved):

PR #229, opened 2019-11-11
lockfiles: nfs-utils has an epoch of 1

dustymabe:
Also we need rpcbind since that was a dependency that was added. This
would get synced over by the config updater eventually but I'd like to
go ahead and add it so I can test.


jlebon:
Yup, thanks for the fix! I had this locally as well, but was trying to whip something up for https://github.com/coreos/fedora-coreos-releng-automation/issues/32#issuecomment-541134070 too so we avoid this in the future. I think I need to write it up in a real language.

Ironically, because the Koji build ID doesn't include the epoch, the tagger did tag it in successfully, so we can wait for CI this time to make sure it gets pulled in alright.

dustymabe:
> Ironically, because the Koji build ID doesn't include the epoch, the tagger did tag it in successfully, so we can wait for CI this time to make sure it gets pulled in alright.

looks like CI passed.. Merge?

PR #231, opened 2019-11-15
overlay: Run coreos-growpart.service in basic.target

cgwalters:
Otherwise, services running on boot may see a much smaller disk
than expected.

We hit this in RHCOS where we needed to order `kubelet.service`
`After=coreos-growpart.service` because we were pulling a lot
of container images.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/coreos-growpart.service
+++ b/overlay.d/05core/usr/lib/systemd/system/coreos-growpart.service
@@ -4,7 +4,8 @@
 Description=Resize root partition
 ConditionPathExists=!/run/ostree-live
 ConditionPathExists=!/var/lib/coreos-growpart.stamp
-Before=sshd.service
+DefaultDependencies=no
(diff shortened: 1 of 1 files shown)


jlebon:
I think default dependencies mean it'll always run after `basic.target`. How about also something like:

```
DefaultDependencies=no
Before=basic.target
```

?

cgwalters:
Oh wow I didn't know that we needed to specify the `Before=`:

```
  Default Dependencies
       The following dependencies are added unless DefaultDependencies=no is set:

       ·   Target units will automatically complement all configured dependencies of type Wants= or Requires= with dependencies of type After= unless DefaultDependencies=no is set in the specified units. Note that Wants= or Requires= must be defined in the target
           unit itself — if you for example define Wants=some.target in some.service, the automatic ordering will not be added.

```

cgwalters:
Thanks, updated :new:

jlebon (approved):

PR #232, opened 2019-11-15
overlay: Move coreos-growpart to the initramfs

cgwalters:
We discussed this in a few forums before (ignition-dracut PRs
and in person); there's no reason not to grow the rootfs
directly as part of our Ignition process.

This closes down all race conditions around having processes
writing to the storage before we grow it (e.g. ignition-files,
journald).

It also means that if growfs fails then provisioning fails, which
is what we want.

(One downside of this is AFAIK dracut doesn't have a way to
 say "take all dependencies of this RPM" so we have to manually
 do that.  Really at this point we should stick all of `coreutils`
 in the initramfs)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
```diff
@@ -0,0 +1,12 @@
...
+Description=Ignition OSTree: Grow root filesystem
+DefaultDependencies=false
+ConditionKernelCommandLine=ostree
+ConditionPathExists=!/run/ostree-live
+Before=initrd-root-fs.target
+After=ignition-ostree-mount-firstboot-sysroot.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-growpart /sysroot
```
cgwalters, line 11: One nice thing here is that all of the growpart code was carefully written before to reference a passed filesystem, so all we needed to do was change `/` to `/sysroot`.

lucab:
I'm overall :+1: on this.

@cgwalters the original service seemed to have some stampfile-logic, which I don't see anymore. Is that on purpose? How often do you intend this to run?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Ignition OSTree: Grow root filesystem
+DefaultDependencies=false
```
lucab, line 3: I think this may be a relic of rootfs ordering. I guess we'd prefer the initramfs to have all basic services in place when we start this. Or do we know we conflict with something?

cgwalters:
> How often do you intend this to run?

It's installed as part of `ignition-complete.target`, so it will only run when that's enabled, same as other units.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
cgwalters: Mmm...this is the same as all of the other units we have like `ignition-mount.service` etc.  I think you're probably right that we don't need this, but I'm a bit uncertain about doing it in this PR instead of doing it in ignition-dracut first.

lucab: Ack, let's defer this.

lucab (approved):
I don't see where we do the Wants/Requires dance here, but I guess you have it already in place somewhere else.
Other than that, LGTM.

cgwalters:
> I don't see where we do the Wants/Requires dance here, but I guess you have it already in place somewhere else.

https://github.com/coreos/fedora-coreos-config/pull/232/files#diff-8cb5ddbe0f8a095150b80718304c21b8R37

jlebon (approved):
Makes sense to me. Will leave open for a bit more before merging.

PR #233, opened 2019-11-16
    overlay/live: Support "fulliso" path as well

cgwalters:
Supports https://github.com/coreos/coreos-assembler/pull/928
   
Here `fulliso` images have a few changes; what this code
is concerned with is finding the squashfs in the ISO, not the initrd.
    
We continue to support `live` images as they existed before
as well.


cgwalters:
I'm going to rework this to support both `live` and `fulliso` cases.

cgwalters:
Now tested in concert with https://github.com/coreos/coreos-assembler/pull/928 !

jlebon:
We can un-hold this one now, right?

cgwalters:
Yep, un-held.

jlebon (approved):
OK I think I understand this better now.

One bikeshed is around naming (should've brought this up in https://github.com/coreos/coreos-assembler/pull/928). Both the `live` and `fulliso` builds are "live", right? Feels like `fulliso` should reflect that too. WDYT about e.g. `fulliso-live`?

Code itself LGTM!

cgwalters:
Maybe `liveinst`?

jlebon:
> Maybe `liveinst`?

Hmm, or `livemedia`? To reflect the fact you need the actual media available.

cgwalters:
In that it doesn't do PXE?  Hrmm.  I think the `inst` is more descriptive in that it's telling you it's for installing, whereas `live` *can* be for that but is also intended for "diskless PXE" type scenarios.

One tangentially related thing here is I'm considering reworking things so that we just boot off of the metal image directly.  This would mean we'd need to decompress the image all at once (making booting a lot slower) but would save a lot of storage/network space.  And it'd more strongly orient the use case towards installing.

cgwalters:
I also want to investigate whether we can do compression at with something like [VDO](https://www.redhat.com/de/blog/look-vdo-new-linux-compression-layer) which would be pretty ideal for this use case.

jlebon:
> In that it doesn't do PXE? Hrmm. I think the `inst` is more descriptive in that it's telling you it's for installing, whereas `live` can be for that but is also intended for "diskless PXE" type scenarios.

Sure, WFM!

dustymabe:
I have to admit I missed the other two PRs regarding fulliso. I think it'd be worth discussing our strategy here regarding what ISOs we provide. It would be nice if we didn't provide multiple ISOs (ironically I've been doing work this week to drop the old installer ISOs so that we could just ship the live ISOs), but I realize there are different use cases that have different requirements. Maybe we can discuss soon.

cgwalters:
FWIW, I was intending fulliso for FCOSB (Fedora Silverblue-based-on-FCOS), not necessarily FCOS.  Having the build capability doesn't mean we need to ship it for FCOS.

dustymabe:
That helps. I'd still like to chat sometime about it just to bring my understanding up to par, if you don't mind. Note: I'm not blocking or intending to block this PR.

cgwalters:
OK I think we can do a rename/changes in a second pass.  I'd like to prove out the functionality first.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -13,7 +15,8 @@ add_requires() {
     ln -sf "../${name}" "${requires_dir}/${name}"
 }
 
-if [ -e /root.squashfs ]; then
+isoroot=$(getarg coreos.liveiso=)
```
SerialVelocity, line 18: @cgwalters I think this is now causing the generator to fail if this isn't specified:
```
/usr/lib/systemd/system-generators/live-generator failed with exit status 1
```

cgwalters: Thanks for the report!  https://github.com/coreos/fedora-coreos-config/pull/260

PR #235, opened 2019-11-20
Switch to coreos.liveiso= karg

cgwalters:
Requires https://github.com/coreos/coreos-assembler/pull/927

Prep for supporting a `fulliso` build that requires locating
the ISO.  Also drops some copy-paste between the different bootloaders.


cgwalters:
Dependency is merged, so this one should be ready now!

jlebon (approved):

PR #237, opened 2019-11-25
[testing] Revert recent promotion to testing branch

dustymabe:
We found an issue [1] with our cgroups v2 strategy [2]. We need to
revert the recent promotions (that move us to F31) and do a new release
of F30 content with a slight modification. See the proposed solution
in the issue comment.

- Revert "tree: promote changes from testing-devel at 20e12225836d8220608feaaf85504441fbc55cb6"
    - This reverts commit 0d8e1884a85c588a754d802744d18c0c254c424d.
- Revert "manifest.yaml: bump to Fedora 31"
    - This reverts commit c11e08fccd9819e17853d2f4455069cee19d1a4b.
- Revert "manifest.yaml: adapt for new path to fedora-coreos.yaml"
    - This reverts commit 01850ff707a6b1a081fd3421fac0234958a00e22.

[1] https://github.com/coreos/fedora-coreos-streams/issues/26#issuecomment-558361931
[2] https://github.com/coreos/fedora-coreos-tracker/issues/292

PR #238, opened 2019-11-26
[testing] Add coreos-keep-cgroup-v1.service

jlebon:
In f31, the default cgroup changed to v2. However, we've decided to stay
on v1 for the time being. Thus, we don't want older nodes upgrading to
f31 to be forced into v2.

Add a tiny service which just scans the BLS configs and injects the
`systemd.unified_cgroup_hierarchy` karg as needed.

For more information, see:
coreos/fedora-coreos-tracker#292
https://github.com/coreos/fedora-coreos-streams/issues/26#issuecomment-558361931

### DIFF
--- a/fedora-coreos-base.yaml
+++ b/fedora-coreos-base.yaml
@@ -146,6 +146,13 @@ packages:
   - whois-nls
   # Parsing/Interacting with JSON data
   - jq
+  # HACK for https://github.com/coreos/fedora-coreos-config/pull/238#issuecomment-558707064
+  - brotli-1.0.7-3.fc30.x86_64
+  - curl-7.65.3-4.fc30.x86_64
+  - libcurl-7.65.3-4.fc30.x86_64
+  - libmetalink-0.1.3-8.fc30.x86_64
+  - libssh-0.9.0-5.fc30.x86_64
+  - libssh-config-0.9.0-5.fc30.noarch
 
 # This thing is crying out to be pulled into systemd, but that hasn't happened
 # yet.  Also we may want to add to rpm-ostree something like arch negation;
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -13,3 +13,4 @@ enable afterburn-firstboot-checkin.service
 enable afterburn-sshkeys@.service
 # Update agent
 enable zincati.service
+enable coreos-keep-cgroup-v1.service
--- a/overlay.d/90cgroups/usr/lib/systemd/system/coreos-keep-cgroup-v1.service
+++ b/overlay.d/90cgroups/usr/lib/systemd/system/coreos-keep-cgroup-v1.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=CoreOS Keep CGroups v1
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/292
+RequiresMountsFor=/boot
+# Note we don't conditionalize on ConditionKernelCommandLine= here because the
+# kernel arg might've been added manually just for this boot; we still want to
+# change the BLS entries in that case.
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/coreos-keep-cgroup-v1
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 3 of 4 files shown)


jlebon:
(Note you'll want to look at the commit itself since the overall PR diff includes the revert from #237.)

dustymabe:
testing now

jlebon:
OK, so one issue here is:

```
Removed:
  brotli-1.0.7-3.fc30.x86_64
  curl-7.65.3-4.fc30.x86_64
  libcurl-7.65.3-4.fc30.x86_64
  libmetalink-0.1.3-8.fc30.x86_64
  libssh-0.9.0-5.fc30.x86_64
  libssh-config-0.9.0-5.fc30.noarch
Added:
  curl-minimal-7.66.0-1.fc31.x86_64
  libcurl-minimal-7.66.0-1.fc31.x86_64
```

I think the issue is similar to what I described in https://github.com/coreos/coreos-assembler/pull/553#issuecomment-501730140 (i.e. the lockfile not being canonical means libsolv is still free to pick newer packages with different names which meet all the same constraints), which would be fixed by https://github.com/coreos/rpm-ostree/pull/1858.

A hack for now though (since we're going to nuke all this right after) is to just add all those NEVRAs directly in the manifest.

dustymabe:
> OK, so one issue here is:
> 
> ```
> Removed:
>   brotli-1.0.7-3.fc30.x86_64
>   curl-7.65.3-4.fc30.x86_64
>   libcurl-7.65.3-4.fc30.x86_64
>   libmetalink-0.1.3-8.fc30.x86_64
>   libssh-0.9.0-5.fc30.x86_64
>   libssh-config-0.9.0-5.fc30.noarch
> Added:
>   curl-minimal-7.66.0-1.fc31.x86_64
>   libcurl-minimal-7.66.0-1.fc31.x86_64
> ```

I see this too. 

> 
> I think the issue is similar to what I described in [coreos/coreos-assembler#553 (comment)](https://github.com/coreos/coreos-assembler/pull/553#issuecomment-501730140) (i.e. the lockfile not being canonical means libsolv is still free to pick newer packages with different names which meet all the same constraints), which would be fixed by [coreos/rpm-ostree#1858](https://github.com/coreos/rpm-ostree/pull/1858).

So because `curl-minimal` provides `curl` but is a newer version, `curl-minimal-7.66.0-1.fc31.x86_64` chosen over `curl-7.66.0-1.fc31.x86_64`?

> 
> A hack for now though (since we're going to nuke all this right after) is to just add all those NEVRAs directly in the manifest.

+1

jlebon:
OK, I just tested an upgrade from `30.20191014.0` to this and then to whatever the latest f31 I had locally and it seems like it worked:

```
[root@coreos repo]# systemctl status coreos-keep-cgroup-v1.service
...
Nov 26 16:43:47 localhost systemd[1]: Starting CoreOS Keep CGroups v1...
Nov 26 16:43:47 localhost coreos-keep-cgroup-v1[744]: ostree-1-fedora-coreos.conf: injected systemd.unified_cgroup_hierarchy=0
Nov 26 16:43:47 localhost coreos-keep-cgroup-v1[744]: ostree-2-fedora-coreos.conf: injected systemd.unified_cgroup_hierarchy=0
Nov 26 16:43:47 localhost systemd[1]: Started CoreOS Keep CGroups v1.
...
[root@coreos repo]# rpm-ostree rebase $PWD:fedora/x86_64/coreos/testing-devel --experimental
...
[root@coreos repo]# reboot
...
[root@coreos ~]# grep -o systemd.unified_cgroup_hierarchy=0 /proc/cmdline
systemd.unified_cgroup_hierarchy=0
```


> So because `curl-minimal` provides curl but is a newer version, `curl-minimal-7.66.0-1.fc31.x86_64` chosen over `curl-7.66.0-1.fc31.x86_64`?

Right, exactly. (That said, we should probably evaluate whether we should switch to `[lib]curl-minimal` -- it's not clear offhand what features we lose).

dustymabe:
The idea here being that `coreos-keep-cgroup-v1.service` would get removed right after we do this next release and thus it will be removed when we update to F31 FCOS as well? i.e. we don't have to worry about this service running forever and possibly causing side effects down the road?

dustymabe:
in my testing this seems to work!

jlebon:
> The idea here being that `coreos-keep-cgroup-v1.service` would get removed right after we do this next release and thus it will be removed when we update to F31 FCOS as well?

Yeah, exactly. It'll only be in that one OSTree commit we build. In fact, would be nice if we could do like RHCOS and have the ability to ship *only* the update (I mean, we have all the pieces in place, we just need to develop the SOP for it).

jlebon:
OK, this is rebased now!

dustymabe (approved):
LGTM

PR #245, opened 2019-11-29
05core: relabel /var from initrd

jlebon:
Now that the kernel supports relabeling from the initrd, we can clean up
some hacks here. After running `systemd-tmpfiles`, immediately relabel
what we just wrote using `setfiles`. This allows us to drop the `Z /var`
tmpfiles.d dropin, and hacking around `systemd-random-seed.service`
ordering.

For compatibility with RHCOS, I added a fallback in `coreos-relabel` so
that if initrd relabeling isn't supported, we just use tmpfiles.d
dropins. (RHCOS today sources the `05core` overlay directly, and I think
doing it this way is cleaner than splitting things out into a separate
overlay).

I've also snuck in a fix for the live case with persistent `/var` so we
don't always relabel everything in there.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-relabel
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-relabel
@@ -1,0 +1,61 @@
+#!/bin/bash
+set -euo pipefail
+
+err() {
+    echo "$@" >&2
+}
+
+fatal() {
+    err "$@"
+    exit 1
+}
+
+if [ $# -eq 0 ]; then
+    err "Usage: $0 [PATTERN...]"
+    err " e.g.: $0 /etc/passwd '/etc/group*'"
+fi
+
+if [ ! -f /sysroot/etc/selinux/config ]; then
+    exit 0
+fi
+
+source /sysroot/etc/selinux/config
+
+if [ -z "${SELINUXTYPE:-}" ]; then
+    fatal "Couldn't find SELINUXTYPE in /sysroot/etc/selinux/config"
+fi
+
+file_contexts="/sysroot/etc/selinux/${SELINUXTYPE}/contexts/files/file_contexts"
+
+# feature detection until RHEL8 kernel has
+# https://lore.kernel.org/selinux/20190912133007.27545-1-jlebon@redhat.com/T/#m5f950ca9fc3ed374cb793fc9aed0435602b1b6ad
+can_setfiles() {
+    CAN_SETFILES_STAMP=/run/.coreos-relabel-can-setfiles
+    if [ ! -f ${CAN_SETFILES_STAMP} ]; then
+        touch /sysroot/etc/.setfiles-test
+        trap 'rm /sysroot/etc/.setfiles-test' EXIT
+        if setfiles -Fr /sysroot "$file_contexts" /sysroot/etc/.setfiles-test &>/dev/null; then
+            echo 1 > ${CAN_SETFILES_STAMP}
+        else
+            echo 0 > ${CAN_SETFILES_STAMP}
+        fi
+    fi
+    grep -q 1 ${CAN_SETFILES_STAMP}
+}
+
+if can_setfiles; then
+    prefixed_patterns=()
+    while [ $# -ne 0 ]; do
+        pattern=$1; shift
+        prefixed_patterns+=("/sysroot/$pattern")
+    done
+    setfiles -vFi0 -r /sysroot "$file_contexts" "${prefixed_patterns[@]}"
+else
+    while [ $# -ne 0 ]; do
+        pattern=$1; shift
+        # Really, we could handle /etc files here earlier by spitting out a
+        # separate unit service like ignition-relabel.service was. But let's not
+        # do that until the need arises.
+        echo "Z $pattern - - -" >> /run/tmpfiles.d/var-relabel.conf
+    done
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-populate-var.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-populate-var.sh
@@ -14,34 +14,28 @@ fi
 # Host and Silverblue:
 # https://github.com/rhinstaller/anaconda/blob/b9ea8ce4e68196b30a524c1cc5680dcdc4b89371/pyanaconda/payload/rpmostreepayload.py#L332
 
-# Simply manually mkdir /var/{lib,log}; the tmpfiles.d entries otherwise reference
-# users/groups which we don't have access to from here (though... we *could*
-# import them from the sysroot, and have nss-altfiles in the initrd, but meh...
-# let's just wait for systemd-sysusers which will make this way easier:
-# https://github.com/coreos/fedora-coreos-config/pull/56/files#r262592361).
-mkdir -p /sysroot/var/lib /sysroot/var/log
+for varsubdir in lib log home roothome opt srv usrlocal mnt media; do
 
-systemd-tmpfiles --create --boot --root=/sysroot \
-    --prefix=/var/home \
-    --prefix=/var/roothome \
-    --prefix=/var/opt \
-    --prefix=/var/srv \
-    --prefix=/var/usrlocal \
-    --prefix=/var/mnt \
-    --prefix=/var/media
+    # If the directory already existed, just ignore. This addresses the live
+    # image case with persistent `/var`; we don't want to relabel all the files
+    # there on each boot.
+    if [ -d "/sysroot/var/${varsubdir}" ]; then
+        continue
+    fi
 
-# Ask for /var to be relabeled.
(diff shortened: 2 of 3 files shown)


jlebon:
So with this + https://github.com/coreos/ignition/pull/846, try:

```
$ curl -LO https://gist.githubusercontent.com/jlebon/889d34b481a8c2abee618cf448a4c6c6/raw/e42dcfa7407d75704c29a5aec2f9d87757a10fdf/config-var-partitions.ign
$ cosa run -i config-var-partitions.ign --size 15
...
[core@coreos ~]$ findmnt
TARGET                           SOURCE     FSTYPE     OPTIONS
...
|-/var                           /dev/vda5  xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
| |-/var/home                    /dev/vda6  xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
| |-/var/log                     /dev/vda9  xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
| | `-/var/log/journal           /dev/vda10 xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
| |-/var/srv                     /dev/vda11 xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
| `-/var/lib                     /dev/vda7  xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
|   `-/var/lib/containers        /dev/vda8  xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
...
[core@coreos ~]$ journalctl | grep avc
[core@coreos ~]$ ls -laZ /var/home/core/hello.txt
-rw-r--r--. 1 root root unconfined_u:object_r:user_home_t:s0 11 Nov 29 17:12 /var/home/core/hello.txt
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-relabel
```diff
@@ -0,0 +1,57 @@
...
+    if [ ! -f ${CAN_SETFILES_STAMP} ]; then
+        touch /sysroot/etc/.setfiles-test
+        trap 'rm /sysroot/etc/.setfiles-test' EXIT
+        if setfiles -Fr /sysroot "$file_contexts" /sysroot/etc/.setfiles-test &>/dev/null; then
+            echo 1 > ${CAN_SETFILES_STAMP}
+        else
+            echo 0 > ${CAN_SETFILES_STAMP}
+        fi
+    fi
+    grep 1 ${CAN_SETFILES_STAMP}
```
cgwalters: `grep -q` to avoid unnecessary output.

cgwalters:
> Now that the kernel supports relabeling from the initrd

Is that in a Fedora kernel build yet?  It looks like it's not in an upstream release yet:

```
walters@toolbox ~/s/linux> git log -n 1 --format=fuller
commit 596cf45cbf6e4fa7bcb0df33e373a7d062b644b5 (HEAD -> master, origin/master, origin/HEAD)
Merge: c3bfc5dd73c6 937790699be9
Author:     Linus Torvalds <torvalds@linux-foundation.org>
AuthorDate: Sun Dec 1 20:36:41 2019 -0800
Commit:     Linus Torvalds <torvalds@linux-foundation.org>
CommitDate: Sun Dec 1 20:36:41 2019 -0800

    Merge branch 'akpm' (patches from Andrew)
    
    Merge updates from Andrew Morton:
     "Incoming:
    
       - a small number of updates to scripts/, ocfs2 and fs/buffer.c
    
       - most of MM
    
      I still have quite a lot of material (mostly not MM) staged after
      linux-next due to -next dependencies. I'll send those across next week
      as the preprequisites get merged up"
    
    * emailed patches from Andrew Morton <akpm@linux-foundation.org>: (135 commits)
      mm/page_io.c: annotate refault stalls from swap_readpage
      mm/Kconfig: fix trivial help text punctuation
      mm/Kconfig: fix indentation
      mm/memory_hotplug.c: remove __online_page_set_limits()
      mm: fix typos in comments when calling __SetPageUptodate()
      mm: fix struct member name in function comments
      mm/shmem.c: cast the type of unmap_start to u64
      mm: shmem: use proper gfp flags for shmem_writepage()
      mm/shmem.c: make array 'values' static const, makes object smaller
      userfaultfd: require CAP_SYS_PTRACE for UFFD_FEATURE_EVENT_FORK
      fs/userfaultfd.c: wp: clear VM_UFFD_MISSING or VM_UFFD_WP during userfaultfd_register()
      userfaultfd: wrap the common dst_vma check into an inlined function
      userfaultfd: remove unnecessary WARN_ON() in __mcopy_atomic_hugetlb()
      userfaultfd: use vma_pagesize for all huge page size calculation
      mm/madvise.c: use PAGE_ALIGN[ED] for range checking
      mm/madvise.c: replace with page_size() in madvise_inject_error()
      mm/mmap.c: make vma_merge() comment more easy to understand
      mm/hwpoison-inject: use DEFINE_DEBUGFS_ATTRIBUTE to define debugfs fops
      autonuma: reduce cache footprint when scanning page tables
      autonuma: fix watermark checking in migrate_balanced_pgdat()
      ...
walters@toolbox ~/s/linux> git tag --contains 3e3e24b42043eceb97ed834102c2d094dfd7aaa6
walters@toolbox ~/s/linux> 
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-relabel
```diff
@@ -0,0 +1,57 @@
...
+set -euo pipefail
+
+err() {
+    echo "$@" >&2
+}
+
+fatal() {
+    err "$@"
+    exit 1
+}
```
cgwalters, line 11: At some point we should get these basics into `dracut-lib.sh` or so.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-relabel
```diff
@@ -0,0 +1,57 @@
...
+
+if [ $# -eq 0 ]; then
+    err "Usage: $0 [PATTERN...]"
+    err " e.g.: $0 /etc/passwd '/etc/group*'"
+fi
+
+source /sysroot/etc/selinux/config
```
cgwalters, line 22: Maybe silently do nothing if that file doesn't exist, to be nicer to people who want to build custom systems without SELinux?  There's a similar pattern with other tools like `restorecon`.

cgwalters (approved):

cgwalters:
Reusing this for a tangentially related q; remind me again why we don't load policy from the initramfs?

The complexity of handling https://github.com/coreos/fedora-coreos-config/pull/184 is much higher than `/var` (this PR) - basically if we have policy loaded we can rely on "default" labels being created.

Although, maybe a crude hammer would be a patch for `setfiles` that only labeled *unlabeled* files - we can rely on ostree to set the contexts for objects.  It's just things like the `repo/config`, the object directories etc. that I don't quite want to patch ostree to explicitly set.

jlebon:
> Is that in a Fedora kernel build yet? It looks like it's not in an upstream release yet:

Yeah, it was backported in https://bugzilla.redhat.com/show_bug.cgi?id=1758597 --> https://src.fedoraproject.org/rpms/kernel/c/8bb21014a55f50b1a734322639fffadafcd79672?branch=f31

> Reusing this for a tangentially related q; remind me again why we don't load policy from the initramfs?

See discussions with Lennart in https://github.com/coreos/ignition/issues/635, starting from https://github.com/coreos/ignition/issues/635#issuecomment-502688675.

> Although, maybe a crude hammer would be a patch for setfiles that only labeled unlabeled files - we can rely on ostree to set the contexts for objects.

OK yeah, I think this may be a problem. The issue is that while we can *set* labels, we can't read them back. So restoring the rootfs using `mv` as is done in #184 will likely not copy back the labels correctly. I'm looking into loosening this too and making sure we can get #184 to work. To compare, this would work if it were a bare-user or archive repo (since OSTree wouldn't have to actually call `getxattr` for `security.selinux`). (I guess we could also get the labels from the policy in the OSTree commit, but... yuck, there's potential for corruption there.)

jlebon:
Fixups! :arrow_up:

cgwalters:
> To compare, this would work if it were a bare-user or archive repo (since OSTree wouldn't have to actually call getxattr for security.selinux).

But that would break the scenario where we're *not* replacing the rootfs.

jlebon:
> But that would break the scenario where we're not replacing the rootfs.

I meant if the source repo we're using to replace the rootfs wasn't bare (e.g. pulled over the network or something). Not that the embedded OSTree itself would be. :)  I was more expanding on why it's problematic rather than suggesting a change.

cgwalters:
> The issue is that while we can set labels, we can't read them back. So restoring the rootfs using mv as is done in #184 will likely not copy back the labels correctly. I'm looking into loosening this too and making sure we can get #184 to work.

Ah, yes.  So we need another kernel patch.

PR #246, opened 2019-12-02
overlay/growpart: Support ext4 too

cgwalters:
I'm using ext4 as a test case for rootfs reprovisioning, and
there's no reason not to add support for it anyways.

As the comments say though, ideally we improve the systemd
code here in the future and use it.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
@@ -11,9 +11,10 @@ set -euo pipefail
 path=$1
 shift
 
-# remove dupes; we'll still error if somehow there are multiple mounts from
-# different devices on there... which is then ambiguous what we should do anyway
-majmin=$(findmnt -nvr -o MAJ:MIN "$path" | sort | uniq)
+# The use of tail is to avoid errors from duplicate mounts;
+# this shouldn't happen for us but we're being conservative.
+src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
+majmin=$(findmnt -nvr -o MAJ:MIN "$path" | tail -n1)
 devpath=$(realpath "/sys/dev/block/$majmin")
 partition=$(cat "$devpath/partition")
 parent_path=$(dirname "$devpath")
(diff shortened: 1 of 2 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -13,6 +13,7 @@ shift
 
 # remove dupes; we'll still error if somehow there are multiple mounts from
 # different devices on there... which is then ambiguous what we should do anyway
+src=$(findmnt -nvr -o SOURCE "$path")
```
jlebon: Minor: let's move this on top of the comment above so it stays together with the corresponding line?

lucab: I guess this one as well may actually need the `uniq | sort` massaging, no?

jlebon: The same issue is definitely possible in theory:

```
[root@coreos ~]# findmnt /sysroot
TARGET   SOURCE    FSTYPE OPTIONS
/sysroot /dev/vda4 xfs    rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota
[root@coreos ~]# mount /dev/vda1 /sysroot
[root@coreos ~]# findmnt /sysroot
TARGET   SOURCE    FSTYPE OPTIONS
/sysroot /dev/vda4 xfs    rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota
/sysroot /dev/vda1 ext4   rw,relatime,seclabel
[root@coreos ~]# findmnt -nvr -o SOURCE /sysroot
/dev/vda4
/dev/vda1
```

Another approach is to just do `tail -n1` here and in the next line. I.e. let the semantics of this script be that `coreos-growpart /path` will target the filesystem/partition of the *topmost* mount point. No strong opinion either way. Though I guess given that this script is just for us right now, probably better to `uniq | sort` since we know it shouldn't happen?

cgwalters: Went with the consistent `tail -n1` path.

jlebon (approved):

PR #247, opened 2019-12-05
coreos-boot-mount-generator: key off of /etc/fstab

jlebon:
Just look at `/etc/fstab` for `/boot` mounts to determine whether to
generate them ourselves or let the `systemd-fstab-generator` do it.

See also: https://github.com/coreos/fedora-coreos-tracker/issues/322

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -11,11 +11,10 @@ add_wants() {
     ln -sf "../${name}" "${wants_dir}/${name}"
 }
 
-# If there's no aleph version from https://github.com/coreos/coreos-assembler/pull/768
-# then this is a non-FCOS system, likely RHCOS pre-4.3.  In
-# that case, we don't want to overwrite what the systemd-fstab-generator
-# will do.
-if ! [ -f /sysroot/.coreos-aleph-version.json ]; then
+# If there's already an /etc/fstab entries for /boot, then this is is a non-FCOS
+# system, likely RHCOS pre-4.3 (which still used Anaconda).  In that case, we
+# don't want to overwrite what the systemd-fstab-generator will do.
+if findmnt --fstab /boot &>/dev/null; then
     exit 0
 fi
 


jlebon:
(Still need to test this.)

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -11,11 +11,10 @@ add_wants() {
...
 
-# If there's no aleph version from https://github.com/coreos/coreos-assembler/pull/768
-# then this is a non-FCOS system, likely RHCOS pre-4.3.  In
-# that case, we don't want to overwrite what the systemd-fstab-generator
-# will do.
-if ! [ -f /sysroot/.coreos-aleph-version.json ]; then
+# If there's already an /etc/fstab entries for /boot, then this is is a non-FCOS
+# system, likely RHCOS pre-4.3 (which still used Anaconda).  In that case, we
+# don't want to overwrite what the systemd-fstab-generator will do.
+if grep '^#' -v /etc/fstab | cut -d' ' -f2 | grep -q /boot; then
```
cgwalters: `findmnt -s /boot`

jlebon: Ahh yup, that's much better.

cgwalters:
Hm, sorry about this, I certainly didn't even think at all about really old FCOS.

Change seems reasonable to me as a better "detect build from Anaconda", but I think we can do better, and it's the same solution as we've been using in RHCOS, which is to key off of the `root=` kernel argument.

jlebon:
> Change seems reasonable to me as a better "detect build from Anaconda", but I think we can do better, and it's the same solution as we've been using in RHCOS, which is to key off of the `root=` kernel argument.

Hmm, I think it's cleaner/more obviously correct to key off the thing we actually care about though. In this case, it's whether we need to emit boot mounts or whether `systemd-fstab-generator` will take care of it for us.

cgwalters:
Sure, that's fine by me too.

jlebon:
OK, tested working from a node that started at `30.20190905.0` (which didn't have `.coreos-aleph-version.json`). Upgraded to the barrier, then to this. `/boot` still being generated.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -11,11 +11,10 @@ add_wants() {
...
 
-# If there's no aleph version from https://github.com/coreos/coreos-assembler/pull/768
-# then this is a non-FCOS system, likely RHCOS pre-4.3.  In
-# that case, we don't want to overwrite what the systemd-fstab-generator
-# will do.
-if ! [ -f /sysroot/.coreos-aleph-version.json ]; then
+# If there's already an /etc/fstab entries for /boot, then this is is a non-FCOS
+# system, likely RHCOS pre-4.3 (which still used Anaconda).  In that case, we
+# don't want to overwrite what the systemd-fstab-generator will do.
+if findmnt --fstab /boot; then
```
cgwalters: `> /dev/null` to avoid spew (though I'm not sure systemd actually logs the output of generators).

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -11,11 +11,10 @@ add_wants() {
     ln -sf "../${name}" "${wants_dir}/${name}"
 }
 
-# If there's no aleph version from https://github.com/coreos/coreos-assembler/pull/768
-# then this is a non-FCOS system, likely RHCOS pre-4.3.  In
-# that case, we don't want to overwrite what the systemd-fstab-generator
-# will do.
-if ! [ -f /sysroot/.coreos-aleph-version.json ]; then
+# If there's already an /etc/fstab entries for /boot, then this is is a non-FCOS
+# system, likely RHCOS pre-4.3 (which still used Anaconda).  In that case, we
```
cgwalters, line 15: ...or a very old FCOS ?  Worth stating here since that's why we're changing it.

cgwalters (approved):
Two nits otherwise LGTM.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
jlebon: Fixed! (I don't think it does either, but seems more correct).

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
jlebon: Not exactly. Even the first "official" FCOS release was post-Anaconda. The change here is to cast a smaller net so we only affect RHCOS, and not old FCOS too.

PR #248, opened 2019-12-05
manifest.yaml: use regular Fedora 31 repos

jlebon:
Looks like the repo structure for f31 have finally stabilized. The
`-next` ones don't work anymore.


dustymabe:
should we also delete the `fedora-next.repo` file?

dustymabe (approved):

jlebon:
> should we also delete the `fedora-next.repo` file?

I think we may still want them in the tree so that other streams like `next` can make use of them.

cgwalters (approved):

PR #249, opened 2019-12-05
fedora-coreos-base: add btrfs-progs

jlebon:
Ignition used to pull it in, but now it's going to become a weak dep.
But we still want it in FCOS so users can have btrfs partitions if
they'd like.

More info in: https://github.com/coreos/fedora-coreos-tracker/issues/323


jlebon:
Note it's already in the lockfile since we're still pulling it in for now.

ajeddeloh (approved):

PR #254, opened 2019-12-13
fedora-coreos-base: add attr package

jlebon:
This is one of those utilities that I consider "core".

Brings no additional deps (we're already shipping libattr as a dep of
coreutils), and has a tiny footprint.

(Also the rpm-ostree testsuite depends on it).


lucab (approved):

lucab:
For reference this also provides `getfattr` and `setfattr` for working with extended attributes, I'm surprised we are not shipping it already.

PR #256, opened 2019-12-18
ci/validate: New file to check syntax

cgwalters:
I've had syntax errors in shell script in `overlay.d` too many
times, let's add a quick check for this.  This is a more lightweight
version of what we're doing in coreos-assembler.

### DIFF
--- a/ci/validate
+++ b/ci/validate
@@ -1,0 +1,39 @@
+#!/usr/bin/python3
+# Validate basic syntax of shell script and yaml.
+
+import os
+import stat
+import subprocess
+import yaml
+
+validated=0
+
+def validate_shell(rootfd, name):
+    subprocess.check_call(['bash', '-n', name], preexec_fn=lambda: os.fchdir(rootfd))
+    global validated
+    validated +=1
+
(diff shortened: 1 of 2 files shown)


cgwalters:
OK, this is passing CI now.

ci/validate
```diff
@@ -0,0 +1,38 @@
...
+
+for root, dirs, files, rootfd in os.fwalk('.'):
+    # Skip .git
+    if '.git' in dirs:
+        dirs.remove('.git')
+    for name in files:
+        if name.endswith(('.yaml', '.yml')):
+            with open(os.open(name, dir_fd=rootfd, flags=os.O_RDONLY)) as f:
+                yaml.safe_load(f)
+                validated +=1
```
jlebon, line 24: Minor: we could just `continue` here, right?

jlebon (approved):
One minor comment, LGTM as is.

jlebon:
```
[fedora-coreos-config] Running shell script
+ set -xeuo pipefail
+ ./ci/validate
Validated 26 files
```

Cool, thanks for adding this check!

PR #257, opened 2019-12-19
dracut: 40ignition-ostree: organize growpart deps

dustymabe:
- Make separate sections for growpart and coreos-growpart
- Break out to multiple lines and sort.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -24,8 +24,35 @@ install() {
         sort \
         uniq
 
+    # coreos-growpart deps
+    inst_multiple \
+        basename  \
+        blkid     \
+        cat       \
+        dirname   \
+        findmnt   \
+        growpart  \
+        realpath  \
+        resize2fs \
+        tail      \
+        touch     \
+        xfs_growfs
+
     # growpart deps
-    inst_multiple sfdisk awk realpath basename dirname sfdisk xfs_growfs resize2fs growpart touch
+    # Mostly generated from the following command:
+    # 	$ bash --rpm-requires /usr/bin/growpart | sort | uniq | grep executable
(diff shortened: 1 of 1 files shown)


dustymabe:
The reason I ended up here is because when I removed the `coreos-installer` and `coreos-installer-dracut` packages to test the new installer `coreos-growpart` started to fail because the dependency on `tail` wasn't explicitly called out.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
```diff
@@ -24,8 +24,35 @@ install() {
...
+        realpath  \
+        resize2fs \
+        tail      \
+        touch     \
+        xfs_growfs
+
     # growpart deps
-    inst_multiple sfdisk awk realpath basename dirname sfdisk xfs_growfs resize2fs growpart touch
+    # Mostly generated from the following command:
+    # 	$ bash --rpm-requires /usr/bin/growpart | sort | uniq | grep executable
```
jlebon, line 43: Woah neat, TIL about `bash --rpm-requires`!

Really weird though. It's not in upstream bash, seems to be a [Fedora patch](https://src.fedoraproject.org/rpms/bash/blob/master/f/bash-requires.patch), and the requires it emits aren't really useful to RPM:

```
$ cat zzz.sh
ls
$ bash --rpm-requires zzz.sh
executable(ls)
$ rpm -q --whatprovides 'executable(ls)'
no package provides executable(ls)
```

I guess whatever needed this patch would translate those executables to e.g. `/bin/ls`?

jlebon (approved):

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
dustymabe: yeah it's also not 100%. It misses things like `$(foo --bar)` and `bar | foo`, but it's better than manually reading a long script.

PR #259, opened 2020-01-03
WIP: NM in the initrd

cgwalters:
This seems not to run for me...I don't quite understand the
initqueue stuff, still debugging.

See also https://github.com/coreos/fedora-coreos-config/pull/200#issuecomment-544592432

Blocked on: https://github.com/dracutdevs/dracut/issues/694

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -18,6 +18,9 @@ initramfs-args:
   - --omit=lvm
   - --omit=multipath
   - --omit=iscsi
+  # Networking is NetworkManager
+  - --omit=systemd-networkd network-legacy
+  - --add=network-manager
 
 # Be minimal
 recommends: false
@@ -55,9 +58,6 @@ remove-from-packages:
               /usr/lib/systemd/system/systemd-networkd-wait-online.service]
   - [systemd-container, /usr/lib/systemd/network/.*]
   - [systemd-udev, /usr/lib/systemd/network/.*]
-  # We're still using the legacy dracut network module for now. Dracut keys off
-  # of this: https://github.com/dracutdevs/dracut/blob/1fcc70fe57eea0ea658aa2de5c0044683fe85cf1/modules.d/40network/module-setup.sh#L11
-  - [NetworkManager, /usr/libexec/nm-initrd-generator]
 
 
 remove-files:


cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1787620

cgwalters:
I've verified that we have the module code in the initrd, but it looks like the initqueue never executes `/usr/lib/dracut/hooks/initqueue/settled/99-nm-run.sh` (I added an `echo` statement to that code).

thom311:
hm. Is there no further information here? Any logs?

What was the kernel command line?

If you don't even get 99-nm-run.sh to execute, then it's a dracut issue (or an issue in NM's dracut module).

dustymabe:
since I'm having networking in the initrd troubles myself, I can try testing this out to see if I get any different results.

cgwalters:
OK so...this is a bit obvious now, but entering `emergency.target` will kill other services - specifically it will kill `dracut-initqueue.service`.  So to better replicate what's going on in the normal case, be sure to  at least `systemctl start dracut-initqueue`.

cgwalters:
> hm. Is there no further information here? Any logs?
> What was the kernel command line?

`rd.neednet ip=dhcp` are the relevant bits.

I can paste the full journal if you want (it's also easy to do a FCOS build from this patch) but AFAICS there's nothing there of use because:

> If you don't even get 99-nm-run.sh to execute, then it's a dracut issue (or an issue in NM's dracut module).

Right.  That's as far as I got - it's not clear to me why that hook isn't being executed.  I don't understand the initqueue really to be honest.

cgwalters:
xref https://github.com/dracutdevs/dracut/pull/702

cgwalters:
As best I can tell what's happening is `/usr/bin/dracut-initqueue` is calling `check_finished` which looks for files in `$hookdir/initqueue/finished/` - and since nothing is there, we exit before trying any of the other hooks.

It seems like other dracut initrd code is dropping files there, but the NM initrd code isn't.

Was this NM initrd code ever tested?  If so, where?

ryncsn:
Hi, for the dracut problem I have filed an issue that is related to your problem sometime ago, maybe we can discuss there:
See https://github.com/dracutdevs/dracut/issues/694

cgwalters:
@ryncsn Thank you, that looks exactly like the bug!

jlebon:
Blocked on: https://github.com/dracutdevs/dracut/issues/694.

LorbusChris:
Hopefully this'll be unblocked with the next dracut build :)

dustymabe:
Now that https://github.com/dracutdevs/dracut/issues/694 is fixed I'm picking this up and running with it. I'm going to close this out and open a new PR with more fixes in it.

dustymabe:
If you'd like to track progress on this you can follow https://github.com/coreos/fedora-coreos-tracker/issues/394

PR #260, opened 2020-01-04
overlay/live-generator: Don't error out if not on live ISO

cgwalters:
Fixes an error in the journal.
(We should have kola check for this)

Reported-by: @SerialVelocity


overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -15,7 +15,7 @@ add_requires() {
     ln -sf "../${name}" "${requires_dir}/${name}"
 }
 
-isoroot=$(getarg coreos.liveiso=)
+isoroot=$(getarg coreos.liveiso= 2>/dev/null || true)
```
SerialVelocity: Is it worth moving this into the `else` clause instead of discarding the error? The generator should probably fail if `coreos.liveiso` isn't set and you are trying to boot an iso, right?

cgwalters: Good point.  I went with a bigger cleanup, moving up the check and exiting early, allowing us to de-indent everything else :arrow_down:

SerialVelocity: I mean, even with the bigger cleanup, this line still has to be in the `else` clause, right? It should only be specified when booting an ISO, not when booting from PXE (which is why I hit this bug. It broke pxe booting. Not urgent since coreos/fedora-coreos-tracker#339 has broken the latest builds too)

cgwalters: There isn't a need for an `else` clause - we exit the whole generator early on if we're not in a live image.  Right?

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -15,17 +15,20 @@ add_requires() {
     ln -sf "../${name}" "${requires_dir}/${name}"
 }
 
+if ! is-live-image; then
+    exit 0
+fi
+
 isoroot=$(getarg coreos.liveiso=)
```
jlebon: Hmm, could we sanity check here that `coreos.liveiso` is not empty? Otherwise if we're somehow in that state, we'll be emitting a malformed mount unit which could be harder to debug.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
cgwalters: I'm sorry, now that I re-read this I understand now - you were talking about the *inner* condition and it indeed makes total sense to move the `getarg` there.  Thanks so much for your review!

jlebon (approved):

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -57,10 +59,16 @@ What=/root.squashfs
 Where=/sysroot
 Type=squashfs
 EOF
-    else
-        # And in this case, it's on the ISO
-        mkdir -p /run/media/iso
-        cat >"${UNIT_DIR}/run-media-iso.mount" <<EOF
+else
+    isoroot=$(getarg coreos.liveiso=)
```
SerialVelocity, line 63: Doesn't really matter that much but a small nit is this should probably be `|| true` just so you get the `Missing required coreos.liveiso kernel argument` message if the parameter doesn't exist. Currently it only covers the "empty" case (having `<other-arg> coreos.liveiso= <other-arg>`)

PR #261, opened 2020-01-09
switch to the rust based coreos-installer

dustymabe:
- manifest: start including coreos-installer-systemd package
- overrides: include the new rust coreos-installer rpm in the repos
- removed the installer directory
    - It was used for creating the old installer ISOs
    - We are now using the Live ISO based approach


manifest-lock.overrides.x86_64.yaml
```diff
@@ -1,7 +1,7 @@
 packages:
-  # New 2.1.1 release of ignition + a backport for:
-  # https://github.com/coreos/ignition/pull/907
-  #
-  # https://bodhi.fedoraproject.org/updates/FEDORA-2019-4bdb83f0af
-  ignition:
-    evra: 2.1.1-3.git40c0b57.fc31.x86_64
+  # New rust based coreos-installer
+  # https://bodhi.fedoraproject.org/updates/FEDORA-2020-xxxx
```
dustymabe: will update this link and then mark ready for review

manifest-lock.overrides.x86_64.yaml
```diff
@@ -1,7 +1,7 @@
 packages:
-  # New 2.1.1 release of ignition + a backport for:
-  # https://github.com/coreos/ignition/pull/907
-  #
-  # https://bodhi.fedoraproject.org/updates/FEDORA-2019-4bdb83f0af
-  ignition:
```
lucab, line 6: Is this `ignition` change expected?

dustymabe: Yeah, see the first commit in the patchset. I probably should have mentioned it in the PR description.

lucab: Ah, my bad, I missed it. :+1:

jlebon (approved):
>```
> - removed the installer directory
>    - It was used for creating the old installer ISOs
>    - We are now using the Live ISO based approach
>```

Cool, this works. I think I would have preferred it as a separate patch after we have the live ISO with the new coreos-installer out and https://github.com/coreos/fedora-coreos-pipeline/pull/182 merged, but this is fine too!

(We should probably also edit `buildextend-installer` to exit with a clear error message if the directory is missing.)

jlebon:
https://github.com/coreos/coreos-assembler/pull/1034

dustymabe:
> coreos/coreos-assembler#1034

thanks @jlebon

PR #264, opened 2020-01-10
Make iptables-nft the default iptables implementation

LorbusChris:
by adding a COSA postprocess script to raise its priority
above (from 5 to 15) iptables-legacy's priority (10)
with `update-alternatives`.

This workaround will be dropped once iptables-nft becomes
the default implementation in F32:
https://fedoraproject.org/wiki/Changes/iptables-nft-default

Tracker Issue:
https://github.com/coreos/fedora-coreos-tracker/issues/342

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -96,6 +96,23 @@ postprocess:
     setsebool -P -N container_use_cephfs on  # RHBZ#1692369
     setsebool -P -N virt_use_samba on  # RHBZ#1754825
 
+  # Make iptables-nft the default iptables implementation by raising its priority
+  # above (from 5 to 15) iptables-legacy's priority (10) with `update-alternatives`.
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Assuming you tested this locally, LGTM.

cgwalters:
There's some lingering "alternatives vs rpm-ostree" issues; see https://bugzilla.redhat.com/show_bug.cgi?id=1657367 at least.  But, hopefully this still works.

dustymabe:
hmm. we've had conversations about this in the past. /hold - I need to dig

LorbusChris:
I had only tested the behaviour on standard Fedora - I've now tried it on a true FCOS and it does not work due to the BZ mentioned by Colin (i.e. `alternatives` saving state in `/var/`). Closing for now.

LorbusChris:
Let's move further discussion over to https://github.com/coreos/fedora-coreos-tracker/issues/342

PR #266, opened 2020-01-13
overrides: pull in crypto-policies without python dependencies

### LINKED ISSUE #280: python3 getting pulled in by crypto-policies
In rawhide right now we grew a dependency on python again:

```
[root@coreos ~]# rpm -e python3
error: Failed dependencies:
        /usr/bin/python3 is needed by (installed) crypto-policies-20190816-3.gitbb9bf99.fc32.noarch
```

Looks like we need to dig into why `crypto-policies` now requires python3.

dustymabe:
We're pulling from rawhide here as this is a brand new change
and the maintainer is not comfortable sending it to F31 yet.

See https://src.fedoraproject.org/rpms/crypto-policies/pull-request/6#comment-35958

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/280

### DIFF
--- a/manifest-lock.overrides.x86_64.yaml
+++ b/manifest-lock.overrides.x86_64.yaml
@@ -5,3 +5,10 @@ packages:
     evra: 0.1.2-1.fc31.x86_64
   coreos-installer-systemd:
     evra: 0.1.2-1.fc31.x86_64
+  # crypto-policies without python dependencies
+  # We're pulling from rawhide here as this is a brand new change
+  # and the maintainer is not comfortable sending it to F31 yet.
+  # https://src.fedoraproject.org/rpms/crypto-policies/pull-request/6#comment-35958
+  crypto-policies:
+    evra: 20191128-3.gitcd267a5.fc32.noarch
+


dustymabe:
Let's see if CI passes :)

jlebon:
Hmm, why can't we do an f31 build and just pull it in FCOS but not submit it to Bodhi?

dustymabe:
> Hmm, why can't we do an f31 build and just pull it in FCOS but not submit it to Bodhi?

I think the maintainer was opposed to merging it to the f31 branch at all. There aren't any changes between the f31 branch and the f32 branch other than the commits for splitting out the scripts so i thought this would be safe for now.

Any other concerns?

jlebon:
OK, gotcha. I was going to suggest exercising the FCOS dist-git branch workflow here, though I think it should be fine in this case since we're not pulling in other changes.

I double checked the provides/requires to make sure it doesn't interfere with the rest of the package set, and it shouldn't (that was my biggest hesitation to just pulling in something from rawhide). Let's make sure we sanity check the diff from CI too before merging!

dustymabe:
CI passed. I don't see a diff in the CI output but here is what it shows when I run it locally:

```
ostree diff commit from: 3d42f4f9efb14b82ffda4f6daa53701eee7b6b06d1f9897a06e306b02dc595d3
ostree diff commit to:   522ae66271aca9a12be0b9afb4817fc774520cb031de559e287284ab7f249e1e
Upgraded:                                                                                                             
  crypto-policies 20191128-2.gitcd267a5.fc31 -> 20191128-3.gitcd267a5.fc32
Removed:                                                  
  gdbm-libs-1:1.18.1-1.fc31.x86_64                         
  python-pip-wheel-19.1.1-7.fc31.noarch       
  python-setuptools-wheel-41.2.0-1.fc31.noarch                                                                        
  python3-3.7.5-2.fc31.x86_64                                                                                         
  python3-libs-3.7.5-2.fc31.x86_64
```

jlebon (approved):

dustymabe:
*crosses fingers* and hopes that the koji tagger works with rawhide packages (I think it will)

jlebon:
> I don't see a diff in the CI output

Ahh right, I gotta polish up https://github.com/coreos/fedora-coreos-config/pull/176.

dustymabe:
> _crosses fingers_ and hopes that the koji tagger works with rawhide packages (I think it will)

yeah looks like it didn't work. I'll look at it in a bit when I get back to keyboard

dustymabe:
> > _crosses fingers_ and hopes that the koji tagger works with rawhide packages (I think it will)
> 
> yeah looks like it didn't work. I'll look at it in a bit when I get back to keyboard

https://pagure.io/releng/issue/9162

PR #269, opened 2020-01-15
kola-blacklist: drop fcos.python

jlebon:
This is fixed now:
https://github.com/coreos/fedora-coreos-tracker/issues/280


dustymabe (approved):
LGTM

PR #271, opened 2020-01-26
overlay: Add an ignition-ostree-generator to fix install path

cgwalters:
The current documentation for the FCOS PXE install path:
https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/#_live_pxe
Uses the `coreos.inst` kernel arguments instead of providing
an Ignition config.  In this case, we fail because
the `mount-subsequent.service` kicks in, and will fail to find
a root device.  (Or worse, may mistakenly find an existing disk
and mount it)

The bug here is a now familiar one - the
`ignition-ostree-mount-subsequent-sysroot.service` already has
`ConditionFileExists=!/run/ostree-live` which is correct for the
Live PXE case, but it also has a hard `Requires`, and as we know
systemd still pulls in requirements for units that aren't enabled.

Fix this by using a generator to enable the unit rather than doing
so as part of ignition-subsequent.

Note if Ignition is provided, then `ignition.firstboot` should be
present which would mean this bug isn't triggered.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-generator
@@ -1,0 +1,16 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+# This generator handles the non-live case; see also live-generator
+
+set -euo pipefail
+
+UNIT_DIR="${1:-/tmp}"
+
+if is-live-image; then
+    exit 0
+fi
+
+requiresdir="${UNIT_DIR}/ignition-subsequent.target.requires"
+mkdir -p "${requiresdir}"
+ln -sr /usr/lib/systemd/system/ignition-ostree-mount-subsequent-sysroot.service ${requiresdir}/
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-subsequent-sysroot.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-subsequent-sysroot.service
@@ -1,3 +1,4 @@
+# Note this unit is conditionally enabled by ignition-ostree-generator
 [Unit]
 Description=CoreOS: Mount (subsequent) /sysroot
 # These dependencies should match the "other" in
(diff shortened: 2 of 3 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-generator
```diff
@@ -0,0 +1,16 @@
...
+
+set -euo pipefail
+
+UNIT_DIR="${1:-/tmp}"
+
+if is-live-image; then
+    exit 0
+fi
+
+wantsdir="${UNIT_DIR}/ignition-subsequent.target.requires"
```
jlebon: Minor: `requiresdir`?

jlebon (approved):
Just one comment, LGTM otherwise!

jlebon:
One thing to note: ignition-dracut already has this concept too for the first boot via the `ignition-diskful.target`: https://github.com/coreos/ignition-dracut/blob/f58f29c6d2fa710371e1b556b906f4c9ca24d886/dracut/30ignition/ignition-generator#L42-L45

So an alternative approach here is adding a `ignition-diskful-subsequent.target` which gets activated under the same conditions. Then we could clean things up here so that `ignition-ostree-mount-firstboot-sysroot.service` links into `ignition-diskful.target` (actually we could do that now too) and `ignition-ostree-mount-subsequent-sysroot.service` to `ignition-diskful-subsequent.target`. An advantage of this is keeping the logic around Ignition/mounting/livefs in a single generator.

cgwalters:
> So an alternative approach here is adding a ignition-diskful-subsequent.target 

Yeah, makes sense - though I tested this, mind getting this in first and we can do a followup cleanup?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-generator
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+# This generator handles the non-live case; see also live-generator
+
+set -euo pipefail
+
+UNIT_DIR="${1:-/tmp}"
+
+if is-live-image; then
```
lucab, line 10: I feel like I'm missing something basic here: what is `is-live-image`?

cgwalters: See d66ed847c4f2eeeefc70727048b5d59622aff18b - it's an "API" between this repo and ignition-dracut.

cgwalters: The idea is e.g. whatever that opensuse thing is that uses ignition+dracut but not ostree/coreos-assembler could have their own `is-live-image` implementation.

lucab: Ack, thanks. I saw this when doing `testing` promotion, and was unsure whether it was a dangling thing. I missed the `/usr/bin/is-live-image` in initramfs.

PR #272, opened 2020-01-27
overlay/ignition-ostree: Use ignition-diskful.target

cgwalters:
For the firstboot mount; this way if we ever decided to add
non-default dependencies to this unit they wouldn't be pulled
into the transaction.

Reference: https://github.com/coreos/fedora-coreos-config/pull/271
Suggested-by: @jlebon


jlebon (approved):
LGTM!

PR #275, opened 2020-01-30
overlay/ignition-ostree: Hook up to ignition-diskful-subsequent

cgwalters:
Per review, rather than having a custom generator.  The code becomes
simpler.  This makes use of https://github.com/coreos/ignition-dracut/pull/151


cgwalters:
Depends on https://github.com/coreos/ignition-dracut/pull/151

jlebon:
https://src.fedoraproject.org/rpms/ignition/pull-request/50

dustymabe:
this should be in now with #278

dustymabe:
tests failed. maybe rebase this PR and let's run them again?

dustymabe (approved):
LGTM

cgwalters:
Hm, we could also remove live-generator I think and ship static units keyed off `diskful` and `diskful-subsequent` possibly too.

cgwalters:
But eh, might as well be incremental.

jlebon:
> Hm, we could also remove live-generator I think and ship static units keyed off `diskful` and `diskful-subsequent` possibly too.

Hmm, though those are meant to run when we are live, so it'd require a new e.g. `ignition-diskless.target` and `ignition-diskless-subsequent.target`. Though those targets would be misleading I think in a live PXE context. And anyway, the logic there is the same regardless of whether Ignition runs.

PR #279, opened 2020-02-05
motd: link to FCOS Discourse forum

jlebon:
Add a link to the Discourse forum directly in the MOTD to make it easier
for new users to figure out where to discuss and ask for help.

### DIFF
--- a/overlay.d/15fcos/usr/lib/motd.d/tracker.motd
+++ b/overlay.d/15fcos/usr/lib/motd.d/tracker.motd
@@ -1,2 +1,3 @@
 Tracker: https://github.com/coreos/fedora-coreos-tracker
+Discuss: https://discussion.fedoraproject.org/c/server/coreos/
 
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@ Base manifest configuration for
 Use https://github.com/coreos/coreos-assembler to build it.
 
 Discussions in
-https://discussion.fedoraproject.org/c/server/coreos and
+https://discussion.fedoraproject.org/c/server/coreos. Bug
+tracking and feature requests at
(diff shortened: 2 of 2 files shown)


jlebon:
Related: https://github.com/coreos/fedora-coreos-docs/pull/41

overlay.d/15fcos/usr/lib/motd.d/tracker.motd
```diff
@@ -1,2 +1,3 @@
 Tracker: https://github.com/coreos/fedora-coreos-tracker
+Discuss: https://discussion.fedoraproject.org/c/server/coreos/
```
dustymabe, line 2: Do the words `Tracker` and `Discuss` give the user enough info to make a decision? before we only had one option so `Tracker` didn't really matter. but now that we have two should we clarify more? I'm torn

dustymabe (approved):
LGTM - one comment

overlay.d/15fcos/usr/lib/motd.d/tracker.motd
cgwalters: Let's `Discuss`!

cgwalters (approved):

overlay.d/15fcos/usr/lib/motd.d/tracker.motd
jlebon: I guess one other option is `Questions`? But then the colons won't line up! :( And it doesn't convey the fact you could just chat there about how awesome FCOS is without actually having a question. :)

Not sure, inclined to just ship as is.

dustymabe: 👍

PR #280, opened 2020-02-07
manifest: Add excludes for python,perl,nodejs,dnf,grubby

cgwalters:
This makes use of the new
https://github.com/coreos/rpm-ostree/pull/1980

And we can then drop the test in kola that boots a whole VM
for this.


jlebon (approved):
Nice!

PR #281, opened 2020-02-07
Live ISO: extend boot prompt timeout to 5 seconds

dustymabe:
At 1 second it's almost impossible to catch the boot prompt if you need
to change the kernel command line parameters. Let's extend it to 5 seconds
so users have a fighting chance to catch the prompt.

### DIFF
--- a/live/EFI/fedora/grub.cfg
+++ b/live/EFI/fedora/grub.cfg
@@ -23,7 +23,7 @@ insmod gzio
 insmod part_gpt
 insmod ext2
 
-set timeout=1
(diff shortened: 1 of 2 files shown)


jamescassell (approved):
Will definitely make users' lives easier.

bgilbert (approved):
I originally set it at one second under the theory that users shouldn't need to mess with kargs.  But with the static IP stuff and the possibility of `coreos.inst` usage, I think that was probably optimistic.

cgwalters (approved):

PR #284, opened 2020-02-17
manifest: add lsof utility

dustymabe:
This PR adds two utilities that I've needed over the past few weeks: `time` and `lsof`. They each are self contained and don't introduce any new dependencies.

- `time`: I needed this when debugging a problem where [a user was reporting operations taking long periods of time](https://discussion.fedoraproject.org/t/cannot-docker-login-to-dockerhub-although-can-pull-public-images/17015/2). I had to work around this by wrapping the command with two `date` calls.

- `lsof`: I've been having some trouble with podman volumes lately and when trying to debug this I thought maybe some files had been deleted on the host but were still being held open by other programs. I've [used lsof to debug this situation in the past](https://dustymabe.com/2012/01/22/recover-space-by-finding-deleted-files-that-are-still-held-open./)

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -141,6 +141,7 @@ packages:
   - bash-completion
   - openssl
   - vim-minimal
+  - lsof
   # file-transfer: note fuse-sshfs is not in RHEL
   # so we can't put it in file-transfer.yaml
   - fuse-sshfs


lucab:
From an out-of-band chat: bash has a built-in for `time`.

```
$ help time
time: time [-p] pipeline
    Report time consumed by pipeline's execution.
    
    Execute PIPELINE and print a summary of the real time, user CPU time,
    and system CPU time spent executing PIPELINE when it terminates.
    
    Options:
      -p        print the timing summary in the portable Posix format
    
    The value of the TIMEFORMAT variable is used as the output format.
    
    Exit Status:
    The return status is the return status of PIPELINE.
```

lucab:
For the other half of the PR: `lsof` is something I would expect to have for emergency debugging, and I haven't seen many creative scripting abuses of that, :+1: from my side.

dustymabe:
> From an out-of-band chat: bash has a built-in for `time`.

+1 - I didn't try it first, just ran `which time` and didn't notice it existing. Dropped the first commit and updated the PR title.

lucab:
LGTM, but leaving some time in case others have comments.

jlebon (approved):
Yup, makes sense to me!

PR #288, opened 2020-02-25
Enable systemd-resolved and use as default resolver

darkmuggle:
When booting with DHCP6 and DHCP4, the most recent DHCP response will
overwrite /etc/resolv.conf. By using systemd-resolvd, we get ability to
have scoped DNS resolvers.

To keep the NetworkManager configurations in the same place, the
overlay.d directory has been renamed.

In the event a user does not want to use systemd-resolved, they can
easily disable it by:
- replacing /etc/resolv.conf with their own
- populate a new /etc/NetworkManager/conf.d/99-user with:
   [main]
   dns=default
   systemd-resolved=false


darkmuggle:
This is a WIP idea, that I wanted to put out there for discussion.  This allows the system to use scoped resolvers. I'm not sure if this is a great idea -- or ifs been discussed -- but I gave it a shot while digging into the way that FCOS's networking works.

overlay.d/14NetworkManager/usr/lib/NetworkManager/conf.d/11-enabled-resolved.conf
```diff
@@ -0,0 +1,10 @@
+# Instruct Network Manager to update and use systemd-resolved.
+# These settings are needed to set the hostname from DNS.
+[main]
+
+# Send the connection DNS configuration to systemd-resolved.
+systemd-resolvd=true
```
lucab: Typo: `systemd-resolved=true`

(according to https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html)

overlay.d/14NetworkManager/usr/lib/NetworkManager/conf.d/11-enabled-resolved.conf
```diff
@@ -0,0 +1,10 @@
...
+
+# Use systemd-resolved for name resolution.
+dns=systemd-resolved
```
lucab, line 9: According to your commit message, it sounds like you'd want the auto-detect behavior of leaving `dns=` unspecified.

lucab:
@darkmuggle not against this, but I fear we may be overly optimistic.

I think the bug you are implicitly referencing is https://github.com/coreos/fedora-coreos-tracker/issues/393. That one is about initramfs networking, where we (currently) don't run NM.

So I guess that:
 * this will only help after we do https://github.com/coreos/fedora-coreos-tracker/issues/394
 * we'd need to do wire the same setup in initramfs

jlebon:
Yeah, I think the first step is to switch to NM in the initrd and see where we land. That specific issue may melt away entirely.

darkmuggle:
> Yeah, I think the first step is to switch to NM in the initrd and see where we land. That specific issue may melt away entirely.

That would be ideal. I threw this out for consideration and debate. I anticipate leaving this as a draft, until after NM in the initrd.

lucab:
As a data-point, Container Linux has been using systemd-resolved for a long time. I don't remember huge issues/regression or user friction with it, but there have been some CVEs in the past. However, it was used in its natural ecosystem (networkd+timesyncd+resolved), not on the side of NM.

dustymabe:
Discussed this with a member of the NM team. They pointed out that systemd-resolved is the best way to do DNS configuration and it is recommended to use it with NM. It also may become the default in the rest of Fedora ([context](https://github.com/coreos/fedora-coreos-config/pull/310)).

gtherond:
Hi everyone, as suggested by @dustymabe on IRC, I would add my use-case to this draft.
I would like to get DNSOverTLS at system scope with FCOS, however, as for now, it seems that it is not possible as it doesn't use systemd-resolved anymore.

systemd-resolved partially support DoT since v239 and fully support it since v243.

darkmuggle:
Closing this a stale. The idea may be desirable for our users, but as @lucab pointed out the motivation on this is moot. If the group wants this to be continued, I can re-open after its been tested again.

jlebon:
FWIW: https://fedoraproject.org/wiki/Changes/systemd-resolved

So any subtle compatibility bug between systemd-resolved and NM will likely be shaken out if this goes forward.

dustymabe:
@jlebon do you think it's worth implementing for `next` now?

travier:
I think there are two points here:
  - Enabling systemd-resolved in the final root: I think we should try this in `next` as this is coming in Fedora 33 and that would bring all the benefits mentioned in https://fedoraproject.org/wiki/Changes/systemd-resolved.
  - Enabling systemd-resolved in the initramfs to keep networking behavior consistent between the two. This might require more work and I don't know if it's part of the F33 changes.

gtherond:
I would still push for this one on my own as I'm still blocked with the DTLS implementation for now and as it's a worthy security measure.

travier:
See ongoing discussion in https://github.com/coreos/fedora-coreos-config/pull/575

PR #294, opened 2020-03-11
s390x: put coreos-installer into initramfs

nikita-dubrovskii:

### DIFF
--- a/overlay.d/README.md
+++ b/overlay.d/README.md
@@ -27,6 +27,19 @@ Disables the Red Hat Linux legacy `ifcfg` format.
 
 Things that are more closely "Fedora CoreOS"; branding, services.
 
+20s390x
+-------
+
+For CoreOS installation on IBM Z we need to have coreos-installer within initramfs. 
+Usually for Linux installation on zVM we fetch kernel, ramdisk and parameters file
+into 'reader queue' and boot kernel like this:
+
+`#cp ipl c`
+
+During `cosa buildextend-live` ramdisk gets generated, but contains no coreos-installer,
+so DASD disk couldn't be formatted and partitioned, and consequently used. 
+That's why we put coreos-installer into initramfs and perform CoreOS installation. 
+
 80experimental
 --------------
 
--- a/overlay.d/20s390x/usr/lib/dracut/modules.d/30coreos-installer/coreos-installer.service
+++ b/overlay.d/20s390x/usr/lib/dracut/modules.d/30coreos-installer/coreos-installer.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=CoreOS Installer
+Before=coreos-installer.target
+After=network-online.target
+Wants=network-online.target
+ConditionKernelCommandLine=coreos.inst.install_dev
+
+Requires=basic.target dracut-cmdline.service dracut-pre-trigger.service dracut-initqueue.service
+After=basic.target dracut-cmdline.service dracut-pre-trigger.service dracut-initqueue.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-installer-service
+StandardOutput=kmsg+console
+StandardError=kmsg+console
+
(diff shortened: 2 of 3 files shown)


nikita-dubrovskii:
This PR depends on: https://github.com/coreos/coreos-installer/pull/153
Both of them allows CoreOS to be installed and used on IBM zVM.

overlay.d/README.md
```diff
@@ -27,6 +27,19 @@ Disables the Red Hat Linux legacy `ifcfg` format.
...
+20s390x
+-------
+
+For CoreOS installation on IBM Z we need to have coreos-installer within initramfs. 
+Usually for Linux installation on zVM we fetch kernel, ramdisk and parameters file
+into 'reader queue' and boot kernel like this:
+
+`#cp ipl c`
+
+During `cosa buildextend-live` ramdisk gets generated, but contains no coreos-installer,
```
jlebon, line 39: Hmm, can you expand on this? The ramdisk contains the root squashfs into which we pivot, right? That rootfs is a regular FCOS system which should have coreos-installer in it.

nikita-dubrovskii: After checking (unpacking) `fedora-coreos-*.dev.*-live-initramfs.s390x.img` i figured out, that it doesn't contain coreos-installer service. Maybe on x86 it's there, haven't checked yet. It means, that on IBM Z we are not able to proceed with CoreOS without installation during 1st boot.

jlebon: > After checking (unpacking) `fedora-coreos-*.dev.*-live-initramfs.s390x.img` i figured out, that it doesn't contain coreos-installer service.

The initramfs is a concatenation of multiple CPIO archives. One of those contains the root squashfs.

Are you able to boot the live image on s390x? Does it not boot all the way to the real root?

dustymabe: After unpacking the initramfs you need to look for a file at `/root.squashfs`. That squashfs has the real root filesystem in it that contains `coreos-installer`. As @jlebon said, the suggestion here is to boot a live cd all the way up and then you'll find `coreos-installer` easily.

jlebon: One thing to watch for if you're doing `cpio -id` is that it doesn't understand concatenated CPIO archives. That's why dracut has `skipcpio` to skip over the first microcode one. But even then, that one just seems to print everything after the first CPIO, regardless if there's just one or multiple ones left.

jamescassell: That explains why I couldn't figure out where root.squashfs was coming from...

tuan-hoang1: I think I had a patch lying somewhere that the big fat initramfs (including normal initramfs and real rootfs at `/root.squashfs`) works as a live image.
This is the (only for z/VM on s390x) way out for kernel + initramfs live boot method for s390x.

nikita-dubrovskii: > Are you able to boot the live image on s390x?

on zVM it's possible with live-initramfs, not live-iso

nikita-dubrovskii: with unpatched ramdisk `ostree-prepare-root.service` fails during boot. 
I'll continue investigation

jlebon: > on zVM it's possible with live-initramfs, not live-iso

Gotcha. How are you booting with the live initramfs directly, via live PXE or some s390x-specific thing?

Either way, if it boot fine all the way to the real root, you should have access to coreos-installer there, right?

nikita-dubrovskii: It's s390x specific: on zVM kernel, ramdisk and boot.prm should be directly copied into `reader queue`, and then `ipl`'ed (booted).
Here some more details about it:
[https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/installation_guide/sect-installation-overview-s390](url)

I wasn't aware about `/root.squashfs` exists within ramdisk, so decided to add `coreos-installer`.  Currently CoreOS failes during boot - mount issues. I'll try fixinig  it today.

nikita-dubrovskii: so, it was my bad. the script i've wrote and used to boot CoreOS on zVM had a bug preventing `ostree-prepare-roo`t to parse cmdline. after fixing it `root.squashfs` mounts without issues

PR #295, opened 2020-03-11
[WIP] multipath support

tuan-hoang1:


lucab:
@tuan-hoang1 @darkmuggle is this PR something that we can drop in favor of the recent multipath work in FCOS?

darkmuggle (changes requested):
@tuan-hoang1 this PR is no longer needed, and IMO, we should not be doing a default multipath configuration for _all_ FCOS/RHCOS instances. With the Dracut support for `rd.multipath=default root=/dev/disk/by-label/dm-mpath-root` full support is in FCOS now. 

https://koji.fedoraproject.org/koji/buildinfo?buildID=1517388 which includes
https://src.fedoraproject.org/rpms/dracut/blob/308914b60e6809c6b644ba4d11eb11a769491019/f/0033.patch

tuan-hoang1:
Yes, thanks for doing this hard work. Closing.

PR #296, opened 2020-03-11
manifests: add ppc64le packages to detect dynamically attached devices

Prashanth684:
While running the Openshift CI encountered: https://bugzilla.redhat.com/show_bug.cgi?id=1811537
which requires some extra ppc64le libraries required to detect dynamically attached pci devices.
These devices show up through the diag events which is why the diag libraries are necessary

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -165,6 +165,9 @@ packages-x86_64:
   - irqbalance
 packages-ppc64le:
   - irqbalance
+  - librtas
+  - powerpc-utils-core
+  - ppc64-diag-rtas
 packages-aarch64:
   - irqbalance
 
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -22,3 +22,7 @@ enable zincati.service
 enable coreos-liveiso-success.service
 # See bootupd.yaml
 enable bootupd.socket
+# Enable rtas_errd for ppc64le to discover dynamically attached pci devices - https://bugzilla.redhat.com/show_bug.cgi?id=1811537
+# The event for the attached device comes as a diag event.
+# Ideally it should have been added as part of base Fedora - but since it was arch specific, it was not added: https://bugzilla.redhat.com/show_bug.cgi?id=1433859
+enable rtas_errd.service


overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
```diff
@@ -13,3 +13,6 @@ enable afterburn-firstboot-checkin.service
 enable afterburn-sshkeys@.service
 # Update agent
 enable zincati.service
+# Enable rtas_errd for ppc64le to discover dynamically attached pci devices - https://bugzilla.redhat.com/show_bug.cgi?id=1811537
+# The event for the attached device comes as a diag event.
+enable rtas_errd.service
```
cgwalters, line 28: This isn't a service specific to FCOS - we should probably enable this [globally in Fedora](https://docs.fedoraproject.org/en-US/packaging-guidelines/DefaultServices/) right?

manifests/fedora-coreos-base.yaml
```diff
@@ -170,6 +170,9 @@ packages-x86_64:
   - irqbalance
 packages-ppc64le:
   - irqbalance
+  - librtas
+  - powerpc-utils
+  - ppc64-diag
```
cgwalters: Can you paste here the additional dependency chain from this?

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
cgwalters: That said, I really wonder what's going on with this - why does just ppc64 need an additional userspace bit for PCI?

cgwalters: Upstream is https://github.com/power-ras/ppc64-diag ...not a lot of docs I can see.  Example code: https://github.com/power-ras/ppc64-diag/blob/master/rtas_errd/hotplug.c

Is this something like a reimplementation of a subset of udev?

Prashanth684: @cgwalters yes it is not specific to fcos, but i believe it is not part of the base package because it is not something that is essential for the system. @andymcc I remember you said this was needed for detecting attached virsh volumes? @mkumatag - could you describe the use case for this in detail ?

mkumatag: All I know is this service will help us adding hotplug devices(cpu, memory, disks), @hegdevasant or someone from his can help us understanding more on this.

Prashanth684: @cgwalters also found this: https://bugzilla.redhat.com/show_bug.cgi?id=1433859 the exact same discussion on adding it to Fedora and looks like it was added as optional.

cgwalters: Even better than writing up the purpose of the software here is to have something modern like a `README.md` in the upstream sources - the current README only talks about how to build it, not why I'd want it etc.

manifests/fedora-coreos-base.yaml
Prashanth684: is this what you're looking for: https://paste.centos.org/view/e4bdea16 ?

cgwalters: `/usr/bin/perl` :cry: 

Well, kind of...`cosa build ` will automatically run `rpm-ostree db diff` between two builds; so usually to get the diff I just do a `cosa build ostree` without the patch (`git stash` or whatever), then `git stash apply` in config git, then cosa build ostree` again.

cgwalters: Or, you can manually `rpm-ostree db diff --repo=tmp/repo fedora/x86_64/coreos/testing-devel` etc.

Prashanth684: oh..duh ..yeah thanks for reminding me about that: https://paste.centos.org/view/68589910

yeah so /usr/bin/perl indeed :)

Prashanth684: this id the fcos diff - the previous one i pasted was rhcos: https://paste.centos.org/view/5524482d

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
hegdevasant: I don't know anything about this project. Here is some generic details about pp64-diag:
  - librtas contains set of API to make RTAS calls from user space
  -  We have `rtas_errd` daemon inside this package.. which monitors for platform generated events like cpu/mem hotplug etc. It then prepared `drmgr` command and makes drmgr call.
  - drmgr (part of powerpc-utils) does actual hotplug.

drmgr code is maintained by Tyrel (tyreld@linux.vnet.ibm.com) . If you need internal details please check with him.

-Vasant

manifests/fedora-coreos-base.yaml
jlebon: FCOS doesn't ship any perl right now, and it'd be preferable to keep it that way if possible (for example, FCOS doesn't ship python either right now). Do we really need Perl though to support hotplugged devices? Is there a more minimal set of packages that we can install that will do what we want? Can we tweak packaging to make the Perl-requiring bits a separate subpackage?

Prashanth684: From the ppc64-diag and powerpc-utils spec files, it looks like it is required. Maybe @sharkcz has a better idea about this, i'm not sure.

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
hegdevasant: > Upstream is https://github.com/power-ras/ppc64-diag ...not a lot of docs I can see. Example code: https://github.com/power-ras/ppc64-diag/blob/master/rtas_errd/hotplug.c
> 
> Is this something like a reimplementation of a subset of udev?

These are platform events and also once we do hotplug we do update device tree etc. Hence I don't think `udev` is sufficient here.

manifests/fedora-coreos-base.yaml
hegdevasant: > `/usr/bin/perl` cry
> 
> Well, kind of...`cosa build ` will automatically run `rpm-ostree db diff` between two builds; so usually to get the diff I just do a `cosa build ostree` without the patch (`git stash` or whatever), then `git stash apply` in config git, then cosa build ostree` again.

Unfortunately original authors used various languages/scripts for various purpose. We have perl script inside this package.. which is mostly used for event notification infrastructure.

If you just need `rtas_errd (/rtas_errd directory) , you should be able to build without perl dependency. It might be little bit of extra work.

Same goes with powerpc-utils. I think if you just need `drmgr` family tools then you will be able to cut the dependency list.

-Vasant

cgwalters:
I feel like I'm in a constant pitched battle to avoid people just adding stuff to the host.  This one is about hardware enablement, which is a usual argument for adding it.

I would really much prefer if someone at least tried to drop the perl dependency.  And it's not at all obvious to me why ppc64le needs a special daemon to handle hotplug events of block devices.  I am guessing there's just a lot of legacy here.  These virtio block devices show up as PCI, right?  Which I'd imagine is handled via the [PCIe hotplug model](https://lwn.net/Articles/767885/).  Are we emulating some sort of older ppc64le hardware or something?

Anyways...short term; I am not totally opposed to just merging this as is (no one is going to notice short term if we ship `perl` just on ppc64le) but hopefully someone can at least try to drop that.

jlebon:
I'm also hesitant to merge this without at least some basic investigation into how hard it would be to drop the perl dependency. If it's not that hard, great! If it turns out to be very complex, we can at least log an issue in the tracker and the relevant upstream (or just RHBZ) and reference them in the manifest here for tracking.

mkumatag:
> I'm also hesitant to merge this without at least some basic investigation into how hard it would be to drop the perl dependency. If it's not that hard, great! If it turns out to be very complex, we can at least log an issue in the tracker and the relevant upstream (or just RHBZ) and reference them in the manifest here for tracking.

I have created BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1814335 to track this issue.

jamescassell:
Bug access denied

mkumatag:
> Bug access denied

can you try now?

manifests/fedora-coreos-base.yaml
ngothan: To avoid the perl dependency in powerpc-utils and ppc64-diag i splitted the rtas_errd daemon into subpackages (ppc64-diag-rtas) and moved the drmgr into subpackage powerpc-utils-core in fedora f33.

You need following change:

```
diff --git a/manifests/fedora-coreos-base.yaml b/manifests/fedora-coreos-base.yaml
index 8fccef9..b72350e 100644
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -166,6 +166,7 @@ packages-x86_64:
   - irqbalance
 packages-ppc64le:
   - irqbalance
+  - ppc64-diag-rtas
 packages-aarch64:
   - irqbalance
```

Could you please test it?

Thank you so much!
Than

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
Prashanth684: added comment as to why it wasn't added into base Fedora.

manifests/fedora-coreos-base.yaml
Prashanth684: now that we have moved to f33 , i have added the packages which are only required for the rtas service to run, thus avoiding the perl dependency.

Prashanth684:
@cgwalters now that we have moved to f33, i have added the core packages which do not pull in perl.

jlebon:
@Prashanth684 Can you print out the pkgdiff from this PR on ppc64le? CI only covers x86_64 unfortunately.

Prashanth684:
> @Prashanth684 Can you print out the pkgdiff from this PR on ppc64le? CI only covers x86_64 unfortunately.

```
diff --git a/packages b/packages
index 6ad2d97..28cef92 100644
--- a/packages
+++ b/packages
@@ -13,6 +13,7 @@
   basesystem-11-10.fc33.noarch (fedora)
   bash-5.0.17-2.fc33.ppc64le (fedora)
   bash-completion-1:2.8-9.fc33.noarch (fedora)
+  bc-1.07.1-11.fc33.ppc64le (fedora)
   bind-libs-32:9.11.24-2.fc33.ppc64le (fedora-updates)
   bind-libs-lite-32:9.11.24-2.fc33.ppc64le (fedora-updates)
   bind-license-32:9.11.24-2.fc33.noarch (fedora-updates)
@@ -223,6 +224,7 @@
   libselinux-utils-3.1-2.fc33.ppc64le (fedora)
   libsemanage-3.1-2.fc33.ppc64le (fedora)
   libsepol-3.1-3.fc33.ppc64le (fedora)
+  libservicelog-1.1.18-9.fc33.ppc64le (fedora)
   libsigsegv-2.11-11.fc33.ppc64le (fedora)
   libslirp-4.3.1-2.fc33.ppc64le (fedora)
   libsmartcols-2.36-3.fc33.ppc64le (fedora)
@@ -315,6 +317,8 @@
   polkit-libs-0.117-2.fc33.ppc64le (fedora)
   polkit-pkla-compat-0.1-18.fc33.ppc64le (fedora)
   popt-1.18-2.fc33.ppc64le (fedora)
+  powerpc-utils-core-1.3.8-3.fc33.ppc64le (fedora-updates)
+  ppc64-diag-rtas-2.7.6-6.fc33.ppc64le (fedora)
   procps-ng-3.3.16-1.fc33.ppc64le (fedora)
   protobuf-c-1.3.3-3.fc33.ppc64le (fedora)
   psmisc-23.3-4.fc33.ppc64le (fedora)
```
looks like bc and servicelog are needed by rtas

jlebon (approved):

PR #298, opened 2020-03-12
[stable] tree: promote changes from testing at 0f1ff85b7cf9455e2075b9023d5794c…

jlebon:
…9491b54ca


jlebon:
Note the second commit here:

```
ci: import from testing-devel at 640ecad

We have to do this to get working upstream CI for this release, because
the old CI code doesn't work with the new CI infra.
```

dustymabe (approved):
LGTM

PR #300, opened 2020-03-12
[bodhi-updates] manifest-lock: temporarily pin ostree to 2019.5

jlebon:
We need some more fixes to the readonly path before we can bump to
2020.2:

https://github.com/coreos/fedora-coreos-tracker/issues/343

### DIFF
--- a/manifest-lock.x86_64.yaml
+++ b/manifest-lock.x86_64.yaml
@@ -1,0 +1,7 @@
+packages:
+  # freezing at 2019.5 until we resolve issues with sysroot.readonly
+  # https://github.com/coreos/fedora-coreos-tracker/issues/343
+  ostree:
+    evra: 2019.5-2.fc31.x86_64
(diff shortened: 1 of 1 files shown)


jlebon:
See https://github.com/coreos/fedora-coreos-tracker/issues/293 for why we need to do this to unblock updates.

dustymabe:
A dumb question: do the failures cause any concern for other editions that use OSTree ? Or is this really only a failure with readonly /sysroot/ which is something new only we are trying to do?

dustymabe (approved):
LGTM

jlebon:
> A dumb question: do the failures cause any concern for other editions that use OSTree ? Or is this really only a failure with readonly /sysroot/ which is something new only we are trying to do?

Right exactly. AFAIK, only FCOS today uses `sysroot.readonly=true`.

PR #301, opened 2020-03-12
[bodhi-updates] manifest-lock: put ostree pin in overrides instead

jlebon:
We want this as an override, not a base lock. A subtle difference is
that versionary uses the presence of a base lockfile as a marker that
(understandably) the full base content is locked, and bases versioning
on that.


jlebon:
This should unblock bodhi-updates builds.
Requires: https://github.com/coreos/coreos-assembler/pull/1233

dustymabe (approved):

PR #302, opened 2020-03-16
[bodhi-updates] overrides: stop pinning ostree

jlebon:
Issues with read-only sysroot should be fixed now!
https://github.com/coreos/fedora-coreos-tracker/issues/343


dustymabe (approved):
LGTM

PR #303, opened 2020-03-16
overrides: add {rpm-,}ostree for read-only /sysroot fixes

jlebon:
Fetch the newer ostree release and rpm-ostree rebuild which complete the
read-only sysroot enablement.


dustymabe (approved):
LGTM

PR #304, opened 2020-03-16
image.yaml: enable sysroot-ro

jlebon:
We want to use the new read-only `/sysroot` feature of libostree. Opt-in
to that to tell cosa we support it and want it.

For more details, see:
https://github.com/ostreedev/ostree/issues/1265
https://github.com/coreos/coreos-assembler/pull/1235

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -13,6 +13,10 @@ extra-kargs:
 # Optional remote by which to prefix the deployed OSTree ref
 ostree-remote: fedora
 
+# We want read-only /sysroot to protect from unintentional damage.
+# https://github.com/ostreedev/ostree/issues/1265
+sysroot-readonly: true
+
 # After this, we plan to add support for the Ignition
 # storage/filesystems sections.  (Although one can do
 # that on boot as well)


dustymabe (approved):
LGTM - do you have a good idea of this breaking anything?

jlebon:
> LGTM - do you have a good idea of this breaking anything?

After https://github.com/coreos/fedora-coreos-config/pull/303, we're essentially already using it. This is more so that we don't go back to not use it once we make it opt-in in cosa (https://github.com/coreos/coreos-assembler/pull/1235).

I think most issues have been shaken out now. At least updates work, so any other remaining issues we should be able to fix in follow-ups if necessary.

jlebon:
Let's hold this until https://github.com/coreos/coreos-assembler/pull/1235 merges in case the field name changes there.

cgwalters:
Force pushed this to update to the new name from https://github.com/coreos/coreos-assembler/pull/1235

ashcrow:
> Let's hold this until [coreos/coreos-assembler#1235](https://github.com/coreos/coreos-assembler/pull/1235) merges in case the field name changes there.

Per https://github.com/coreos/coreos-assembler/pull/1235#issuecomment-630397247 it sounds like this needs to merge first.

ashcrow (approved):

ashcrow:
LGTM merging

PR #307, opened 2020-03-17
overlay/ignition-ostree-growfs: Don't run if root= karg

cgwalters:
This occurs in the currently artificial scenario of the RHCOS
upgrade tests from previous 4.2 Anaconda-based builds.  This wouldn't
happen in practice because growfs is already keyed off only executing
on the Ignition boot, and there's no way we'd rebase this code and
release updated 4.2 bootimages.  But, we already have similar exclusions in
the mount code, so let's do the same for growfs.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
@@ -2,6 +2,9 @@
 Description=Ignition OSTree: Grow root filesystem
 DefaultDependencies=false
 ConditionKernelCommandLine=ostree
+# Similar to the other mount rules, suppress invocation if we detect
+# we are running from a legacy setup created by Anaconda.
+ConditionKernelCommandLine=!root
 ConditionPathExists=!/run/ostree-live
 Before=initrd-root-fs.target
 After=ignition-ostree-mount-firstboot-sysroot.service


dustymabe:
can we add a comment that gives guidance on when this can be removed? Seems like something we'll only need for a short period of time.

cgwalters:
> can we add a comment that gives guidance on when this can be removed? Seems like something we'll only need for a short period of time.

It's not any different from

>  But, we already have similar exclusions in the mount code, so let's do the same for growfs.

I don't think it's worth trying to remove these anytime soon.

jlebon (approved):
This is fine, but it's essentially RHCOS-specific implementation details leaking into FCOS. I think we should work towards supporting RHCOS in the `run-upgrade` suite instead (see https://github.com/coreos/coreos-assembler/issues/1158) and drop that artificial test.

dustymabe:
> Similar to the other mount rules, suppress invocation if we detect
> we are running from a legacy setup created by Anaconda.

This is good info, but would be better if we say something like what was the last release we used Anaconda for. Then it's real easy to know when we no longer need it.

dustymabe (approved):

cgwalters:
> This is fine, but it's essentially RHCOS-specific implementation details leaking into FCOS. 

[Not quite](https://github.com/coreos/fedora-coreos-config/pull/247) right?

> I think we should work towards supporting RHCOS in the run-upgrade suite instead (see coreos/coreos-assembler#1158) and drop that artificial test.

This I obviously agree with!  But while *this* case was "synthetic", others won't be I think.

> This is good info, but would be better if we say something like what was the last release we used Anaconda for. Then it's real easy to know when we no longer need it.

I think we need to support in-place updates *from* RHCOS 4.2 until such time as we have [bootimage updates for OpenShift 4](https://github.com/openshift/enhancements/pull/201).

So...I'm not seeing us removing this anytime soon.  It's not a huge tech debt at least.

PR #308, opened 2020-03-19
manifest/base: add dhcp-client

jlebon:
The latest dracut (050) switched to a rich dependency to accept either
NM or `dhclient` for doing DHCP in the initrd:

https://github.com/dracutdevs/dracut/commit/e8638076857621b28eb4dc36cdd49a94992835b7

which caused us to lose it in FCOS since we do ship NM in the real root.

The dependency there should really be something like "require NM if it's
used in the initrd and the real root, otherwise require dhclient", but
you can't really do that in a spec file.

Until we have NetworkManager in the initrd, we definitely need
`dhclient` available for DHCP there.

Resolves: https://github.com/coreos/fedora-coreos-tracker/issues/429

PR #309, opened 2020-03-19
manifest/ignition: require Ignition dracut module

jlebon:
Explicitly do `--add=ignition` when calling `dracut` so that we fail if
for whatever reason it couldn't be included.

For more background, see:
https://github.com/coreos/fedora-coreos-tracker/issues/429

### DIFF
--- a/manifests/ignition-and-ostree.yaml
+++ b/manifests/ignition-and-ostree.yaml
@@ -8,6 +8,10 @@
 # Include rpm-ostree + kernel + bootloader
 include: bootable-rpm-ostree.yaml
 
+initramfs-args:
+  # make it a hard error if Ignition can't be included
+  - --add=ignition
+
 # Modern defaults we want
 boot-location: modules
 tmp-is-dir: true


cgwalters (approved):

cgwalters:
Though maybe this should go in `ignition-and-ostree.yaml` ?

dustymabe (approved):

dustymabe:
so ignition wasn't added because dhclient wasn't installed ?

jlebon:
> Though maybe this should go in `ignition-and-ostree.yaml` ?

Done!

> so ignition wasn't added because dhclient wasn't installed ?

Yup, you can see the full chain of dependencies here: https://github.com/coreos/fedora-coreos-tracker/issues/429#issuecomment-601302105

With this commit (but without #308):

```
dracut: Executing: /usr/bin/dracut --reproducible --gzip -v --add ostree --tmpdir=/tmp/dracut -f /tmp/initramfs.img --add=ignition --no-hostonly --omit=nfs --omit=lvm --omit=multipath --omit=iscsi --kver 5.5.9-200.fc31.x86_64
...
dracut: dracut module 'network-legacy' will not be installed, because command 'dhclient' could not be found!
dracut: dracut module 'network' depends on 'network-legacy', which can't be installed
dracut: dracut module 'afterburn' depends on 'network', which can't be installed
dracut: dracut module 'url-lib' depends on 'network', which can't be installed
dracut: dracut module 'ignition' depends on 'url-lib', which can't be installed
dracut: dracut module 'ignition' depends on 'url-lib', which can't be installed
dracut: dracut module 'ignition-conf' depends on 'ignition', which can't be installed
dracut: dracut module 'ignition-ostree' depends on 'ignition', which can't be installed
...
dracut: dracut module 'ignition' cannot be found or installed.
error: Finalizing rootfs: During kernel processing: Executing bwrap(rpmostree-dracut-wrapper): Child process killed by signal 1
```

PR #305, opened 2020-03-16
tests/kola: New `basic` test using kola external tests

cgwalters:
Pairs with https://github.com/coreos/coreos-assembler/pull/1252

The motivation here is to get the ball rolling for (where it
makes sense) to have a place to add tests corresponding to changes
in this repository.

PR #310, opened 2020-03-19
fedora-coreos-base: use NetworkManager for networking in the initramfs

### LINKED ISSUE #394: Using NetworkManager in the initramfs
We had been planning to move to using Networkmanager in the initrd for a while. We now have a host of issues that have cropped up that we think would be solved by making this move.

This is a separate ticket (related to https://github.com/coreos/fedora-coreos-tracker/issues/24) that specifically can track moving to networkmanager in the initrd.

dustymabe:
Moving to NetworkManager in the initrd should help us solve some
problems we've been having with Networking. For what we want to
do in Fedora CoreOS doing this right requires the network to be
torn down in the initrd and also possibly propagated forward.

Requires: https://github.com/coreos/ignition-dracut/pull/159
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/394

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -55,9 +55,6 @@ remove-from-packages:
               /usr/lib/systemd/system/systemd-networkd-wait-online.service]
   - [systemd-container, /usr/lib/systemd/network/.*]
   - [systemd-udev, /usr/lib/systemd/network/.*]
-  # We're still using the legacy dracut network module for now. Dracut keys off
-  # of this: https://github.com/dracutdevs/dracut/blob/1fcc70fe57eea0ea658aa2de5c0044683fe85cf1/modules.d/40network/module-setup.sh#L11
-  - [NetworkManager, /usr/libexec/nm-initrd-generator]
 
 
 remove-files:
@@ -116,10 +113,6 @@ packages:
   - NetworkManager hostname
   - iproute-tc
   - adcli
-  # We still need this for now, but may drop it once we have NM in the initrd
-  # https://github.com/coreos/fedora-coreos-tracker/issues/394
-  # https://github.com/coreos/fedora-coreos-tracker/issues/429
-  - dhcp-client
   ## Teaming https://github.com/coreos/fedora-coreos-config/pull/289 and http://bugzilla.redhat.com/1758162
   - NetworkManager-team teamd
   # Static firewalling
--- a/manifest-lock.overrides.x86_64.yaml
+++ b/manifest-lock.overrides.x86_64.yaml
@@ -22,3 +22,7 @@ packages:
     evra: 2020.1.21.ge9011530-2.fc31.x86_64
   rpm-ostree-libs:
     evra: 2020.1.21.ge9011530-2.fc31.x86_64
+  # Fast track new Ignition with initramfs network takedown
+  # https://bodhi.fedoraproject.org/updates/FEDORA-2020-6db5551bf6
+  ignition:
+    evra: 2.2.1-1.git2d3ff58.fc31.x86_64


dustymabe:
this worked in some early testing. I've still got more testing to do and maybe we can automate some of it and put it in CI. I just rebased on top of latest master so respinning everything locall too.

Also one issue that came up is how NM processes the dracut commandline args for networking: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/391 We'll see what they think about the behavior.

manifests/fedora-coreos-base.yaml
```diff
@@ -18,6 +18,9 @@ initramfs-args:
   - --omit=lvm
   - --omit=multipath
   - --omit=iscsi
+  # Networking is NetworkManager
+  - --omit=systemd-networkd network-legacy
+  - --add=network-manager
```
jlebon: Question: do we actually want to be explicit about this? A big part of FCOS is following Fedora defaults where possible, and previously we were actively working to avoid the default of NM-in-initrd. If we just drop that part (the `nm-initrd-generator` bit), we essentially get the same thing, right? My personal opinion is this would be cleaner, but don't feel strongly about this though.

jlebon (approved):

manifests/fedora-coreos-base.yaml
dustymabe: IOW don't be explicit here and it *should* revert to the default of NM in the initrd because we are no longer deleting the nm-initrd-generator binary?

jlebon: Right, exactly.

jlebon: Actually, initramfs args is one place where default vs explicit propagates to the user now that rpm-ostree knows to re-use the same base args when regenerating the initrd (https://github.com/coreos/rpm-ostree/pull/1997). In the default case, one can override it with `rpm-ostree initramfs --add=network-legacy`. In the explicit case, I think one would have to do `--omit=network-manager --add=network-legacy`. (Not saying one would want to do this, but it demonstrates the difference well.)

cgwalters (approved):
It'd be good to have someone besides Dusty play with this before we land it I think.

jlebon:
Yup, sanity checking this on top of https://github.com/coreos/ignition-dracut/pull/159 right now!

manifests/fedora-coreos-base.yaml
dustymabe: ok I pushed up a rebase that removes the dracut args. Tests look like it still works with nm in the initrd.

jlebon (approved):
Dusty tested this on top of a bunch of combinations around bonding, teaming, bridging, and static networking. I tested it on top of various DHCPv4 and v6 configurations. All seems to be working well! (And any regression I think we can work with the NM team to get fixed as follow-ups.)

Thanks a lot for this!

cgwalters:
BTW you could probably create a new test like https://github.com/coreos/fedora-coreos-config/pull/305 that maybe uses e.g. `journalctl` to scrape out some bits from NetworkManager to verify it ran in the initramfs or so.

dustymabe:
> BTW you could probably create a new test like #305 that maybe uses e.g. `journalctl` to scrape out some bits from NetworkManager to verify it ran in the initramfs or so.

yeah I don't see a good string in the journal logs that indicate NM is running in the initramfs (vs in the real root) other than things like

```
NetworkManager[525]: <info>  [1585084135.2919] auth[0x5651fccfaca0]: create auth-manager: D-Bus connection not available.
```

Maybe I can look at the strings from the teardown service:

```
info: propagating initramfs networking config to the real root
```

If I add it I'll add it in a follow up since I'd like to get this out today.

dustymabe:
ok I added a second commit that adds an override for the ignition rpm to get the latest bits that have https://github.com/coreos/ignition-dracut/pull/159. Will merge this after tests pass.

dustymabe:
ok I've been testing this using an ugly script for now (better than hand testing at least 😄). It goes through and makes sure variations of initrd passed networking config and ignition passed networking config works. See https://github.com/coreos/fedora-coreos-config/pull/316

dustymabe:
tests passed!

PR #311, opened 2020-03-20
overlay/05core: Set SYSTEMD_SULOGIN_FORCE=1 for emergency.service

cgwalters:
This makes
https://github.com/coreos/coreos-installer/commit/15a79263d0bd5d72056a6080f6687dc10cba2dda
global.

We want booting with `systemd.unit=emergency.target` to Just Work even with our locked
root account.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/emergency.service.d/coreos-sulogin-force.conf
+++ b/overlay.d/05core/usr/lib/systemd/system/emergency.service.d/coreos-sulogin-force.conf
@@ -1,0 +1,6 @@
+# https://github.com/coreos/coreos-installer/commit/15a79263d0bd5d72056a6080f6687dc10cba2dda
+# https://github.com/systemd/systemd/pull/10397
+# We want `systemd.unit=emergency.target` to Just Work even with our locked
+# root account.
+[Service]
+Environment=SYSTEMD_SULOGIN_FORCE=1
--- a/overlay.d/05core/usr/lib/systemd/system/rescue.service.d/coreos-sulogin-force.conf
+++ b/overlay.d/05core/usr/lib/systemd/system/rescue.service.d/coreos-sulogin-force.conf
@@ -1,0 +1,1 @@
+../emergency.service.d/coreos-sulogin-force.conf


jlebon (approved):

dustymabe:
Thanks @cgwalters - a few questions:

- Does this make `single` work?
- Do you think after we merge this we should see if the rest of Fedora should want this change?

cgwalters:
> Does this make single work?

No, seems `single` is something else.

> Do you think after we merge this we should see if the rest of Fedora should want this change?

Probably; the reason this isn't the default in systemd upstream is I think they want to support setups where the bootloader is locked, and someone with access to the physical console shouldn't be able to just get a root shell if bootup fails.

cgwalters:
> No, seems single is something else.

Looks like it's `rescue.target` so I'll update this to do the same thing there.

cgwalters:
Done, working for `rescue` too.

dustymabe:
thanks - let me give this a spin

dustymabe:
seems to work:

```
You are in rescue mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, "systemctl default" or "exit"
to boot into default mode.
Press Enter for maintenance
(or press Control-D to continue): 
[root@localhost ~]#
```

cc @bgilbert for review

dustymabe (approved):
LGTM

bgilbert (approved):
LGTM!

strigazi:
Hello,

I have a machine that does not have the proposed config in place, so `single` doesn't work. Is there a way around it? i.e. do smth in grub, get access to root to change the passed, login in the console.

cc @dustymabe

jlebon:
> I have a machine that does not have the proposed config in place

Is this on an old FCOS boot image? Or are you using a different variant/distro entirely?

strigazi:
I'm on Fedora CoreOS 31:
```
                   Version: 31.20200113.3.1 (2020-01-14T00:20:15Z)
                    Commit: f480038412cba26ab010d2cd5a09ddec736204a6e9faa8370edaa943cf33c932
              GPGSignature: Valid signature by 7D22D5867F2A4236474BF7B850CB390B3C3359C4
```

jlebon:
Can you update to a newer boot image? You should also be able to drop those same files via your Ignition config.

Otherwise, it's also possible to override `init` as described here: https://github.com/coreos/fedora-coreos-docs/pull/48 (this was then superseded by `single` which this PR fixes).

strigazi:
This is a host that is running since weeks, first we need to get access to it. Then we can update to a newer version. I don't if I can do it now. Can I update ignition on a host that has already booted?

I followed https://github.com/coreos/fedora-coreos-docs/pull/48 but i had to do some changes.
* replace `console=tty0 console=ttyts0,115200n8` `rw init=/bin/bash`
** just adding init=/bin/sh or /bin/bash was not working. I couldn't get the console.
* I didn't notice the selinux part in the first attenpt and I believe I corrupted /etc/shadow and /etc/passwd, so I had to disable selinux to get access. I was getting an avc denied wihtout it.

dustymabe:
So you were able to get access eventually?

Now that you have access can you update the node? `31.20200113.3.1` is really out of date.

strigazi:
I managed to get access. As soon as the node is fixed (HW issue) will try to update.

PR #306, opened 2020-03-17
[next-devel] manifest: point next-devel at f32

jlebon:
This branch is forked from the testing-devel branch to start a
next-devel stream which will currently track f32. This is something we
should've done for a while now.

PR #312, opened 2020-03-20
ci: buildprep from stream matching branch name

jlebon:
E.g. on `next-devel` we want CI to `buildprep` from the `next-devel`
stream.

Required for green CI in:
https://github.com/coreos/fedora-coreos-config/pull/306


jlebon:
Cool, looks like that works!

dustymabe (approved):
LGTM - if you say it works I trust you 😸

PR #313, opened 2020-03-23
overlay/ignition-ostree: Add a Requires: for sysroot in growpart

cgwalters:
A user reported on IRC seeing this service crash; I think that
might happen if the service runs even without the rootfs mounted
which could happen if the firstboot mount failed.

Add a `Requires:` to express the fact this is a hard dependency.


cgwalters:
OK actually I was wrong it wasn't growpart that was failing, it was the main mount.  So this wouldn't have helped.  (But eh, it's also not wrong)

jlebon (approved):
Yup, makes sense to me!
Needs a rebase.

cgwalters:
> Needs a rebase.

:surfing_man:

jlebon (approved):

PR #316, opened 2020-03-24
tests: add a test that can be run manually for net testing

dustymabe:
This is how I've been doing network testing for the NM in
the initrd PR (https://github.com/coreos/fedora-coreos-config/pull/310).
One day I'd like to convert this into some more sophisticated
test that can be run in our CI but AFAIU right now that's not
possibly because our CI is limited to qemu unprivileged networking.

For now let's put this $somewhere so that we can pull it out
at least when we suspect problems.

### DIFF
--- a/tests/manual/fcos-network-testing.sh
+++ b/tests/manual/fcos-network-testing.sh
@@ -1,0 +1,486 @@
+#!/usr/bin/bash
+set -eu -o pipefail
+#set -x
+
+# This script attempts to test networking configuration in various
+# scenarios/configurations. It tries to test what should happen
+# when initramfs networking is passed and/or networking config is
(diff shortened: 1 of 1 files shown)


dustymabe:
I went though and talked to @jlebon about where would be the best place to put something like this. We came up with `tests/manual` in this repo for now. I don't know if there are problems with doing that or not, but this is an attempt at that.

cgwalters:
OK yeah, this looks like a net win.  See some discussions in https://github.com/coreos/fedora-coreos-tracker/issues/340 for libvirt.

lucab:
Isn't this something that belongs to coreos-assembler / mantle? Even if not part of automated CI, it would helpful not to start sprawling scripts around.

PR #318, opened 2020-03-25
overrides: fasttrack rust-zincati-0.0.9-1.fc31

dustymabe:
New version that includes various fixes and also support
for running behind a proxy.


dustymabe:
will update the link to bodhi once the update is made

dustymabe:
> will update the link to bodhi once the update is made

done - ready for review

jlebon (approved):

jlebon:
Upgrade testing should at least sanity-check basic functionality in the new Zincati. (But since the zincati repo itself has that test in its CI, we shouldn't expect any surprises here. :) )

dustymabe:
will merge once CI passes

PR #320, opened 2020-03-25
manifests: drop whois-nls

dustymabe:
No one really knows what it's needed for and no other
package seems to require it.


dustymabe:
this is a follow on from the discussion in https://github.com/coreos/fedora-coreos-tracker/issues/432#issuecomment-602310619

cc @bgilbert

cgwalters (approved):

bgilbert (approved):

jlebon:
Number of packages going down... these moments should be celebrated! Just gonna drop a :tada: at least. :)

PR #321, opened 2020-03-26
WIP: Run NM via systemd unit, don't depend on ip=dhcp kargs

cgwalters:
See https://github.com/coreos/ignition-dracut/issues/94
and https://github.com/coreos/ignition/pull/948

Needs pairing with a cosa PR to drop the default `ip=dhcp` kargs.

And yes we really want to upstream this into NM by default or so.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/coreos-NetworkManager.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/coreos-NetworkManager.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Network Manager (CoreOS initrd)
+Wants=network.target
+After=network-pre.target dbus.service
+Before=network.target network.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/nm-initrd-generator rd.neednet=1 ip=dhcp,dhcp6
+ExecStart=/usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
+CapabilityBoundingSet=CAP_NET_ADMIN CAP_DAC_OVERRIDE CAP_NET_RAW CAP_NET_BIND_SERVICE CAP_SETGID CAP_SETUID CAP_SYS_MODULE CAP_AUDIT_WRITE CAP_KILL CAP_SYS_CHROOT
--- a/overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/module-setup.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+depends() {
+    echo ignition network-manager
+}
+
+install_and_enable_unit() {
+    unit="$1"; shift
+    target="$1"; shift
+    inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
+    mkdir -p "$initdir/$systemdsystemunitdir/$target.requires"
+    ln_r "../$unit" "$systemdsystemunitdir/$target.requires/$unit"
+}
+
+install() {
+    # We're forcibly overriding NM to be run as a service
+    install_and_enable_unit coreos-NetworkManager.service network-online.target
+}
--- a/live/EFI/fedora/grub.cfg
+++ b/live/EFI/fedora/grub.cfg
@@ -28,6 +28,6 @@ set timeout=5
 
 ### BEGIN /etc/grub.d/10_linux ###
 menuentry 'Fedora CoreOS (Live)' --class fedora --class gnu-linux --class gnu --class os {
-	linux /images/vmlinuz @@KERNEL-ARGS@@ rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
+	linux /images/vmlinuz @@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal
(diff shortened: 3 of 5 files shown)


dustymabe:
I do really want NM in the initrd and I've talked with the NM team about it (see [the RFE they asked me to open](https://bugzilla.redhat.com/show_bug.cgi?id=1814038)), but I think it's risky to move away from the current NM initrd implementation today. I'd prefer if we kept the initrd implementation to the defaults (and hopefully NM upstream will move to NM started by systemd in the initrd soon).

An alternative way to achieve the goal of booting a live ISO without networking is to dynamically add some dracut networking args at runtime if some specific conditions are met. The POC is in #326. This way will allow us to keep the same NM initrd implementation that is the default in NM upstream.

jlebon:
https://github.com/coreos/ignition-dracut/pull/164 works on top of this by only activating `network.target` if the Ignition config requires it.

Hmm so actually we can delete the `nm-online.conf` dropin from here.

cgwalters:
> but I think it's risky to move away from the current NM initrd implementation today.

That's a reasonable position.  OTOH, this approach is still running NM as a *oneshot* unit - we're not trying to run as a daemon, which would definitely be a bigger change.

IOW, we're just lifting the execution out of the abandoned swap of the initqueue into the shiny modern glass office tower of systemd units.

overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/module-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+    unit="$1"; shift
+    target="$1"; shift
+    inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
+    mkdir -p "$initdir/$systemdsystemunitdir/$target.requires"
+    ln_r "../$unit" "$systemdsystemunitdir/$target.requires/$unit"
+}
+
+install() {
+    # We're forcibly overriding NM to be run as a service
+    install_and_enable_unit coreos-NetworkManager.service network-online.target
```
jlebon, line 19: This should be `network.target`, no?

cgwalters:
Another highly important thing going on here is that we can drop the default kargs from coreos-assembler.  Hopefully everyone agress with that.

But, doing so will have a tricky interlock with ensuring RHCOS keeps building.  This gets back into the `features.yaml` thing so we can "ratchet" changes safely.

overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/module-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+    fetchd="$initdir/$systemdsystemunitdir/ignition-fetch.service.d/"
+    mkdir -p "${fetchd}"
+    cat >${fetchd}/nm-online.conf << EOF
+[Unit]
+After=network-online.target
+Wants=network-online.target
+EOF
```
jlebon: Can we drop this part? With the separate `ignition-fetch-offline.service`, we can make `ignition-fetch.service` hard require `network.target` (the tricky bit is still the "systemd always starts dependencies of units", but I'll work around that there).

cgwalters: Overall, feel free to push whatever changes you want to this PR if it helps you iterate!

overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/module-setup.sh
cgwalters: Mmm...debatable.  See https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/
I think what Ignition wants is much closer to `network-online.target`.

jlebon:
I cherry-pick the key hunks from #326 we need for the live ISO and opened the corresponding cosa patch: https://github.com/coreos/coreos-assembler/pull/1298

overlay.d/05core/usr/lib/dracut/modules.d/39coreos-ignitionnetwork/coreos-NetworkManager.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Network Manager (CoreOS initrd)
+Wants=network.target
+After=network-pre.target dbus.service
+Before=network.target network.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/nm-initrd-generator rd.neednet=1 ip=dhcp,dhcp6
```
lucab, line 10: I don't think this is going to work. I fear this is going to miss:
 * custom kargs for static networking (e.g. the ones injected by coreos-installer)
 * runtime kargs from `/etc/cmdline.d.` (e.g. the ones forwarded from vmware guestinfo)

cgwalters: > custom kargs for static networking (e.g. the ones injected by coreos-installer)

Right, we probably need to make this an `ExecStartPre=` that looks at `/proc/cmdline` and defers to that if any relevant arguments exist.

> runtime kargs from /etc/cmdline.d. (e.g. the ones forwarded from vmware guestinfo)

This only exists in https://github.com/coreos/afterburn/pull/379 right?

But hmm...either way, we should order `After=dracut-cmdline.service` and parse...`/tmp/networking_opts` it looks like?

lucab: Yes, the vmware guestinfo bits specifically are only in a not-yet-merged PR, but generally I think the `cmdline.d` primitive is a useful one to keep.

cgwalters:
Most of this is obsolete with the current afterburn handling; the core idea of running NM as a unit still stands, but yeah, better done upstream.

PR #322, opened 2020-03-26
overlay/20live: persist osmet files

jlebon:
We need to make sure that the osmet files in the initrd are kept
somewhere accessible in the real root where coreos-installer will find
them. Have a simple unit which copies them over to the right spot.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-live-persist-osmet.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-live-persist-osmet.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Persist osmet files
+DefaultDependencies=false
+ConditionPathExists=/run/ostree-live
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/mkdir -p /run/coreos-installer/osmet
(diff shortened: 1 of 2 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1244
Requires: https://github.com/coreos/coreos-installer/pull/187

cgwalters:
This will persistently take up RAM even if the user isn't going to do an install, right?

We could fix that by doing e.g. `lsinitrd --unpack` or something to get it instead.

jlebon:
We don't have access to the initrd in a live scenario though, right? (I mean, other than it already being loaded in RAM in the first place).  I guess in the live ISO case where the media isn't removed yet, you could use that, though for 5M seems cleaner to just have it work the same across all live scenarios.

jlebon:
One related thing is `/run/initramfs`, which systemd can use to go back to the initrd on shutdown. I don't think we make use of that in Fedora at least though, since the whole initrd filesystem is nuked on pivot: https://github.com/systemd/systemd/blob/6281c6e56c51f45edbf41a83fd13c9d290f6a691/src/shared/switch-root.c#L123.

cgwalters (approved):
OK yeah, not worth worrying about a few MB right now.

bgilbert (approved):

jlebon:
I just tweaked this slightly so that it doesn't error out if no osmet files are present. That way it's safe to merge this right now so that we can use it in the CI for https://github.com/coreos/coreos-installer/pull/187.

Feel free to do another review. Otherwise will merge on green!

PR #323, opened 2020-03-27
overrides: fasttrack coreos-installer 0.1.3-1.fc31.x86_64

bgilbert:


dustymabe (approved):

jamescassell:
Great, but why?

dustymabe:
> Great, but why?

so we can test the installer in our development stream for more time before it hits our testing stream and then our stable stream.

jamescassell:
Yes makes sense. I should have checked the branch.

jlebon:
Can you add a link to the Bodhi update? It's helpful for tracking when to remove the override (so long as it's humans doing that job).

Edit: accompanied by a short rationale. :)

PR #326, opened 2020-03-29
overlay/live: support booting from live ISO without networking

### LINKED ISSUE #349: Booting from ISO w/o DHCP server fails and ends in emergency mode
Hello!
I try to boot a KVM from _fedora-coreos-31.20200113.3.1-live.x86_64.iso_.
However, there's no DHCP service in the network, and therefore the boot shows many DHCPDISCOVER issues.
In the end the console is opening in emergency mode only.

I have created a screenshot documenting this.

In my opinion there's an issue booting from ISO w/o DHCP service available.
But I want to use a static IP and configure this with the ignition file accordingly.

I proceeded this way with CoreOS (_coreos_production_iso_image.iso_) w/o any problems.

Please advise how to fix this.

THX

dustymabe:
There is a scenario where the user wants to configure networking
after they get to the interactive bash prompt. Let's support this.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/349

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
@@ -1,0 +1,50 @@
+# This unit will run very early before the dracut-cmdline
+# service and detect if we want to request dracut bring up
+# networking or not. We do want to request networking if:
+#
+# - the user is booting the live ISO
+# - the user didn't already request networking via rd.neednet
+# - the user provided a ignition.config.url karg, implying
+#   the need for networking
+# - there is an embedded ignition config
+#
+# For the case of the embedded Ignition config there could be a
+# case where the user embeds an Ignition config (via coreos-installer
+# iso embed) but doesn't want networking. In that case we'll have
+# smarter detection in the future (https://github.com/coreos/fedora-coreos-tracker/issues/443)
+# but the user can override with `rd.neednet=0` now if needed.
+#
+# If we do determine we need network and there are no other
+# `ip=` kargs then we'll use `ip=dhcp,dhcp6` by default.
+#
+# The requesting of network will be done by writing relevant
+# dracut networking args into /etc/cmdline.d/coreos-live-network-kargs.conf
+# so that it gets picked up by the dracut networking scripts later
+# on in boot.
+#
+# This is all done because we want to support a mode where
+# the user can boot the live ISO and get to an interactive
+# prompt without requiring networking on boot. The user can
+# then configure the networking interactively.
+#
+[Unit]
+Description=Request live ISO networking
+DefaultDependencies=no
+Before=dracut-cmdline.service
+
+# Make sure we are in the initramfs. We are booted to the live
+# ISO and rd.neednet wasn't already passed somewhere else
+ConditionPathExists=/usr/lib/initrd-release
+ConditionKernelCommandLine=!rd.neednet
+ConditionKernelCommandLine=coreos.liveiso
+ConditionPathExists=/run/ostree-live
+
+# We'll assume we need network in either of these two following
+# cases (see description from above)
+ConditionKernelCommandLine=|ignition.config.url
+ConditionPathExists=|/config.ign
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/coreos-liveiso-network-kargs
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/bash
+
+# For a description of how this is used see coreos-liveiso-network-kargs.service
+
+# Load the dracut library for getarg
+source /usr/lib/dracut-lib.sh
+
+main() {
+
+    # If we're running this script we already know we need networking
+    # (determined by conditionals in the systemd unit).
+    echo 'info: Requesting networking in the initramfs'
+    echo 'rd.neednet=1' > /etc/cmdline.d/10-coreos-liveiso-network-kargs.conf
+
+    # If there is not already a ip= CLI arg use ip=dhcp,dhcp6 as default.
+    if ! getarg 'ip' &>/dev/null; then
+        echo 'info: using ip=dhcp,dhcp6 default networking configuration'
+        echo 'ip=dhcp,dhcp6' >> /etc/cmdline.d/10-coreos-liveiso-network-kargs.conf
+    else
+        echo 'info: using already set karg ip= information'
+    fi
+}
+
+main
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-reconfigure-nm-wait-online.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-reconfigure-nm-wait-online.service
@@ -1,0 +1,23 @@
+# Configure NetworkManager-wait-online in the real root for the
+# Live ISO to timeout quicker and also not explicitly fail since
+# booting the Live ISO without network is a valid use case.
+#
+# Doing this improves the user experience when booting the
+# Live ISO without network.
+
+[Unit]
+Description=Reconfigure NetworkManager-wait-online service
+DefaultDependencies=no
+# Make sure we are in the initramfs and we are booted to the live ISO
+ConditionPathExists=/usr/lib/initrd-release
+ConditionKernelCommandLine=coreos.liveiso
+ConditionPathExists=/run/ostree-live
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+# Note keep this in sync with NetworkManager-wait-online.service
+# Right now we are keeping the same ExecStart but we are making it
+# OK to fail (`-`) and timeout sooner (5 seconds vs 30).
+ExecStartPre=/usr/bin/mkdir -p /run/systemd/system/NetworkManager-wait-online.service.d
+ExecStart=/bin/bash -c 'echo -e "[Service]\nExecStart=\nExecStart=-/usr/bin/nm-online -s -q --timeout=5" > /run/systemd/system/NetworkManager-wait-online.service.d/liveiso.conf'
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
@@ -36,4 +36,7 @@ install() {
 
     inst_simple "$moddir/coreos-liveiso-network-kargs.service" \
         "$systemdsystemunitdir/coreos-liveiso-network-kargs.service"
+
+    inst_simple "$moddir/coreos-liveiso-reconfigure-nm-wait-online.service" \
+        "$systemdsystemunitdir/coreos-liveiso-reconfigure-nm-wait-online.service"
 }
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
@@ -30,4 +30,10 @@ install() {
 
     inst_simple "$moddir/writable.mount" \
         "$systemdsystemunitdir/writable.mount"
+
+    inst_simple "$moddir/coreos-liveiso-network-kargs.sh" \
+        "/usr/sbin/coreos-liveiso-network-kargs"
+
+    inst_simple "$moddir/coreos-liveiso-network-kargs.service" \
+        "$systemdsystemunitdir/coreos-liveiso-network-kargs.service"
 }
--- a/live/EFI/fedora/grub.cfg
+++ b/live/EFI/fedora/grub.cfg
@@ -28,6 +28,6 @@ set timeout=5
 
 ### BEGIN /etc/grub.d/10_linux ###
 menuentry 'Fedora CoreOS (Live)' --class fedora --class gnu-linux --class gnu --class os {
-	linux /images/vmlinuz @@KERNEL-ARGS@@ rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
+	linux /images/vmlinuz @@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal
 	initrd /images/initramfs.img
 }
--- a/live/isolinux/isolinux.cfg
+++ b/live/isolinux/isolinux.cfg
@@ -67,7 +67,7 @@ label linux
   menu label ^Fedora CoreOS (Live)
   menu default
   kernel /images/vmlinuz
-  append initrd=/images/initramfs.img @@KERNEL-ARGS@@ rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
+  append initrd=/images/initramfs.img @@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal
 
 menu separator # insert an empty line
 
--- a/live/zipl.prm
+++ b/live/zipl.prm
@@ -1,1 +1,1 @@
-@@KERNEL-ARGS@@ rd.neednet=1 ip=dhcp ignition.firstboot ignition.platform.id=metal
+@@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -27,6 +27,8 @@ add_requires sysroot.mount     initrd-root-fs.target
 add_requires sysroot-etc.mount initrd-root-fs.target
 add_requires sysroot-var.mount initrd-root-fs.target
 
+add_requires coreos-liveiso-network-kargs.service              initrd.target
+
 mkdir -p "${UNIT_DIR}/ostree-prepare-root.service.d"
 cat > "${UNIT_DIR}/ostree-prepare-root.service.d/10-live.conf" <<EOF
 # With live PXE there's no ostree= argument on the kernel command line, so
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -28,6 +28,7 @@ add_requires sysroot-etc.mount initrd-root-fs.target
 add_requires sysroot-var.mount initrd-root-fs.target
 
 add_requires coreos-liveiso-network-kargs.service              initrd.target
+add_requires coreos-liveiso-reconfigure-nm-wait-online.service initrd.target
 
 mkdir -p "${UNIT_DIR}/ostree-prepare-root.service.d"
 cat > "${UNIT_DIR}/ostree-prepare-root.service.d/10-live.conf" <<EOF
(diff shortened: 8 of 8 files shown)


cgwalters:
This looks sane but it heavily overlaps with https://github.com/coreos/fedora-coreos-config/pull/321 right?

dustymabe:
> This looks sane but it heavily overlaps with #321 right?

correct. I'm in the middle of writing up a response to #321. I wanted to post up this WIP first to show what my alternative proposal was.

cgwalters:
I guess #321 is trying to be more ambitious.

But, you're missing an important case - this needs to support `coreos-installer iso embed`, see discussion in https://github.com/coreos/ignition-dracut/pull/161

Something like
```
ConditionPathExists=/config.ign
```

(Edit, the `!` was wrong)

dustymabe:
> I guess #321 is trying to be more ambitious.
> 
> But, you're missing an important case - this needs to support `coreos-installer iso embed`, see discussion in [coreos/ignition-dracut#161](https://github.com/coreos/ignition-dracut/pull/161)
> 
> Something like
> 
> ```
> ConditionPathExists=!/config.ign
> ```


Is there a case where someone would provide both `ignition.config.url` on the kernel command line and also embed an ignition config using `coreos-installer iso embed`?

cgwalters:
> Is there a case where someone would provide both ignition.config.url on the kernel command line and also embed an ignition config using coreos-installer iso embed?

I can't think of a good use case for that, feels like we can leave it as unspecified behavior.

bgilbert:
Ignition prioritizes the karg over `user.ign`.

bgilbert:
Makes sense to me.  Given the `config.ign` case, I think we'd want:

```
ConditionPathExists=/usr/lib/initrd-release
ConditionKernelCommandLine=!ip
ConditionKernelCommandLine=coreos.liveiso

ConditionKernelCommandLine=|ignition.config.url
ConditionPathExists=|/config.ign
```

dustymabe:
> Makes sense to me. Given the `config.ign` case, I think we'd want:

So in this scenario you are assuming that `config.ign` could have remote references and that we need network? I was really hoping to keep the "do we need network" logic down to just a dumb check for if `ignition.config.url=` exists. The problem with the embedded `config.ign` is that I could see a case where someone embedded a config because they didn't have network (offline installer) and also a case where they embedded `config.ign` just because it was convenient to not to have to catch the boot prompt and add a karg.

jlebon:
WDYT about https://github.com/coreos/ignition/pull/956 instead of this? This would solve the generic conditional networking problem.

Though we could get something like this in as a short-term fix too if there isn't consensus on the approach taken there.

jlebon:
Cross-linking https://github.com/coreos/ignition/pull/956#issuecomment-606229675. I sanity checked that the live ISO now boots fully offline. (Haven't tried with an embedded Ignition config that pulls in networking yet, but that should Just Work.)

cgwalters:
> (Haven't tried with an embedded Ignition config that pulls in networking yet, but that should Just Work.)

That's what `kola testiso` does!

dustymabe:
OK i'm going to revive this PR and move it forward. One thing I'm thinking about doing here is leave `ip=dhcp` on the kernel command line and just remove the `rd.neednet=1` so that the `coreos-liveiso-network-kargs.service` only adds back `rd.neednet=1`.  I'm not sure if that is better or worse than what I'm currently doing.

Thoughts?

dustymabe:
> OK i'm going to revive this PR and move it forward. One thing I'm thinking about doing here is leave `ip=dhcp` on the kernel command line and just remove the `rd.neednet=1` so that the `coreos-liveiso-network-kargs.service` only adds back `rd.neednet=1`. I'm not sure if that is better or worse than what I'm currently doing.
> 
> Thoughts?

I think the argument for doing this is incase someone provided `ip=` kargs but forgot to specify `rd.neednet=1` themselves.

cgwalters:
> I think the argument for doing this is incase someone provided ip= kargs but forgot to specify rd.neednet=1 themselves.

Doesn't seem like a strong argument to me.  Using `rd.neednet=1` is very well documented.  And I think the elegance of reducing our default kernel arguments wins out.

dustymabe:
> > I think the argument for doing this is incase someone provided ip= kargs but forgot to specify rd.neednet=1 themselves.
> 
> Doesn't seem like a strong argument to me. Using `rd.neednet=1` is very well documented. And I think the elegance of reducing our default kernel arguments wins out.

A couple of us stayed longer in the open discussion today and came up with a way to get the behavior we wanted without having leaving `ip=` on the kernel command line in the isolinux or grub configs. I'll post up an update shortly.

dustymabe:
ok - marking this as ready for review. The logic and the reasoning is captured in this description:

```
# This unit will run very early before the dracut-cmdline
# service and detect if we want to request dracut bring up
# networking or not. We do want to request networking if:
#
# - the user is booting the live ISO
# - the user didn't already request networking via rd.neednet
# - the user provided a ignition.config.url karg, implying
#   the need for networking
# - there is an embedded ignition config
#
# For the case of the embedded Ignition config there could be a
# case where the user embeds an Ignition config (via coreos-installer
# iso embed) but doesn't want networking. In that case we'll have
# smarter detection in the future (https://github.com/coreos/fedora-coreos-tracker/issues/443)
# but the user can override with `rd.neednet=0` now if needed.
#
# If we do determine we need network and there are no other
# `ip=` kargs then we'll use `ip=dhcp,dhcp6` by default.
#
# The requesting of network will be done by writing relevant
# dracut networking args into /etc/cmdline.d/coreos-live-network-kargs.conf
# so that it gets picked up by the dracut networking scripts later
# on in boot.
#
# This is all done because we want to support a mode where
# the user can boot the live ISO and get to an interactive
# prompt without requiring networking on boot. The user can
# then configure the networking interactively.

```

In this I also added a unit to change NetworkManager-wait-online.service in the real root to not show a failure if it can't get the network and also to timeout in 5 vs 30 seconds. This is done to improve the user experience of a user that boots the Live ISO without networking.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-reconfigure-nm-wait-online.service
```diff
@@ -0,0 +1,23 @@
+# Configure NetworkManager-wait-online in the real root for the
+# Live ISO timeout quicker and also not explicitly fail since
+# booting the Live ISO without network is a valid use case.
```
cgwalters, line 3: I think we should support an explicit API to turn off networking instead of this.

Basically all we need is something like:

`coreos-installer iso embed --no-initramfs-network`
which would also create `/etc/coreos-install-nonet` in the cpio archive along with `/config.ign`.

Then we'd just have a:

```
ConditionPathExists=!/etc/coreos-install-nonet
```

in the other service above.

jlebon: Note this configures the service in the real root, not in the initrd.

I think we should match what the Fedora live ISO does here, which is to opportunistically set up networking if available, but otherwise not fail. Was playing around with that in the Fedora 32 Workstation live ISO in both a VM without any network adapters and one connected to an isolated network and `NetworkManager-wait-online.service` worked just fine either way. I don't see any timeout overrides or configuration tweaks. But clearly, *something* is different.

cgwalters: > Note this configures the service in the real root, not in the initrd.

Oh, right.

> I think we should match what the Fedora live ISO does here, which is to opportunistically set up networking if available, but otherwise not fail.

Hmm.  OK yeah I guess so.

cgwalters (approved):
OK, this looks good to me!  I assume you tested it manually?

It shouldn't be terribly hard to add `kola testiso --offline` that starts the VM with no NICs at all; I actually did [this recent PR](https://github.com/coreos/coreos-assembler/pull/1307) with that in mind because we can use virtio-channels to talk to a VM with no NIC.

dustymabe:
thanks @cgwalters! Yep, I've been testing this over and over today. Note that I'm testing this by setting up libvirt network with no DHCP and then running nmtui (future PR) once I get the system up to configure a static IP. Here's the libvirt network XML I'm using:

```
<network>
  <name>nodhcp</name>
  <forward mode="nat">
    <nat>
      <port start="1024" end="65535"/>
    </nat>
  </forward>
  <bridge name="virbr100" stp="on" delay="0"/>
  <ip address="192.168.130.1" netmask="255.255.255.0">
  </ip>
</network>
```

I'll have to checkout `kola testiso`. 

@jlebon if you have any more suggestions on what I can change to make it more like the Live ISO let me know. When you boot the Fedora live ISO does it timeout before you're able to get to a console? Maybe you just don't see the timeout/failure because it's a GUI interface?

dustymabe:
Since I know he had voiced earlier concerns about this I talked with @arithx earlier today and he said he was good with the following logic:

```
# For the case of the embedded Ignition config there could be a
# case where the user embeds an Ignition config (via coreos-installer
# iso embed) but doesn't want networking. In that case we'll have
# smarter detection in the future (https://github.com/coreos/fedora-coreos-tracker/issues/443)
# but the user can override with `rd.neednet=0` now if needed.
```

overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
```diff
@@ -30,4 +30,13 @@ install() {
...
         "$systemdsystemunitdir/writable.mount"
+
+    inst_simple "$moddir/coreos-liveiso-network-kargs.sh" \
+        "/usr/sbin/coreos-liveiso-network-kargs"
+
+    install_and_enable_unit "coreos-liveiso-network-kargs.service" \
+        "initrd.target"
+
+    install_and_enable_unit "coreos-liveiso-reconfigure-nm-wait-online.service" \
+        "initrd.target"
```
jlebon: One suggestion here: there's a generator already for the live ISO, so we could instead dynamically enable this. An advantage of doing that is that systemd doesn't even bother with the service unit and doesn't spam the "Skipped ..." message on every boot on all other platforms. Definitely not a blocker of course. :)

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
```diff
@@ -0,0 +1,50 @@
...
+#
+# The requesting of network will be done by writing relevant
+# dracut networking args into /etc/cmdline.d/coreos-live-network-kargs.conf
+# so that it gets picked up by the dracut networking scripts later
+# on in boot.
+#
+# This is all done because we want to support a mode where
+# the user can boot the live ISO and get to an interactive
+# prompt without requiring networking on boot. The user can
+# then configure the networking interactively.
```
jlebon, line 28: Awesome documentation!

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
```diff
@@ -0,0 +1,50 @@
...
+#
+[Unit]
+Description=conditionally add networking kargs for live ISO
```
jlebon: Maybe "Request live ISO networking" ?

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
```diff
@@ -0,0 +1,28 @@
...
+    # If we're running this script we already know we need networking
+    # (determined by conditionals in the systemd unit).
+    echo 'info: Requesting networking in the initramfs'
+    echo 'rd.neednet=1' > /etc/cmdline.d/10-coreos-liveiso-network-kargs.conf
+
+    # If there is not already a ip= CLI arg use ip=dhcp,dhcp6 as default.
+    # Note that because of the priority of /etc/cmdline.d/*.conf it doesn't
+    # matter if we do this check or if we unconditionally write ip=dhcp,dhcp6
+    # because it will never take precedence over an ip= arg on the kernel
+    # command line.
```
jlebon: Hmm, not sure I follow this... doesn't that mean we don't need to check for `ip` at all then and can just unconditionally print `ip=dhcp,dhcp6`?

jlebon (commented):
> @jlebon if you have any more suggestions on what I can change to make it more like the Live ISO let me know. When you boot the Fedora live ISO does it timeout before you're able to get to a console? Maybe you just don't see the timeout/failure because it's a GUI interface?

I add `console=ttyS0` and attach to the VM console so I can see the logs. And yeah it's weird, it doesn't time out, it just... succeeds. Doesn't even seem to take more than a few seconds either. I'll have to poke around more on this, but anyway, I don't think it's a blocker!

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
dustymabe: Does this message show up in the logs even if it doesn't run? If so then I'd like to keep some words in there that indicates it is conditional

dustymabe:
> > @jlebon if you have any more suggestions on what I can change to make it more like the Live ISO let me know. When you boot the Fedora live ISO does it timeout before you're able to get to a console? Maybe you just don't see the timeout/failure because it's a GUI interface?
> 
> I add `console=ttyS0` and attach to the VM console so I can see the logs. And yeah it's weird, it doesn't time out, it just... succeeds. Doesn't even seem to take more than a few seconds either. I'll have to poke around more on this, but anyway, I don't think it's a blocker!

This is on a system with a network without DHCP? See https://github.com/coreos/fedora-coreos-config/pull/326#issuecomment-608197359 for how I'm setting up a libvirt network to test this. On the Fedora Live ISO what does `systemctl cat NetworkManager-wait-online.service` show. I wonder if they are doing some overrides.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
dustymabe: thanks :)

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
jlebon: Yeah, it'll print:

> Condition check resulted in <descrition> being skipped.

> If so then I'd like to keep some words in there that indicates it is conditional

Is it though? Once the unit is activated, we're definitely activating networking, no?

overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
dustymabe: I hadn't thought of that. If it's just an extra message in the log I think I'd like to keep the unit separate and prevent yet another heredoc in the generator and also retesting this all again.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
dustymabe: ~~yep, basically because of the way /etc/cmdline.d/*.conf gets merged with the kernel command line we could unconditionally write `ip=dhcp,dhcp6` and the right thing would still happen but I think it would be misleading. IMHO future developers who come in here and try to figure out what is going on are better off with the current conditional logic and comment.~~

scratch all of that. I now think it *does* matter because it's valid for a user to provide multiple `ip=` kargs so if we unconditionally add it then we'd end up with something like 

```
ip=dhcp,dhcp6 ip=192.168.130.2::192.168.130.1:255.255.255.0:fcos:eth0:none:192.168.130.1
```

that dracut/NM would then parse and that's not what we want.

I'll delete all but the first line of the comment if you agree.

jlebon:
> This is on a system with a network without DHCP? See #326 (comment) for how I'm setting up a libvirt network to test this.

Ahh yup, good catch! I tested no network adapters, and with a network adapter isolated, but still with DHCP and those worked fine. An isolated network adapter without DHCP does indeed cause `NetworkManager-wait-online.service` to time out and error out after 30s.

overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
jlebon: Right, definitely don't want heredocs either. I meant more doing the `ln -s` in the generator. See e.g. https://github.com/coreos/ignition-dracut/blob/6136be3d9d38d7926a61cd4d1b4ba5f9baf0892f/dracut/30ignition/ignition-generator#L39-L40. Anyway, as is is fine with me too!

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
jlebon: OK, yup that makes sense!

overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
dustymabe: ahh I just realized you were referring to both units and not just the `reconfigure-nm-wait-online` one. ok let me look into that.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
dustymabe: > Is it though? Once the unit is activated, we're definitely activating networking, no?

It think that is a "no" answer to my question: "Does this message show up in the logs even if it doesn't run?" If that is the case I'm 👍 to changing the wording.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
dustymabe: see.. if I hadn't try to "do the right thing" and also comment the hell out of this (even though the comment was wrong) I would have unconditionally added `ip=dhcp,dhcp6` 😜

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
jlebon: Ahhh hehe, the GitHub markdown elided the critical part of my message. It prints:

> Condition check resulted in $descrition being skipped.

What I mean is that all the conditionals are part of the systemd unit itself already, so whether the unit is skipped or not is already reflected in what systemd does (and prints). Since the script itself unconditionally enables `rd.neednet`, it'd be cleaner to say something like "Request live ISO networking".

overlay.d/05core/usr/lib/dracut/modules.d/20live/module-setup.sh
dustymabe: so right now the add_requires in the live generator for live makes them all a requirement of `initrd-root-fs.target`. Currently I had made them a target of `initrd.target`. I could modify the `add_requires()` function like is done in other generators or I could just make them required by `initrd-root-fs.target`. Thoughts?

jlebon: Yeah, I think enhancing `add_requires` like in Ignition to allow specifying the target makes sense (and keeping the new units to `initrd.target`).

dustymabe:
ok pushed up a change to address code review comments. Once you say it looks good I'll give a final round of testing.

jlebon (approved):
Nice, LGTM!

dustymabe:
rebased on top of latest testing-devel - now doing final testing

dustymabe:
ok pushed one final commit to fix a problem introduced in the last change.. merging

PR #329, opened 2020-04-03
manifests: add NetworkManager-tui

dustymabe:
`nmtui` allows for a user to interactively set up networking. Using Fedora
CoreOS interactively is usually not what we encourage with Fedora CoreOS, relying
on automation instead. The one exception to this is the Live ISO environment
where our current documented workflow is for a user to boot the Live ISO and
run coreos-installer interactively (of course, there is an automation workflow
as well). In this flow it would be useful for a user to be able to have an
interactive tool that helped them set up networking if they need static networking.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -117,6 +117,8 @@ packages:
   - NetworkManager-team teamd
   # Static firewalling
   - iptables nftables iptables-nft iptables-services
+  # Interactive Networking configuration during coreos-install
+  - NetworkManager-tui
   # Storage
   - cloud-utils-growpart
   - lvm2 iscsi-initiator-utils sg3_utils


cgwalters:
We can install it only on the Live ISO, but it would require more work on the buildsystem side.

dustymabe:
For some background see: https://github.com/coreos/fedora-coreos-tracker/issues/427

Current depchain:

```
Added:
  NetworkManager-tui-1:1.20.10-1.fc31.x86_64
  newt-0.52.21-2.fc31.x86_64
  slang-2.3.2-6.fc31.x86_64
```

dustymabe:
> We can install it only on the Live ISO, but it would require more work on the buildsystem side.

yep

andrewdunndev:
It'd be nice to have this. As someone coming over from systemd-networkd (realizing that in [this issue](https://github.com/coreos/fedora-coreos-tracker/issues/24) you've decided to remove it)... having more assistance for generating keyfiles would be welcome.

jlebon (approved):
Makes sense to me!

PR #332, opened 2020-04-07
overlay/live-generator: Avoid overlayfs on squashfs in RHEL8

cgwalters:
It looks like the `squashfs` in RHEL8 is too old to support
being used as a "lowerdir" for `overlayfs`, which is too bad
because the two go together really well.  Hack around this
by copying `/etc` instead.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -119,14 +119,42 @@ Before=ignition-mount.service
 EOF
 }
 
-common_etcvar_unit > "${UNIT_DIR}/sysroot-etc.mount"
-cat >>"${UNIT_DIR}/sysroot-etc.mount" <<EOF
+supports_squashfs_overlayfs=1
+case "$(uname -r)" in
+    4.18.*) supports_squashfs_overlayfs=0
+esac
+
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -119,14 +119,43 @@ Before=ignition-mount.service
...
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/cp -a /sysroot/etc /writable/etc-copy
+EOF
+
+    common_etcvar_unit > "${UNIT_DIR}/sysroot-etc.mount"
+cat >>"${UNIT_DIR}/sysroot-etc.mount" <<EOF
+After=sysroot-etc-copy.service
+Requires=sysroot-etc-copy.service
+Before=ignition-ostree-populate-var.service
```
bgilbert: Why is this needed here but not in the other branch?

cgwalters: I think this is leftover cruft from when I was confused and thought we were doing an overlayfs for `/var` too.  Testing removing it...

cgwalters: OK &lt;some distractions later&gt; yep removing that works fine!

PR #335, opened 2020-04-07
[bodhi-updates] manifest: drop fedora-coreos-pool repo

jlebon:
So... one thing I just realized is that now that we have a bunch of f32
packages in the pool, we really can't have `bodhi-updates` picking up
packages from there unlocked.

So this is sort of forcing our hand to get
https://github.com/coreos/fedora-coreos-tracker/issues/293 fixed very
soon so that we can unblock package bumps again.

As a first step and to prevent `testing-devel` ending up with a bunch of
f32 packages, let's remove the pool from the list of repos here.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -13,9 +13,6 @@ repos:
   - fedora-updates
   - fedora-modular
   - fedora-updates-modular
-  # We don't actually want this here... it's only until we update the lockfile a
-  # better way, see: https://github.com/coreos/fedora-coreos-tracker/issues/293
-  - fedora-coreos-pool
 
 add-commit-metadata:
   fedora-coreos.stream: bodhi-updates


dustymabe (approved):
LGTM

jlebon:
And ahhh yup, bodhi-updates is failing now. The reason is that the python-less crypto-policies only exists in f32 and we have [an `exclude` on `python`](https://github.com/coreos/fedora-coreos-config/pull/280).

One thing we could do is have an FCOS-only f31 version of that crypto-policies package with the Python-less patch backported. (That's probably what we would've done anyway instead of pulling in the f32 version directly if it had other changes incompatible with f31). Another approach is enhancing lockfiles to allow cherry-picking only some packages from a specific repo.

Or maybe... once we have https://github.com/coreos/fedora-coreos-tracker/issues/293, we could just get rid of `bodhi-updates` entirely? There's also `bodhi-updates-testing` though, which I think could still be useful to have.

Anyway, we can keep discussing this, but for now I'm just going to merge this to prevent `testing-devel` from inheriting f32 packages.

dustymabe:
with https://github.com/coreos/rpm-ostree/pull/1847 we should just be able to link directly to koji urls for packages as a short term workaround for this.

PR #336, opened 2020-04-08
[testing] use old version of coreos-pool yum repo

dustymabe:
We need this because the f32 packages from next-devel that are now
in coreos-pool are causing packages to leak into our f31 based Fedora
CoreOS builds. Use a version of the yum repo from before the packages
were tagged in for now.

### DIFF
--- a/fedora-coreos-pool.repo
+++ b/fedora-coreos-pool.repo
@@ -1,6 +1,6 @@
 [fedora-coreos-pool]
 name=Fedora coreos pool repository - $basearch
-baseurl=https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/latest/$basearch/
+baseurl=https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/1427579/$basearch/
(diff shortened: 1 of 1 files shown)


jlebon (approved):
```
$ dnf repoquery libbrotli libcurl --repofrompath=pool,https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/1427579/x86_64/ --disablerepo '*' --enablerepo pool  --refresh
Added pool repo from https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/1427579/x86_64/
pool                                                                      795 kB/s |  27 MB     00:35
Last metadata expiration check: 0:00:10 ago on Wed Apr  8 09:33:49 2020.
libcurl-0:7.65.3-3.fc30.x86_64
libcurl-0:7.65.3-4.fc30.x86_64
libcurl-0:7.66.0-1.fc31.x86_64
```

So it's definitely old enough. CI will verify that it's new enough to contain all the packages this testing release needs.

PR #338, opened 2020-04-08
overlay: Introduce /run/bin

cgwalters:
See: https://github.com/coreos/fedora-coreos-tracker/issues/354

Basically we want to support things like containers that contain
binaries designed to execute on the host.  These should really
be "lifecycle bound" to the container image.  Let's at least
have an obvious place to drop them that goes away on reboot.

### DIFF
--- a/overlay.d/05core/usr/lib/tmpfiles.d/coreos-run-bin.conf
+++ b/overlay.d/05core/usr/lib/tmpfiles.d/coreos-run-bin.conf
@@ -1,0 +1,3 @@
+# See https://github.com/coreos/fedora-coreos-tracker/issues/354
+# TL;DR we want people to be able to run *transient* binaries from the host context
+d /run/bin 0755 - - -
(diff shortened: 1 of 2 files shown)


jlebon:
Hmm, I think it's worth fleshing out https://github.com/coreos/fedora-coreos-tracker/issues/354 some more first?

I'm still not sure if we're talking about a host mechanism which knows how to fetch binaries from containers and run them, or just having privileged containers place them in `/run` on start and e.g. use `systemctl` to start them. But I might be missing context.

lucab:
If we really want to go for this, I fear this has to plumbed a bit deeper inside the Fedora ecosystem as I suspect we'd need matching SELinux labels/policies too.

cgwalters:
Yeah, fair enough.  So I then lean a bit more towards [live "underlays"](https://github.com/ostreedev/ostree/pull/2103).

PR #339, opened 2020-04-08
overlay: Add code to verify interactive login for Live ISO

cgwalters:
This is similar to: https://github.com/coreos/ignition-dracut/pull/146

For our test system, it generally works really well to inject
things via Ignition.  That PR was about handling failures in the
initramfs *before* Ignition runs.

This PR is trying to help us test the scenario where no Ignition
is injected into the Live ISO.  Let's also use the virtio-channel
approach.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/coreos-liveiso-success.service
+++ b/overlay.d/05core/usr/lib/systemd/system/coreos-liveiso-success.service
@@ -1,0 +1,30 @@
+# This is used by our test framework in coreos-assembler
+# since for the "live ISO without Ignition" case we
+# don't have an easy way to test it.
+[Unit]
+Description=CoreOS Live ISO virtio success
+Documentation=https://github.com/coreos/fedora-coreos-config
+# Only run on the Live ISO, and only if there's no Ignition config;
+# the second bits here invert the conditionals in coreos-liveiso-network-kargs.service.
+ConditionKernelCommandLine=coreos.liveiso
+ConditionPathExists=!/config.ign
+ConditionVirtualization=|kvm
+ConditionVirtualization=|qemu
+# Start running late to help ensure that the below conditional works
+After=systemd-user-sessions.service
+ConditionPathExists=/dev/virtio-ports/coreos.liveiso-success
+
+[Service]
+Type=simple
+# https://stackoverflow.com/questions/44358723/systemd-unit-file-problems-with-tr
+IgnoreSIGPIPE=false
+# See https://cgit.freedesktop.org/systemd/systemd/plain/src/systemd/sd-messages.h for the MESSAGE_ID source.
+# The logic here is that we're doing a streaming journalctl query (-f to follow)
+# and the `| head` bit will cause the pipeline to wait until at least one line is
+# emitted, which will happen when a user login starts.  We then just write a static
+# message to the virtio channel, which https://github.com/coreos/coreos-assembler/pull/1330
+# knows how to read.
+ExecStart=/bin/sh -c 'journalctl -b -q -f --no-tail -o cat -u systemd-logind.service MESSAGE_ID=8d45620c1a4348dbb17410da57c60c66 | head -1; echo coreos-liveiso-success > /dev/virtio-ports/coreos.liveiso-success'
+
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -13,3 +13,5 @@ enable afterburn-firstboot-checkin.service
 enable afterburn-sshkeys@.service
 # Update agent
 enable zincati.service
+# Testing aid
+enable coreos-liveiso-success.service


dustymabe:
Will this new service run on every boot when booted in kvm/qemu ? Can we make it only run when we're doing testing? The reason I ask is because I don't understand the implications of writing to /dev/virtio-ports/coreos.liveiso-success if we're not testing.

overlay.d/05core/usr/lib/systemd/system/coreos-liveiso-success.service
```diff
@@ -0,0 +1,17 @@
...
+Description=CoreOS Live ISO virtio success
+Documentation=https://github.com/coreos/fedora-coreos-config
+ConditionPathExists=/run/ostree-live
+ConditionVirtualization=|kvm
+ConditionVirtualization=|qemu
+
+[Service]
+Type=simple
+# See https://cgit.freedesktop.org/systemd/systemd/plain/src/systemd/sd-messages.h
+ExecStart=/bin/sh -c 'journalctl -b -o cat -u systemd-logind.service MESSAGE_ID=8d45620c1a4348dbb17410da57c60c66 | head -1 >dev/null; echo coreos-liveiso-success > /dev/virtio-ports/coreos.liveiso-success'
```
dustymabe: what's the point of the `| head -1 >/dev/null` ?

cgwalters:
> Will this new service run on every boot when booted in kvm/qemu ? 

Only on the Live ISO, and actually we could also make it run on only if no Ignition config is provided; will do.

> Can we make it only run when we're doing testing?
 
The rationale for this PR is precisely that that's a bit tricky to do.  Now of course, there's a lot of other things we could do that would be a lot more invasive, like actually copying the ISO and editing the kernel cmdline, but...

>The reason I ask is because I don't understand the implications of writing to /dev/virtio-ports/coreos.liveiso-success if we're not testing.

Well...we'll create a file named that if the port doesn't exist...will fix that.  Otherwise, there's not much that's going to happen.

overlay.d/05core/usr/lib/systemd/system/coreos-liveiso-success.service
cgwalters: It delays until `journalctl` prints at least one line.

cgwalters: Oops, though we're missing `-f` which makes this incorrectly pass.

jlebon:
Since this is in the real root, could we just do `ConditionPathExists=/dev/virtio-ports/coreos.liveiso-success`?

cgwalters:
An entirely different approach would be to add a general mechanism that allows the host to inject arbitrary code into the initramfs.  But eh, so far there are just two of these things and I can't imagine growing many more, because again "inject arbitrary code" is really what Ignition is anyways, we're just trying to cover the few !Ignition cases we have.

jlebon:
Hmm, we're going to hit a very similar situation for testing https://github.com/coreos/ignition/pull/958#pullrequestreview-389514697 too.

I think this is 80% of the way to implementing a streaming journal log as discussed in https://github.com/coreos/ignition-dracut/pull/146 (but just scoped to real root for now). WDYT?

cgwalters:
> I think this is 80% of the way to implementing a streaming journal log as discussed in coreos/ignition-dracut#146 (but just scoped to real root for now). WDYT?

Strong preference here for https://github.com/coreos/fedora-coreos-config/pull/339#issuecomment-611181965 instead; or a variant of it that's basically something like:

```
[Unit]
After=ignition-complete.target
Before=initrd-switch-root.target
ConditionVirtualization=|kvm
ConditionVirtualization=|qemu
Wants=/dev/virtio-ports/coreos-qemu-code-injection

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=cp < /dev/virtio-ports/coreos-qemu-code-injection > /run/injected && chmod a+x /run/injected && /run/injected
```

This maximizes flexibility for the test framework.

jlebon:
Yeah definitely more flexible though... are there security concerns here? I feel a bit uneasy adding a code injection path by design.

cgwalters:
> are there security concerns here? I feel a bit uneasy adding a code injection path by design.

We already read and execute code from the Ignition qemu mechanisms (`-fw_cfg` and the newer block device); anyone who can control the qemu commandline can control the guest (e.g. switching to direct kernel booting).    

Things like [AMD SEV](https://developer.amd.com/sev/) I think try to address hypervisor attacks on guests but I think that's more a "line of defense" (think e.g. some kernel flaw that allows userspace an arbitrary process read gadget), and not a magic tool.

[enarx](https://github.com/enarx/enarx/wiki/Enarx-Introduction) is more in the "protect code from the hypervisor" game.

overlay.d/05core/usr/lib/systemd/system/coreos-liveiso-success.service
dustymabe: thanks for the explanation here. I really think we should add that information into the code. It's really not obvious. basically the -f says to follow and the head blocks until it gets one line of output. That's really useful for someone coming in and trying to understand what's going on here.

bgilbert:
@cgwalters https://github.com/coreos/fedora-coreos-config/pull/339#issuecomment-611198212 is all true, but I'm still uneasy about adding a new mechanism to inject arbitrary code.  Those other mechanisms are maintained, documented, and have some degree of auditability.  The code injection port would be obscure, undocumented, and basically unmaintained, and such mechanisms have been abused before.

cgwalters:
@bgilbert what's your suggested alternative then?  journal streaming?  just leave all this stuff untested?

bgilbert:
If I understand correctly that the test under consideration just checks whether https://github.com/coreos/ignition/pull/958 reports that a config wasn't provided, I'd rather see that code untested than add code injection to the OS so we can test it.  If journal streaming is sufficient to test it, then that SGTM.

This PR itself LGTM though.

cgwalters:
Though one aspect to this injection is that specifically for ISOs, the entire filesystem is read-only.  

For regular qcow2/non-live qemu scenarios it's not too hard to create a CoW image and inject code into the target OSTree; the main annoying trap there is (wait for it) SELinux labeling, but we can write the tools to address that.  That's probably fine for testing things like https://github.com/coreos/ignition/pull/958#pullrequestreview-389514697

So we may end up with just this one and the ignition-dracut one.

cgwalters:
Oh wow gah, this test only works when I pass `console=ttyS0` to debug it...

cgwalters:
OK whee, now this *actually* works.  That missing `-f` before made it spuriously pass :cry: which was a copy-paste error from when I was testing the one-liner interactively before moving it to the systemd unit.

overlay.d/05core/usr/lib/systemd/system/coreos-liveiso-success.service
```diff
@@ -0,0 +1,30 @@
...
+Type=simple
+# https://stackoverflow.com/questions/44358723/systemd-unit-file-problems-with-tr
+IgnoreSIGPIPE=false
+# See https://cgit.freedesktop.org/systemd/systemd/plain/src/systemd/sd-messages.h for the MESSAGE_ID source.
+# The logic here is that we're doing a streaming journalctl query (-f to follow)
+# and the `| head` bit will cause the pipeline to wait until at least one line is
+# emitted, which will happen when a user login starts.  We then just write a static
+# message to the virtio channel, which https://github.com/coreos/coreos-assembler/pull/1330
+# knows how to read.
+ExecStart=/bin/sh -c 'journalctl -b -q -f --no-tail -o cat -u systemd-logind.service MESSAGE_ID=8d45620c1a4348dbb17410da57c60c66 | head -1; echo coreos-liveiso-success > /dev/virtio-ports/coreos.liveiso-success'
```
jlebon, line 27: Instead of `head -1`, couldn't we do `journalctl -n 1` instead? Hmm, and once we do that, I think we could boil it down to

```
ExecStart=/usr/bin/journalctl ...
StandardOutput=file:/dev/virtio-ports/coreos.liveiso-success
```

And then kola could also check that it's e.g. the `core` user that was logged in.

jlebon (approved):
Had a chat with Colin about this. Long-term we might be better served by moving to a "journal streaming over virtio-serial" model. This would help with testing cases where no Ignition config is provided (live ISO and https://github.com/coreos/ignition/pull/958), early Ignition/initrd debugging (https://github.com/coreos/ignition-dracut/pull/146), and overall stronger kola journal gathering abilities on qemu.

Anyway, one suggestion though LGTM as is. Definitely good to get coverage for this!

overlay.d/05core/usr/lib/systemd/system/coreos-liveiso-success.service
cgwalters: Not as far as I can tell, because `-f` overrides `-n`.

cgwalters:
I think there's still a race condition here; I sometimes see the service still running, but I don't understand why.  Possibly related to [this](https://superuser.com/questions/696855/why-doesnt-a-bash-while-loop-exit-when-piping-to-terminated-subcommand)?

bgilbert:
https://github.com/coreos/fedora-coreos-config/pull/1208 should fix the service hanging.

PR #340, opened 2020-04-09
manifests: add initscripts to exclude-packages

dustymabe:
Let's just try to make sure nothing tries to bring it back in.


jlebon (approved):
Yes, agreed!

cgwalters:
Worth adding a comment link to the audit BZ I think, but yeah, LGTM!

dustymabe:
> Worth adding a comment link to the audit BZ I think, but yeah, LGTM!

done

PR #343, opened 2020-04-14
[testing-devel] manual update of lockfile

dustymabe:
While we currently have the bodhi-updates pipeline down let's manually
update the lockfile so we can get a new build going.

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -91,7 +91,7 @@
       "evra": "5.2-34.fc31.x86_64"
     },
     "conmon": {
-      "evra": "2:2.0.14-1.fc31.x86_64"
+      "evra": "2:2.0.15-1.fc31.x86_64"
     },
     "console-login-helper-messages": {
       "evra": "0.17-1.fc31.noarch"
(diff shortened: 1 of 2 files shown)


manifest-lock.x86_64.json
```diff
@@ -642,6 +642,9 @@
     "libsigsegv": {
       "evra": "2.11-8.fc31.x86_64"
     },
+    "libslirp": {
```
jlebon, line 645: New package. Worth doing a sanity-check re. size/content, though going to guess that's from podman.

jlebon (approved):
Thanks for this!

dustymabe:
CI is failing because the new fedora-release build:

```
error: Package 'fedora-release-common' has (currently) unsupported <lua> script in '%posttrans'
```

Looks like a lua scriptlet was added in https://src.fedoraproject.org/rpms/fedora-release/c/bfc4c31d8f625d4963a8169c9ba65686da7a4ce0?branch=f31 and [rpm-ostree doesn't support lua](https://github.com/coreos/rpm-ostree/issues/749). The good news is that it was only implemented for f31.

manifest-lock.x86_64.json
dustymabe: about 200K: 

```
$ rpm -ql libslirp | xargs -I{}  bash -c "test -f {} && ls -lh {}"
lrwxrwxrwx. 2 root root 39 Apr 14 14:20 /usr/lib/.build-id/b5/a077ba537e65969c9caeda966f0ca27f228375 -> ../../../../usr/lib64/libslirp.so.0.1.0
lrwxrwxrwx. 2 root root 17 Apr 14 14:20 /usr/lib64/libslirp.so.0 -> libslirp.so.0.1.0
-rwxr-xr-x. 2 root root 194K Jan  1  1970 /usr/lib64/libslirp.so.0.1.0
-rw-r--r--. 2 root root 2.9K Jan  1  1970 /usr/share/licenses/libslirp/COPYRIGHT
```

dustymabe:
> CI is failing because the new fedora-release build:

added a commit to hold the fedora-release rpms back while we sort out https://github.com/coreos/fedora-coreos-tracker/issues/459

PR #344, opened 2020-04-14
overlay/15fcos: Print information regarding an ignition config and authorized keys

sohankunkerkar:
Fixes #[279](https://github.com/coreos/fedora-coreos-tracker/issues/279)
Note:
As per the resolution provided in the above issue(https://github.com/coreos/fedora-coreos-tracker/issues/279#issuecomment-533731715),
-  The first bullet point is yet to finish (waiting for https://github.com/coreos/ignition/pull/958 to get merged)
- have already pushed the code for the second bullet point.

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
@@ -1,0 +1,11 @@
+# This service is used for printing a message if 
+# no ignition config is provided.
+[Unit]
+Description=Check if ignition config is provided
+Before=console-login-helper-messages-issuegen.service
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-check-ignition-config.sh
+RemainAfterExit=yes
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
@@ -1,0 +1,17 @@
+# This service is used for printing a message if no ssh keys were added 
+# by ignition/afterburn
+[Unit]
+Description=Check that ssh-keys are added by aferburn/ignition
+Before=console-login-helper-messages-issuegen.service
+# https://github.com/coreos/afterburn/issues/417 is created
+# to track the issue that would allow other units to synchronize 
+# around any instance of `afterburn-sshkeys@` and not just the 
+# `core` user.
+After=afterburn-sshkeys@core.service
+ProtectHome=read-only
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-check-ssh-keys.sh
+RemainAfterExit=yes
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/bash
+# The logic for the message_id is handled in 
+# https://github.com/coreos/ignition/pull/958
+# In this script, we need to capture the journald 
+# log with the particular message_id and query using 
+#`jq` utility to check if a user config is provided.
+
+# Change the output color to yellow
+warn='\033[0;33m'
+# No color
+nc='\033[0m'
+
+# See https://github.com/coreos/ignition/pull/958 for the MESSAGE_ID source.
+# It will track the journal messages related to an ignition config provided 
+# by the user.
+output=$(journalctl -o json-pretty MESSAGE_ID=57124006b5c94805b77ce473e92a8aeb | jq -s '.[] | select(.IGNITION_CONFIG_TYPE == "user")'| wc -l)
+
+if [[ $output -gt  0 ]];then
+   echo "Ignition: user provided config was applied" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+else
+  echo -e "${warn}Ignition: no config provided by user${nc}" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+fi
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/bash
+# This script will print a message in the serial console 
+# if no ssh keys were added by ignition/afterburn.
+main() {
+    # Change the output color to yellow
+    warn='\033[0;33m'
+    # No color
+    nc='\033[0m'
+
+    # See https://github.com/coreos/ignition/pull/964 for the MESSAGE_ID 
+    # source. It will track the authorized-ssh-keys entries in journald 
+    # provided via ignition.
+    ignitionusers=$(journalctl -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | jq -r '.MESSAGE')
+
+    # See https://github.com/coreos/afterburn/pull/397 for the MESSAGE_ID 
+    # source. It will track the authorized-ssh-keys entries in journald  
+    # provided via afterburn.
+    afterburnusers=$(journalctl -o json-pretty MESSAGE_ID=0f7d7a502f2d433caa1323440a6b4190 | jq -r '.MESSAGE')
+
+    output=''
+    if [ -n "$ignitionusers" ]; then
+        output+="$ignitionusers"
+    fi
+    if [ -n "$afterburnusers" ]; then
+        output+="$afterburnusers"
+    fi
+
+    if [ -n "$output" ]; then
+        echo "$output" > /run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue
+    else
+        echo -e "${warn}No ssh authorized keys provided by Ignition or Afterburn${nc}" \
+             > /run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue
+    fi
+}
+
+main
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -1,2 +1,5 @@
 # User metrics client
 enable fedora-coreos-pinger.service
+# Provide information if no ignition is provided
+enable coreos-check-ignition-config.service
+enable coreos-check-ssh-keys.service


overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Check that ssh-keys are added by aferburn/ignition
+After=console-login-helper-messages-issuegen.service afterburn-sshkeys@core.service
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/ssh_authorized_keys.sh
```
cgwalters: Let's namespace this more cleanly like `/usr/libexec/coreos-check-ssh-keys` or so.

overlay.d/15fcos/usr/libexec/ssh_authorized_keys.sh
```diff
@@ -0,0 +1,53 @@
...
+generate_issue_message(){
+   echo "ssh authorized keys were provided via ${1} for users: ${2}"
+}
+main() {
+
+    ignitionusers=''
+    afterburnusers=''
+
+    # Check for keys for all users with a home directory
+    for userhome in /home/*; do
```
cgwalters: Looking at user's home directories is a highly privileged thing to be doing.  It can also be problematic in automounting and NFS scenarios; this is why system services like GDM ended up depending on projects like [accountsservice](https://www.freedesktop.org/wiki/Software/AccountsService/) which store data out-of-band from the home directory.

Where possible we should be dropping privileges.  Let's at least use `ProtectHome=read-only` in the systemd unit - systemd makes it so easy to that kind of thing!

cgwalters (commented):
I am uncomfortable with poking into all of `/home` for the reasons below.  I think it'd be safer instead if we changed afterburn and Ignition to write out "stamp files" like:
`/var/lib/ignition/ssh-keys-written`
`/run/afterburn/ssh-keys-written`

or so.  Then all this service needs to do is run `test -f` on those files, and it doesn't need any special privileges.

The downside of this approach is it requires patching those projects.  For the Ignition case we discussed caching the config somewhere; if we did that we could just parse the config.

Maybe it's simplest to start with patching afterburn to write such a stamp file.

overlay.d/15fcos/usr/libexec/ssh_authorized_keys.sh
cgwalters: (~Also, this is omitting `/root` which is pretty relevant here~) EDIT: nevermind

cgwalters:
Or, building on the idea from https://github.com/coreos/ignition/pull/958 instead of the "stamp files" we could use a journal entry.

sohankunkerkar:
It looks like we need to cut a new release of `ignition` and `afterburn` repo to proceed with the CI testing. This PR is dependent on the following PRs which are merged recently. 

https://github.com/coreos/ignition/pull/958
https://github.com/coreos/ignition/pull/964
https://github.com/coreos/afterburn/pull/397

cgwalters:
> It looks like we need to cut a new release of ignition and afterburn repo to proceed with the CI testing. This PR is dependent on the following PRs which are merged recently.

Yeah, that's currently a manual process.

overlay.d/15fcos/usr/libexec/coreos-check-user-config.sh
```diff
@@ -0,0 +1,20 @@
...
+# The logic for the message_id is handled in 
+# https://github.com/coreos/ignition/pull/958
+# In this script, we need to capture the journald 
+# log with the particular message_id and query using 
+#`jq` utility to check if a user config is provided.
+
+# See https://github.com/coreos/ignition/pull/958 for the MESSAGE_ID source.
+# It will track the journal messages related to an ignition config provided 
+# by the user.
+exec journalctl -o json-pretty MESSAGE_ID=57124006b5c94805b77ce473e92a8aeb > output.json
```
dustymabe: a few questions:

- why does this need to be run as part of `exec` ?
- instead of writing to `output.json` let's just capture the output in a variable or something. If we do stick with writing a file we should write it to a temporary file somewhere.

overlay.d/15fcos/usr/libexec/coreos-check-user-config.sh
```diff
@@ -0,0 +1,20 @@
...
+
+val=$(jq '.IGNITION_CONFIG_TYPE' output.json | grep -c "user")
+
+if [[ $val -eq  0 ]];then
```
dustymabe: here you are capturing the string output of `grep`. We can just check the return code instead.

```
if jq '.IGNITION_CONFIG_TYPE' output.json | grep -c "user"; then
 echo "user-config is provided"
else
 echo "No user-config is provided"
 exit 1
fi
```

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-user-config.service
```diff
@@ -0,0 +1,11 @@
+# This service is used for printing a message if 
+# no ignition config is provided.
+[Unit]
+Description=Check if ignition config is provided
+Before=console-login-helper-messages-issuegen.service
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-check-user-config.sh
```
dustymabe: where does this script write its output to? Doesn't look like it's writing it to anywhere that will get picked up.

overlay.d/15fcos/usr/libexec/coreos-check-user-config.sh
```diff
@@ -0,0 +1,20 @@
...
+ echo "No user-config is provided"
+ exit 1
+else
+ echo "user-config is provided"
```
dustymabe: How will this information be presented to the user? I'm thinking we can improve the message text.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-user-config.service
dustymabe: Also can we change it to `coreos-check-ignition-config.sh` ?

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
```diff
@@ -0,0 +1,12 @@
+# This service is used for printing a message if no ssh keys were added 
+# by ignition/afterburn
+[Unit]
+Description=Check that ssh-keys are added by aferburn/ignition
+After=console-login-helper-messages-issuegen.service afterburn-sshkeys@core.service
```
rfairley: `Before=console-login-helper-messages-issuegen.service` would be best here I believe (like done for `coreos-check-user-config.service`) - to give the best chance of `console-login-helper-messages-issuegen.service` of picking up the dropped `30_ssh_authorized_keys.issue` snippet.

The current way of integrating with c-l-h-m that I'm suggesting with `Before=` isn't very obvious or robust (there could be a race, if writing the snippet from this unit somehow got delayed until after the c-l-h-m service finished, causing the snippet to not get displayed). This would be improved I think by https://github.com/coreos/console-login-helper-messages/pull/47 - which I will aim to include soon in a new release of c-l-h-m. That way, `After=console-login-helper-messages-issuegen.path` could be specified, which would guarantee the displayed snippet gets updated properly.

My suggestion for now is `Before=...-issuegen.service`, but once the above linked PR is in, we should change it to `After=...-issuegen.path`.

sohankunkerkar: @rfairley Thanks for the information. I think we can wait here until your PR(https://github.com/coreos/console-login-helper-messages/pull/47) gets in the new release of `c-l-h-m`.

rfairley: To avoid blocking on my PR (https://github.com/coreos/console-login-helper-messages/pull/47) doing `Before=console-login-helper-messages-issuegen.service` would be best I believe. With this, there'd be no guarantee by systemd that `issuegen` will execute after `30_ssh_authorized_keys.issue` is written, and `30_ssh_authorized_keys.issue` may not get picked up. But using `Before=` (for both systemd units in this PR) would make it unlikely that this occurs. If testing locally, the messages do show properly (at least for a good number of times, say 5 in a row) when using `Before=`, then we should be good here.

rfairley: > good number of times, say 5 in a row

By this - on the first 5 consecutive boots where we expect the message to display

sohankunkerkar: @rfairley Thanks for your inputs. I think that makes sense to me and reasonable enough to test this feature multiple times (5 sgtm) to see if we're getting the right information displayed on the serial console.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
```diff
@@ -0,0 +1,12 @@
+# This service is used for printing a message if no ssh keys were added 
+# by ignition/afterburn
+[Unit]
+Description=Check that ssh-keys are added by aferburn/ignition
+Before=console-login-helper-messages-issuegen.service afterburn-sshkeys@core.service
```
rfairley: Should we still keep `After=afterburn-sshkeys@core.service` (I believe initially this was the case in this PR)?

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# See https://github.com/coreos/ignition/pull/958 for the MESSAGE_ID source.
+# It will track the journal messages related to an ignition config provided 
+# by the user.
+output=$(journalctl -o json-pretty MESSAGE_ID=57124006b5c94805b77ce473e92a8aeb | jq -s '.[] | select(.IGNITION_CONFIG_TYPE == "user")'| wc -l)
+
+if [[ $output -gt  0 ]];then
+   echo "Ignition config is provided by the user" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+else
+  echo "Ignition config isn't provided by the user" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
```
dustymabe: I think we can improve on these messages slightly:

- `Ignition: user provided config was applied`
- `Ignition: no config provided by user`


If there is an opportunity for us use a yellow color (warning) or something for that "no config" case would be excellent.

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -0,0 +1,32 @@
...
+    output=''
+    if [ -z "$ignitionusers" ]; then
+        output+="$ignitionusers"
+    fi
+    if [ -z "$afterburnusers" ]; then
+        output+="$afterburnusers"
+    fi
+
+    if [ -n $output ]; then
+        echo "$output" > /run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue
```
dustymabe, line 29: what does this look like on the console of the machine? Seems like we need to output more than just `"$output"`. Seems like it should be:

- `ssh authorized keys provided via Afterburn for user(s): core foo bar`
- `ssh authorized keys provided via Ignition for user(s): core foo bar`

OR if no keys provided what you have below with a slight tweak

- `No ssh authorized keys provided via Ignition or Afterburn`

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
dustymabe: yeah I'm thinking we need `After=afterburn-sshkeys@core.service`, though is that an instance for a specific user? 

I think we probably also need `Before=systemd-user-sessions.service` so the issue gets written out before the console starts up.

dustymabe: > I think we probably also need `Before=systemd-user-sessions.service` so the issue gets written out before the console starts up.

maybe `Before=console-login-helper-messages-issuegen.service` gives us that for free?

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/bash
+# This script will print a message in the serial console 
+# if no ssh keys were added by ignition/afterburn.
+main() {
+
+    # See https://github.com/coreos/ignition/pull/964 for the MESSAGE_ID 
+    # source.It will track the authorized-ssh-keys entries in journald
```
dustymabe:
```suggestion
    # source. It will track the authorized-ssh-keys entries in journald 
```

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
rfairley: > > I think we probably also need `Before=systemd-user-sessions.service` so the issue gets written out before the console starts up.
> 
> maybe `Before=console-login-helper-messages-issuegen.service` gives us that for free?

Right yes, issuegen will give us this in https://github.com/coreos/console-login-helper-messages/blob/00b17df7f11b361a86eecd34c33a1be2bfe4e2ae/usr/lib/systemd/system/console-login-helper-messages-issuegen.service#L3. No harm in including it explicitly here, but we can rely on CLHM-issuegen keeping `Before=systemd-user-sessions.service`.

rfairley: > yeah I'm thinking we need After=afterburn-sshkeys@core.service, though is that an instance for a specific user?

Yes, this would only apply for the `@core` instance of the unit and so the `core` user. It doesn't seem there is a way to refer to all instances of a template unit. To apply for other users, we could add another target in afterburn, `afterburn-sshkeys.target`, which requires all the `afterburn-sshkeys@.service` instances (i.e. the `afterburn-sshkeys@.service` template specifies `RequiredBy=afterburn-sshkeys.target` [here](https://github.com/coreos/afterburn/blob/9717ea8eab198359c61481b1cb9c7ea04b12ba29/systemd/afterburn-sshkeys%40.service.in)). Then, I think we could do `After=afterburn-sshkeys.target` here which would cover other users.

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
sohankunkerkar: > what does this look like on the console of the machine? Seems like we need to output more than just `"$output"`. Seems like it should be:
> 
>     * `ssh authorized keys provided via Afterburn for user(s): core foo bar`
> 
>     * `ssh authorized keys provided via Ignition for user(s): core foo bar`
> 
I think we're already sending structured entry into the jorunald log for both [ignition](https://github.com/coreos/ignition/pull/964/files#diff-e035c66c974fdac1b7f18ec980142566R192)` and `[afterburn](https://github.com/coreos/afterburn/pull/397/files#diff-0b2226915abcd4a043e0dcf9bd303a7eR74) users. And that message will be printed separately for each of the users. For e.g.
```shell
$ journalctl -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | jq '.MESSAGE'
"wrote ssh authorized keys file for user: core"
"wrote ssh authorized keys file for user: xyz"
```

dustymabe: ahh ok - yeah that looks fine - I didn't realize the journal entry had all of that output.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
```diff
@@ -0,0 +1,13 @@
+# This service is used for printing a message if no ssh keys were added 
+# by ignition/afterburn
+[Unit]
+Description=Check that ssh-keys are added by aferburn/ignition
+Before=console-login-helper-messages-issuegen.service 
+After=afterburn-sshkeys.target
```
rfairley: Sorry - before I may not have been clear in https://github.com/coreos/fedora-coreos-config/pull/344#discussion_r428844257. Was proposing adding `afterburn-sshkeys.target` as an enhancement to Afterburn - which would live in https://github.com/coreos/afterburn/tree/master/systemd. If we had it, we could then use the target here and it'd let us execute after getting SSH keys for all users, not just `core`. Not sure if we should block this PR on adding that functionality, it may need some discussion (cc @lucab, @bgilbert).

lucab: Assuming that works, I think it's a reasonable idea. The dependency between the target and the instances should be manually tested first though, as I'm not sure how they play together.

dustymabe: maybe we can just depend on `After=afterburn-sshkeys@core.service` for now and get this in and then switch to the target once it exists

dustymabe: IOW - existing code in this PR LGTM - let's merge and change it later when we get the target.

dustymabe (approved):
LGTM - this was a long time coming. Nice work getting changes into all relevant upstreams and then bringing it all together here.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
lucab: Indeed, we can come back to this later, just keeping a ticket about it somewhere.

dustymabe: @sohankunkerkar - can you open a ticket against afterburn for this and maybe add a comment here in the unit file that links to the ticket (requesting the systemd target)?

lucab (approved):

rfairley (approved):

dustymabe:
when I test this using `cosa run` I don't see the output for the ssh keys check the issue.d file is empty: 

/run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -0,0 +1,36 @@
...
+
+    # See https://github.com/coreos/ignition/pull/964 for the MESSAGE_ID 
+    # source. It will track the authorized-ssh-keys entries in journald 
+    # provided via ignition.
+    ignitionusers=$(journalctl -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | jq '.MESSAGE')
+
+    # See https://github.com/coreos/afterburn/pull/397 for the MESSAGE_ID 
+    # source. It will track the authorized-ssh-keys entries in journald  
+    # provided via afterburn.
+    afterburnusers=$(journalctl -o json-pretty MESSAGE_ID=0f7d7a502f2d433caa1323440a6b4190 | jq '.MESSAGE')
```
dustymabe: let's switch this to `jq -r`

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -0,0 +1,36 @@
...
+        output+="$ignitionusers"
+    fi
+    if [ -z "$afterburnusers" ]; then
+        output+="$afterburnusers"
+    fi
+
+    if [ -n $output ]; then
+        echo "$output" > /run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue
+    else
+        echo "${warn}No ssh authorized keys provided by Ignition or Afterburn${nc}" \
```
dustymabe: I think we need to make this `echo -e` so that the color codes get processed.

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# See https://github.com/coreos/ignition/pull/958 for the MESSAGE_ID source.
+# It will track the journal messages related to an ignition config provided 
+# by the user.
+output=$(journalctl -o json-pretty MESSAGE_ID=57124006b5c94805b77ce473e92a8aeb | jq -s '.[] | select(.IGNITION_CONFIG_TYPE == "user")'| wc -l)
+
+if [[ $output -gt  0 ]];then
+   echo "Ignition: user provided config was applied" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+else
+  echo "${warn}Ignition: no config provided by user${nc}" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
```
dustymabe: I think we need to make this `echo -e` so that the color codes get processed.

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -0,0 +1,36 @@
...
+    # provided via ignition.
+    ignitionusers=$(journalctl -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | jq '.MESSAGE')
+
+    # See https://github.com/coreos/afterburn/pull/397 for the MESSAGE_ID 
+    # source. It will track the authorized-ssh-keys entries in journald  
+    # provided via afterburn.
+    afterburnusers=$(journalctl -o json-pretty MESSAGE_ID=0f7d7a502f2d433caa1323440a6b4190 | jq '.MESSAGE')
+
+    output=''
+    if [ -z "$ignitionusers" ]; then
```
dustymabe: this should be `-n`, not `-z`

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -0,0 +1,36 @@
...
+        output+="$ignitionusers"
+    fi
+    if [ -z "$afterburnusers" ]; then
```
dustymabe: switch to `-n`

dustymabe:
> when I test this using `cosa run` I don't see the output for the ssh keys check the issue.d file is empty:
> 
> /run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue

turned out to be https://github.com/coreos/fedora-coreos-config/pull/344/files#r429340945

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
sohankunkerkar: Thanks, @rfairley for creating a ticket to track this issue.

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -0,0 +1,36 @@
...
+
+    output=''
+    if [ -n "$ignitionusers" ]; then
+        output+="$ignitionusers"
+    fi
+    if [ -n "$afterburnusers" ]; then
+        output+="$afterburnusers"
+    fi
+
+    if [ -n $output ]; then
```
dustymabe:
```suggestion
    if [ -n "$output" ]; then
```

Final issue is that ^^ needs to be quoted. Otherwise I've been getting blank `/run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue` files

dustymabe:
ok all problems have been addressed!

dustymabe:
and on GCP I now see:

gcp


Nice work Sohan! Thanks everyone for all the reviews!

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
```diff
@@ -0,0 +1,11 @@
+# This service is used for printing a message if 
+# no ignition config is provided.
+[Unit]
+Description=Check if ignition config is provided
```
bgilbert, line 4: Ignition is always capitalized.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
```diff
@@ -0,0 +1,17 @@
+# This service is used for printing a message if no ssh keys were added 
+# by ignition/afterburn
+[Unit]
+Description=Check that ssh-keys are added by aferburn/ignition
```
bgilbert, line 4: Afterburn/Ignition

bgilbert (commented):
Couple spelling nits, FWIW.

sohankunkerkar:
> Couple spelling nits, FWIW.

My bad. Thanks for point this out. I will create a follow-up PR to fix it.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
```diff
@@ -0,0 +1,11 @@
...
+Before=console-login-helper-messages-issuegen.service
```
rfairley, line 5: Noting that with https://github.com/coreos/console-login-helper-messages/pull/55 we'll be able to drop this and have no explicit systemd dependency on CLHM.

sohankunkerkar: @rfairley Thanks! Let me create a follow-up PR to address this part.

PR #346, opened 2020-04-15
overlay: add 15copy-installer-network dracut module

dustymabe:
This dracut module delivers a coreos-copy-installer-network
systemd service and script that will detect when files have
been placed into /boot/ by `coreos-installer install --copy-network`
and appropriately copy them in place to be used by the initramfs
networking. If files are detected within /boot/coreos-installer-network/
then they will be considered to be the only source of networking
for that ignition boot (i.e. no networking kargs will be applied).

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.service
@@ -1,0 +1,45 @@
+# This unit will run early in boot and detect if the user copied
+# in firstboot networking config files into the installed system
+# (most likely by using `coreos-installer install --copy-network`).
+# Since this unit is modifying network configuration there are some
+# dependencies that we have:
+#
+# - Need to look for networking configuration on the /boot partition
+#     - i.e. after /dev/disk/by-label/boot is available
+# - Need to run before networking is brought up.
+#     - This is done in nm-run.sh [1] that runs as part of dracut-initqueue [2]
+#     - i.e. Before=dracut-initqueue.service
+# - Need to make sure karg networking configuration isn't applied
+#     - There are two ways to do this.
+#         - One is to run *before* the nm-config.sh [3] that runs as part of
+#           dracut-cmdline [4] and `ln -sf /bin/true /usr/libexec/nm-initrd-generator`.
+#             - i.e. Before=dracut-cmdline.service
+#         - Another is to run *after* nm-config.sh [3] in dracut-cmdline [4]
+#           and just delete all the files created by nm-initrd-generator.
+#             - i.e. After=dracut-cmdline.service, but Before=dracut-initqueue.service
+#     - We'll go with the second option here because the need for the /boot
+#       device (mentioned above) means we can't start before dracut-cmdline.service
+#
+# [1] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/nm-run.sh
+# [2] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/module-setup.sh#L37
+# [3] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/nm-config.sh
+# [4] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/module-setup.sh#L36
+#
+[Unit]
+Description=Copy CoreOS Firstboot Networking Config
+ConditionPathExists=/usr/lib/initrd-release
+DefaultDependencies=false
+Before=ignition-diskful.target
+Before=dracut-initqueue.service
+After=dracut-cmdline.service
+# Since we are mounting /boot/, require the device first
+Requires=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device   
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+# The MountFlags=slave is so the umount of /boot is guaranteed to happen
+# /boot will only be mounted for the lifetime of the unit.
+MountFlags=slave
+ExecStart=/usr/sbin/coreos-copy-firstboot-network
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
@@ -4,8 +4,10 @@ set -euo pipefail
 # For a description of how this is used see coreos-copy-firstboot-network.service
 
 bootmnt=/mnt/boot_partition
-firstboot_network_dir="${bootmnt}/coreos-firstboot-network/"
+firstboot_network_dir_basename="coreos-firstboot-network"
+initramfs_firstboot_network_dir="${bootmnt}/${firstboot_network_dir_basename}"
 initramfs_network_dir="/run/NetworkManager/system-connections/"
+realroot_firstboot_network_dir="/boot/${firstboot_network_dir_basename}"
 
 # Mount /boot. Note that we mount /boot but we don't unmount boot because we
 # are run in a systemd unit with MountFlags=slave so it is unmounted for us.
@@ -14,15 +16,19 @@ mkdir -p ${bootmnt}
 # running alongside other code that also has it mounted ro
 mount -o ro /dev/disk/by-label/boot ${bootmnt}
 
-if [ -n "$(ls -A ${firstboot_network_dir} 2>/dev/null)" ]; then
+if [ -n "$(ls -A ${initramfs_firstboot_network_dir} 2>/dev/null)" ]; then
     # Clear out any files that may have already been generated from
     # kargs by nm-initrd-generator
     rm -f ${initramfs_network_dir}/*
     # Copy files that were placed into boot (most likely by coreos-installer)
     # to the appropriate location for NetworkManager to use the configuration.
-    echo "info: copying files from ${firstboot_network_dir} to ${initramfs_network_dir}"
+    echo "info: copying files from ${initramfs_firstboot_network_dir} to ${initramfs_network_dir}"
     mkdir -p ${initramfs_network_dir}
-    cp -v ${firstboot_network_dir}/* ${initramfs_network_dir}/
+    cp -v ${initramfs_firstboot_network_dir}/* ${initramfs_network_dir}/
+    # If we make it to the realroot (successfully ran ignition) then
+    # clean up the files in the firstboot network dir
+    echo "R ${realroot_firstboot_network_dir} - - - - -" > \
+        /run/tmpfiles.d/15-coreos-firstboot-network.conf
 else
-    echo "info: no files to copy from ${firstboot_network_dir}. skipping"
+    echo "info: no files to copy from ${initramfs_firstboot_network_dir}. skipping"
 fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+
+# For a description of how this is used see coreos-copy-firstboot-network.service
+
+bootmnt=/mnt/boot_partition
+firstboot_network_dir="${bootmnt}/coreos-firstboot-network/"
+initramfs_network_dir="/run/NetworkManager/system-connections/"
+
+# Mount /boot. Note that we mount /boot but we don't unmount boot because we
+# are run in a systemd unit with MountFlags=slave so it is unmounted for us.
+mkdir -p ${bootmnt}
+# mount as read-only since we don't strictly need write access and we may be
+# running alongside other code that also has it mounted ro
+mount -o ro /dev/disk/by-label/boot ${bootmnt}
+
+if [ -n "$(ls -A ${firstboot_network_dir} 2>/dev/null)" ]; then
+    # Clear out any files that may have already been generated from
+    # kargs by nm-initrd-generator
+    rm -f ${initramfs_network_dir}/*
+    # Copy files that were placed into boot (most likely by coreos-installer)
+    # to the appropriate location for NetworkManager to use the configuration.
+    echo "info: copying files from ${firstboot_network_dir} to ${initramfs_network_dir}"
+    mkdir -p ${initramfs_network_dir}
+    cp -v ${firstboot_network_dir}/* ${initramfs_network_dir}/
+else
+    echo "info: no files to copy from ${firstboot_network_dir}. skipping"
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
@@ -4,6 +4,8 @@ set -euo pipefail
 # For a description of how this is used see coreos-copy-firstboot-network.service
 
 bootmnt=/mnt/boot_partition
+mkdir -p ${bootmnt}
+bootdev=/dev/disk/by-label/boot
 firstboot_network_dir_basename="coreos-firstboot-network"
 initramfs_firstboot_network_dir="${bootmnt}/${firstboot_network_dir_basename}"
 initramfs_network_dir="/run/NetworkManager/system-connections/"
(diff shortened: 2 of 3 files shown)


dustymabe:
In order to propagate networking from the live ISO installer environment the installer will place files into a directory in /boot/ (https://github.com/coreos/coreos-installer/pull/212) and the systemd service/script from this PR will pick that up and apply it during the initramfs boot process on the ignition boot (firstboot) of the installed system.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [ -n "$(ls -A ${installer_network_dir} 2>/dev/null)" ]; then
+    # Clear out any files that may have already been generated from
+    # kargs by nm-initrd-generator
+    rm -f ${initramfs_network_dir}/*
+    # Copy files that were placed by the installer to the appropriate 
+    # location for NetworkManager to use the configuration.
+    echo "info: copying files from ${installer_network_dir} to ${initramfs_network_dir}"
+    mkdir -p $initramfs_network_dir
+    cp -v ${installer_network_dir}/* ${initramfs_network_dir}/
```
jlebon: I think we also want to delete `installer_network_dir` here, no? Otherwise, it'll just hang around forever.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.sh
```diff
@@ -0,0 +1,28 @@
...
+mount -o ro /dev/disk/by-label/boot $bootmnt
+
+if [ -n "$(ls -A ${installer_network_dir} 2>/dev/null)" ]; then
+    # Clear out any files that may have already been generated from
+    # kargs by nm-initrd-generator
+    rm -f ${initramfs_network_dir}/*
+    # Copy files that were placed by the installer to the appropriate 
+    # location for NetworkManager to use the configuration.
+    echo "info: copying files from ${installer_network_dir} to ${initramfs_network_dir}"
+    mkdir -p $initramfs_network_dir
```
jlebon: This is minor, though we're consistently using the `${var}` convention throughout except here.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.service
```diff
@@ -0,0 +1,45 @@
...
+[Unit]
+Description=Copy Live ISO Installer Networking Config
+ConditionPathExists=/usr/lib/initrd-release
+DefaultDependencies=false
+Before=ignition-diskful.target
+Before=dracut-initqueue.service
+After=dracut-cmdline.service
+# Since we are mounting /boot/, require the device first
+Requires=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device
```
jlebon: Shouldn't we also have e.g. `ConditionKernelCommandLine=ignition.platform.id=metal` here?

jlebon (commented):
It still feels wrong to me that this mechanism takes precedence over karg network options the user might've input. Though I guess we can fix that once we do https://github.com/coreos/coreos-assembler/pull/1298 as discussed in https://github.com/coreos/fedora-coreos-tracker/issues/460.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.sh
dustymabe: I'd rather keep /boot/ mounted read-only here. I don't think it hurts to keep the directory around especially since nothing after firstboot will consume it anyway.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.service
dustymabe: I don't know if it'd be safe to do that. You can change the platform id using the installer.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.sh
dustymabe: 👍 - I'll update it

dustymabe:
> It still feels wrong to me that this mechanism takes precedence over karg network options the user might've input.

I don't really feel like it's that wrong considering the user just did an install and explicitly told the installer to bake specific networking configuration in specifically for this first boot. We do have to do it this way for now (ignoring kargs completely) because we do embed default networking args, but as you mention below 👇🏼 we maybe be able to rework it when we no longer do that.

> Though I guess we can fix that once we do [coreos/coreos-assembler#1298](https://github.com/coreos/coreos-assembler/pull/1298) as discussed in [coreos/fedora-coreos-tracker#460](https://github.com/coreos/fedora-coreos-tracker/issues/460).

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.sh
dustymabe: updated in most recent force-push

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.service
bgilbert: Right, we can't assume `metal`.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.service
```diff
@@ -0,0 +1,45 @@
...
+#     - We'll go with the second option here because the need for the /boot
+#       device (mentioned above) means we can't start before dracut-cmdline.service
+#
+# [1] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/nm-run.sh
+# [2] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/module-setup.sh#L37
+# [3] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/nm-config.sh
+# [4] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/module-setup.sh#L36
+#
+[Unit]
+Description=Copy Live ISO Installer Networking Config
```
bgilbert: `Live ISO` is a misnomer here.  `--copy-network` can be done from live PXE, or in principle from a Fedora live image running the coreos-installer container.

dustymabe: Technically you are right. How about `Copy CoreOS Installer Networking Config` ?

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.sh
jlebon: Oh right, forgot about the ro mount.

> I don't think it hurts to keep the directory around especially since nothing after firstboot will consume it anyway.

Right yeah, not so much that it conflicts with anything. It's just bad optics IMO to have it stay in the root of `/boot` forever. For example, back when we used Anaconda, we would actively remove everything in `/sysroot` except what was needed (https://github.com/coreos/coreos-assembler/pull/492). Nowadays we just have `/boot`, `/ostree`, and `.coreos-aleph-version.json`, which is nice. Similarly an `ls /boot` right now only has the files we actually need.

WDYT about adding a `/run/tmpfiles.d/10-coreos-installer-network.conf` file which has:

```
R /boot/coreos-installer-network - - - - -
```

?

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.service
jlebon: We could also keep it generic as just a mechanism for baking in network configs in the image and `coreos-installer` just happens to use it. So e.g. it'd be `Copy Network Configs` and the directory would be `${bootmnt}/coreos-network-configs` or something.

bgilbert: I'm okay with either of those options.

jlebon: (This is similar to how we support adding a `$boot/ignition/config.ign` without the help of coreos-installer, and in fact kola knows how to do this today.)

dustymabe: ok if we want to change the name of the directory now is the time to do it. I named it something specific because I didn't see other uses for this (or at least not other encouraged uses). What do you all think we should do?

dustymabe: bikeshed: `/boot/firstboot-network-configs`

dustymabe: we decided in IRC to go with: `/boot/coreos-firstboot-network`

dustymabe: coreos-installer PR to change the dir: https://github.com/coreos/coreos-installer/pull/216

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.sh
dustymabe: ok added a second commit to clean up via tmpfiles.d. I can squash it if desired.

overlay.d/05core/usr/lib/dracut/modules.d/15copy-installer-network/coreos-copy-installer-network.service
dustymabe: updated this PR to use `/boot/coreos-firstboot-network`

dustymabe:
bringing this out of draft mode. testing looks good (combined with https://github.com/coreos/coreos-installer/pull/216)

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
```diff
@@ -0,0 +1,34 @@
...
+    rm -f ${initramfs_network_dir}/*
+    # Copy files that were placed into boot (most likely by coreos-installer)
+    # to the appropriate location for NetworkManager to use the configuration.
+    echo "info: copying files from ${initramfs_firstboot_network_dir} to ${initramfs_network_dir}"
+    mkdir -p ${initramfs_network_dir}
+    cp -v ${initramfs_firstboot_network_dir}/* ${initramfs_network_dir}/
+    # If we make it to the realroot (successfully ran ignition) then
+    # clean up the files in the firstboot network dir
+    echo "R ${realroot_firstboot_network_dir} - - - - -" > \
+        /run/tmpfiles.d/15-coreos-firstboot-network.conf
```
jlebon, line 63: Can we move this out so we always do this if the directory exists, not just if it's not empty?

jlebon (approved):
One comment, LGTM otherwise.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh
dustymabe: I'd prefer not. Deleting the files (presumably on success) will already make things hard to debug. This would make it worse I think.

dustymabe:
```
--- FAIL: ostree.remote (337.27s)

    --- PASS: ostree.remote/add (0.26s)

    --- PASS: ostree.remote/list (0.11s)

    --- PASS: ostree.remote/show-url (0.10s)

    --- PASS: ostree.remote/refs (1.01s)

    --- PASS: ostree.remote/summary (1.83s)

    --- PASS: ostree.remote/delete (0.41s)

        harness.go:731: Found emergency shell on machine 894d3b2a-eaff-4256-93bf-3d442af4d1fe console

```

This test passes for me when I run it locally so maybe let's retry.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.service
```diff
@@ -0,0 +1,45 @@
...
+#     - This is done in nm-run.sh [1] that runs as part of dracut-initqueue [2]
+#     - i.e. Before=dracut-initqueue.service
+# - Need to make sure karg networking configuration isn't applied
+#     - There are two ways to do this.
+#         - One is to run *before* the nm-config.sh [3] that runs as part of
+#           dracut-cmdline [4] and `ln -sf /bin/true /usr/libexec/nm-initrd-generator`.
+#             - i.e. Before=dracut-cmdline.service
+#         - Another is to run *after* nm-config.sh [3] in dracut-cmdline [4]
+#           and just delete all the files created by nm-initrd-generator.
+#             - i.e. After=dracut-cmdline.service, but Before=dracut-initqueue.service
```
cgwalters, line 19: This is a clear and obvious downside of the "running NM via dracut initqueue"...

dustymabe:
The CI failure This appears to be a race condition. The `After=...*boot.device` in the systemd unit should be enough, but apparently not (kernel race, fun!):
       
```                                                                                                                                     
  [    4.045181] systemd[1]: Found device /dev/disk/by-label/boot.                                                                     
  [  OK  ] Found device /dev/disk/by-label/boot                                                                                        
  [    4.051500] systemd[1]: Starting Copy CoreOS Firstboot Networking Config...                                                       
        Starting  Copy CoreOS Firstboot Networking Config                                                                              
  [    4.060573]  vda: vda1 vda2 vda3 vda4                                                                                             
  [    4.063296] coreos-copy-firstboot-network[479]: mount: /mnt/boot_partition: special device /dev/disk/by-label/boot does not exist.
```

will add a short loop to wait for the device before continuing

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.service
dustymabe: Agree. Though when we do finally get NM running via systemd we'll need to make sure we work with the team such that there are two separate services: one for running `nm-initrd-generator` and one for starting NM. This will enable us to hook in between them OR maybe be able to just say `Before=nm-initrd-generator.service`

dustymabe:
yay! CI passed now

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.service
```diff
@@ -0,0 +1,45 @@
+# This unit will run early in boot and detect if the user copied
+# in firstboot networking config files into the installed system
+# (most likely by using `coreos-installer install --copy-network`.
```
bgilbert: )

bgilbert (approved):
LGTM

dustymabe:
CI re-ran again because I pushed up a comment change for https://github.com/coreos/fedora-coreos-config/pull/346#discussion_r409965301 and it failed. It appears the same fallacy that plagues `After=dev-disk-by\x2dlabel-boot.device` also plagues `[ -e /dev/disk/by-label/boot ]` so I switched the loop to just retry the actual `mount` command instead.

dustymabe:
if CI does pass I'll look at the logs and hopefully we'll see at least one `info: retrying ${bootdev} mount in 1 second...` message

dustymabe:
> if CI does pass I'll look at the logs and hopefully we'll see at least one `info: retrying ${bootdev} mount in 1 second...` message

It did pass! and we do see it! 

```
ostree.unlock/c0895ef6-5c96-417c-a6d7-171050a4c29a/journal.txt:Apr 17 14:21:24.260616 coreos-copy-firstboot-network[481]: info: retrying /dev/disk/by-label/boot mount in 1 second...
ostree.unlock/c0895ef6-5c96-417c-a6d7-171050a4c29a/journal.txt:Apr 17 14:21:25.269617 coreos-copy-firstboot-network[481]: info: retrying /dev/disk/by-label/boot mount in 1 second...
ostree.unlock/c0895ef6-5c96-417c-a6d7-171050a4c29a/console.txt:[    3.755254] coreos-copy-firstboot-network[481]: info: retrying /dev/disk/by-label/boot mount in 1 second...
ostree.unlock/c0895ef6-5c96-417c-a6d7-171050a4c29a/console.txt:[    4.764259] coreos-copy-firstboot-network[481]: info: retrying /dev/disk/by-label/boot mount in 1 second...
```

👏 👏 👏 👏

jlebon (approved):

cgwalters:
I think the races may be due to `coreos-gpt-setup@dev-disk-by\x2dlabel-root.service`.  We need more ordering against that.

cgwalters:
Still racy, this PR failed: https://jenkins-coreos-ci.apps.ci.centos.org/blue/organizations/jenkins/github-ci%2Fostreedev%2Fostree/detail/PR-2121/3/pipeline/154

cgwalters:
Filed as https://github.com/coreos/fedora-coreos-tracker/issues/523

PR #347, opened 2020-04-15
live: disable audit on Live ISO

dustymabe:
This is the nuclear option to fix the issue [1] where we have
audit messages coming to the console of our machines. We
specifically need the messages to be disabled on the Live ISO
because we are trying to support a workflow where a user can
interactively do an install and use TUI interfaces for configuring
networking. The increased user experience provided by a TUI is
completely lost if we have audit messages scrolling on the console
every 10 seconds or so.

### DIFF
--- a/live/EFI/fedora/grub.cfg
+++ b/live/EFI/fedora/grub.cfg
@@ -28,6 +28,6 @@ set timeout=5
 
 ### BEGIN /etc/grub.d/10_linux ###
 menuentry 'Fedora CoreOS (Live)' --class fedora --class gnu-linux --class gnu --class os {
-	linux /images/vmlinuz @@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal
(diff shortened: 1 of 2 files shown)


dustymabe:
The change in this PR is one of two possible options to fix https://github.com/coreos/fedora-coreos-tracker/issues/220. This change was proposed to spark discussion. Please see https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-613803086 for more information.

dustymabe:
drawbacks of this approach:

- auditing is disabled for Live ISO workloads
- Live ISO will different from every other platform we run

dustymabe:
We are going with https://github.com/coreos/fedora-coreos-config/pull/352 see https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-614241078. Regarding the future of audit in FCOS see https://github.com/coreos/fedora-coreos-tracker/issues/461

PR #348, opened 2020-04-15
manifest: include audit rpm

dustymabe:
This change will cause the auditd daemon to get included in
Fedora CoreOS and be enabled by default. This is a roundabout
way of fixing [1] where we have audit messages coming to the
primary console of our FCOS machines.

This need for this is increased right now because we are trying
to support a workflow where a user can interactively do an install
using the Live ISO and use TUI interfaces for configuring networking.
The increased user experience provided by a TUI is completely lost
if we have audit messages scrolling on the console every 10 seconds
or so.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -55,6 +55,9 @@ remove-from-packages:
               /usr/lib/systemd/system/systemd-networkd-wait-online.service]
   - [systemd-container, /usr/lib/systemd/network/.*]
   - [systemd-udev, /usr/lib/systemd/network/.*]
+  # We don't want initscripts but audit insists on requiring it
+  # https://src.fedoraproject.org/rpms/audit/pull-request/5
+  - [initscripts, .*]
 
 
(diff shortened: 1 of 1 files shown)


dustymabe:
The change in this PR is one of two possible options to fix https://github.com/coreos/fedora-coreos-tracker/issues/220. This change was proposed to spark discussion. Please see https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-613803086 for more information.

dustymabe:
drawbacks of this approach:

- we include an extra rpm (audit), that includes auditd
- removing the audit rpm in the future will be hard

silver lining:

- auditd and tools are required by some compliance benchmarks anyway

dustymabe:
We are going with https://github.com/coreos/fedora-coreos-config/pull/352 see https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-614241078. Regarding the future of audit in FCOS see https://github.com/coreos/fedora-coreos-tracker/issues/461

PR #349, opened 2020-04-15
manifest: move sshd config fragments to overlay sshd_config.d on F32

bgilbert:
Fedora 32 supports `sshd_config.d`.  Use it.  This allows users to easily re-enable password authentication if desired.

We still need to disable the default `AuthorizedKeysFile` directive, since the `Include` directive appears after it in `sshd_config`.

On Fedora 31, the `sshd_config.d` fragments will be ignored, so continue to edit `sshd_config` there.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -77,13 +77,20 @@ postprocess:
     for x in /etc/yum.repos.d/*modular.repo; do
       sed -i -e 's,enabled=[01],enabled=0,' ${x}
     done
-  # Read the Ignition 2 and Afterburn SSH key fragments, pending better tooling
+  # Disable the default-enabled AuthorizedKeysFile directive so we can
+  # override it in a config fragment
   # https://github.com/coreos/fedora-coreos-tracker/issues/139
+  # https://bugzilla.redhat.com/1824913
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
     sed -i 's/^AuthorizedKeysFile[[:blank:]]/#&/' /etc/ssh/sshd_config
-    echo -e '\n# Read authorized_keys fragments written by Ignition and Afterburn\nAuthorizedKeysFile .ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn' >> /etc/ssh/sshd_config
+    . /etc/os-release
+    if [ "${VERSION_ID}" = 31 ]; then
+      # Fedora 31 doesn't read the config fragment, so we need to do this
+      # here.
+      echo -e '\n# Read authorized_keys fragments written by Ignition and Afterburn\nAuthorizedKeysFile .ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn' >> /etc/ssh/sshd_config
+    fi
 
   # Enable SELinux booleans used by OpenShift
   # https://github.com/coreos/fedora-coreos-tracker/issues/284
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -50,8 +50,8 @@ postprocess:
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
-    sed -Ei 's/^(PasswordAuthentication|PermitRootLogin)[[:blank:]]/#&/' /etc/ssh/sshd_config
-    echo -e '\n# Disable password logins by default\nPasswordAuthentication no\nPermitRootLogin prohibit-password' >> /etc/ssh/sshd_config
+    sed -Ei 's/^PasswordAuthentication[[:blank:]]/#&/' /etc/ssh/sshd_config
+    echo -e '\n# Disable password logins by default\nPasswordAuthentication no' >> /etc/ssh/sshd_config
   # Users shouldn't be configuring `rpm-ostreed.conf`
   # https://github.com/coreos/fedora-coreos-tracker/issues/271
   - |
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -44,14 +44,17 @@ postprocess:
       echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/212\nreporting.enabled = false' > /etc/fedora-coreos-pinger/config.d/95-disable-on-dev.toml
       echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/212\nupdates.enabled = false' > /etc/zincati/config.d/95-disable-on-dev.toml
     fi
-  # Disable SSH password logins by default
-  # Move to overlay once sshd_config fragments are supported
+  # Disable SSH password logins by default on Fedora 31
+  # On newer Fedora we handle this via sshd_config.d
   # https://github.com/coreos/fedora-coreos-tracker/issues/138
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
-    sed -Ei 's/^PasswordAuthentication[[:blank:]]/#&/' /etc/ssh/sshd_config
-    echo -e '\n# Disable password logins by default\nPasswordAuthentication no' >> /etc/ssh/sshd_config
+    . /etc/os-release
+    if [ "${VERSION_ID}" = 31 ]; then
+      sed -Ei 's/^PasswordAuthentication[[:blank:]]/#&/' /etc/ssh/sshd_config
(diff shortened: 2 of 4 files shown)


jamescassell:
Don't suppose it would work on RHEL 8... But glad to see sshd finally has that support.

manifests/fedora-coreos-base.yaml
```diff
@@ -77,13 +77,13 @@ postprocess:
     for x in /etc/yum.repos.d/*modular.repo; do
       sed -i -e 's,enabled=[01],enabled=0,' ${x}
     done
-  # Read the Ignition 2 and Afterburn SSH key fragments, pending better tooling
+  # Disable the default-enabled AuthorizedKeysFile directive so we can
+  # override it in a config fragment
```
dustymabe, line 81: I wish sshd had the concept of being able to override something without having to blank it out first

dustymabe (approved):
LGTM

dustymabe:
@jlebon are we safe to merge things to next-devel without fear of them getting stomped on by config-bot ?

manifests/fedora-coreos-base.yaml
bgilbert: Well, it does.  But the `Include` directive would need to be at the beginning of `sshd_config`, not the end.

jlebon:
Right, this will indeed get stomped on by config-bot. See https://github.com/coreos/fedora-coreos-config/pull/180#issuecomment-534697400. As mentioned in https://github.com/coreos/fedora-coreos-config/pull/180#issuecomment-534706218, I think folding the overlay stuff into the manifest would be a cleaner way to solve this.

For now though... one hack is to make the postprocess scripts you're modifying just conditionalize on `VERSION_ID` and for the `05core` bits, just emit those directly from `manifest.yaml`?

bgilbert:
Updated, rebased on testing-devel, and tested on top of both testing-devel and next-devel.

manifests/fedora-coreos-base.yaml
dustymabe: hmm I think that is a misconfiguration since sshd will use the first value it sees for any particular configuration key (i.e. first one wins VS last one wins) the *Include* should probably go at the top of sshd_config or at least before any uncommented out values in sshd_config. cc @Jakuje - WDYT?

manifests/fedora-coreos.yaml
```diff
@@ -44,14 +44,16 @@ postprocess:
...
+  # On newer Fedora we handle this via sshd_config.d
   # https://github.com/coreos/fedora-coreos-tracker/issues/138
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
-    sed -Ei 's/^(PasswordAuthentication|PermitRootLogin)[[:blank:]]/#&/' /etc/ssh/sshd_config
-    echo -e '\n# Disable password logins by default\nPasswordAuthentication no\nPermitRootLogin prohibit-password' >> /etc/ssh/sshd_config
+    if grep -q VERSION_ID=31 /etc/os-release; then
+      sed -Ei 's/^(PasswordAuthentication|PermitRootLogin)[[:blank:]]/#&/' /etc/ssh/sshd_config
+      echo -e '\n# Disable password logins by default\nPasswordAuthentication no\nPermitRootLogin prohibit-password' >> /etc/ssh/sshd_config
```
Jakuje: `PermitRootLogin prohibit-password` is default since Fedora 31: https://fedoraproject.org/wiki/Changes/DisableRootPasswordLoginInSshd

manifests/fedora-coreos-base.yaml
```diff
@@ -77,13 +77,18 @@ postprocess:
...
   # https://github.com/coreos/fedora-coreos-tracker/issues/139
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
     sed -i 's/^AuthorizedKeysFile[[:blank:]]/#&/' /etc/ssh/sshd_config
-    echo -e '\n# Read authorized_keys fragments written by Ignition and Afterburn\nAuthorizedKeysFile .ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn' >> /etc/ssh/sshd_config
+    if grep -q VERSION_ID=31 /etc/os-release; then
```
jlebon: `^VERSION_ID=31$` maybe?

Though I think usually one would usually just source `/etc/os-release` and use `VERSION_ID` directly.

manifests/fedora-coreos-base.yaml
Jakuje: You are right. I took the configuration file from upstream and missed that this particular option is there already and it might be an issue for you. The include is in the end mostly because it was the same way in ssh_config and it was not a problem (but that one is really empty except for comments). For Fedora 32 it is probably late to change this but feel free to open Fedora bug for this.

overlay.d/05core/etc/ssh/sshd_config.d/20-authorized-keys.conf
```diff
@@ -0,0 +1,5 @@
+# This file is ignored on Fedora 31.
```
jlebon, line 1: Ahh right, makes sense. I think we can drop these headers. We're moving off of f31 soon-ish anyway. :)

This is fine too, but we'll probably want to drop it at some point after we move into f32.

Jakuje (commented):
Otherwise LGTM

jlebon (approved):
Some comments, but LGTM as is too!

manifests/fedora-coreos-base.yaml
dustymabe: ok will open a fedora bug and we'll workaround however we need to for Fedora 32.

dustymabe: https://bugzilla.redhat.com/show_bug.cgi?id=1824913

@bgilbert - do we want to link to that bug in the comment here?

overlay.d/05core/etc/ssh/sshd_config.d/20-authorized-keys.conf
bgilbert: Agreed that we'll want to drop eventually.  I'd like to leave the headers for now, to avoid confusing users.

bgilbert:
Updated!

manifests/fedora-coreos-base.yaml
```diff
@@ -77,13 +77,20 @@ postprocess:
...
+  # override it in a config fragment
   # https://github.com/coreos/fedora-coreos-tracker/issues/139
+  # https://bugzilla.redhat.com/1824913
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
     sed -i 's/^AuthorizedKeysFile[[:blank:]]/#&/' /etc/ssh/sshd_config
-    echo -e '\n# Read authorized_keys fragments written by Ignition and Afterburn\nAuthorizedKeysFile .ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn' >> /etc/ssh/sshd_config
+    . /etc/os-release
+    if [ "${VERSION_ID}" = 31 ]; then
```
jlebon, line 89: Minor:

```suggestion
    if [ "${VERSION_ID}" -lt 32 ]; then
```

This isn't relevant *right now*, but will be if e.g. we set up the rawhide stream tracking f33 and this snippet is still here.

manifests/fedora-coreos.yaml
```diff
@@ -44,14 +44,17 @@ postprocess:
...
+  # Disable SSH password logins by default on Fedora 31
+  # On newer Fedora we handle this via sshd_config.d
   # https://github.com/coreos/fedora-coreos-tracker/issues/138
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
-    sed -Ei 's/^PasswordAuthentication[[:blank:]]/#&/' /etc/ssh/sshd_config
-    echo -e '\n# Disable password logins by default\nPasswordAuthentication no' >> /etc/ssh/sshd_config
+    . /etc/os-release
+    if [ "${VERSION_ID}" = 31 ]; then
```
jlebon, line 54: And here too.

manifests/fedora-coreos-base.yaml
bgilbert: The test gates the legacy behavior, and we're not going to add any older Fedora versions that require it.  So this should be fine?

jlebon: Yup, you're right. I got my conditions mixed up!

jlebon (approved):

dustymabe (approved):
LGTM

PR #350, opened 2020-04-15
overlay/05core: for DM root-device get path from lsblk, instead of using /dev/disk/by-label/root

nikita-dubrovskii:


cgwalters:
This is part of dm-multipath support for the rootfs?  See discussion in https://github.com/coreos/ignition-dracut/issues/154

nikita-dubrovskii:
> This is part of dm-multipath support for the rootfs? 

Yes. 
Btw, i made one more PR - https://github.com/coreos/ignition-dracut/pull/169

PR #351, opened 2020-04-15
[next-devel] lockfiles: remove for now

jlebon:
F32 is more or less frozen right now. And it'd be really nice if we had
*something* f32-based in time for the f32 release. So let's just drop
all the lockfiles for now so we can do an unlocked build and avoid all
the issues that are still being fixed.

Our overrides should all already be in the pool, and since they're still
the highest NEVRA, they still get picked.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1352

dustymabe (approved):
LGTM

miabbott (approved):

bgilbert:
For the record, what issues does this address that couldn't be fixed by bumping the lockfiles to the NEVRAs that an unlocked build would have picked anyway?

jlebon:
> For the record, what issues does this address that couldn't be fixed by bumping the lockfiles to the NEVRAs that an unlocked build would have picked anyway?

Basically cosa today will automatically try to enforce lockfiles if they're present in the repo. But we also don't want to tag packages into the pool because of https://github.com/coreos/fedora-coreos-tracker/issues/454. Which means that if we have a lockfile, there's no guarantee that we'll be able to actually perform that build in the pipeline. So for now, we're just banking on the fact that f32 doesn't change much anyway and that CI will give us coverage.

Anyway, this is all super messy and there's different ways to do this, but the bottom line is this is a temporary hack until we get the rpm-ostree lockfile patches in (https://github.com/coreos/rpm-ostree/pull/1858 and https://github.com/coreos/rpm-ostree/pull/2058).

Anyway, now that https://github.com/coreos/coreos-assembler/pull/1352 has merged, I restarted CI here to make sure we're green before merging!

jlebon:
For posterity, note lockfiles were added back in https://github.com/coreos/fedora-coreos-config/pull/370.

PR #352, opened 2020-04-15
overlay: lower kernel console logging level on Live ISO

dustymabe:
We are trying to support a workflow where a user can interactively
do an install and use TUI interfaces for configuring networking.
The increased user experience provided by a TUI is completely lost
if we have audit messages scrolling on the console every 10 seconds
or so [1]. This change changes the kernel logging level on the
console to be WARNING and not DEBUG.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
@@ -32,6 +32,18 @@ ExecStart=-/sbin/agetty --autologin core -o '-p -f core' ${args} %I \$TERM
 EOF
 }
 
+# We can remove this when we fix https://github.com/coreos/fedora-coreos-tracker/issues/220
+silence_audit_on_console() {
+    mkdir -p /run/sysctl.d
+    cat <<'EOF' > /etc/sysctl.d/20-coreos-autologin-kernel-printk.conf
+# Raise console message logging level from DEBUG (7) to WARNING (4)
+# so that audit messages don't get interspersed on the console that
+# may frustrate a user trying to interactively do an install with
+# nmtui and coreos-installer.
+kernel.printk=4
+EOF
+}
+
 # Only allow automatic autologin on live systems
 if [ ! -e /run/ostree-live ]; then
     exit 0
@@ -53,3 +65,6 @@ fi
 write_dropin "getty@.service"        "--noclear"
 # Also autologin on serial console if someone enables that
 write_dropin "serial-getty@.service" "--keep-baud 115200,38400,9600"
+
+# We can remove this when we fix https://github.com/coreos/fedora-coreos-tracker/issues/220
+silence_audit_on_console


overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
```diff
@@ -32,6 +32,18 @@ ExecStart=-/sbin/agetty --autologin core -o '-p -f core' ${args} %I \$TERM
 EOF
 }
 
+# We can remove this when we fix https://github.com/coreos/fedora-coreos-tracker/issues/220
+silence_audit_on_console() {
```
jlebon, line 36: We discussed this in the meeting, though IMO we should reframe this PR as just lowering verbosity for the sake of the UX, not specifically targeting audit messages only. IOW, I think it'd make sense to do this even if https://github.com/coreos/fedora-coreos-tracker/issues/220 weren't an issue.

jlebon (approved):
One comment, but LGTM as is too!

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
jlebon: So then, this function name would instead be e.g. `reduce_kernel_logging_on_console`.

dustymabe: I think we don't really know if we want to permanently reduce logging on the console for this case or not. Basically the question is: "when https://github.com/coreos/fedora-coreos-tracker/issues/220 is fixed, do we have a need for continued reduced kernel console logging?" I don't know the answer to that question. When I installed auditd (which means no audit messages to the console) I didn't notice any more messages getting printed to the screen, which would mean if we fix 220 then we can remove this code. I guess we'd need to wait and see to get the answer to those questions and at that point we can update this code to be more appropriate.

bgilbert (approved):

dustymabe:
great - i'm getting an SELinux denial with `/run/` 

```
Apr 15 19:55:29 localhost audit[769]: AVC avc:  denied  { read } for  pid=769 comm="systemd-sysctl" name="20-coreos-autologin-kernel-printk.conf" dev="tmpfs" ino=18569 scontext=system_u:system_r:systemd_sysctl_t:s0 tcontext=system_u:object_r:init_var_run_t:s0 tclass=file permissive=0
```

I'm going to try with `/etc/` (since it's a live system it's all ephemeral anyway).

bgilbert:
Do we need to add a relabel for that path in `/run`?

jlebon:
This looks like the policy needs to be tweaked to allow `systemd-sysctl` access to `var_run_t`. Or alternatively a rule needs to be added for `/run/sysctl.d` to have a better label. But yeah, dropping it in `/etc` *should* work.

dustymabe:
> I'm going to try with `/etc/` (since it's a live system it's all ephemeral anyway).

It works in `/etc/`.

> This looks like the policy needs to be tweaked to allow `systemd-sysctl` access to `var_run_t`. Or alternatively a rule needs to be added for `/run/sysctl.d` to have a better label. But yeah, dropping it in `/etc` _should_ work.

Right. Even if I created the file manually after the system was up the
denial still happened which leads me to think we either need 

- a policy that defines a different ontext for files that go in `/run/sysctl.d`
- OR a policy that changes what contexts systemd-sysctl can access

So probably an SELinux bug.

dustymabe:
> So probably an SELinux bug.

i.e. a policy bug. I opened this to see if we can get to the bottom of it: https://github.com/fedora-selinux/selinux-policy/issues/348

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
dustymabe: you were right. see https://github.com/coreos/fedora-coreos-config/pull/368#issuecomment-620698885

PR #354, opened 2020-04-17
dracut/ignition-ostree: Regenerate UUIDs for /boot and /

cgwalters:
This is a general best practice; the intention of filesystem
UUIDs is that they're unique.  It helps backup systems and the like
if we change this.

But in the future, we may also switch to using these UUIDs for subsequent
boots; see: https://github.com/coreos/fedora-coreos-tracker/issues/465

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -euo pipefail
+# https://github.com/coreos/fedora-coreos-tracker/issues/465
+# coreos-assembler generates disk images which are installed bit-for-bit
+# or booted directly in the cloud.
+# Generate new UUID on firstboot; this is general best practice, but in the future
+# we may use this for mounting by e.g. adding a boot=<uuid> and root=<uuid> kernel args.
+# Note this code should be changed when we land https://github.com/coreos/fedora-coreos-tracker/issues/94
+# so that we only generate a new UUID if we're using the default one.
+
+label=$1
+
+# Keep this in sync with https://github.com/coreos/coreos-assembler/blob/e3905fd2e138de04184c1cd86b99b0fd83cbe5cf/src/create_disk.sh#L17
+bootfs_uuid="96d15588-3596-4b3c-adca-a2ff7279ea63"
+rootfs_uuid="910678ff-f77e-4a7d-8d53-86f2ac47a823"
+
+target=/dev/disk/by-label/${label}
+if ! [ -b "${target}" ]; then
+  echo "$0: Failed to find block device ${target}" 1>&2
+  exit 1
+fi
+
+eval $(blkid -o export ${target})
+case "${label}" in
+  root) orig_uuid="${rootfs_uuid}"; orig_type=xfs ;;
+  boot) orig_uuid="${bootfs_uuid}"; orig_type=ext4 ;;
+  *) echo "unexpected ${label}"; exit 1 ;;
+esac
+
+if [ "${TYPE}" == "${orig_type}" ] && [ "${UUID}" == "${orig_uuid}" ]; then
+  case "${TYPE}" in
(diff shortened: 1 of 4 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-boot.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Ignition OSTree: Regenerate filesystem UUID (boot)
+DefaultDependencies=false
+ConditionPathExists=/usr/lib/initrd-release
+ConditionKernelCommandLine=ostree
+ConditionPathExists=!/run/ostree-live
+# We run pretty early
+Before=coreos-copy-firstboot-network.service
```
cgwalters, line 8: If we switch to having a `boot.mount` in the initramfs then this would just be:
`Before=boot.mount`.

dustymabe:
> If we switch to having a `boot.mount` in the initramfs then this would just be:
> `Before=boot.mount`.

Yeah I'm in favor of a `boot.mount` I think.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-boot.service
```diff
@@ -0,0 +1,19 @@
...
+Before=ignition-fetch.service
```
jlebon, line 9: Minor: we don't need this ordering here. It's already implied by being ordered `Before=ignition-setup-{base,user}.service` below. (This is harmless of course, but it's a pet peeve of mine to have ordering requirements like this which seem to imply the order of all the Ignition stages aren't in fact strictly ordered.)

jlebon (approved):
Independent of https://github.com/coreos/fedora-coreos-tracker/issues/465, I think this is a good idea. LGTM overall, just one nit!

darkmuggle (approved):
I like this idea. Let's ship it.

cgwalters:
Whee, some sort of race in an unrelated test:

```
"Found device /dev/disk/by-label/boot."
"Starting Ignition OSTree: Regenerate filesystem UUID (boot)..."
"Missed 5 kernel messages"
" vda: vda1 vda2 vda3 vda4"
"audit: type=1130 audit(1589409301.519:10): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=ignition-ostree-uuid-boot comm=\"systemd\" exe=\"/usr/lib/systemd/systemd\" hostname=? addr=? terminal=? res=failed'"
"SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=ignition-ostree-uuid-boot comm=\"systemd\" exe=\"/usr/lib/systemd/systemd\" hostname=? addr=? terminal=? res=failed'"
"ignition-ostree-uuid-boot.service: Main process exited, code=exited, status=1/FAILURE"
"/usr/sbin/ignition-ostree-firstboot-uuid: Failed to find block device /dev/disk/by-label/boot"
"ignition-ostree-uuid-boot.service: Failed with result 'exit-code'."
"Failed to start Ignition OSTree: Regenerate filesystem UUID (boot)."
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
```diff
@@ -0,0 +1,49 @@
...
+bootfs_uuid="96d15588-3596-4b3c-adca-a2ff7279ea63"
+rootfs_uuid="910678ff-f77e-4a7d-8d53-86f2ac47a823"
+
+regenerate_uuid_if() {
+  local path=$1
+  shift
+
+
+}
+
```
jlebon: Extraneous?

jlebon (commented):
Looks sane to me! Just one minor comment (there was also https://github.com/coreos/fedora-coreos-config/pull/354#discussion_r411622858 from a previous review).

jlebon:
This is now part of https://github.com/coreos/fedora-coreos-config/pull/503.

PR #355, opened 2020-04-17
manifest: move pool repo to lockfile-repos

jlebon:
The pool repo isn't semantically an ordinary like the others. It
contains past and present locked packages used in any Fedora CoreOS
stream. Mark it as a lockfile repo instead to tell rpm-ostree to only
use this repo for fetching locked packages and never to try to freely
depsolve with it on.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -9,7 +9,6 @@ rojig:
   summary: Fedora CoreOS testing-devel
 
 repos:
-  - fedora-coreos-pool
   # these repos are there to make it easier to add new packages to the OS and to
   # use `cosa fetch --update-lockfile`; but note that all package versions are
   # still pinned
@@ -17,6 +16,12 @@ repos:
   - fedora-updates
   - fedora-modular
   - fedora-updates-modular
+
+# All Fedora CoreOS streams share the same pool for locked files.
+# This will be in fedora-coreos.yaml in the future so it can be more easily be
+# shared between all the streams
+lockfile-repos:
+  - fedora-coreos-pool
 
 add-commit-metadata:
   fedora-coreos.stream: testing-devel


jlebon:
Requires: https://github.com/coreos/rpm-ostree/pull/2058

dustymabe (approved):
LGTM

jlebon:
OK, this is ready now! Though it requires https://github.com/coreos/coreos-assembler/pull/1387 first. (Well, technically it doesn't require it, but they'll work together in the context of the lockfile updater.)

jlebon:
OK, I kept this localized to `manifest.yaml` for now. We'll move it to `fedora-coreos.yaml` once we have https://github.com/coreos/rpm-ostree/pull/2071 so we can drop `repos` entirely from the production branches' `manifest.yaml` files.

Once this merges, I'm going to do the same thing to `next-devel` and introduce back lockfiles there.

dustymabe (approved):
LGTM

jcajka:
This breaks non-intel as there are no lock files there yet. This effectively disabled the fedora-coreos-pool repo there. @jlebon do you have any advice how to proceed?

jlebon:
> This breaks non-intel as there are no lock files there yet.

Ahh right, yeah. All builds are unlocked there, but I'm guessing you're not getting a depsolve without the pool because of https://github.com/coreos/fedora-coreos-config/pull/335#issuecomment-610634917, right?

Are you getting a successful compose with:

```diff
diff --git a/manifest-lock.overrides.s390x.yaml b/manifest-lock.overrides.s390x.yaml
new file mode 100644
index 0000000..637eb4e
--- /dev/null
+++ b/manifest-lock.overrides.s390x.yaml
@@ -0,0 +1,3 @@
+packages:
+  crypto-policies:
+    evra: 20191128-5.gitcd267a5.fc32.noarch
```

?

>  This effectively disabled the fedora-coreos-pool repo there. @jlebon do you have any advice how to proceed?

The right fix is adding lockfiles for all the arches :). But yeah, we're not there yet.

For now though, I think the right fix is simply maintaining overrides for all the arches, not just x86_64. Otherwise, even if you did get a depsolve without the pool, you'd still be missing out on all the other packages we fast-track.

Will open a PR!

jlebon:
https://github.com/coreos/fedora-coreos-config/pull/374

jcajka:
That should be it. I have not tried that on the "lock" level yet, just the repo level. I have proposed in the past to have not arch-full lock as if I correctly understand that it is really more of a "override" list and currently there is no arch dependent override(correct me if I'm wrong). It seems to me as rather tedious and unnecessary job to maintain locks in current format. IMO it seems that this should be done on the "override" repo side, only provide packages that are needed there.

jlebon:
> I have proposed in the past to have not arch-full lock as if I correctly understand that it is really more of a "override" list and currently there is no arch dependent override(correct me if I'm wrong).

I don't think it's come up yet, but we want to retain that flexibility. (This is similar to how e.g. the stream metadata can in theory support arch-dependent overrides).

> It seems to me as rather tedious and unnecessary job to maintain locks in current format.

Well, there are two bits; the base lockfiles and the overrides. The former is automated (well, not right now because of some issues, but it normally is), while the latter is manual, by definition.

> IMO it seems that this should be done on the "override" repo side, only provide packages that are needed there.

I'm not sure if I fully understand what you mean here. Are you suggesting leaving the base unlocked and only maintain an overrides repo with e.g. `priority=1` or something?

(Though in general, yes it's possible to have "locked package sets" on the yum repo side and that would solve issues and create others. We went with what best matches the current state of Fedora today. See somewhat related discussions in https://github.com/coreos/rpm-ostree/issues/415.)

PR #356, opened 2020-04-17
overlay/mounts: Mount /boot and /boot/efi ro,nodev,nosuid

cgwalters:
Ironically ostree has had support for a `ro` boot for a long time,
and only more recently did we land the [sysroot readonly](https://github.com/coreos/coreos-assembler/pull/736).

But we never actually went and made `/boot` `ro` for FCOS, so let's
do it now.

This was actually motivated by someone wanting to "security harden" RHCOS
running through a checklist saying certain mounts should be `nodev`.
Let's add `nosuid` while we're here.

### DIFF
--- a/tests/kola/basic
+++ b/tests/kola/basic
@@ -1,2 +1,20 @@
 #!/bin/bash
-exec systemctl is-enabled logrotate.service
+set -euo pipefail
+systemctl is-enabled logrotate.service
+echo "ok logrotate"
+
+validate_not_writable_mount() {
+    local mnt=$1
+    shift
+    findmnt "${mnt}" -o OPTIONS | grep -q ro
+    if test -w "${mnt}"; then
+        echo "writable ${mnt}"
+        exit 1
+    fi
+    echo "ok not writable ${mnt}"
+}
+
+validate_not_writable_mount /boot
+if test -d /boot/efi; then
+    validate_not_writable_mount /boot/efi
+fi
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -18,6 +18,10 @@ if findmnt --fstab /boot &>/dev/null; then
     exit 0
 fi
 
+# Some security hardening guidelines like to see this.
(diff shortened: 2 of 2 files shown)


tests/kola/basic
```diff
@@ -1,2 +1,20 @@
...
+        echo "writable ${mnt}"
+        exit 1
+    fi
+    echo "ok not writable ${mnt}"
+}
+
+validate_not_writable_mount /boot
+if test -d /boot/efi; then
+    validate_not_writable_mount /boot/efi
+fi
```
dustymabe, line 20: should this go in a separate file?

dustymabe: i.e. a different test and maybe add some text about why we have this test

dustymabe:
If we mount boot read only we probably need to delete at least: 

https://github.com/coreos/fedora-coreos-config/blob/49de0c72b082c3e3b6bb1e9015691246a2fd8727/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh#L60-L63

tests/kola/basic
cgwalters: But then we end up booting a whole other VM for it.

cgwalters: (What we really want are nondestructive tests and teach kola to reuse existing VMs for them)

cgwalters:
> If we mount boot read only we probably need to delete at least:

Hmm; and have the files leak?  We already have `ignition-firstboot-complete.service` which is remounting writable (privately/temporarily) too from the real root.  Maybe we should coalesce the "clean up stuff from /boot" there.

tests/kola/basic
dustymabe: yeah ok that's a good reason to keep it in the same file. Can we add some text at the top about an RFE for nondestructive tests (open issue somewhere?) and then we can break these out once that is in place,

It would still be nice to have a comment at the top of this particular test you are adding about 
`# in coreos-boot-mount-generator  we mount boot readonly, verify it is readonly`.

dustymabe:
> Maybe we should coalesce the "clean up stuff from /boot" there.

SGTM

cgwalters:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/659

PR #363, opened 2020-04-21
40ignition-ostree: add service for deleting /etc enablements

### LINKED ISSUE #392: Unable to disable zincati.service using Ignition
As reported by a user in [a discussion forum thread](https://discussion.fedoraproject.org/t/how-do-we-kick-off-manual-updates/17089/12?u=dustymabe) for some reason disabling `zincati.service` via Ignition doesn't seem to be working right. There is probably a corner case in here that isn't being accounted for somehow.

Here is my fcc:

```
variant: fcos
version: 1.0.0
systemd:
  units:
    - name: serial-getty@ttyS0.service
      dropins:
      - name: autologin-core.conf
        contents: |
          [Service]
          # Override Execstart in main unit
          ExecStart=
          # Add new Execstart with `-` prefix to ignore failure
          ExecStart=-/usr/sbin/agetty --autologin core --noclear %I $TERM
          TTYVTDisallocate=no
    - name: zincati.service
      enabled: false
```

Here is what I see on the booted system:

```
[core@localhost ~]$ systemctl is-enabled zincati
enabled
[core@localhost ~]$ systemctl is-active zincati
active
[core@localhost ~]$ 
[core@localhost ~]$ grep zincati /etc/systemd/system-preset/20-ignition.preset
disable zincati.service
[core@localhost ~]$ grep zincati /usr/lib/systemd/system-preset/40-coreos.preset
enable zincati.service
```

So the `disable` preset did get put in `/etc/` appropriately but for some reason it wasn't applied properly. If I ask systemd to re-evaluate it seems to do the right thing:

```
[core@localhost ~]$ sudo systemctl preset zincati.service
Removed /etc/systemd/system/multi-user.target.wants/zincati.service.
[core@localhost ~]$ systemctl is-enabled zincati
disabled
```

**Version Info**:

```
[core@localhost ~]$ rpm-ostree status
State: idle
AutomaticUpdates: disabled
Deployments:
● ostree://fedora:fedora/x86_64/coreos/testing
                   Version: 31.20200210.2.0 (2020-02-10T21:52:48Z)
                    Commit: 3e4a6a48ed8d6817081c902bc2aa2bbe9df3302e659d4f42d933f9abb22914e8
              GPGSignature: Valid signature by 7D22D5867F2A4236474BF7B850CB390B3C3359C4
[core@localhost ~]$ rpm -q ignition systemd
ignition-2.1.1-5.git40c0b57.fc31.x86_64
systemd-243.6-1.fc31.x86_64
```

jlebon:
A lot of backstory on this, but essentially right now, we always bake a
run of `systemctl preset-all` into the OSTree because upgrading hosts
rely on these links for service enablement.

In hindsight, we should've just stuck with pure systemd preset only as
canonicaly from the get go, though it's a bit difficult now to
transition from one to the other without breaking things. (Though I'll
note not impossible, since we do have update barriers which could allow
us to e.g. run a script to restore lost symlinks).

For now though, let's at least fix the ability to disable services,
which is a pretty big gap in our Ignition configuration story right now.

Related: https://github.com/systemd/systemd/pull/15205
Related: https://github.com/coreos/fedora-coreos-config/pull/77

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/392

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-delete-etc-services.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-delete-etc-services.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=Delete /etc Service Enablements
+DefaultDependencies=false
+ConditionKernelCommandLine=|ostree
+ConditionPathExists=|/run/ostree-live
+
+# Really, we don't really expect mount points for /etc, so we could just order
+# after initrd-root-fs.target, but meh... doesn't hurt to pretend we support
+# that.
+After=ignition-mount.service
+
+# But *before* any user-specified enablement
+Before=ignition-files.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+# use glob instead of nuking /etc/systemd/system itself so that we don't end up
+# recreating it from the initrd as unlabeled for RHCOS
+ExecStart=/usr/bin/sh -c "rm -rf /sysroot/etc/systemd/system/*"
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -68,4 +68,6 @@ install() {
     inst_script "$moddir/coreos-growpart" /usr/libexec/coreos-growpart
 
     inst_script "$moddir/coreos-relabel" /usr/bin/coreos-relabel
+
+    install_ignition_unit ignition-ostree-delete-etc-services.service
 }


jlebon:
This is a short-term solution to https://github.com/coreos/fedora-coreos-tracker/issues/392 while we figure out what the long-term strategy should be (discussions ongoing in https://github.com/systemd/systemd/pull/15205).

(To re-iterate, as mentioned in https://github.com/systemd/systemd/pull/15205#issuecomment-617418106, yes, this is really awkward to basically undo what the OSTree commit is baking in.)

jlebon:
> In hindsight, we should've just stuck with pure systemd preset only as
canonicaly from the get go, though it's a bit difficult now to
transition from one to the other without breaking things. (Though I'll
note not impossible, since we do have update barriers which could allow
us to e.g. run a script to restore lost symlinks).

This might actually not be as hard as I initially imagined. I didn't want to write systemd services for this that would muck around with analyzing a previous deployment's `/etc` and the ambiguity there, but I think we could avoid that and do it in two steps:
1. an update barrier with a systemd service which runs right before `ostree-finalize-staged.service` on shutdown and uses `ostree admin config-diff` to find all added and modified files in `/etc/systemd/system` and copies those to e.g. `/etc/systemd/system.new`.
2. another update barrier where we switch to presets as canonical and we have a systemd service which runs in the initrd that looks for `/etc/systemd/system.new`, restores it, and deletes it.

Still a sensitive operation of course, given that failure potentially means a broken boot. Though it'd be a nice legacy cleanup.

jlebon:
Thoughts on just merging this until https://github.com/systemd/systemd/pull/15205 is decided (and possibly backported to Fedora/RHEL8)?

cgwalters:
To be honest I am scared of this :smile: 
I believe you it works but we should be sure multiple people review it and think through it carefully and we've tested it locally.

Not to mention that the systemd PR seemed to come to a conclusion and looks likely to merge.  (Once it does, we can just e.g. detect the systemd version in postprocess and not do the "copy units to /etc" right?)

Plus we have a workaround for the zincati thing today (it's not a difficult workaround, but it's also admittedly not an *obvious* workaround - people are going to keep hitting that issue, but hopefully web search takes them to it quickly)

dustymabe:
Yeah I'm a bit wary. Since we have an easy workaround and hopefully some momentum upstream maybe let's just see if we can get that in sooner than later. Obviously if upstream lingers we can revisit.

jlebon:
Sure, we can wait some more. I think there was a misconception on my part on how bad this actually is.

jlebon:
Looks like upstream is lingering. And the service disablement issue continues to be UX wart. So I think we should consider doing something here.

I actually don't think this PR is that scary. The main thing to note is that it's purely a first boot thing which means (1) lots of CI coverage, and (2) it cannot affect upgrading nodes.

jlebon:
Actually, I think another approach here is to tweak Ignition so that `enabled: false` will also make sure that no enablement symlink exists in the real root on top of emitting a `disable` in the presets.

travier:
> Actually, I think another approach here is to tweak Ignition so that `enabled: false` will also make sure that no enablement symlink exists in the real root on top of emitting a `disable` in the presets.

While I agree that this PR is closer to what we actually want in the end, doing the above seems to be safer as a first step.

jlebon:
> Actually, I think another approach here is to tweak Ignition so that `enabled: false` will also make sure that no enablement symlink exists in the real root.

This is https://github.com/coreos/ignition/issues/588.

travier:
Closing as per https://github.com/coreos/fedora-coreos-tracker/issues/392#issuecomment-1090710309

PR #367, opened 2020-04-27
Override chronyd config on Azure to use PHC

cgwalters:
See https://github.com/openshift/installer/pull/3513


cgwalters:
Totally untested (I need to dig out my Azure creds etc. - filing for discussion/arch review)

cgwalters:
OK I fixed all of the bugs I found by manually hacking `ignition.platform.id=azure` in `cosa run` (qemu) and rebooting.  Just for general interest I found a fast workflow for this was:

```
cosa# cp overlay.d/20azure/usr/lib/systemd/system-generators/azure-phc tmp
cosa# cosa run
guest# mount -o remount,rw /usr && cp /var/mnt/workdir-tmp/azure-phc /usr/lib/systemd/system-generators
guest# restorecon -v /usr/lib/systemd/system-generators/azure-phc
guest# systemctl reboot
```

cgwalters:
The other alternative is to change coreos-assembler to have some mechanism to change the default deployment per-platform; it'd be very much like what the MCO has in its templates.  Something like in `src/config` we'd have `overlay-${platform}.d`, e.g.

```
overlay-azure.d/20phc/etc/chrony.conf.d/azure-phc.conf
```

And coreos-assembler would automatically overlay config files from there into the disk image during e.g. `buildextend-azure`.  One downside of that is it wouldn't show up on upgrades.  I think I prefer this way myself but could be convinced the other way - (either way is ugly).

jlebon:
Why not just add `include /etc/chrony.d/*.conf` to the chrony config ourselves and use it? (See https://github.com/openshift/installer/pull/3513#issuecomment-620130892.)

cgwalters:
> Why not just add include /etc/chrony.d/*.conf to the chrony config ourselves and use it?

`/etc` is persistent and will show up as admin-created changes; I was trying to avoid that because "source of truth" here is derived from kernel cmdline + no changes to `/etc/chrony.conf`.

I tried adding `/run/chrony.conf.d` but we need to add SELinux labeling for that to make it `etc_t`...

jlebon:
> `/etc` is persistent and will show up as admin-created changes; I was trying to avoid that because "source of truth" here is derived from kernel cmdline + no changes to `/etc/chrony.conf`.

Hmm, not sure I follow. The suggestion in https://github.com/openshift/installer/pull/3513#issuecomment-620130892 is to add `include /etc/chrony.d/*.conf` at compose time from a postprocess script. So it wouldn't show up as a modified file, right?

The `/etc/chrony.d/azure.conf` would, but that's the same as the bunch of other cloud-specific paths which write to `/etc`. (E.g. the recent networking propagation bits or Afterburn's hostname service).

cgwalters:
> The /etc/chrony.d/azure.conf would, but that's the same as the bunch of other cloud-specific paths which write to /etc. (E.g. the recent networking propagation bits or Afterburn's hostname service).

I haven't looked at the details but I think those would be best done in `/run` as well. 

Take for example [the NM hostname](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/419#note_473418).

The way I think of this is that if e.g. a service is disabled later, it avoids having cruft accumulate in the root (after a reboot).

On the other hand it just doesn't matter in most cases either I'd admit.

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
```diff
@@ -0,0 +1,34 @@
...
+    exit 0
+fi
+
+if ! cmp {/usr,}/etc/chrony.conf; then
+    echo "/etc/chrony.conf is modified; not changing the default"
+    exit 0
+fi
+(echo '# Generated by coreos-azure-ptp'
+ cat /etc/chrony.conf
+ echo '# Generated by coreos-azure-ptp, see also https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync'
```
miabbott: nit: the generator filename is `coreos-azure-phc`

cgwalters: Yeah I got PTP and PHC confused and wrote this whole patch the first time using the former and only noticed I got it wrong just before pushing the PR...

cgwalters:
Currently waiting on an Azure dev account to test, though if that takes too long I may just try this out with my personal one.

cgwalters:
OK, dusted off my Azure creds and tested this:
`Apr 28 15:39:21 walters-fcos chronyd[728]: Selected source PHC0`

jlebon:
> The way I think of this is that if e.g. a service is disabled later, it avoids having cruft accumulate in the root (after a reboot).

Right OK, that makes sense. I agree our ideal should be to try to keep `/etc` for admin configurations (I guess this goes with https://github.com/coreos/fedora-coreos-tracker/issues/52).

OK, let's circle back to this:

> I tried adding `/run/chrony.conf.d` but we need to add SELinux labeling for that to make it `etc_t`...

Can't we still do this and use the generator logic to `chcon --reference`?

cgwalters:
> Can't we still do this and use the generator logic to chcon --reference?

Probably...feels a bit unclean though since the role of a systemd generator should be more to just generate units than affect system state, although I guess we could generate a unit that does an `ExecStart=/bin/sh -c 'mkdir -p /run/chrony.conf.d && chcon --reference=/etc /run/chrony.conf.d'` or so...

cgwalters:
OK I just carefully read https://bugzilla.redhat.com/show_bug.cgi?id=1828434#c2 and I think we need to change this.

cgwalters:
OK updated for that and re-tested.  I think the fact we need to change the base argues for not trying to do configuration drop-ins.

lucab:
>  I think the fact we need to change the base argues for not trying to do configuration drop-ins.

For configuration fragments to work in this scenario, chrony has to support real drop-ins (i.e. logic for same-filename overrides). Instead, it seems to lack that feature and to only support include directives.

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
```diff
@@ -0,0 +1,33 @@
...
+ echo '# and https://github.com/openshift/installer/pull/3513'
+ sed -e s,'^pool,#pool,' < /etc/chrony.conf
+ echo 'refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0'
+) > "${confpath}"
+# Policy doesn't allow chronyd to read run_t
+chcon --reference=/etc/chrony.conf "${confpath}"
+
+UNIT_DIR="${1:-/tmp}"
+
+unitconfpath="${UNIT_DIR}/chronyd.service.d/coreos-azure-phc.conf"
```
sinnykumari, line 34: questions to improve my systemd understanding: 
1. Does systemd looks for services available in /tmp/ or /var/tmp/ by default or we do some magic in coreos-assmbler?
2. Is this service going to just override defined section in /etc/systemd/system/multi-user.target.wants/chronyd.service or will run as independent service?

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -euo pipefail
+# See https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync
+# and https://github.com/openshift/installer/pull/3513
+
+# Yeah this isn't a completely accurate kernel argument parser but
+# we don't have one shared across shell services at the moment.
+if ! grep -qF ignition.platform.id=azure /proc/cmdline; then
```
jlebon: Maybe at least we can do

```suggestion
if ! grep -qF ' ignition.platform.id=azure' /proc/cmdline; then
```

?

There's also `--word-regexp` which would've been nice here, but I think this breaks on dots.

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
```diff
@@ -0,0 +1,33 @@
...
+    exit 0
+fi
+
+if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
+    echo "/etc/chrony.conf is modified; not changing the default"
+    exit 0
+fi
```
jlebon: I think that makes sense overall, but... it's not obvious either that changing some totally unrelated setting will cause chrony to stop using the PHC. I guess this comes back to chrony not properly supporting dropins though. :(

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
```diff
@@ -0,0 +1,33 @@
...
+# and https://github.com/openshift/installer/pull/3513
+
+# Yeah this isn't a completely accurate kernel argument parser but
+# we don't have one shared across shell services at the moment.
+if ! grep -qF ignition.platform.id=azure /proc/cmdline; then
+    exit 0
+fi
+
+if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
+    echo "/etc/chrony.conf is modified; not changing the default"
```
jlebon: Related to my previous comment, maybe we can print something in the positive case too, e.g. `Changing chrony to use Azure PTP hardware clock`? And in the negative test here `Not changing chrony to use Azure PTP hardware clock; /etc/chrony.conf is modified` or something.

jlebon: Also, shouldn't this be writing to `/dev/kmsg`?

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
jlebon: Re. 1., systemd passes the directory in which to output units as the first argument. I suspect the `/tmp` fallback is for testing it manually.

Re. 2., it adds an override to the existing `chronyd.service`. That way e.g. it shows up in `systemctl status chronyd.service` too.

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
cgwalters: Yep.

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
cgwalters: We need to fix systemd to have sane logging for generators, but yes.

jlebon (approved):
LGTM! Looks like CentOS CI is hitting issues right now though. Trying to figure that out.

jlebon:
Oh hey, looks like this PR didn't hit the issues. It has good timing I guess!

overlay.d/20azure-chrony/usr/lib/systemd/system-generators/coreos-azure-phc
sinnykumari: understood, thank you!

PR #369, opened 2020-04-28
tests/manual: update the net test script to support RHCOS

dustymabe:
Update the net test to support RHCOS and also add some new tests.


dustymabe:
Yes. At least some of this coverage needs to be ported into kola tests so that they run as part of our CI runs.

lucab:
> At least some of this coverage needs to be ported into kola tests so that they run as part of our CI runs.

Unless we are in a hurry, this looks like a great time and incentive to move this "testing logic for RHCOS" out of this "configuration repo for FCOS", doesn't it?

cgwalters:
> Unless we are in a hurry, this looks like a great time and incentive to move this "testing logic for RHCOS" out of this "configuration repo for FCOS", doesn't it?

Hmm...there's a lot of shared code though.

The subject of a separate repo for tests also came up in https://github.com/coreos/coreos-assembler/pull/1407#discussion_r417278585

dustymabe:
I don't have any magic bullets here on where this should live. I don't know of a better place right now.

lucab:
I may have misunderstood something. Wasn't `kola` the planned place where these kinds of tests (and setup logic too) would live?
My understanding is that it was here temporarily as a manual stopgap during development, to be moved into automated CI once we were comfortable with the direction of the underlying logic.

cgwalters:
Moving to the coreos-assembler repository does make more sense than this repo I would agree.  (But, conflicts a bit with the [exttest](https://github.com/coreos/coreos-assembler/pull/1252) direction of having cosa be a mechanism and not as much policy)

cgwalters:
https://github.com/coreos/fedora-coreos-tracker/issues/340 effectively covers "our testing and libvirt".

PR #370, opened 2020-04-29
[next-devel] Add back lockfiles

jlebon:
We previously removed these because rpm-ostree couldn't deal with an
influx of f32 packages being added to the pool repo and ending up in
streams still based on f31. This was fixed now (see [1]). So let's add
back lockfiles here.

[1] coreos/fedora-coreos-tracker#454

### DIFF
--- a/manifest-lock.overrides.x86_64.yaml
+++ b/manifest-lock.overrides.x86_64.yaml
@@ -1,0 +1,28 @@
+packages:
+  # Cherry-pick rpm-ostree git snapshot to resolve issues with /read-only
+  # sysroot. We can nuke this when we cut a new release.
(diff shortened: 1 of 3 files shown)


jlebon:
Yay, all green!

dustymabe:
does this mostly address https://github.com/coreos/fedora-coreos-tracker/issues/463 ?

dustymabe (approved):
LGTM

jlebon:
> does this mostly address coreos/fedora-coreos-tracker#463 ?

Yeah, I think so. Though as you can see from the override, there's still coreos-installer that we're pulling in from f31 since the latest release was never actually built for f32.

PR #368, opened 2020-04-28
overlay: coreos-autologin-generator: add /etc/issue.d for interactive live env

dustymabe:
This issue.d will give the user some information about what is possible
to do in the interactive live environment, including some information
about installing to disk.

PR #371, opened 2020-04-29
overlay: coreos-autologin-generator: switch to motd

dustymabe:
- Switch to using motd vs issue so all logins will get the message.
  This also overwrites the existing motd, which we want.
- Support customizing the message for RHCOS/FCOS


overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
```diff
@@ -43,24 +43,31 @@ kernel.printk=4
 EOF
 }
 
-write_interactive_live_issue() {
-    cat <<'EOF' > /etc/issue.d/80-interactive-live.issue
+write_interactive_live_motd() {
+    if grep 'RHEL' /etc/os-release &>/dev/null; then
```
cgwalters: Why not use my suggestion in https://github.com/coreos/fedora-coreos-config/pull/368#discussion_r416892640 ?

dustymabe: @cgwalters - just saw that suggestion, sorry. The reason is primarily that we need roughly the same number of characters for either Fedora or RHEL to stick within ideal character width limits of a serial console. The problem with using `${PRETTY_NAME}` is that on RHCOS that is very long:

```
$ cat /etc/os-release 
NAME="Red Hat Enterprise Linux CoreOS"
VERSION="46.82.202004290440-0"
VERSION_ID="4.6"
OPENSHIFT_VERSION="4.6"
RHEL_VERSION="8.2"
PRETTY_NAME="Red Hat Enterprise Linux CoreOS 46.82.202004290440-0 (Ootpa)"
ID="rhcos"
ID_LIKE="rhel fedora"
ANSI_COLOR="0;31"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="OpenShift Container Platform"
REDHAT_BUGZILLA_PRODUCT_VERSION="4.6"
REDHAT_SUPPORT_PRODUCT="OpenShift Container Platform"
REDHAT_SUPPORT_PRODUCT_VERSION="4.6"
OSTREE_VERSION='46.82.202004290440-0'
```

`$ID` looks like it could work but on FCOS the `$ID` is just `Fedora`. This if/else statement is the best thing I could come up with given the constraints.

cgwalters: The whole point of having `/usr/lib/os-release` is to avoid lots of random hardcoding of the operating system identifier.  I would take the word wrapping hit over hardcoding the OS name.

But anyways, we can have both by [piping to fold](https://unix.stackexchange.com/questions/25173/how-can-i-wrap-text-at-a-certain-column-size).

cgwalters: Also, I would use `/usr/lib/os-release` over `/etc/os-release` in new code.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
```diff
@@ -43,24 +43,31 @@ kernel.printk=4
...
+    else
+        OS='Fedora'
+    fi
+    # Note this intentionally overwrites the existing motd, which is
+    # blank on FCOS and populated on RHCOS.
+    cat <<EOF > /etc/motd
 
 ###########################################################################
-Welcome to the Fedora CoreOS Live environment. This system is running
+Welcome to the ${OS} CoreOS Live environment. This system is running
```
cgwalters: Or instead of `fold` to allow RHEL, break these two lines up:
```
Welcome to: ${PRETTY_NAME} (Live OS)
This system is running completely ...
```

cgwalters: Don't we already print the OS name anyways from other MOTD fragments?

cgwalters:
BTW from IRC:
> for some reason writing /etc/motd fails with `Permission denied

I think that's because this will race with the live generator generating the overlayfs mount.  Basically...it's bad practice for generators to do anything *other* than generate unit files.

So we likely need to change this to generate a unit file like:
```
[Unit]
RequiresMountsFor=/etc
[Service]
Type=oneshot
ExecStart=/usr/libexec/coreos-live-motd
```
or something.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
dustymabe: The PRETTY_NAME for RHCOS is kind of absurd. It's 60 characters by itself. I guess we could leave out the `Fedora` or `RHEL` in front of `CoreOS` here but it felt slightly wrong because we've been asking people to be specific when they say the word CoreOS.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
dustymabe: +1 for `/usr/lib/os-release` - will use that

dustymabe:
> BTW from IRC:
> 
> > for some reason writing /etc/motd fails with `Permission denied
> 
> I think that's because this will race with the live generator generating the overlayfs mount. Basically...it's bad practice for generators to do anything _other_ than generate unit files.

It turned out to be an SELinux denial because `init_t` doesn't have access to write `etc_t` so I'll probably have to do something like you suggest below  


> So we likely need to change this to generate a unit file like:
> 
> ```
> [Unit]
> RequiresMountsFor=/etc
> [Service]
> Type=oneshot
> ExecStart=/usr/libexec/coreos-live-motd
> ```
> 
> or something.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
dustymabe: > Don't we already print the OS name anyways from other MOTD fragments?

Yep. We do so we could leave it out.

dustymabe:
ok pushed up an update to just say `CoreOS` without qualification and let the other message fragments from motd or issue say RHCOS or FCOS and the version.

dustymabe:
updated git commit message:

```
overlay: coreos-autologin-generator: switch to motd

Switch to using motd vs issue so all logins will get the message.
This also overwrites the existing motd. On FCOS the /etc/motd is
empty by default but on RHCOS it is populated and we want to get
rid of that and use this message instead.

```

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
```diff
@@ -43,28 +43,46 @@ kernel.printk=4
...
+    # Write motd to a tmp file and not directly to /etc/motd because
+    # SELinux denies write from init_t to etc_t
+    cat <<EOF > /run/interactive-live-motd
 
 ###########################################################################
-Welcome to the Fedora CoreOS Live environment. This system is running
-completely from memory, making it a good candidate for hardware discovery
-and installing persistently to disk. Here is an example of running an
-install to disk via `coreos-installer`:
+Welcome to the CoreOS Live environment. This system is running completely
```
cgwalters, line 52: This is fine as is though I think we could omit "CoreOS" entirely and just say something like:
```
This is a "live" environment, running completely from memory.
You can use this e.g. for hardware discovery, or to install persistently to disk.
Here is...

```

dustymabe: Yeah. I'm torn. Do you think there is some value in people knowing it *is* the real thing though? and not some other live environment like Anaconda or just a dracut generated initramfs?

dustymabe:
ok tested this on FCOS and RHCOS and it looks good.. moving out of draft!

jamescassell:
Any support for motd.d?

dustymabe:
> Any support for motd.d?

we do support motd.d but in my tests using motd.d with our autologin doesn't seem to work. I think for motd.d you have to go through pam or something.

dustymabe:
rebased on top of latest master

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
```diff
@@ -43,28 +43,46 @@ kernel.printk=4
...
+'--copy-network' argument to 'coreos-installer install'. Please run
+'coreos-installer install --help' for more information on the possible
 install options.
 ###########################################################################
 
 EOF
+    # Create cp-interactive-live-motd.service to copy over the motd in
+    # place. Note this intentionally overwrites the existing motd, which is
+    # blank on FCOS and populated on RHCOS.
+    service="cp-interactive-live-motd.service"
```
jlebon: Minor: I'd prefix the service with `coreos-`.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
miabbott: +1 to keeping CoreOS included so folks understand they are in "the real thing"

jlebon (commented):
Hmm, this is somewhat abusing systemd generators though. ISTM there's nothing stopping us from just doing this from a regular systemd unit with `ConditionPathExists=/run/ostree-live` (and I guess `ConditionKernelCommandLine=coreos.liveiso` too) that runs `Before=systemd-user-sessions.service` and just writes out the `motd`, right?

dustymabe:
> Hmm, this is somewhat abusing systemd generators though. ISTM there's nothing stopping us from just doing this from a regular systemd unit with `ConditionPathExists=/run/ostree-live` (and I guess `ConditionKernelCommandLine=coreos.liveiso` too) that runs `Before=systemd-user-sessions.service` and just writes out the `motd`, right?

I think the attractive thing to me is to keep the logic for "autologin" together in the same file. There are 3 conditions that are checked in here: 

https://github.com/coreos/fedora-coreos-config/blob/37ab510af8539097d618b7b790b5f8d57fae1818/overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator#L70-L86

I guess we can move all of the logic in this file into a unit, but I am a bit confused about what is the difference between https://github.com/coreos/fedora-coreos-config/pull/368 and this PR to raise concerns.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
dustymabe: 👍

jlebon:
> I guess we can move all of the logic in this file into a unit, but I am a bit confused about what is the difference between #368 and this PR to raise concerns.

Oh no, it equally applied to #368 too. :)  It's just that the hacking around SELinux issues here made it more obvious. And we're writing a systemd service that doesn't have any dynamic content.

> I think the attractive thing to me is to keep the logic for "autologin" together in the same file. There are 3 conditions that are checked in here:

Yeah, I agree this is nice. Though we could keep them in the same `overlay.d` (and actually that'll allow us to group the live dracut module with those too). Re. the condition checks, we could do what Ignition does and have the generator enable an existing unit if conditions are met. I guess in general, IMO we should use proper systemd unit files if there's nothing dynamic about them.

Big picture this is fine too and we can keep discussing as a follow-up. I know it's a pain to hack on the live ISO because mksquashfs takes ages.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-autologin-generator
dustymabe: done

dustymabe:
> Big picture this is fine too and we can keep discussing as a follow-up. I know it's a pain to hack on the live ISO because mksquashfs takes ages.

yeah. I spent a lot of time yesterday on this (between the obscure SELinux issue, the no logging by default from generators, and my local ISP going out). I'd like to go work on some bigger issues 😄

jlebon (approved):

cgwalters (approved):

rfairley:
> > Any support for motd.d?
> 
> we do support motd.d but in my tests using motd.d with our autologin doesn't seem to work. I think for motd.d you have to go through pam or something.

Just to add - `login` invoked by `agetty --autologin` only supports `/etc/motd` by default. Other paths could be configured with `MOTD_FILE` in `/etc/login.defs`, but only individual filepaths can be entered there. Could be worthwhile to have `util-linux` patched at some point to support directories so that motd.d could be configured (similar as [done for `issue.d`](https://github.com/karelzak/util-linux/commit/456bcbca6b55fbed33d9f86e69a51abd0e1b8f0b#diff-d7efd2b3dbb10e54185f001dc21d43db)). That would make motd.d work by default for most login cases I think (the other major case is `ssh` where `pam_motd.so` provides the motd.d support). +1 for overwriting `/etc/motd` here though.

dustymabe:
Thanks @rfairley for explaining. Do you think that is worth a feature request upstream in `util-linux`?

rfairley:
I think so! Will open an issue there shortly.

rfairley:
Opened: https://github.com/karelzak/util-linux/issues/1034

PR #373, opened 2020-04-30
manifests: Move grub2-removals include to fedora-coreos

LorbusChris:
Move the include statement for grub2-removals from bootable-rpm-ostree
which is pulled in by fedora-coreos-base to the fedora-coreos manifest.

The exclusion of files in grub2-removals is specific to Fedora
CoreOS but not other potentially derivative operating systems:
- Fedora Silverblue uses `/etc/grub.d/10_reset_boot_success`
and `/etc/grub.d/12_menu_auto_hide`
- Fedora IoT uses `/etc/grub.d/08_fallback_counting` and
`/etc/grub.d/10_reset_boot_success`

This is a step towards making fedora-coreos-base re-usable as base
manifest for these operating systems.

### DIFF
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -23,8 +23,3 @@ packages-s390x:
 packages-x86_64:
   - grub2 grub2-efi-x64 efibootmgr shim
   - microcode_ctl
-# And remove some cruft from grub2
-arch-include:
-  x86_64: grub2-removals.yaml
-  aarch64: grub2-removals.yaml
-  ppc64le: grub2-removals.yaml
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -92,3 +92,9 @@ exclude-packages:
   # Let's make sure initscripts doesn't get pulled back in
   # https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-611566254
   - initscripts
+
+# And remove some cruft from grub2
+arch-include:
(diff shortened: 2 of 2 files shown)


dustymabe:
I don't have too strong of an opinion on this (defer to @cgwalters). This file does get included in RHCOS so there are implications there.

LorbusChris:
I can open a PR to include this in the rhcos manifest as well. I haven't checked whether grub actually ships these files in RHEL, though.

cgwalters:
```
/usr/lib/systemd/user/grub-boot-success.service
/usr/lib/systemd/user/grub-boot-success.timer
```
exists on RHEL8 and we definitely don't want it for RHCOS; it's basically desktop specific.

cgwalters:
Maybe create a new `grub2-removals.yaml` here, make FCOS include it and then when we bump RHCOS we include it there too?

cgwalters:
Or I guess we should evaluate whether we still need to strip out this stuff for CoreOS?  I think we know we don't want the menu auto hiding but IMO that should move into `fedora-release-workstation` or something.

LorbusChris:
> Maybe create a new `grub2-removals.yaml` here, make FCOS include it and then when we bump RHCOS we include it there too?

Wait, I'm not entirely following. The `grub2-removals.yaml` already exists here.

Don't we just need to also add the include directive for `fedora-coreos-config/manifests/grub2-removals.yaml` in the RHCOS manifest once we bump the submodule there?

cgwalters:
Oh sorry, yep you're right!

PR #375, opened 2020-05-01
manual update of lockfile

dustymabe:
Bump the lockfiles one more time while we get the new automated
lockfile bumper in place.

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -148,7 +148,7 @@
       "evra": "2.3.0-1.fc31.x86_64"
     },
     "cups-libs": {
-      "evra": "1:2.2.12-6.fc31.x86_64"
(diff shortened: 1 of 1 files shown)


manifest-lock.x86_64.json
```diff
@@ -289,7 +289,7 @@
       "evra": "2.9.9-8.fc31.x86_64"
     },
     "fuse-overlayfs": {
-      "evra": "0.7.8-1.fc31.x86_64"
+      "evra": "1.0.0-1.fc31.x86_64"
```
dustymabe, line 292: I guess this is safe

dustymabe:
Note that some packages in the lockfile went backwards in version because now `cosa fetch --update-lockfile` takes into account the overrides lockfile. The older versions that are displayed in this diff represent ones that are frozen.

jlebon (approved):

PR #376, opened 2020-05-02
[next] enable automatic updates for next stream

dustymabe:
Our next stream is a production stream that we want to have automatic
updates enabled on.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/473

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -20,13 +20,3 @@ repos:
 
 add-commit-metadata:
   fedora-coreos.stream: next
-
-postprocess:
-  # Disable Zincati and fedora-coreos-pinger on non-production streams
-  # https://github.com/coreos/fedora-coreos-tracker/issues/163
-  - |
-    #!/usr/bin/env bash
-    set -xeuo pipefail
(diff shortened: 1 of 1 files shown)


bgilbert (approved):
LGTM.

We'll want to have a coreos-status post about this.  The same post could announce the `next` stream as well.

lucab:
@dustymabe thanks! I've left it up in the ticket yesterday planning to revisit today, nice to wake up and see a PR already :)

@bgilbert indeed. Do we have a timeline for that? Once we have two releases in the stream, the oldest one should be able to update, after manually removing the file and restarting `zincati.service`.

lucab (approved):

dustymabe:
> @bgilbert indeed. Do we have a timeline for that? Once we have two releases in the stream, the oldest one should be able to update, after manually removing the file and restarting `zincati.service`.

I was planning to send a coreos-status email this week after we do this next round of releases to draw more attention to the `next` stream. We can add this information to that email.

PR #377, opened 2020-05-04
Add gcp-routes script and service to overlay.d/05core

LorbusChris:
This includes the gcp-routes script and service in the 05core overlay.
Any differing versions of these files, layered on by e.g. RHCOS, should be overriding these.

### DIFF
--- a/overlay.d/05core/usr/libexec/coreos-gcp-routes
+++ b/overlay.d/05core/usr/libexec/coreos-gcp-routes
@@ -1,91 +1,113 @@
 #!/bin/bash
 
-declare -A routes
+# Update iptables rules based on google cloud load balancer VIPS
+#
+# This is needed because the GCP L3 load balancer doesn't actually do DNAT; 
+# the destination IP address is still the VIP. Normally, there is an agent that
+# adds the vip to the local routing table, tricking the kernel in to thinking
+# it's a local IP and allowing processes doing an accept(0.0.0.0) to receive
+# the packets. Clever.
+#
+# We don't do that. Instead, we DNAT with conntrack. This is so we don't break
+# existing connections when the vip is removed. This is useful for draining
+# connections - take ourselves out of the vip, but service existing conns.
+#
+# ~cdc~
+
+set -e
+
+# the list of load balancer IPs that are assigned to this node
+# keys = values, for easy searching
+declare -A vips
+
 curler() {
    curl --silent -L -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/${1}"
 }
 
-get_ifname() {
-  sysfs_path="/sys/class/net"
-  for dev in $(find ${sysfs_path} -maxdepth 1  -mindepth 1);
-  do
-      local mac=$(<${dev}/address);
-      local name="$(basename ${dev})"
-      if [ "${mac}" == "${1}" ];
-      then
-          echo "${name}"
-          return;
-      fi
-  done
+CHAIN_NAME="gcp-vips"
+
+
+# Create a chan if it doesn't exist
+ensure_chain() {
+    local table="${1}"
+    local chain="${2}"
+
+    if ! iptables -w -t "${table}" -S "${chain}" &> /dev/null ; then
+        iptables -w -t "${table}" -N "${chain}";
+    fi;
 }
 
-set_routes() {
-  local dev="${1}"
-  read -a dev_routes <<< "${routes[$dev]}"
-  for cur_route in $(ip route show dev ${dev} table local proto 66 | awk '{print$2}');
-  do
-      if [[ ! "${dev_routes[@]}" =~ "${cur_route}" ]];
-      then
-          echo "Removing stale forwarded IP ${cur_route}/32"
-          ip route del ${cur_route}/32 dev ${dev} table local proto 66
-      fi
-  done
-  for route in ${dev_routes[@]}
-  do
-      ip route replace to local ${route} dev $dev proto 66
-  done
-  unset dev_routes
+ensure_rule() {
+    local table="${1}"
+    local chain="${2}"
+    shift 2
+
+    if ! iptables -w -t "${table}" -C "${chain}" "$@" &> /dev/null; then
+        iptables -w -t "${table}" -A "${chain}" "$@"
+    fi
 }
 
-del_routes() {
-  local dev="${1}"
-  read -a dev_routes <<< "${routes[$dev]}"
-  for cur_route in $(ip route show dev ${dev} table local proto 66 | awk '{print$2}');
-  do
-      if [[ "${dev_routes[@]}" =~ "${cur_route}" ]];
-      then
-          echo "Removing forwarded IP ${cur_route}/32"
-          ip route del ${cur_route}/32 dev ${dev} table local proto 66
-      fi
-  done
-  unset dev_routes
+# set the chain, ensure entry rules, ensure ESTABLISHED rule
+initialize() {
+    ensure_chain nat "${CHAIN_NAME}"
+    ensure_rule nat PREROUTING -m comment --comment 'gcp LB vip DNAT' -j ${CHAIN_NAME}
+
+    # Need this so that existing flows (with an entry in conntrack) continue to be
+    # balanced, even if the DNAT entry is removed
+    ensure_rule filter INPUT -m comment --comment 'gcp LB vip existing' -m addrtype ! --dst-type LOCAL -m state --state ESTABLISHED,RELATED -j ACCEPT
 }
 
-run() {
-   net_path="network-interfaces/"
+remove_stale() {
+    ## find extra iptables rules
+    for ipt_vip in $(iptables -w -t nat -S "${CHAIN_NAME}" | awk '$4{print $4}' | awk -F/ '{print $1}'); do
+        echo checking if "${ipt_vip}" is stale
+        if [[ -z "${vips[${ipt_vip}]}" ]]; then
+            iptables -w -t nat -D "${CHAIN_NAME}" --dst "${ipt_vip}" -j REDIRECT
+        fi
+    done
+}
+
+add_rules() {
+    for vip in "${vips[@]}"; do
+        echo "ensuring rule for ${vip}"
+        ensure_rule nat "${CHAIN_NAME}" --dst "${vip}" -j REDIRECT
+    done
+}
+
+clear_rules() {
+    iptables -t nat -F "${CHAIN_NAME}" || true
+}
+
+# out paramater: vips
+list_lb_ips() {
+   local net_path="network-interfaces/"
    for vif in $(curler ${net_path}); do
-      hw_addr=$(curler "${net_path}${vif}mac")
-      fwip_path="${net_path}${vif}forwarded-ips/"
-      dev_name="$(get_ifname ${hw_addr})"
-      for level in $(curler ${fwip_path})
-      do
-          for fwip in $(curler ${fwip_path}${level})
-          do
-              echo "Processing route for NIC ${vif}${hw_addr} as ${dev_name} for ${fwip}"
-              routes[$dev_name]+="${fwip} "
+      local hw_addr; hw_addr=$(curler "${net_path}${vif}mac")
+      local fwip_path; fwip_path="${net_path}${vif}forwarded-ips/"
+      for level in $(curler "${fwip_path}"); do
+          for fwip in $(curler "${fwip_path}${level}"); do
+              echo "Processing route for NIC ${vif}${hw_addr} for ${fwip}"
+              vips[${fwip}]="${fwip}"
           done
       done
-      $"${1}" ${dev_name}
-
-      routes[$dev_name]=""
-      unset hw_addr
-      unset fwip_path
-      unset dev_name
    done
 }
 
 case "$1" in
   start)
+    initialize
     while :; do
-      run set_routes
+      list_lb_ips
+      remove_stale
+      add_rules
+      echo "done applying vip rules"
       sleep 30
     done
     ;;
   cleanup)
-    run del_routes
+    clear_rules
     ;;
   *)
     echo $"Usage: $0 {start|cleanup}"
     exit 1
-
 esac
--- a/overlay.d/05core/usr/libexec/coreos-gcp-routes
+++ b/overlay.d/05core/usr/libexec/coreos-gcp-routes
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
 
 # Update iptables rules based on google cloud load balancer VIPS
 #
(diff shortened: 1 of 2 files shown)


cgwalters:
Positives:

:heavy_plus_sign:  Ensures OKD can share code with OCP
:heavy_plus_sign:  Moves this script out of the "dark matter" of the RHCOS internal gitlab repo and into the sunlight 🌞 of public Github

Negatives:

:heavy_minus_sign: This script is really tied into the MCO - changes to this will need "ratcheting" and become an upgrade hazard

Neutral:

:black_circle: No OpenShift CI on this repo yet (but there isn't good CI for rhcos git either yet sadly)

---

Since we're planning to move this to NetworkManager at some point (hopefully soon) and this is just a short term thing, :+1: from me.

miabbott:
Can we drop a comment in one or both of these files with a link to the NM issue where we are discussing moving this functionality? 

https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/407

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Update GCP routes for forwarded IPs.
+ConditionKernelCommandLine=|ignition.platform.id=gce
+ConditionKernelCommandLine=|ignition.platform.id=gcp
```
dustymabe: one of these (I think `gce`) is legacy - can we add a comment about removing it in the future?

LorbusChris:
Added the comments per requests above

overlay.d/05core/usr/sbin/gcp-routes.sh
```diff
@@ -0,0 +1,165 @@
...
+declare -A vips
+
+curler() {
+   curl --silent -L -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/${1}"
+}
+
+CHAIN_NAME="gcp-vips"
+RUN_DIR="/run/gcp-routes"
+
+# Create a chan if it doesn't exist
```
dustymabe: chain*

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
```diff
@@ -0,0 +1,22 @@
...
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/bin/bash /usr/sbin/gcp-routes.sh start
+ExecStopPost=/bin/bash /usr/sbin/gcp-routes.sh cleanup
+User=root
+RestartSec=30
+Restart=always
```
dustymabe: Is this something that should execute and exit or should it run continuously? Feels like if it is failing we don't want this to always restart.

dustymabe:
I don't have much else to add to this code review (I don't know the intricacies of GCP LB networking) other than to ask if we tested this and what is our confidence we're not going to break existing users on upgrade? so I guess LGTM, but would really like someone with more knowledge to review.

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
LorbusChris: I'd rather not change this at this stage. It's been likes this in OKD and OCP.

LorbusChris:
I don't think it will interfere with upgrades or anything current setups do. The service is disabled by default and is only started/stopped by MCO's `gcp-routes-controller`: https://github.com/openshift/machine-config-operator/blob/master/cmd/gcp-routes-controller/run.go#L92

overlay.d/05core/usr/sbin/gcp-routes.sh
LorbusChris: fixed

dustymabe:
> the service is disabled by default and is only started/stopped by MCO's `gcp-routes-controller`

That makes me feel more comfortable with the change.

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
```diff
@@ -0,0 +1,22 @@
+# NOTE:
```
jlebon: Can we prefix the service name and script with `coreos-`?

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
```diff
@@ -0,0 +1,22 @@
...
+# This functionality should eventually be moved into nm-cloud-setup, see:
+# https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/407
+
+[Unit]
+Description=Update GCP routes for forwarded IPs.
```
jlebon: `Documentation=https://github.com/coreos/fedora-coreos-config` ?

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
```diff
@@ -0,0 +1,22 @@
...
+# and will likely be removed in the future 
+ConditionKernelCommandLine=|ignition.platform.id=gce
+ConditionKernelCommandLine=|ignition.platform.id=gcp
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/bin/bash /usr/sbin/gcp-routes.sh start
+ExecStopPost=/bin/bash /usr/sbin/gcp-routes.sh cleanup
+User=root
```
jlebon: Isn't that the default?

overlay.d/05core/usr/sbin/gcp-routes.sh
```diff
@@ -0,0 +1,165 @@
...
+# Additionally, clients can write a file to /run/gcp-routes/$IP.down to force
+# a VIP as down. This is useful for graceful shutdown / upgrade.
+#
+# ~cdc~
+
+# NOTE: 
+# This functionality should eventually be moved into nm-cloud-setup, see:
+# https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/407
+
+set -e
```
jlebon: Let's do `set -euo pipefail` here?

Also, can we put it as the second line of the script?

jlebon (commented):
Following up on https://github.com/coreos/fedora-coreos-config/pull/377#issuecomment-623583932, if this is tied to the MCO, then shouldn't it live there instead for now (until the base functionality moves to NM)?

Another way to ask this: does this script have any value outside of its use to MCO? Is this something that other users/Kube installs could use as is, or is it MCO-specific in implementation?

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
LorbusChris: This would need ratcheting code changes on the MCO side where the service invoked.

LorbusChris:
@jlebon yes, this is specific to MCO and for OKD we currently have it live there.
However upstream MCO will not take it in, and the goal is to to re-use MCO from canonical branches in OKD.

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
LorbusChris: Maybe this is due to the way it is invoked by MCO's controller: https://github.com/openshift/machine-config-operator/blob/master/cmd/gcp-routes-controller/run.go#L92 (could it be that it runs under `core` user otherwise?)

LorbusChris:
> heavy_minus_sign This script is really tied into the MCO - changes to this will need "ratcheting" and become an upgrade hazard

@cgwalters I want to just note that this is the status quo already with the script living in RHCOS.
Whenever we plan to move to this version of the script with OCP (be it 4.5 or 4.6), the RHCOS base will just have to drop the script from it's overlay.d in order to pick it up from here

cgwalters:
Yep, that part is easy, we can do it in one transactional commit as part of updating the git submodule.

cgwalters:
> if this is tied to the MCO, then shouldn't it live there instead for now (until the base functionality moves to NM)?

Hmm...the history here is...long.  Related original bits are:
- https://github.com/openshift/machine-config-operator/pull/1031
- https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/448

I forget exactly why we didn't end up with it in the MCO originally.  It might simply have been it just happened to start as part of RHCOS because it was obvious how to write a shell script we ship in the OS there.  One general rule here is https://github.com/openshift/machine-config-operator/pull/1696#issuecomment-622024637

We may also need this on the bootstrap node, I forget.

Given the MCO has driver code for it today too I honestly just don't recall why it didn't end up in the MCO but someone else may know.

> Another way to ask this: does this script have any value outside of its use to MCO? Is this something that other users/Kube installs could use as is, or is it MCO-specific in implementation?

The *idea* is something other Kuberenetes implementations could want, but I don't think we want to commit to it as API.

LorbusChris:
It's also needed on the bootstrap node, which is why in OKD right now, we add it A in the installer, and B in the MCO.

Should the comment I added more clearly state that this is not a supported api and definitely subject to future changes?

jlebon:
> Should the comment I added more clearly state that this is not a supported api and definitely subject to future changes?

Yes, agreed. Also, let's put it in `/usr/libexec` or something?

overlay.d/05core/usr/lib/systemd/system/gcp-routes.service
jlebon: But OKD isn't relying on it being in FCOS yet anyway, right?

LorbusChris: No, not yet. Right now it's added by the Installer on bootstrap nodes, and MCO on masters/workers.
I think we can prefix the files (which would mean they wouldn't be overridden by the RHCOS config and would be placed on all RHCOS and FCOS images) and then later make the switch on the MCO side by simply switching the command over to use the new coreos prefixed service and script

LorbusChris:
Added `coreos-` prefixes and changed the script's location to `/usr/libexec`

overlay.d/05core/usr/libexec/coreos-gcp-routes
```diff
@@ -0,0 +1,164 @@
...
+        done
+    done
+}
+
+sleep_or_watch() {
+    if hash inotifywait &> /dev/null; then
+        inotifywait -t 30 -r "${RUN_DIR}" &> /dev/null || true
+    else
+        # no inotify, need to manually poll
+        for (( tries=0; tries<6; tries++ )); do
```
jlebon, line 132: Minor: I think it's more idiomatic to use e.g. `for i in {0..5}` or even `for i in $(seq 6)`.

overlay.d/05core/usr/libexec/coreos-gcp-routes
```diff
@@ -0,0 +1,164 @@
...
+            for vip in "${!vips[@]}"; do
+                if [[ "${vips[${vip}]}" != down ]] && [[ -e "${RUN_DIR}/${vip}.down" ]]; then
+                    echo "new downfile detected"
+                    break 2
+                elif [[ "${vips[${vip}]}" = down ]] && ! [[ -e "${RUN_DIR}/${vip}.down" ]]; then
+                    echo "downfile disappeared"
+                    break 2
+                fi
+            done
```
jlebon, line 141: Minor: this IMO would be cleaner as a separate function named e.g. `detected_vip_changes`, so that it could just use `return 0` instead of `break 2` and the main condition in this loop would just be:

```sh
if detected_vip_changes; then
  break
fi
```

jlebon (approved):
Some minor comments about the bash script itself, but LGTM overall (I haven't really reviewed the iptables-related bits of the script in depth, since I'm not familiar with GCP routes). If you've tested this final version of the script, then I'm good with just leaving the comments as separate follow-ups instead.

squeed:
For upgrade / ratcheting, we could call it openshift-gcp-routes.service and add a `Conflicts=gcp-routes.service` in the unit.

squeed:
(deleted incorrect comment)

squeed:
Wow, this got complicated - I didn't think about the bootstrap node.

We don't actually need the same sophisticated, openshift-specific script on the bootstrap node, since we don't need to support any sort of in-place upgrades. It would be fine for the bootstrap node to have a generic nm-cloud-setup / route-tables script that should be part of FCOS / RHEL / RHCOS / CentOS on GCP.

Then, the MCO can install an openshift-gcp-routes that conflicts with the generic gcp-routes, which improves the specific experience on the control plane.

LorbusChris:
@squeed is this so specific to OpenShift that we can't move all of its functionality into NM eventually?

In any way, in 4.6 we'll need this in the FCOS base image so we can drop MCO's `fcos` branch.

squeed:
> @squeed is this so specific to OpenShift that we can't move all of its functionality into NM eventually?

Yup. NM doesn't want to touch iptables, and rightly so. But we do.

jlebon:
>> @squeed is this so specific to OpenShift that we can't move all of its functionality into NM eventually?
>
> Yup. NM doesn't want to touch iptables, and rightly so. But we do.

Given that, I think this should indeed belong in the MCO. You then get a strong binding between the routes controller and the script (i.e. you *know* exactly which version of the script is on the host), you can make atomic changes to both of them in one commit, etc...

But anyway, if folks prefer it here, that's fine too.

LorbusChris:
I'd be equally ok with it moving into MCO eventually. It's just that right now it would then be missing on the bootstrap hosts because afaik NM hasn't implemented any of this yet.

Having this in the base was only ever meant to be temporary and a compromise anyway.

How about we do this here now for 4.6, and revisit once NM has gained functionality we could use on boostrap nodes, moving it into MCO then?
/cc @runcom

squeed:
I think I'll create openshift-gcp-routes.service in the MCO now, and disable gcp-routes. You can push this to fcos now, and whenever the mco change rolls out, it will cleanly apply.

cgwalters:
> I'd be equally ok with it moving into MCO eventually. It's just that right now it would then be missing on the bootstrap hosts because afaik NM hasn't implemented any of this yet.

On the bootstrap host, the MCO already runs with root privileges - it can write the code to the host the same way it does when run as a pod.

It'd be an ugly special case, but definitely tractable.

LorbusChris:
@squeed I'm not 100% what the reduced script would have to look like on the bootstrap nodes.

Would you mind creating a replacement PR on this repo? I'll close this one here then.

squeed:
Since this is now in the MCO, I'd suggest closing this and enabling nm-cloud-setup **only** for the bootstrap node. Seems the simplest way; the bootstrap node doesn't need the openshift-specific gcp setup.

jlebon:
Obsoleted by https://github.com/coreos/fedora-coreos-config/pull/502.

sohankunkerkar:
@squeed Thanks for your reply. 
I'm thinking about handling this feature by:
[1] Adding `NetworkManager-cloud-setup` package in the fedora-coreos-base.yaml file (https://github.com/coreos/fedora-coreos-config/pull/502)
[2] ~Configuring https://github.com/openshift/installer/blob/master/data/data/bootstrap/gcp/files/usr/local/bin/report-progress.sh by enabling `NetworkManager-cloud-setup` at the start of the script and stop after  https://github.com/openshift/installer/blob/master/data/data/bootstrap/gcp/files/usr/local/bin/report-progress.sh#L16~

PR #378, opened 2020-05-05
[next-devel] Fix manifest lockfile extension from yaml to json

jlebon:
The lockfile mistakenly had a `.yaml` extension instead of `.json`. It
still worked fine of course, because rpm-ostree supports both, and cosa
looks for both.

Though it's misleading because the base lockfile is generated by
rpm-ostree itself, which technically generates JSON and not YAML (even
though JSON is in fact valid YAML).

So let's rename it. (We also have some tools today which would still
need to be adapted for handling YAML base lockfiles if we transitioned.)


jlebon:
See https://github.com/coreos/coreos-assembler/pull/1426 for the cosa side of this (though the PRs themselves are independent from each other).

rfairley (approved):
LGTM! Learned in reviewing this that `{}` and `[]` are valid YAML and compatible with JSON (https://yaml.org/spec/1.2/spec.html#id2759963). Makes sense to name the file explicitly JSON though, given rpm-ostree generates the file in this format (and won't use YAML features in the generated file).

PR #379, opened 2020-05-05
coreos-autologin-generator: rename to coreos-liveiso-autologin-generator

jlebon:
An `autologin` generator can sound scary. Let's add `liveiso` to its
name to make it more obvious that it only kicks in on the live ISO path.

While we're here, paint the bikeshed some more by lowercasing the "live"
word in the MOTD since having it capital looks off.


jlebon:
> While we're here, paint the bikeshed some more by lowercasing the "live"
word in the MOTD since having it capital looks off.

Happy to drop that bit if others prefer it as it is! It just didn't look right to me.

cgwalters (approved):

bgilbert (approved):
LGTM

PR #380, opened 2020-05-05
[next-devel] fasttrack coreos-installer-0.2.0-1.fc32 and afterburn-4.3.3-1.fc32

rfairley:
See commit messages for details.

Currently not verified locally - testing in a COSA build locally gives `error: Couldn't find locked package 'afterburn-4.3.3-1.fc32.x86_64' (pkgs matching NEVRA: 0; mismatched checksums: 0)`). I think `coreos-koji-tagger` will pick this change up once merged, which will fix this.

PR #381, opened 2020-05-05
[testing-devel] fasttrack coreos-installer-0.2.0-1.fc32 and afterburn-4.3.3-1.fc32

rfairley:
See commit messages for details.

Equivalent to #380, but in `testing-devel`.


dustymabe (approved):
LGTM

PR #385, opened 2020-05-06
[next-devel] manifest: point to release repos

jlebon:
The f32 repos are no longer in the `development/` path. Use the release
repos now.


dustymabe (approved):
LGTM

PR #387, opened 2020-05-06
overlay: 14NetworkManager-plugins: update comment in conf dropin

dustymabe:
There is an upstream proposal for fixing the bug referenced in the
configuration dropin. Let's link to the bug so we can follow and fix
it.


cgwalters (approved):

dustymabe:
Thanks @cgwalters!

PR #392, opened 2020-05-13
Add support for root device on multipath via `rd.multipath=default` kernel argument 

darkmuggle:
This provides for an ergonomic means for a user to enable multipath without having to manually define ugly kargs. When rd.multipath= is set for a "on" option, then the ostree mount units will use   `/dev/disk/by-label/dm-mpath-root` instead.
    
This can be tested with:
      `cosa run --kargs multipath=default --qemu-multipath`

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
@@ -14,15 +14,37 @@ shift
 # The use of tail is to avoid errors from duplicate mounts;
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
-majmin=$(findmnt -nvr -o MAJ:MIN "$path" | tail -n1)
-devpath=$(realpath "/sys/dev/block/$majmin")
-partition=$(cat "$devpath/partition")
-parent_path=$(dirname "$devpath")
-parent_device=/dev/$(basename "${parent_path}")
 
-# TODO: make this idempotent, and don't error out if
-# we can't resize.
-growpart "${parent_device}" "${partition}" || true
+if [[ "${src}" =~ "/dev/mapper" ]]; then
+    eval $(udevadm info --query property --export "${src}")
+    # get the partition, if any, and the name for the device mapper
+    partition="${ID_PART_ENTRY_NUMBER:-}"
+    dm_name="${DM_NAME//$partition/}"
+    # identify the type of device mapper.
+    subsystem=$(dmsetup info ${dm_name} -C -o subsystem --noheadings)
+
+    # for now, we only support multipath devices
+    if [ "${subsystem}" == "mpath" ] && [ -n "${partition}" ]; then
+        # growpart does not understand device mapper, instead of having sfdisk inform the kernel,
+        # use kpartx to inform the kernel and the device mapper linear maps.
+        echo ", +" | sfdisk --no-reread --no-tell-kernel --force -N "${ID_PART_ENTRY_NUMBER}" "/dev/mapper/${dm_name}"
+        kpartx -fu /dev/mapper/${dm_name}
+    else
+        echo "coreos-growpart: unsupported device-mapper target: ${dm_name}"
+        exit 0
+    fi
+else
+    # Handle traditional disk/partitions
+    majmin=$(findmnt -nvr -o MAJ:MIN "$path" | tail -n1)
+    devpath=$(realpath "/sys/dev/block/$majmin")
+    partition="${partition:-$(cat "$devpath/partition")}"
+    parent_path=$(dirname "$devpath")
+    parent_device=/dev/$(basename "${parent_path}")
+
+    # TODO: make this idempotent, and don't error out if
+    # we can't resize.
+    growpart "${parent_device}" "${partition}" || true
+fi
 
 eval $(blkid -o export "${src}")
 # TODO: Add XFS to https://github.com/systemd/systemd/blob/master/src/partition/growfs.c
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
@@ -1,0 +1,83 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# On a regular boot we rely on `/dev/disk/by-label/{boot,root} to exist.
+# However, if multipath is enabled, then automatic discovery of boot/root
+# gets racy. If the user has not put `rd.multipath` then generator exists
+# without making any modifications.
+
+set -e
+
+# Generators don't have logging right now
+# https://github.com/systemd/systemd/issues/15638
+exec 1>/dev/kmsg; exec 2>&1
+
+UNIT_DIR="${1:-/tmp}"
+mkdir -p "${UNIT_DIR}"
+
+skip() {
+    echo "multipath-generator: skipping root on multipath: ${@}"
+    exit 0
+}
+
+# Check if we expect multipath root. These are the commdline options
+# checked by Dracut in https://github.com/dracutdevs/dracut/blob/master/modules.d/90multipath/multipathd.service#L9-L12
+[ -f /etc/multipath.conf ] && mpath=1 || mpath=0
+
+for arg in $(</proc/cmdline);
+do
+    case "${arg}" in
+        multipath=off)                   skip "karg multipath=off";;
+        rd.multipath=0|rd.multipath=off) skip "karg rd.multipath=(0|off)";;
+        rd_NO_MULTIPATH)                 skip "karg rd_NO_MULTIPATH";;
+        root=*)                          skip "karg root= explicit";;
+        rd.multipath=*)                  mpath=1;;
+    esac
+done
+
+# Unless there is a multipath.conf and automatic multipath,
+# then bail.
+if [ "${mpath}" == "0" ]; then
+    echo "no indicator that multipath should be used"
+    exit 0
+fi
+
+echo "multipath-generator: requiring root to be on a multipath device"
+
+cat > /run/sysroot.env <<EOF
+ROOT_DEVICE_PATH=/dev/disk/by-label/dm-mpath-root
+EOF
+
+boot_device="dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device"
+root_device="dev-disk-by\x2dlabel-dm\x2dmpath\x2droot.device"
+
+# Reconfigure ignition-diskful.target.d
+firstboot_cfg="${UNIT_DIR}/ignition-diskful.target.d"
+mkdir -p "${firstboot_cfg}"
+cat > "${firstboot_cfg}/ignition-mpath.conf" <<EOF
+# Added by multipath-generator
+[Unit]
+After=multipathd.service
+Wants=multipathd.service
+
+Requires=${boot_device}
+After=${boot_device}
+
+Requires=${root_device}
+After=${root_device}
+EOF
+
+# Reconfigure ignition-ostree-mount-subsequent-sysroot.service.d, which is
+# required by ignition-diskful-subsquent.target.
+secondboot_cfg="${UNIT_DIR}/ignition-ostree-mount-subsequent-sysroot.service.d"
+mkdir -p "${secondboot_cfg}"
+cat >> "${secondboot_cfg}/ignition-mpath.conf" <<EOF
+# Added by multipath-generator
+[Unit]
+After=multipathd.service
+Wants=multipathd.service
+
+After=${root_device}
+Requires=${root_device}
+EOF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+export PATH="/usr/bin:/usr/sbin:${PATH}"
 set -euo pipefail
 
 # Generators don't have logging right now
@@ -7,6 +7,12 @@ set -euo pipefail
 exec 1>/dev/kmsg; exec 2>&1
 
 UNIT_DIR="${1:-/tmp}"
+
+# Turn out if you boot with "root=..." $UNIT_DIR is not writable.
+[ -w "${UNIT_DIR}" ] || {
+    echo "skipping coreos-boot-mount-generator: ${UNIT_DIR} is not writable"
+    exit 0
+}
 
 add_wants() {
     local name="$1"; shift
@@ -22,42 +28,54 @@ if findmnt --fstab /boot &>/dev/null; then
     exit 0
 fi
 
+# Generate mount units that work with device mapper. The traditional
+# device unit (dev-disk-by\x2dlabel...) does not work since it is not the
+# device that systemd will fsck. This code ensures that if the label
+# is backed by a device-mapper target the dev-mapper.*.device is used.
+mk_mount() {
+    local unit_name="${1}.mount"; shift
+    local label="${1}"; shift
+    local mount_pt="${1:-/$label}"
+    local path="/dev/disk/by-label/${label}"
+
+    eval $(udevadm info --query property --export "${path}")
+    device="$(systemd-escape ${path})"
+    if [ "${DM_NAME:-x}" != "x" ]; then
+        path="/dev/mapper/${DM_NAME}"
+        device="$(systemd-escape dev/mapper/${DM_NAME})"
+    fi
+    device="${device//-dev/dev}"
+    echo "coreos-boot-mount-generator: using ${device} for ${label} mount to ${mount_pt}"
+
+    cat > "${UNIT_DIR}/${unit_name}" <<EOF
+# Automatically created by coreos-boot-mount-generator
+[Unit]
+Description=CoreOS Dynamic Mount for ${mount_pt}
+Documentation=https://github.com/coreos/fedora-coreos-config
+
+Before=local-fs.target
+Requires=systemd-fsck@${device}.service
+After=systemd-fsck@${device}.service
+
+[Mount]
+What=${path}
+Where=${mount_pt}
+EOF
+
+    add_wants "${unit_name}"
+}
+
+
 # Don't create mount units for /boot or /boot/efi on live systems.
 # ConditionPathExists won't work here because conditions don't affect
 # the dependency on the underlying device unit.
 if [ ! -f /run/ostree-live ]; then
-    add_wants boot.mount
-    cat > "${UNIT_DIR}/boot.mount" <<EOF
-# Automatically created by coreos-boot-mount-generator
-[Unit]
-Description=Boot partition
-Documentation=https://github.com/coreos/fedora-coreos-config
-Before=local-fs.target
-Requires=systemd-fsck@dev-disk-by\x2dlabel-boot.service
-After=systemd-fsck@dev-disk-by\x2dlabel-boot.service
-
-[Mount]
-What=/dev/disk/by-label/boot
-Where=/boot
-EOF
+    mk_mount boot boot
 
     # Only mount the EFI System Partition on machines where it exists,
     # which are 1) machines actually booted through EFI, and 2) x86_64
     # when booted through BIOS.
     if [ "$(uname -m)" = "x86_64" -o -d /sys/firmware/efi ]; then
-        add_wants boot-efi.mount
-        cat > "${UNIT_DIR}/boot-efi.mount" <<EOF
-# Automatically created by coreos-boot-mount-generator
-[Unit]
-Description=EFI System Partition
-Documentation=https://github.com/coreos/fedora-coreos-config
-Before=local-fs.target
-Requires=systemd-fsck@dev-disk-by\x2dlabel-EFI\x2dSYSTEM.service
-After=systemd-fsck@dev-disk-by\x2dlabel-EFI\x2dSYSTEM.service
-
-[Mount]
-What=/dev/disk/by-label/EFI-SYSTEM
-Where=/boot/efi
-EOF
+        mk_mount boot-efi EFI-SYSTEM "/boot/efi"
     fi
 fi
--- a/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
+++ b/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
@@ -1,0 +1,20 @@
+ACTION=="remove", GOTO="dm_label_end"
+SUBSYSTEM!="block", GOTO="dm_label_end"
+KERNEL!="dm-*", GOTO="dm_label_end"
+
+# Ensure that the device mapper target is active
+ENV{DM_ACTIVATION}!="1", GOTO="dm_label_end"
+
+# Unless we have a label, skip
+ENV{ID_FS_LABEL_ENC}!="?*", GOTO="dm_label_end"
+
+# Only act on filesystems
+ENV{ID_FS_USAGE}!="filesystem", GOTO="dm_label_end"
+
+ENV{DM_MPATH}=="?*", \
+    ENV{DM_SUSPENDED}=="Active"
+    SYMLINK+="disk/by-label/dm-mpath-$env{ID_FS_LABEL_ENC}",
+    SYMLINK+="disk/by-uuid/dm-mpath-$env{ID_FS_UUID_ENC}"
+
+LABEL="dm_label_end"
+
--- a/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
+++ b/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
@@ -12,7 +12,7 @@ ENV{ID_FS_LABEL_ENC}!="?*", GOTO="dm_label_end"
 ENV{ID_FS_USAGE}!="filesystem", GOTO="dm_label_end"
 
 ENV{DM_MPATH}=="?*", \
-    ENV{DM_SUSPENDED}=="Active"
+    ENV{DM_SUSPENDED}=="Active",
     SYMLINK+="disk/by-label/dm-mpath-$env{ID_FS_LABEL_ENC}",
     SYMLINK+="disk/by-uuid/dm-mpath-$env{ID_FS_UUID_ENC}"
 
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -16,7 +16,6 @@ initramfs-args:
   - --omit=nfs
   # Omit these since we don't use them
   - --omit=lvm
-  - --omit=multipath
   - --omit=iscsi
 
 # Be minimal
@@ -135,7 +134,7 @@ packages:
   - cifs-utils
   # Time sync
   - chrony
-  # Allow communication between sudo and SSSD 
+  # Allow communication between sudo and SSSD
   # for caching sudo rules by SSSD.
   # https://github.com/coreos/fedora-coreos-tracker/issues/445
   - libsss_sudo
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-firstboot-sysroot.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-firstboot-sysroot.service
@@ -22,4 +22,5 @@ Before=ostree-prepare-root.service ignition-remount-sysroot.service
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+EnvironmentFile=-/run/sysroot.env
 ExecStart=/usr/sbin/ignition-ostree-mount-sysroot
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-subsequent-sysroot.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-subsequent-sysroot.service
@@ -21,4 +21,5 @@ Before=ostree-prepare-root.service
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+EnvironmentFile=-/run/sysroot.env
 ExecStart=/usr/sbin/ignition-ostree-mount-sysroot
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
@@ -7,7 +7,10 @@ set -euo pipefail
 # In the future this will be augmented with a check for whether
 # or not we've reprovisioned the rootfs, since we don't want to
 # force on prjquota there.
-rootpath=/dev/disk/by-label/root
+rootpath="${ROOT_DEVICE_PATH:-/dev/disk/by-label/root}"
+
+# If root is on a multipath device, we use that one.
+# this link is created by our own udev rule.
 if ! [ -b "${rootpath}" ]; then
   echo "ignition-ostree-mount-sysroot: Failed to find ${rootpath}" 1>&2
   exit 1
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -59,6 +59,13 @@ install() {
         inst_script "$moddir/ignition-ostree-${x}-var.sh" "/usr/sbin/ignition-ostree-${x}-var"
     done
 
+    inst_simple \
+        /usr/lib/udev/rules.d/90-coreos-device-mapper.rules
+
+    inst_simple "$moddir/multipath-generator" \
+        "$systemdutildir/system-generators/multipath-generator"
+
+    # Disk support
     install_ignition_unit ignition-ostree-mount-firstboot-sysroot.service diskful
     install_ignition_unit ignition-ostree-mount-subsequent-sysroot.service diskful-subsequent
     inst_script "$moddir/ignition-ostree-mount-sysroot.sh" \


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -11,12 +11,30 @@ set -euo pipefail
...
 
+
 # The use of tail is to avoid errors from duplicate mounts;
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
 majmin=$(findmnt -nvr -o MAJ:MIN "$path" | tail -n1)
+
+
+# Detect if the device is on multipath and then _find_ the active target
+if [[ "${src}" =~ "/dev/mapper" ]]; then
```
darkmuggle, line 18: this does not work yet

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -22,42 +22,52 @@ if findmnt --fstab /boot &>/dev/null; then
...
+    local mount_pt="/${1:-$label}"
+    local path="/dev/disk/by-label/${label}"
+
+    eval $(udevadm info --query property --export "${path}")
+    local device="dev-disk-by\x2dlabel-${label//-\x2d}"
+    if [ "${DM_NAME:-x}" != "x" ]; then
+        device="dev-mapper-${DM_NAME}"
+        path="/dev/mapper/${DM_NAME}"
+    fi
+    echo "***** coreos-boot-mount-generator: using ${device} for ${label} mount to ${mount_pt}"
```
darkmuggle: this is for debugging; will be dropped

darkmuggle:
@jlebon when used with https://github.com/coreos/ignition-dracut/pull/183 this works end-to-end

image.yaml
```diff
@@ -8,6 +8,7 @@ extra-kargs:
     # https://github.com/coreos/fedora-coreos-tracker/issues/292
     # https://fedoraproject.org/wiki/Changes/CGroupsV2
     - systemd.unified_cgroup_hierarchy=0
+    - rd.multipath=default
```
jlebon: Need to drop this too.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -22,42 +22,52 @@ if findmnt --fstab /boot &>/dev/null; then
...
-    add_wants boot.mount
-    cat > "${UNIT_DIR}/boot.mount" <<EOF
+mk_mount() {
+    local unit_name="${1}.mount"; shift
+    local label="${1}"; shift
+    local mount_pt="/${1:-$label}"
+    local path="/dev/disk/by-label/${label}"
+
+    eval $(udevadm info --query property --export "${path}")
+    local device="dev-disk-by\x2dlabel-${label//-\x2d}"
```
jlebon: Shouldn't that be

```suggestion
    local device="dev-disk-by\x2dlabel-${label/-/\x2d}"
```

?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -22,42 +22,52 @@ if findmnt --fstab /boot &>/dev/null; then
...
-# Don't create mount units for /boot or /boot/efi on live systems.
-# ConditionPathExists won't work here because conditions don't affect
-# the dependency on the underlying device unit.
-if [ ! -f /run/ostree-live ]; then
-    add_wants boot.mount
-    cat > "${UNIT_DIR}/boot.mount" <<EOF
+mk_mount() {
+    local unit_name="${1}.mount"; shift
+    local label="${1}"; shift
+    local mount_pt="/${1:-$label}"
```
jlebon: I think it'd be cleaner to pass the full mount point path instead.

jlebon (commented):
Just did a quick respin with this and I can confirm this fixes the `/boot` mount at least. `/boot/efi` doesn't get mounted though.

Only did a light review below. Need to play with it some more, though... let me ask a silly question: do you understand why fsck is failing without this? When I fsck manually it works fine:

```
# fsck.ext4 /dev/mapper/mpatha1
e2fsck 1.45.5 (07-Jan-2020)
boot: clean, 306/98304 files, 104701/393216 blocks
```

Haven't prodded much further though.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: Intentional, but re-worked.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: Done

darkmuggle:
Note: this does give a pattern to handle dm-raid on the root device as well. All we would need to do is to:
- add knowledge for software raid
- add a udev rule

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
ashcrow: Should this PR be WIP?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# On a regular boot we rely on `/dev/disk/by-label/{boot,root} to exist.
+# However, if multipath is enabled, then automatic discovery of boot/root
+# gets racy. If the user has not put `rd.multipath`
```
cgwalters: ... "then this generator exits without changing anything." ?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -22,42 +22,53 @@ if findmnt --fstab /boot &>/dev/null; then
...
+    local path="/dev/disk/by-label/${label}"
+
+    eval $(udevadm info --query property --export "${path}")
+    device="$(systemd-escape ${path})"
+    if [ "${DM_NAME:-x}" != "x" ]; then
+        path="/dev/mapper/${DM_NAME}"
+        device="$(systemd-escape dev/mapper/${DM_NAME})"
+    fi
+    device="${device//-dev/dev}"
+    echo "coreos-boot-mount-generator: using ${device} for ${label} mount to ${mount_pt}"
```
cgwalters, line 48: If we used uuids for the target filesystems, then we wouldn't need all of this right?

cgwalters (commented):
Overall looks OK, didn't dive into the details of parsing dmsetup etc. though.  I do suspect if we switched to using UUIDs we could basically just reuse all of the existing dracut infrastructure, right?

cgwalters:
Also forgot to say it seems clearer to title this work "Add support for root device on multipath" or so - the term *automatic* implies something stronger, but per discussion we require an opt-in kernel argument.  (And that will be set by `coreos-installer` right?  Probably convenient to test with e.g. `cosa run --kargs rd.multipath=default` too though)

darkmuggle:
> Overall looks OK, didn't dive into the details of parsing dmsetup etc. though. I do suspect if we switched to using UUIDs we could basically just reuse all of the existing dracut infrastructure, right?

Not likely. We would have the same racing condition around `/dev/disk/by-label` except with `/dev/disk/by-uuid`. The only way around this would be to use `/dev/mapper/<DM NAME>` or `/dev/disk/by-id/dm-mpath-<WWN>`.  In light of that, I've updated the UDEV rule to include creating `/dev/disk/by-uuid/dm-mpath-<UUID>`.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: Not likely. We would have the same racing condition around /dev/disk/by-label except with /dev/disk/by-uuid. The only way around this would be to use /dev/mapper/<DM NAME> or /dev/disk/by-id/dm-mpath-<WWN>. In light of that, I've updated the UDEV rule to include creating /dev/disk/by-uuid/dm-mpath-<UUID>.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
darkmuggle: Good catch, and updated.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
cgwalters: I guess my question is: if I use Anaconda and install onto a multipath system, why isn't there a race there too?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
darkmuggle: It works now. See the latest.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: https://anaconda-installer.readthedocs.io/en/latest/multipath.html#how-anaconda-handles-multipath

tl;dr: Anaconda uses `/dev/mapper/<....>` and uses the multipath device for partitioning 

Per the docs, it would appear that root becomes `/dev/mapper/<NAME><PART>`

cgwalters: Oh wow hey docs!  Nice!  I've been thinking lately we should try to codify such an "internals" document - we have a lot of great stuff scattered around our PRs and commit messages, but not the same as that type of doc.

Anyways...hmm.  You're saying that the resulting `root=` kernel argument isn't `root=UUID=` but `root=/dev/mapper/...` ?  Would it help us to do the same?

cgwalters: To be clear, this isn't a blocker for merging necessarily, we can always clean up an implementation later.  But just thinking through this, doesn't it help us to know *which* multipath device is holding the rootfs via a karg like that instead of just `rd.multipath=1` ?

darkmuggle: The reason I didn't go down the `/dev/mapper/...` route and used a `/dev/disk/by-id/dm-mpath-root` was because of identifying which multipath device to use. This removes the need for us to guess which multipath (e.x multiple mpaths) and the WWN. 

Now, that said, I think the step is that we make the code that switches to UUID we should make it aware of using the multipath UUID.

darkmuggle:
Okay, I've put this through its paces and I'm satisfied that it works as intended. 

I tested with the latest upstream Dracut using the following:
1. `cosa run --kargs rd.multipath=default --qemu-multipath`: should be automatic mulitpath 
1. `cosa run  --qemu-multipath`: should have no multipath, regular boot
1. `cosa run`: no multipath, regular boot

In all cases I tested first boot and second boot. And finally, I confirmed that resizing/growing the rootfs works.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: > But just thinking through this, doesn't it help us to know which multipath device is holding the rootfs via a karg....

I'm not sure I understand this statement. However, a couple of points:
1. A udev rule creates a stable link between the label and the device mapper target. https://github.com/coreos/fedora-coreos-config/pull/392/files#diff-a9bdb2d43b35de24f89ce871ca5f40fdR14-R17
1. `mount` and `/proc/mounts` will actually show `/dev/mapper/mpatha4` regardless of whether you mount by `/dev/dm-4`, `/dev/disk/by-label/dm-mpath-root` or `/dev/mapper/mpatha4`. 
1. `sudo dmsetup table /dev/mapper/mpatha` will show something like this: `0 16777216 multipath 0 0 2 1 service-time 0 1 2 8:0 1 1 service-time 0 1 2 8:16 1 1` gives you the major and minor of the targets

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
cgwalters: This condition can also be true for e.g. LUKS right?   I guess what we really need to think about here is how this intersects with Ignition rootfs reprovisioning.  Would we do multipath setup before or after that?

Probably, if the rootfs has been reprovisioned then the rootfs will already have been grown there, so perhaps `ignition-disks` would write out a stamp file like `/run/igntion-disks-rootfs-done.stamp` and we'd skip growpart entirely if that was done.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
cgwalters: > The reason I didn't go down the /dev/mapper/... route and used a /dev/disk/by-id/dm-mpath-root was because of identifying which multipath device to use. This removes the need for us to guess which multipath (e.x multiple mpaths) and the WWN.

In a `coreos-installer` scenario, we don't need to guess anything - the user must have explicitly provided a target disk.

Actually something that's still not clear to me here is how `multipath.conf` gets written.  The admin is writing that manually right?  If so, their config file is defining a multipath device - but unless we parse that `multipath.conf` file ourselves, we don't know which one.

Another way to say this is - when I'm arguing for UUIDs what I'm really saying is that on *subsequent* boots we move away from depending on the labels at all.   This code is basically doing "multipath over labels".

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
darkmuggle: For our current LUKS in RHCOS, there is a separate resize logic since we need to do it a bit differently. See `rhcos-growpart`. 

That said, we should absolutely have multipath active before running Ignition rootfs does its thing.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -22,42 +22,53 @@ if findmnt --fstab /boot &>/dev/null; then
...
+# device unit (dev-disk-by\x2dlabel...) does not work since its not the
+# device that systemd will fsck. This code ensures that if the label
+# is backed by a device-mapper target the dev-mapper.*.device is used.
+mk_mount() {
+    local unit_name="${1}.mount"; shift
+    local label="${1}"; shift
+    local mount_pt="${1:-/$label}"
+    local path="/dev/disk/by-label/${label}"
+
+    eval $(udevadm info --query property --export "${path}")
```
cgwalters, line 41: Something I don't understand here is - how do we know that `/dev/disk/by-label/boot` has appeared by the time this generator has run?  Are we relying on udev settling somehow?

darkmuggle: This generator is not run in the initramfs. So
- `/dev/disk/by-label/{boot,root}` are left over from the initramfs
-  `udevadm info` is querying the udev database found in `/run/dev/data` (which is left from the initramfs). You can confirmed this via `strace udevadm info....`.  The database is a just a bunch of flat files. For example:
```
$ cat /run/udev/b253\:4
S:disk/by-id/dm-name-mpatha4
S:disk/by-id/dm-uuid-part4-mpath-0x2e21c8a45250f752
S:disk/by-partlabel/root
S:mapper/mpatha4
S:disk/by-label/root
S:disk/by-label/dm-mpath-root
S:disk/by-uuid/28ef6336-bc5a-4955-bd31-966497c068c6
S:disk/by-id/scsi-0x2e21c8a45250f752-part4
S:disk/by-id/wwn-0xx2e21c8a45250f752-part4
S:disk/by-uuid/dm-mpath-28ef6336-bc5a-4955-bd31-966497c068c6
S:disk/by-partuuid/e54705f6-d6a1-4f4f-b479-56ce7151ddd2
L:50
W:25
I:2552345
E:DM_UDEV_DISABLE_LIBRARY_FALLBACK_FLAG=1
E:DM_UDEV_PRIMARY_SOURCE_FLAG=1
E:DM_SUBSYSTEM_UDEV_FLAG0=1
E:DM_ACTIVATION=1
E:DM_NAME=mpatha4
E:DM_UUID=part4-mpath-0x2e21c8a45250f752
E:DM_SUSPENDED=0
E:DM_UDEV_RULES_VSN=2
E:ID_FS_USAGE=filesystem
E:ID_FS_UUID_ENC=28ef6336-bc5a-4955-bd31-966497c068c6
E:ID_FS_LABEL_ENC=root
E:ID_PART_ENTRY_NAME=root
E:ID_PART_ENTRY_UUID=e54705f6-d6a1-4f4f-b479-56ce7151ddd2
E:ID_PART_ENTRY_SCHEME=gpt
E:DM_MPATH=mpatha
E:DM_PART=4
E:DM_TYPE=scsi
E:DM_WWN=0xx2e21c8a45250f752
E:DM_SERIAL=0x2e21c8a45250f752
E:ID_FS_LABEL=root
E:ID_FS_UUID=28ef6336-bc5a-4955-bd31-966497c068c6
E:ID_FS_TYPE=xfs
E:ID_PART_ENTRY_TYPE=0fc63daf-8483-4772-8e79-3d69d8477de4
E:ID_PART_ENTRY_NUMBER=4
E:ID_PART_ENTRY_OFFSET=1050624
E:ID_PART_ENTRY_SIZE=15726559
E:ID_PART_ENTRY_DISK=253:0
G:systemd
```

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: > Actually something that's still not clear to me here is how multipath.conf gets written.

Per https://github.com/dracutdevs/dracut/commit/b8a92b715677d52dbc2b27a710b9816fd8b9a63b `rd.multipath=default` will generate a default multipath configuration. This change relies on in-coming new version of Dracut and https://github.com/darkmuggle/ignition-dracut/blob/c8c631a2b7a7bf10f7b5b1cb037b6d3e2608c909/dracut/30ignition/coreos-teardown-initramfs.sh#L107-L117 

> Another way to say this is - when I'm arguing for UUIDs what I'm really saying is that on subsequent boots we move away from depending on the labels at all.

The label that switches over to UUID's will need to understand either multipath or the `/dev/disk/by-uuid/dm-mpath-<UUID>`.  Since that code has not landed, I supported the current state.

cgwalters (approved):
OK thanks for working on this!  I think it's good to merge as is but I'd like to chase down the concept of having the target disk be specified via karg as a potential followup.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
cgwalters: > Per dracutdevs/dracut@b8a92b7 rd.multipath=default will generate a default multipath configuration.

I think I don't understand that basically.  What's a "default" multipath?  Was that for the s390x case where we can somehow detect configuration from the underlying platform?

Or stated at a higher level, I don't understand what
`cosa run --kargs rd.multipath=default --qemu-multipath:`
is doing with that dracut patch compared to just
`cosa run --kargs rd.multipath=1 --qemu-multipath`
i.e. the previous dracut option?

darkmuggle: `rd.multipath=1` is legacy [1] and is no longer used [4] by Fedora since we have systemd. `rd.multipath=0` turns it off [2].

`rd.multipath=default` was added to run `mpathconf --enable` [3] and only triggers when there is NO configuration file. `rd.multipath` is a boolean for turning it on or off; `default` says turn it on _and_ use the default configuration.  The default configuration is what most users will want and may negate the need for users to provide custom configurations. 

My change in here was to extend the meaning of `rd.multipath=default` to mean:
- follow the dracut behavior in [3]
- _expect root to be on multipath_ 
- if the user has defined `root=/dev/mapper/mpatha4` (which is the default mapping for `root` on Fedora CoreOS) do nothing. 

[1] https://github.com/dracutdevs/dracut/blob/master/modules.d/90multipath/multipathd.sh#L8
[2] https://github.com/dracutdevs/dracut/blob/master/modules.d/90multipath/multipathd.service#L9-L12
[3] https://github.com/dracutdevs/dracut/blob/master/modules.d/90multipath/multipathd-configure.service#L10-L11
[4] https://github.com/dracutdevs/dracut/blob/master/modules.d/90multipath/module-setup.sh#L112-L120

darkmuggle:
I'll wait before merging until @jlebon or another review this PR.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
cgwalters: > if the user has defined root=/dev/mapper/mpatha4 (which is the default mapping for root on Fedora CoreOS)

Where is `/dev/mapper/mpatha4` defined?  How does it map to physical disks?

cgwalters:
OK I think I understand the state a bit more now, without
https://github.com/coreos/fedora-coreos-config/pull/364
this can't configure very much, right?

So probably it makes sense to revisit the implementation details of this (such as specifying a specific multipath config via kernel argument) once we also support actually configuring `multipath.conf`.

darkmuggle:
> OK I think I understand the state a bit more now, without
> #364
> this can't configure very much, right?

Well, #364 is clearly dead. But yeah, this does not configure much other than the defaults which amounts to: 
```
defaults {
	user_friendly_names yes
	find_multipaths yes
}
blacklist_exceptions {
        property "(SCSI_IDENT_|ID_WWN)"
}
blacklist {
}
``` 
and then multipathd uses its built in defaults. 

After @jlebon and I had a video chat I went and dealt with the generic path. My assertion is that this will satisfy the needs of most users. Running `multipath -t` shows the built-in defaults (see https://gist.github.com/darkmuggle/80fdd5d17bfb9b1a357bfa66e59a2675) which supports some 120+ SANS and iSCSI targets.  FWIW, this is why https://github.com/coreos/coreos-assembler/commit/72458e60ed200c50517153819386acb1654f67e9#diff-cd4cf2dc4290b10e445cb3d643c39bc0R293 uses the vendor of `NVME` since the vendor definition https://gist.github.com/darkmuggle/80fdd5d17bfb9b1a357bfa66e59a2675 is simple and was innocent enough. 

> 
> So probably it makes sense to revisit the implementation details of this (such as specifying a specific multipath config via kernel argument) once we also support actually configuring `multipath.conf`.

WDYM?

darkmuggle:
Push a minor cosmetic fix.

jlebon:
Currently testing this!

cgwalters:
> Well, #364 is clearly dead. 

I didn't think so - We need *some* mechanism eventually for "initramfs configuration that can't be compressed into the kernel cmdline" right?

(I guess also this is a core tension in what I was arguing above for "Do what Anaconda does" for this - we can't do exactly the same thing unless we regenerate the initramfs client side)

> Running multipath -t shows the built-in defaults (see https://gist.github.com/darkmuggle/80fdd5d17bfb9b1a357bfa66e59a2675) which supports some 120+ SANS and iSCSI targets.

OK well, I know nothing about those.  If with this we can cover a lot of stuff out of the box with zero configuration - great!  My main concern then is stuff like "what happens when a machine has two different SANs attached", or how often it's necessary to tune any configuration for them.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
jlebon, line 1: I'm confused by the need for this generator. It seems to be mostly concerned with detecting whether multipath is used and if so assuming that the rootfs is on multipath and forcing the sysroot services to mount that instead, right?

I thought what we were settling on in https://github.com/coreos/ignition-dracut/pull/183 (see https://github.com/coreos/ignition-dracut/pull/183#issuecomment-628079503 and onwards) was to require users to put both `rd.multipath=default` and `root=/dev/disk/by-label/dm-mpath-root` on the kernel cmdline, right? Then systemd will automatically mount the rootfs once it comes online.

overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
```diff
@@ -0,0 +1,20 @@
...
+ENV{DM_ACTIVATION}!="1", GOTO="dm_label_end"
+
+# Unless we have a label, skip
+ENV{ID_FS_LABEL_ENC}!="?*", GOTO="dm_label_end"
+
+# Only act on filesystems
+ENV{ID_FS_USAGE}!="filesystem", GOTO="dm_label_end"
+
+ENV{DM_MPATH}=="?*", \
+    ENV{DM_SUSPENDED}=="Active"
```
jlebon: Hmm, noticed `systemd-udevd` says:

> /usr/lib/udev/rules.d/90-coreos-device-mapper.rules:15 The line takes no effect, ignoring.

jlebon (commented):
Seeing `ignition-ostree-mount-firstboot-sysroot` fail with:

> ignition-ostree-mount-sysroot: Failed to find /dev/disk/by-label/dm-mpath-root

Edit: sigh... OK right forgot to include the dracut patch for `rd.multipath=default`. Retesting!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
cgwalters: > root=/dev/disk/by-label/dm-mpath-root on the kernel cmdline

This makes total sense to start with, but I think the reason we're ending up with this "multipath labels" approach is because we also need to handle the boot partition.

And hmm in the general case with rootfs reprovisioning, we basically need to do something like use the `root=` karg to *start*, then handle rootfs, then potentially change the value of `root=` right?  For dm-crypt on multipath for example.

jlebon:
>> Well, #364 is clearly dead.
>
> I didn't think so - We need some mechanism eventually for "initramfs configuration that can't be compressed into the kernel cmdline" right?

The motivation for #364 isn't dead, but I think the last time we discussed this, we had rough agreement doing initrd regens at coreos-installer time for now and exploring enhancing that with initrd overlays? (See last line in https://github.com/coreos/fedora-coreos-config/pull/364#issuecomment-621520679).

cgwalters:
> The motivation for #364 isn't dead, but I think the last time we discussed this, we had rough agreement doing initrd regens at coreos-installer time for now and exploring enhancing that with initrd overlays? 

Hm I don't think I was there at that meeting.  Right, I'm not opposed to regenerating the initramfs - but doing so is a whole new level of "stuff that happens at coreos-installer time" and cuts into the story of "image system by default" and the cloud/metal symmetry.   But yeah, we can figure out initramfs overlays down the line.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
jlebon: > This makes total sense to start with, but I think the reason we're ending up with this "multipath labels" approach is because we also need to handle the boot partition.

Ahh gotcha.

Hmm, let's think about this. We have services which access the boot partition in read-only before multipath starts, and nothing in the initrd which needs rw access to it (until we move firstboot stamp file deletion into the initrd). And by the time real root comes up, the regular `/dev` symlinks should work fine, no?

jlebon:
Unthreading https://github.com/coreos/fedora-coreos-config/pull/392#discussion_r427521052 for visibility... testing this PR minus the `multipath-generator` with kargs `rd.multipath=default root=/dev/disk/by-label/dm-mpath-root` is working great!

jlebon:
So here's a question that might motivate the karg approach some more: shouldn't it be acceptable to have a multipathed partition on e.g. `/var` or `/var/lib/containers` or whatever? Right now with the generator, any presence of a multipath configuration assumes that it's for root, whereas ISTM like multipathing of devices and rootfs selection are strongly related but still independent things (and the same goes for e.g. RAID or LUKS or whatever).

darkmuggle:
> My main concern then is stuff like "what happens when a machine has two different SANs attached"

That's not uncommon in my experience.  And it's very common to have a whole bunch of disks. 
What would be uncommon would be to have two different filesystems present with labels of `root` (which, I think is far more unlikely than someone plugging in a USB disk with the label of root). For that the generator configures the diskful targets to use `/dev/disk/by-label/dm-mpath-{boot,root}`.  

> (I guess also this is a core tension in what I was arguing above for "Do what Anaconda does" for this - we can't do exactly the same thing unless we regenerate the initramfs client-side)

We are more or less doing that. 
- Anaconda does propagate the multipath configuration [1], and the WWIDs file is populated in root
- We're using the friendly name default in Mulitpath, which matches [2]

[1] https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/storage/devicetree/fsset.py#L655-L658
[2] https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/argument_parsing.py#L588-L589

cgwalters:
> So here's a question that might motivate the karg approach some more: shouldn't it be acceptable to have a multipathed partition on e.g. /var or /var/lib/containers or whatever? 

But multipath is about block devices, not partitions.  Why wouldn't that case be covered by having `/etc/multipath.conf` in the real root and `.mount` units etc?

jlebon:
>  Why wouldn't that case be covered by having /etc/multipath.conf in the real root and .mount units etc?

But don't we need to be able to e.g. create the partition and filesystem via Ignition? So we'd still need multipathd doing the binding in the initrd (and also `ignition-mount` mounting it so that `ignition-files` drops the right files on the right filesystems).

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
darkmuggle: The reason I added the generator was because things could get racy when booting with `rd.multipath=default` _without_ `root=....`. In testing, it would cause a race condition on `/dev/disk/by-label/{root,boot}`. 

Further, since `rd.multipath` as a karg only affects the loading of `multipathd` as it pertains to finding the root disk, teaching `ingition-diskful*.target` to use the mulitpathed devices seemed sensible. The rationale was that using `rd.multipath=default` as both the indicator for using multipath root and to set sensible defaults seemed sensible. 

The generator ensures that we don't start the `ignition-diskful*targets` until after multipath. For example:
```
ignition-ostree-mount-subsequent-sysroot.service
● ├─dev-disk-by\x2dlabel-dm\x2dmpath\x2droot.device
● ├─dev-disk-by\x2dlabel-root.device
● ├─multipathd.service
● └─system.slice
```

> Then systemd will automatically mount the rootfs once it comes online.

The other reason I went down this path was I looked at what `root=` actually does. The short story is that dracut creates a `sysroot.mount` dynamically like our diskful targets. 

IMO, the benefit to the generator is that:
- We preserve the first boot `ignition-diskful.target` work
- We don't train nor do we require users to use two kargs. Requiring users to know that they have to do `rd.multipath=default root=/dev/disk/by-label/dm-mpath-root` is better than `rd.multipath=default root=/dev/disk/by-id/wwn-0xx42834aa2e4158a1c-part4` but not by much.
- The "API" is a single karg. Should users wish to turn it on _after_ the fact, they can do so with an MCO kernel arg making it manageable at the cluster level for OCP. 
- We make it an all-or-none operation: if root is not on multipath then we fail and more to the point, its enforced. This seems to follow the CoreOS idea that if the user expects multipath, then we enforce that.

darkmuggle:
> > So here's a question that might motivate the karg approach some more: shouldn't it be acceptable to have a multipathed partition on e.g. /var or /var/lib/containers or whatever?
> 
> But multipath is about block devices, not partitions. Why wouldn't that case be covered by having `/etc/multipath.conf` in the real root and `.mount` units etc?

Multipath manages the device, and device-mapper uses a linear map to the partitions. As implemented here, `rd.multipath=default` is used to manage the device for root and any others. If `/var/lib/containers` is somewhere else, and its multipathed, then it will be managed as well. 

As far as ignition is concerned, `/dev/mapper/mpathf` is just a device. However, multipath must have claimed the device. The race for which device to use for root is the same one that affects Igntiion and why I went through the hoops with the generator: I wanted to block Ingition-Disks until _after_ multipath and the root multipath device was claimed.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: > Where is /dev/mapper/mpatha4 defined? How does it map to physical disks?

`mpath` is the default prefix for multipath when using the default of friendly names. When Mutliapth loads and sees a device with multiple paths it will create device-mapper targets counting from `a`. Since we ship with root on partition 4, we get ` mpatha4`. 

Perhaps this will help: 
```
# multipath -l /dev/mapper/mpatha
mpatha (0x267aea2de8329860) dm-0 NVME,VirtualMultipath
size=8.0G features='0' hwhandler='0' wp=rw
|-+- policy='service-time 0' prio=0 status=active
| `- 2:0:0:0 sda 8:0  active undef running
`-+- policy='service-time 0' prio=0 status=enabled
  `- 3:0:0:0 sdb 8:16 active undef running
```

darkmuggle: I should clarify that `/dev/mapper/mpatha4` is _assumed_ for the purposes of conversation within this PR with the assumption of one disk with two paths. The name is dependent upon the path discovery and it could be `/dev/mapper/mpath[a-z]4` Since we don't know what the multipath device housing root will be named, this PR includes a udev rule to create symlink. 

Also, for those following at home that is a virtual device create by Kola.

darkmuggle:
I did catch a race between `sfdisk` informing the kernel and `kpartx`.  Pushed a few comment changes, with the race fixed, and rebased.

overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
```diff
@@ -0,0 +1,20 @@
...
+KERNEL!="dm-*", GOTO="dm_label_end"
+
+# Ensure that the device mapper target is active
+ENV{DM_ACTIVATION}!="1", GOTO="dm_label_end"
+
+# Unless we have a label, skip
+ENV{ID_FS_LABEL_ENC}!="?*", GOTO="dm_label_end"
+
+# Only act on filesystems
+ENV{ID_FS_USAGE}!="filesystem", GOTO="dm_label_end"
```
dustymabe, line 12: thanks for these comments.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -14,15 +14,37 @@ shift
...
+    eval $(udevadm info --query property --export "${src}")
+    # get the partition, if any, and the name for the device mapper
+    partition="${ID_PART_ENTRY_NUMBER:-}"
+    dm_name="${DM_NAME//$partition/}"
+    # identify the type of device mapper.
+    subsystem=$(dmsetup info ${dm_name} -C -o subsystem --noheadings)
+
+    # for now, we only support multipath devices
+    if [ "${subsystem}" == "mpath" ] && [ -n "${partition}" ]; then
+        # growpart does not understand device mapper, instead of having sfdisk inform the kernel,
```
dustymabe, line 28: hmm - upstream bug anywhere ? I can't remember if we used growpart for Atomic Host or not but we did use LVM for Atomic Host so that might be something to look at

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -22,42 +28,54 @@ if findmnt --fstab /boot &>/dev/null; then
...
 fi
 
-# Don't create mount units for /boot or /boot/efi on live systems.
-# ConditionPathExists won't work here because conditions don't affect
-# the dependency on the underlying device unit.
-if [ ! -f /run/ostree-live ]; then
-    add_wants boot.mount
-    cat > "${UNIT_DIR}/boot.mount" <<EOF
+# Generate mount units that work with device mapper. The traditional
+# device unit (dev-disk-by\x2dlabel...) does not work since its not the
```
dustymabe: it's

dustymabe (approved):
I don't have any concrete objections other than small typos or questions. Thanks for the education in the comment discussions.

dustymabe:
typo in `Automatic Mutltipath with rd.mutlipath kargs ` commit title.

`rd.mutlipath` -> `rd.multipath`

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
darkmuggle: Thanks, and fixed.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
darkmuggle: https://bugs.launchpad.net/cloud-utils/+bug/1799953

`growpart` follows the link to the real target. In this specific case, it maps `/dev/mapper/mpatha` to `/dev/dm-0` and then blindly tries to resize `/dev/dm-04`.

darkmuggle:
@cgwalters @jlebon @dustymabe okay, I think we've reached the point having debated and discussed and the next step is to commit? @cgwalters and @dustymabe have approved, but in the spirit of teamwork I want to double-check that there are no substantive objections or improvements requested.

jlebon (approved):
> As implemented here, rd.multipath=default is used to manage the device for root and any others. If /var/lib/containers is somewhere else, and its multipathed, then it will be managed as well.

Right, but what if we *only* want e.g. `/var` multipathed?

> I wanted to block Ingition-Disks until after multipath and the root multipath device was claimed.

Ignition configs need to use the same `/dev/disk/by-label/dm-mpath-...` pointer (or using the wwn) and `ignition-disks` already knows to wait until devices it doesn't see show up.

> @cgwalters @jlebon @dustymabe okay, I think we've reached the point having debated and discussed and the next step is to commit?

My concern is that with this PR we're setting an API where `rd.multipath=default` is the only thing users need to add to get root on multipath (i.e. without an explicit `root=` karg) and that might bind us in the future if we stray from that assumption.

That said, if this is blocking more downstream work, I'm also fine to merge this and discuss that bit further outside this PR and do whatever fixups that come out of it separately! Thanks *so much* for working on this!

darkmuggle:
> > As implemented here, rd.multipath=default is used to manage the device for root and any others. If /var/lib/containers is somewhere else, and its multipathed, then it will be managed as well.
> 
> Right, but what if we _only_ want e.g. `/var` multipathed?

Well, that an easy answer. `rd.multipath=default` ensures that root is multipath. It setups the configuration for multipath. The users would provide an Ignition payload with their custom configuration. After first boot the user would regenerate their `initramfs`.

In all reality, if the user is looking to multipath `/var`  only (why?) then they will only present multiple paths for that LUN that will house `/var` which would prevent the other LUNS from being managed by mutlipathd. 

The default configuration is only propagated if there is `/sysroot/etc/mutlipath.conf`  (https://github.com/coreos/ignition-dracut/blob/master/dracut/30ignition/coreos-teardown-initramfs.sh#L108)

> My concern is that with this PR we're setting an API where `rd.multipath=default` is the only thing users need to add to get root on multipath (i.e. without an explicit `root=` karg) and that might bind us in the future if we stray from that assumption.

To counter that point, the udev rule becomes the API since we'd be telling them to use either the `/dev/disk/by-id/...<WWN>` or `/dev/disk/by-label/dm-mpath-root`.

darkmuggle:
@cgwalters @jlebon @dustymabe thank you all for your input, consideration, comments, debate and feedback.

PR #393, opened 2020-05-13
 20platform-chrony: Add support for AWS and GCP

cgwalters:
overlay.d: Rename 20azure-chrony → 20platform-chrony

In prep for adding AWS.

---

20platform-chrony: Add support for AWS

In AWS there's an NTP source accessible via the link-local IP,
which means it works in restricted networks (i.e. no Internet) too.

Should fix https://bugzilla.redhat.com/show_bug.cgi?id=1834252


cgwalters:
Tested this in AWS, but not yet re-tested in Azure.

cgwalters:
OK now that I look, GCP has similar recommendations too:
https://cloud.google.com/compute/docs/instances/managing-instances#configure-ntp

I'll do that next.

cgwalters:
(Should this be part of https://github.com/coreos/afterburn/ ?  )

cgwalters:
OK, tested on Azure, AWS, and GCP!

miabbott:
This looks pretty promising for solving clock issues in the cloud!  :+1: 

Regarding the RHCOS downstream, should we backport this to 4.3/4.4?

tests/kola/chrony/kola.json
```diff
@@ -0,0 +1,3 @@
+{
+    "platforms": "aws,azure,gcp"
```
cgwalters, line 2: Note of course kola isn't running on any of these by default; I tested the test manually by just copy-pasting the script to an instance and running it directly.  (This is an advantage of ext tests)

lucab:
@cgwalters yes, 3 is a clear pattern, I think it could fit into Afterburn. It would be a normal unit in initramfs, scheduled before ignition-files.

The string-replacing logic is quite brittle though, it would be great if upstream chrony landed logic for config fragment soon-ish (I know we have a BZ somewhere).

cgwalters:
Though, this configuration isn't *actually* dynamic; we just happen to do it on boot because of the "single update stream" issue.  So that's an argument against afterburn doing it.

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+set -euo pipefail
+# Configuring the timeserver for the platform is often handled
+# by pre-baking a config into a particular image for a platform, but
+# that doesn't work for us because we have a single update stream.  Hence
+# this generator dynamically inspects the platform and reconfigures chrony.
+#
+# AWS: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html
+# Azure: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync
```
rfairley, line 9: Nitpick: also note GCP URLs up here (same as written to the confpath already below)

rfairley (commented):
Minor comment - otherwise LGTM!

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -0,0 +1,67 @@
...
+    *) exit 0 ;;
+esac
+
+if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
+    echo "$self: /etc/chrony.conf is modified; not changing the default"
+    exit 0
+fi
+
+# Yeah this isn't a completely accurate kernel argument parser but
+# we don't have one shared across shell services at the moment.
```
jlebon: I think this comment was meant to be higher up?

jlebon (approved):
Minor comment, but LGTM as is too!

cgwalters:
Fixed the comment comments!

miabbott (approved):

miabbott:
> @cgwalters yes, 3 is a clear pattern, I think it could fit into Afterburn. It would be a normal unit in initramfs, scheduled before ignition-files.
> 
> The string-replacing logic is quite brittle though, it would be great if upstream chrony landed logic for config fragment soon-ish (I know we have a BZ somewhere).

https://bugzilla.redhat.com/show_bug.cgi?id=1828434

cgwalters:
In looking further into this I found
https://opensource.com/article/17/6/timekeeping-linux-vms
which basically says to `modprobe ptp_kvm` and then configure chrony in exactly the same way we're doing here which covers KVM (and presumably OpenStack for example).

It certainly seems to work for me; though loading a kernel module is probably arguing for refactoring this into cleaner "not bash" code.

PR #395, opened 2020-05-13
manifest: add catatonit

### LINKED ISSUE #478: Include catatonit
Right now, `podman run --init` doesn't work:

```
[root@cosa-devsh ~]# sudo podman run -ti --privileged --rm --init registry.fedoraproject.org/fedora:31
Trying to pull registry.fedoraproject.org/fedora:31...
Getting image source signatures
Copying blob a523108545c2 done
Copying config fac75bdc1c done
Writing manifest to image destination
Storing signatures
Error: container-init binary not found on the host: stat /usr/libexec/podman/catatonit: no such file or directory
```

This used to be the status quo because catatonic wasn't bundled nor shipped in Fedora: https://github.com/containers/libpod/issues/4159. It's shipped now: https://bodhi.fedoraproject.org/updates/?packages=catatonit.

We should consider adding it to FCOS because `--init` is important for long-running containerized services so they don't collect tons of zombie processes.

jlebon:
podman needs catatonit in order to use `--init`, which is important for
long-running containerized services.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/478

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -78,6 +78,9 @@
     "ca-certificates": {
       "evra": "2020.2.40-1.1.fc31.noarch"
     },
+    "catatonit": {
+        "evra": "0.1.5-2.fc31.x86_64"
+    },
     "chrony": {
       "evra": "3.5-4.fc31.x86_64"
     },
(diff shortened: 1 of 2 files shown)


jlebon:
Just putting this out since I think we should include it and so CI gives us a diff.

jlebon:
```
# rpm -qi catatonit | grep -i size
Size        : 786574
```

No added deps.

jamescassell (approved):
confirmed no extra deps, 780k on disk.

cgwalters (approved):

PR #398, opened 2020-05-14
coreos-copy-firstboot-network.service: run before multipathd

jlebon:
This is required because the unit needs read-only access to `/boot` in
order to extract any baked in network config and multipathd might claim
ownership.

See similar patch for `ignition-setup-user.service`:
https://github.com/coreos/ignition-dracut/pull/185

For more information, see:
https://github.com/coreos/ignition-dracut/pull/183#issuecomment-627628526

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
@@ -34,7 +34,11 @@ Before=dracut-initqueue.service
 After=dracut-cmdline.service
 # Since we are mounting /boot/, require the device first
 Requires=dev-disk-by\x2dlabel-boot.device
-After=dev-disk-by\x2dlabel-boot.device   
+After=dev-disk-by\x2dlabel-boot.device
+
+# We want to make sure we're not racing with multipath taking ownership of the
+# boot device.
+Before=multipathd.service
 
 [Service]
 Type=oneshot


overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
```diff
@@ -34,7 +34,11 @@ Before=dracut-initqueue.service
 After=dracut-cmdline.service
 # Since we are mounting /boot/, require the device first
 Requires=dev-disk-by\x2dlabel-boot.device
-After=dev-disk-by\x2dlabel-boot.device   
+After=dev-disk-by\x2dlabel-boot.device
+
+# We want to make sure we're not racing with multipath taking ownership of the
+# boot device.
+Before=multipathd.service
```
cgwalters, line 41: Seems OK for now, but is it really just multipathd?  There's also iSCSI.  I think the generalization of this is probably going to be some sort of target like Ben was iterating on.

But seems fine for now!

dustymabe:
My understanding from our conversations was that multipathd ran after dracut-initqueue. This clearly runs before dracut-initqueue. Was my understanding incorrect?

jlebon:
> My understanding from our conversations was that multipathd ran after dracut-initqueue. This clearly runs before dracut-initqueue. Was my understanding incorrect?

AFAICT, they're not strictly ordered wrt each other. Both of them have `After=systemd-udev-trigger.service`:

https://github.com/dracutdevs/dracut/blob/1ec7b694b55c70d06f262a96c320709b70d20bcb/modules.d/90multipath/multipathd.service#L5
https://github.com/dracutdevs/dracut/blob/1ec7b694b55c70d06f262a96c320709b70d20bcb/modules.d/98dracut-systemd/dracut-initqueue.service#L11

jlebon:
Nowadays, the approach is to have multipath claim the bootfs first before we try any access due to the potential non-optimized path issue.

PR #403, opened 2020-05-18
coreos-update-ca-trust: Run After=sysinit.target

cgwalters:
I am seeing this service fail in RHCOS when rebasing ostree
to a version that supports https://github.com/ostreedev/ostree/pull/1767

When using `DefaultDependencies=no`, one really wants to specify *some*
dependency - very few units have no dependencies at all.

In this case, let's run "early" which means after `sysinit.target`
but before `basic.target`, which will ensure we run after `ostree-prepare-root.service`.


lucab (approved):

dustymabe:
> In this case, let's run "early" which means after `sysinit.target`
> but before `basic.target`, which will ensure we run after `ostree-prepare-root.service`.

could we add this comment to the unit file?

dustymabe (approved):
LGTM

PR #407, opened 2020-05-18
WIP: overlay: Don't mount /boot/efi by default

cgwalters:
Nothing in the OS touches the ESP by default, so there's
no reason to mount it by default, particularly writable.
This is good for avoiding wear&tear on the filesystem, but
I am specifically doing this as preparation for potentially
removing the ESP from AWS images, because AWS `ImportImage`
chokes on its presence:
https://github.com/openshift/os/pull/396

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -33,10 +33,11 @@ fi
 # device that systemd will fsck. This code ensures that if the label
 # is backed by a device-mapper target the dev-mapper.*.device is used.
 mk_mount() {
-    local unit_name="${1}.mount"; shift
+    local mount_pt=${1}; shift
     local label="${1}"; shift
-    local mount_pt="${1:-/$label}"
+    local conditions="${1:-}"
     local path="/dev/disk/by-label/${label}"
+    local unit_name=$(systemd-escape -p ${mount_pt} --suffix=mount)
 
     eval $(udevadm info --query property --export "${path}")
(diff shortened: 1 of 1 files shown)


jlebon:
> I am specifically doing this as preparation for potentially
removing the ESP from AWS images, because AWS `ImportImage`
chokes on its presence:

Yeah... I was looking into the RHBZ behind this this morning. We're probably going to have to do this. :(

Hmm, if we do do this, I wonder if it'd work to instead just put the symlink in `$UNIT_DIR/dev-disk-by\x2dlabel-EFI\x2dSYSTEM.device.wants/`? That way, we still mount it on platforms that have it. (Not against just not mounting it outright, though I do like the symmetry.)

cgwalters:
> Hmm, if we do do this, I wonder if it'd work to instead just put the symlink in $UNIT_DIR/dev-disk-by\x2dlabel-EFI\x2dSYSTEM.device.wants/?

Err...that's backwards right?  I guess it'd work because actually in systemd `Wants=` is distinct from `After=` but...

For now I pushed some code to just not mount on `aws` but I need to actually test this.

jlebon:
>  I guess it'd work because actually in systemd `Wants=` is distinct from `After=` but...

Right, it's just about pulling the unit into the transaction.

Making it conditional on `ignition.platform.id` WFM though it seems nicer to just automatically mount it if we find the partition. There's probably a cleaner way to do this than the `.device.wants` hack I suggested above.

bgilbert:
Should be obsoleted as part of https://github.com/coreos/fedora-coreos-config/pull/794.

bgilbert:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/794.

PR #408, opened 2020-05-18
20platform-chrony: Always resynchronize after drift of one second

sohankunkerkar:
Context: [Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1780165]). This fix is applicable to only three cloud providers i.e. `aws/gcp/azure`. 
/cc @cgwalters


overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -36,17 +36,20 @@ case "${platform}" in
     azure) 
         (echo '# See also https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync'
          sed -e s,'^pool,#pool,' < /etc/chrony.conf
+         sed -e 's/makestep\ 1.0\ 3/makestep\ 1\ -1/g' < /etc/chrony.conf
```
cgwalters: I think actually we could move these "change default chrony config" bits to above.  When it was just one line the duplication wasn't so bad, but with two I think it's worth centralizing.

cgwalters (changes requested):
Can you please match commit message style here, specifically:

- Include "topic prefix" (like `20platform-chrony:`)
- Imperative style, like "Always resynchronize after drift of one second"
- No period at end

Putting it together e.g.

`20platform-chrony: Always resynchronize after drift of one second`

You can see plenty of examples of this style in e.g. [this repository](https://github.com/coreos/fedora-coreos-config/commits/testing-devel), the [linux kernel git](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/), [ostree](https://github.com/ostreedev/ostree/commits/master), [systemd](https://github.com/systemd/systemd/commits), etc.

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
sohankunkerkar: Yeah, it makes sense to me.

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -31,7 +31,9 @@ if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
     exit 0
 fi
 
-echo "# Generated by $self - do not edit directly" > "${confpath}"
+(echo '# Generated by $self - do not edit directly' 
+ sed -e 's/makestep\ 1.0\ 3/makestep\ 1.0\ -1/g' < /etc/chrony.conf
```
cgwalters: I think we need to also move up the `sed -e s,'^pool,#pool,' < /etc/chrony.conf` here, otherwise we're duplicating the contents of the file.  Something like:
`sed -e 's/makestep\ 1.0\ 3/makestep\ 1.0\ -1/g' -e s,'^pool,#pool,' < /etc/chrony.conf`
right?

BTW if you want to test this out, a simple hack is to e.g. in `cosa run` (qemu) just `vi /boot/loader/entries/` and change `ignition.platform.id=azure`, then `systemctl reboot`, then:

```
rpm-ostree usroverlay
vi /usr/lib/systemd/system-generators/coreos-platform-chrony
# I use copy/paste from my host editor, but edit however you want, e.g. you can use the 9p mount in /var/mnt/workdir or ssh or whatever
systemctl daemon-reload
```

cgwalters: Or if you want to test this "for real", use e.g. `@cluster-bot` via e.g. `launch 4.5 azure`, then perform the same procedure above, but then the resulting chrony config should actually work, e.g. you can validate it with `chronyc sources`.

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -31,22 +31,21 @@ if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
     exit 0
 fi
 
-echo "# Generated by $self - do not edit directly" > "${confpath}"
+(echo '# Generated by $self - do not edit directly'
```
dustymabe: `$self` is now inside of single quotes, which means the variable won't expand. I think you want to use double quotes here.

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -31,22 +31,21 @@ if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
     exit 0
 fi
 
-echo "# Generated by $self - do not edit directly" > "${confpath}"
+(echo '# Generated by $self - do not edit directly' 
+ sed -e 's/makestep\ 1.0\ 3/makestep\ 1.0\ -1/g' -e s,'^pool,#pool,' < /etc/chrony.conf
```
dustymabe: chrony.conf currently has:

```
# Allow the system clock to be stepped in the first three updates
# if its offset is larger than 1 second.
makestep 1.0 3
```

So making this sed substitution makes the comment wrong :( Any ideas for how to correct or delete the comment? One option would be to have the sed statement comment out any lines starting with `makestep` and then append the new `makestep` setting at the bottom along with a new more accurate comment.

dustymabe:
It would help if your commit message summarized the BZ and the desired outcome and why making this change achieves that goal.

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
cgwalters: I'm not sure I'd say we should block on fixing that but if someone knows the requisite `sed` incantations then yeah.

dustymabe: this should work:

```
$ (echo '# Generated by $self - do not edit directly' ; sed -e s,'^makestep,#makestep,' -e s,'^pool,#pool,' < /etc/chrony.conf; echo -e "\n# put nice comment here\n# and here\nmakestep 1.0 -1\n")
# Generated by $self - do not edit directly
# Use public servers from the pool.ntp.org project.
# Please consider joining the pool (http://www.pool.ntp.org/join.html).
#pool 2.fedora.pool.ntp.org iburst

# Record the rate at which the system clock gains/losses time.
driftfile /var/lib/chrony/drift

# Allow the system clock to be stepped in the first three updates
# if its offset is larger than 1 second.
#makestep 1.0 3

# Enable kernel synchronization of the real-time clock (RTC).
rtcsync

# Enable hardware timestamping on all interfaces that support it.
#hwtimestamp *

# Increase the minimum number of selectable sources required to adjust
# the system clock.
#minsources 2

# Allow NTP client access from local network.
#allow 192.168.0.0/16

# Serve time even if not synchronized to a time source.
#local stratum 10

# Specify file containing keys for NTP authentication.
keyfile /etc/chrony.keys

# Get TAI-UTC offset and leap seconds from the system tz database.
leapsectz right/UTC

# Specify directory for log files.
logdir /var/log/chrony

# Select which information is logged.
#log measurements statistics tracking

# put nice comment here
# and here
makestep 1.0 -1
```

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -31,22 +31,21 @@ if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
     exit 0
 fi
 
-echo "# Generated by $self - do not edit directly" > "${confpath}"
+(echo "# Generated by $self - do not edit directly" 
+ sed -e 's/makestep\ 1.0\ 3/makestep\ 1.0\ -1/g' -e s,'^pool,#pool,' < /etc/chrony.conf
+) >> "${confpath}"
```
dustymabe: we should leave this as `>` since previously we had `echo "# Generated by $self - do not edit directly" > "${confpath}"`

dustymabe (approved):
current set of changes LGTM - we'll want to try to recreate the failure scenario and verify the change if possible

sohankunkerkar:
> current set of changes LGTM - we'll want to try to recreate the failure scenario and verify the change if possible

Just confirmed that by introducing `makestep 1.0 -1` in the `coreos-platform-chrony.conf` and changing the system's time to some earlier time by stopping the `chronyd` service, and then starting that service again solves the time resynchronization issue.

jlebon (approved):

dustymabe:
> It would help if your commit message summarized the BZ and the desired outcome and why making this change achieves that goal.

Nice update to the commit message. The added context is so helpful!

cgwalters (approved):

PR #406, opened 2020-05-18
tree: promote changes from next-devel at a1c57faa13ed9c65e2679b7076aa…

lucab:
…7ab076682182

PR #409, opened 2020-05-18
[next] manifest: remove non fedora-coreos-pool repos

dustymabe:
We've done this for the `testing` and `stable` branches already
so it should probably be done here for `next` too.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -10,13 +10,6 @@ rojig:
 
 repos:
   - fedora-coreos-pool
-  # these repos are there to make it easier to add new packages to the OS and to
-  # use `cosa fetch --update-lockfile`; but note that all package versions are
-  # still pinned
-  - fedora
-  - fedora-updates
-  - fedora-modular
-  - fedora-updates-modular
 
 add-commit-metadata:
   fedora-coreos.stream: next


lucab:
It looks like CI detected an Ignition-related regression after this:
```
--- FAIL: coreos.ignition.journald-log (23.45s)
journaldEntry.go:46: Ignition didn't write 57124006b5c94805b77ce473e92a8aeb
```

dustymabe:
yeah I think this is because #406 hadn't merged yet (which brought the new ignition version into `next`). Just rebased and it should pass CI now.

lucab:
@dustymabe I possibly have some minor confusion here, shouldn't this be targeted at `next-devel` or `testing-devel` branch?

dustymabe:
If you look at the manifest.yaml in the `testing` and `stable` branches you see they don't have any non `fedora-coreos-pool` repos. This PR makes `next` match what those are already doing.

jlebon (approved):
Makes sense, thanks!

jlebon:
> @dustymabe I possibly have some minor confusion here, shouldn't this be targeted at next-devel or testing-devel branch?

Specifically, note that `manifest.yaml` in each branch is one of the few files that aren't synced/promoted between branches. It's what contains the "identity" of the stream (see https://github.com/coreos/fedora-coreos-config#about-this-repo). So for production branches, we really do expect now all the packages to be locked and thus come from the pool.

A follow-up to this is to have the pool repo in a shared manifest since that's in common (see https://github.com/coreos/fedora-coreos-config/pull/355#issuecomment-620860762), though we need the new rpm-ostree in cosa first -- just tagged it into the continuous tag now.

PR #410, opened 2020-05-19
overrides: prune overrides that have made it to stable

dustymabe:


dustymabe:
I'll do a separate PR for `next` once the podman fix goes into stable (in the next bodhi run)

lucab (approved):

PR #411, opened 2020-05-19
overlay/live-generator: Add requires/after for rootfs on ISO

cgwalters:
We are seeing RHCOS ISO bootup fail sometimes, I think we're racing
with some udev probe transiently inspecting the CD.
I believe that by adding this `After=`, we're serializing on
udev probing being complete for that device.  Plus, it's
just best practice anyways.


cgwalters:
(Testing this locally)

dustymabe:
seems reasonable. This does appear to be a race condition so it might be worth trying to loop to see how consistent it is before and after this change. For example I just ran my install again and it booted fine so I'm 1/2 on hitting the issue. Hopefully after this change we don't hit it at all.

cgwalters:
OK tested and I'm so far 8/8 on boots.

dustymabe (approved):

PR #414, opened 2020-05-20
README: additions to the new Fedora major SOP notes

dustymabe:
- README: update new major version SOP to include barriers
- README: update new major version SOP to include CoreOS Installer

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,13 @@ Update server changes:
 1. Set a new update barrier for N-2 on all streams.
    See [discussion](https://github.com/coreos/fedora-coreos-tracker/issues/480#issuecomment-631724629).
 
+CoreOS Installer changes:
+
+1. Update CoreOS Installer to know about the signing key used for the
+   future new major version of Fedora (N+1). Note that the signing
+   keys for N+1 won't get created until releng branches and rawhide
+   becomes N+1.
+
 Release engineering changes:
 
 1. verify that the `f${releasever}-coreos-signing-pending` Koji tag has
--- a/README.md
+++ b/README.md
@@ -88,16 +88,21 @@ one easy way to do this is for now:
 - run `cosa fetch --update-lockfile`
 - commit only the new package entries
 
-## Moving to a new major version of Fedora
+## Moving to a new major version (N) of Fedora
 
-Updating this repo itself is easy:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,6 +102,11 @@ Update server changes:
 1. Set a new update barrier for N-2 on all streams.
    See [discussion](https://github.com/coreos/fedora-coreos-tracker/issues/480#issuecomment-631724629).
 
+CoreOS Installer changes:
+
+1. Update CoreOS Installer to know about the signing key used for the
+   new major version of Fedora (N).
```
jlebon: Hmm, should this be about N+1 instead? I would expect coreos-installer to already have the key for N much earlier than when this SOP needs to happen since normally `next` would've been on N already before and so coreos-installer would already need to know about the N key. It feels a bit odd too to block the SOP on a new coreos-installer release and respin (whereas if it's N+1, it's more something that can be PR'ed but doesn't require a release right away).

bgilbert: I agree, assuming the N+1 key will be available at that time.

dustymabe: ok pushed an update to try to correctly articulate this. LMK what you think.

README.md
```diff
@@ -88,16 +88,28 @@ one easy way to do this is for now:
...
+Update server changes:
+
+1. Set a new update barrier for N-2 on all streams.
+   See [discussion](https://github.com/coreos/fedora-coreos-tracker/issues/480#issuecomment-631724629).
+
+CoreOS Installer changes:
+
+1. Update CoreOS Installer to know about the signing key used for the
+   future new major version of Fedora (N+1). Note that the signing
+   keys for N+1 won't get created until releng branches and rawide
```
jlebon:
```suggestion
   keys for N+1 won't get created until releng branches and rawhide
```

jlebon (approved):
Thinking more on this, once we have a rawhide mechanical stream, I think we'll just have to add the keys as they become available so that we can keep running that stream with signing on, right? But anyway, we can refine these instructions if that happens.

One comment, LGTM otherwise!

README.md
dustymabe: fixed

dustymabe:
skipping CI tests since it was a one line change in the readme

PR #415, opened 2020-05-21
overlay: Support coreos-installer iso embed --disable-initramfs-networking

cgwalters:
We want people to be able to take our ISO and inject
Ignition so they can fully automate installs - all that
should be needed is to e.g. attach the ISO via an IMPI system,
or written to a USB stick etc.

Today though, we enable the default DHCP if Ignition is
provided, which breaks automating static IP installs.

We've debated a lot of approaches for this, and general
consensus is that the best is probably "conditional networking"
in Ignition.  However that's a much larger amount of work that
also touches every artifact type.  This approach is
small and targeted and only affects the Live ISO path.

We also discussed generalizing this into arbitrary kernel
arguments, but the problem with that is that today we can only
inject into the initramfs, so e.g. one can't enable SMT
that way.  To fix that we probably need to enhance our ISO
model to be able to actually modify the kernel config too.
That would be another potential good future enhancement.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
@@ -6,13 +6,14 @@
 # - the user didn't already request networking via rd.neednet
 # - the user provided a ignition.config.url karg, implying
 #   the need for networking
-# - there is an embedded ignition config
+# - there is an embedded ignition config (unless disabled, see below)
 #
 # For the case of the embedded Ignition config there could be a
 # case where the user embeds an Ignition config (via coreos-installer
-# iso embed) but doesn't want networking. In that case we'll have
-# smarter detection in the future (https://github.com/coreos/fedora-coreos-tracker/issues/443)
-# but the user can override with `rd.neednet=0` now if needed.
+# iso embed) but doesn't want networking (in particular the DHCP default).
+# As of recently we support:
+# `coreos-installer iso embed --disable-initramfs-networking`
+# but the goal is smarter detection in the future (https://github.com/coreos/fedora-coreos-tracker/issues/443).
 #
 # If we do determine we need network and there are no other
 # `ip=` kargs then we'll use `ip=dhcp,dhcp6` by default.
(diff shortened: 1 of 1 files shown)


cgwalters:
Pairs with https://github.com/coreos/fedora-coreos-config/pull/415

cgwalters:
Tested this with 
```
$ cp --reflink=auto builds/latest/x86_64/fedora-coreos-*-live.x86_64.iso tmp/live.iso && ~/src/github/coreos/coreos-installer/target/debug/coreos-installer iso embed -c ~/src/github/cgwalters/playground/fcct/autologin.ign tmp/live.iso --disable-initramfs-networking
```
Then in libvirt I created a network with no DHCP and booted the ISO in it.

cgwalters:
Honestly the more I think about it the more I feel like perhaps we should require people to explicitly *enable* networking in the ISO initramfs by default.  This would admittedly be a behavioral change affecting latest stable FCOS but...OTOH this is all basically all brand new and I doubt many people would be affected.

cgwalters:
> Tested this

BTW, I did double check that today `kola testiso --scenarios iso-offline-install` today fully disables networking in qemu.  But *that* case happens to work because NM just does nothing if no devices were found.  The problem arises when we have network devices (as in a physical server), we just don't want to do DHCP on them.

jlebon:
> Honestly the more I think about it the more I feel like perhaps we should require people to explicitly enable networking in the ISO initramfs by default.

Hmm, yeah maybe that's cleaner. The mental model would definitely be simpler. Because then we can say "networking is disabled on the live ISO unless explicitly enabled", instead of "networking is disabled on the live iSO unless you inject an Ignition config... unless you pass `--disable-initramfs-networking`".

I think it might flow better with the long-term goal of conditional networking as well, because making networking opt-in is not incompatible with us automatically turning it on if we deem necessary. OTOH, if it's opt-out, then it's a footgun for if the Ignition config *does* require networking.

dustymabe:
maybe we can circle back on this in IRC in 30 minutes? (got a 1:1, then I'm free).

cgwalters:
Yep.  Though one other downside of no initramfs networking on ISO by default is that ISO and PXE become less symmetrical.  But eh, I am increasingly feeling it's cleaner.  We could drop a lot of special code in our generator, it's easier to explain.

cgwalters:
Moving this to https://github.com/coreos/fedora-coreos-config/pull/417

PR #417, opened 2020-05-21
overlay: Don't enable Live ISO networking w/injected config

cgwalters:
We want people to be able to take our ISO and inject
Ignition so they can fully automate installs - all that
should be needed is to e.g. attach the ISO via an IPMI system,
or written to a USB stick etc.

Today though, we enable the default DHCP if Ignition is
provided via `coreos-installer iso embed`, which breaks
automating static IP installs.

We've debated a lot of approaches for this, and general
consensus is that the best is probably "conditional networking"
in Ignition.  However that's a much larger amount of work that
also touches every artifact type.  This approach is
small and targeted and only affects the Live ISO path.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
@@ -6,13 +6,10 @@
 # - the user didn't already request networking via rd.neednet
 # - the user provided a ignition.config.url karg, implying
 #   the need for networking
-# - there is an embedded ignition config
 #
-# For the case of the embedded Ignition config there could be a
-# case where the user embeds an Ignition config (via coreos-installer
-# iso embed) but doesn't want networking. In that case we'll have
-# smarter detection in the future (https://github.com/coreos/fedora-coreos-tracker/issues/443)
-# but the user can override with `rd.neednet=0` now if needed.
+# As of recently we skip networking if a config is injected via
+# `coreos-installer iso embed` in order to allow automating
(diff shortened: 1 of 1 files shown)


cgwalters:
Migrated from https://github.com/coreos/fedora-coreos-config/pull/415 based on IRC discussion.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
```diff
@@ -42,7 +39,6 @@ ConditionPathExists=/run/ostree-live
 # We'll assume we need network in either of these two following
 # cases (see description from above)
```
dustymabe:
```suggestion
```

cgwalters:
The basic concept here is that the use cases for initramfs networking with the ISO are very obscure - at the cost of symmetry w/our other paths (including PXE, with which this shares code), we make it easier for people to use the ISO for automation of arbitrarily complex installation paths solely via `coreos-installer iso embed`.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
```diff
@@ -42,7 +39,6 @@ ConditionPathExists=/run/ostree-live
 # We'll assume we need network in either of these two following
```
dustymabe:
```suggestion
# We'll assume we need network in the following case (see description from above)
```

dustymabe (approved):
LGTM with suggested comment tweaks to make them accurate.

jlebon (approved):

PR #418, opened 2020-05-21
90-coreos-device-mapper.rules: refinement on udev for device-mapper.

darkmuggle:
Cleanup on the device-mapper rules to make it a bit clearer and follow
other rules conventions.

Signed-off-by: Ben Howard <ben.howard@redhat.com>

### DIFF
--- a/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
+++ b/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
@@ -3,18 +3,23 @@ SUBSYSTEM!="block", GOTO="dm_label_end"
 KERNEL!="dm-*", GOTO="dm_label_end"
 
 # Ensure that the device mapper target is active
+# And the required attributes exist.
 ENV{DM_ACTIVATION}!="1", GOTO="dm_label_end"
-
-# Unless we have a label, skip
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
```diff
@@ -3,18 +3,23 @@ SUBSYSTEM!="block", GOTO="dm_label_end"
...
+# And the required attributes exist.
 ENV{DM_ACTIVATION}!="1", GOTO="dm_label_end"
-
-# Unless we have a label, skip
 ENV{ID_FS_LABEL_ENC}!="?*", GOTO="dm_label_end"
+ENV{ID_FS_UUID_ENC}!="?*", GOTO="dm_label_end"
 
-# Only act on filesystems
+# Only act on filesystems. This should prevent layered devices
+# such as Raid on Mulitpath devices from appearing.
```
miabbott: s/Mulitpath/Multipath/

cgwalters (approved):
I don't really know udev rules (I try hard to avoid them), but I'd trust you and this seems at least superficially sane.

overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules
darkmuggle: ah, fixed

PR #419, opened 2020-05-22
multipath-generator: configure boot.mount

darkmuggle:
With a recent change in Ingition-Dracut, the boot label race was
re-introduced. With https://github.com/coreos/ignition-dracut/pull/187,
boot.mount was added, giving the ability to control which device is
mounted as /boot.

This eliminates any race-condition chance on /dev/disk/by-label/boot.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
@@ -68,7 +68,7 @@ Requires=${root_device}
 After=${root_device}
 EOF
 
-# Reconfigure ignition-ostree-mount-subsequent-sysroot.service.d, which is
+# Reconfigure ignition-ostree-mount-subsequent-sysroot.service.d
 # required by ignition-diskful-subsquent.target.
 secondboot_cfg="${UNIT_DIR}/ignition-ostree-mount-subsequent-sysroot.service.d"
 mkdir -p "${secondboot_cfg}"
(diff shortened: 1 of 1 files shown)


cgwalters:
Is this waiting on https://github.com/coreos/ignition-dracut/pull/187 or will it also work with the existing code?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/multipath-generator
```diff
@@ -81,3 +81,18 @@ Wants=multipathd.service
 After=${root_device}
 Requires=${root_device}
 EOF
+
+boot_cfg="${UNIT_DIR}/boot.mount.d"
+mkdir -p "${boot_cfg}"
+cat > "${boot_cfg}/ignition-mpath.conf" <<EOF
+[Unit]
+After=
```
cgwalters, line 89: Entirely blowing away the original `After/Requires=` seems like it's going to be prone to problems if at some point the unit has an `After=` for something other than the `label-boot.device`.   Not a blocker necessarily though.

jlebon: Hmm, I think we could keep the original `After=` and `Requires=` here since those technically do still exist even on `multipathd`. (But of course this unit would still add a `Requires`/`After` and override `What` to point to the devicemapper version.)

darkmuggle:
Dropping this since `boot.mount` conflicts with the dependencies on ignition-setup. I haven't found a way to make the mount unit work with needing to update the udev symlinks.

PR #421, opened 2020-05-22
[next-devel] overrides: fast-track ostree-2020.3-4

jlebon:
Which includes a patch to backport `sysroot.readonly` neutering:

coreos/fedora-coreos-tracker#488
ostreedev/ostree#2108

PR #422, opened 2020-05-22
[next-devel] overrides: prune overrides that have made it to stable

dustymabe:


jlebon (approved):
I can rebase https://github.com/coreos/fedora-coreos-config/pull/421 on top of this if it goes in first.

dustymabe:
> I can rebase #421 on top of this if it goes in first.

thanks!

PR #423, opened 2020-05-22
overlay/15fcos: Make coreos-check-ssh-keys denote where key came from

dustymabe:
This will make it so that the Issue will show which source the ssh
came from. It shows:

```
Ignition: wrote ssh authorized keys file for user: core
Ignition: wrote ssh authorized keys file for user: elroy
Afterburn: wrote ssh authorized keys file for user: elmer
```

Versus:

```
wrote ssh authorized keys file for user: core
wrote ssh authorized keys file for user: elroy
wrote ssh authorized keys file for user: elmer
```


sohankunkerkar (approved):
This looks much better. That way we can identify the message/user based on the provider name. 
I think I like `ssh authorized keys provided via ${provider-name} for user: core ` UI too.

dustymabe:
> This looks much better. That way we can identify the message/user based on the provider name.

Thanks!

> I think I like `ssh authorized keys provided via ${provider-name} for user: core ` UI too.

Yeah doing that would be trickier.

PR #424, opened 2020-05-22
[hold] Use new cosa build naming model

cgwalters:
https://github.com/coreos/coreos-assembler/pull/1459 is
going to add support for more convenient multiple builds.

Here the `name` is derived simply from the name of the manifest file.
We use the summary for cloud uploads, so that moves to `image.yaml`.


cgwalters:
Pairs with https://github.com/coreos/coreos-assembler/pull/1474 - don't merge until that merges.

fedora-coreos.yaml
```diff
@@ -0,0 +1,32 @@
+ref: fedora/${basearch}/coreos/testing-devel
```
jlebon, line 1: Hmm, this is a bit awkward, because there is also a `manifests/fedora-coreos.yaml`. Maybe let's rename that one to reduce confusion?

This will also require updating config-bot, which right now special-cases `manifest.yaml` (see https://github.com/coreos/fedora-coreos-config#about-this-repo).

jlebon: Hmm, I wonder if it'd be simpler to just make https://github.com/coreos/coreos-assembler/pull/1474 use a `name` field in `image.yaml` instead? It's not strictly correct since e.g. the OSTree tarball has that name too, though it does seem cleaner to have it be more explicit and sitting alongside `summary`.

cgwalters:
Closing this out until we finalize the cosa model.

PR #426, opened 2020-05-25
Move to conditional networking

### LINKED ISSUE #443: Don't bring up networking in the initramfs on first boot by default
Currently, Fedora CoreOS and RHEL CoreOS always try to bring up networking on first boot in the initramfs: https://github.com/coreos/coreos-assembler/blob/87098329e7d3112c8544d3706de45581ce0c4d59/src/grub.cfg#L48.

The reason for this is that we can't be sure whether Ignition will require access to the network to fetch remote resources.

However, this causes a host of problems. Essentially, in any environment that strays from DHCP (e.g. static IP or no networking), it's a pain or sometimes impossible to have to change kernel arguments either at install time, or during the first boot itself. The networking kargs are also not a nice interface.

One obvious example is the live ISO; if no Ignition config is provided (or if one is provided but doesn't require networking), there's no reason to bring up networking (https://github.com/coreos/fedora-coreos-tracker/issues/349).

We should make networking transparently optional by only bringing it up when Ignition needs it.

jlebon:
We have all the piece in place now to move to conditional networking. So
let's drop the firstboot kargs, as well as
coreos-liveiso-network-kargs.service, which is no longer needed (i.e.
the live ISO will now enable initrd networking as required given the
embedded Ignition config).

Fixes: coreos/fedora-coreos-tracker#443

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -31,7 +31,11 @@ add_requires sysroot.mount     initrd-root-fs.target
 add_requires sysroot-etc.mount initrd-root-fs.target
 add_requires sysroot-var.mount initrd-root-fs.target
 
-add_requires coreos-liveiso-network-kargs.service              initrd.target
+# Need to be flexible here until RHCOS moves to spec3.
+if ! (ignition -help || :) |& grep -q 'fetch-offline'; then
+    add_requires coreos-liveiso-network-kargs.service              initrd.target
+fi
+
 add_requires coreos-liveiso-reconfigure-nm-wait-online.service initrd.target
 
 mkdir -p "${UNIT_DIR}/ostree-prepare-root.service.d"
--- a/image.yaml
+++ b/image.yaml
@@ -9,9 +9,9 @@ extra-kargs:
     # https://fedoraproject.org/wiki/Changes/CGroupsV2
     - systemd.unified_cgroup_hierarchy=0
 
-# Kernel arguments to be used on first-boot.
-ignition-network-kcmdline:
-    - 'rd.neednet=1'
+# Disable networking by default on firstboot. We can drop this once cosa stops
+# defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
+ignition-network-kcmdline: []
 
 # Optional remote by which to prefix the deployed OSTree ref
 ostree-remote: fedora
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
@@ -43,6 +43,10 @@ After=coreos-gpt-setup@dev-disk-by\x2dlabel-root.service
 # Since we are mounting /boot/, require the device first
 Requires=dev-disk-by\x2dlabel-boot.device
 After=dev-disk-by\x2dlabel-boot.device
+# Need to run after fetch-offline stage since it may re-run the NM cmdline
+# hook which will generate NM configs from the network kargs, but we want to
+# have precedence.
+After=ignition-fetch-offline.service
 
 [Service]
 Type=oneshot
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.service
@@ -24,6 +24,10 @@
 # prompt without requiring networking on boot. The user can
 # then configure the networking interactively.
 #
+# Note that this script is only used on RHCOS now. We should be able to remove
+# it once RHCOS moves to spec3. On FCOS/spec3, this is replaced by the
+# conditional networking work:
+# https://github.com/coreos/fedora-coreos-config/pull/426
 [Unit]
 Description=Request live ISO networking
 DefaultDependencies=no
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-network-kargs.sh
@@ -2,6 +2,11 @@
 
 # For a description of how this is used see coreos-liveiso-network-kargs.service
 
+# Note that this script is only used on RHCOS now. We should be able to remove
+# it once RHCOS moves to spec3. On FCOS/spec3, this is replaced by the
+# conditional networking work:
+# https://github.com/coreos/fedora-coreos-config/pull/426
+
 # Load the dracut library for getarg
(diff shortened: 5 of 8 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1373
Requires: https://github.com/coreos/ignition/pull/956
Requires: https://github.com/coreos/ignition-dracut/pull/164/

jlebon:
Rebased this on top of https://github.com/coreos/fedora-coreos-config/pull/427!

darkmuggle (commented):
:100: I _really_ like this.

jlebon:
This will be the final ratcheting point for FCOS. Will add a test so we can validate it before merging.

dustymabe (approved):
LGTM

jlebon:
Almost there! All remaining patches are merged, we just need an Ignition release at this point.

dustymabe:
> Almost there! All remaining patches are merged, we just need an Ignition release at this point.

Do we at least have any plans for that release to happen? /me really wants this in :smile:

jlebon:
> Do we at least have any plans for that release to happen? /me really wants this in smile

I think once https://github.com/coreos/ignition/pull/960 is in, we should cut a release. /cc @bgilbert @arithx

dustymabe:
New ignition version just landed! https://github.com/coreos/fedora-coreos-config/pull/513

jlebon:
Rebased this and marking ready for review! Still sanity-checking ISO behaviour locally as well.

dustymabe:
I'll take a re-look at this PR as well and try to build locally

jlebon:
> Still sanity-checking ISO behaviour locally as well.

Working well on FCOS!

But yeah I think for now we'll need to keep the `liveiso-network-kargs` bits for RHCOS until we move to spec3 there and gain the Ignition fetch-offline work. Will just tweak those to make them active on RHCOS only.

dustymabe:
This seems to be working well in my testing too! Though we did find https://github.com/coreos/fedora-coreos-config/issues/514 which needs to be addressed (though not in this PR). There is one subtle change here, which is if people were relying on networking configuration getting persisted into the real root and they were using `ip=` kargs for that in the past then they'll now need to make sure they also add `rd.neednet=1`, otherwise their `ip=` settings won't get applied persistently to the real root. Maybe a good point for a release note.

> But yeah I think for now we'll need to keep the `liveiso-network-kargs` bits for RHCOS until we move to spec3 there and gain the Ignition fetch-offline work. Will just tweak those to make them active on RHCOS only.

Does that mean you're going to do another push to this PR?

cgwalters:
> There is one subtle change here, which is if people were relying on networking configuration getting persisted into the real root and they were using ip= kargs for that in the past then they'll now need to make sure they also add rd.neednet=1, otherwise their ip= settings won't get applied persistently to the real root. Maybe a good point for a release note.

Nice catch!  Hmm...could we distinguish between "our default kargs" and "kargs provided by user" in that case though?

dustymabe:
> > There is one subtle change here, which is if people were relying on networking configuration getting persisted into the real root and they were using ip= kargs for that in the past then they'll now need to make sure they also add rd.neednet=1, otherwise their ip= settings won't get applied persistently to the real root. Maybe a good point for a release note.
> 
> Nice catch! Hmm...could we distinguish between "our default kargs" and "kargs provided by user" in that case though?

We might be able to do some trickery (@jlebon would probably know best since he did this most recent round of work), but I'm on the fence. Since it's a straightforward change I think I'd be more inclined to request the users who need that to just add the `rd.neednet=1` or even better use it as an opportunity to encourage them to put the config in the ignition config instead since they obviously don't need networking in the initramfs.

dustymabe:
the fix for #514 seems to be working great!

jlebon:
Yeah, though I'm hitting another issue now with the live ISO where it *seems* like `/var` isn't mounted when `ignition-files.service` runs. But it might be more subtle than that. (Iterating on the live ISO is painful. Need like a `cosa buildliveinitramfs-fast` or something which just disassembles the ISO and re-assembles it)

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -31,7 +31,11 @@ add_requires sysroot.mount     initrd-root-fs.target
 add_requires sysroot-etc.mount initrd-root-fs.target
 add_requires sysroot-var.mount initrd-root-fs.target
 
-add_requires coreos-liveiso-network-kargs.service              initrd.target
+# Need to be flexible here until RHCOS moves to spec3.
+if ! (ignition -help || :) |& grep -q 'fetch-offline'; then
+    add_requires coreos-liveiso-network-kargs.service              initrd.target
```
dustymabe, line 36: can we add a header to that service and shell script that it's only used for RHCOS now?

dustymabe (approved):
LGTM - one optional comment

jlebon:
> Yeah, though I'm hitting another issue now with the live ISO where it seems like /var isn't mounted when ignition-files.service runs. But it might be more subtle than that. (Iterating on the live ISO is painful. Need like a cosa buildliveinitramfs-fast or something which just disassembles the ISO and re-assembles it)

Ahh OK, this is another regression from v2.4.0: https://github.com/coreos/ignition/pull/1029. One can reproduce this by trying to add an SSH key for the `root` user.

jlebon:
> Since it's a straightforward change I think I'd be more inclined to request the users who need that to just add the `rd.neednet=1` or even better use it as an opportunity to encourage them to put the config in the ignition config instead since they obviously don't need networking in the initramfs.

Heh came to the same conclusion when I was thinking about this yesterday.

The thing is that without `rd.neednet=1` provided by either us (via `cmdline.d`) or the user (via kargs), the NM generator won't run, so there's no network config to propagate. I guess one thing we could do is automatically inject `rd.neednet=1` if we detect kargs like `ip=`? But I'd strongly prefer pushing them towards NM keyfiles via Ignition.

jlebon:
OK, the fix for https://github.com/coreos/fedora-coreos-config/issues/514 ended up being more involved again because RHCOS has older Ignition. See commit message for full details.

Tested this on both FCOS and RHCOS both with and without an embedded Ignition config.

dustymabe:
> OK, the fix for #514 ended up being more involved again because RHCOS has older Ignition. See commit message for full details.
> 
> Tested this on both FCOS and RHCOS both with and without an embedded Ignition config.

👍

dustymabe:
Still LGTM

bgilbert:
The `coreos.inst` wrapper script [appends `rd.neednet=1`](https://github.com/coreos/coreos-installer/blob/d7108829f25c6e37e7cd67ac8d5990db06df94df/scripts/coreos-installer-service#L86) if it's forwarding any networking kargs to the first boot.  AIUI that's still okay?  We'll need to update [the README](https://github.com/coreos/coreos-installer/blob/d7108829f25c6e37e7cd67ac8d5990db06df94df/README.md#installing-from-pxe) though.

bgilbert (approved):
LGTM

jlebon:
> The coreos.inst wrapper script appends `rd.neednet=1` if it's forwarding any networking kargs to the first boot. AIUI that's still okay?

Yup, that should be fine.

PR #427, opened 2020-05-26
image: inject firstboot kargs (ignition-network-kcmdline)

lucab:
This updates `image` manifest to capture current firstboot kernel
arguments. Those are effectively the same parameters used by GRUB
and overridable by coreos-installer, which so far have been
hardcoded inside `grub.cfg`.

Ref: https://github.com/coreos/coreos-assembler/pull/1373


ashcrow (approved):

jlebon (approved):
LGTM! I like the new key name much better.

PR #428, opened 2020-05-26
liveiso: make motd install hint consistent with docs

jjasghar:
- Edited the /etc/motd to be consistent with the help docs.
REF:
https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/#_installing_from_live_iso

It seems that you don't need the image line if you've booted off the
ISO, which this clears that up.

Signed-off-by: JJ Asghar <jjasghar@gmail.com>

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
@@ -59,9 +59,7 @@ installing persistently to disk. Here is an example of running an install
 to disk via coreos-installer:
 
 curl -O https://example.com/example.ign
-sudo coreos-installer install /dev/sda \\
-         --ignition-file ./example.ign \\
-         --image-url https://example.com/image.raw.xz
+sudo coreos-installer install /dev/sda --ignition-file ./example.ign 
 
 You may configure networking via 'sudo nmcli' or 'sudo nmtui' and have
 that configuration persist into the installed system by passing the


dustymabe:
@jjasghar you are correct, we should change this as you suggested. @jlebon, since this text is shared with RHCOS should we hold off on changing it until we get the osmet (offline install) stuff working there?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
```diff
@@ -60,8 +60,7 @@ to disk via coreos-installer:
 
 curl -O https://example.com/example.ign
 sudo coreos-installer install /dev/sda \\
-         --ignition-file ./example.ign \\
-         --image-url https://example.com/image.raw.xz
+         --ignition-file ./example.ign
```
jlebon: Hmm, let's just make the whole invocation one line now since I think it fits within the limit?

jlebon (approved):
Thanks for the patch! I have one suggestion below. Could you also change the commit title to something like:

```
liveiso: make motd install hint consistent with docs
```

?

> @jlebon, since this text is shared with RHCOS should we hold off on changing it until we get the osmet (offline install) stuff working there?

Yeah, I guess we could. Though I think @cgwalters is working on that now. I'd say let's get this in and if needed, we can tweak it based on `$ID`?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
jjasghar: updating now :)

PR #429, opened 2020-05-27
overlay: handle unbound var in coreos-growpart

dustymabe:
This will make it so we get the desired error message
rather than an error about an unbound variable.

Ref: https://discussion.fedoraproject.org/t/installer-not-creating-xfs-on-sda4/20141

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
@@ -49,10 +49,10 @@ fi
 eval $(blkid -o export "${src}")
 # TODO: Add XFS to https://github.com/systemd/systemd/blob/master/src/partition/growfs.c
 # and use it instead.
-case "${TYPE}" in
+case "${TYPE:-}" in
     xfs) xfs_growfs /sysroot ;;
     ext4) resize2fs "${src}" ;;
-    *) echo "error: Unsupported filesystem for /sysroot: ${TYPE}" 1>&2; exit 1 ;;
+    *) echo "error: Unsupported filesystem for ${path}: '${TYPE:-}'" 1>&2; exit 1 ;;
 esac
 
 touch /var/lib/coreos-growpart.stamp


dustymabe:
untested - will test when I get back to my test set up (next week)

lucab:
Patch looks sane to me.

zonggen:
looks :+1: from my side too

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -49,10 +49,10 @@ fi
 eval $(blkid -o export "${src}")
 # TODO: Add XFS to https://github.com/systemd/systemd/blob/master/src/partition/growfs.c
 # and use it instead.
-case "${TYPE}" in
+case "${TYPE:-}" in
     xfs) xfs_growfs /sysroot ;;
     ext4) resize2fs "${src}" ;;
-    *) echo "error: Unsupported filesystem for /sysroot: ${TYPE}" 1>&2; exit 1 ;;
+    *) echo "error: Unsupported filesystem for /sysroot: ${TYPE:-}" 1>&2; exit 1 ;;
```
jlebon:
```suggestion
    *) echo "error: Unsupported filesystem for $path: '${TYPE:-}'" 1>&2; exit 1 ;;
```

That way it's more obvious if it's somehow empty.

jlebon (commented):
Did you hit this somehow or you just noticed it by inspection?

lucab:
@jlebon this comes from a forum discussion (I noticed it this morning), I added a reference link to the PR text.

cgwalters:
I think this `TYPE` of problem will only happen when something else fails; it seems weird to me though because this service
```
After=ignition-ostree-mount-firstboot-sysroot.service
Requires=ignition-ostree-mount-firstboot-sysroot.service
```

and if that succeeds we should have a valid `/sysroot`.

dustymabe:
> I think this `TYPE` of problem will only happen when something else fails; it seems weird to me though because this service
> 
> ```
> After=ignition-ostree-mount-firstboot-sysroot.service
> Requires=ignition-ostree-mount-firstboot-sysroot.service
> ```
> 
> and if that succeeds we should have a valid `/sysroot`.

agree.. see https://github.com/coreos/fedora-coreos-tracker/issues/500#issuecomment-635075306

cgwalters (approved):

bgilbert:
Included in https://github.com/coreos/fedora-coreos-config/pull/453.

dustymabe:
+1 closing in favor of #453

dustymabe:
actually since @bgilbert included my commit verbatim I can try to address @jlebon's comment in https://github.com/coreos/fedora-coreos-config/pull/429#discussion_r431142673 here and he can pick it up.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
dustymabe: done in latest upload

bgilbert:
Pulled the updated commit into #453.  Closing.

Issue #430, opened 2020-05-27
Installation on qemu with multipath fails


nikita-dubrovskii:
I've been trying to install FCOS on qemu + multipath (using kargs: `coreos.inst=yes multipath=default rd.multipath=on coreos.inst.install_dev=/dev/mapper/mpatha`) and got an error: `no such device`

Despite [https://github.com/coreos/fedora-coreos-config/pull/392/commits/2a37f64a37aed8d4e9568a11b9a43a41ba17473f](url) includes multipathd, without config daemon doesn't create any DM-device.

darkmuggle:
This is caused by needing an updated Dracut in Fedora. We are currently waiting on that. The correct syntax will `rd.multipath=default`.

darkmuggle:
This should be fixed now. The correct syntax is `rd.multipath=defauilt root=/dev/disk/by-label/dm-mpath-root`.

PR #431, opened 2020-05-27
Add "/etc/multipath.conf" to the Initramfs

nikita-dubrovskii:


nikita-dubrovskii:
This PR helps with [https://github.com/coreos/fedora-coreos-config/issues/430](url)

darkmuggle (changes requested):
IMO, we really do not want to have a default `multipath.conf` present. I've commented in #430 but we are waiting an updated Dracut module that supports `rd.multipath=default` that will provide for this feature without enabling it for everyone. 

Per https://github.com/dracutdevs/dracut/blob/master/modules.d/90multipath/multipathd.service#L8-L12 this change would effectively make Multipath on by default since it checks for an `/etc/multipath.conf`  and will run unless there is "off" karg.

There are some good reasons why people may not want it to be on by default.

darkmuggle:
@nikita-dubrovskii when the bits land, would you be willing to help with the testing on s390x? And most importantly, thank you for taking a stab at this problem.

nikita-dubrovskii:
> @nikita-dubrovskii when the bits land, would you be willing to help with the testing on s390x? And most importantly, thank you for taking a stab at this problem.

sure!

PR #433, opened 2020-05-28
Drop multipath-generator

darkmuggle:
The multipath generator was a nice idea to help make things automatic
based on `rd.multipath=default`. However, changes in Ignition Dracut and
the direction of using explict kargs makes this well-intentioned module
cruft.

See: https://github.com/coreos/fedora-coreos-tracker/issues/465


darkmuggle:
Users wishing to use multipath will need to do `rd.mutlipath=default root=/dev/disk/by-label/dm-mpath-root`

cgwalters:
I'm always in favor of less shell script run as root!

Having things be more explicit makes sense to me.  

I am wondering though how mpath will interact with rootfs reprovisioning, but perhaps that code can do "use current root= arg if present, otherwise just /dev/disk/by-label/root".

darkmuggle:
> I am wondering though how mpath will interact with rootfs reprovisioning, but perhaps that code can do "use current root= arg if present, otherwise just /dev/disk/by-label/root".

See: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-device-mapper.rules#L19-L23 

For root reprovisioning, then would have to use `/dev/mapper/mpatha`

jlebon:
A bit late, but LGTM!
I think we can also remove the `sysroot.env` bits too with this.

PR #434, opened 2020-05-29
Revert "ignition-ostree-mount-sysroot"

darkmuggle:
This reverts commit 899ce2eade053bbe508e8cf35602c791f758392f.

Reverting in light of the move to explicit kargs.


darkmuggle:
Proposed to drop in light of:
https://github.com/coreos/ignition/pull/960
https://github.com/coreos/fedora-coreos-tracker/issues/465
https://github.com/coreos/fedora-coreos-config/pull/433

In short, its cruft and should be removed since the direction is moving away from automatic detection of root. And with the complex disk work, this couldn't in the way.

PR #435, opened 2020-05-29
live: remove serial line from isolinux.cfg

### LINKED ISSUE #506: iso: not able to view isolinux menu on serial console
`fedora-coreos-31.20200505.3.0-live.x86_64.iso`

I recently acquired a pcengines x86_64 SoC. It has no graphical console (only serial console). When I tried to boot our ISO (after dd to USB) I start to get the isolinux menu but it gets stuck:

```
SeaBIOS (version rel-1.12.1.3-0-g300e8b70)

Press F10 key now for boot menu

Select boot device:

1. AHCI/0: KINGSTON SUV500MS480G ATA-11 Hard-Disk (447 GiBytes)
2. SD card SD02G 1938MiB
3. USB MSC Drive SanDisk Cruzer Facet 1.26
4. Payload [setup]
5. Payload [memtest]

Booting from Hard Disk...

ISOLINUX 6.04  EHDD Copyright (C) 1994-2015 H. Peter Anvin et a
```

After `Peter Anvin et a` the console just stays there forever.

dustymabe:
When testing out installing FCOS on a new mini PC that only contains
a serial port (no graphical console) I noticed the isolinux menu wasn't
displaying correctly. I compared this with the Fedora Workstation and
Fedora Server configs and neither of those have the `serial 0` line in
them and the isolinux menu shows up fine via serial console there.

Let's remove it since it seems like serial auto-detection works better
than the `0` setting, which is actively causing some consoles to not
work.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/506

### DIFF
--- a/live/isolinux/isolinux.cfg
+++ b/live/isolinux/isolinux.cfg
@@ -1,7 +1,6 @@
 # Note this file mostly matches the isolinux.cfg file from the Fedora 
 # Server DVD iso. Diff this file with that file in the future to pick up
 # changes.
-serial 0
 default vesamenu.c32
 # timeout in units of 1/10s. 50 == 5 seconds
 timeout 50


lucab:
This seems reasonable. Did you try the resulting ISO (both on that machine and on one with a graphical console)?

dustymabe:
> This seems reasonable. Did you try the resulting ISO (both on that machine and on one with a graphical console)?

yes (assuming a VM with graphical console counts 😇)

lucab (approved):

bgilbert (approved):
SGTM.

dustymabe:
> > This seems reasonable. Did you try the resulting ISO (both on that machine and on one with a graphical console)?
> 
> yes (assuming a VM with graphical console counts innocent)

Though I didn't test on a VM with a serial console (which happens to be how I do most of my testing). Using the new ISO images with `virt-install --graphics=none` gives me nothing on the console output and it does look like the Fedora Server DVD has that same behavior. I'm going to need to dig in more to see if there is a way to support all the different cases with a single config. For now let's revert: https://github.com/coreos/fedora-coreos-config/pull/447

PR #442, opened 2020-06-02
liveiso-autologin-generator: showcase --ignition-url

jlebon:
Now that coreos-installer supports it, we can simplify the example
`coreos-installer` command here further.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
@@ -58,8 +58,8 @@ from memory, making it a good candidate for hardware discovery and
 installing persistently to disk. Here is an example of running an install
 to disk via coreos-installer:
 
-curl -O https://example.com/example.ign
-sudo coreos-installer install /dev/sda --ignition-file ./example.ign 
+sudo coreos-installer install /dev/sda \
+    --ignition-url https://example.com/example.ign
 
 You may configure networking via 'sudo nmcli' or 'sudo nmtui' and have
 that configuration persist into the installed system by passing the


jlebon:
Requires: https://github.com/coreos/coreos-installer/pull/126

overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
```diff
@@ -58,8 +58,8 @@ from memory, making it a good candidate for hardware discovery and
 installing persistently to disk. Here is an example of running an install
 to disk via coreos-installer:
 
-curl -O https://example.com/example.ign
-sudo coreos-installer install /dev/sda --ignition-file ./example.ign 
+sudo coreos-installer install /dev/sda \
+    --ignition-url https://example.com/example.ign
```
lucab, line 62: Can we please add `--ignition-hash sha512-<DIGEST>` to the example too?

jlebon: Hmm, that's not really something we expect users to type though, right?

lucab: Possibly not on a live-ISO prompt (i.e. without copy-paste support), you are right.

dustymabe: I do get the sentiment of trying to encourage some validation. We could mention it in the notes below maybe. I'm OK either way.

dustymabe: I guess our previous example didn't show us checking it against a checksum either, so this at least is not a removal of any validation bits.. Probably good as-is.

jlebon: Yeah, maybe. Though there's a lot of things we could mention. I feel like this is covered by `"Please run 'coreos-installer install --help' for more information on the possible
install options."` ?

dustymabe (approved):
LGTM

overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
dustymabe: we were commenting at the same time :)

lucab: Let's keep it as is, and sorry for the noise.

jlebon:
Unheld; this should be good for merge now!

PR #444, opened 2020-06-02
overlay: only run ignition/afterburn checks on first boot

dustymabe:
Only perform checks on the first (Ignition) boot as they are
mostly useful only on that boot. This ensures systems started
before Ignition/Afterburn started logging structured data don't
get misleading messages. Also handles the case where the journal
gets rotated and no longer has the structured log messages.

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
@@ -3,6 +3,12 @@
 [Unit]
 Description=Check if ignition config is provided
 Before=console-login-helper-messages-issuegen.service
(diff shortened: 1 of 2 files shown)


dustymabe:
I hit this problem when doing upgrade testing. I started from a `testing` version before the Ignition 2.3.0 release and upgraded to latest `testing` and then rebased to latest `testing-devel`. Because the system originally ran with Ignition < 2.3.0 it never got the structured journal log entry, so it will forever complain that ignition didn't run, even though it did.

A big hammer way to solve this problem is to just run these checks on first boot.

jlebon (approved):
Seems reasonable to me!

sohankunkerkar (approved):
+1
Yeah, it makes sense to me.

dustymabe:
ok cool. will merge on green CI

PR #445, opened 2020-06-02
[testing] tree: promote changes from testing-devel at 6bb676b

dustymabe:


jlebon (approved):
> manifest: switch to f32 for releasever

Hmm right... I think we can re-organize things so that this lives in an included manifest. Then e.g. `rawhide` and `next-devel` (after branching) could override it in its `manifest.yaml`.

PR #446, opened 2020-06-02
[next] tree: promote changes from testing-devel at 6bb676b

dustymabe:


dustymabe:
Note that we are promoting from `testing-devel` and not `next-devel` because `next` and `testing` are just shadows of each other for now since we are promoting to f32.

jlebon (approved):

PR #454, opened 2020-06-04
20platform-chrony: make generator executable

rfairley:
Add missing executable permissions found in latest FCOS release:
coreos/fedora-coreos-tracker#517 (comment)

This has bit us a few times in the past - I thought I had
double-checked this at some time during review but maybe hadn't
clicked through to "View file". The GitHub PR "Files changed" page
alone doesn't indicate whether the file is executable. Will be more
thorough in future.


cgwalters (approved):

cgwalters:
:cry: This is the problem with manual testing.  Need to get some !qemu coverage here.

PR #457, opened 2020-06-05
coreos-copy-firstboot-network.service: Squash race with GPT generator

### LINKED ISSUE #523: races in boot.mount
https://github.com/coreos/fedora-coreos-config/pull/346#issuecomment-639779100

We have multiple things mounting `/boot` and we're not cleanly waiting for the device.  https://github.com/coreos/fedora-coreos-config/pull/420 tried to fix this but was withdrawn; we will need some solution.

cgwalters:
I believe this is the core issue; on one failing run I see the
services being interleaved.  Remove the "retry mounting" hack
to prove it.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/523

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
@@ -6,6 +6,7 @@
 #
 # - Need to look for networking configuration on the /boot partition
 #     - i.e. after /dev/disk/by-label/boot is available
+#     - and after the ignition-dracut GPT generator (see below)
 # - Need to run before networking is brought up.
 #     - This is done in nm-run.sh [1] that runs as part of dracut-initqueue [2]
 #     - i.e. Before=dracut-initqueue.service
@@ -32,9 +33,12 @@ DefaultDependencies=false
 Before=ignition-diskful.target
 Before=dracut-initqueue.service
 After=dracut-cmdline.service
+# Any services looking at mounts need to order after this
+# because it causes device re-probing.
+After=coreos-gpt-setup@dev-disk-by\x2dlabel-root.service
 # Since we are mounting /boot/, require the device first
 Requires=dev-disk-by\x2dlabel-boot.device
-After=dev-disk-by\x2dlabel-boot.device   
+After=dev-disk-by\x2dlabel-boot.device
 
 [Service]
 Type=oneshot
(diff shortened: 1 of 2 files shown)


dustymabe:
> Remove the "retry mounting" hack
> to prove it.

i.e., we're not sure if this will surely fix the race, but removing the retry will cause issues to surface faster?

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
```diff
@@ -32,9 +32,11 @@ DefaultDependencies=false
 Before=ignition-diskful.target
 Before=dracut-initqueue.service
 After=dracut-cmdline.service
+# Any services looking at mounts need to order after this
+After=coreos-gpt-setup@dev-disk-by\x2dlabel-root.service
```
dustymabe, line 38: line 7-8 mention the deps needed for mounting boot, maybe we should mention this there?

dustymabe (approved):
Seems fine. I'm hoping you're right and we found the race condition that's been haunting us a bit.

dustymabe:
cc @darkmuggle @jlebon

cgwalters:
> i.e., we're not sure if this will surely fix the race, but removing the retry will cause issues to surface faster?

I'm pretty confident it's the fix.  I didn't exhaustively but looking at your PR history originally you just couldn't land it until adding the retry, now that CI is passing without retries is a strong signal.

jlebon (approved):
Yeah makes sense to me. The `coreos-gpt-setup` service itself has:

```
# Run before services that use device nodes, preventing them from racing
# with udev activity generated by sgdisk
Before=ignition-setup-base.service ignition-setup-user.service ignition-disks.service
```

Which... I don't see why `coreos-copy-firstboot-network.service` would be any different.

jlebon:
> I'm pretty confident it's the fix. I didn't exhaustively but looking at your PR history originally you just couldn't land it until adding the retry, now that CI is passing without retries is a strong signal.

One easy way to check is to drop the `After=` hunk here but keep the mount simplification hunk and see if CI fails. But I agree this is likely the race.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
cgwalters: Sure updated and re-pushed which will also re-trigger CI.

dustymabe:
The fix for this went into testing stream release `32.20200615.2.0`.  Please try out the new release and report issues.

PR #458, opened 2020-06-09
image: drop hardcoded firstboot network kargs

lucab:
This removes the hardcoded `ip=dhcp,dhcp6` from bootloader configuration.
The default value has been moved to Afterburn (>= 4.4.0), making
this dynamically injected at runtime. On some platforms (e.g. VMware),
this also allows user-controlled overriding.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/460

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -12,7 +12,6 @@ extra-kargs:
 # Kernel arguments to be used on first-boot.
 ignition-network-kcmdline:
     - 'rd.neednet=1'
-    - 'ip=dhcp,dhcp6'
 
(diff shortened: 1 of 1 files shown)


dustymabe:
are we not dropping the `rd.neednet=1` too?

lucab:
@dustymabe soon! We'll need (at least) https://github.com/coreos/ignition/pull/956 merged and packaged. The two concerns are decoupled though, so we can land the pieces at different times.

dustymabe (approved):
LGTM

jlebon (approved):
:tada:

PR #459, opened 2020-06-09
Various overlay fixes

travier:
Collection of overlay fixes to simplify ignition specv3 support in RHCOS and futur LUKS support in FCOS.


overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
```diff
@@ -35,7 +35,7 @@ Before=dracut-initqueue.service
 After=dracut-cmdline.service
 # Any services looking at mounts need to order after this
 # because it causes device re-probing.
-After=coreos-gpt-setup@dev-disk-by\x2dlabel-root.service
+After=coreos-gpt-setup.service
```
cgwalters, line 38: Let's keep the old ordering too, otherwise we have a "version interlock" problem where new configs with older ignition-dracut will re-introduce the race.

And having a new config but older ignition-dracut is exactly what is happening in CI for this PR, and will happen for everyone buliding FCOS locally until we get it through Bodhi etc.

(systemd is fine with having `After=some-nonexistent-unit.service`)

cgwalters (changes requested):
Overall LGTM, though I would make "cleanup" PRs separate from functional changes in general.  But OK as is too.

travier:
Split into https://github.com/coreos/fedora-coreos-config/pull/460 https://github.com/coreos/fedora-coreos-config/pull/461 https://github.com/coreos/fedora-coreos-config/pull/462

PR #461, opened 2020-06-09
overlay.d: Merge ignition-conf module with 05core

travier:
RHCOS is moving to ignition specv3 thus the split is no longer needed.

Also updates the README to match current overlay content.


overlay.d/README.md
```diff
@@ -25,9 +25,15 @@ Disables the Red Hat Linux legacy `ifcfg` format.
 15fcos
 ------
 
-Things that are more closely "Fedora CoreOS"; branding, services.
+Things that are more closely "Fedora CoreOS":
+
+* disable password logins by default over SSH
+* branding (MOTD)
+* enable services by default (fedora-coreos-pinger)
+* display warnings on the console if no ignition config was provided or no ssh
```
dustymabe: should we make these warnings go for RHCOS too?

dustymabe (approved):
Seems OK to me

overlay.d/README.md
jlebon: Yeah, I think we should, though let's do that in a separate PR?

jlebon (approved):

cgwalters:
One hesitation I have here is that as soon as we merge this, it means if we need to backport any other changes from this repo we'll have to do it "manually" by e.g. copying rather than just updating the submodule.

I lean a bit towards waiting until we've fully validated RHCOS+Ignition spec3 but I'm also OK merging now if someone else has a stronger preference.

dustymabe:
> I lean a bit towards waiting until we've fully validated RHCOS+Ignition spec3

+1 - how close are we on being able to commit to that?

travier:
I agree, this should be merged only when we are ready to commit to specv3 in RHCOS.

travier:
This is the second PR to merge in order to move to spec3 for RHCOS.

PR #462, opened 2020-06-09
coreos-copy-firstboot-network.service: Update GPT generator unit name

travier:
The GPT setup unit was simplified to reflect the fact that it will only
be called once for the disk with the partition labeled 'boot'.

This keeps the old unit name for now for compatibility.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-firstboot-network/coreos-copy-firstboot-network.service
@@ -35,6 +35,10 @@ Before=dracut-initqueue.service
 After=dracut-cmdline.service
 # Any services looking at mounts need to order after this
 # because it causes device re-probing.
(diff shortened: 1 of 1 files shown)


travier:
Resolves https://github.com/coreos/fedora-coreos-config/pull/459#discussion_r437431010

cgwalters (approved):

dustymabe (approved):
LGTM - optionally would be nice to add some words that explain when we can drop the compat `After=`.

dustymabe:
~~Note this requires https://github.com/coreos/ignition-dracut/pull/191~~ - actually it doesn't because of the compat code. So I'll convert this into an FYI - the generator was converted to a non instance unit in https://github.com/coreos/ignition-dracut/pull/191

travier:
Note added. Thanks

dustymabe:
> Note added. Thanks

Looks great!

jlebon (approved):

jlebon:
CI on this repo is broken right now because of the Koji migration. If these PRs can wait until it's back on, then let's do that.

PR #463, opened 2020-06-09
overlay/15fcos: coreos-check-ssh-keys: add missing newline to output

dustymabe:
Currently we're missing a newline when printing information
about the SSH keys on the console:

```
Ignition: user provided config was applied
Ignition: wrote ssh authorized keys file for user: coreAfterburn: wrote ssh authorized keys file for user: core
```

Let's add a newline if needed.

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
@@ -28,6 +28,10 @@ main() {
         output+="$ignitionusers"
     fi
     if [ -n "$afterburnusers" ]; then
+        # add newline if needed
+        if [ -n "$output" ]; then
+            output+=$'\n'
+        fi
         output+="$afterburnusers"
     fi
 


overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -28,6 +28,10 @@ main() {
         output+="$ignitionusers"
     fi
     if [ -n "$afterburnusers" ]; then
+        # add linebreak if needed
+        if [ -n "$output" ]; then
+            output+=$'\n'
```
cgwalters, line 33: I don't understand this bash syntax, what does `$` mean before a literal string?

dustymabe: It's a bash literal. Look for `Words of the form $'string' are treated specially` in [the man page](http://linuxcommand.org/lc3_man_pages/bash1.html). In this case, it's just a newline.

dustymabe: that also reminds me that I should probably `s/linebreak/newline/` in the comment

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
```diff
@@ -28,6 +28,10 @@ main() {
         output+="$ignitionusers"
     fi
     if [ -n "$afterburnusers" ]; then
+        # add newline if needed
+        if [ -n "$output" ]; then
+            output+=$'\n'
```
jlebon, line 33: Another less verbose approach is to just pipe the `journalctl | jq | xargs` pipelines directly into the issue file and then below just use `if [ ! -s "$issue" ]; then echo -e "${warn}..."; fi`. This is fine too though!

jlebon (approved):

PR #465, opened 2020-06-10
tests: Add gshadow test, migrated from coreos-assembler

cgwalters:
See https://github.com/coreos/coreos-assembler/pull/1522

The idea is our default for new external tests would be this repo.


jlebon (approved):
Restarted CI on this since Koji should be back up now.
LGTM! Will leave open for a bit though in case anyone objects to this direction.

cgwalters:
OK, sounds like no objections!  Would like to keep up momentum on testing stuff so merging.

PR #467, opened 2020-06-11
[testing] overrides: fast-track fedora-coreos-pinger-0.0.4-4.fc32

dustymabe:
(cherry picked from commit 8eceea9a79be6d8e89776fc29c36bfa79ab109ea)

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -70,7 +70,7 @@
       "evra": "1.0.8-2.fc32.x86_64"
     },
     "c-ares": {
-      "evra": "1.16.0-1.module_f32+8723+aeaf79ed.x86_64"
+      "evra": "1.15.0-5.fc32.x86_64"
     },
     "ca-certificates": {
       "evra": "2020.2.40-3.fc32.noarch"
@@ -235,7 +235,7 @@
       "evra": "2.2.8-2.fc32.x86_64"
     },
     "fedora-coreos-pinger": {
-      "evra": "0.0.4-1.module_f32+6507+a4e4adf6.x86_64"
+      "evra": "0.0.4-3.fc32.x86_64"
     },
     "fedora-gpg-keys": {
       "evra": "32-2.noarch"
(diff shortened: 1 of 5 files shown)


dustymabe:
ok this should be ready now.

jlebon (approved):

dustymabe:
CI is failing because robosig has been stuck signing an llvm build for 10 hours - infra team is looking at it now.

manifest-lock.x86_64.json
```diff
@@ -70,7 +70,7 @@
       "evra": "1.0.8-2.fc32.x86_64"
     },
     "c-ares": {
-      "evra": "1.16.0-1.module_f32+8723+aeaf79ed.x86_64"
+      "evra": "1.15.0-5.fc32.x86_64"
```
rfairley, line 73: Hmm, this downgrade is probably not expected. (was apparent already in https://github.com/coreos/fedora-coreos-tracker/issues/525#issuecomment-640154198 - I shouldn't have left it until here to notice)

I see in the package builds (https://koji.fedoraproject.org/koji/packageinfo?packageID=1434), we're sourcing the non-modular F32 version from a time when F32 was rawhide (https://koji.fedoraproject.org/koji/buildinfo?buildID=1434406), i.e. the original RPM build which the branched modular builds are based on. Without modular content we wouldn't be getting an updated 1.16 version until `f33` becomes the new `testing`. I think this would be okay to live with for now, and there is a long term plan to fully support modularity in rpm-ostree (https://github.com/coreos/rpm-ostree/issues/1435). Just something to be aware of (if others weren't aware already).

rfairley (approved):

manifest-lock.x86_64.json
dustymabe: Yeah I knew about this (see the diff in https://github.com/coreos/fedora-coreos-tracker/issues/525#issuecomment-640154198). I'm not sure about why it happened (obviously you looked closer at it and explained it more in your comment above), but it's pretty clear it shouldn't have been in the compose to begin with so I was fine with it going backwards.

dustymabe: it's better to fix it now in testing since it never hit stable.

rfairley: :+1: understood now - fixing it early in testing before the modular version gets to stable makes sense.

PR #468, opened 2020-06-11
overrides: fast-track fedora-coreos-pinger-0.0.4-4.fc32

rfairley:
Fast-track a new build of fedora-coreos-pinger which doesn't use
modules and now has an associated Bodhi update.


dustymabe (approved):
LGTM

PR #476, opened 2020-06-16
WIP: add AWS EBS NVMe udev rules

bh7cw:
Add the udev rules for AWS EBS NVMe.

### DIFF
--- a/overlay.d/05core/usr/lib/udev/bin/aws-ebs-nvme-id
+++ b/overlay.d/05core/usr/lib/udev/bin/aws-ebs-nvme-id
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2020 The CoreOS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+usage() {
+    echo "Usage: $0 [-d|-p|-h] <device>" 1>&2
+}
+
+# Get partition ID from device name, e.g. '/dev/nvme0n1p2' -> '1'
+namespace_id() {
+    NSID=$(echo -n "$1" | cut -f 3 -d 'n' | cut -f 1 -d 'p')
+    echo "_NS_ID=${NSID}"
+}
+
+# Get device name from raw metadata,
+devname() {
+    RAWVOL=$(nvme id-ctrl --raw-binary "$1" | cut -c3073-3104 | tr -s ' ' | sed 's/ $//g')
+    VOL="${RAWVOL#/dev/}"
+    if [[ -n "$VOL" ]]; then
+        echo "${VOL}"
+    else
+        exit 1
+    fi
+}
+
+while getopts "hd:n:" o; do
+    case "${o}" in
+        d)
+            devname "${OPTARG}"
+            exit 0
+            ;;
+        n)
+            namespace_id "${OPTARG}"
+            exit 0
+            ;;
+        h)
+            usage
+            exit 0
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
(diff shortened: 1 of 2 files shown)


overlay.d/05core/usr/lib/udev/bin/aws-ebs-nvme-id
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2018 The CoreOS Authors. All rights reserved.
```
jlebon: 2020? :)

overlay.d/05core/usr/lib/udev/bin/aws-ebs-nvme-id
```diff
@@ -0,0 +1,47 @@
...
+    RAWVOL=$(nvme id-ctrl --raw-binary "$1" | cut -c3073-3104 | tr -s ' ' | sed 's/ $//g')
+    VOL="${RAWVOL#/dev/}"
+    if [[ -n "$VOL" ]]; then
+        echo "${VOL}"
+    else
+        exit 1
+    fi
+}
+
+while getopts "hd:n:" o; do
```
jlebon, line 28: These options are pretty cryptic. Can we use long opts instead?

jlebon (commented):
Thanks for the contribution! Can you squash all your commits into a single one and add a commit message there on why we want these rules? Also, was this script taken from somewhere? If so, could you add a link to it in the commit message as well?

dustymabe:
In https://github.com/coreos/fedora-coreos-tracker/issues/104#issuecomment-456201265 we had a suggestion from the community to maybe try to get these udev rules into systemd. So I [started a discussion upstream](https://github.com/coreos/fedora-coreos-tracker/issues/104#issuecomment-456515167) about that. Upstream systemd seems willing to accept rules on a case by case basis. Should we try to see if they'll accept rules upstream first?

overlay.d/05core/usr/lib/udev/bin/aws-ebs-nvme-id
bh7cw: I am working on as part of https://github.com/coreos/fedora-coreos-tracker/issues/104#issuecomment-456515167
The script is from Container Linux, I changed it and makes it work on fedora.
https://github.com/coreos/init/pull/268/files

overlay.d/05core/usr/lib/udev/bin/aws-ebs-nvme-id
```diff
@@ -0,0 +1,47 @@
...
+
+# Get partition ID from device name, e.g. '/dev/nvme0n1p2' -> '1'
+namespace_id() {
+    NSID=$(echo -n "$1" | cut -f 3 -d 'n' | cut -f 1 -d 'p')
+    echo "_NS_ID=${NSID}"
+}
+
+# Get device name from raw metadata,
+devname() {
+    RAWVOL=$(nvme id-ctrl --raw-binary "$1" | cut -c3073-3104 | tr -s ' ' | sed 's/ $//g')
```
velothump, line 19: The `nvme` binary isn't available on FCOS.

dustymabe: The `nvme` binary was added in https://github.com/coreos/fedora-coreos-config/pull/564 and should be in the next `testing` release.

bh7cw: Thanks for reply. @dustymabe :smiley_cat: Shall I close this PR now that we will move to systemd?

dustymabe: We can close this PR and re-open if the efforts to get this upstream don't work for some reason. 👍

bh7cw: Great, thanks. :+1:

PR #478, opened 2020-06-16
overlay.d: Move sshd_config.d configs to 15fcos

travier:
Keep all sshd_config.d config snippets in 15fcos (FCOS specific) as
OpenSSH version in RHCOS is not capable of handling them yet.

In this case, RHCOS also does not want to read keys added by afterburn.

This PR also include a README update extracted from https://github.com/coreos/fedora-coreos-config/pull/461 that is not merge ready.


bgilbert:
For what it's worth, older OpenSSH will just ignore the fragment.

jamescassell (approved):

cgwalters:
> For what it's worth, older OpenSSH will just ignore the fragment.

Ah, I see.  It is still a bit deceptive to ship it and have it not take effect though, so probably better to remove it for now to avoid confusion?

lucab:
I agree it's better to not forward the fragment at all to RCHOS (where it is unsupported), because empirically this exact point already caused confusion when discussing this topic internally among us yesterday.

bgilbert (approved):

travier:
How do I retry the CI checks here?

bgilbert:
@travier If you click the Details link and log into Jenkins, there's a rerun button in the top bar.

lucab:
I rerun this manually, I think @travier is missing permissions to access CI.

PR #479, opened 2020-06-16
kola-denylist: rename from kola-blacklist

bgilbert:
For https://github.com/coreos/coreos-assembler/pull/1530.

Is there any other ratcheting that needs to be done before this lands?


cgwalters (approved):
I can't think of any reason not to land this.

dustymabe (approved):
LGTM

PR #482, opened 2020-06-17
[WIP]podman rc7

baude:
ensure we pass tests only

Signed-off-by: Brent Baude <bbaude@redhat.com>


lucab:
Please note that we had to [disable](https://github.com/coreos/fedora-coreos-config/pull/425) most `podman` tests due to registry rate-limiting.

jlebon:
@baude WDYT about collaborating some time next week or so to set up CoreOS CI on containers/libpod. Then you won't have to do this each time. :)

cgwalters:
This looks obsolete.

PR #485, opened 2020-06-19
40ignition-ostree: relabel /root after copying skel files

jlebon:
Otherwise they can end up unlabeled. We don't hit this right now because
Ignition is just doing a blind relabeling on all of `/var/roothome`
whenever any user is configured (which is always because we have a base
config which configures the `core` user). We shouldn't rely on this
though:

https://github.com/coreos/ignition/pull/996

And more generally, each piece of software should be responsible for
relabeling the bits that it touched.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-populate-var.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-populate-var.sh
@@ -35,9 +35,11 @@ for varsubdir in lib log home roothome opt srv usrlocal mnt media; do
         systemd-tmpfiles --create --boot --root=/sysroot --prefix="/var/${varsubdir}"
     fi
 
+    if [[ $varsubdir == roothome ]]; then
+        # TODO move this to tmpfiles.d once systemd-tmpfiles handles C! with --root correctly.
+        # See https://github.com/coreos/fedora-coreos-config/pull/137
+        cp /sysroot/etc/skel/.bash* /sysroot/var/${varsubdir}
+    fi
+
     coreos-relabel "/var/${varsubdir}"
 done
-
-# TODO move this to tmpfiles.d once systemd-tmpfiles handles C! with --root correctly.
-# See https://github.com/coreos/fedora-coreos-config/pull/137
-cp /sysroot/etc/skel/.bash* /sysroot/root


jlebon:
> Otherwise they can end up unlabeled. We don't hit this right now because
Ignition is just doing a blind relabeling on all of /var/roothome
whenever any user is configured (which is always because we have a base
config which configures the core user).

One can make this happen by booting with an Ignition config which doesn't set any users, and passing `rd.break=cmdline` to delete the base config. Then:

```
[root@localhost ~]# ls -laZ
total 20
drwx------.  3 root root system_u:object_r:admin_home_t:s0   92 Jun 19 19:32 .
drwxr-xr-x. 23 root root system_u:object_r:var_t:s0        4096 Jun 19 19:32 ..
-rw-r--r--.  1 root root system_u:object_r:unlabeled_t:s0    18 Jun 19 19:32 .bash_logout
-rw-r--r--.  1 root root system_u:object_r:unlabeled_t:s0   141 Jun 19 19:32 .bash_profile
-rw-r--r--.  1 root root system_u:object_r:unlabeled_t:s0   376 Jun 19 19:32 .bashrc
drwxr-xr-x.  2 root root system_u:object_r:ssh_home_t:s0     29 Jun 19 19:32 .ssh
-rw-r--r--.  1 root root system_u:object_r:admin_home_t:s0 3531 Jun 19 19:32 foobar.txt
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-populate-var.sh
```diff
@@ -35,9 +35,11 @@ for varsubdir in lib log home roothome opt srv usrlocal mnt media; do
         systemd-tmpfiles --create --boot --root=/sysroot --prefix="/var/${varsubdir}"
     fi
 
+    if [[ $varsubdir == roothome ]]; then
+        # TODO move this to tmpfiles.d once systemd-tmpfiles handles C! with --root correctly.
+        # See https://github.com/coreos/fedora-coreos-config/pull/137
+        cp /sysroot/etc/skel/.bash* /sysroot/root
```
bgilbert: `/sysroot/var/${varsubdir}` would be clearer than `/sysroot/root`.

bgilbert (approved):

PR #486, opened 2020-06-19
tests/kola/basic: check for unlabeled files in /var and /etc

jlebon:
Inspired by https://github.com/coreos/fedora-coreos-config/pull/485.

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -41,3 +41,12 @@ elif [[ $nm_ts -gt $switchroot_ts ]] && on_platform aws; then
     fatal "NetworkManager not started in initramfs!"
 fi
 ok conditional initrd networking
+
+# check that no files are unlabeled
+unlabeled=$(find /var /etc -context '*:unlabeled_t:*')
+if [ -n "${unlabeled}" ]; then
+    echo "Found unlabeled files:"
+    echo "${unlabeled}"
+    exit 1
+fi
+ok no files with unlabeled_t SELinux label


jlebon:
This actually uncovered an unlabeled directory: https://github.com/coreos/ignition/pull/1006. So it'll probably break CI until we respin Ignition with that fix.

jlebon:
(That said, there doesn't seem to be any repercussions from that directory being unlabeled. Which makes sense, otherwise we would've noticed this much faster.)

dustymabe (approved):
LGTM

tests/kola/basic
```diff
@@ -1,2 +1,14 @@
 #!/bin/bash
-exec systemctl is-enabled logrotate.service
+set -xeuo pipefail
+
+# check that logrotate is enabled
+systemctl is-enabled logrotate.service
+
+# check that no files are unlabeled
+for dir in /var /etc; do
+    find ${dir} -context '*:unlabeled_t:*' |& tee out.txt
```
bgilbert: Can't we just `find /var /etc -context ...`?

jlebon: Thanks, that's clearly better! :)

bgilbert:
Test failure:

```
tee: out.txt: Operation not permitted
```

bgilbert (approved):

cgwalters:
I took the opportunity to rebase :surfing_man: this, LGTM

PR #488, opened 2020-06-22
overrides: drop crypto-policies pin

### LINKED ISSUE #540: crypto-policies pinned to avoid Lua script
crypto-policies &ge; 20200527-3.gitb234a47 has a Lua script for FIPS mode, which causes a failure bumping the lockfile:

```
error: Package 'crypto-policies' has (currently) unsupported <lua> script in '%post'
```

The previous crypto-policies is pinned for now.

xref https://github.com/coreos/rpm-ostree/issues/749

bgilbert:
No longer needed with rpm-ostree 2020.3 in the COSA container.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/540.


bgilbert:
Tested locally by dropping locks for `crypto-policies` and `openssh-*`.

jlebon (approved):

PR #490, opened 2020-06-24
overlay/fcos: pin legacy ifnames on existing machines

lucab:
shipping such unit in the OS.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/484

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
+++ b/overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
@@ -1,0 +1,34 @@
+#!/bin/bash
+# This is a pinning script to keep legacy ifnames on machines that were
+# installed on an OS version where `99-default.link` was missing due to
+# https://github.com/coreos/fedora-coreos-tracker/issues/484
+
+set -euo pipefail
+
+injected=0
+
+if test -e /usr/lib/systemd/network/99-default.link; then
+    echo "unit /usr/lib/systemd/network/99-default.link found, no actions performed"
+    exit 0;
+fi
+
+for f in /boot/loader/entries/*.conf; do
+    options=$(grep '^options ' "$f" | cut -f2- -d' ')
+
+    # If it is already specified, do not touch whatever value is there.
+    # This is in order to avoid messing with user customization, and to
+    # make the logic idempotent.
+    if grep -q "net.ifnames" <<< "$options"; then
+        continue
+    fi
+
+    # Otherwise, make sure we stay on legacy ifnames.
+    sed -e "/^options / s/$/ net.ifnames=0/" -i "$f"
+    echo "$(basename "$f"): injected net.ifnames=0"
+    injected=1
+done
+
+if [[ $injected -eq 0 ]]; then
+    echo "existing net.ifnames karg detected, no actions performed"
+fi
+
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -3,3 +3,6 @@ enable fedora-coreos-pinger.service
 # Provide information if no ignition is provided
 enable coreos-check-ignition-config.service
 enable coreos-check-ssh-keys.service
+# Pinning script for legacy interface names, see
+# https://github.com/coreos/fedora-coreos-tracker/issues/484
+enable coreos-keep-legacy-ifnames.service
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-keep-legacy-ifnames.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-keep-legacy-ifnames.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=CoreOS Keep Legacy Network Interface Names
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/484
+RequiresMountsFor=/boot
+ConditionPathExists=!/usr/lib/systemd/network/99-default.link
+# NOTE: we do not conditionalize on ConditionKernelCommandLine= here because
+# kernel args might have been tweaked manually just for this boot; we still
+# want to change the BLS entries in that case.
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/coreos-keep-legacy-ifnames
+
+[Install]
+WantedBy=multi-user.target


overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
```diff
@@ -0,0 +1,42 @@
...
+
+    # If it is already specified, do not touch whatever value is there.
+    # This is in order to avoid messing with user customization, and to
+    # makes the logic idempotent.
+    if grep -q "net.ifnames" <<< "$options"; then
+        continue
+    fi
+
+    # Otherwise, make sure we stay on legacy ifnames.
+    sed -e "/^options / s/$/ net.ifnames=0/" -i "$f"
```
lucab, line 26: @jlebon please double-check the my knowledge here: if injected this way, the `net.ifnames=0` will be retained even if the next update payload does not have it in its embedded kargs. Correct?

overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
```diff
@@ -0,0 +1,42 @@
...
+    echo "$(basename "$f"): injected net.ifnames=0"
+    injected=1
+done
+
+if [[ $injected -eq 1 ]]; then
+    if test -e /etc/systemd/network/99-default.link; then
```
cgwalters: Note `test -e` will exit `1` for broken symlinks; in that case the `ln -s` below would fail.  I mention this because it's not uncommon for systemd units in e.g .`wants.d` directories to be broken symlinks because systemd doesn't care, and so I have learned to more look for "existence" here.

overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
cgwalters: The update payload doesn't have kargs.   Supporting that is https://github.com/ostreedev/ostree/issues/479

So yes it will be retained because there's nothing to override it.

overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
lucab: Ah, manpage explicitly says so but I failed to appreciate the implications of that. I've augmented that with a `-L || -e` in a fixup commit.

overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
```diff
@@ -0,0 +1,43 @@
...
+
+if [[ $injected -eq 1 ]]; then
+    if test -L /etc/systemd/network/99-default.link ||
+       test -e /etc/systemd/network/99-default.link; then
+        echo "unit /etc/systemd/network/99-default.link already exists, not touching it"
+        exit 0;
+    fi
+
+    ln -s /dev/null /etc/systemd/network/99-default.link
+    echo "null-linked /etc/systemd/network/99-default.link"
```
dustymabe: Creating a `/etc/systemd/network/99-default.link` file is a bit confusing to me because it doesn't have any effect IIUC because we are using the karg to direct the behavior we want. It is convenient to have the file, though, so we can "detect" if we've already run. Maybe we could use a different file to "detect" if we had already run if we did want to keep that functionality.

Do we think keeping the /dev/null symlink is confusing? Is it worth keeping it around?

jlebon: Yeah, I'm not even sure if we need a stamp file for determining whether we already ran. The script is already idempotent by the fact that it checks the BLS config for `net.ifnames`.

lucab: Ack, let me drop all the `/etc` fiddling in here.

overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
```diff
@@ -0,0 +1,34 @@
...
+    fi
+
+    # Otherwise, make sure we stay on legacy ifnames.
+    sed -e "/^options / s/$/ net.ifnames=0/" -i "$f"
+    echo "$(basename "$f"): injected net.ifnames=0"
+    injected=1
+done
+
+if [[ $injected -eq 0 ]]; then
+    echo "existing ifnames kargs detected, no actions performed"
```
jlebon:
```suggestion
    echo "existing net.ifnames karg detected, no actions performed"
```

jlebon (approved):
Cool, thanks! One optional tweak, but LGTM as is too.

lucab:
Rebased. The PR to drop this logic and get back to systemd default persistent names is at https://github.com/coreos/fedora-coreos-config/pull/491.

overlay.d/15fcos/usr/libexec/coreos-keep-legacy-ifnames
```diff
@@ -0,0 +1,34 @@
...
+    echo "unit /usr/lib/systemd/network/99-default.link found, no actions performed"
+    exit 0;
+fi
+
+for f in /boot/loader/entries/*.conf; do
+    options=$(grep '^options ' "$f" | cut -f2- -d' ')
+
+    # If it is already specified, do not touch whatever value is there.
+    # This is in order to avoid messing with user customization, and to
+    # makes the logic idempotent.
```
dustymabe:
```suggestion
    # make the logic idempotent.
```

dustymabe (approved):
LGTM

cgwalters:
Writing an ext test for this probably wouldn't be too hard (not a blocker though)

lucab:
@cgwalters do you mean literally on this logic, or are you maybe thinking of https://github.com/coreos/fedora-coreos-config/pull/491? I was planning to have it on the latter, as this one requires a reboot to be really tested and the old behavior will be gone from CI in a couple of releases.

lucab:
Rebased to fix the typos too.

dustymabe:
will merge this in a bit unless anyone has anything else

cgwalters:
FTR, this service fails on the live ISO because `/boot` is read-only:
```
Jul 01 11:59:18 localhost systemd[1]: Starting CoreOS Keep Legacy Network Interface Names...
Jul 01 11:59:18 localhost coreos-keep-legacy-ifnames[956]: sed: couldn't open temporary file /boot/loader/entries/sedTIEkyv: Read-only file system
Jul 01 11:59:18 localhost systemd[1]: coreos-keep-legacy-ifnames.service: Main process exited, code=exited, status=4/NOPERMISSION
Jul 01 11:59:18 localhost systemd[1]: coreos-keep-legacy-ifnames.service: Failed with result 'exit-code'.
Jul 01 11:59:18 localhost systemd[1]: Failed to start CoreOS Keep Legacy Network Interface Names.
```

jlebon:
Ahh of course. This should've had something like `ConditionPathExists=!/run/ostree-live`.

We should add some basic test of the live path as part of `fcosKola` since it's so radically different in setup from qemu. I guess https://github.com/coreos/coreos-assembler/pull/1571 will make that easier too.

tmartin-git:
(https://github.com/dustymabe/fedora-coreos-config/commit/aaf68674e5bc46ce67621519388b82b8dd25319b#commitcomment-40293518)

(Just a quit note about this - someone may encouter the same issue I had)

I noticed that all kernel configurations are impacted by this change and as a consequence it actually kind of "corrupts" your rollback.

As an example, I had setup a /etc/systemd/network/99-default.link configuration file (not /usr/lib/systemd/network/99-default.link as expected) to avoid conflict with package systemd-dev.
When the update took place to FCOS v>=32, it also added net.ifnames="0" to my previous kernel-attached configuration deployement.

Then I did a rollback (to my previously working FCOS 31) and unfortunatly, my system was broken (i.e. I had legacy names on my network interfaces).

cgwalters:
> As an example, I had setup a /etc/systemd/network/99-default.link configuration file

In other words, you'd manually enabled the new names, and our change reverted it?

Yes...we clearly should have been checking for that.  I suspect a relatively small number of people have done this, and those who have can drop this override and use the new names going forward.

If we get more reports of this we may need to try to adjust the code to detect this case.  Sorry about that, and thank you for following up here!

tmartin-git:
I was more concerned about the changes applied to **all** bootable kernel cmdline parameters. (by the way don't really know if this can be done a different way)
As a result, a rollback did not work - and as far as I understand, rpm-ostree rollback should always work, shouldn't it ?

PR #491, opened 2020-06-24
manifests: stop removing systemd-udev link units 

### LINKED ISSUE #484: network interface name differs between Fedora CoreOS and RedHat CoreOS
I've noticed that Fedora coreos use the old eth* standard while RedHat CoreOS use the net.ifnames=1 naming.
I.E. on vmware FedoraCOS expose eth0 while RHCOS expose ens192 .

lucab:
This reverts parts of #129 in order to bring back systemd default configuration for interface
names.
It pulls into the OS a single link unit:

```
$ ls /usr/lib/systemd/network/
99-default.link
```

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/484


lucab:
I've manually checked this on a local testing-devel build, verification steps below.

Checking the legacy-pinning logic on upgrade:
```
$ ip addr | grep BROADCAST
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000

sudo ostree remote add --no-gpg-verify --force fedora https://ostree.fedoraproject.org
sudo tar -xf /var/local/build/fedora-coreos-32.20200625.dev.1-ostree.x86_64.tar -C /var/srv
sudo ostree --repo=/sysroot/ostree/repo pull-local /var/srv fedora/x86_64/coreos/testing-devel
sudo rpm-ostree rebase -r <NEWCOMMIT>

$ ip addr | grep BROADCAST
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000

```

Checking the default systemd behavior:
```
$ ip addr | grep BROADCAST
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000

sudo rpm-ostree kargs --reboot --delete 'net.ifnames=0'

$ ip addr | grep BROADCAST
2: ens9: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
```

jlebon (approved):
Nice, thanks for testing this! One test we should also add is in the upgrade test, we verify that whatever interface name we had when we first booted the system is the same name after a reboot. That will exercise the legacy script once your other PR goes in, and for the future, it'll sanity-check that we're not unknowingly changing interface naming schemes across updates.

dustymabe (approved):
LGTM

PR #492, opened 2020-06-24
[next] overlay/fcos: pin legacy ifnames on existing machines

dustymabe:
shipping such unit in the OS.

(cherry picked from commit d67f11e05f22c14b85079eaf6af7161d9925aa6f)


dustymabe:
context: https://github.com/coreos/fedora-coreos-tracker/issues/484#issuecomment-649009580

jlebon (approved):

PR #493, opened 2020-06-25
[testing] overlay/fcos: pin legacy ifnames on existing machines

jlebon:
shipping such unit in the OS.

(cherry picked from commit d67f11e)


dustymabe:
any chance you could run `git cherry-pick` with the `-x` arg?

dustymabe (approved):
LGTM

PR #498, opened 2020-06-26
{manifests,overlay}: add ssh-key-dir

bgilbert:
Drop `sshd_config.d` fragment overriding `AuthorizedKeysFile`, since ssh-key-dir will pull in the Ignition and Afterburn fragments now.

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -1,2 +1,5 @@
-# no overrides at this time
-packages: {}
+packages:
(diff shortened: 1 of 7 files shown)


jlebon (commented):
Hmm, isn't this missing a patch to `sshd_config` to make it use `AuthorizedKeysCommand=/usr/libexec/ssh-key-dir` or something?

bgilbert:
Nope, the ssh-key-dir RPM drops a fragment in `/etc/ssh/sshd_config.d`.

jlebon (approved):

PR #499, opened 2020-06-26
 live-generator: Avoid tmpfs/overlayfs, add stronger deps

cgwalters:
This has many things rolled into one commit because I just
needed to test them together.  The high level goal is SELinux
enforcing for RHCOS and to mitigate a boot time race (still
not fully debugged, ended up adding `After=systemd-udev-settle.service`)

The SELinux enforcing one turned up a huge mess because as
it happens SELinux will nuke all labels on `tmpfs` after policy
is loaded.  Which is very problematic in general (FCOS included)
but FCOS was papering over this problem by using the systemd `relabel-extra.d`
bits.

But we don't really want to use that anymore IMO - we should
have a clean model where files are always labeled correctly in the initramfs
before we switch root.  Anything else going to lead to pain.

In order to work around the SELinux/`tmpfs` bug, instead make a loopback-mounted
`xfs` filesystem (on `tmpfs`).

When the kernel is fixed to retain labels in `tmpfs` we can drop that
hack.


cgwalters:
Still debugging this.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -144,20 +144,24 @@ EOF
...
     common_etcvar_unit > "${UNIT_DIR}/sysroot-etc-copy.service"
+    # Note we use /etccopy because coreos-populate-writable already makes
+    # /writable/etc (which we could clean up but hopefully we'll delete
+    # this code entirely later once RHEL8 can do ovl on squashfs)
     cat >>"${UNIT_DIR}/sysroot-etc-copy.service" <<EOF
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/bin/cp -a /sysroot/etc /writable/etc-copy
+ExecStart=/bin/sh -c 'mkdir -p /writable/etccopy && /bin/cp -a /sysroot/etc /writable/etccopy/etc && /sbin/setfiles -F -r /writable/etccopy /sysroot/etc/selinux/targeted/contexts/files/file_contexts /writable/etccopy'
```
jlebon: WDYT about just adding another unit which runs after `sysroot-etc.mount` and does `ExecStart=/usr/bin/coreos-relabel /etc`? Feels cleaner since it's then part of the `/sysroot` rootfs proper too instead of having to create one just for `setfiles`. Not against doing it all in one unit though!

cgwalters: I think if we can be sure no other units ran trying to operate on `/sysroot/etc` until the relabeling was done.  Which...I think we can.  I will investigate this.

jlebon: The guideline I try to follow for relabeling is that every component is responsible for relabeling what it touched. So ideally it shouldn't matter whether something needs to modify a file in `/etc` before or after the mass relabeling.

cgwalters: I agree that relabeling can happen before or after files are changed, but it clearly is going to lead to races if we relabel *concurrently* with files changing right?

jlebon: Hmm, can you expand on this? If every service always makes sure to call `coreos-relabel` after doing things in `/etc`, then it shouldn't matter if e.g. one service is changing a file while it's being relabeled since it will also trigger a relabel once it's done, right?

cgwalters: Yeah, I guess so.  In the end though these services are now moved to be `Before=initrd-root-fs.target` so nothing else (most notably `ignition-files.service` should be touching them).

cgwalters:
OK so...`coreos-relabel` here seems not to be working, it's not obvious to me why.  I see a spam of "relabeling" messages...but when we get to the real root it's all still `tmpfs_t`.  The absolutely maddening thing is 
1) Missing tooling for inspecting xattrs in the initramfs, I am manually adding `attr`
2) Even after getting that, current kernel disallows reading them...

cgwalters:
If I manually do:

```
chroot /sysroot
mount -t sysfs sysfs /sys
mount -t selinuxfs selinuxfs /sys/fs/selinux
/sbin/load_policy
<ctrl-d> back to initramfs
```
Then I can see the labels are still `tmpfs_t`, but now running `coreos-relabel` works.

Hum...I notice there's a zero `0` in the setfiles arguments but we don't seem to be `NUL` separating them?

cgwalters:
And notably this seems to be common to the things done via `ignition-ostree-populate-var.service`; I see e.g.
```
# journalctl -u ignition-ostree-populate-var | grep 'var/usrlocal '
Jun 30 16:10:49 localhost ignition-ostree-populate-var[823]: Relabeled /sysroot//var/usrlocal from unlabeled to system_u:object_r:var_t:s0
# ls -alZd /sysroot/var/usrlocal 
drwxr-xr-x. 11 root root system_u:object_r:tmpfs_t:s0 220 Jun 30 16:10 /sysroot/var/usrlocal
#
```

cgwalters:
With strace I see this:
```
[pid  1051] lgetxattr("/sysroot//etc/nsswitch.conf", "security.selinux", "unlabeled", 255) = 10
[pid  1051] access("/var/run/setrans/.setrans-unix", F_OK) = -1 ENOENT (No such file or directory)
[pid  1051] futex(0x7ff06b2315d0, FUTEX_WAKE_PRIVATE, 2147483647) = 0
[pid  1051] lsetxattr("/sysroot//etc/nsswitch.conf", "security.selinux", "system_u:object_r:etc_t:s0", 27, 0) = 0
[pid  1051] fstat(1, {st_mode=S_IFCHR|0600, st_rdev=makedev(0x5, 0x1), ...}) = 0
[pid  1051] ioctl(1, TCGETS, {B9600 opost isig icanon echo ...}) = 0
[pid  1051] write(1, "Relabeled /sysroot//etc/nsswitch"..., 83Relabeled /sysroot//etc/nsswitch.conf from unlabeled to system_u:object_r:etc_t:s0
) = 83
```

But...after doing the bits to load the policy so I can inspect the xattrs...it appears not to be set.  With ftrace I can see that `selinux_inode_setxattr` is being invoked at least, but not what it's returning.

cgwalters:
Reading some of the kernel code here...I think this may be a behavioral difference in the initramfs between `tmpfs` and `xfs`?

When policy is loaded, the kernel walks over all the inodes from mounted filesystems and ensures they're initialized, and I think this is wiping out any changes we make to `tmpfs` inodes or something like that.

If so, then this would explain why this further code for rhel8 is only broken there - for FCOS we're doing overlayfs on the squashfs, and this is mostly going to work because it would only affect files we've modified.

Yeah, getting more confident in this theory, look at the policy:
https://github.com/fedora-selinux/selinux-policy/blob/862368c92def52e3bccce571a46cd99dce34fc78/policy/modules/kernel/filesystem.te
e.g.
`fs_use_xattr xfs gen_context(system_u:object_r:fs_t,s0);`
versus what's happening for `tmpfs_t` below.

cgwalters:
Blah...maybe the simplest is to try to rework this so that rather than copying everything into `tmpfs` we use devmapper CoW device on the squashfs?

Clearly in parallel we need to get the fhandle bits backported into RHEL8's squashfs so that one can create an overlayfs on top - they go together so well.

cgwalters:
OK I just tested, rather than `tmpfs` it works to use loopback-mounted `xfs` filesystems (backed by a file in `tmpfs`).  So awesome.

cgwalters:
Ahh but FCOS is also using a `tmpfs` for `/var` so shouldn't we be hitting the failure there too.  Theory: FCOS `/var` is only working because it has `/run/systemd/relabel-extra.d/`

jlebon:
> Hum...I notice there's a zero 0 in the setfiles arguments but we don't seem to be NUL separating them?

Ahh yeah, that's left over from when I copied this from the Ignition work where we feed the file list through `setfiles`' standard input. We pass it on the command-line here though, so I think that's probably a red herring (but still worth cleaning up).

Your comments about `tmpfs` ring a bell. I think @bgilbert might've encountered this as well when he was initially hacking on the live arfifacts?

> Theory: FCOS `/var` is only working because it has `/run/systemd/relabel-extra.d/`

Hmm good catch. I thought we had gotten rid of all post-switchroot relabeling in FCOS (but at least `relabel-extra.d` is the best alternative of all of those). Let me play with this locally as well.

jlebon:
Yup, this is easy to reproduce on FCOS by booting with `rd.break` and then:

```
sh-5.0# rm /run/systemd/relabel-extra.d/coreos-writable.relabel
sh-5.0# coreos-relabel /etc
...
sh-5.0# coreos-relabel /var
...
sh-5.0# exit
...
(lots of AVC denials with tcontext=system_u:object_r:tmpfs_t:s0)
```

cgwalters:
OK so...I'll rework both paths to use the loopback-mounted xfs trick instead of `tmpfs`, unless anyone has other ideas.

(Now, I think we *should* fix `tmpfs` + SELinux but...not going to block on that today)

jlebon:
> OK so...I'll rework both paths to use the loopback-mounted xfs trick instead of tmpfs, unless anyone has other ideas.

Hmm, I'm torn. I think using `relabel-extra.d/` for FCOS is better than the loopback, though I know how nice it'd be to unify the approaches here. WDYT about trying to keep it to RHEL8 as a first approach and see how it goes?

cgwalters:
It's going to be a huge mess in the code to have two paths though.  I think the real bug here is to fix `tmpfs` + SELinux in the initramfs.  It may not even be too hard; after that we can drop the loopback mounts.

cgwalters:
I feel like this is making progress but now what I'm seeing...seems to be `systemd-tmpfiles-setup.service` running twice in the initramfs and not in the real root?  A bunch of services fail because e.g. the `/var/run` symlink is missing.

I am currently baffled by what would cause it to run again in the initramfs.  It's not happening outside of live runs AFAICS.

cgwalters:
> seems to be systemd-tmpfiles-setup.service running twice in the initramfs and not in the real root? A bunch of services fail because e.g. the /var/run symlink is missing.

And *that* problem turns out to be that we really do need the `DefaultDependencies=false`, otherwise the `initrd-reload.service` causes just that unit to start again...which really has to be some sort of bug.

Man, that was painful to figure out.  I only stumbled into doing so because when initially testing changes for the race a week or two ago the first thing I tried was removing those `DefaultDependencies=` and got the same failure, but I'd forgotten about it and then only dimly recalled it when debugging this.

cgwalters:
YES
```
[core@localhost ~]$ sudo su -
Last login: Wed Jul  1 13:26:56 UTC 2020 on tty1
[root@localhost ~]# rpm-ostree status
State: idle
Deployments:
* ostree://0959f1d922a78d3e58f9ad064e73b9dc82fcbf151b28d3d7d9c349f79a3457d1
                   Version: 46.82.202007011315-0 (2020-07-01T13:18:39Z)
[root@localhost ~]# getenforce 
Enforcing
[root@localhost ~]# ls -al /run/ostree-live 
-rw-r--r--. 1 root root 0 Jul  1 13:26 /run/ostree-live
[root@localhost ~]# 
```

OK now I've introduced some other race in bootup - seeing `run-ephemeral.mount` failed with `Operation not permitted` 
and its dependency `sysroot-xfs-ephemeral-mkfs.service` be `Inactive (dead)` which shouldn't happen because it's `RemainAfterExit=yes`.

But this is feeling close...

cgwalters:
> OK now I've introduced some other race in bootup - seeing run-ephemeral.mount failed with Operation not permitted
and its dependency sysroot-xfs-ephemeral-mkfs.service be Inactive (dead) which shouldn't happen because it's RemainAfterExit=yes.

That seemed to somehow be related to the service starting too early.

cgwalters:
OK, lifting WIP on this!  Tested with both FCOS/RHCOS, we have SELinux enforcing and haven't seen a boot race yet, but I'm working on some coreos-assembler patches to make testing that in an automated fashion easier.

cgwalters:
And now building on top of https://github.com/coreos/coreos-assembler/pull/1571/ I have this going:

```
walters@toolbox /v/s/w/b/rhcos-master> jobs
Job	Group	CPU	State	Command
4	1866544	0%	running	bash -c 'while kola run --output-dir=tmp/kola-iso3 --qemu-image builds/latest/x86_64/rhcos-46.82.202007011714-0-live.x86_64.iso --qemu-memory 8192 basic; do :; done'&
3	1866037	0%	running	bash -c 'while kola run --output-dir=tmp/kola-iso2 --qemu-image builds/latest/x86_64/rhcos-46.82.202007011714-0-live.x86_64.iso --qemu-memory 8192 basic; do :; done'&
2	1865987	0%	running	bash -c 'while kola run --output-dir=tmp/kola-iso1 --qemu-image builds/latest/x86_64/rhcos-46.82.202007011714-0-live.x86_64.iso --qemu-memory 8192 basic; do :; done'&
1	1865911	0%	running	bash -c 'while kola run --output-dir=tmp/kola-iso0 --qemu-image builds/latest/x86_64/rhcos-46.82.202007011714-0-live.x86_64.iso --qemu-memory 8192 basic; do :; done'&
```

So let's see if we can find any races...

OK, did over 100 boots this time, no failures.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -109,67 +117,105 @@ Type=squashfs
...
-# Make sure our tmpfs is available
-RequiresMountsFor=/writable
-
-# Need to do this before Ignition mounts any other filesystems (potentially
-# shadowing our own mount).
-Before=ignition-mount.service
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/sh -c 'set -euo pipefail; mem=$$(grep "^MemTotal" /proc/meminfo | sed -e "s,^.*: *\\([0-9]*\\) .*,\\1,") && /bin/truncate -s $${mem}k /run/ephemeral.xfsloop'
```
darkmuggle: What do you think about `awk '/^MemTotal/{print$$2;exit;}' /proc/meminfo`?  I think we already have `awk` in the initramfs and its a bit _easier_ to read. 

And if I am gawking this correctly this allows for 100% of the total memory. I think that having a percentage of RAM would better. 

edit: https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt says that it will be capped at half the RAM. So it won't OOM. But it could have write errors since the backing memory is smaller than the allocated space.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -109,67 +117,105 @@ Type=squashfs
...
+add_requires sysroot-etc.mount initrd-root-fs.target
 
 common_etcvar_unit >"${UNIT_DIR}/sysroot-var.mount"
 cat >>"${UNIT_DIR}/sysroot-var.mount" <<EOF
 [Mount]
-What=/writable/var
+What=/run/ephemeral/var
 Where=/sysroot/var
 Type=none
 Options=bind
```
darkmuggle, line 204: `Options=bind,discard` Could be really helpful here. 

Since the filesystem is being swapped from squashfs (compressed) to XFS, this change runs the risk of causing memory exhaustion.

To whit:  
```
[root@porpentina mnt]# df -h tmpfs/
Filesystem      Size  Used Avail Use% Mounted on
none            500M     0  500M   0% /var/mnt/tmpfs
[root@porpentina tmpfs]# truncate -s 500M tmpfs/test.img
[root@porpentina tmpfs]# df -h /mnt/tmpfs
Filesystem      Size  Used Avail Use% Mounted on
none            500M     0  500M   0% /var/mnt/tmpfs

[root@porpentina tmpfs]# mkfs.xfs test.img
<REDACTED>

[root@porpentina mnt]# mkdir test
[root@porpentina mnt]# mount -o loop tmpfs/test.img test
[root@porpentina mnt]# dd if=/dev/urandom of=test/file1 bs=1M count=100
100+0 records in
100+0 records out
104857600 bytes (105 MB, 100 MiB) copied, 2.81126 s, 37.3 MB/s
[root@porpentina mnt]# df tmpfs
Filesystem     1K-blocks   Used Available Use% Mounted on
none              512000 108556    403444  22% /var/mnt/tmpfs

[root@porpentina mnt]# rm test/file1
[root@porpentina mnt]# sync
[root@porpentina mnt]# df tmpfs
Filesystem     1K-blocks   Used Available Use% Mounted on
none              512000 108556    403444  22% /var/mnt/tmpfs
```

darkmuggle (commented):
See my comment about disk-space exhaustion and discards. 

As is, this could be interesting. Consider:
1. the `tmpfs` is capped at 50% of the RAM
2. the filesystem is allocated for the full amount of RAM
3. writes could fail since the filesystem is bigger than the backing disk
4. Removing files won't help since space is still allocated.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
cgwalters: > Since the filesystem is being swapped from squashfs (compressed) to XFS,

We're still using squashfs for `/` - this is changing the backing storage for the writable/ephemeral layer only (i.e. `/etc` and `/var`).

But, discard definitely makes sense!

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
cgwalters: Right, actually `tmpfs` uses [50% of RAM by default](https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt) (totally arbitrary and I think anyone using the Live ISO/PXE would want to reconfigure this in some situations).  So what we're doing here is wrong in that we're creating a sparse file larger than could be written.

How about sizing the filesystem to the total available in `/run`?  Something like
```
[root@cosa-devsh ~]# echo $(($(stat -f -c '%b * %s / 1024' /run)))
495516
```

None of this is really right but I think for most use cases it doesn't matter, and for those it does (say a lot of container images in `/var/lib/containers`), an admin can make that a `tmpfs` mount point with a specific size.

cgwalters: > Removing files won't help since space is still allocated.

Keep in mind an admin can still invoke `fstrim` on demand (and requiring them to do this is the current default) - `discard` is just turning it on by default *synchronously* (which is totally fine in this case because this is all just in memory and so there's no latency concerns).

cgwalters:
Updated per above discussion.  I still want to try to diagnose exactly the `tmpfs`/SELinux interaction suitable for a bug report; i'm not 100% sure it's the kernel and not e.g. how systemd relabels `/run`.  But, I don't think we should block on that either.

darkmuggle (approved):
lgtm

Good work @cgwalters

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -78,14 +78,22 @@ else
...
+    cat > "${initrd_rootdev_target_d}/50-root-device.conf" <<EOF
+[Unit]
+After=${isosrc_escaped}
+Requires=${isosrc_escaped}
+EOF
     cat >"${UNIT_DIR}/run-media-iso.mount" <<EOF
 # Automatically generated by live-generator
 
 [Unit]
 DefaultDependencies=false
```
jlebon, line 92: Not new, though since we're on the hunt for unnecessary `DefaultDependencies=`, I think this is one of them since `initrd-root-device.target` is *after* `basic.target`.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -78,14 +78,22 @@ else
...
+After=initrd-root-device.target
+# HACK for https://github.com/coreos/fedora-coreos-config/issues/437
+After=systemd-udev-settle.service
```
jlebon, line 95: Are you sure this is still needed? `systemd-udev-settle.service` is before `sysinit.target`, which is before `basic.target` (which is before `initrd-root-device.target`).

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -109,67 +117,105 @@ Type=squashfs
...
+# It turns out that `tmpfs` currently munches all SELinux labels
+# we set before policy is loaded, so we make an XFS filesystem
+# loopback mounted that's sized the same as /run.
+# https://github.com/coreos/fedora-coreos-config/pull/499
+cat >"${UNIT_DIR}/sysroot-xfs-ephemeral-mkfs.service" <<'EOF'
 [Unit]
 DefaultDependencies=false
+# Let's be sure we have basic devices, but other than that we
+# can run really early.
+After=systemd-tmpfiles-setup-dev.service
```
jlebon, line 129: Hmm, is this for `/run`? It's mounted by systemd itself on startup.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -109,67 +117,105 @@ Type=squashfs
...
+EOF
+
+common_etcvar_unit() {
+    cat << EOF
+# Automatically generated by live-generator
+[Unit]
+DefaultDependencies=false
+# Make sure our tmpfs is available
+Requires=sysroot-xfs-ephemeral-setup.service
+After=sysroot-xfs-ephemeral-setup.service
```
jlebon, line 184: I think this should also have `Before=initrd-root-fs.target` since the `etc` and `var` mounts should be considered part of setting up the root fs too, right?

jlebon (approved):
Nice job debugging this! Just a couple of comments but LGTM overall.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
cgwalters: I don't see much ordering around `initrd-root-device.target`:
```
sh-5.0# systemctl show initrd-root-device.target |grep -Ee 'Before|After'
Before=initrd.target
After=ignition-disks.service
```

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
cgwalters: Right, I thought so too but I was seeing weird failures until I added this, see https://github.com/coreos/fedora-coreos-config/pull/499#issuecomment-652552890

cgwalters: In general I'm skeptical of units with absolutely no dependencies at all.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
cgwalters: Yeah, no harm in adding that, but we already have that implicitly because `sysroot-relabel.service` (the final service in the chain) has it, and is ordered after those units, so it forces the whole chain.

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
jlebon: Hmm weird. I was going according to `bootup(7)`: https://github.com/systemd/systemd/blob/c03ef420fa7157b8d4881636fe72596a06e08bb6/man/bootup.xml#L243-L251. But you're right, there's no obvious ordering that seems to implement that.

cgwalters: Hmm, I think actually this may be not sufficient though, the idea is `systemd-udev-settle.service` doesn't run by default, and so one should actually use both `Wants=` and `Requires=` or so.  Ah right, `multipathd.service` is:
```
sh-5.0# systemctl show systemd-udev-settle.service |grep WantedBy
WantedBy=multipathd.service multipathd-configure.service
```

cgwalters:
If we're not seeing anything seriously wrong I'd like to merge this (tested extensively as is) PR, and do other cleanups as a followup that can be tested independently.  Iterating on this is painful!

cgwalters:
Testing this followup now:
```
From 56696592d5fec57e5362bfb5397e4512a937f476 Mon Sep 17 00:00:00 2001
From: Colin Walters <walters@verbum.org>
Date: Thu, 2 Jul 2020 16:20:44 +0000
Subject: [PATCH] wip

---
 .../usr/lib/dracut/modules.d/20live/live-generator   | 12 +++++++-----
 1 file changed, 7 insertions(+), 5 deletions(-)

diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
index 2a9c91e..855dfcc 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -90,9 +90,12 @@ EOF
 
 [Unit]
 DefaultDependencies=false
-After=initrd-root-device.target
 # HACK for https://github.com/coreos/fedora-coreos-config/issues/437
-After=systemd-udev-settle.service
+Wants=systemd-udev-settle.service
+# Note that `man bootup` implies that initrd-root-device is After=basic.target
+# but that appears to not be the case.  We explicitly order after sysinit.target
+After=sysinit.target
+After=initrd-root-device.target
 Before=initrd-root-fs.target
 
 [Mount]
@@ -164,9 +167,6 @@ After=sysroot.mount
 # And after OSTree has set up the chroot() equivalent
 After=ostree-prepare-root.service
 
-# We're part of assembling the root fs
-Before=initrd-root-fs.target
-
 [Service]
 Type=oneshot
 RemainAfterExit=yes
@@ -182,6 +182,8 @@ DefaultDependencies=false
 # Make sure our tmpfs is available
 Requires=sysroot-xfs-ephemeral-setup.service
 After=sysroot-xfs-ephemeral-setup.service
+# We're part of assembling the root fs
+Before=initrd-root-fs.target
 EOF
 }
 
-- 
2.26.2

```

jlebon:
> If we're not seeing anything seriously wrong I'd like to merge this (tested extensively as is) PR, and do other cleanups as a followup that can be tested independently. Iterating on this is painful!

Yup, WFM!

PR #500, opened 2020-06-26
tests/manual: nettest: update FCOS to use ens2/ens3

dustymabe:
Now that we are using persistent NIC naming in FCOS we'll update the
test to also use those interface names. See:
https://github.com/coreos/fedora-coreos-tracker/issues/484

Also add a test that makes sure that net.ifnames=0 works as desired.


dustymabe:
Just so there is no confusion: the ultimate goal is to have this entire test suite in kola tests and this script would go away.

jlebon (approved):

Issue #437, opened 2020-05-29
ISO mount still racing?


cgwalters:
Even after https://github.com/coreos/fedora-coreos-config/pull/411 I just saw this in a CI run for RHCOS:
```
"Found device QEMU_DVD-ROM rhcos-46.82.202005292045-0."
"Mounting /run/media/iso..."
"mount: /run/media/iso: /dev/sr0 already mounted or mount point busy."
"run-media-iso.mount: Mount process exited, code=exited status=32"
"run-media-iso.mount: Failed with result 'exit-code'."
"Failed to mount /run/media/iso."
```

I'm 87.2% sure we have that PR, though I need to triple check.

dustymabe:
I just saw it locally:

```
[    5.131654] ata1.01: ATAPI: QEMU DVD-ROM, 2.5+, max UDMA/100                                                                                                                                                                                                                                          
[    5.134278] scsi 0:0:0:0: Direct-Access     ATA      QEMU HARDDISK    2.5+ PQ: 0 ANSI: 5                                                                                                                                                                                                              
[    5.136662] scsi 0:0:1:0: CD-ROM            QEMU     QEMU DVD-ROM     2.5+ PQ: 0 ANSI: 5                                                                                                                                                                                                              
[    5.167197] scsi 0:0:0:0: Attached scsi generic sg0 type 0                                                                                                                                                                                                                                            
[    5.168497] scsi 0:0:1:0: Attached scsi generic sg1 type 5                                                                                                                                                                                                                                            
[    5.175112] sd 0:0:0:0: [sda] 41943040 512-byte logical blocks: (21.5 GB/20.0 GiB)                                                                                                                                                                                                                    
[    5.176458] sd 0:0:0:0: [sda] Write Protect is off                                                                                               
[    5.177427] sd 0:0:0:0: [sda] Write cache: enabled, read cache: enabled, doesn't support DPO or FUA
[    5.190230] sr 0:0:1:0: [sr0] scsi3-mmc drive: 4x/4x cd/rw xa/form2 tray
[    5.191576] cdrom: Uniform CD-ROM driver Revision: 3.20
[    5.306952] e1000 0000:00:02.0 eth0: (PCI:33MHz:32-bit) 52:54:00:47:bc:0e
[    5.308249] e1000 0000:00:02.0 eth0: Intel(R) PRO/1000 Network Connection
[...]

PR #501, opened 2020-06-26
live-generator: Add Options=ro to our mounts

cgwalters:
The default is a writable mount, and right now the
kernel emits a warning when the mount succeeds
because the underlying block device is read-only.

Which...actually turns out to *also* be the cause behind
the races we've been seeing mounting the ISO
https://github.com/coreos/fedora-coreos-config/issues/437

I believe what's happening is one of the bits of udev
is probing the ISO and doing a read-only mount at the same time we're
trying to mount it.

From the kernel perspective it's fine to have multiple concurrent
read-only mounts, but because ours was requested as writable,
if we happend to be trying to do our mount at the same time
as the probing the kernel will reject it.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -90,6 +90,7 @@ Requires=${isosrc_escaped}
 [Mount]
 What=/${isosrc}
 Where=/run/media/iso
+Options=ro
 Type=iso9660
 EOF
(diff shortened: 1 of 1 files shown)


cgwalters:
I don't know why this is seemingly OK in FCOS...I think there might be some other udev rules in RHCOS, or something else in the way boot ordering is happening.

But I am 10/10 RHCOS ISO boots with this.

jlebon (approved):
Ahhh so evil!

PR #502, opened 2020-06-29
WIP manifests: add nm-cloud-setup

sohankunkerkar:
Reference: https://github.com/coreos/fedora-coreos-config/pull/377#issuecomment-651183953
Now that https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/546 got merged, NetworkManager-cloud-setup can support ` GCP` load-balancing routes. We'd need this package on an FCOS machine to run it on the bootstrap node.

More references:
- https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/905
- https://bugzilla.redhat.com/show_bug.cgi?id=1822603

/cc @LorbusChris

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -111,6 +111,9 @@ packages:
   - iptables nftables iptables-nft iptables-services
   # Interactive Networking configuration during coreos-install
   - NetworkManager-tui
+  # Automatically configure NetworkManager in cloud
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1822603
+  - NetworkManager-cloud-setup
   # Storage
   - cloud-utils-growpart
   - lvm2 iscsi-initiator-utils sg3_utils


jlebon (commented):
This will need to be added to the lockfiles as well. See https://github.com/coreos/fedora-coreos-config#adding-packages-to-the-os.

Re. the references, it saddens me that 2 of the 3 listed are private. At least the RHBZ it seems like the first comment can be unmarked private?

sohankunkerkar:
I think the changes are still in the testing phase and have not made their way into a `stable` build.
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/a2b699f40f29fb1d37cda8c9a10365229f55b0c8#note_551816.

miabbott:
> I think the changes are still in the testing phase and have not made their way into a `stable` build.
> https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/a2b699f40f29fb1d37cda8c9a10365229f55b0c8#note_551816.

I think the GCP functionality is coming in NM 1.26 based on the name of the branch and where the commits got picked to - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commits/nm-1-26

sohankunkerkar:
> I think the GCP functionality is coming in NM 1.26 based on the name of the branch and where the commits got picked to - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commits/nm-1-26

Seeing it here https://github.com/NetworkManager/NetworkManager/compare/1.27.0-dev...master

jlebon:
Looks like f32 is still at 1.22: https://bodhi.fedoraproject.org/updates/FEDORA-2020-3857463d30, while f33 is at 1.26: https://bodhi.fedoraproject.org/updates/FEDORA-2020-2db8e9b937. We should ask the NM folks whether they're planning to rebase f32.

squeed:
How is this configured? We only want it enabled on the bootstrap node. You may have to update the MCO to disable nm-cloud-setup on the real cluster.

lucab:
> How is this configured?

Good question. It looks like the service unit has the following comment:
```
# Opt-in by setting the right environment variable for
# the provider.
#Environment=NM_CLOUD_SETUP_EC2=yes
#Environment=NM_CLOUD_SETUP_GCP=yes
```

I think we should have some glue to forward our platform ID into these environment flags by default so that it works out of the box, and then disabling the service would go via Ignition configuration. By the way, is there a reason why this shouldn't be enabled on non-bootstrap nodes, other than not doing anything strictly useful there?

sohankunkerkar:
> Looks like f32 is still at 1.22: https://bodhi.fedoraproject.org/updates/FEDORA-2020-3857463d30, while f33 is at 1.26: https://bodhi.fedoraproject.org/updates/FEDORA-2020-2db8e9b937. We should ask the NM folks whether they're planning to rebase f32.

Alright, I had a discussion with Thomas Haller from the NetworkManager team. A quick summary of that:
- They(NM Maintainers) are not planning to rebase to F32 after the `Fedora devel freeze` which means, F32 will stick to 1.22.x, but it will get rebased to future (minor) 1.22 releases.
- An ideal solution is to backport that change, but in this case, it would be challenging. While the nm-cloud-setup change is self-contained, it requires a new API. Backporting new API to old branches is very much cumbersome.
- In theory, they could rebase F32 to 1.26.0. But in practice that only has the potential to break stuff for users. One workaround here is to build a 1.26.0 NetworkManager package for CoreOS/F32 i.e. rebuild the F33 package against F32 and use that.
/cc @jlebon @LorbusChris

jlebon:
> One workaround here is to build a 1.26.0 NetworkManager package for CoreOS/F32 i.e. rebuild the F33 package against F32 and use that.

We have the ability to do this, but haven't had to yet. If NM maintainers think it's risky, then I hesitate to carry that delta as well. Feels safer in that case to carry whatever minimal script we need here? Curious what others think. (Or as Colin [suggested](https://github.com/coreos/fedora-coreos-config/pull/377#issuecomment-627464103), just move it to the MCO and special-case the bootstrap node.)

miabbott:
> We have the ability to do this, but haven't had to yet. If NM maintainers think it's risky, then I hesitate to carry that delta as well. Feels safer in that case to carry whatever minimal script we need here? Curious what others think. (Or as Colin [suggested](https://github.com/coreos/fedora-coreos-config/pull/377#issuecomment-627464103), just move it to the MCO and special-case the bootstrap node.)

Since networking is such a crucial piece of our stack, I'm also hesitant to try to pull off front-running 1.26 in FCOS.  (And we'd need something similar for RHCOS, too).

Developing a minimal script that we can include in both places (FCOS/RHCOS) seems like the best path forward.

LorbusChris:
> Developing a minimal script that we can include in both places (FCOS/RHCOS) seems like the best path forward.

:+1: 
That sounds like the next best alternative to me for now

cgwalters:
If we're going to do something short term, why not stick it in the MCO? 
https://github.com/coreos/fedora-coreos-config/pull/377#issuecomment-627464103
Again we totally can solve the issue of running on the bootstrap node because the MCO *does* get invoked by `bootkube.sh` today, all we'd need to do is start another thing there.

miabbott:
> If we're going to do something short term, why not stick it in the MCO?
> [#377 (comment)](https://github.com/coreos/fedora-coreos-config/pull/377#issuecomment-627464103)
> Again we totally can solve the issue of running on the bootstrap node because the MCO _does_ get invoked by `bootkube.sh` today, all we'd need to do is start another thing there.

A script delivered via the MCO meets my desire to get something in FCOS/RHCOS for the bootstrap case.  And it means we just have to write it once in a single repo (not having to worry about getting RHCOS synced to FCOS etc).  So no objections to this path.

cgwalters:
Closing in favor of https://github.com/openshift/machine-config-operator/issues/1883

cgwalters:
Well sorry we should *still* do this...just later.

sohankunkerkar:
closing this in favor of https://github.com/coreos/fedora-coreos-config/pull/760

PR #466, opened 2020-06-10
*: add Ignition LUKS support

arithx:
Adds LUKS related packages

PR #503, opened 2020-06-29
40ignition-ostree: add coreos-inject-rootmap.service

### LINKED ISSUE #94: Support for reconfiguring the root storage
FCOS will have a static, server-side generated initramfs. This means the initramfs needs to know how to find the root filesystem to mount it, including if that means starting RAID devices. However, we do NOT want to start other raid devices (unless Ignition is starting them as part of the creation process) since users might be writing out configuration to things like `/etc/mdadm.conf` that they want to use.

On CL we require that RAIDed devices be partitions with the a special GPT partition type guid and have a udev rule to start only those.

For FCOS I can see two options:
 - Do something similar to what CL did. This has the limitation of being unable to use whole disks for raid and requires partition tables where they would not otherwise be needed.
 - Have some sort of file that gets written out to `/boot` which is a "rootmap". This would describe the "path to root" including what services like mdadm need to be started. Ignition could generate this if we include a way of flagging partitions/RAIDs as "containing root" or it could be user written. If it is user written we should unmount root after Ignition files/umount and then let whatever tooling reads it mount the root itself to ensure the first boot is not "special" in any way.

### LINKED ISSUE #465: switch to root=<uuid> post-firstboot?
Currently we rely on finding the rootfs by just assuming `/dev/disk/by-label/root`, on both firstboot and subsequent boots.

One problem I hit is that if you e.g. image a USB stick with the FCOS metal image and put it in a machine that has an existing install, you can easily boot from the main hard drive but have the OS mount the USB stick as `/`, which is confusing to say the least.

One option would be for us to extend the existing [GPT uuid regeneration](https://github.com/coreos/ignition-dracut/blob/master/dracut/30ignition/coreos-gpt-setup@.service) to also generate a new XFS uuid for the rootfs, and inject a `root=<uuid>` karg using it by default.

It's also just cleaner for e.g. backup/snapshot systems if they don't have to deal with potentially duplicate UUIDs.

I can't think of any downsides to this really.

The other reason to do this is in order to support [complex root devices](https://github.com/coreos/fedora-coreos-tracker/issues/94) we'll end up changing the kernel cmdline, and so we might as well do it always.

jlebon:
This implements the rootmap functionality that figures out all the
dependencies required to find `/sysroot`, and injects them into the BLS
config. For more information, see:

https://github.com/coreos/fedora-coreos-tracker/issues/94#issuecomment-650433166

The `rdcore` code supports RAID and LUKS devices, though the latter
needs a new Clevis release with the following patches to be fully
supported:

latchset/clevis#211
latchset/clevis#217

This also implements the `root=UUID=$uuid` inject patch proposed in
coreos/fedora-coreos-tracker#465.

On its own, this unlocks reprovisioning FCOS with root on a RAID device,
or e.g. in-place reprovisioning of root on btrfs.

Closes: coreos/fedora-coreos-tracker#465
Closes: coreos/fedora-coreos-tracker#94

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
@@ -87,4 +87,5 @@ case "${TYPE}" in
     ext4) resize2fs "${src}" ;;
 esac
 
-touch /var/lib/coreos-growpart.stamp
+# this is useful for tests
+touch /run/coreos-growpart.stamp
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
@@ -8,12 +8,32 @@ set -euo pipefail
 # Growpart /, until we can fix Ignition for separate /var
 # (And eventually we want ignition-disks)
 
+saved_partstate=/run/ignition-ostree-rootfs-partstate.json
+
 path=$1
 shift
 
 # The use of tail is to avoid errors from duplicate mounts;
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
+
+if [ "${path}" == /sysroot ] && [ -f "${saved_partstate}" ]; then
+    # We're still ironing out our rootfs automatic growpart story, see e.g.:
+    # https://github.com/coreos/fedora-coreos-tracker/issues/570
+    # https://github.com/coreos/fedora-coreos-tracker/issues/586
+    #
+    # In the context of rootfs reprovisioning, for now our rule is the
+    # following: if the rootfs partition was moved off of the boot disk or it
+    # was resized, then we don't growpart.
+    #
+    # To detect this, we compare the output of `lsblk -o PATH,SIZE` before and
+    # after `ignition-disks.service`.
+    partstate=$(lsblk "${src}" --nodeps --json -b -o PATH,SIZE | jq -c .)
+    if [ "${partstate}" != "$(cat "${saved_partstate}")" ]; then
+        echo "coreos-growpart: detected rootfs partition changes; not auto-growing"
+        exit 0
+    fi
+fi
 
 # Get the filesystem type before extending the partition.  This matters
 # because the partition, once extended, might include leftover superblocks
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=CoreOS Inject Rootmap
+# If root is specified, assume rootmap is already configured
+ConditionKernelCommandLine=!root
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+After=ignition-files.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/rdcore rootmap /sysroot --boot-device /dev/disk/by-label/boot
+RemainAfterExit=yes
+MountFlags=slave
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
@@ -7,6 +7,7 @@ set -euo pipefail
 ignition_cfg=/run/ignition.json
 rootdisk=/dev/disk/by-label/root
 saved_sysroot=/run/ignition-ostree-rootfs
+partstate=/run/ignition-ostree-rootfs-partstate.json
 
 case "${1:-}" in
     detect)
@@ -21,6 +22,8 @@ case "${1:-}" in
         mount "${rootdisk}" /sysroot
         echo "Moving rootfs to RAM..."
         cp -aT /sysroot "${saved_sysroot}"
+        # also store the state of the partition
+        lsblk "${rootdisk}" --nodeps --json -b -o PATH,SIZE | jq -c . > "${partstate}"
         ;;
     restore)
         # This one is in a private mount namespace since we're not "offically" mounting
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+set -euo pipefail
+
+# This is implementation details of Ignition; in the future, we should figure
+# out a way to ask Ignition directly whether there's a filesystem with label
+# "root" being set up.
+ignition_cfg=/run/ignition.json
+rootdisk=/dev/disk/by-label/root
+saved_sysroot=/run/ignition-ostree-rootfs
+
+case "${1:-}" in
+    detect)
+        wipes_root=$(jq '.storage?.filesystems? // [] | map(select(.label == "root" and .wipeFilesystem == true)) | length' "${ignition_cfg}")
+        if [ "${wipes_root}" = "0" ]; then
+            exit 0
+        fi
+        echo "Detected rootfs replacement in fetched Ignition config: /run/ignition.json"
+        mkdir "${saved_sysroot}"
+        ;;
+    save)
+        mount "${rootdisk}" /sysroot
+        echo "Moving rootfs to RAM..."
+        cp -aT /sysroot "${saved_sysroot}"
+        ;;
+    restore)
+        # This one is in a private mount namespace since we're not "offically" mounting
+        mount "${rootdisk}" /sysroot
+        echo "Restoring rootfs from RAM..."
+        cd "${saved_sysroot}"
+        find . -mindepth 1 -maxdepth 1 -exec mv -t /sysroot {} \;
+        chattr +i $(ls -d /sysroot/ostree/deploy/*/deploy/*/)
+        ;;
+    *)
+        echo "Unsupported operation: ${1:-}" 1>&2; exit 1
+        ;;
+esac
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
@@ -1,20 +1,17 @@
 #!/bin/bash
 set -euo pipefail
 
-# We use prjquota on XFS by default to aid multi-tenant
-# Kubernetes (and other container) clusters.  See
-# https://github.com/coreos/coreos-assembler/pull/303/commits/6103effbd006bb6109467830d6a3e42dd847668d
-# In the future this will be augmented with a check for whether
-# or not we've reprovisioned the rootfs, since we don't want to
-# force on prjquota there.
+# Note that on *new machines* this script is now only ever used on firstboot. On
+# subsequent boots, systemd-fstab-generator mounts /sysroot from the
+# root=UUID=... and rootflags=... kargs.
+
+# We may do a migration window at some point where older machines have these
+# kargs injected so that we can simplify the model further.
+
 rootpath=/dev/disk/by-label/root
 if ! [ -b "${rootpath}" ]; then
   echo "ignition-ostree-mount-sysroot: Failed to find ${rootpath}" 1>&2
   exit 1
 fi
-eval $(blkid -o export ${rootpath})
-mountflags=
-if [ "${TYPE}" == "xfs" ]; then
-  mountflags=prjquota
-fi
-mount -o "${mountflags}" "${rootpath}" /sysroot
+
+mount -o "$(coreos-rootflags)" "${rootpath}" /sysroot
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-rootfs-detect.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-rootfs-detect.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=Ignition OSTree: detect rootfs replacement
+DefaultDependencies=false
+After=ignition-fetch.service
+Before=ignition-disks.service
+Before=initrd-root-fs.target
+Before=sysroot.mount
+ConditionKernelCommandLine=ostree
+
+# This stage requires udevd to detect disks
+Requires=systemd-udevd.service
+After=systemd-udevd.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+EnvironmentFile=/run/ignition.env
+ExecStart=/usr/libexec/ignition-ostree-dracut-rootfs detect
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-boot.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-boot.service
@@ -1,0 +1,23 @@
+[Unit]
+Description=Ignition OSTree: Regenerate filesystem UUID (boot)
+DefaultDependencies=false
+ConditionPathExists=/usr/lib/initrd-release
+ConditionKernelCommandLine=ostree
+ConditionPathExists=!/run/ostree-live
+# We run pretty early
+Before=coreos-copy-firstboot-network.service
+Before=ignition-fetch.service
+Before=ignition-setup-base.service
+Before=ignition-setup-user.service
+# Any services looking at mounts need to order after this
+# because it causes device re-probing.
+After=coreos-gpt-setup.service
+
+Before=systemd-fsck@dev-disk-by\x2dlabel-boot.service
+Requires=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/ignition-ostree-firstboot-uuid boot
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-copy-firstboot-network.service
@@ -36,10 +36,6 @@ After=dracut-cmdline.service
 # Any services looking at mounts need to order after this
 # because it causes device re-probing.
 After=coreos-gpt-setup.service
-# Older unit name kept for compatibility. This will be safe to remove once the
-# dracut-ignition changes from [1] are available in the latest ignition RPM.
-# [1] https://github.com/coreos/ignition-dracut/pull/191
-After=coreos-gpt-setup@dev-disk-by\x2dlabel-root.service
 # Since we are mounting /boot/, require the device first
 Requires=dev-disk-by\x2dlabel-boot.device
 After=dev-disk-by\x2dlabel-boot.device
--- a/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-gpt-setup.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-gpt-setup.service
@@ -7,7 +7,7 @@ Before=ignition-diskful.target
 Wants=systemd-udevd.service
 After=systemd-udevd.service
 
-# This unit must the first to run when the disk holding the root partition
+# This unit must be the first to run when the disk holding the root partition
 # becomes available. To avoid relying on the name of the root partition which
 # is different between RHCOS LUKS setup and current FCOS setup, we wait for the
 # partition labeled 'boot' to become available. This is reliable as we don't
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-rootflags.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-rootflags.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+
+rootpath=/dev/disk/by-label/root
+
+# If the rootfs was reprovisioned, then the user is free to define their own
+# rootflags.
+if [ -d /run/ignition-ostree-rootfs ]; then
+    exit 0
+fi
+
+eval $(blkid -o export ${rootpath})
+# this really should always be true, but let's be conservative
+if [ "${TYPE}" == "xfs" ]; then
+    # We use prjquota on XFS by default to aid multi-tenant Kubernetes (and
+    # other container) clusters.  See
+    # https://github.com/coreos/coreos-assembler/pull/303/commits/6103effbd006bb6109467830d6a3e42dd847668d
+    echo "prjquota"
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -euo pipefail
+# https://github.com/coreos/fedora-coreos-tracker/issues/465
+# coreos-assembler generates disk images which are installed bit-for-bit
+# or booted directly in the cloud.
+# Generate new UUID on firstboot; this is general best practice, but in the future
+# we may use this for mounting by e.g. adding a boot=<uuid> and root=<uuid> kernel args.
+# Note this code should be changed when we land https://github.com/coreos/fedora-coreos-tracker/issues/94
+# so that we only generate a new UUID if we're using the default one.
+
+label=$1
+
+# Keep this in sync with https://github.com/coreos/coreos-assembler/blob/e3905fd2e138de04184c1cd86b99b0fd83cbe5cf/src/create_disk.sh#L17
+bootfs_uuid="96d15588-3596-4b3c-adca-a2ff7279ea63"
+rootfs_uuid="910678ff-f77e-4a7d-8d53-86f2ac47a823"
+
+target=/dev/disk/by-label/${label}
+if ! [ -b "${target}" ]; then
+  echo "$0: Failed to find block device ${target}" 1>&2
+  exit 1
+fi
+
+eval $(blkid -o export ${target})
+case "${label}" in
+  root) orig_uuid="${rootfs_uuid}"; orig_type=xfs ;;
+  boot) orig_uuid="${bootfs_uuid}"; orig_type=ext4 ;;
+  *) echo "unexpected ${label}"; exit 1 ;;
+esac
+
+if [ "${TYPE}" == "${orig_type}" ] && [ "${UUID}" == "${orig_uuid}" ]; then
+  case "${TYPE}" in
+    # For now we need to fsck first, see https://github.com/coreos/coreos-assembler/pull/1452
+    # Basically we're not passing `metadata_csum_seed` as a mkfs.ext4 option
+    # because grub2 barfs on it.
+    ext4) e2fsck -fy "${target}" && tune2fs -U random "${target}" ;;
+    xfs) xfs_admin -U generate "${target}" ;;
+    *) echo "unexpected filesystem type ${TYPE}" 1>&2; exit 1 ;;
+  esac
+  echo "Regenerated UUID for ${target}"
+else
+  echo "No changes required for ${target} TYPE=${TYPE} UUID=${UUID}"
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-rootfs-restore.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-rootfs-restore.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Ignition OSTree: restore rootfs
+DefaultDependencies=false
+After=ignition-disks.service
+Before=ignition-ostree-growfs.service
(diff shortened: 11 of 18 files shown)


jlebon:
Requires #184, #354, and the [SELinux patch](https://bugzilla.redhat.com/show_bug.cgi?id=1845210).

> With this, it's now possible to configure FCOS with root on a RAID device!

Example Ignition config:

```json
{
  "ignition": {
    "version": "3.2.0-experimental"
  },
  "storage": {
    "disks": [
      {
        "device": "/dev/disk/by-id/virtio-disk1",
        "partitions": [
          {
            "label": "foo"
          }
        ],
        "wipeTable": true
      },
      {
        "device": "/dev/disk/by-id/virtio-disk2",
        "partitions": [
          {
            "label": "bar"
          }
        ],
        "wipeTable": true
      }
    ],
    "raid": [
      {
        "devices": [
          "/dev/disk/by-partlabel/foo",
          "/dev/disk/by-partlabel/bar"
        ],
        "level": "raid1",
        "name": "myroot"
      }
    ],
    "filesystems": [
      {
        "device": "/dev/md/myroot",
        "format": "xfs",
        "wipeFilesystem": true,
        "label": "root"
      }
    ]
  }
}
```

Running with cosa:

```
(cosa)$ kola qemuexec -i config.ign --ignition-direct --memory 4096 --add-disk 5G --add-disk 5G
(vm)$ findmnt /sysroot
TARGET   SOURCE     FSTYPE OPTIONS
/sysroot /dev/md127 xfs    rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize
(vm)$ lsblk --inverse /dev/md127
NAME    MAJ:MIN RM SIZE RO TYPE  MOUNTPOINT
md127     9:127  0   5G  0 raid1 /sysroot
├─vda1  252:1    0   5G  0 part
│ └─vda 252:0    0   5G  0 disk
└─vdb1  252:17   0   5G  0 part
  └─vdb 252:16   0   5G  0 disk
(vm)$ reboot
...
(vm)$ lsblk --inverse /dev/md127
NAME    MAJ:MIN RM SIZE RO TYPE  MOUNTPOINT
md127     9:127  0   5G  0 raid1 /sysroot
├─vda1  252:1    0   5G  0 part
│ └─vda 252:0    0   5G  0 disk
└─vdb1  252:17   0   5G  0 part
  └─vdb 252:16   0   5G  0 disk
(vm)$ cat /proc/cmdline
BOOT_IMAGE=(hd0,gpt1)/ostree/fedora-coreos-cb9be02cb0885ea378b9763a62c5ca5a596b280d182553fb4f2522725f4257be/vmlinuz-5.6.19-300.jl.fc32.x86_64 mitigations=auto,nosmt systemd.unified_cgroup_hierarchy=0 console=tty0 console=ttyS0,115200n8 ignition.platform.id=qemu rd.md.uuid=139abc1a:9fde35c5:1021a2b6:748ce991 root=UUID=ff2ac1ab-f449c-4de0-8396-b19305747456 rw
```

jlebon:
For fun, with https://github.com/coreos/ignition/pull/1010, it's possible to do root on RAID10:

```json
{
  "ignition": {
    "version": "3.2.0-experimental"
  },
  "storage": {
    "disks": [
      {
        "device": "/dev/disk/by-id/virtio-disk1",
        "partitions": [
          {
            "label": "foo"
          }
        ],
        "wipeTable": true
      },
      {
        "device": "/dev/disk/by-id/virtio-disk2",
        "partitions": [
          {
            "label": "bar"
          }
        ],
        "wipeTable": true
      },
      {
        "device": "/dev/disk/by-id/virtio-disk3",
        "partitions": [
          {
            "label": "baz"
          }
        ],
        "wipeTable": true
      },
      {
        "device": "/dev/disk/by-id/virtio-disk4",
        "partitions": [
          {
            "label": "boo"
          }
        ],
        "wipeTable": true
      }
    ],
    "raid": [
      {
        "devices": [
          "/dev/disk/by-partlabel/foo",
          "/dev/disk/by-partlabel/bar"
        ],
        "level": "raid1",
        "name": "foobar"
      },
      {
        "devices": [
          "/dev/disk/by-partlabel/baz",
          "/dev/disk/by-partlabel/boo"
        ],
        "level": "raid1",
        "name": "bazboo"
      },
      {
        "devices": [
          "/dev/md/foobar",
          "/dev/md/bazboo"
        ],
        "level": "raid0",
        "name": "myroot"
      }
    ],
    "filesystems": [
      {
        "device": "/dev/md/myroot",
        "format": "xfs",
        "wipeFilesystem": true,
        "label": "root"
      }
    ]
  }
}
```

Running with cosa:

```
(cosa)$ kola qemuexec -i config.ign --ignition-direct --memory 4096 --add-disk 5G --add-disk 5G --add-disk 5G --add-disk 5G
(vm)$ findmnt /sysroot
TARGET   SOURCE     FSTYPE OPTIONS
/sysroot /dev/md125 xfs    rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize
(vm)$ reboot
...
(vm)$ lsblk --inverse /dev/md125
NAME      MAJ:MIN RM SIZE RO TYPE  MOUNTPOINT
md125       9:125  0  10G  0 raid0 /sysroot
├─md126     9:126  0   5G  0 raid1
│ ├─vdc1  252:33   0   5G  0 part
│ │ └─vdc 252:32   0   5G  0 disk
│ └─vdd1  252:49   0   5G  0 part
│   └─vdd 252:48   0   5G  0 disk
└─md127     9:127  0   5G  0 raid1
  ├─vda1  252:1    0   5G  0 part
  │ └─vda 252:0    0   5G  0 disk
  └─vdb1  252:17   0   5G  0 part
    └─vdb 252:16   0   5G  0 disk
(vm)$ cat /proc/cmdline
BOOT_IMAGE=(hd0,gpt1)/ostree/fedora-coreos-cb9be02cb0885ea378b9763a62c5ca5a596b280d182553fb4f2522725f4257be/vmlinuz-5.6.19-300.jl.fc32.x86_64 mitigations=auto,nosmt systemd.unified_cgroup_hierarchy=0 console=tty0 console=ttyS0,115200n8 ignition.platform.id=qemu ostree=/ostree/boot.1/fedora-coreos/cb9be02cb0885ea378b9763a62c5ca5a596b280d182553fb4f2522725f4257be/0  rd.md.uuid=2bbba6df:428aea5c:763ec8b4:9a2a357b rd.md.uuid=7e1033f7:117cde9d8:8cf8492d:619c679b rd.md.uuid=ac2ed991:60d98db2:7cef9638:773d4426 root=UUUID=1b184d4a-f273-46c7-a359-b84f257f07b8 rw
```

jlebon:
So... I initially wanted to write this in rpm-ostree so it can be in not-bash, and then call it from the initramfs, but it's hard to beat bash in prototyping speed. And then it ended up being much simpler than expected thanks to `lsblk`'s `--inverse` switch, which gives us exactly what we need. One thing I was tempted to do was providing the ability to have Rust code in this repo which cosa compiles at build-time, but didn't want to tackle that discussion yet.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=CoreOS Inject Rootmap
+DefaultDependencies=false
```
cgwalters: Do we really need this?  I feel like we are using this way too much.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-boot.service
```diff
@@ -0,0 +1,20 @@
...
+ConditionPathExists=/usr/lib/initrd-release
+ConditionKernelCommandLine=ostree
+ConditionPathExists=!/run/ostree-live
+# We run pretty early
+Before=coreos-copy-firstboot-network.service
+Before=ignition-fetch.service
+Before=ignition-setup-base.service
+Before=ignition-setup-user.service
+
+Before=systemd-fsck@dev-disk-by\x2dlabel-boot.service
```
cgwalters, line 16: I think these services also need the GPT ordering fix from
https://github.com/coreos/fedora-coreos-config/pull/457

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -2,57 +2,31 @@
...
-        done
-        # And restore the immutable bits
-        chattr +i ${rootmnt}/ostree/deploy/*/deploy/*.0 ${rootmnt}
-        echo "...done"
-        umount $rootmnt
-        umount $saved_rootmnt
-        rmmod brd
-        rm -rf "${tmproot}"
+        cd "${saved_sysroot}"
+        find . -mindepth 1 -maxdepth 1 -exec mv {} /sysroot \;
```
cgwalters: This loses the immutable bits?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap.service
```diff
@@ -0,0 +1,15 @@
...
+# If root is specified, assume rootmap is already configured
+ConditionKernelCommandLine=!root
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+After=initrd-root-fs.target
```
cgwalters: We also need something like:
`Before=initrd-switch-root.target`
I think.  Almost all of our units in the initramfs should be doing that.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap.service
cgwalters: (It turns out we may need it.... https://github.com/coreos/fedora-coreos-config/pull/499/#issuecomment-652398341 )

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
```diff
@@ -0,0 +1,75 @@
...
+inject_bls_kargs() {
+    local tmpf
+
+    # Note we mount /boot but don't unmount since we run with MountFlags=slave
+    mkdir -p /boot
+    mount -o rw ${BOOT_DEV} /boot
+    for f in /boot/loader/entries/*.conf; do
+        tmpf=$(mktemp -p /boot)
+        sed -e "/^options / s|$| $*|" < "${f}" > "${tmpf}"
+        mv "${tmpf}" "${f}"
```
cgwalters: Let's usually add `-T` in our `mv` invocations to prevent it doing the wrong thing if the target is a directory.  I usually add `-f` too just for good measure.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
```diff
@@ -0,0 +1,75 @@
...
+
+get_luks_karg() {
+    local name=$1; shift
+    echo -n " rd.luks.uuid=${name}"
+}
+
+inject_bls_kargs() {
+    local tmpf
+
+    # Note we mount /boot but don't unmount since we run with MountFlags=slave
```
cgwalters: Everything doing this will need to be serialized I think.  There was some discussion related to this around https://github.com/coreos/fedora-coreos-config/pull/346

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
```diff
@@ -0,0 +1,75 @@
...
+get_rootmap_kargs() {
+    local root_dev=$1; shift
+    lsblk --pairs -o NAME,TYPE --inverse "${root_dev}" | while IFS= read -r line; do
+        # note this is a pipeline, so we're in a subprocess
+        eval "${line}"
+        # shellcheck disable=SC2153
+        case "${TYPE}" in
+            raid*) get_raid_karg "${NAME}";;
+            crypt) get_luks_karg "${NAME}";;
+            part|disk) ;;
```
cgwalters: Hmm...how about we call out to something like `/usr/bin/ignition generate-kargs-for ${root_dev}`?

jlebon:
Re. root-on-LUKS, see https://github.com/coreos/ignition/pull/960#issuecomment-653082793.

jlebon:
OK, rebased this and now with https://github.com/coreos/fedora-coreos-config/pull/466, https://github.com/coreos/fedora-coreos-config/pull/184, and https://github.com/coreos/fedora-coreos-config/pull/354 folded in!

Will see how quickly we can get a respin of Clevis with https://github.com/latchset/clevis/pull/211 included. Otherwise, we should be able to hack around it for now by carrying our own version of the path and service units until that happens.

I haven't addressed the comments yet. I'd also like to write a kola test for this before everything goes in. Also need to adapt growpart.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -8,6 +8,9 @@ set -euo pipefail
 # Growpart /, until we can fix Ignition for separate /var
 # (And eventually we want ignition-disks)
 
+echo "nope"
```
cgwalters: Shouldn't the rootfs reprovisioning have also done resizing?  IOW I think we just want to skip growpart if rootfs has been reprovisioned.

jlebon: I think if the rootfs has been moved, we should skip growpart. But if it's been replaced in place, we still want to do that. (In Ignition terms, e.g. moving from xfs to ext4 doesn't involve ignition-disks changing the partition table at all.)

cgwalters: Seems like this would all be cleaner if growpart is part of `ignition-disks` always.  It's weird to provision a new FS, load a bunch of data into it, then grow it immediately.

jlebon: Yeah, I guess that's https://github.com/coreos/ignition/issues/924 and part of what was https://github.com/coreos/fedora-coreos-tracker/issues/45. We probably need a new issue now to track lowering the growpart bit to a base config or something.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-rootfs-detect.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Ignition OSTree: detect rootfs replacement
+DefaultDependencies=false
+After=ignition-fetch.service
```
arithx, line 4: Will this work in cases where we are doing a full offline Ignition config? My understanding of the new process is that the fetch stage only runs if we've marked that we neednet.

jlebon:
This is blocked on https://bugzilla.redhat.com/show_bug.cgi?id=1845210.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-rootfs-detect.service
jlebon: Yeah, systemd respects ordering even if services aren't actually started in the end due to their conditionals. See https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Conditions%20and%20Asserts.

manifest-lock.x86_64.json
```diff
@@ -135,6 +147,9 @@
     "cracklib": {
       "evra": "2.9.6-22.fc32.x86_64"
     },
+    "cracklib-dicts": {
```
bgilbert: If the clevis package maintainer accepts https://src.fedoraproject.org/rpms/clevis/pull-request/7, we can ship a config fragment to disable cracklib dict lookup in clevis and avoid this dependency.

cgwalters:
I checked out this branch and also dropped in a locally built kernel from [this commit](https://src.fedoraproject.org/rpms/kernel/c/e6fc315293820bc53857634588302f951d1442f9?branch=f32), then provided:
```
# Example of reprovisioning / with btrfs
variant: fcos
version: 1.0.0
storage:
  filesystems:
    - device: /dev/disk/by-partlabel/root
      format: btrfs
      wipe_filesystem: true
      label: root
```

And it just worked, rootfs on btrfs was that easy!  (And same for `ext4` etc.)

(I did get a failure from `ignition-ostree-uuid-boot.service` but that's https://github.com/coreos/fedora-coreos-config/pull/503#discussion_r448511369 I believe)

cgwalters:
Another bug I just hit with this is when doing an install without providing an Ignition config:
```
[root@localhost ~]# systemctl status ignition-ostree-rootfs-detect
● ignition-ostree-rootfs-detect.service
     Loaded: not-found (Reason: Unit ignition-ostree-rootfs-detect.service not found.)
     Active: failed (Result: exit-code) since Mon 2020-07-13 15:08:54 UTC; 3min 16s ago
   Main PID: 537 (code=exited, status=1/FAILURE)
        CPU: 2ms

Jul 13 15:08:54 localhost systemd[1]: Starting Ignition OSTree: detect rootfs replacement...
Jul 13 15:08:54 localhost systemd[1]: ignition-ostree-rootfs-detect.service: Main process exited, code=exited, status=1/FAILURE
Jul 13 15:08:54 localhost ignition-ostree-dracut-rootfs[539]: /usr/libexec/ignition-ostree-dracut-rootfs: line 12: /run/ignition.json: No such file or directory
Jul 13 15:08:54 localhost systemd[1]: ignition-ostree-rootfs-detect.service: Failed with result 'exit-code'.
Jul 13 15:08:54 localhost systemd[1]: Failed to start Ignition OSTree: detect rootfs replacement.
```

arithx:
> Another bug I just hit with this is when doing an install without providing an Ignition config:

At least for the no such file part with the new Ignition release the cache config (`/run/ignition.json` by default) should always exist (coreos/ignition#1002)

jlebon:
This kernel update has the patch we need: https://bodhi.fedoraproject.org/updates/FEDORA-2020-c5938abe2d

jlebon:
Status update on this:
- The rootmap injection logic is now implemented as part of `rdcore` in https://github.com/coreos/coreos-installer/pull/304.
- The kernel in FCOS now supports reading SELinux labels.
- Ignition v2.5.0 with experimental LUKS support is in FCOS, but full rootfs-on-LUKS support is waiting on a Clevis release with https://github.com/latchset/clevis/pull/211 and https://github.com/latchset/clevis/pull/217 (though in the worst case, I think we can work around these until that happens).

In light of that last point, what I'd like to do for this PR is to address rootfs-on-LUKS as a follow-up. There's already a lot going on here and I'd like to get it in soon (plus, it's not like we aren't unlocking many other things, such as moving root, changing filesystems, root-on-RAID, etc..).

I dropped the commit which adds the `clevis` packages.

So next steps:
- Get new `rdcore` into FCOS.
- Address review comments.
- Punt on https://github.com/coreos/fedora-coreos-tracker/issues/570: for now, just key off of whether the rootfs was reprovisioned in-place or not (and skip `growpart` in the latter case).

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap.service
jlebon: You're right, we don't need it. Dropped!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
jlebon: This logic is in `rdcore` now.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
jlebon: This logic is in `rdcore` now.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
jlebon: This logic is in `rdcore` now.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap.service
jlebon: Hmm, I don't think that's required. This service is pulled in by the `diskful`, which is pulled in by `complete`, which is pulled in via `initrd.target`. So we wouldn't even get to `initrd.target` before running these services.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
jlebon: Pending more discussions in https://github.com/coreos/fedora-coreos-tracker/issues/570, for now I went with: if the rootfs changed backing device or the partition changed size, we don't automatically growpart. Otherwise, we still grow. That will allow an Ignition config which *only* does e.g. `xfs` -> `btrfs` to still keep resizing.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
jlebon: We're already serialized against the other `/boot` services by virtue of running after Ignition because all those other services happen much earlier in the boot process before Ignition runs in earnest. I'll add some details to https://github.com/coreos/fedora-coreos-tracker/blob/master/internals/README-initramfs.md about this.

jlebon:
OK, dropped WIP on this and marked ready for review! Addressed all review comments and also added reprovisioning tests. Final items needed:
- https://github.com/coreos/coreos-assembler/pull/1651
- `coreos-installer` respin with https://github.com/coreos/coreos-installer/pull/347

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-inject-rootmap
jlebon: Added docs about this in https://github.com/coreos/fedora-coreos-tracker/pull/600.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -8,13 +8,33 @@ set -euo pipefail
...
+if [ "${path}" == /sysroot ] && [ -f "${saved_partstate}" ]; then
+    # We're still ironing out our rootfs automatic growpart story, see e.g.:
+    # https://github.com/coreos/fedora-coreos-tracker/issues/570
+    # https://github.com/coreos/fedora-coreos-tracker/issues/586
+    #
+    # In the context of rootfs reprovisioning, for now our rule is the
+    # following: if the rootfs partition was moved off of the boot disk or it
+    # was resized, then we don't growpart.
+    #
+    # To detect this, we compare the output of `lsblk -o PATH,SIZE` before and
```
cgwalters, line 29: Ah, clever.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -67,4 +87,5 @@ case "${TYPE}" in
     ext4) resize2fs "${src}" ;;
 esac
 
-touch /var/lib/coreos-growpart.stamp
```
cgwalters, line 70: OK, and we can make this change because as part of adapting this code to run in the initramfs, it also now only runs on the ignition boot.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -0,0 +1,39 @@
...
+        # also store the state of the partition
+        lsblk "${rootdisk}" --nodeps --json -b -o PATH,SIZE | jq -c . > "${partstate}"
+        ;;
+    restore)
+        # This one is in a private mount namespace since we're not "offically" mounting
+        mount "${rootdisk}" /sysroot
+        echo "Restoring rootfs from RAM..."
+        cd "${saved_sysroot}"
+        find . -mindepth 1 -maxdepth 1 -exec mv -t /sysroot {} \;
+        chattr +i /sysroot/ostree/deploy/*/deploy/*
```
cgwalters: Optional/followup: Feels like this should move into `rdcore` too.  It's not totally clear to me that `mv` is going to do exactly what we want always (xattrs/selinux), but OTOH this clearly works now.

cgwalters (commented):
This looks good at a high level!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
jlebon: Yup, see also the message for the commit which does this:

```
40ignition-ostree: put growpart stamp file in /run

This stamp file was used to make sure coreos-growpart only ran on the
first boot when it ran in the real root. Nowadays, it runs in the
initramfs as part of `ignition-complete.target` before we even run
`ostree-prepare-root` (which meant we were actually creating the stamp
file in the initrd filesystem).

Having a stamp file is useful though for writing tests. So let's
repurpose the idea and put it in `/run` instead.
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
jlebon: Yeah, I think this script would be a good candidate for Rust-ification. I did sanity-check `mv` semantics for this. (Apart from xattrs, one thing I wasn't 100% sure about was hardlinks, but yes it does preserve those as well.)

Related to this, it's worth mentioning I did initially benchmark different approaches (from what I recall right now: the brd approach we originally had vs `rsync` vs plain `mv`) and there wasn't really a large difference between them. So I opted for the simplest one.

(An interesting one I didn't try is ZRAM which was discussed earlier, and if that one is fast enough and gives us appreciable RAM savings, it's worth playing with it.)

jlebon:
OK, so one thing I just realized is that with `root=UUID=...` always being appended, we lose the `prjquota` option on subsequent boots because now it's systemd that mounts and not us. And more generally, reprovisioning to a different filesystem with custom mount options will need those persisted as well. So I think `rdcore rootmap` needs to also inject `rootflags=`. Working on a patch for that. (Another approach is making it part of `/etc/fstab` since systemd is capable of re-applying mount options after the initial mount, but we're moving away from `/etc/fstab` and also some mount options can only be set on the first mount, including quota-related options on XFS).

(Edit: sorry I was kinda confused on this. Mount options for root can't live in the Ignition config. So in the reprovisioning case, it's up to users to add an appropriate `rootflags` karg.)

jlebon:
> OK, so one thing I just realized is that with root=UUID=... always being appended, we lose the prjquota option on subsequent boots because now it's systemd that mounts and not us.

This is fixed now in https://github.com/coreos/coreos-installer/pull/358 and the latest commit here which adds `coreos-rootflags`. (Definitely a downside of the `rdcore` approach is that it's more work to get fixes in vs just some script in the overlay.)

jlebon:
```
--- PASS: ext.fedora-coreos-config_PR-503-4MMBZDZR33QM3J24KCIOGR3JPDUXCUHF5DYIZB4C3WVNVMA5OEJA.root-reprovision.raid1 (95.30s)
...
--- PASS: ext.fedora-coreos-config_PR-503-4MMBZDZR33QM3J24KCIOGR3JPDUXCUHF5DYIZB4C3WVNVMA5OEJA.root-reprovision.filesystem-only (80.73s)
```

:tada: 

Who wants to give an approving review on this so we can merge it in?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
```diff
@@ -1,20 +1,17 @@
...
   exit 1
 fi
-eval $(blkid -o export ${rootpath})
-mountflags=
-if [ "${TYPE}" == "xfs" ]; then
-  mountflags=prjquota
-fi
-mount -o "${mountflags}" "${rootpath}" /sysroot
+
+mount -o "$(coreos-rootflags)" "${rootpath}" /sysroot
```
ashcrow, line 17: This took a second to realize it was sensing options based on a script but :+1:!

cgwalters (approved):
LGTM

ashcrow (approved):
Great work here! I seriously appreciate the additional commenting through the code :pray:

Conan-Kudo:
@jlebon Thanks for making this happen! 🎉 🙏

sohankunkerkar:
Somehow, this change broke the `Ignition` CI
```
--- FAIL: ext.config.root-reprovision.raid1 (5107.03s)

        harness.go:916: Cluster failed starting machines: machine "4d94e9a8-d77f-41b6-bbe7-d08d1c14f4d7" failed to start: ssh journalctl failed: ssh: handshake failed: read tcp 127.0.0.1:52754->127.0.0.1:42537: read: connection reset by peer

--- FAIL: ext.config.root-reprovision.filesystem-only (5155.02s)

        harness.go:916: Cluster failed starting machines: machine "abe27985-e26a-457a-a4b0-711fc1fbafce" failed to start: ssh journalctl failed: ssh: handshake failed: read tcp 127.0.0.1:33566->127.0.0.1:39207: read: connection reset by peer

```

fulminemizzega:
> (Edit: sorry I was kinda confused on this. Mount options for root can't live in the Ignition config. So in the reprovisioning case, it's up to users to add an appropriate `rootflags` karg.)

Hello,
I think that this should be mentioned in the related documentation section "Reconfiguring the root filesystem". After reading this PR and issue #94 I think I understand why filesystems.mount_options in ignition does not do anything for root fs, but it was not that intuitive (at least to me). In my case, I changed the root fs to btrfs and enabled compression. I also understand this is at least frowned upon, so I guess I got what I deserved.

jlebon:
Hmm, I think if we have magic for reprovisioning root, then we should also just handle `mount_options` too and translate that into `rootflags`. This is similar to how `with_mount_unit` also injects the mount options into the unit.

PR #504, opened 2020-06-30
tree: promote changes from testing-devel at 2ad4cef901d0d69027542b5e3…

travier:
…25c387bf702dd2c


manifest-lock.x86_64.json
```diff
@@ -106,7 +106,7 @@
       "evra": "0.18.1-1.fc32.noarch"
     },
     "container-selinux": {
-      "evra": "2:2.135.0-1.fc32.noarch"
+      "evra": "2:2.132.0-1.fc32.noarch"
```
lucab, line 109: This seems correct according to bodhi. At the same this downgrade feels a bit weird.

lucab:
> error: Couldn't find locked package 'rpm-ostree-2020.3-1.fc32.x86_64' (pkgs matching NEVRA: 0; mismatched checksums: 0)

Hmm, it looks like this has reached F32 stable but it has not been tagged into `coreos-pool` (last version there is `rpm-ostree-2020.2-3.fc32`).

manifest-lock.x86_64.json
travier: The current stable PR/build is doing:
```
-      "evra": "2:2.132.0-1.fc32.noarch"
+      "evra": "2:2.135.0-1.fc32.noarch"
```

jlebon:
> Hmm, it looks like this has reached F32 stable but it has not been tagged into coreos-pool (last version there is rpm-ostree-2020.2-3.fc32).

I tickled the tagger and restarted CI here.

manifest-lock.x86_64.json
travier: Happened in: https://github.com/coreos/fedora-coreos-config/commit/c57085a0e0e75ac6047a33b2a6e4839e97e472d1

jlebon: It's possible that it's a blip due to the DC move. I'm actually seeing the same thing here on my Silverblue (from 32.20200619.0 to 32.20200629.0):

```
Downgraded:
  container-selinux 2:2.135.0-1.fc32 -> 2:2.132.0-1.fc32
```

So... let's just roll with it!

jlebon (approved):

PR #506, opened 2020-07-02
live-generator: A few systemd ordering tweaks

cgwalters:
No actual functional changes I believe.  The main thing
here is explicitly doing the `Wants=systemd-udev-settle.service`
to pull it in (even though it already is from `multipathd.service`,
that might change).

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -88,9 +88,12 @@ EOF
 
 [Unit]
 DefaultDependencies=false
-After=initrd-root-device.target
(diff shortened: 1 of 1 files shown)


cgwalters:
This pulls in some review feedback from https://github.com/coreos/fedora-coreos-config/pull/499
So far I've sanity checked this works with FCOS and RHCOS, though I only just started the "boot 100 times" loop for the latter.

jlebon (approved):
Makes sense to me!

I think before we merge this, we should add some basic `testiso` tests to CI here. Except CI is broken right now. :|

cgwalters:
Rebased :surfer: and we have testiso running here now..

jlebon (approved):

PR #507, opened 2020-07-02
Rename tests/kola/basic to misc-ro

cgwalters:
And add a test case for the NIC naming.  I think we should
make "extend misc-ro" a standard procedure for a lot of recent changes.


miabbott:
We are using this naming/implementation in RHCOS and seems to fit well.

/approve

jlebon (approved):

PR #508, opened 2020-07-07
overlay.d/05core: add preset for CLHM-{issuegen,motdgen}.path

rfairley:
Add a preset for `console-login-helper-messages-issuegen.path` and
`console-login-helper-messages-motdgen.path`, which enable
dynamic regeneration of the issue/MOTD after dropping files into
the runtime directory. E.g.:

```
echo "test" > /run/console-login-helper-messages/issue.d/test.issue
```

will cause `test` to appear in the displayed issue from agetty.
This was added in the v0.18.2 release: https://github.com/coreos/console-login-helper-messages/releases/tag/v0.18.2

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -1,6 +1,9 @@
 # Presets here that eventually should live in the generic fedora presets
 enable coreos-growpart.service
+# console-login-helper-messages - https://github.com/coreos/console-login-helper-messages
+enable console-login-helper-messages-issuegen.path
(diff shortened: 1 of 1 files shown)


rfairley:
Note the corresponding `.service` units need to remain enabled currently, but this may not be required in the next CLHM release, with https://github.com/coreos/console-login-helper-messages/pull/55.

Tested this in FCOS and RHCOS.

This will soon land in FCOS: https://bodhi.fedoraproject.org/updates/FEDORA-2020-98a61ed152. I didn't fast-track this as there should be no harm enabling the .path before the new CLHM RPM lands in FCOS - just making sure this is in place for when it does land.

jlebon (approved):

PR #521, opened 2020-07-16
overlay/15fcos: remove CLHM dependency

sohankunkerkar:
Follow up PR to address https://github.com/coreos/fedora-coreos-config/pull/344#discussion_r455952922
With https://github.com/coreos/console-login-helper-messages/pull/55 change, there's no explicit systemd dependency on CLHM
/cc @rfairley

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
@@ -2,7 +2,6 @@
 # no Ignition config is provided.
 [Unit]
 Description=Check if Ignition config is provided
-Before=console-login-helper-messages-issuegen.service
 # Only perform checks on the first (Ignition) boot as they are
 # mostly useful only on that boot. This ensures systems started
 # before Ignition/Afterburn started logging structured data don't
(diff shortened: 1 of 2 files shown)


rfairley:
Thanks for opening this! We should also do the same for the coreos-check-ssh-keys (in general, any service that previously had the dependency on CLHM can remove the dependency once the next CLHM release is made) https://github.com/coreos/fedora-coreos-config/blob/6bf6e8b634dfa09ca13020e58368e784b9b0c879/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service#L5

Let's also mark this PR as on hold, until https://github.com/coreos/console-login-helper-messages/pull/55 makes it into the next CLHM release (v0.19 in https://github.com/coreos/console-login-helper-messages/releases).

rfairley (commented):
Review in https://github.com/coreos/fedora-coreos-config/pull/521#issuecomment-659804594

kelvinfan001 (approved):
LGTM. These two services should be the only units that had any dependency on CLHM.

sohankunkerkar:
> These two services should be the only units that had any dependency on CLHM.

Yeah

PR #527, opened 2020-07-22
Move conditional network startup from ignition-dracut

bgilbert:
The mechanism for starting network is too distro-specific to live in ignition-dracut.  In addition, PXE rootfs fetch will need to invoke the same machinery.  Add a unit with triggering conditions for the cases that need network, and use `After=ignition-fetch-offline.service` to force that service to run early enough.

Safe to merge before https://github.com/coreos/ignition-dracut/pull/200 lands in the distro.


jlebon (approved):
Thanks, makes sense to me!

PR #532, opened 2020-07-25
overrides: fast-track coreos-installer, afterburn

dustymabe:
- coreos-installer-0.4.0-1.fc32.x86_64
- rust-afterburn-4.4.2-1.fc32.x86_64

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -19,10 +19,10 @@
       "evra": "0.9.0-1.fc32.x86_64"
     },
     "afterburn": {
-      "evra": "4.4.0-1.fc32.x86_64"
+      "evra": "4.4.2-1.fc32.x86_64"
     },
     "afterburn-dracut": {
-      "evra": "4.4.0-1.fc32.x86_64"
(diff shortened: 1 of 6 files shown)


manifest-lock.x86_64.json
```diff
@@ -19,10 +19,10 @@
       "evra": "0.9.0-1.fc32.x86_64"
     },
     "afterburn": {
-      "evra": "4.4.0-1.fc32.x86_64"
+      "evra": "4.4.2-1.fc32.x86_64"
```
bgilbert, line 22: Is this correct?  Usually fast-track PRs don't bump the lockfile, only the overrides.

dustymabe: I decided to bump the lockfile because of the coreos-installer-bootinfra name change from coreos-installer-systemd. Didn't know if it would break CI to have that not correct in the lockfile.

I'm pretty sure it won't hurt to have it updated.

bgilbert (approved):

dustymabe:
rebased - will merge on green CI

PR #545, opened 2020-08-04
live: add karg embed areas in GRUB/ISOLINUX configs

jlebon:
This provides a buffer into which `coreos-installer iso embed-kargs` can
write additional kernel arguments for the live ISO. It must be located
right after the karg line in configuration files.

Note this doesn't touch `zipl.prm`, which is where ISO kernel args are
stored for s390x. Supporting this for s390x would need more
investigation and testing.


jlebon:
See https://github.com/coreos/coreos-assembler/pull/1637 and https://github.com/coreos/coreos-installer/pull/341 for how this will be used.

cgwalters (approved):
Just two comments, otherwise LGTM

bgilbert:
Apologies for the late comment, but would it make sense to use the bootloaders' respective include directives to put these in separate files?

jlebon:
> Apologies for the late comment, but would it make sense to use the bootloaders' respective include directives to put these in separate files?

We need the buffer to be physically located near the kernel cmdline though. The way this works is below the level of the interpreters; `coreos-installer` just moves the newline character rightwards into the buffer as needed.

I think we could do a more "native" thing for GRUB at least. Something like:

```
# kargs_embed_area.cfg
set embedded_kargs="" ######...
```

and then

```
# grub.cfg
source kargs_embed_area.cfg
linux /images/vmlinuz @@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal $embedded_kargs
```

For ISOLINUX though, I don't see a way to play the same trick since it doesn't support variables.

bgilbert:
Ah, right.  Okay, seems fine as is.  :+1:

PR #548, opened 2020-08-06
Ship live PXE artifacts inside live ISO

bgilbert:
Update bootloaders and 20live for https://github.com/coreos/coreos-assembler/pull/1643.

This adds `bsdtar` to the package list; it's a wrapper around libarchive (and can handle cpio files).  We use it instead of `cpio` to unpack the rootfs image, because in ISO boots we need to extract the osmet files, and `cpio` can't do that without reading the entire rootfs image (including the root squashfs) from the boot medium.

This is interlocked with https://github.com/coreos/coreos-assembler/pull/1643 and there is no ratchet.


dustymabe:
```
commit 0a3571ab6c8d73ab02d9bce6c3f758330079182c                                                                                                                                                                                              
Author:     Benjamin Gilbert <bgilbert@redhat.com>                                                                                                                                                                                           
AuthorDate: Thu Aug 6 06:20:35 2020 -0400                                                                                                                                                                                                    
Commit:     Dusty Mabe <dusty@dustymabe.com>                                                                                                                                                                                                 
CommitDate: Thu Aug 6 16:23:36 2020 -0400                                                                                                                                                                                                    

    live: move kernel/initramfs and rename initramfs

    It seems to be conventional in Fedora/RHEL to put the kernel in
    /images/pxeboot/vmlinuz and the initramfs in /images/pxeboot/initrd.img.

    Requires matching coreos-assembler buildextend-installer changes.
```

It feels a bit weird to me to change from `images/vmlinuz` to `images/pxeboot/vmlinuz` here for the ISO boot case. I don't disagree that we should do it, but wondering if we can use that iso9660 hardlink trick to have them both be a hardlink. So

- `images/vmlinuz` hardlink with `images/pxeboot/vmlinuz`
- `images/initramfs.img` hardlink with `images/pxeboot/initrd.img`

live/EFI/fedora/grub.cfg
```diff
@@ -28,6 +28,6 @@ set timeout=5
 
 ### BEGIN /etc/grub.d/10_linux ###
 menuentry 'Fedora CoreOS (Live)' --class fedora --class gnu-linux --class gnu --class os {
-	linux /images/vmlinuz @@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal
-	initrd /images/initramfs.img
+	linux /images/pxeboot/vmlinuz @@KERNEL-ARGS@@ ignition.firstboot ignition.platform.id=metal
+	initrd /images/pxeboot/initrd.img /images/ignition.img
```
dustymabe, line 33: Regarding having a second initrd for ignition.img: Do we think this will cause issues on other architectures by chance? I'm wondering if having a secondary file with the padding is worth it or if just putting the padding in the initramfs (like we were doing in the ISO case in the past) would cause us less trouble, even though we'd be copying around bits unnecessarily.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-livepxe-rootfs.sh
```diff
@@ -34,8 +34,7 @@ elif [[ -n "${rootfs_url}" ]]; then
     # image hash.
     if ! curl --silent --insecure --location --retry 5 "${rootfs_url}" | \
             rdcore stream-hash /etc/coreos-live-want-rootfs | \
-            gunzip -c | \
-            cpio -i -H newc -D / --quiet ; then
+            bsdtar -xf - -C / ; then
```
dustymabe, line 39: since we're not using any advanced features of bsdtar here, does it make more sense to keep using `cpio` in this one case? I assume cpio is already in the initramfs so we wouldn't really be adding a dep.

Otherwise maybe we should add a comment. Sorry, it just looks surprising because I'm not used to using `bsdtar` and it throws me off.

dustymabe: I guess it makes this particular module's dependencies larger, but removing `cpio` from the list probably doesn't actually get it out of the initramfs.

If we keep it maybe just a comment that bsdtar is just extracting a `cpio`.

dustymabe: "Use `bsdtar` instead of `cpio` here because it knows how to extract a cpio archive and we need it in other places for advanced features. Rather than depend on both `cpio` and `bsdtar`, just use `bsdtar` for everything."

It's verbose, I know :(

bgilbert:
> It feels a bit weird to me to change from `images/vmlinuz` to `images/pxeboot/vmlinuz` here for the ISO boot case.

How so?  Is anything likely to depend on the old paths?  Up until now, everything in the ISO image was supposed to be an implementation detail.  :slightly_smiling_face:  And if we add the hardlink, _new_ things may acquire a dependency on the old path.

live/EFI/fedora/grub.cfg
bgilbert: It might, but it seemed worthwhile trying to do this properly.  We can always conditionally pad the initramfs for architectures that need it, or revert the change entirely.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-livepxe-rootfs.sh
bgilbert: > I assume cpio is already in the initramfs so we wouldn't really be adding a dep.

It's not.  The initramfs doesn't usually need to unpack itself.  :stuck_out_tongue_closed_eyes: 

I'll add a comment.

dustymabe:
> > It feels a bit weird to me to change from `images/vmlinuz` to `images/pxeboot/vmlinuz` here for the ISO boot case.
> 
> How so? Is anything likely to depend on the old paths? Up until now, everything in the ISO image was supposed to be an implementation detail. slightly_smiling_face And if we add the hardlink, _new_ things may acquire a dependency on the old path.

Mostly that it just feels wrong. Naming is hard, but it feels awkard that we would move to a less appropriate name for compat even if it is just an internal implementation detail. Also our live PXE initramfs we put on the website is `initramfs.img` and making it `initrd.img` could be confusing too. It would feel more ergonomic if we just supported the old "expected" (i.e. people have hardcoded it for other Red Hat distributions) location as a hardlink or symlink (are symlinks supported?), but didn't make it the preferred location.

Something similar seems to be done for the Fedora 32 server DVD where the same initramfs is in two different locations:

```
$ find /mnt/ | grep initrd.img | xargs md5sum
90b9ce13217024c456c27c78034f4bab  /mnt/images/pxeboot/initrd.img
90b9ce13217024c456c27c78034f4bab  /mnt/isolinux/initrd.img
$
$ cat /mnt/isolinux/isolinux.cfg | grep initrd.img | head -n 1
  append initrd=initrd.img inst.stage2=hd:LABEL=Fedora-S-dvd-x86_64-32 quiet
```


If I were investigating a problem it might throw me down the wrong path because in the ISO case we clearly aren't PXE booting and I would say "wait a minute, why are we using the `pxeboot` initrd!", but it would be the intended image to use.

I know it doesn't really matter and I don't really know why I'm defending this so hard.

dustymabe:
oh wow, but if I look at the grub.cfg (for EFI) on that F32 ISO I see:

```
menuentry 'Install Fedora 32' --class fedora --class gnu-linux --class gnu --class os {                                                                                                                                                      
    linuxefi /images/pxeboot/vmlinuz inst.stage2=hd:LABEL=Fedora-S-dvd-x86_64-32 quiet                                                                                                                                                       
    initrdefi /images/pxeboot/initrd.img                                                                                                                                                                                                     
}
```

So I guess there is a case where it's set up counter to what I would have expected. That certainly weakens my argument.

manifests/fedora-coreos-base.yaml
```diff
@@ -92,6 +92,8 @@ packages:
   - afterburn
   - afterburn-dracut
   - passwd
+  # Dracut modules
```
dustymabe: Optional. Feel free to resolve/close with no questions asked.

```suggestion
  # Needed by 20live dracut module
```

bgilbert:
> it feels awkard that we would move to a less appropriate name for compat even if it is just an internal implementation detail.

I agree that it's weird.  I'd argue the alias is actually worse for two reasons: 1) if there's only one name for a thing, it's obvious which name is being used at any given time, and 2) if there are two names for the thing, people might think there's a difference between them.

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-livepxe-rootfs.sh
```diff
@@ -32,10 +32,11 @@ elif [[ -n "${rootfs_url}" ]]; then
     fi
     # We don't need to verify TLS certificates because we're checking the
     # image hash.
+    # bsdtar can read cpio archives and we already depend on it for
+    # coreos-liveiso-persist-osmet.service.
```
dustymabe: Optional. Feel free to resolve/close with no questions asked.
```suggestion
    # Note, bsdtar can read cpio archives and we already depend on it for
    # coreos-liveiso-persist-osmet.service so use it instead of cpio.
```

overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-liveiso-persist-osmet.service
```diff
@@ -0,0 +1,15 @@
...
+RequiresMountsFor=/run/media/iso
+Before=initrd-switch-root.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/mkdir -p /run/coreos-installer/osmet
+# bsdtar reads cpio archives, and unlike cpio(1L), knows how to seek over
+# members it isn't reading
+ExecStart=/usr/bin/bsdtar -x -C /run/coreos-installer/osmet -f /run/media/iso/images/pxeboot/rootfs.img *.osmet
```
dustymabe, line 15: My understanding is that using `bsdtar` here saves us the memory overhead of unpacking the entire archive. Is that correct?

bgilbert: It saves I/O.  On my system, `cpio` took something like 20 seconds to do the unpack at boot.  And of course it'd be much worse with an actual CD.

dustymabe:
> > it feels awkard that we would move to a less appropriate name for compat even if it is just an internal implementation detail.
> 
> I agree that it's weird. I'd argue the alias is actually worse for two reasons: 1) if there's only one name for a thing, it's obvious which name is being used at any given time, and 2) if there are two names for the thing, people might think there's a difference between them.

Yeah, I'll concede. I think the only point that still really sticks is the difference between `initrd.img` and `fedora-coreos-32.20200804.dev.1-live-initramfs.x86_64.img`, but I think it doesn't matter much either.

dustymabe (approved):
LGTM, only few optional comments left. Nice work @bgilbert

live/EFI/fedora/grub.cfg
jlebon: From my research while hacking on https://github.com/coreos/rpm-ostree/pull/2170, AFAIK the only arch multiple initrds breaks on is s390x because it doesn't use GRUB and zipl doesn't support it: https://github.com/ibm-s390-tools/s390-tools/issues/49.

However, we already don't support Ignition config injection on s390x ISOs today, because it works differently there. So this is furthering the drift, and making it less likely that we'll get back to feature parity there. OTOH, I don't think support for Ignition config injection is a "make or break" feature, and there's always ways to work around it (similarly for https://github.com/coreos/rpm-ostree/pull/2170, on s390x users can just keep regenerating the initramfs). (Or if there's strong enough motivation down the road, maybe we could try to push for https://github.com/ibm-s390-tools/s390-tools/issues/49 again.)

overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
```diff
@@ -116,9 +111,12 @@ Before=initrd-root-fs.target
 RequiresMountsFor=/run/media/iso
 
 [Mount]
-What=/run/media/iso/root.squashfs
+What=/run/media/iso/images/pxeboot/rootfs.img
 Where=/sysroot
 Type=squashfs
+# Offset of the squashfs within the rootfs cpio.  Assumes newc format
+# and that a file named "root.squashfs" is the first member.
```
jlebon: Maybe also e.g.:

> This is ensured by coreos-assembler at build time.

jlebon (approved):
One optional comment, but LGTM as is! Needs rebase on top of #545.

PR #550, opened 2020-08-07
release-notes.d: initial directory for release note yaml snippets

zonggen:
This directory stores the Fedora CoreOS release note yaml snippets which will be picked up by `fedora-coreos-releng-automation/coreos-release-note-generator` [1] script to produce the final `release-notes.yaml` for a release.

Release notes will be organized according to the origin of the change, i.e. the project name. Otherwise miscellaneous changes will be stored under `miscellaneous`. Therefore, the schema of each yaml snippet is designed as follows:

```yaml
# Each yaml file consists of a list of dictionaries which looks like this
- component (required): [Custom Project Name] | miscellaneous
  subject (required): xxx
  body (optional): xxxxx
```
[1] https://github.com/coreos/fedora-coreos-releng-automation

Related: https://github.com/coreos/fedora-coreos-tracker/issues/194
Signed-off-by: Allen Bai <abai@redhat.com>

### DIFF
--- a/release-notes.d/README.md
+++ b/release-notes.d/README.md
@@ -1,0 +1,32 @@
+# Fedora CoreOS Release Note
+
+This directory stores the Fedora CoreOS release note yaml snippets which will be picked up by `fedora-coreos-releng-automation/coreos-release-note-generator` [1] script to produce `release-notes.yaml` with the corresponding latest release ID for the target stream.
+
+Release notes will be organized according to the origin of the change, i.e. the project name. Otherwise miscellaneous changes will be stored under `miscellaneous`. Therefore, the schema of each yaml snippet is designed as follows:
+
+```yaml
+# Each yaml file consists of a list of dictionaries which looks like this
+- component (required): [Custom Project Name] | miscellaneous
+  subject (required): xxx
+  body (optional): xxxxx
+```
+
+An example `release-notes.yaml`:
+```
+- 32.20200715.3.0:
+    coreos-assembler:
+      - subject: add a new sub-command that automates xxx
+      - subject: fix a bug that result in https://github.com/coreos/fedora-coreos-tracker/issues/xxx
+    miscellaneous:
+      - subject: introduce a new config file to facilitate xxx workflow
+        body: the config file as described in https://github.com/coreos/fedora-coreos-tracker/issues/xxx helps users to monitor xxx
+- 32.20200706.3.0:
+    afterburn:
+      - subject: add support for platform xxx
+- 32.20200620.1.0:
+    ignition:
+      - subject: fix a minor issue https://github.com/coreos/fedora-coreos-tracker/issues/xxx
+...
+```
+
+[1] https://github.com/coreos/fedora-coreos-releng-automation
--- a/release-notes.d/example.yaml
+++ b/release-notes.d/example.yaml
@@ -1,0 +1,14 @@
+# Each list item should be a dictionary with structure:
+# - component (required): [Custom Project Name] | miscellaneous
+#   subject (required): xxx
+#   body (optional): xxx
+#
+# For example:
+# - component: coreos-installer
+#   subject: Retired the xxx sub-command
+#   body: The sub-command was constantly causing xxx failure (https://github.com/coreos/fedora-coreos-tracker/issues/ISSUE_NUMBER)
+# - component: miscellaneous
+#   subject: Added package abc as per (https://github.com/coreos/fedora-coreos-tracker/issues/ISSUE_NUMBER)
+- component: ""
+  subject: ""
+  body: ""


release-notes.d/README
```diff
@@ -0,0 +1,21 @@
+# Fedora CoreOS Release Note
+
+This directory stores the Fedora CoreOS release note yaml snippets which will be picked up by
+`fedora-coreos-releng-automation/coreos-release-note-generator` [1] script to produce the final
+`release-notes.yaml` for a release. After `release-notes.yaml` has been generated / updated,
+all snippets under `release-notes.d` are expected to be cleaned up for future release notes.
+
+Please name the yaml file with the corresponding project name, e.g. afterburn.yaml, since the project name
```
dustymabe: So if we use names like `afterburn.yaml` it increases the likelihood of merge conflicts, which is what we're trying to avoid by using a release-notes.d directory to begin with. We should be able to support completely random file names here (as long as they end with `.yaml`), though I see what you're trying to do by trying to group things together. How about instead of a filename we add another entry into the yaml:

```
# Each list item should be a dictionary with structure:
# - component (required): (i.e. afterburn or ignition or coreos-installer. if generic enough, just use `general`)
#   subject (required): removed package xyz
#   body (optional): xyz introduced bug which is related to (https://example.com/bug3).
```

and then we group them based on `component`.

zonggen: Like the idea of a `component` better than enforcing a naming convention, like you said I think it makes sense to organize things together based on project and display them accordingly..
Will update with the suggested method :+1:

zonggen:
Updated with a `component` field instead of enforcing naming convention. Currently thought on `component` is to support four types of components: `[Custom Project Name] | general | bugfix | security`.

dustymabe:
> Updated with a `component` field instead of enforcing naming convention. Currently thought on `component` is to support four types of components: `[Custom Project Name] | general | bugfix | security`.

I think there might be too much collision there. For example, what if I have a significant bugfix in Ignition? I think we have two options here:

1. add a `type:` field with options like `enhancement | bugfix | security`.
2. just ignore `type:` for now and add it later

For `component:` maybe the word `miscellaneous` would be better than `general`. So our options would be `component_name | miscellaneous`.

zonggen:
Updated the `component` to allow either `[Project Name]` or `miscellaneous`.

release-notes.d/README
```diff
@@ -0,0 +1,31 @@
...
+```
+
+The final rendered version is expected to be:
+
+```
+32.20200715.3.0:
+ - `coreos-assembler`:
+    add a new sub-command that automates xxx (details)
+ - Miscellaneous:
+    introduce a new config file to facilitate xxx workflow (details)
```
dustymabe: we could probably just keep these simple and say it would be lower case with no '`' characters

```suggestion
 - coreos-assembler:
    add a new sub-command that automates xxx (details)
 - miscellaneous:
    introduce a new config file to facilitate xxx workflow (details)
```

dustymabe (approved):
LGTM - nice work @zonggen

release-notes.d/README
zonggen: Sure, SGTM :+1:

dustymabe:
Maybe let's squash the two commits into one?

zonggen:
@dustymabe I formatted the doc a little bit and added an example for the intermediate `release-notes.yaml` file that consists a `next-release` field (https://github.com/coreos/fedora-coreos-config/pull/550/files#diff-501cdb98b3adaa694d31c5775aa02a46R14-R29), could you take another look?

release-notes.d/README
```diff
@@ -0,0 +1,50 @@
...
+    miscellaneous:
+      - subject: introduce a new config file to facilitate xxx workflow (details)
+        body: the config file as described in https://github.com/coreos/fedora-coreos-tracker/issues/xxx helps users to monitor xxx
+- 32.20200706.3.0:
+    afterburn:
+      - subject: add support for platform xxx
+- 32.20200620.1.0:
+    ignition:
+      - subject: fix a minor issue https://github.com/coreos/fedora-coreos-tracker/issues/xxx
+```
```
dustymabe: I'm not sure we need this "intermediate" example for a few reasons:

- I'm not sure we need `next-release` at all. See https://github.com/coreos/fedora-coreos-pipeline/pull/261#issuecomment-675104281
- Even if we do end up needing `next-release` the intermediate state example seems like something we should put in the README.md in https://github.com/coreos/fedora-coreos-releng-automation/pull/119

So maybe we don't need the intermediate state, but even if we do we probably don't need to document it here.

dustymabe (commented):
Can we rename `README` to `README.md` so GitHub will format the file?

zonggen:
Simplified `README.md` and squashed commits.

dustymabe (dismissed):
LGTM - I think we'll need to also do this for `next-devel`?

dustymabe:
Will having the `example.yaml` file make things harder in the implementation? Were you planning to just ignore that file in the code? I like having it there so people can copy and paste, but want to make sure we account for it.

zonggen:
> Were you planning to just ignore that file in the code?

@dustymabe Yes, I was writing the releng code such that:
 - empty files are ignored
 - note item with empty component / subject is ignored (since they are required fields)
 - empty body is removed to clean up the final release note

zonggen:
Will open another PR to `next-devel`, thanks for the reminder!

EDIT: Opened https://github.com/coreos/fedora-coreos-config/pull/570

PR #551, opened 2020-08-07
20live: add missing coreos-livepxe-rootfs dep to persist-osmet

bgilbert:
If we're downloading the rootfs ourselves, and we don't wait for that to finish before copying osmet files, coreos-installer will fall back to installing FCOS stable.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-live-persist-osmet.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-live-persist-osmet.service
@@ -2,6 +2,8 @@
 Description=Persist osmet files
 DefaultDependencies=false
 ConditionPathExists=/run/ostree-live
+# Downloads and unpacks the osmet files if not already appended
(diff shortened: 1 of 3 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-live-persist-osmet.service
```diff
@@ -2,6 +2,8 @@
 Description=Persist osmet files
 DefaultDependencies=false
 ConditionPathExists=/run/ostree-live
+# Downloads and unpacks the osmet files if not already appended
+After=coreos-livepxe-rootfs.service
```
cgwalters, line 6: Right, makes sense.   This unit also had no dependencies at all before, which is almost always a huge red flag.

cgwalters (approved):
LGTM

bgilbert:
xref: https://bugzilla.redhat.com/show_bug.cgi?id=1867091

PR #556, opened 2020-08-08
kola-denylist: fix terminology for better wording

dustymabe:


kola-denylist.yaml
```diff
@@ -1,5 +1,5 @@
 # This file documents currently known-to-fail kola tests. It is consumed by
-# coreos-assembler to automatically blacklist some tests. For more information,
+# coreos-assembler to automatically deny some tests. For more information,
```
bgilbert: Maybe "skip" would be clearer?

dustymabe: done

bgilbert (approved):

dustymabe:
skipping CI since this is a comment change.

PR #564, opened 2020-08-14
manifests: Addition of nvme-cli

nasirhm:
It adds `nvme-cli` to provide an interface that allows host software utilities to communicate with nvme disks.
 
The missing `nvme` binary was brought up over in a tracker issue [1] and also discussed during the 08/12 meeting [2] where there was agreement that we should include this utility in the host.
 
[1] https://github.com/coreos/fedora-coreos-tracker/issues/601
[2] https://lists.fedoraproject.org/archives/list/coreos@lists.fedoraproject.org/message/5E66QVX7F6HRZXMY4GEWQW6SU3GUOQ3G/

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -157,6 +157,8 @@ packages:
   - kbd
   # Parsing/Interacting with JSON data
   - jq
+  # nvme-cli for managing nvme disks
+  - nvme-cli
 
 # This thing is crying out to be pulled into systemd, but that hasn't happened
 # yet.  Also we may want to add to rpm-ostree something like arch negation;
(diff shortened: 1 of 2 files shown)


manifests/fedora-coreos-base.yaml
```diff
@@ -157,6 +157,8 @@ packages:
   - kbd
   # Parsing/Interacting with JSON data
   - jq
+  # nvme-cli for udev rules
```
dustymabe: Since we're adding it independent of the udev rules let's just make the comment more generic. Something like:

```suggestion
  # nvme-cli for managing nvme disks
```

dustymabe (commented):
Nice work @nasirhm - let's squash those two commits into one if possible.

manifests/fedora-coreos-base.yaml
nasirhm: Sure, Thank you for the review @dustymabe .

dustymabe (approved):
LGTM

PR #567, opened 2020-08-17
coreos-gpt-setup: Exit if PTUUID is not set

Prashanth684:
On s390x zVM with DASD disks, coreos-gpt-setup fails because there are no partition table attributes available for DASD disks.
Bail out if PTUUID variable is unset.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-gpt-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-gpt-setup.sh
@@ -19,9 +19,12 @@ else
     eval $(lsblk --output PTUUID,PKNAME --pairs --paths --nodeps "$1")
 fi
 
-# s390x DASD disks don't have PTUUID and any of the other traditional partition
-# table attributes of GPT disks
-([ -z ${PTUUID:-} ] || [ "$PTUUID" != "$UNINITIALIZED_GUID" ]) && exit 0
+# Skip in the following two cases:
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-gpt-setup.sh
```diff
@@ -19,7 +19,9 @@ else
     eval $(lsblk --output PTUUID,PKNAME --pairs --paths --nodeps "$1")
 fi
 
-[ "$PTUUID" != "$UNINITIALIZED_GUID" ] && exit 0
+# s390x DASD disks don't have PTUUID and any of the other traditional partition
+# table attributes of GPT disks
+([ -z ${PTUUID:-} ] || [ "$PTUUID" != "$UNINITIALIZED_GUID" ]) && exit 0
```
dustymabe: Would this also work (untested)?

```suggestion
# Skip in the following two cases:
#  - The PTUUID is != $UNINITIALIZED_GUID
#  - The PTUUID is empty. This happens on s390x where DASD disks don't
#    have PTUUID or any of the other traditional partition table
#    attributes of GPT disks.
[ "${PTUUID:-}" != "$UNINITIALIZED_GUID" ] && exit 0
```

Prashanth684: yup..it works and is cleaner..thanks!!

dustymabe (approved):
LGTM

PR #543, opened 2020-08-03
manifest: chmod files to unset the writable permissions

sohankunkerkar:
Not mergeable until #412 gets merged
In order to make chrony use NTP settings from DHCP (https://github.com/coreos/fedora-coreos-config/pull/412), we'd need to chmod the `overlay.d/21dhcp-chrony/etc/NetworkManager/dispatcher.d/20-coreos-chrony-dhcp`, `overlay.d/21dhcp-chrony/usr/libexec/coreos-chrony-helper`, and `overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony` to unset the writable permissions. Git tracks only the executable bit of the permissions for the user who owns that file which means that all permissions (execute, write and read) for other users in the file’s group (g) and other users not in the files group (o) are not tracked.

PR #569, opened 2020-08-20
tests/kola/chrony: add test to verify the DHCP propagation of NTP servers

sohankunkerkar:
Writing this test to verify coreos/fedora-coreos-config#412 to enable the DHCP propagation of NTP servers.
CI won't pass because it needs #412 and #543

### DIFF
--- a/tests/kola/chrony/dhcp-propagation
+++ b/tests/kola/chrony/dhcp-propagation
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# This script creates two veth interfaces i.e. one for the host machine 
+# and other for the container(dnsmasq server). This setup will be helpful
+# to verify the DHCP propagation of NTP servers. This will also avoid any 
+# regression that might cause in RHCOS or FCOS when the upstream changes 
+# come down and obsolete the temporary work (https://github.com/coreos/fedora-coreos-config/pull/412)
+
+set -xeuo pipefail
+
+# kola: { "tags": "needs-internet" }
+
+test_setup() {
+
+    # This is needed to run a container with systemd
+    setsebool container_manage_cgroup 1
+
+    # create a network namespace
+    ip netns add container
+
+    # create veth pair and assign a namespace to veth-container
+    ip link add veth-host type veth peer name veth-container
+    ip link set veth-container netns container
+
+    # assign an IP address to the `veth-container` interface and bring it up
+    ip netns exec container ip address add 172.16.0.1/24 dev veth-container
+    ip netns exec container ip link set veth-container up
+
+    # create a static ethernet connection for the `veth-host`
+    nmcli dev set veth-host managed yes
+    ip link set veth-host up
+
+    # run podman commands to set up dnsmasq server
+    pushd $(mktemp -d)
+    NTPHOSTIP=$(getent hosts time-c-g.nist.gov | cut -d ' ' -f 1)
+    cat <<EOF >Dockerfile
+FROM registry.fedoraproject.org/fedora:32
+RUN dnf -y install systemd dnsmasq iproute iputils \
+&& dnf clean all \
+&& systemctl enable dnsmasq
+RUN echo -e 'dhcp-range=172.16.0.10,172.16.0.20,12h\nbind-interfaces\ninterface=veth-container\ndhcp-option=option:ntp-server,$NTPHOSTIP' > /etc/dnsmasq.d/dhcp
+CMD [ "/sbin/init" ]
+EOF
+    podman build -t dnsmasq .
+    popd
+    podman run -d --rm --name dnsmasq --privileged --network ns:/var/run/netns/container dnsmasq
+
+}
+
+main() {
+
+    test_setup
+
+    # check for NTP server information
+    # Name: time-c-g.nist.gov  IP address: 129.6.15.30
+    NTPSERVER=""
+    retries=300
+    while [[ $retries -gt 0 && ! $NTPSERVER =~ "time-c-g.nist.gov" ]]; do
+         NTPSERVER=$(chronyc sources)
+         if [[ $NTPSERVER != *"time-c-g.nist.gov"* ]]; then
+             echo "waiting for ntp server to appear"
+             sleep 1
+             retries=$((retries - 1))
+         fi
+    done
+
+    if [ $retries -eq  0 ]; then
+        echo "propagation of ntp server information via dhcp failed"
+        exit 1
+    fi
+}
+
+main
--- a/tests/kola/chrony/coreos-platform-chrony-generator
+++ b/tests/kola/chrony/coreos-platform-chrony-generator
@@ -2,6 +2,8 @@
 set -euo pipefail
 # Test the coreos-platform-chrony generator.
 
+# kola: { "platforms": "aws,azure,gcp" }
+
 cd $(mktemp -d)
 
 platform="$(grep -Eo ' ignition.platform.id=[a-z]+' /proc/cmdline | cut -f 2 -d =)"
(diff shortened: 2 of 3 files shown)


tests/kola/chrony/chronydhcp
```diff
@@ -0,0 +1,71 @@
...
+
+# This script creates two veth interfaces i.e. one for the host machine 
+# and other for the container(dnsmasq server). This setup will be helpful
+# to verify the DHCP propagation of NTP servers. This will also avoid any 
+# regression that might cause in RHCOS or FCOS when the upstream changes 
+# come down and obsolete the temporary work (https://github.com/coreos/fedora-coreos-config/pull/412)
+
+set -euo pipefail
+
+# kola: { "architectures": "x86_64", "platforms": "qemu", "tags": "needs-internet" }
```
cgwalters: This isn't specific to x86_64 right?  I think we could also in theory run it on platforms other than qemu.

dustymabe:
```
2020-08-20T20:52:57Z kola: Found non-executable file with shebang: chronydhcp
```

Thanks @kelvinfan001 for writing that helpful check. 

@sohankunkerkar you need to make the file executable.

dustymabe:
regarding the files: since we're re-using the `chrony` directory, let's move the `kola.json` content (which limits the platforms) into the other file and also rename it to something more appropriate. Maybe `coreos-platform-chrony-generator`?

If we do this I would do it in a separate commit in this PR before adding your file, which maybe we should just rename to `dhcp-propagation`, so the name of the test would be `ext.config.chrony.dhcp-propagation`.

tests/kola/chrony/chronydhcp
```diff
@@ -0,0 +1,72 @@
...
+# assign an IP address to the `veth-container` interface and bring it up
+ip netns exec container ip address add 172.16.0.1/24 dev veth-container
+ip netns exec container ip link set veth-container up
+
+# create a static ethernet connection for the `veth-host`
+nmcli dev set veth-host managed yes
+ip link set veth-host up
+
+# run podman commands to set up dnsmasq server
+cd $(mktemp -d)
```
dustymabe:
```suggestion
pushd $(mktemp -d)
```

tests/kola/chrony/chronydhcp
```diff
@@ -0,0 +1,72 @@
...
+cat <<EOF >Dockerfile
+FROM registry.fedoraproject.org/fedora:32
+RUN dnf -y install systemd dnsmasq iproute iputils \
+&& dnf clean all \
+&& systemctl enable dnsmasq
+RUN echo -e 'dhcp-range=172.16.0.10,172.16.0.20,12h\nbind-interfaces\ninterface=veth-container\ndhcp-option=option:ntp-server,129.6.15.30' > /etc/dnsmasq.d/dhcp
+CMD [ "/sbin/init" ]
+EOF
+podman build -t dnsmasq .
+cd -
```
dustymabe:
```suggestion
popd
```

tests/kola/chrony/chronydhcp
```diff
@@ -0,0 +1,72 @@
...
+podman run -d --rm --name dnsmasq --privileged --network ns:/var/run/netns/container dnsmasq
+
+}
+
+main(){
```
dustymabe: Indent the content inside the `main()` function?

tests/kola/chrony/chronydhcp
```diff
@@ -0,0 +1,72 @@
...
+ip link set veth-host up
+
+# run podman commands to set up dnsmasq server
+cd $(mktemp -d)
+cat <<EOF >Dockerfile
+FROM registry.fedoraproject.org/fedora:32
+RUN dnf -y install systemd dnsmasq iproute iputils \
+&& dnf clean all \
+&& systemctl enable dnsmasq
+RUN echo -e 'dhcp-range=172.16.0.10,172.16.0.20,12h\nbind-interfaces\ninterface=veth-container\ndhcp-option=option:ntp-server,129.6.15.30' > /etc/dnsmasq.d/dhcp
```
dustymabe: I wonder if we should auto detect the ip address to use here in case it changes.

```
$ getent hosts time-c-g.nist.gov
129.6.15.30     time-c-g.nist.gov
```

```
$ getent hosts time-c-g.nist.gov | cut -d ' ' -f 1
129.6.15.30
```

tests/kola/chrony/chronydhcp
```diff
@@ -0,0 +1,72 @@
...
+while [[ $retries -gt 0 && ! $NTPSERVER =~ "time-c-g.nist.gov" ]]; do
+  NTPSERVER=$(chronyc sources)
+  if [[ $NTPSERVER != *"time-c-g.nist.gov"* ]]; then
+      echo "waiting for ntp server to appear"
+      sleep 1
+      retries=$((retries - 1))
+  fi
+done
+
+ if [ $retries == 0 ]; then
```
dustymabe:
```suggestion
 if [ $retries -eq 0 ]; then
```

Using `-eq` for arithmetic operations.

dustymabe:
Nice work Sohan! One final thing and then I think this will be ready for merge!

Can you also indent the code within the `test_setup()` function. The only part of that function that won't be indented is the heredoc; the `cat <<EOF >Dockerfile` would be indented, but all other lines up to and including the closing `EOF` would not be indented.

sohankunkerkar:
> the `cat <<EOF >Dockerfile` would be indented, but all other lines up to and including the closing `EOF` would not be indented.

Yeah, I was having the same issue yesterday that's why I didn't indent the entire `test_setup()` function. How about know? 
Does that look any better?

tests/kola/chrony/dhcp-propagation
```diff
@@ -0,0 +1,73 @@
...
+    pushd $(mktemp -d)
+    NTPHOSTIP=$(getent hosts time-c-g.nist.gov | cut -d ' ' -f 1)
+    cat <<EOF >Dockerfile
+    FROM registry.fedoraproject.org/fedora:32
+    RUN dnf -y install systemd dnsmasq iproute iputils \
+    && dnf clean all \
+    && systemctl enable dnsmasq
+    RUN echo -e 'dhcp-range=172.16.0.10,172.16.0.20,12h\nbind-interfaces\ninterface=veth-container\ndhcp-option=option:ntp-server,$NTPHOSTIP' > /etc/dnsmasq.d/dhcp
+    CMD [ "/sbin/init" ]
+EOF
```
dustymabe: The code lines after the `<<EOF` need to stay non-indented.

```suggestion
    cat <<EOF >Dockerfile
FROM registry.fedoraproject.org/fedora:32
RUN dnf -y install systemd dnsmasq iproute iputils \
&& dnf clean all \
&& systemctl enable dnsmasq
RUN echo -e 'dhcp-range=172.16.0.10,172.16.0.20,12h\nbind-interfaces\ninterface=veth-container\ndhcp-option=option:ntp-server,$NTPHOSTIP' > /etc/dnsmasq.d/dhcp
CMD [ "/sbin/init" ]
EOF
```

dustymabe (approved):
LGTM

The CI tests are failing (this is a good thing). So I'll focus on review for #412 and #543 and then we'll circle back here for merge.

dustymabe:
ok - re-kicked CI now that #412 and #543 are in.

dustymabe:
From CI:

```
--- PASS: ext.fedora-coreos-config_PR-569-SVYAZ36FP6NF4RKSEFPAEHHD4SF2WAUCIV2BDXEVQBDY5X7I3W3A.chrony.dhcp-propagation (132.24s)
```

🎉🎉🎉

PR #572, opened 2020-08-20
liveiso-autologin-generator: fix up motd formatting

dustymabe:
Fix formatting for motd line break. We want it to show in the output
so we need to escape it. Followup to 24adb7a.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
@@ -58,7 +58,7 @@ from memory, making it a good candidate for hardware discovery and
 installing persistently to disk. Here is an example of running an install
 to disk via coreos-installer:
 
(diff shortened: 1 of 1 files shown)


dustymabe:
This is what I see currently:

```
Last login: Thu Aug 20 22:03:49 on tty1

###########################################################################
Welcome to the CoreOS live environment. This system is running completely
from memory, making it a good candidate for hardware discovery and
installing persistently to disk. Here is an example of running an install
to disk via coreos-installer:

sudo coreos-installer install /dev/sda     --ignition-url https://example.com/example.ign

You may configure networking via 'sudo nmcli' or 'sudo nmtui' and have
that configuration persist into the installed system by passing the
'--copy-network' argument to 'coreos-installer install'. Please run
'coreos-installer install --help' for more information on the possible
install options.
###########################################################################

[core@localhost ~]$
```

arithx (approved):

sohankunkerkar (approved):

PR #573, opened 2020-08-21
misc-ro: Validate that we have the immutable bit on /

cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1867601 for motivation;
basically there's a likely ostree bug affecting s390x, and this
would have caught it.

(Sadly I don't think ostree's upstream suite has this, but
 adding it there wouldn't be useful until we start actually
 running upstream suites on our main releases and/or have
 upstream CI on !x86_64)

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -34,6 +34,11 @@ if test -w /sysroot; then
 fi
 ok sysroot ro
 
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
Seems reasonable. And look how easy it is to add a test. Nice work everyone who helped put that plumbing in place.

jlebon (approved):

PR #576, opened 2020-08-24
[testing] tree: promote changes from testing-devel at 16da480786222d407a33d944952b5c8e6312028b

sohankunkerkar:

PR #578, opened 2020-08-24
[next] tree: promote changes from next-devel at 6854e9f03405333e73c26304ac0ddf2708bdd6e5

sohankunkerkar:


jlebon:
Let's hold this; I think something went wrong with config-bot which should've resulted in the same diff between this one and https://github.com/coreos/fedora-coreos-config/pull/576. Digging.

jlebon:
https://github.com/coreos/fedora-coreos-releng-automation/pull/122

Once this is merged, it'll restart the bot, and the `next-devel` diff should match https://github.com/coreos/fedora-coreos-config/pull/576.

jlebon (approved):

PR #579, opened 2020-08-25
tests/kola: restrict DHCP propagation test to qemu

jlebon:
On cloud platforms, we have platform-specific code which takes
precedence of NTP-via-DHCP so this test won't work (specifically because
we set `PEERNTP=no`).

If there are other platforms which natively use DHCP for NTP setting, we
should enhance this test to handle those (so that e.g. we don't have to
set up dnsmasq at all).

### DIFF
--- a/tests/kola/chrony/dhcp-propagation
+++ b/tests/kola/chrony/dhcp-propagation
@@ -8,7 +8,9 @@
 
 set -xeuo pipefail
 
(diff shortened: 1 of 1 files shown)


sohankunkerkar (approved):
LGTM

cgwalters (approved):
Sohan and I talked about merging the two tests, but this is a much easier fix for now.

PR #580, opened 2020-08-25
[testing-devel] overrides: drop graduated overrides

sohankunkerkar:


cgwalters (approved):
Tearfully watching them walk away with their graduation hats on...

PR #581, opened 2020-08-25
[next] overrides: drop graduated overrides

sohankunkerkar:


cgwalters (approved):
LGTM

jlebon:
Note this should've been opened against `next-devel`, not `next`. Had a chat with @sohankunkerkar, and we'll just let it roll for now instead of reverting. The base lockfiles already have the same matching packages, so it doesn't actually affect the package set in case we do need to respin `next` before the next promotion. Sohan is working on a PR for `next-devel`.

sohankunkerkar:
https://github.com/coreos/fedora-coreos-config/pull/582

PR #585, opened 2020-08-27
99emergency-timeout/virtio-dump: sync before dumping

jlebon:
When systemd activates us as a result of switching to
`emergency.target`, journald might still be collecting and writing back
messages from various services. If we run too early, we'll be missing
out on those.

Let's explicitly make journald sync its logs before dumping the whole
journal.


jlebon:
This is a much cleaner version than https://github.com/coreos/ignition-dracut/pull/201.

cgwalters (approved):

PR #586, opened 2020-08-28
tests/kola: add test for separate /var mount

jlebon:
We regressed on this a few times. Let's add a test for it.

### DIFF
--- a/tests/kola/var-mount/test.sh
+++ b/tests/kola/var-mount/test.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# restrict to qemu for now because the primary disk path is platform-dependent
+# kola: {"platforms": "qemu"}
+
+src=$(findmnt -nvr /var -o SOURCE)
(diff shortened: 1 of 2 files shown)


tests/kola/var-mount/test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# restrict to qemu for now because the primary disk path is platform-dependent
```
cgwalters, line 4: Not strictly related to this PR but I think we really need to introduce something like `/dev/disk/coreos-root` or so.

cgwalters (approved):
LGTM

jlebon:
CI here will fail until https://github.com/ostreedev/ostree/pull/2186 reaches FCOS.

cgwalters:
I rebased :surfing_man: this now that we have https://github.com/coreos/fedora-coreos-config/pull/596

cgwalters:
OK hmm that last CI run did successfully get the latest ostree and the new test passed; the upgrade run just flaked downloading an image.

cgwalters:
Hooray, green here!  Let's make sure we don't break this again.

PR #588, opened 2020-08-28
tests/kola: add rootless-systemd podman test

dustymabe:
Let's try to make sure we don't regress on running systemd in a rootless
container. Prevents seeing issues like:
https://github.com/containers/podman/issues/7441

### DIFF
--- a/tests/kola/podman/rootless-systemd
+++ b/tests/kola/podman/rootless-systemd
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# This script runs a rootless podman container with systemd inside
+# that brings up httpd. Tests that rootless+systemd works. See issue:
+# https://github.com/containers/podman/issues/7441
+#
+# If it gets easy to change the kargs in the future we should try this
+# both on cgroups v1 and cgroups v2.
+
+set -euxo pipefail
+
+# Just run on QEMU for now. If this test works in one place it should
+# work everywhere.
+# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv" }
+
+# Script to be executed as the `core` user to build and
+# run a rootless podman container that uses systemd inside.
+runascoreuserscript='
+#!/bin/bash
+set -euxo pipefail
+
+# build systemd container (runs httpd via systemd)
+# NOTE: we may want to not build this in the future and run
+#       from a common testutils container. See
+#       https://github.com/coreos/coreos-assembler/issues/1645
+cd $(mktemp -d)
+cat <<EOF > Containerfile
+FROM registry.fedoraproject.org/fedora:32
+RUN dnf -y install systemd httpd \
+&& dnf clean all \
+&& systemctl enable httpd
+ENTRYPOINT [ "/sbin/init" ]
+EOF
+podman build -t httpd .
+
+# Run systemd container
+# Pass `-t` so systemd will print logs so `podman logs` will work
+podman run -t -d --name httpd --publish 8080:80 httpd
+'
+
+runascoreuser() {
+    # NOTE: If we don't use `| cat` the output won't get copied
+    # to our unit and won't show up in the `systemctl status` output
+    # of the ext test.
+    sudo -u core "$@" | cat
+}
+
+main() {
+
+    # Execute script as the core user to build/run the container
+    echo "$runascoreuserscript" > /tmp/runascoreuserscript
+    chmod +x /tmp/runascoreuserscript
+    runascoreuser /tmp/runascoreuserscript
+
+    # Let it come up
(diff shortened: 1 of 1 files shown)


dustymabe:
We should merge this at the same time we update to podman 2.0.6.

tests/kola/podman/rootless-systemd
```diff
@@ -0,0 +1,59 @@
...
+    echo "$runascoreuserscript" > /tmp/runascoreuserscript
+    chmod +x /tmp/runascoreuserscript
+    sudo -u core -- /tmp/runascoreuserscript
+
+    # Let it come up
+    sleep 5
+
+    if ! curl http://localhost:8080 1>/dev/null; then
+        echo TEST FAILED 1>&2
+        su - core -c "podman logs httpd"
```
dustymabe: For some reason this output doesn't make it into the `systemctl status kola-runext.service` output. The output is in the journal but it doesn't get attributed to the `kola-runext.service` unfortunately. `journalctl -t kola-runext-rootless-systemd` does show it, though.

tests/kola/podman/rootless-systemd
```diff
@@ -0,0 +1,56 @@
...
+# work everywhere.
+# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv" }
+
+# Script to be executed as the `core` user to build and
+# run a rootless podman container that uses systemd inside.
+runascoreuserscript='
+#!/bin/bash
+set -eux -o pipefail
+
+# build systemd container (runs httpd via systemd)
```
cgwalters, line 22: Add a link to https://github.com/coreos/coreos-assembler/issues/1645 ? Since this is another case of 2 potential flake sources (container fetch *and* rpm-md fetch).

tests/kola/podman/rootless-systemd
cgwalters: I think using `runuser` will avoid that problem.  Alternatively, cleanly split it out to a separate service via:
`systemd-run -P --wait --uid=core --gid=core --no-ask-password ...`.

tests/kola/podman/rootless-systemd
dustymabe: Added a link in the latest push.

tests/kola/podman/rootless-systemd
dustymabe: I experimented with a lot of different options here. Nothing short of `| cat` in order to copy the stdout of the command to the current unit worked so I'm just going with that for now so I don't waste more time on it.

Check out the latest push to see what works in local testing.

tests/kola/podman/rootless-systemd
```diff
@@ -0,0 +1,68 @@
...
+#!/bin/bash
+set -euxo pipefail
+
+# build systemd container (runs httpd via systemd)
+# NOTE: we may want to not build this in the future and run
+#       from a common testutils container. See
+#       https://github.com/coreos/coreos-assembler/issues/1645
+cd $(mktemp -d)
+cat <<EOF > Containerfile
+FROM registry.fedoraproject.org/fedora:32
```
arithx, line 28: Is there any reason this couldn't either be a scratch container or a more lightweight pre-built (hopefully not maintained by us) container?

dustymabe: I don't know of any that exist that would be not maintained by us. See the comment above this that links to https://github.com/coreos/coreos-assembler/issues/1645. A single container maintained by us to try to handle these one off uses. I think that's the path to go for us in the future.

dustymabe:
This should get unblocked when https://github.com/coreos/fedora-coreos-config/pull/596 merges.

tests/kola/podman/rootless-systemd
arithx: Sure, I was mostly thinking short term while we wait for that container to exist we could use something like https://hub.docker.com/_/httpd but given it's short term I wouldn't block on anything.

arithx (approved):
wfm

tests/kola/podman/rootless-systemd
dustymabe: Right. The problem is that what we're testing here is "does systemd in a container work?". I just happened to use `httpd` here, but any systemd enabled service should do. I don't know of any containers that exist that are maintained that run systemd that we could use.

cgwalters:
Looks like this needs a rebase to pass CI?

dustymabe:
> Looks like this needs a rebase to pass CI?

rebased!

cgwalters (approved):
LGTM after green CI (it'd be nice to have Prow on more repos)

PR #592, opened 2020-08-31
40ignition-ostree: Fix Before= for fsck for root

cgwalters:
I was investigating a rare flake in repeated runs of rootfs reprovisioning,
and I noticed that the `Before=` line here incorrectly referenced `boot`
instead of `root`.

Then, I noticed in fact for root it's special and should simply be
`systemd-fsck-root.service` - which due to a different bug, we
aren't even running at the moment.

Anyways, let's fix this so that if we do fix that in the future
we won't race.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-root.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-root.service
@@ -9,7 +9,7 @@ Before=initrd-root-fs.target
 After=ignition-disks.service
 
 # Avoid racing with fsck
-Before=systemd-fsck@dev-disk-by\x2dlabel-boot.service
+Before=systemd-fsck@dev-disk-by\x2dlabel-root.service
 
 # Note we don't have a Requires: /dev/disk/by-label/root here like
 # the -subsequent service does because ignition-disks may have


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-root.service
```diff
@@ -9,7 +9,8 @@ Before=initrd-root-fs.target
 After=ignition-disks.service
 
 # Avoid racing with fsck
-Before=systemd-fsck@dev-disk-by\x2dlabel-boot.service
+Before=systemd-fsck@dev-disk-by\x2dlabel-root.service
+Before=systemd-fsck-root.service
```
jlebon: Seems like that's only a thing in the real root. `systemd-fsck-root.service(8)` says:

> systemd-fsck-root.service is responsible for file system checks on the root file system, but in only if the root filesystem wasn't checked in the initramfs.

Doesn't seem like it's shipped in the initramfs either:

```
[root@cosa-devsh ~]# lsinitrd /boot/ostree/fedora-coreos-0d3dc72b1764b855a618de5ef68603839b1bde0ad9b42e4e29afeac742b6aa43/initramfs-5.7.17-200.fc32.x86_64.img | grep systemd-fsck-root
[root@cosa-devsh ~]#
```

And `systemd-fsck` itself when called from that service just uses `/`: https://github.com/systemd/systemd/blob/a1447e77a77b80db84cadc7b03f80f653f74118d/src/fsck/fsck.c#L300-L303.

cgwalters: OK, dropped that clause.

jlebon (approved):
Commit message is still referencing `systemd-fsck-root.service`. LGTM otherwise.

cgwalters:
Fixed the commit message.

PR #595, opened 2020-09-02
manifest: Add bootupd for x86_64|aarch64

cgwalters:
This converts the EFI update directory from (rpm-)ostree specific
layout to bootupd layout.  Required for both initial
provisioning with bootupd as well as later updates.

Required by: https://github.com/coreos/coreos-assembler/pull/1695


cgwalters:
OK I reworked this so that it's a no-op if bootupd isn't installed.  Hopefully we can merge this now and testing out bootupd will involve less juggling of git branches.

jlebon:
I'm having trouble understanding how this all ties together. I would've expected bootupd on the compose side to interact in coreos-assembler somewhere in `create_disk.sh`. Doesn't it need to fiddle with `/boot`?

cgwalters:
> I'm having trouble understanding how this all ties together.

No worries, me too :smile:   I messed this all up initially.

Basically this part replaces `/usr/lib/ostree-boot`.

> I would've expected bootupd on the compose side to interact in coreos-assembler somewhere in create_disk.sh. Doesn't it need to fiddle with /boot?

That's a separate phase in https://github.com/coreos/coreos-assembler/pull/1695

*This* phase needs to happen for every single ostree commit build.  The coreos-assembler one happens only when we make a bootimage.

cgwalters:
Side note: the reason this is useful now is I am testing bootupd via
`make install DESTDIR=/var/srv/walters/builds/fcos/overrides/rootfs/` so it doesn't matter that it's not listed in the manifest too.  But I do plan to package it soon.

cgwalters:
OK bootupd is in the repos; I reworked this to make it explicitly required for `x86_64` and `aarch64` (though I need to fix the RPM `ExclusiveArchitecture` for the latter still).

manifests/bootupd.yaml
```diff
@@ -0,0 +1,10 @@
+# Integration with https://github.com/coreos/bootupd
+# xref https://github.com/coreos/fedora-coreos-tracker/issues/510
+packages:
+  - bootupd
```
jlebon, line 4: Need a matching entry in the lockfiles for this.

manifests/bootupd.yaml
```diff
@@ -0,0 +1,10 @@
...
+
+postprocess:
+  - |
+    #!/bin/bash
+    set -xeuo pipefail
+    /usr/bin/bootupctl backend generate-update-metadata /
```
jlebon, line 11: Can you add a comment here with a short explanation of what this does? E.g. (IIUC):

```
# Transforms /usr/lib/ostree-boot into a bootupd-compatible update payload
```

jlebon (approved):
LGTM overall!

cgwalters:
https://github.com/coreos/coreos-assembler/pull/1695#issuecomment-696241376
so that PR needs to land first.

cgwalters:
OK, :heavy_check_mark: CI here.  We should probably debate whether to run with this right now; bootupd is still a WIP but I know it works for the "install" case, and having this in will help a lot developing it.  OTOH it does mean shipping it to all FCOS users in WIP status.

Hmm...maybe what we want is support for "build options" so that we can merge to testing-devel but turn off the option for stable?  Not sure.

jlebon:
> We should probably debate whether to run with this right now; bootupd is still a WIP but I know it works for the "install" case, and having this in will help a lot developing it. OTOH it does mean shipping it to all FCOS users in WIP status.

I'm personally fine with merging it. We don't publicize/document the new CLI tool anywhere super user visible right now, and otherwise AFAICT bootupd on its own doesn't change the behaviour of a booted FCOS in any meaningful way, right?

Do you see any risks/potential areas for regression from this change?

manifests/bootupd.yaml
jlebon: Did you mean to address this one?

darkmuggle:
> OTOH it does mean shipping it to all FCOS users in WIP status.

How much of a WIP is bootupd right now? Is it "alpha" or "beta" quality? If is the former, then I think we should hold off. And what's the risk of bricking a user? FCOS is stable, but we would be introducing something that unstable by definition into the code. 

> AFAICT bootupd on its own doesn't change the behaviour of a booted FCOS in any meaningful way, right

This PR enables the socket. I would call that a meaningful change. 

If we do no foresee a major refactor or changes, then I would be okay shipping this once we have Kola tests to make sure we're good. I think we're probably safe, but let's put up the guard rails first to catch our human mistakes.

cgwalters:
> How much of a WIP is bootupd right now? Is it "alpha" or "beta" quality? If is the former, then I think we should hold off. 

I'd say alpha.

> And what's the risk of bricking a user?

If they somehow discover and run `bootupctl update` manually then non-zero, otherwise zero.  So...effectively zero.

> This PR enables the socket. I would call that a meaningful change.

That's mostly an implementation detail I think, it's a unix domain socket, not listening on TCP or anything like that.

But perhaps the simplest alternative is to keep the socket disabled?  That would make it easy to turn on via Ignition or by hand and test out.

manifests/bootupd.yaml
cgwalters: I put the comment above, but I guess it is better inline in the script.

cgwalters:
> But perhaps the simplest alternative is to keep the socket disabled? That would make it easy to turn on via Ignition or by hand and test out.

Updated with that change.

overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
```diff
@@ -3,3 +3,4 @@ enable fedora-coreos-pinger.service
 # Provide information if no ignition is provided
 enable coreos-check-ignition-config.service
 enable coreos-check-ssh-keys.service
+
```
jlebon: Minor/optional: spurious new line.

jlebon (approved):
LGTM. Though to be safe let's hold until the current round of releases for this week is done. Otherwise, this'll get immediately promoted to `testing` and `next` e.g. tomorrow.

cgwalters:
Also did https://github.com/coreos/bootupd/pull/27

jlebon:
Releases are out now. Restarted CI here.

PR #599, opened 2020-09-04
Revert "Add root reprovisioning tests"

cgwalters:
This reverts commit 522ca223a19fcf731dbfbcb4cbf26c7ea6c3456e.

These are constantly failing in CI, we don't know why yet.
ref https://github.com/coreos/fedora-coreos-config/issues/591

PR #602, opened 2020-09-08
[next] tree: promote changes from next-devel at 350e0f6

travier:


jlebon (approved):
Sanity-checked this includes https://github.com/coreos/fedora-coreos-config/pull/599.

PR #603, opened 2020-09-08
20live: explicitly mount squashfs with `loop` option

jlebon:
This should be implied by the fact that we're trying to mount a
filepath, but let's be explicit. For comparison, this is how it's
documented in both squashfs-tools and in tldp:

https://github.com/plougher/squashfs-tools/blob/c570c6188811088b12ffdd9665487a2960c997a0/USAGE#L182

https://tldp.org/HOWTO/html_single/SquashFS-HOWTO/#creating

Spurred by RHBZ#1863466 where seemingly passing `-o loop` makes a
difference, though the root cause is likely somewhere else.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/live-generator
@@ -69,6 +69,7 @@ Before=initrd-root-fs.target
 What=/root.squashfs
 Where=/sysroot
 Type=squashfs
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
LGTM

jlebon:
Actually, let's do https://github.com/coreos/fedora-coreos-config/pull/606 first.

bgilbert (approved):

PR #604, opened 2020-09-08
40ignition-ostree: always save root partition state

jlebon:
Right now, there's special logic in coreos-growpart so that we skip
`growpart` if the rootfs backing block device or size changed during
rootfs reprovisioning (e.g. if one simply changed from xfs to ext4).

Let's make that logic more generic so that it's not actually tied to
rootfs reprovisioning. That way, if a user changes the partition size
via Ignition (pending https://github.com/coreos/ignition/issues/924), we
know not to automatically `growpart`.


jlebon:
Not tested yet.

jlebon:
I think this could still make sense but let's revive it when we see a clearer need for it.

PR #606, opened 2020-09-08
ci: Also test live images

jlebon:
This should be part of coreos-ci-lib proper, but let's do this for now.


bgilbert (approved):

cgwalters:
> qemu-system-x86_64: -drive if=none,id=cache,discard=unmap,file=/srv/fcos/cache/cache2.qcow2: Failed to get "write" lock

Right...I think we need to only mount the `cache2.qcow2` for the main build.

cgwalters:
https://github.com/coreos/coreos-assembler/pull/1704

PR #609, opened 2020-09-09
Add support for root-on-LUKS

jlebon:
We have a new Clevis release now with the fixes we need, so add the
packages to the manifest. This is all that's needed to support
root-on-LUKS since the rest of the rootfs replacement stack is already
LUKS-aware.

See added test for sample Ignition config.

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -27,3 +27,14 @@ packages:
     evra: 4.5.1-1.fc32.aarch64
   afterburn-dracut:
     evra: 4.5.1-1.fc32.aarch64
+  # https://bodhi.fedoraproject.org/updates/FEDORA-2020-af9f9ccb12
+  # To get root-on-LUKS earlier for testing
+  # 14-1 is in stable, but is missing some fixes
+  clevis:
+    evra: 14-4.fc32.aarch64
+  clevis-dracut:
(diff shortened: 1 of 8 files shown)


arithx (approved):
🎉

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
```diff
@@ -18,3 +18,5 @@ enable afterburn-sshkeys@.service
 enable zincati.service
 # Testing aid
 enable coreos-liveiso-success.service
+# For unlocking non-root devices via Clevis
+enable clevis-luks-askpass.path
```
cgwalters: Shouldn't this be enabled in the baseline Fedora presets?  (No objection to doing it here in addition temporarily though)

cgwalters (approved):
Nice!  An epic amount of work went on to get here.

bgilbert (approved):

jlebon:
OK, updated this now and added a test!

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
jlebon: Turns out we actually don't need this at all!

jlebon:
Since I added a test here, would appreciate another review before merge.

cgwalters (approved):

PR #610, opened 2020-09-09
overlay: Install /usr/share/licenses/fedora-coreos-config/LICENSE

cgwalters:
Thought about this after seeing
https://discussion.fedoraproject.org/t/fedora-core-os-redistribution/23134

Today fedora-coreos-config is special in that it's one of the only things
not an RPM.  Today one can use e.g. `rpm -qf /path/to/file` to find the
source RPM then map that to `/usr/share/licenses/$pkg/` to find the license,
but that's not true of stuff we ship here.

This doesn't solve the `rpm -qf` part, but it does put a file in the
expected place.

### DIFF
--- a/overlay.d/05core/usr/share/licenses/fedora-coreos-config/LICENSE
+++ b/overlay.d/05core/usr/share/licenses/fedora-coreos-config/LICENSE
@@ -1,0 +1,21 @@
+Copyright 2018 Fedora CoreOS Authors.
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be included
+in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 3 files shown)


overlay.d/05core/usr/share/licenses/fedora-coreos-config/README.md
```diff
@@ -0,0 +1,8 @@
+# fedora-coreos-config
+
+Today most components of Fedora CoreOS are built as RPMs; this
+is the main exception.  fedora-coreos-config is "architecture-independent glue"
+and the overhead of building an RPM for each change is onerous.
+
+It's also *the* central point of management (e.g. it contains lockfiles), so having it be
+an RPM too would become circular.  Instead, coreos-assembler directly consumes it.
```
jlebon, line 8: Should we add a link to this repo in this file?

overlay.d/05core/usr/share/licenses/fedora-coreos-config/LICENSE
```diff
@@ -0,0 +1,21 @@
+Copyright 2018 Fedora CoreOS Authors.
```
jlebon, line 1: 2020

cgwalters: This is a copy of the `LICENSE` from the toplevel...which, it'd be nice to avoid duplicating but would require either some sort of "build" step or a way to tell the overlay commit to follow (some) symlinks or so.

overlay.d/05core/usr/share/licenses/fedora-coreos-config/README.md
cgwalters: Done and also added a `jq` invocation to find the source commit.

jlebon (approved):

overlay.d/05core/usr/share/licenses/fedora-coreos-config/LICENSE
jlebon: We could also make the toplevel `LICENSE` be a symlink to the overlay file. But meh, not worth blocking on this.

PR #611, opened 2020-09-11
Revert "Revert "Add root reprovisioning tests""

jlebon:
This reverts commit 77787a021688720b90036cf95323b6c31ff04ccf.

I can't reproduce the failures in
coreos/fedora-coreos-tracker#619 locally or on
the CI cluster, and the logs from previous failures are stale. Let's
re-enable the tests and if it comes back we'll debug more deeply.


cgwalters:
Looks like the CI run was aborted, retried.

jlebon:
All green on this now!

cgwalters (approved):

PR #616, opened 2020-09-15
tests/kola: use wipeVolume for root-on-LUKS test

jlebon:
Otherwise, Ignition will refuse to overwrite the original xfs filesystem
on there.


cgwalters (approved):

sohankunkerkar (approved):
LGTM

PR #617, opened 2020-09-16
overrides: fast-track catatonit 0.1.5-3

dustymabe:
For some reason the package went backwards in Fedora because
rhcontainerbot built 0.1.5-1 *after* lokesh had built 0.1.5-2.
Let's fast-track 0.1.5-3 so we don't downgrade catatonit,
which is currently what is happening.

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -1,11 +1,4 @@
 packages:
-  # Fast track ostree 2020.6
-  # https://github.com/coreos/fedora-coreos-tracker/issues/617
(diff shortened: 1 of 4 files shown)


jlebon (approved):
Let's add a link to https://bodhi.fedoraproject.org/updates/FEDORA-2020-33358ae219? Makes it easier for releasers to check if they're in.

LGTM otherwise!

dustymabe:
sigh.. I should have done that first thing. Thanks. Will do here in a minute.

dustymabe:
Added the link. I'm not going to wait for CI again because it's a waste since I only updated a comment.

PR #618, opened 2020-09-17
[next-devel] Fast-track multiple packages 

jlebon:
This matches `testing-devel`.

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -1,8 +1,1 @@
 packages:
-  # Fast track ostree 2020.6
-  # https://github.com/coreos/fedora-coreos-tracker/issues/617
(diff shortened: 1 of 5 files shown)


jlebon:
Hmm, I think we need some mechanism for sharing overrides more easily between `next-devel` and `testing-devel` during the phase in which they're the same (but while still allowing stream specific overrides).

dustymabe:
Thanks @jlebon!

> Hmm, I think we need some mechanism for sharing overrides more easily between `next-devel` and `testing-devel` during the phase in which they're the same (but while still allowing stream specific overrides).

We could have a file in the next branch that tells config bot to sync them or not.

dustymabe (approved):
LGTM

dustymabe:
> We could have a file in the next branch that tells config bot to sync them or not.

though, they are about to differ again soon: https://github.com/coreos/fedora-coreos-streams/issues/188#issuecomment-693688560

PR #626, opened 2020-09-18
`core` user: specifically enable afterburn-sshkeys@core.service

bh7cw:
This only enables afterburn-sshkeys@core.service, which is the
default instance of afterburn-sshkeys@.service, when the `core`
user is created.
Tested it works. Ready for review.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
@@ -1,0 +1,7 @@
+FCOS enables `afterburn-sshkeys@core.service` from `base.ign`, allowing the user
+to prevent Ignition from enabling the service with a user config if the user
+wants to change the username. Unlike FCOS, RHCOS doesn't fetch SSH keys and thus
+doesn't need `afterburn-sshkeys@core.service`. Therefore, RHCOS maintains its
+own copy of `base.ign`, and changes to one copy need to be synced to the other
+copy.
+See https://github.com/coreos/fedora-coreos-config/pull/626
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/base.ign
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/base.ign
@@ -15,5 +15,13 @@
         ]
       }
     ]
+  },
+  "systemd": {
+    "units": [
+      {
+        "enabled": true,
+        "name": "afterburn-sshkeys@core.service"
+      }
+    ]
   }
 }
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -16,8 +16,6 @@ enable afterburn-checkin.service
 enable afterburn-firstboot-checkin.service
 # Target to write SSH key snippets from cloud providers.
 enable afterburn-sshkeys.target
-# Service to write SSH key snippets from cloud providers.
-enable afterburn-sshkeys@.service
 # Update agent
 enable zincati.service
 # Testing aid


bgilbert:
Is the idea that the user's Ignition config can override the enabling of that unit?

bh7cw:
> Is the idea that the user's Ignition config can override the enabling of that unit?

Yes. Exactly.

bh7cw:
Test result:
```
[core@ip-172-31-1-153 ~]$ sudo journalctl | grep afterburn
Sep 18 19:35:21 localhost audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=afterburn-network-kargs comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Sep 18 19:35:21 localhost kernel: audit: type=1130 audit(1600457721.701:2): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=afterburn-network-kargs comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Sep 18 19:35:21 localhost kernel: audit: type=1131 audit(1600457721.719:3): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=afterburn-network-kargs comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Sep 18 19:35:21 localhost audit[1]: SERVICE_STOP pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=afterburn-network-kargs comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Sep 18 19:35:31 ip-172-31-1-153 ignition[626]: INFO     : files: op(3): [started]  processing unit "afterburn-sshkeys@core.service"
Sep 18 19:35:31 ip-172-31-1-153 ignition[626]: INFO     : files: op(3): [finished] processing unit "afterburn-sshkeys@core.service"
Sep 18 19:35:31 ip-172-31-1-153 ignition[626]: INFO     : files: op(6): [started]  setting preset to enabled for "afterburn-sshkeys@.service core"
Sep 18 19:35:31 ip-172-31-1-153 ignition[626]: INFO     : files: op(6): [finished] setting preset to enabled for "afterburn-sshkeys@.service core"
Sep 18 19:35:35 ip-172-31-1-153 systemd[1]: Created slice system-afterburn\x2dsshkeys.slice.
Sep 18 19:35:39 ip-172-31-1-153 afterburn[886]: Sep 18 19:35:39.804 INFO Putting http://169.254.169.254/latest/api/token: Attempt #1
Sep 18 19:35:40 ip-172-31-1-153 afterburn[886]: Sep 18 19:35:40.850 INFO Putting http://169.254.169.254/latest/api/token: Attempt #2
Sep 18 19:35:40 ip-172-31-1-153 afterburn[886]: Sep 18 19:35:40.865 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/public-keys: Attempt #1
Sep 18 19:35:40 ip-172-31-1-153 afterburn[886]: Sep 18 19:35:40.867 INFO Fetch successful
Sep 18 19:35:40 ip-172-31-1-153 afterburn[886]: Sep 18 19:35:40.867 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/public-keys/0/openssh-key: Attempt #1
Sep 18 19:35:40 ip-172-31-1-153 afterburn[886]: Sep 18 19:35:40.871 INFO Fetch successful
Sep 18 19:35:40 ip-172-31-1-153 afterburn[886]: wrote ssh authorized keys file for user: core
Sep 18 19:35:40 ip-172-31-1-153 systemd[1]: afterburn-sshkeys@core.service: Succeeded.
Sep 18 19:35:40 ip-172-31-1-153 systemd[1]: Reached target Synchronize afterburn-sshkeys@.service template instances.
Sep 18 19:35:40 ip-172-31-1-153 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=afterburn-sshkeys@core comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Sep 18 19:35:40 ip-172-31-1-153 kernel: audit: type=1130 audit(1600457740.897:188): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=afterburn-sshkeys@core comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Sep 18 19:35:40 ip-172-31-1-153 audit[1]: SERVICE_STOP pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=afterburn-sshkeys@core comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Sep 18 19:35:41 ip-172-31-1-153 kernel: audit: type=1131 audit(1600457740.897:189): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=afterburn-sshkeys@core comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
```

cgwalters (approved):
Overall this sounds good to me, though presumably you wrote an fcct/Ignition to test this behavior?  If so we can add it as a test case here too, just drop it in `tests/kola/afterburn-sshkeys` or so - see the `var-mount` tests for examples that use an Ignition config and then sanity check the target state.

bh7cw:
> Overall this sounds good to me, though presumably you wrote an fcct/Ignition to test this behavior? If so we can add it as a test case here too, just drop it in `tests/kola/afterburn-sshkeys` or so - see the `var-mount` tests for examples that use an Ignition config and then sanity check the target state.

Good idea. :+1:  I learned more.

dustymabe (approved):
LGTM

jlebon:
Hmm, I was trying to figure out how `afterburn-sshkeys@.service` gets instantiated for the core service. And I think it's via [DefaultInstance=](https://github.com/coreos/afterburn/blob/5950143b474dad48f9338f018051384e1a519166/systemd/afterburn-sshkeys%40.service.in#L24)[core](https://github.com/coreos/afterburn/blob/5950143b474dad48f9338f018051384e1a519166/Makefile#L4), right?

Which means that previously users who wanted to change the username which gets the cloud SSH keys could add a drop-in for overriding `DefaultInstance`. But now they have to also disable this systemd service. And if they don't, the `core` user can be SSH'ed in.

>> Is the idea that the user's Ignition config can override the enabling of that unit?
>
> Yes. Exactly.

Couldn't they do that before by just disabling the preset?

darkmuggle:
> Which means that previously users who wanted to change the username which gets the cloud SSH keys could add a drop-in for overriding DefaultInstance. But now they have to also disable this systemd service. And if they don't, the core user can be SSH'ed in.

I don't believe this behavior is documented. Providing a drop-in unit is a worse UX than enabling then unit:
```
variant: fcos
version: 1.0.0
passwd:
  users:
   - name: darkmuggle
     ssh_authorized_keys:
      - "ssh-rsa...."
systemd:
  units:                                                                                                                               
    - name: afterburn-sshkeys@.service
      dropins:
      - name: autologin-core.conf
        contents: |
            [Install]
            DefaultInstance=
            DefautlInstance=darkmuggle
````
vs
```
variant: fcos
version: 1.0.0
passwd:
  users:
   - name: darkmuggle
     ssh_authorized_keys:
      - "ssh-rsa...."
systemd:
  units:                                                                                                                               
    - name: afterburn-sshkeys@darkmuggle.service
      enabled: true
````

> Couldn't they do that before by just disabling the preset?

The problem with the current default present is that the core user is not always created and therefore the default unit will fail.

**Review by darkmuggle — APPROVED** [review]
/lgtm

**Review by jlebon — APPROVED** [review]
My initial confusion here is that users should today already be able to do e.g.

```yaml
systemd:
  units:
    - name: afterburn-sshkeys@.service
      enabled: false
```

But testing this, I realize this is hitting up against https://github.com/systemd/systemd/pull/15205 again which makes the above not actually disable the service. We should try to find a way forward on that issue soon.

lucab:
Isn't this change aimed at a directory which is also going to affect RHCOS? How does it behave there, where the SSH keys service shouldn't be used?

bh7cw:
> Isn't this change aimed at a directory which is also going to affect RHCOS? How does it behave there, where the SSH keys service shouldn't be used?

Good point. Will open a PR at rhcos after this is completed.

bgilbert:
@darkmuggle, re https://github.com/coreos/fedora-coreos-config/pull/626#issuecomment-696812396, in the second case you'd also want to disable the default instance:

```yaml
variant: fcos
version: 1.0.0
passwd:
  users:
   - name: darkmuggle
     ssh_authorized_keys:
      - "ssh-rsa...."
systemd:
  units:
    - name: afterburn-sshkeys@darkmuggle.service
      enabled: true
    - name: afterburn-sshkeys@core.service
      enabled: false
```

@bh7cw, if the PR mentioned in https://github.com/coreos/fedora-coreos-config/pull/626#issuecomment-696939685 is to carry an RHCOS override for `base.ign`, we should consider having a README here that points out the need to keep RHCOS in sync.

bgilbert (approved):

bgilbert:
For now, I think it's okay for RHCOS to carry a base config override with some common elements hardcoded.  As our base config becomes more complex, we may want to make it easier for distros to override portions of a default base config; I've filed https://github.com/coreos/ignition/issues/1101 for this.

lucab:
Overall this is ok to land, but it's quite likely that we make uncoordinated changes when merging/ratcheting this, thus ending up with behavioral side-effects in RHCOS.
Before merging this, @bh7cw it would be great if you could:
 1. already prepare and get merged the override in RHCOS config repo
 1. add a README here pointing back to that copy to be kept in sync

bh7cw:
>     1. already prepare and get merged the override in RHCOS config repo
>     2. add a README here pointing back to that copy to be kept in sync

Make sense. Have submitted PR there. And added README.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,3 @@
+Since `afterburn-sshkeys@core.service` is enabled in `base.ign` in **FCOS**,
```
bgilbert: This is a true statement, but it'd be better to be explicit about the cause and the consequence of this fact.  I think this README should say two things: 1) `afterburn-sshkeys@core.service` is not used in RHCOS, so RHCOS carries its own copy of `base.ign` without it, and 2) these two files need to be kept in sync.  As an exercise, try writing that in your own words.  :slightly_smiling_face: 

Nit: we don't normally write FCOS and RHCOS in **bold**.

bh7cw: Good point. Thanks for review. Have updated the PR.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,6 @@
+FCOS changes to specially enable `afterburn-sshkeys@core.service` after
+create the `core` user in the `base.ign` instead of enable all the 
+instances of `afterburn-sshkeys@.service` via preset.
```
bgilbert: Most of the time I think it's not useful to document how code has worked in the past.  It's better to say why code works the way it does now.  Why does FCOS enable this service from the Ignition config instead of the preset?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,6 @@
+FCOS changes to specially enable `afterburn-sshkeys@core.service` after
+create the `core` user in the `base.ign` instead of enable all the 
+instances of `afterburn-sshkeys@.service` via preset. 
+RHCOS requires the same `base.ign` except `afterburn-sshkeys@core.service`.
```
bgilbert: Why does RHCOS not also enable `afterburn-sshkeys@core.service`?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,6 @@
...
+So, RHCOS keeps `base.ign` itself and needs to be synchronized with FCOS
```
bgilbert: s/keeps `base.ign` itself/maintains its own copy of `base.ign`/

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,6 @@
...
+in the future.
```
bgilbert: "in the future" is vague.  The important part is that changes to one copy need to be synced to the other copy.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
bh7cw: Thanks for review. The preset enables all the instances of `afterburn-sshkeys@.service`, but actually we just want to enable `afterburn-sshkeys@core.service`, and let the user to define what they want to enable as needed. So, it makes more sense to have it after create the `core` user IMO. What do you think?

bgilbert: Well, a couple things:
- Enabling `foo@.service` actually enables `foo@<some default>.service`, where the default is specified in a `DefaultInstance` key.  Afterburn is currently [defaulting](https://github.com/coreos/afterburn/blob/b93b18797d00e0aee5acbaae51f5e23d08825300/systemd/afterburn-sshkeys%40.service.in#L25) to the [core](https://github.com/coreos/afterburn/blob/b93b18797d00e0aee5acbaae51f5e23d08825300/Makefile#L26) [user](https://github.com/coreos/afterburn/blob/b93b18797d00e0aee5acbaae51f5e23d08825300/Makefile#L4).
- Presets run in the real root, but Ignition creates the `core` user beforehand, from the initramfs.

So that's not the reason.  The actual issue is https://github.com/coreos/fedora-coreos-config/pull/626#pullrequestreview-493654904.

bh7cw: Thanks for the quick response. So, for now `disable afterburn-sshkeys@.service` is not supportable, but if we put `enable afterburn-sshkeys@core.service` at base.ign, it can be disabled. Is that correct?

bgilbert: Almost.  If we enable `afterburn-sshkeys@core.service` in the base config, then when the user config is [merged](https://github.com/coreos/ignition/blob/master/docs/operator-notes.md#config-merging), the user config can override that enablement.  So it's not that Ignition is writing out `enable afterburn-sshkeys@core.service` followed by `disable afterburn-sshkeys@core.service`, it's that the user can prevent Ignition from enabling the service at all.

bh7cw: Thank you! That makes sense.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,8 @@
+FCOS changes to specially enable `afterburn-sshkeys@core.service` after
```
bgilbert: How about: "FCOS enables `afterburn-sshkeys@core.service` from `base.ign`, allowing the user to prevent..."

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,8 @@
...
+create the `core` user in the `base.ign`. In this case, the user can
+prevent Ignition from enabling the service with user config if the user
```
bgilbert: "with a user config"

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -0,0 +1,8 @@
...
+wants to change the username. Unlike FCOS, RHCOS doesn't fetch SSH keys
+and thus does't need `afterburn-sshkeys@core.service`. Therefore, RHCOS
```
bgilbert: "doesn't"

bh7cw: Oops, good catch.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
bh7cw: Sounds good. Thanks for review.

bgilbert (approved):
:+1:

lucab (approved):

PR #629, opened 2020-09-22
tests/kola/misc-ro: check that rpmdb is using bdb

jlebon:
Make sure we don't unknowingly move over to sqlite until we're ready.
See: https://github.com/coreos/fedora-coreos-tracker/issues/623

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -64,3 +64,10 @@ if [ -n "${unlabeled}" ]; then
     exit 1
 fi
 ok no files with unlabeled_t SELinux label
+
+# make sure we stick with bdb until we're ready to move to sqlite
+# https://github.com/coreos/fedora-coreos-tracker/issues/623
+if [ ! -f /usr/share/rpm/Packages ]; then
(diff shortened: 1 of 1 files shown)


tests/kola/misc-ro
```diff
@@ -64,3 +64,9 @@ if [ -n "${unlabeled}" ]; then
     exit 1
 fi
 ok no files with unlabeled_t SELinux label
+
+# make sure we stick with bdb until we're ready to move to sqlite
+# https://github.com/coreos/fedora-coreos-tracker/issues/623
+if [ ! -f /usr/share/rpm/Packages ]; then
+    fatal "Didn't find bdb file /usr/share/rpm/Packages"
+fi
```
cgwalters, line 72:
```suggestion
fi
ok rpmdb is bdb
```

cgwalters (changes requested):

tests/kola/misc-ro
cgwalters: (We're not actually doing TAP right now but in the future we should)

dustymabe (approved):
LGTM

PR #635, opened 2020-09-23
[next-devel] overrides: Fast-track podman-2.1.0-2.fc32 for baking on the `next` stream.

dustymabe:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-76fcd0ba34


dustymabe:
Also pulled over a few other overrides commits that needed to come from the `testing-devel` branch.

dustymabe:
I'm not 100% sure we should do this but figured I'd see if CI passed.

cverna:
I would be generally +1 to merge this considering this is security update and this is the for the `next` stream.

lucab (approved):

PR #636, opened 2020-09-24
tests/manual: nettest: add nameserver testing

dustymabe:
The primary commit in this patch set adds testing for setting the dns
nameserver. The other commits are relatively minor.

```
commit de11a1da180bdec570cc6ed4724d69db14b32a05
Date:   Wed Sep 23 23:15:32 2020 -0400

    tests/manual: nettest: add nameserver testing
    
    This commit attempts to add tests that confirm the nameserver is getting
    passed in correctly in multiple ways.
    
    - via ip=${ip}::${gateway}:${netmask}:${hostname}:${devname}:none:${nameserver}
    - via nameserver=${nameserver}
        - specifically see bug https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/391
    
    It also tries to verify the nameserver got propagated into either
    resolv.conf (RHEL 8 and Fedora less than 32) or systemd-resolved (F33+).

commit 16586de4eae9e960663f7a989572d93dc01bc13c
Date:   Wed Sep 23 23:13:32 2020 -0400

    tests/manual: nettest: workaround NM 1.26 bug
    
    There is an upstream NM bug that causes NM to wait for the second
    interface to come up even though we explicitly set it to `off`.
    Conversations are ongoing upstream about how to fix it.
    
    https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/539

commit f543b679ee3d595dc980eeb522b67e1b39d78257
Date:   Wed Sep 23 23:10:44 2020 -0400

    tests/manual: nettest: use virsh shutdown instead of reboot
    
    I saw some weird behavior where the reboot was actually managing
    to start the VM as well, which caused the `virsh start` to fail.
    Let's just make things clearer by using an explicit `shutdown` and
    then `start`.

commit 0b8d0ad7e3f1afb55b21079694281ddcc04fce48
Date:   Wed Sep 23 23:07:59 2020 -0400

    tests/manual: nettest: stop using ign-converter for RHCOS
    
    Now that RHCOS uses Ignition spec V3 let's drop the use of
    `ign-converter`. We leave the important bits in there as a comment
    in case the user wants to uncomment and run a test against <4.6.
```

### DIFF
--- a/tests/manual/coreos-network-testing.sh
+++ b/tests/manual/coreos-network-testing.sh
@@ -418,11 +418,9 @@ destroy_vm() {
 create_ignition_file() {
     local fcctconfig=$1
     local ignitionfile=$2
-    if [ "$rhcos" == 1 ]; then
-        echo "$fcctconfig" | fcct --strict | ign-converter -downtranslate -output $ignitionfile
-    else
+    # uncomment and use ign-converter instead if on rhcos less than 4.6
+    #echo "$fcctconfig" | fcct --strict | ign-converter -downtranslate -output $ignitionfile
         echo "$fcctconfig" | fcct --strict --output $ignitionfile
-    fi
     chcon --verbose unconfined_u:object_r:svirt_home_t:s0 $ignitionfile &>/dev/null
 }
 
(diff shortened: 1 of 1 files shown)


dustymabe:
The reason I dug into this was because I needed to gain a better understanding of the `nameserver=` argument and its current behavior in RHCOS as part of the investigation for https://bugzilla.redhat.com/show_bug.cgi?id=1870871. Which lead me into revisiting https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/391 and trying to verify the upstream fix a bit more (since it has now landed in F33, and probably will be in RHEL 8.3. While I was in there understanding things I wrote this up to automate it and also comment things.

Also happened to find a new bug in the process: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/539

cgwalters (approved):
Only skimmed this but looks good.

dustymabe:
Thanks @cgwalters!

PR #639, opened 2020-09-24
overlay: initramfs teardown: read hostname from file

dustymabe:
In 1.26 NM started [1] writing out the detected hostname to a file
(/run/NetworkManager/initrd/hostname). Let's pick up the hostname
from that file if we are on 1.26+.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/ff70adf

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
@@ -64,34 +64,60 @@ propagate_initramfs_networking() {
 
 # Propagate the ip= karg hostname if desired. The policy here is:
 #
-#     - IF a hostname is specified in static networking ip= kargs
+#     - IF a hostname was detected in ip= kargs by NetworkManager
 #     - AND no hostname was set via Ignition (realroot `/etc/hostname`)
-#     - THEN we make the last hostname specified in an ip= karg apply
-#       permanently by writing it into `/etc/hostname`
+#     - THEN we make the hostname detected by NM apply permanently
+#       by writing it into `/etc/hostname`
 #
-# This may no longer be needed when the following bug is fixed:
-# https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/419
 propagate_initramfs_hostname() {
     if [ -e '/sysroot/etc/hostname' ]; then
         echo "info: hostname is defined in the real root"
         echo "info: will not attempt to propagate initramfs hostname"
         return 0
     fi
-    # Detect if any hostname was provided via static ip= kargs
-    # run in a subshell so we don't pollute our environment
-    hostnamefile=$(mktemp)
-    (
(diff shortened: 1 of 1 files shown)


dustymabe:
This is prep for f33 but I'm opening against the testing-devel branch because the compat code should work just fine here and I'd like to exercise the compat code properly by putting it into testing-devel.

dustymabe:
If you "Hide whitespace changes" when viewing the diff it's easier to see the changes that were made.

dustymabe:
example of compat code getting executed:

```
[    5.145903] coreos-teardown-initramfs[729]: info: NM version is older than 1.26.0. Executing compat code path.
[    5.147317] coreos-teardown-initramfs[729]: info: propagating initramfs hostname (initrdhost) to the real root
```

lucab:
This looks fine to me, thanks for the compat logic. Reviewing shell script is always spotty, you may want to double-check locally whether `shellcheck` finds anything major (I'd expect lot of minor noise though). I'm stamping this in the meanwhile, feel free to self-merge when done.

lucab (approved):

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
```diff
@@ -64,34 +64,60 @@ propagate_initramfs_networking() {
...
+
+    # COMPAT: keep two code paths, one for older NetworkManager and
+    #         one for newer NetworkManager that supports writing to
+    #         /run/NetworkManager/initrd/hostname. We can delete this
+    #         block once RHCOS and FCOS minimum NM version is >= 1.26.0
+    #         See https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/ff70adf
+    barrierversion='1.26.0'
+    nmversion=$(/usr/sbin/NetworkManager --version)
+    sorted=$((echo $barrierversion; echo $nmversion) | sort -V | tail -n 1)
+    if [ $sorted == $barrierversion ]; then
```
jlebon, line 87: Shouldn't we check that `nmversion` and `barrierversion` aren't equal? Otherwise we'll execute the compat code if the version is exactly `1.26.0`, no?

It's too bad there isn't a clean way to do feature detection for this.

dustymabe: Let me see if I can alleviate your concerns (or you tell me if it makes things worse). When you run `/usr/sbin/NetworkManager --version` you actually get back a longer string:

```
[dustymabe@media ~]$ /usr/sbin/NetworkManager --version
1.26.2-1.fc33.1
```

So they will never be exactly equal strings. Does that make things worse or better?

dustymabe: Also note that the compat code still works on newer versions, just that we prefer to delete the compat code in the future and this is the first step towards doing that.

PR #644, opened 2020-09-28
tests/manual: Clarify expected state

cgwalters:
This is failing for me with both current fcos/rhcos
and seeing more clearly what is expected vs actual is always helpful.

### DIFF
--- a/tests/manual/coreos-network-testing.sh
+++ b/tests/manual/coreos-network-testing.sh
@@ -376,7 +376,7 @@ check_vm() {
     # with `wc -l <<< $keyfiles`.
     if [ ${detectedkeyfiles} != ${numkeyfiles} ]; then
         rc=1
-        echo "ERROR: Incorrect number of NM keyfiles created" 1>&2 
+        echo "ERROR: Expected ${numkeyfiles} NM keyfiles, found ${detectedkeyfiles}" 1>&2
     fi
 
     # verify that the nameserver got set


cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1870871

cgwalters:
@dustymabe can you verify that this test does pass for you?  And if so with which FCOS did you test?

Hmm debugging a bit it seems like my test VM has *two* `e1000` interfaces for some reason...

cgwalters:
Oh I see from https://github.com/coreos/fedora-coreos-config/pull/636#issuecomment-698100756 this won't work against current.

dustymabe:
> xref https://bugzilla.redhat.com/show_bug.cgi?id=1870871

Also xref: https://bugzilla.redhat.com/show_bug.cgi?id=1882781 which has more details on current behavior of `nameserver=` option in FCOS (based on F32) and RHCOS 4.6. 

> @dustymabe can you verify that this test does pass for you? And if so with which FCOS did you test?

It passes on Fedora 33 (which I just recently started working on) and RHCOS based on RHEL 8.3 (using the builds Timothee started), Though, note the [workaround](https://github.com/coreos/fedora-coreos-config/blob/aaed71c1d7a7c436811aacdab575442c2c234df7/tests/manual/coreos-network-testing.sh#L542-L543) embedded in the tests for the [other bug I found](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/539) while working on this.



> Oh I see from [#636 (comment)](https://github.com/coreos/fedora-coreos-config/pull/636#issuecomment-698100756) this won't work against current.

Right. Maybe I should comment out that test for now?

dustymabe (approved):
The proposed error message change LGTM.

PR #645, opened 2020-09-30
coreos-boot-mount-generator: Use Options=umask=077 for /boot/efi

### LINKED ISSUE #640: accessibility of /boot/efi
Splitting this from https://github.com/coreos/bootupd/pull/48#discussion_r497607504

Today:
```
$ rpm -q efi-filesystem
efi-filesystem-4-4.fc32.noarch
$ rpm -qlv efi-filesystem
drwx------    2 root     root                        0 Jan 28  2020 /boot/efi
drwxr-xr-x    2 root     root                        0 Jan 28  2020 /boot/efi/EFI
drwx------    2 root     root                        0 Jan 28  2020 /boot/efi/EFI/BOOT
drwx------    2 root     root                        0 Jan 28  2020 /boot/efi/EFI/fedora
```

I also verified that an unprivileged user can't `ls -al /boot/efi` on Fedora (traditional) Workstation 32.

Yet on FCOS:
```
[core@cosa-devsh ~]$ ls -al /boot/efi/EFI/
total 22
drwxr-xr-x. 4 root root  2048 Sep 23 08:21 .
drwxr-xr-x. 3 root root 16384 Jan  1  1970 ..
drwxr-xr-x. 2 root root  2048 Sep 23 08:21 BOOT
drwxr-xr-x. 3 root root  2048 Sep 23 08:21 fedora
```

We weren't paying attention to perms in https://github.com/coreos/coreos-assembler/blob/master/src/create_disk.sh before, nor are we currently with bootupd.

Perhaps we should be `0700 `in all of these directories.  (Though if the toplevel `/boot/efi` is `0700` then most of the other perms don't matter)  

AFAICS those perms came in with https://src.fedoraproject.org/rpms/efi-rpm-macros/c/4d38fc38961a39f6109eb6b5897c6c5a406b436e though there's no rationale for `0700` in the commit or comments in the code.

As far as I know there's not any private/secret data there.

cgwalters:
This is basically a consensus; the default systemd mount generator
does this.  While it's not a concern for CoreOS specifically
on general dual boot and other scenarios one might have private
or machine-specific data in there.

While there's no reason to prevent unprivileged users from
reading our pre-built bootloader binaries, I think it's reasonable
for e.g. security scanning systems to check this.

I'm mostly doing this because for bootupd I want to validate
our permissions handling and having things be common here
is better.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/640

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -77,5 +77,12 @@ if [ ! -f /run/ostree-live ]; then
     # when booted through BIOS.
     if [ "$(uname -m)" = "x86_64" -o -d /sys/firmware/efi ]; then
         mk_mount boot-efi EFI-SYSTEM "/boot/efi"
+        # In the general case the ESP might have per-machine or private
+        # data on it.  Let's not make it world readable on general
+        # principle.
+        # https://github.com/coreos/fedora-coreos-tracker/issues/640
+        cat >>${UNIT_DIR}/boot-efi.mount << EOF
+Options=umask=0077
+EOF
     fi
 fi
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -59,6 +59,9 @@ ok LICENSE
 
 case "$(arch)" in
     x86_64|aarch64)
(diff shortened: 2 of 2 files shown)


jlebon (approved):
Seems reasonable to me!

PR #646, opened 2020-09-30
tests: misc-ro: make systemd-resolved service check conditional

dustymabe:
Make the systemd-resolved service check conditional on Fedora major
version. In Fedora 32 it should be disabled. In Fedora 33+ it should
be enabled by default.

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -20,11 +20,25 @@ get_journal_msg_timestamp() {
         | jq -r --slurp '.[0]["__MONOTONIC_TIMESTAMP"]'
 }
 
-for unit in logrotate systemd-resolved; do
-    if ! systemctl is-enabled ${unit}; then
+# Test some services are enabled or disabled appropriately
+for unit in logrotate; do
+    if ! systemctl is-enabled ${unit} 1>/dev/null; then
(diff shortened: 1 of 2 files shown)


dustymabe:
will merge after #575

tests/kola/misc-ro
```diff
@@ -20,11 +20,25 @@ get_journal_msg_timestamp() {
...
-    if ! systemctl is-enabled ${unit}; then
+# Test some services are enabled or disabled appropriately
+for unit in logrotate; do
+    if ! systemctl is-enabled ${unit} 1>/dev/null; then
         fatal "Unit ${unit} should be enabled"
     fi
 done
+# systemd-resolved should be disabled on f32 but
+# enabled on f33+.
+source /etc/os-release
```
jlebon, line 31: I usually prefer to source this in a subshell to avoid polluting the env, but this is OK too.

jlebon (approved):
LGTM! I think we can just merge this one and close the other one included here? That avoids sending out a webhook event for a build state which is gonna fail tests anyway.

dustymabe:
> LGTM! I think we can just merge this one and close the other one included here? That avoids sending out a webhook event for a build state which is gonna fail tests anyway.

OK. Will do.

PR #648, opened 2020-09-30
manifest.yaml: host systemd-networkd removal bits in manifest.yaml

dustymabe:


dustymabe:
Assuming this gets merged I'll port over to `stable` and `testing` as well.

jlebon (approved):
I think we might not need this in practice, because since it's a separate subpackage the directive here won't match. But makes sense to just do this to keep our sanity.

We need to make sure that if we do a `next` release before it moves to f33 that we open a PR like this against `next-devel` too.

PR #653, opened 2020-10-01
manifests/base: add ncurses

jlebon:
This is currently getting pulled into the base by `coreutils` but won't
be in f33. We do want it though, so let's explicitly add it to the
manifest.

This provides common terminal tools like `clear`, `reset`, `tput`, and
`tset`.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -128,6 +128,8 @@ packages:
   - openssl
   - vim-minimal
   - lsof
+  # Provides terminal tools like clear, reset, tput, and tset
+  - ncurses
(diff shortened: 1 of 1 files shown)


manifests/fedora-coreos-base.yaml
```diff
@@ -128,6 +128,7 @@ packages:
   - openssl
   - vim-minimal
   - lsof
+  - ncurses
```
dustymabe, line 132: The comment in the git commit would be useful in here IMHO.

cgwalters (approved):
Yeah we `clear`ly want this, otherwise admins would be cursing...

PR #654, opened 2020-10-01
[testing] manifest.yaml: host systemd-networkd removal bits in manifest.yaml

dustymabe:

(cherry picked from commit f7c0d22426355b1e1d51cd760f2409a54df90423)

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -13,3 +13,17 @@ repos:
 
 add-commit-metadata:
   fedora-coreos.stream: testing
+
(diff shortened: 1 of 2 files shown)


cgwalters:
Maybe rpm-ostree should grow something like:

```
releasever-include:
  "32": removals-32.yaml
```

to conditionalize the manifest on `$releasever`?

cgwalters (approved):

jlebon (approved):

dustymabe:
re-stamped!

PR #655, opened 2020-10-01
[stable] manifest.yaml: host systemd-networkd removal bits in manifest.yaml

dustymabe:

(cherry picked from commit f7c0d22426355b1e1d51cd760f2409a54df90423)

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -13,3 +13,17 @@ repos:
 
 add-commit-metadata:
   fedora-coreos.stream: stable
+
+# Put this in manifest.yaml for now and we'll delete when we move to
+# F33 where all of these files have been broken out to the systemd-networkd
+# subpackage.
+remove-from-packages:
+  # We're not using networkd.
+  - [systemd, /etc/systemd/networkd.conf,
+              /usr/lib/systemd/systemd-networkd,
+              /usr/lib/systemd/systemd-networkd-wait-online,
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
A test would be as easy as 
```
diff --git a/tests/kola/misc-ro b/tests/kola/misc-ro
index 5d4a00a..f1887c1 100755
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -40,6 +40,11 @@ else
 fi
 ok services
 
+if test -x /usr/lib/systemd/systemd-networkd; then
+    fatal found networkd
+fi
+ok no networkd
+
 # https://github.com/coreos/fedora-coreos-config/commit/2a5c2abc796ac645d705700bf445b50d4cda8f5f
 if ip link | grep -o -e " eth[0-9]:"; then
     fatal "detected eth* NIC naming on node"
```

too

jlebon (approved):

dustymabe:
re-stamped!

dustymabe:
```
--- FAIL: coreos.ignition.instantiated.enable-unit (36.23s)
        cluster.go:162: "systemctl -q is-active echo@foo.timer" failed: output , status Process exited with status 3
```

I'm pretty sure we can ignore this CI failure because the promotion in #662 will fix it.

dustymabe:
Going to merge this and we'll verify CI passes before we merge #662

PR #656, opened 2020-10-02
overlay/coreos-boot-mount-generator: Small code simplification

cgwalters:
Prep for further work.  systemd requires an exact relationship
between the unit name and mount point, so rather than having
the caller pass both, compute the unit name from the mount.


darkmuggle (approved):
lgtm

This is a really nice clean-up. `systemd-escape` is way nicer.

jlebon (approved):

PR #640, opened 2020-09-24
[next-devel] rebase to Fedora 33

dustymabe:
These commits perform the rebase to Fedora 33.

Part of the work for https://github.com/coreos/fedora-coreos-tracker/issues/611

PR #658, opened 2020-10-02
tests/kola/root-reprovision: bump the amount of RAM to 6G as 4G is not enough on ppc64le

jcajka:
Re-provision tests runs will fail(log in attachment), with something that seems as not enough memory on ppc64le. Increasing the memory size to the 6G enables the tests pass(obtained by trail and error). I assume that this is most probably caused by the image being roughly ~200M bigger than on the x86_64 and by usage of the 64k pages on the ppc64le.`
[ppc64le-raid1.log](https://github.com/coreos/fedora-coreos-config/files/5317325/ppc64le-raid1.log)

### DIFF
--- a/tests/kola/root-reprovision/filesystem-only/test.sh
+++ b/tests/kola/root-reprovision/filesystem-only/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "minMemory": 4096}
+# kola: {"platforms": "qemu", "minMemory": 6144}
 set -xeuo pipefail
 
 fstype=$(findmnt -nvr / -o FSTYPE)
--- a/tests/kola/root-reprovision/raid1/test.sh
+++ b/tests/kola/root-reprovision/raid1/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "minMemory": 4096, "additionalDisks": ["5G", "5G"]}
+# kola: {"platforms": "qemu", "minMemory": 6144, "additionalDisks": ["5G", "5G"]}
 set -xeuo pipefail
 
 srcdev=$(findmnt -nvr / -o SOURCE)


cgwalters (approved):

jlebon (approved):
Hmm, I wish there was a way to make it conditional on the architecture. 4G already meant a lot relative to my puny laptop. 6G is half my total memory.

Maybe what we want is a `root-reprovision` tag which kola knows to mean bumping the memory accordingly?

Anyway, we can get this in for now.

cgwalters:
I think if we added support for instance sizes https://github.com/coreos/coreos-assembler/pull/1651#issuecomment-675719475 then we could scale those up by 20% or whatever for architectures with 64k pages.

arithx:
> Hmm, I wish there was a way to make it conditional on the architecture. 4G already meant a lot relative to my puny laptop. 6G is half my total memory.
> 
> Maybe what we want is a `root-reprovision` tag which kola knows to mean bumping the memory accordingly?
> 
> Anyway, we can get this in for now.

As a temporary workaround we could always do the hacky method of having architecture specific duplicated tests with different memory requirements.

jcajka:
@jlebon @cgwalters: I will look on that, it makes most sense. I think that we can even bump the minimal amount of memory on 64k arches to something like 8G, safely(it should be ppc64le only and those are usually pretty beefy, whatever VMs or whole hosts). What do you think? CC'ing @Prashanth684 
From other side what is the minimal amount of memory that we target for the COS hosts? From my rather naive perceptive I would assume something in line with 8G+(with the use cases that we mention, maybe with exception of the aarch64 as there is still rather rare to see sbcs with more than 4G).
@jlebon: tbh I haven't been able to build COS from scratch locally(on x86_64) until like a week ago when I bought myself a new bigger disk for my laptop(podman's storage driver needs like ~50G of free space to build the container image, although the storage taken afterwards is way, way less...).

jlebon:
Cross-linking: https://github.com/coreos/fedora-coreos-config/pull/640#issuecomment-703751994

Restamped this to trigger CI here.

jcajka:
Moved the solution in to the kola via https://github.com/coreos/coreos-assembler/pull/1768

PR #659, opened 2020-10-02
overlay/boot-mount-generator: Mount /boot{,efi} read-only,nodev,nosuid 

cgwalters:
ostree has had support for leaving `/boot` mounted read-only
for a long time: https://github.com/ostreedev/ostree/pull/1767
(And then later extended to `/sysroot`)

Particularly for CoreOS, only a few things should be touching
`/boot`, and we control all of them.  Those projects should
create a new mount namespace and remount these partitions
writable just while they need it.

The main thing we're accomplishing here is making the system
more resilient against accidental damage from a sysadmin
root shell as well as configuration management tools like
Puppet/Ansible.  None of those should be directly manipulating
files on these partitions, they should go through the API
of one of our projects (e.g. `rpm-ostree kargs`, `bootupctl`) etc.

While we're here, also andd `nodev,nosuid` because some
OS hardening scanners like to see this.  IMO it's of minimal
value, but hey, might as well.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -35,8 +35,9 @@ fi
 mk_mount() {
     local mount_pt="${1}"; shift
     local label="${1}"; shift
-    local path="/dev/disk/by-label/${label}"
+    local options="${1}"; shift
 
+    local path="/dev/disk/by-label/${label}"
     local unit_name=$(systemd-escape -p ${mount_pt} --suffix=mount)
 
     eval $(udevadm info --query property --export "${path}")
@@ -61,15 +62,20 @@ After=systemd-fsck@${device}.service
 [Mount]
 What=${path}
 Where=${mount_pt}
+Options=${options}
 EOF
 
     add_wants "${unit_name}"
 }
 
-
-# Don't create mount units for /boot or /boot/efi on live systems.
+# Don't create mount units for /boot on live systems.
 # ConditionPathExists won't work here because conditions don't affect
 # the dependency on the underlying device unit.
 if [ ! -f /run/ostree-live ]; then
-    mk_mount /boot boot
+    # We mount read-only by default mostly to protect
+    # against accidental damage.  Only a few things
+    # owned by CoreOS should be touching /boot or the ESP.
+    # Use nodev,nosuid because some hardening guides want
+    # that even though it's of minimal value.
+    mk_mount /boot boot ro,nodev,nosuid
 fi
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -51,10 +51,15 @@ if ip link | grep -o -e " eth[0-9]:"; then
 fi
 ok nic naming
 
-if test -w /sysroot; then
-    fatal "found writable /sysroot"
-fi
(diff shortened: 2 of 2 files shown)


cgwalters:
Heh I forgot about https://github.com/coreos/fedora-coreos-config/pull/356 and just saw it - incorporated that too.

cgwalters:
This needs to resolve
https://github.com/coreos/fedora-coreos-config/pull/356#issuecomment-615470538
and at least bootupd needs to learn to do the remounting dance.

cgwalters:
One thing this does break is that I think it would be nice if sysadmins could just `vi /boot/loader/entries` - having a standardized text format is part of the idea.  But...eh.  We want fcct sugar for kargs which would use `rpm-ostree kargs` probably anyways, and for OpenShift the MCO already uses `rpm-ostree kargs` today too.

cgwalters:
Also after this let's revisit https://github.com/coreos/fedora-coreos-config/pull/407

kelvinfan001:
@cgwalters With the [updates to how `copy-firstboot-network` is done](https://github.com/coreos/fedora-coreos-config/pull/743) and also the [patch in bootupd](https://github.com/coreos/bootupd/pull/118) to ensure `/boot` and `/boot/efi` writable, is this ready to be merged once the next release of bootupd is out?

jlebon:
This is remounting `/boot/efi`, which we stopped mounting in https://github.com/coreos/fedora-coreos-config/pull/794 as per https://github.com/coreos/fedora-coreos-tracker/issues/694.

jlebon:
Ahh OK, is this related to https://github.com/openshift/os/pull/480#issuecomment-754728968 ? I.e. are we backtracking for the time being and going back to mounting `/boot/efi` by default until bootupd learns to mount it itself?

kelvinfan001:
@jlebon  Actually that was my bad, the latest release of bootupd should be able to handle a not-mounted EFI and a read-only `/boot`. I simply rebased this PR on the latest testing-devel. We should indeed remove the EFI mount section from this PR.
All this PR should be doing now is mounting `/boot` as RO and adding `nodev,nosuid`.

> are we backtracking for the time being and going back to mounting /boot/efi by default until bootupd learns to mount it itself?

I think https://github.com/coreos/bootupd/releases/tag/v0.2.5 should mean we don't need to backtrack.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -61,15 +62,21 @@ After=systemd-fsck@${device}.service
...
 # the dependency on the underlying device unit.
 if [ ! -f /run/ostree-live ]; then
-    mk_mount /boot boot
+    # We mount read-only by default mostly to protect
+    # against accidental damage.  Only a few things
+    # owned by CoreOS should be touching /boot or the ESP.
+    # Use nodev,nosuid because some hardening guides want
+    # that even though it's of minimal value.
+    boot_mount_options=ro,nodev,nosuid
+    mk_mount /boot boot "${boot_mount_options}"
```
jlebon: Minor/optional: meh, I'd just inline these. It made sense before because we were re-using it for the EFI mount as well.

jlebon (approved):
Commit title still references `efi` and also `andd` -> `add` in the commit message.

LGTM otherwise!

PR #664, opened 2020-10-05
manifests: move the permissions on chrony files to fedora-coreos-base.yaml

dustymabe:
This code was originally introduced in 0352445.

We need to move it out of manifest.yaml because that file doesn't get
synced across branches. We want it to get synced everywhere.


jlebon (approved):
LGTM!

Aside: we should probably add a big notice to the top of the `manifest.yaml` files in all the branches to explain the semantics there (or at least link to the README section).

PR #669, opened 2020-10-06
ci: checkout repo to the config/ directory

jlebon:
Instead of checking out the repo directly into the workspace, let's get
in the habit of checking it out into a subdirectory to keep things
clean.

This also helps code which look at the name of the directory in which
the repo is checked out, such as kola external host test handling.

This makes use of a new custom step:

https://github.com/coreos/coreos-ci-lib/pull/40


dustymabe (approved):
LGTM

dustymabe:
still LGTM

PR #673, opened 2020-10-07
manifests: fedora-coreos: add archive repo to FCOS

### LINKED ISSUE #400: package layering: split versions between OSTree base vs yum repo
- **NOTE**: we currently discourage package layering in FCOS as it makes updates less reliable. The root of this ticket is part of the reason why, but this issue is not specific to FCOS as mentioned below.

This is a problem that periodically plagued the Atomic Host community, periodically does plague the Silverblue community and is starting to creep into FCOS as users attempt to package layer rpms that aren't included in the base. The user attempts a package layering operation that fails because the versions of packages in the base OS are out of sync with the versions in the yum repositories. There is also [an issue against the RPM-OSTree repo](https://github.com/coreos/rpm-ostree/issues/415) for this problem, but it's a high level problem and a solution will probably take some coordination at multiple levels so we're creating a new ticket here to track that coordination.

The user ends up with an error like:

```
Forbidden base package replacements:
  libgcc 8.2.1-6.fc29 -> 8.3.1-2.fc29 (updates)
```

We need to coordinate with Fedora releng, Fedora Silverblue working group, and the RPM-OSTree maintainers, in order to try to come up with a better solution to this problem. @owtaylor from the Silverblue team did a nice writeup on the problem last year in this [public google document](https://docs.google.com/document/d/1yS0PTaUPmD-CkQkdlJ9OvY-Z5hBOIMlRKDyqutEG_ks), so there is much more context about the problem and possible solutions in there.

dustymabe:
This is the culmination of a lot of work to make package layering
more reliable. This archive repo provides all packages that have
ever been in the updates repository, which means there should always
be a solution that will depsolve given the existing set of base layer
packages.

Pairing this along with https://github.com/coreos/rpm-ostree/pull/2125
means that we should finally see less of the split base layer vs update
repo problem and see less `Forbidden base package replacements` errors.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/400

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -22,3 +22,13 @@ packages:
     evra: 0.20.1-1.fc32.noarch
   console-login-helper-messages-profile:
     evra: 0.20.1-1.fc32.noarch
+  # Fast-track fedora-repos-32-7 to get archive repo
+  # https://bodhi.fedoraproject.org/updates/FEDORA-2020-ab02250f9d
+  fedora-repos:
+    evra: 32-7.noarch
+  fedora-repos-ostree:
+    evra: 32-7.noarch
+  fedora-repos-archive:
+    evra: 32-7.noarch
+  fedora-gpg-keys:
+    evra: 32-7.noarch
--- a/manifest-lock.overrides.ppc64le.yaml
+++ b/manifest-lock.overrides.ppc64le.yaml
@@ -22,3 +22,13 @@ packages:
     evra: 0.20.1-1.fc32.noarch
   console-login-helper-messages-profile:
     evra: 0.20.1-1.fc32.noarch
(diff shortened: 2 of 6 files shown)


jlebon (approved):
Awesome! :tada: :tada: :tada:

I was going to suggest doing this in `next` first, but meh... a `testing` release just went out, so it'll bake for at least 4 weeks before it reaches `stable`.

dustymabe:
> Awesome! tada tada tada

:tada: 

> I was going to suggest doing this in `next` first, but meh... a `testing` release just went out, so it'll bake for at least 4 weeks before it reaches `stable`.

Yeah, and it really has no effect in `next` for now, which is f33. Because Fedora 33 is still in "pre-release", there haven't been any rpms in the "updates" repo yet so there are none in the archive. In the future we could expand the archive repo a bit for pre-release content, but for now I'm not going there.

PR #675, opened 2020-10-08
Revert "image.yaml: set osmet-pack-with-cosa-coreinst to true"

jlebon:
This reverts commit 838269245977192b9a4e79f2f0df6fbb6e5fcd7c.

We don't need this anymore now that coreos-assembler correctly uses the
target rootfs directly to run coreos-installer (see
https://github.com/coreos/coreos-assembler/pull/1774).


dustymabe (approved):
LGTM

PR #676, opened 2020-10-09
40ignition-conf: support .d dir for base config fragments

### LINKED ISSUE #1101: Consider supporting .d directory for base config fragments
We're starting to have divergence between the FCOS and RHCOS base configs (https://github.com/coreos/fedora-coreos-config/pull/626), which requires the two distros to carry different `base.ign` files with common elements.  Consider supporting a `.d`-style base config directory, where Ignition walks the directory and merges every config fragment in it.

sohankunkerkar:
Fixes https://github.com/coreos/ignition/issues/1101
This change is required for https://github.com/coreos/ignition/pull/1108
Ignition PR allows the two distros i.e. RHCOS and FCOS to carry different `base.ign` files with common elements and will provide some flexibility as we start to see more divergence there.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/module-setup.sh
@@ -7,6 +7,9 @@ depends() {
 }
 
 install() {
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+    inst "$moddir/00-core.ign" \
+        "/usr/lib/ignition/base.d/00-core.ign"
     inst "$moddir/base.ign" \
         "/usr/lib/ignition/base.ign"
 }
--- a/overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/README.md
+++ b/overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/README.md
@@ -1,0 +1,1 @@
+FCOS enables `afterburn-sshkeys@core.service` from `30-afterburn-sshkeys-core.ign`, allowing the user to prevent Ignition from enabling the service with a user config if the user wants to change the username. Unlike FCOS, RHCOS doesn't fetch SSH keys from cloud providers and thus doesn't need `afterburn-sshkeys@core.service`.
--- a/overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/module-setup.sh
+++ b/overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/module-setup.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+depends() {
+    echo ignition
+}
+
+install() {
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+    inst "$moddir/30-afterburn-sshkeys-core.ign" \
+        "/usr/lib/ignition/base.d/30-afterburn-sshkeys-core.ign"
+}
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/00-core.ign
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/00-core.ign
@@ -1,0 +1,19 @@
+{
+  "ignition": {
+    "version": "3.0.0"
+  },
+  "passwd": {
+    "users": [
+      {
+        "name": "core",
+        "gecos": "CoreOS Admin",
+        "groups": [
+          "adm",
+          "sudo",
+          "systemd-journal",
+          "wheel"
+        ]
+      }
+    ]
+  }
+}
--- a/overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/30-afterburn-sshkeys-core.ign
+++ b/overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/30-afterburn-sshkeys-core.ign
@@ -1,0 +1,13 @@
+{
+  "ignition": {
+    "version": "3.0.0"
+  },
+  "systemd": {
+    "units": [
+      {
+        "enabled": true,
+        "name": "afterburn-sshkeys@core.service"
+      }
+    ]
+  }
+}


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/afterburn-sshkeys.ign
```diff
@@ -0,0 +1,13 @@
...
+  },
+  "systemd": {
+    "units": [
+      {
+        "enabled": true,
+        "name": "afterburn-sshkeys@core.service"
+      }
+    ]
+  }
+}
```
lucab: It would be better to give this file a numeric prefix (i.e. `NN-afterburn-sshkeys.ign`).

jlebon: That makes sense to me. The way Ignition merges these configs will be order-sensitive since the later ones will override earlier ones per object key, right? We should make sure we document that.

We should probably also rename `base.ign` to `00-base.ign`.

sohankunkerkar: @lucab Yeah, that's a good idea. Any preferences for a numeric prefix? I'd imagine something like this:
```
00-base.ign
10-afterburn-sshkeys.ign
```
WDYT?

cgwalters:
I'm a bit confused, do we really need this change?
If we land https://github.com/coreos/ignition/pull/1108 then we basically just need a separate overlay that only enables the afterburn-sshkeys bit from https://github.com/coreos/fedora-coreos-config/pull/626/ in this repo that *isn't* inherited/included by RHCOS.

IOW for FCOS we have:

 - /usr/lib/ignition/base.d/05core.ign
 - /usr/lib/ignition/base.d/30core-afterburn-sshkeys.ign

And RHCOS just has the first or so?

bgilbert:
@cgwalters Yup, I agree with your proposal.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/10-afterburn-sshkeys.ign
```diff
@@ -0,0 +1,13 @@
+{
+  "ignition": {
+    "version": "3.0.0"
+  },
+  "systemd": {
+    "units": [
+      {
+        "enabled": true,
+        "name": "afterburn-sshkeys@core.service"
```
cgwalters: To clarify this bit looks good except it needs to move into `overlay.d/15fcos` I think.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/module-setup.sh
```diff
@@ -7,6 +7,10 @@ depends() {
 }
 
 install() {
-    inst "$moddir/base.ign" \
-        "/usr/lib/ignition/base.ign"
+    # See https://github.com/coreos/ignition/issues/1101
+    # Creating a `base.d` dir for placing the base config fragments.
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+    for file in "$moddir"/*.ign; do
```
cgwalters: And it's this bit I don't think is correct - instead we need two dracut modules that each install one file or so.

overlay.d/15fcos/usr/lib/dracut/modules.d/configs/30core-afterburn-sshkeys.ign
```diff
@@ -0,0 +1,13 @@
+{
```
cgwalters: "configs" is a too generic name for a dracut module.  How about `fcos-ignition`?

(The comment isn't related to this line in the file, but I can't comment on the name of the file in GH)

overlay.d/15fcos/usr/lib/dracut/modules.d/configs/module-setup.sh
```diff
@@ -0,0 +1,16 @@
...
+
+depends() {
+    echo ignition
+}
+
+install() {
+    # See https://github.com/coreos/ignition/issues/1101
+    # Creating a `base.d` dir for placing the base config fragments.
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+    for file in "$moddir"/*.ign; do
```
cgwalters: I'm OK with this as is, but it seems somewhat unlikely we'd add multiple fragments here in the future.  IOW we could just `inst 30core-afterburn-sshkeys.ign /usr/lib/ignition/base.d/` I think?

But still, OK as is.

cgwalters (approved):

overlay.d/15fcos/usr/lib/dracut/modules.d/configs/module-setup.sh
sohankunkerkar: > we could just inst 30core-afterburn-sshkeys.ign /usr/lib/ignition/base.d/

 Yeah, I think that makes sense to me.

cgwalters (approved):
LGTM!

overlay.d/15fcos/usr/lib/dracut/modules.d/fcos-ignition/module-setup.sh
```diff
@@ -0,0 +1,15 @@
...
+
+depends() {
+    echo ignition
+}
+
+install() {
+    # See https://github.com/coreos/ignition/issues/1101
+    # Creating a `base.d` dir for placing the base config fragment.
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+       inst "$moddir/30core-afterburn-sshkeys.ign" \
```
jlebon: Minor: this line looks overindented?

jlebon (approved):
One minor comment, LGTM as is!

jlebon:
Is this still WIP? PR title says it is. :)

sohankunkerkar:
> Is this still WIP? PR title says it is. :)

Actually, I wanted to get [this](https://github.com/coreos/ignition/pull/1108) PR in before merging this one. But I'm not sure if it's a valid concern.

jlebon:
Oh right, we need to hold this until the Ignition support is there. Otherwise we'd essentially be reverting https://github.com/coreos/fedora-coreos-config/pull/626.

Added a `hold` label.

> 15fcos: copy all base config fragments to base.d dir

Hmm, is this accurate? We're just splicing out a part of the base config into a `base.d`, right?

> This change is required for coreos/ignition#1101

How something like:

> Make use of the new support for config fragments in Ignition (https://github.com/coreos/ignition/pull/1108) to move out the Afterburn bit added in #626 to a separate overlay directory so that RHCOS can more easily avoid it.

sohankunkerkar:
> > 15fcos: copy all base config fragments to base.d dir
> 
> Hmm, is this accurate? We're just splicing out a part of the base config into a `base.d`, right?
> 
@jlebon Yeah, you're right. I think I missed that point to change the commit message and the commit body message after my last changes.

> > This change is required for [coreos/ignition#1101](https://github.com/coreos/ignition/issues/1101)
> 
> How something like:
> 
> > Make use of the new support for config fragments in Ignition ([coreos/ignition#1108](https://github.com/coreos/ignition/pull/1108)) to move out the Afterburn bit added in #626 to a separate overlay directory so that RHCOS can more easily avoid it.

+1

overlay.d/15fcos/usr/lib/dracut/modules.d/fcos-ignition/module-setup.sh
```diff
@@ -0,0 +1,15 @@
...
+
+depends() {
+    echo ignition
+}
+
+install() {
+    # See https://github.com/coreos/ignition/issues/1101
+    # Creating a `base.d` dir for placing the base config fragment.
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+    inst "$moddir/30core-afterburn-sshkeys.ign" \
```
bgilbert: The `module-setup.sh` in `40ignition-conf` should also be updated to write into `base.d`, no?

overlay.d/15fcos/usr/lib/dracut/modules.d/fcos-ignition/module-setup.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+depends() {
+    echo ignition
+}
+
+install() {
+    # See https://github.com/coreos/ignition/issues/1101
```
bgilbert: I don't think the bug link is helpful here, actually.  It implies that the bug context is necessary to understand what's going on here, but this code isn't a workaround for an obscure problem; it's just the way base configs will be stored in the initrd now.

overlay.d/15fcos/usr/lib/dracut/modules.d/fcos-ignition/module-setup.sh
```diff
@@ -0,0 +1,15 @@
...
+    # Creating a `base.d` dir for placing the base config fragment.
```
bgilbert: Nit: IMO this doesn't add anything that isn't clear from the `mkdir` command.

(I'm generally not in favor of "noise" comments.  Other folks may have a different view.)

overlay.d/15fcos/usr/lib/dracut/modules.d/fcos-ignition/module-setup.sh
```diff
@@ -0,0 +1,15 @@
...
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+    inst "$moddir/30core-afterburn-sshkeys.ign" \
+        "/usr/lib/ignition/base.d/30core-afterburn-sshkeys.ign"
```
bgilbert: Nit: putting "core" first makes it sound like this is e.g. "core functionality", rather than a reference to the core user.  I'd be inclined to call this `30afterburn-sshkeys-core.ign`.

bgilbert (commented):
Dracut module directory names usually start with a two-digit sequence number.

Remember to move and update the README.md from `40ignition-conf`.

overlay.d/15fcos/usr/lib/dracut/modules.d/10fcos-ignition/README.md
```diff
@@ -0,0 +1,4 @@
+With the new support for config fragments in Ignition [1], it would be easier for RHCOS to avoid the `afterburn-sshkeys@core.service` ( since RHCOS doesn't support SSH keys) by moving out the Afterburn change added in [2] to a seperate overlay directory i.e. `15fcos/usr/lib/dracut/modules.d/10-fcos-ignition`.
```
bgilbert: In most cases, code comments and READMEs should document what happens now, not how things have changed over time.  (Commit messages should document how things have changed.)  This text won't make sense in six months or two years.

Also, this text no longer describes _why_ we enable the Afterburn service from Ignition rather than from an OS preset.

bgilbert (commented):
10fcos-ignition will run before 40ignition-conf, which is currently harmless but a little confusing.  Maybe call it 50ignition-conf-fcos?

bgilbert (approved):
The core config has a dash after the sequence number in the filename, but the Afterburn fragment doesn't.  LGTM otherwise.

overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/README.md
```diff
@@ -0,0 +1 @@
+FCOS enables `afterburn-sshkeys@core.service` from `30afterburn-sshkeys-core.ign`, allowing the user to prevent Ignition from enabling the service with a user config if the user wants to change the username. Unlike FCOS, RHCOS doesn't fetch SSH keys from cloud providers and thus doesn't need `afterburn-sshkeys@core.service`.
```
bgilbert: `30-afterburn-sshkeys-core.ign`

bgilbert (approved):

overlay.d/15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/module-setup.sh
```diff
@@ -0,0 +1,13 @@
...
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+depends() {
+    echo ignition
+}
+
+install() {
+    mkdir -p "$initdir/usr/lib/ignition/base.d"
+    inst "$moddir/30afterburn-sshkeys-core.ign" \
+        "/usr/lib/ignition/base.d/30afterburn-sshkeys-core.ign"
```
jlebon: Hmm, this needs to be

```suggestion
    inst "$moddir/30-afterburn-sshkeys-core.ign" \
        "/usr/lib/ignition/base.d/30-afterburn-sshkeys-core.ign"
```

otherwise dracut won't find it, no?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/README.md
```diff
@@ -1,7 +0,0 @@
-FCOS enables `afterburn-sshkeys@core.service` from `base.ign`, allowing the user
```
jlebon: Might be worth still having a README.md here which just makes the reader aware of `50ignition-conf-fcos`.

bgilbert (approved):

sohankunkerkar:
After having some discussion with @bgilbert and @arithx, it appears to me that this PR can't be merged in its current form since it requires https://github.com/coreos/ignition/pull/1108, which leads to cutting another Ignition release. I think we already released a new version of Ignition a few days ago, and we wouldn't want to rush to create another release. Without the new Ignition release, this PR will certainly break CI.
In order to avoid CI breakage, we've decided to drop the `base.ign` removal from this PR and split it out to a separate PR (https://github.com/coreos/fedora-coreos-config/pull/717).

arithx (approved):

bgilbert (approved):

PR #677, opened 2020-10-09
ci: Use the exttests container

cgwalters:
This is an incremental step to test out running exttests.
But hopefully soon want we do is go multi-pod/multi-stage
and split off exttests separately - we may not always
want to gate on them, etc.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,6 +1,8 @@
 // Documentation: https://github.com/coreos/coreos-ci/blob/master/README-upstream-ci.md
 
-cosaPod {
+// Use the exttests container here in CI to get extra coverage; xref
+// https://github.com/coreos/coreos-assembler/pull/1745
+cosaPod(image: "registry.svc.ci.openshift.org/coreos/cosa-exttests:latest") {
     checkoutToDir(scm, 'config')
 
     shwrap("cd config && ci/validate")


cgwalters:
From Jenkins logs looks like it worked!

> --- PASS: ext.ostree.destructive-rs.transactionality (397.98s)

cgwalters:
To recap with this we have a flow where:

- Change to tests (or new test for new feature) in upstream ostree/ignition/etc git repo is submitted
- Change is tested in CI on that repo
- Change merges
- (Later, or maybe sooner) Submit PR to bump the tag on https://github.com/coreos/coreos-assembler/blob/master/exttests/build.sh (could be automated)
- (hmm, that change should be covered by CI but isn't today)
- New exttests container builds
- New test is run on our main builds too

Or in other words, we run upstream tests on both upstream builds and our pipeline without duplicating the tests.

.cci.jenkinsfile
```diff
@@ -1,6 +1,8 @@
 // Documentation: https://github.com/coreos/coreos-ci/blob/master/README-upstream-ci.md
 
-cosaPod {
+// Use the exttests container here in CI to get extra coverage; xref
+// https://github.com/coreos/coreos-assembler/pull/1745
+cosaPod(image: "registry.svc.ci.openshift.org/coreos/cosa-exttests:latest") {
```
jlebon, line 5: I have a strong preference for having CI keep using the canonical cosa image from quay.io for all the main steps. It's what devs use locally. We've also had multiple encounters with OpenShift CI having stale versions of cosa or rebuilds of layered images like this not triggering. It's a major pain to debug. 

How about instead we add sugar to coreos-ci-lib to have the exttest container be added to the pod and then for the kola run, it knows to use the exttest container just for the installed tests? (I.e. using `container("exttests") { ... }`).

Hmm, or maybe it is indeed easier to just fold this into cosa for now until we do something fancier as discussed in https://github.com/coreos/coreos-assembler/pull/1745.

PR #679, opened 2020-10-13
README: update notes for moving to new Fedora Major

dustymabe:


dustymabe:
I killed the CI run because this is just an update to the README.

README.md
```diff
@@ -111,12 +112,41 @@ CoreOS Installer changes:
...
 
-1. verify that the `f${releasever}-coreos-signing-pending` Koji tag has
-   been created (this should have already been done by releng scripts on
-   branching)
-2. `koji untag` N-2 packages from the pool (at some point we'll have GC
+1. Verify that a few tags have been created. These should have been created
+   by releng scripts on branching: 
+
+- `f${releasever}-coreos-signing-pending`
+- `f${releasever}-coreos-continuous`
```
jlebon, line 119: This one would technically belong better in the SOP for bumping cosa since FCOS itself doesn't use that repo at all. But meh, here WFM too.

README.md
```diff
@@ -111,12 +112,41 @@ CoreOS Installer changes:
...
+
+2. The tag info for the coreos-pool tag has the new release (N) signing
```
jlebon: Should this be N+1 instead? So we stay ahead of the curve. Seems like that's what the example is doing too. :)

README.md
```diff
@@ -111,12 +112,41 @@ CoreOS Installer changes:
...
+- `koji edit-tag coreos-pool -x tag2distrepo.keys="12c944d0 9570ff31 45719a39"`
+
+
+3. `koji untag` N-2 packages from the pool (at some point we'll have GC
    in place to do this for us, but for now we must remember to do this
    manually or otherwise distRepo will fail once the signed packages are
-   GC'ed). For example:
-    - `koji list-tagged coreos-pool --quiet | grep fc30 | cut -f1 -d' ' | sort | uniq`
-    - Sanity-check the output, then pipe it to `xargs koji untag-build coreos-pool`
+   GC'ed). For example the following snippet finds all rpms signed by the
```
jlebon: Minor/optional:

```suggestion
   GC'ed). For example the following snippet finds all RPMs signed by the
```

README.md
```diff
@@ -111,12 +112,41 @@ CoreOS Installer changes:
...
+   Fedora 31 key and untags them.
+
+```
+f31key=3c3359c4
+key=$f31key
+untaglist=''
+for build in $(koji list-tagged --quiet coreos-pool | cut -f1 -d' '); do
+    if koji buildinfo $build | grep $key 1>/dev/null; then
+        untaglist+="${build} "
+        echo "Adding $build to untag list"
```
jlebon, line 145: Do we need this `echo`? We're already printing the full list below.

dustymabe: This set of commands actually takes some time to complete so it's nice to get some feedback along the way. If we remove an echo statement maybe it should be the one after the `for` loop. WDYT?

README.md
dustymabe: Yeah I think we can make it `N+1`

README.md
jlebon: Sure, that makes sense to me!

dustymabe:
OK pushed up a new update to address code review comments.

jlebon (approved):

PR #681, opened 2020-10-13
Drop bits in overlay to chmod (use cosa fix), add test

cgwalters:
The coreos-assembler fix here is stronger, let's
just rely on it and add a test that it's working.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -61,20 +61,6 @@ postprocess:
     setsebool -P -N container_use_cephfs on  # RHBZ#1692369
     setsebool -P -N virt_use_samba on  # RHBZ#1754825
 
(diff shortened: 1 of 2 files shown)


cgwalters:
Requires: https://github.com/coreos/coreos-assembler/pull/1786

(So CI here will likely fail, we can probably merge this tomorrow or so?)

jlebon (approved):

cgwalters:
This is stalling out on
```
oc image info quay.io/coreos-assembler/coreos-assembler:latest | grep Created
Created:     5d ago
```

cgwalters:
OK, CI is green here!  We have one review here.  TBH I'd like to see more people doing reviews on PRs like this...I think we should also consider hooking up Prow so that we have the /lgtm flow to more clearly differentiate between "this looks OK but I'm waiting for another person to approve" versus "merge this".

PR #685, opened 2020-10-14
ignition-ostree-rootfs: use our own tmpfs with size=80%

### LINKED ISSUE #644: F33: ext.config.root-reprovision* tests failing
These tests are failing in the Fedora 33 rebase (coreos/fedora-coreos-config#640). We'll disable the tests for now to unblock.

@jlebon has done some preliminary investigation over in: https://github.com/coreos/fedora-coreos-config/pull/640#issuecomment-703751994

jlebon:
In f33 systemd, /run is now mounted with a smaller size cap instead of
the default 50%:

systemd/systemd#15424

This was causing the rootfs reprovisioning code to run out of space.

Let's use this as an opportunity to be more explicit about the memory
semantics we want. Notably, we don't care about starving the host here;
either we have enough memory to reprovision the rootfs, or we should
fail. So let's just mount our own tmpfs with `size=80%`.

This in turn means that machines need even *less* RAM for rootfs
reprovisioning to work. A local test shows 2.5G seems to be enough.
Though let's keep recommending 4G in the docs to be safe.

A minor thing this also fixes is that we now clean up a leftover
directory and file from `/run` before we switchroot.

Closes: coreos/fedora-coreos-tracker#644

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
@@ -17,6 +17,10 @@ case "${1:-}" in
         fi
         echo "Detected rootfs replacement in fetched Ignition config: /run/ignition.json"
         mkdir "${saved_sysroot}"
+        # use 80% of RAM: we want to be greedy since the boot breaks anyway, but
+        # we still want to leave room for everything else so it hits ENOSPC and
+        # doesn't invoke the OOM killer
+        mount -t tmpfs tmpfs "${saved_sysroot}" -o size=80%
         ;;
     save)
         mount "${rootdisk}" /sysroot
@@ -31,6 +35,12 @@ case "${1:-}" in
         echo "Restoring rootfs from RAM..."
         find "${saved_sysroot}" -mindepth 1 -maxdepth 1 -exec mv -t /sysroot {} \;
         chattr +i $(ls -d /sysroot/ostree/deploy/*/deploy/*/)
+        ;;
+    cleanup)
+        if [ -d "${saved_sysroot}" ]; then
+            umount "${saved_sysroot}"
+            rm -rf "${saved_sysroot}" "${partstate}"
+        fi
         ;;
     *)
         echo "Unsupported operation: ${1:-}" 1>&2; exit 1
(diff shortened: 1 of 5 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -17,6 +17,10 @@ case "${1:-}" in
         fi
         echo "Detected rootfs replacement in fetched Ignition config: /run/ignition.json"
         mkdir "${saved_sysroot}"
+        # use 100% of RAM: we don't care about breaking the boot here (or
+        # another way to say this: we *want* to break the boot if reprovisioning
+        # fails)
+        mount -t tmpfs tmpfs "${saved_sysroot}" -o size=100%
```
cgwalters: Maybe `90%` or even `80%`?  We need to save *some* RAM for the host system...

cgwalters (changes requested):

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
jlebon: That's the thing though. This isn't greedy (it doesn't reserve RAM for the tmpfs), it just caps what the maximum size of the tmpfs can be. Either it's large enough for us or it's not.

jlebon: Let me rephrase this way: there isn't a strong incentive to cap the size given that the semantics we want is already "use as much RAM as we need, or die trying".

cgwalters: Right, but the failure scenario here if we fill up the tmpfs entirely is likely to be the OOM killer striking random critical userspace processes and the system dying in a hard to debug way versus getting a clearer `ENOSPC` from this systemd unit.

cgwalters: Testing this out quickly locally:
```
# mount -t tmpfs tmpfs /run/mytmpfs -o size=100%
# dd if=/dev/urandom of=/run/mytmpfs/blah bs=8M
```

The OOM killer did take out the `dd` the first try but that may be kind of luck?

jlebon: It's pretty clear when it kicks in though:

```
[   10.731379] Out of memory: Killed process 503 (ignition) total-vm:943116kB, anon-rss:2896kB, file-rss:4kB, shmem-rss:10896kB, UID:0 pgtables:140kB oom_score_adj:0
[   10.734876] oom_reaper: reaped process 503 (ignition), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

That said, yes I can imagine a scenario where the OOM killer actually kills some non-critical process which doesn't break the boot outright but has some subtle effect which users only discover later on.

OK, changed to 80%! :arrow_down:

cgwalters (approved):

dustymabe:
Thanks for tracking this down @jlebon

PR #687, opened 2020-10-14
manifests: fedora-coreos-base: add zram-generator

### LINKED ISSUE #509: F33 feature/change proposal SwapOnZRAM by default
This will be proposed for Fedora 33 all editions and spins
https://fedoraproject.org/wiki/Changes/SwapOnZRAM

I'd like to make the case that some swap, especially if it's fast, is better than no swap. In the no swap case, if the system comes under any memory pressure, it means the system must resort to reclaim of file pages because it's not possible to evict even inactive anonymous pages due to lack of swap. Such a system starts to do a kind of swap thrashing, which is really not at all swap related since it doesn't exist, but it's this churn of reading file pages on demand, and then almost immediately they get dropped out of memory. Whereas if there were a swap-on-ZRAM device, those inactive anonymous pages would get evicted, compressed, and free up memory to avoid reclaiming file pages.

Other than totally opting out, I think there are two options Fedora CoreOS could consider:

1. include zram-generator, but not include a configuration file. Without a configuration file, there is no setup of a swap-on-zram device.

2. include both zram-generator and configuration; but you could have a different configuration than the default if you want to go even more minimalist than proposed. e.g. maybe use a fraction of 20% RAM instead of 50%.

Note that Fedora IoT has been using swap-on-ZRAM for some time and they're defaulting to 50% RAM which is the same as this proposal.

dustymabe:
This was part of a F33 proposed change. We'll include the generator
but not the defaults subpackage because we don't want it enabled by
default just yet. We'll add docs for users instructing them how to
enable it by dropping down a file via Ignition/FCCT. The following
snippet is an example:

```
variant: fcos
version: 1.1.0
passwd:
  users:
    - name: core
      ssh_authorized_keys:
        - ssh-rsa AAAAB user@example.com
storage:
  files:
    - path: /etc/systemd/zram-generator.conf
      mode: 0644
      contents:
        inline: |
          # This config file enables a /dev/zram0 device with the default settings
          [zram0]
```

Closes https://github.com/coreos/fedora-coreos-tracker/issues/509

### DIFF
--- a/tests/kola/swap-on-zram/test.sh
+++ b/tests/kola/swap-on-zram/test.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# This test makes sure that swap on zram devices can be set up
+# using the zram-generator as defined in the docs at
+# https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-configure-swaponzram/
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if ! grep -q 'zram0' /proc/swaps; then
+    fatal "expected zram0 to be set up"
+fi
+ok "swap on zram was set up correctly"
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -1130,6 +1130,9 @@
     },
     "zlib": {
       "evra": "1.2.11-21.fc32.x86_64"
+    },
+    "zram-generator": {
+      "evra": "0.2.0-1.fc32.x86_64"
     }
   },
   "metadata": {
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -159,6 +159,9 @@ packages:
   - jq
   # nvme-cli for managing nvme disks
   - nvme-cli
+  # zram-generator (but not zram-generator-defaults) for F33 change
+  # https://github.com/coreos/fedora-coreos-tracker/issues/509
+  - zram-generator
 
 # This thing is crying out to be pulled into systemd, but that hasn't happened
 # yet.  Also we may want to add to rpm-ostree something like arch negation;
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -115,3 +115,11 @@ if [ ! -f /usr/share/rpm/Packages ]; then
     fatal "Didn't find bdb file /usr/share/rpm/Packages"
 fi
 ok rpmdb is bdb
+
+# make sure we don't default to having swap on zram
+# https://github.com/coreos/fedora-coreos-tracker/issues/509
+# https://github.com/coreos/fedora-coreos-config/pull/687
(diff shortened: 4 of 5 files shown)


cgwalters:
This seems like a simple thing to add a test to misc-ro that the unit is not enabled by default.

cgwalters (changes requested):
See above re test

lucab:
There is no service unit to test here (it's a generator), but we can similarly check that by default there are no `/dev/zram*` on the running instance.

dustymabe:
now with a test that fails if `zram0` exists.

```
=== RUN   ext.config.dusty
systemctl status kola-runext.service:
● kola-runext.service
     Loaded: loaded (/etc/systemd/system/kola-runext.service; static; vendor preset: disabled)
     Active: failed (Result: exit-code) since Thu 2020-10-15 14:38:01 UTC; 87ms ago
    Process: 1250 ExecStart=/usr/local/bin/kola-runext-test.sh (code=exited, status=1/FAILURE)
   Main PID: 1250 (code=exited, status=1/FAILURE)

Oct 15 14:38:01 localhost.localdomain systemd[1]: Started kola-runext.service.
Oct 15 14:38:01 localhost.localdomain kola-runext-test.sh[1250]: + '[' -e /dev/zram0 ']'
Oct 15 14:38:01 localhost.localdomain kola-runext-test.sh[1250]: + fatal 'zram0 swap device set up on default install'
Oct 15 14:38:01 localhost.localdomain kola-runext-test.sh[1250]: + echo 'zram0 swap device set up on default install'
Oct 15 14:38:01 localhost.localdomain kola-runext-test.sh[1250]: zram0 swap device set up on default install
Oct 15 14:38:01 localhost.localdomain kola-runext-test.sh[1250]: + exit 1
Oct 15 14:38:01 localhost.localdomain systemd[1]: kola-runext.service: Main process exited, code=exited, status=1/FAILURE
Oct 15 14:38:01 localhost.localdomain systemd[1]: kola-runext.service: Failed with result 'exit-code'.
```

cgwalters (approved):

jlebon (approved):
Stretch-goal is a separate test which enables zram using that FCC you wrote. :) But we can do that as a follow-up! (I'd say we should definitely have that though before it gets added to the docs.)

dustymabe:
Ok added another test to verify configuring zram works too.

tests/kola/swap-on-zram/test.sh
```diff
@@ -0,0 +1,20 @@
...
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if ! grep 'zram0' /proc/swaps >/dev/null; then
```
jlebon: Minor/optional:

```suggestion
if ! grep -q 'zram0' /proc/swaps; then
```

jlebon (approved):
One optional comment, LGTM as is!

tests/kola/swap-on-zram/test.sh
dustymabe: done

PR #694, opened 2020-10-16
tests/kola/ignition: add ignition.systemd-unmasking test

dustymabe:
To make sure ignition can unmask units.

See https://github.com/coreos/ignition/pull/1113

### DIFF
--- a/tests/kola/ignition/systemd-unmasking/config.fcc
+++ b/tests/kola/ignition/systemd-unmasking/config.fcc
@@ -1,0 +1,7 @@
+variant: fcos
+version: 1.1.0
+systemd:
+  units:
+    - name: dnsmasq.service
+      mask: false
+      enabled: true
--- a/tests/kola/ignition/systemd-unmasking/test.sh
+++ b/tests/kola/ignition/systemd-unmasking/test.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# This test makes sure that ignition is able to unmask units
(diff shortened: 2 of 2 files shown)


dustymabe:
blocked on https://github.com/coreos/ignition/pull/1113

bgilbert (approved):
LGTM

jlebon (approved):
LGTM, though for these "configure something via Ignition and check that it happened in `test.sh`" types of tests I made a new `misc-ign-ro` where we can combine these since they're completely independent so that we try to keep the number of VM bring-ups to a reasonable number: https://github.com/coreos/fedora-coreos-config/commit/1f38b844d3b245d3a837edd98a2480475eb8d0a2. This would be a good candidate. Edit: OK I see you have some opinions about that here: https://github.com/coreos/fedora-coreos-config/pull/723#issuecomment-721872431. So let's not block on this.

dustymabe:
adding the `hold` label on this. It won't pass until https://github.com/coreos/ignition/pull/1113 makes it into a release that makes it into FCOS.

dustymabe:
removing hold label, latest ignition should be capable of this now.

dustymabe:
From the [last test run](https://jenkins-coreos-ci.apps.ocp.ci.centos.org/blue/organizations/jenkins/github-ci%2Fcoreos%2Ffedora-coreos-config/detail/PR-694/25/pipeline/56):

```
--- PASS: ext.config.ignition.systemd-unmasking (35.92s)
```

🎉

PR #696, opened 2020-10-19
[next-devel] lockfiles: drop graduated override fedora-repos

jlebon:
This is causing dependency resolution issues in `bump-lockfile` because
the latest `fedora-release` relies on the latest `fedora-repos`.

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -5,18 +5,6 @@ packages:
     evra: 0.7.0-4.fc33.aarch64
   coreos-installer-bootinfra:
     evra: 0.7.0-4.fc33.aarch64
-  # Fast-track fedora-repos-33-0.14 to get archive repo
-  # https://bodhi.fedoraproject.org/updates/FEDORA-2020-868d76a5eb
-  fedora-repos:
-    evra: 33-0.14.noarch
-  fedora-repos-ostree:
-    evra: 33-0.14.noarch
-  fedora-repos-archive:
-    evra: 33-0.14.noarch
-  fedora-repos-modular:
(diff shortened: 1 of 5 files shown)


manifest-lock.overrides.aarch64.yaml
```diff
@@ -5,18 +5,6 @@ packages:
     evra: 0.7.0-4.fc33.aarch64
   coreos-installer-bootinfra:
     evra: 0.7.0-4.fc33.aarch64
-  # Fast-track fedora-repos-33-0.14 to get archive repo
-  # https://bodhi.fedoraproject.org/updates/FEDORA-2020-868d76a5eb
```
dustymabe, line 9: This still shows up as not in stable yet. Because of "freeze" it probably won't go to stable until after release IIUC.

dustymabe:
> This is causing dependency resolution issues in `bump-lockfile` because
> the latest `fedora-release` relies on the latest `fedora-repos`.

The problem is that `fedora-repos-archive` depends on this version (or newer). I think we'll just create a new problem by removing the pin.

manifest-lock.overrides.aarch64.yaml
jlebon: I'm not sure why that Bodhi update wasn't marked obsoleted. 33-1 was pushed to stable in https://bodhi.fedoraproject.org/updates/FEDORA-2020-941b0d5a2e.

dustymabe: Ahh, nice. OK, that solves the problem then :)

dustymabe (approved):
LGTM

dustymabe:
still LGTM

PR #699, opened 2020-10-19
coreos-update-ca-trust: order before first-boot-complete.target

jlebon:
Otherwise, it's possible we'll never run if the system is rebooted after
`systemd-machine-id-commit.service` but before this unit.

This will be recommended by systemd starting from:
https://github.com/systemd/systemd/pull/16939

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/coreos-update-ca-trust.service
+++ b/overlay.d/05core/usr/lib/systemd/system/coreos-update-ca-trust.service
@@ -6,6 +6,14 @@
 [Unit]
 Description=Run update-ca-trust
 ConditionFirstBoot=true
+# All services which use ConditionFirstBoot=yes should use
+# Before=first-boot-complete.target, which is a target that
+# was introduced in https://github.com/systemd/systemd/issues/4511
+# and hasn't propagated everywhere yet. Once the target propagates
+# everywhere, we can drop the systemd-machine-id-commit.service
+# from the Before= line.
+Before=first-boot-complete.target systemd-machine-id-commit.service
+Wants=first-boot-complete.target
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/systemd/system/coreos-update-ca-trust.service
```diff
@@ -6,6 +6,11 @@
 [Unit]
 Description=Run update-ca-trust
 ConditionFirstBoot=true
+# All services which use ConditionFirstBoot=yes should use this.
+# Once the target propagates everywhere, we can drop the service.
+# See: https://github.com/systemd/systemd/issues/4511
+Before=first-boot-complete.target systemd-machine-id-commit.service
+Wants=first-boot-complete.target
```
dustymabe: So `first-boot-complete.target` doesn't exist right now but will in the future?
And when it does exist we can remove `systemd-machine-id-commit.service` ?

If that's the case maybe we can make that a little more clear:

```suggestion
# All services which use ConditionFirstBoot=yes should use 
# Before=first-boot-complete.target, which is a target that
# was introduced in https://github.com/systemd/systemd/issues/4511
# and hasn't propagated everywhere yet. Once the target propagates
# everywhere, we can drop the systemd-machine-id-commit.service
# from the Before= line.
Before=first-boot-complete.target systemd-machine-id-commit.service
Wants=first-boot-complete.target
```

jlebon: Copied verbatim, modulo an errant trailing space.

dustymabe (approved):
LGTM

PR #700, opened 2020-10-19
overlay: 15fcos: add systemd unit to migrate to systemd-resolved

### LINKED ISSUE #646: F33: transitioning existing systems to systemd-resolved on upgrade
With the change to systemd-resolved we need to do some sort of intervention to make the [systemd-resolved change](https://fedoraproject.org/wiki/Changes/systemd-resolved) take effect on existing systems that are upgraded.

The scriptlets for the `systemd` rpm have something for this but "upgrade" logic for OSTree systems doesn't really work because the compose starts with a fresh world view every time:

```
$ rpm -q --scripts systemd
<snip>
# Create /etc/resolv.conf symlink.
# We would also create it using tmpfiles, but let's do this here
# too before NetworkManager gets a chance. (systemd-tmpfiles invocation above
# does not do this, because it's marked with ! and we don't specify --boot.)
# https://bugzilla.redhat.com/show_bug.cgi?id=1873856
if systemctl -q is-enabled systemd-resolved.service &>/dev/null; then
  ln -fsv ../run/systemd/resolve/stub-resolv.conf /etc/resolv.conf
fi
<snip>
```

For us on existing systems the resolv.conf file will already exist and contain some contents like:

```
[core@fedora ~]$ cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 192.168.1.1
```

I suggest we write some migration logic that basically detects the `# Generated by NetworkManager` and runs the `ln -fsv ../run/systemd/resolve/stub-resolv.conf /etc/resolv.conf` if it has that in it. According to the change document, as long as that symlink is set up NetworkManager knows what to do to take advantage of systemd-resolved.

Any `resolv.conf` that had been hand edited and not managed by NetworkManager would be left alone.

dustymabe:
This systemd unit migrates the /etc/resolv.conf file on systems to
point to ../run/systemd/resolve/stub-resolv.conf if users haven't
set up a custom resolv.conf. It will only run on Fedora 33 systems
and will only execute once (a single migration).

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/646


jamescassell:
Can you make a backup at `/etc/resolv.conf.rpmsave`?

dustymabe:
> Can you make a backup at `/etc/resolv.conf.rpmsave`?

I'm not against making a backup but am interested in other opinions. 

If we do make a backup I'm not sure about using the `rpmsave` suffix. It might be misleading since rpm isn't involved here. Other suggestions?

dustymabe:
> > Can you make a backup at `/etc/resolv.conf.rpmsave`?
> 
> I'm not against making a backup but am interested in other opinions.

Though.. maybe I should be against making a backup. We're being careful to only do the migration if the file is "managed by NetworkManager", which means nothing should be in there that isn't in NetworkManager's logic, right?

cgwalters:
I'd be very surprised if any FCOS users had *manually* edited the file; they should be storing their config via Ignition and so they'd need to rework their configs to adapt to this change.  A backup file seems unnecessary.

overlay.d/15fcos/usr/lib/systemd/system/coreos-migrate-systemd-resolved.service
```diff
@@ -0,0 +1,20 @@
...
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/646
+# Run before NetworkManager and systemd-resolved start
+Before=NetworkManager.service
+Before=systemd-resolved.service
+# Ensure we only run once on a F33 system. This file is created
+# by the coreos-migrate-systemd-resolved script
+ConditionPathExists=!/var/lib/coreos-migrate-systemd-resolved-ran
+
+[Service]
+Type=simple
```
cgwalters:
```suggestion
Type=oneshot
```

I don't think we want this to act like a daemon.

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/bash
+
+# This script will migrate a FCOS system running the default configuration
+# to systemd-resolved, which is standard for Fedora 33. For non OSTree Fedora
+# systems the rpm post scriptlet takes care of the transition.
+# https://fedoraproject.org/wiki/Changes/systemd-resolved
+# https://github.com/coreos/fedora-coreos-tracker/issues/646
+
+# If we're not on f33 yet then hard exit.
+source /etc/os-release
```
cgwalters: Mostly pointless nit but `/usr/lib/os-release` I think should be preferred in new code - the "/etc should be empty by default" purists prefer this though I am skeptical of them reaching that goal in the next 5 years :wink: 

(does glibc's resolver support `/run/resolv.conf` yet?)

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
```diff
@@ -0,0 +1,34 @@
...
+# Here we have a bit of logic:
+#
+#   - Does /etc/resolv.conf exist?
+#      - It should since tmpfiles will create it otherwise
+#      - If it doesn't, do nothing.
+#   - Is /etc/resolv.conf currently being managed by NetworkManager?
+#      - This is the default configuration in <F33.
+#      - If yes, then create symlink to stub-resolv.conf
+#   
+if [ -e /etc/resolv.conf ]; then
```
cgwalters:
```suggestion
if [ -f /etc/resolv.conf ]; then
```
maybe?  (`/etc/resolv.conf` shouldn't ever be a directory or fifo but if it somehow is we should just ignore it)

cgwalters (changes requested):

overlay.d/15fcos/usr/lib/systemd/system/coreos-migrate-systemd-resolved.service
dustymabe: Yeah I copied the unit file from another unit (`coreos-liveiso-success.service`) and forgot to update it.

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
dustymabe: updated to `/usr/lib/os-release`. Not sure about the glibc resolver question.

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
dustymabe: fixed

overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
```diff
@@ -3,3 +3,5 @@ enable fedora-coreos-pinger.service
 # Provide information if no ignition is provided
 enable coreos-check-ignition-config.service
 enable coreos-check-ssh-keys.service
+# Enable the migration service to systemd-resolved
+enable coreos-migrate-systemd-resolved.service
```
jlebon: Super minor/optional bikeshed: how about

```suggestion
enable coreos-migrate-to-systemd-resolved.service
```

?

overlay.d/15fcos/usr/lib/systemd/system/coreos-migrate-systemd-resolved.service
```diff
@@ -0,0 +1,20 @@
+# This unit is used to migrate existing systems running default
+# configurations to systemd-resolved as part of the Fedora 33 change.
+# https://fedoraproject.org/wiki/Changes/systemd-resolved
+[Unit]
+Description=CoreOS Migrate to Systemd Resolved
```
jlebon:
```suggestion
Description=CoreOS Migrate to systemd-resolved
```

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/bash
```
jlebon: `set -euo pipefail`

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
```diff
@@ -0,0 +1,34 @@
...
+
+# This script will migrate a FCOS system running the default configuration
+# to systemd-resolved, which is standard for Fedora 33. For non OSTree Fedora
+# systems the rpm post scriptlet takes care of the transition.
```
jlebon: Minor/optional:

```suggestion
# systems the RPM post scriptlet takes care of the transition.
```

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
```diff
@@ -0,0 +1,34 @@
...
+if [ -f /etc/resolv.conf ]; then
+    if [ "$(head -n 1 /etc/resolv.conf)" == '# Generated by NetworkManager' ]; then
+       echo 'Detected that NetworkManager is managing resolv.conf'
+       echo 'Migrating to systemd-resolved for Fedora 33 change'
+       ln -fsv ../run/systemd/resolve/stub-resolv.conf /etc/resolv.conf
+    fi
+fi
+
+# Only run this on Fedora 33 once
+touch /var/lib/coreos-migrate-systemd-resolved-ran
```
jlebon: Using stamp files for this makes sense, though knowing we'll remove this unit eventually and the stamp file will just lay there forever makes me a bit sad. Not sure what the best way to remove it is though. Maybe we can just ship a `tmpfiles.d` drop-in when we remove it to also delete it?

jlebon (approved):
Really just minor nits, LGTM overall!

overlay.d/15fcos/usr/libexec/coreos-migrate-systemd-resolved
dustymabe: I guess another option is to just `systemctl disable` it instead of using a stamp file?

jlebon: Heh, I was thinking about that when I made that comment, and thought it would cause the symlink to linger in `/etc` forever but I think I was confused with `systemctl mask`. `systemctl disable` should just remove the symlink, so yeah you're right I think that'd work.

dustymabe:
all code review comments should be addressed now. I still need to do some testing so please just LGTM, but don't merge.

jlebon (approved):

PR #702, opened 2020-10-20
manifests: fedora-coreos-base: remove temporary support for RSA-SHA1 keys on F33

dustymabe:
We added this in be947c2 because we needed to still be able to run kola
tests on AWS. Now we have a new workaround in COSA [1] so we don't need
to downgrade the policy for SSH in the image.

[1] https://github.com/coreos/coreos-assembler/pull/1797

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -61,27 +61,6 @@ postprocess:
     setsebool -P -N container_use_cephfs on  # RHBZ#1692369
     setsebool -P -N virt_use_samba on  # RHBZ#1754825
 
-  # For F33 let's re-enable RSA-SHA1 keys for now so our kola tests will
-  # work. The plan is to only re-enable this briefly while we wait for
-  # an upstream feature [1] to be implemented. We had moved to an ecdsa
-  # key [2] but AWS doesn't support non RSA keys [3] so we reverted it
-  # for now in [4].
-  #
-  # [1] https://github.com/golang/go/issues/37278#issuecomment-704287991
-  # [2] https://github.com/coreos/coreos-assembler/pull/1749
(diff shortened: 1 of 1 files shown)


dustymabe:
I opened this against `next-devel` for now. Will rebase onto `testing-devel` once the tests pass (which will need https://github.com/coreos/coreos-assembler/pull/1797 before they will pass).

cgwalters:
Maybe I'm missing something here but...isn't this more fundamentally that the crypto policy change is probably going to break a lot of Fedora users in general on AWS?

bgilbert:
> Maybe I'm missing something here but...isn't this more fundamentally that the crypto policy change is probably going to break a lot of Fedora users in general on AWS?

No.  RSA _keys_ are still valid, but the RSA authentication handshake now requires a different hash algorithm.  So the problem only affects old client implementations (the Go SSH client, the Ruby SSH for Vagrant, etc.) that don't support the new algos.

cgwalters:
Right, I understand.  I guess the debate here comes down to the values of "a lot".

cgwalters:
The relationship with AWS is that on other clouds ecdsa or hopefully also ed25519 keys are valid which sidesteps the problem.  Although have we done that analysis on a per-cloud basis?

In a quick spot check it looks like [GCP supports all those key types](https://cloud.google.com/compute/docs/instances/connecting-to-instance).

cgwalters:
https://forums.aws.amazon.com/thread.jspa?threadID=250753

dustymabe:
https://github.com/coreos/coreos-assembler/pull/1797 was obsoleted by https://github.com/coreos/coreos-assembler/pull/1799 (which merged). Kicking off a CI run here.

bgilbert (approved):

dustymabe:
CI passed. Rebasing on `testing-devel` now and if CI passes will merge.

dustymabe:
hmm and it looks like CI gets in a weird state if you switch the target branch. Since CI already passed once I'm going to merge this and deal with the consequences (hopefully none) later.

PR #703, opened 2020-10-20
overlay.d: revert chrony-dhcp temp fix 

sohankunkerkar:
Now that the upstream patch for chrony change has merged (https://src.fedoraproject.org/rpms/chrony/pull-request/3) and
available in F-33, we can safely remove this code, which was a temporary fix to allow chrony propagate NTP servers via DHCP.


dustymabe:
We need to hold this until the new changes (f33) lands in `testing-devel` IIUC.

Also we might need to think about how this affects RHCOS, since it inherits from FCOS and probably will still need the workaround for some time.

jlebon:
Yeah, we could do this cleanly if we folded the overlay semantics into rpm-ostree (because then we could make it a `manifest.yaml` thing which doesn't live on `next-devel`; or even teach rpm-ostree to do conditional includes based on `$releasever`).

One hack for now we could do is have a postprocessing script which deletes those files on f33.

dustymabe:
> Yeah, we could do this cleanly if we folded the overlay semantics into rpm-ostree (because then we could make it a `manifest.yaml` thing which doesn't live on `next-devel`; or even teach rpm-ostree to do conditional includes based on `$releasever`).

:+1:

> 
> One hack for now we could do is have a postprocessing script which deletes those files on f33.

Yeah. Or we could just forklift them over into the RHCOS configs and delete them from here like this commit is doing.

cgwalters:
> Yeah. Or we could just forklift them over into the RHCOS configs and delete them from here like this commit is doing.

Seems fine to me!

dustymabe:
> We need to hold this until the new changes (f33) lands in `testing-devel` IIUC.

no longer blocked on this ^^

dustymabe (approved):
LGTM

PR #701, opened 2020-10-20
[next-devel] overrides: fast-track selinux-policy 3.14.6-29

sohankunkerkar:
related to https://github.com/coreos/fedora-coreos-tracker/issues/643

PR #704, opened 2020-10-20
kola-denylist: remove chrony.dhcp-propagation reference

sohankunkerkar:
Needs https://github.com/coreos/fedora-coreos-config/pull/701
The new version SELinux policy(3.14.6-29) for f-33 fixes an issue with the `chronyd` from reading the sources file that allows chrony to read NTP server data propagated via DHCP.


jlebon (approved):
Should this have a footer to close https://github.com/coreos/fedora-coreos-tracker/issues/643?

dustymabe (approved):
LGTM

PR #708, opened 2020-10-23
manifest: add kexec-tools for kdump

jlebon:
This was discussed in:
https://github.com/coreos/fedora-coreos-tracker/issues/622

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -228,6 +228,9 @@
     "dracut-network": {
       "evra": "050-61.git20200529.fc32.x86_64"
     },
+    "dracut-squash": {
+      "evra": "050-61.git20200529.fc32.x86_64"
+    },
     "e2fsprogs": {
       "evra": "1.45.5-3.fc32.x86_64"
     },
@@ -251,6 +254,9 @@
     },
     "elfutils-libs": {
       "evra": "0.181-1.fc32.x86_64"
+    },
+    "ethtool": {
+      "evra": "2:5.9-1.fc32.x86_64"
     },
     "expat": {
       "evra": "2.2.8-2.fc32.x86_64"
@@ -452,6 +458,9 @@
     },
     "kernel-modules": {
       "evra": "5.8.15-201.fc32.x86_64"
+    },
+    "kexec-tools": {
+      "evra": "2.0.20-17.fc32.x86_64"
     },
     "keyutils": {
       "evra": "1.6-4.fc32.x86_64"
@@ -1026,11 +1035,17 @@
     "slirp4netns": {
       "evra": "1.1.4-1.fc32.x86_64"
     },
+    "snappy": {
+      "evra": "1.1.8-2.fc32.x86_64"
+    },
     "socat": {
       "evra": "1.7.3.4-2.fc32.x86_64"
     },
     "sqlite-libs": {
       "evra": "3.33.0-2.fc32.x86_64"
+    },
+    "squashfs-tools": {
+      "evra": "4.3-25.fc32.x86_64"
     },
     "ssh-key-dir": {
       "evra": "0.1.2-2.fc32.x86_64"
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -152,6 +152,8 @@ packages:
   # zram-generator (but not zram-generator-defaults) for F33 change
   # https://github.com/coreos/fedora-coreos-tracker/issues/509
   - zram-generator
+  # kdump (https://github.com/coreos/fedora-coreos-tracker/issues/622)
+  - kexec-tools
 
 # This thing is crying out to be pulled into systemd, but that hasn't happened
 # yet.  Also we may want to add to rpm-ostree something like arch negation;


dustymabe (approved):
LGTM

manifest-lock.x86_64.json
```diff
@@ -252,6 +255,9 @@
     "elfutils-libs": {
       "evra": "0.181-1.fc32.x86_64"
     },
+    "ethtool": {
```
jlebon, line 258: The only reason this appears to be needed is to handle bonding and teaming: https://src.fedoraproject.org/rpms/kexec-tools/blob/master/f/dracut-module-setup.sh#_189, which is unfortunate. I think we have code which also deals with those, so maybe we can submit a patch to drop that dep.

manifest-lock.x86_64.json
```diff
@@ -1026,12 +1035,18 @@
     "slirp4netns": {
       "evra": "1.1.4-1.fc32.x86_64"
     },
+    "snappy": {
```
jlebon, line 1038: For compressing dump with snappy. Dynamically linked.

manifest-lock.x86_64.json
```diff
@@ -1026,12 +1035,18 @@
...
+      "evra": "1.1.8-2.fc32.x86_64"
+    },
     "socat": {
       "evra": "1.7.3.4-2.fc32.x86_64"
     },
     "sqlite-libs": {
       "evra": "3.33.0-2.fc32.x86_64"
     },
+    "squashfs-tools": {
```
jlebon, line 1047: It uses dracut's built-in support for putting files in a squashfs.

manifest-lock.x86_64.json
```diff
@@ -228,6 +228,9 @@
     "dracut-network": {
       "evra": "050-61.git20200529.fc32.x86_64"
     },
+    "dracut-squash": {
```
jlebon, line 231: It uses dracut's built-in support for putting files in a squashfs.

manifest-lock.x86_64.json
cgwalters: Right but *why* does it do that?  What's the squashfs image?  I am not finding docs for this; the main hit I see so far is:

```
grep -ri squash /usr/lib/dracut/modules.d/99kdumpbase/
/usr/lib/dracut/modules.d/99kdumpbase/module-setup.sh:    is_squash_available() {
/usr/lib/dracut/modules.d/99kdumpbase/module-setup.sh:        for kmodule in squashfs overlay loop; do
/usr/lib/dracut/modules.d/99kdumpbase/module-setup.sh:    if is_squash_available && ! is_fadump_capable; then
/usr/lib/dracut/modules.d/99kdumpbase/module-setup.sh:        _dep="$_dep squash"
/usr/lib/dracut/modules.d/99kdumpbase/module-setup.sh:        dwarning "Required modules to build a squashed kdump image is missing!"
```

but that doesn't help.

jlebon: I'll admit I only actually looked over the dracut code for this to get a sense of why it pulls it in. What I gathered is that it puts the meat of the initramfs files in a squashfs to save space.

OK, some dist-git sleuthing led me to https://src.fedoraproject.org/rpms/kexec-tools/c/9b6e3124477212b8321392b2f365beddd7881e22?branch=master and https://github.com/dracutdevs/dracut/commit/095e1f37c41e8995c95635a47dbe2bf61d8ee2bc, which more or less confirms this. The squashfs is part of the initrd, and dracut mounts it early on and things carry on transparently.

From https://src.fedoraproject.org/rpms/kexec-tools/c/4a44eee472745f8dd43b18815851bd668b7497da?branch=master though, it clearly can work fine without it. So we could try to see if the maintainers are interested in making it a weak dep. I'm not sure if I'd block inclusion into RHCOS for this though.

jlebon: https://src.fedoraproject.org/rpms/kexec-tools/pull-request/4

jlebon: And of course, there's the question of whether we actually want to opt out of those RAM savings. From https://github.com/dracutdevs/dracut/commit/095e1f37c41e8995c95635a47dbe2bf61d8ee2bc, it seems like not a lot, though somewhat significant in proportion to the total amount of RAM to reserve. If that PR gets merged, we can ship without it as a first cut at least.

jlebon:
https://src.fedoraproject.org/rpms/kexec-tools/pull-request/4 was merged now, so in the future `dracut-squash` and `squashfs-tools` should drop out.

liutgnu:
Hello CoreOS team,

I have reverted the commit "Make dracut-squash a weak dep" to make
dracut-squash as Required for kexec-tools again [1].

Although kexec-tools can work without dracut-squash, but it is essential
for kdump to run properly in some cases [2]. If it's the case in CoreOS,
then dracut-squash is needed.

Besides, kdump is evolving. Currently new features [3] of kdump to enhance
vmcore security is in development. If patch were merged, minimal amount
of memory consumption is still expected for kdump, thus dracut-squash is
needed for that.

Thanks!

[1] https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/thread/MX7INNHDHXKS4JXARMSYYHVEV2HOFHID/
[2] https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/message/SJX7CW3WLOYSFI2YJKGTUGDBWSCMZXVZ/
[3] https://www.spinics.net/lists/systemd-devel/msg05864.html

jlebon:
@liutgnu Thanks for following up here. IIUC, this will also solve `crashkernel=auto` not estimating the right amount of memory to reserve.

/cc @kelvinfan001

kelvinfan001:
On our end, we'll need to update the docs to instruct users to reserve less memory for the crash kernel (`auto` should be fine) once `kexec-tools` pulls in `dracut-squash` again.

liutgnu:
Thanks for the updating. @jlebon @kelvinfan001

PR #711, opened 2020-10-26
lockfile: Bump bootupd

cgwalters:
Since we have 0.1.3 in cosa we need the patching version in the
target.


jlebon (approved):
Version is in stable repos already.

jlebon:
Hmm, hitting the same issue here:

```
+ /usr/bin/bootupctl backend install --src-root=/tmp/rootfs/ostree/deploy/fedora-coreos/deploy/ca8254e5b37b6b3a430f02276afa1fb77f9c434aaeca0242696c16fa93dc47ba.0 /tmp/rootfs
error: boot data installation failed: trailing input at line 1 column 35
```

(Which isn't surprising; `bump-lockfile` where I originally saw this already does `cosa fetch --update-lockfile`, which tests with the latest `bootupd` in the repos).

cgwalters:
I reproduced the error locally, and it fixed it for me to drop that rpm into `overrides/rpm`.  Looking.

cgwalters:
OK tentatively I think this is related to https://github.com/coreos/bootupd/pull/29

cgwalters:
bootupd will need to learn how to handle old versions of itself for sure (so far CI is only covering upgrading from "pre-bootupd" and "bootupd 0.2.0" images) but this is actually an inverse because currently it's new bootupd in cosa trying to parse old-bootupd in the ostree.

I have just re-verified again that applying this patch fixes things locally.  I am confused.

cgwalters:
No no wait...the error message here is inverted!  

Notice in [the cosa PR](https://github.com/coreos/coreos-assembler/pull/1806) we have:
>  error: boot data installation failed: premature end of input at line 1 column 34

Yet here it's:

> error: boot data installation failed: trailing input at line 1 column 35

Which makes me think somehow this PR ran with an old cosa (so we have the inverted case).  Retrying this PR.

jlebon:
> Which makes me think somehow this PR ran with an old cosa (so we have the inverted case). Retrying this PR.

Hmm, I think this is Quay caching again biting us. Looking at a cosa pod I provisioned on CentOS CI this morning:

```
$ oc exec -ti coreos-assembler-sleep-jenkins /bin/bash
bash-5.0$ rpm -q bootupd
bootupd-0.1.2-2.fc32.x86_64
```

And looking at the recent Quay builds, none in the last two days were a full rebuild (and the Bodhi update only was pushed to stable two days ago).

So I think we need https://github.com/coreos/coreos-assembler/pull/1809 ?

cgwalters:
That PR merged, restarting CI here.

cgwalters:
Hooray, green CI!

PR #712, opened 2020-10-26
WIP: Add an eln-devel branch

cgwalters:
Currently blocked on:

```
error: Packages not found: adcli, afterburn, afterburn-dracut, bootupd, bsdtar, catatonit, clevis, clevis-dracut, clevis-luks, clevis-systemd, console-login-helper-messages-issuegen, console-login-helper-messages-motdgen, console-login-helper-messages-profile, coreos-installer, coreos-installer-bootinfra, fedora-coreos-pinger, fedora-release-coreos, fedora-repos-archive, fedora-repos-ostree, fuse-overlayfs, fuse-sshfs, ignition, iptables-services, libsss_sudo, libvarlink-util, moby-engine, nfs-utils-coreos, nss-altfiles, nvme-cli, podman-plugins, runc, skopeo, ssh-key-dir, toolbox, wireguard-tools, zincati, zram-generator
```

(And not all of these we may actually want to ship in this hypothetical
 ELN CoreOS)


bgilbert:
This PR was directed to the `testing-devel` branch.  Was that the intention?

Could you add some context to the commit message/PR about what this is and why we want it?

cgwalters:
> This PR was directed to the testing-devel branch. Was that the intention?

Nah, it'd need to be a new branch (like `next-devel`).

> Could you add some context to the commit message/PR about what this is and why we want it?

See https://fedoraproject.org/wiki/Changes/ELN_Buildroot_and_Compose
And then this PR is basically fleshing out what is necessary to make an "ELN CoreOS" which is an important component of influencing the relationship between ELN and CoreOS (and from there also, the relationship between the next RHEL and CoreOS), and the "next RHEL" can (if we make it) relate to the *current* RHEL CoreOS.

bgilbert:
I'd like to see a concrete proposal so we're all on the same page about what the goals are.  Are you proposing a [mechanical, development, or production stream](https://github.com/coreos/fedora-coreos-tracker/blob/master/stream-tooling.md#proposal)?  We originally specified a mechanical Rawhide stream, and a mechanical ELN stream would be the closest analog to that.  If you're proposing something more substantial (i.e. something that would do more than automatically compose and run tests), we should talk about use cases, which probably wants a fedora-coreos-tracker ticket with a `meeting` keyword.

cgwalters:
No intention to try merging this anytime soon.  This would just be a mechanical stream I think to start.

PR #716, opened 2020-10-28
tests/kola/misc-ro: verify that kdump isn't active

jlebon:
For now, it should only be enabled on user request. In the future, it
will be enabled all the time, but still conditional on `crashkernel`:

src.fedoraproject.org/rpms/kexec-tools/pull-request/5


cgwalters (approved):
LGTM

PR #717, opened 2020-10-29
40ignition-conf: remove the legacy base config reference

sohankunkerkar:
If we go ahead with https://github.com/coreos/fedora-coreos-config/pull/676 in its current form, then it not only breaks CI but also fails the distro builds for both RHCOS and FCOS.
1. https://github.com/coreos/fedora-coreos-config/pull/676/commits/90d1f01dcc856da386de4968fc8572816c679113
2. https://github.com/coreos/fedora-coreos-config/pull/676/commits/fc7a6f01117f87a976000a3aa38971a366ce9c7a

So the idea is to carry both the legacy base config and new configs(`15fcos/usr/lib/dracut/modules.d/50ignition-conf-fcos/`) in f-c-c until we cut the next Ignition release.

This PR is not mergeable until we create a new Ignition release (with https://github.com/coreos/ignition/pull/1108 change included in it).


bgilbert (approved):

sohankunkerkar (commented):
With [this change](https://github.com/coreos/fedora-coreos-config/pull/749), we're ready to merge this PR.

PR #720, opened 2020-11-02
[stable] tree: promote changes from testing at 7529172f4e91ea670577c8a15d422a88a06faf44

sohankunkerkar:


jlebon:
```
+ /usr/bin/bootupctl backend install --src-root=/tmp/rootfs/ostree/deploy/fedora-coreos/deploy/9f4587fc047096ad8bb3c0f503d6c34a1450c00b3a1ae22857bc67e691dc8318.0 /tmp/rootfs
error: boot data installation failed: premature end of input at line 1 column 34
```

Hmm right, this is the opposite of https://github.com/coreos/fedora-coreos-config/pull/711 (i.e. new bootupd in cosa but old one in the tree). So either we special-build a cosa with older bootupd which matches the one here, or we cherry-pick #711 too here. Since it basically doesn't do anything on its own yet, I'm inclined to take the easier path and just do the latter. @dustymabe Agreed?

dustymabe:
> we cherry-pick #711 too here. Since it basically doesn't do anything on its own yet, I'm inclined to take the easier path and just do the latter. @dustymabe Agreed?

Assuming that's the only package change it seems safe enough to me.

jlebon:
@sohankunkerkar Want to cherry-pick #711 here?

sohankunkerkar:
> @sohankunkerkar Want to cherry-pick #711 here?

Done!
Should I add a comment here to mention the reason behind cherry-picking this commit?

jlebon (approved):
LGTM!

> Should I add a comment here to mention the reason behind cherry-picking this commit?

Yeah, you can use `cherry-pick -x` and add a reason for the cherry-pick as part of the commit message. OK to proceed as is though!

PR #722, opened 2020-11-03
uuid-root: Add an After= for the device

cgwalters:
This should help with RHCOS updating to the latest, and
is general best practice anyways.


cgwalters:
This is the semi-inverse of https://github.com/openshift/os/pull/425/files#diff-aff6ce12d3d273df5c6b6076b9816dc279f705b5ccd4b028aae072970c816b2cR17 but I think this way is cleaner.

jlebon (approved):

PR #723, opened 2020-11-04
coreos-update-ca-trust: order after local-fs.target

### LINKED ISSUE #660: [next] Found ordering cycle on systemd-machine-id-commit.service/start
**Describe the bug**
Ordering cycle reported by systemd on bootup.

**Reproduction steps**
1. Boot the latest Next FCOS release (33.20201101.1.0).

**System details**
 - QEMU
 - Fedora CoreOS version: Next release 33.20201101.1.0

**Ignition config**
Default COSA config.

**Additional information**

```
Nov 04 13:59:35 fedora systemd[1]: sysinit.target: Found ordering cycle on systemd-machine-id-commit.service/start
Nov 04 13:59:35 fedora systemd[1]: sysinit.target: Found dependency on coreos-update-ca-trust.service/start
Nov 04 13:59:35 fedora systemd[1]: sysinit.target: Found dependency on sysinit.target/start
Nov 04 13:59:36 fedora systemd[1]: Starting Apply Kernel Variables...
Nov 04 13:59:35 fedora systemd[1]: sysinit.target: Job systemd-machine-id-commit.service/start deleted to break ordering cycle starting with sysinit.target/start
Nov 04 13:59:35 fedora systemd[1]: Queued start job for default target Multi-User System.
```

I can provide the full journal log for this boot if needed.

jlebon:
All we really need is `/etc`, so we could run even earlier than
`sysinit.target`.

Drop the comment about `ostree-prepare-root.service` since it's
misleading (that service runs in the initramfs).

This also fixes an ordering cycle between this service and
`systemd-machine-id-commit.service` which wants to run quite early.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/660


jlebon:
Let's wait until we have https://github.com/coreos/coreos-assembler/pull/1834 before merging this.

travier:
> Let's wait until we have [coreos/coreos-assembler#1834](https://github.com/coreos/coreos-assembler/pull/1834) before merging this.

I'm afraid we will have to do it the other way as the cosa PR is currently blocked on that.

travier (approved):
LGTM

jlebon:
OK, and to be safe I also added a test for `coreos-update-ca-trust.service` since we didn't have one yet.

> I'm afraid we will have to do it the other way as the cosa PR is currently blocked on that.

Right yeah, let's do that.

travier (approved):

dustymabe:
I don't really like the `misc-ign-ro` concept and I don't really like moving the existing `swap-on-zram` test into it. The tests are really nice right now because it is so easy to add a new test with a very small scoped world view. `swap-on-zram` was a perfect example of that. `misc-ro` works fine because it's just a big script, but `misc-ign-ro` I think falls down real quick because now you're dealing with a fcct config and you have to consider everything else in the config.

I'd personally prefer reverting the `misc-ign-ro` consolidation and create individual tests for anything that needs an ignition config with only the bits needed for that test. If you really want the added value of running them on a single VM vs spawning multiple then we could look at merging configs together into a final config, but keeping the broken out configs I think has a lot of value.

jlebon:
Sure, we can revert that bit if you'd like. My argument is that they're all really orthogonal tests and it's an easy way to keep down the number of VMs kola brings up. Imagine having e.g. 20 individual tests like this which are just a few FCC lines and a few bash "read-only" lines, that translates to 20 more VMs to bring up, which translates into developer time and CI/build time, esp. when testing on clouds.

> The tests are really nice right now because it is so easy to add a new test with a very small scoped world view.

Heh, I think it makes it *easier* to add others of that type because the files already exist. I agree there's a concern re. making sure tests don't conflict, though I wouldn't want it to grow to e.g. 50 tests like this either. Clearly we'll need some balance.

That said, feel free to answer with "I still want to revert" and I'll go ahead with it.

dustymabe:
There are benefits and negatives to both approaches as we've both emphasized. How about a brainstorm how we can combine them? Maybe we could devise a way to mark tests as `exclusive: false` (default would be true, like it is today) and they can be combined together with other tests on a single VM, but each test is defined separately.

PR #724, opened 2020-11-04
[testing] coreos-update-ca-trust: order after local-fs.target

jlebon:
All we really need is `/etc`, so we could run even earlier than
`sysinit.target`.

Drop the comment about `ostree-prepare-root.service` since it's
misleading (that service runs in the initramfs).

This also fixes an ordering cycle between this service and
`systemd-machine-id-commit.service` which wants to run quite early.

Closes: coreos/fedora-coreos-tracker#660
(cherry picked from commit aa7b539)


jlebon:
Or meh... let's just rerun `promote-config.sh` for these respins since it's basically almost the same anyway.

PR #731, opened 2020-11-09
99emergency-timeout/virtio-dump: add dependency on journald service

Prashanth684:
On ppc64le/s390x I am noticing that there is sort of a race condition where the virtio-dump service starts when the journald
is down. Adding a dependency on it so the unit starts only after journald is up

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/ignition-virtio-dump-journal.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/ignition-virtio-dump-journal.service
@@ -4,6 +4,8 @@ ConditionPathExists=/etc/initrd-release
 DefaultDependencies=false
 ConditionVirtualization=|kvm
 ConditionVirtualization=|qemu
+Requires=systemd-journald.service
+After=systemd-journald.service
 After=basic.target
 
 [Service]


travier:
To start it after journald gets started, you probably want both `Requires` and `After`:
```
Requires=systemd-journald.service
After=systemd-journald.service
```

overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/ignition-virtio-dump-journal.service
```diff
@@ -4,6 +4,7 @@ ConditionPathExists=/etc/initrd-release
 DefaultDependencies=false
 ConditionVirtualization=|kvm
 ConditionVirtualization=|qemu
+Requires=systemd-journald.service
```
travier, line 7: To start it after journald gets started, you probably want both `Requires` and `After`:
```
Requires=systemd-journald.service
After=systemd-journald.service
```

Prashanth684: done. thanks!

jlebon:
Can you describe more what you're seeing? Are you missing journal output in the virtio dump?

jlebon:
The way this works is that this service only kicks in if the initrd goes to `emergency.target`. It shouldn't strictly need the journal running since it just dumps what's in the catalog (though it does do `journalctl --sync`, which I think will start up `journald` via socket activation if it's somehow not running).

Prashanth684:
> Can you describe more what you're seeing? Are you missing journal output in the virtio dump?

so, i found this when the `coreos.ignition.failure` kola test was failing on ppc64l3/s390x and was timing out. I manually did a `cosa run` with the ignition config used in the ignition failure test and saw that the `journalctl --sync` was failing.

```
[    9.119136] systemd[1]: Starting Dump journal to virtio port...
         Starting Dump journal to virtio port...
[    9.119905] systemd[1]: Stopp[ed target System Initialization.  OK  
] Stopped target System Initialization.
  OK  .120491] systemd[1]: Stopp[ed target Swap.
] Stopped target Swap.
[    9.121149] systemd[1]: Stopped Create Volati[le Files and Directories.
ile Systems.ped Create Volatile Files and Directories.
m] Stopped target Local File Systems.arget Local   OK  
[  OK  ] Stopped target Local File Systems (Pre).
[    9.122357] ignition-virtio-dump-journal[963]: + port=/dev/virtio-ports/com.coreos.ignition.journal
[    9.122473] systemd[1]: Stopping Journal Service...
[    9.123308] ignition-virtio-dump-journal[963]: + '[' -e /dev/virtio-ports/com.coreos.ignition.journal ']'
[    9.124782] ignition-virtio-dump-journal[963]: + journalctl -         -sync
Stopping Journal Service...
[    9.125339] systemd[1]: Stopped Apply Kernel Variables.
[    9.125358] systemd[1]: Stopped target Local File Systems (Pre).
...
[    9.177498] systemd[1]: ignition-virtio-dump-journal.service: Main process exited, code=exited, status=1/FAILURE
[    9.178727] systemd[1]: ignition-virtio-dump-journal.service: Failed with result 'exit-code'.
[    9.179668] systemd[1]: Failed to start Dump journal to virtio port.
[FAILED] Failed to start Dump journal to virtio port.
```
and the systemctl status 

```
● ignition-virtio-dump-journal.service - Dump journal to virtio port
   Loaded: loaded (/usr/lib/systemd/system/ignition-virtio-dump-journal.service>
   Active: failed (Result: exit-code) since Tue 2020-11-10 15:45:36 UTC; 3min 1>
  Process: 963 ExecStart=/usr/bin/ignition-virtio-dump-journal (code=exited, st>
 Main PID: 963 (code=exited, status=1/FAILURE)

Nov 10 15:45:36 localhost systemd[1]: Starting Dump journal to virtio port...
Nov 10 15:45:36 localhost ignition-virtio-dump-journal[963]: + port=/dev/virtio>
Nov 10 15:45:36 localhost ignition-virtio-dump-journal[963]: + '[' -e /dev/virt>
Nov 10 15:45:36 localhost ignition-virtio-dump-journal[963]: + journalctl --sync
```

BTW, if i remove the sync , it works fine.

Prashanth684:
> The way this works is that this service only kicks in if the initrd goes to `emergency.target`. It shouldn't strictly need the journal running since it just dumps what's in the catalog (though it does do `journalctl --sync`, which I think will start up `journald` via socket activation if it's somehow not running).

yes - which is was i thought too but for some reason the sync command was failing . i just experimented manually stopping  journald and running sync and it gave me:
```
:/# journalctl --sync
Failed to kill journal service: No main process to kill
```

jlebon:
Ahh OK, in that case let's just do something like

```sh
journactl --sync || : 
```

?

I.e. let's not error out if we fail to tell the journal to sync since it likely means it's already down, which means all the journal entries we'll get are already on disk (well, really in the `/run` tmpfs).

Prashanth684:
> Ahh OK, in that case let's just do something like
> 
> ```shell
> journactl --sync || : 
> ```
> 
> ?
> 
> I.e. let's not error out if we fail to tell the journal to sync since it likely means it's already down, which means all the journal entries we'll get are already on disk (well, really in the `/run` tmpfs).

hmm..i tried this change and it looks like the sync doesn't error out but is actually stuck? not sure what's going on because the unit reports success:

```
:/# systemctl status ignition-virtio-dump-journal.service 
● ignition-virtio-dump-journal.service - Dump journal to virtio port
   Loaded: loaded (/usr/lib/systemd/system/ignition-virtio-dump-journal.service>
   Active: active (exited) since Tue 2020-11-10 16:51:14 UTC; 1min 46s ago
  Process: 972 ExecStart=/usr/bin/ignition-virtio-dump-journal (code=exited, st>
 Main PID: 972 (code=exited, status=0/SUCCESS)
    Tasks: 0 (limit: 11904)
   Memory: 0B
   CGroup: /system.slice/ignition-virtio-dump-journal.service

Nov 10 16:51:13 localhost systemd[1]: Starting Dump journal to virtio port...
Nov 10 16:51:13 localhost ignition-virtio-dump-journal[972]: + port=/dev/virtio>
Nov 10 16:51:13 localhost ignition-virtio-dump-journal[972]: + '[' -e /dev/virt>
Nov 10 16:51:13 localhost ignition-virtio-dump-journal[972]: + journalctl --sync
```
But the coreos.ignition.failure test still fails with this change.

jlebon:
> hmm..i tried this change and it looks like the sync doesn't error out but is actually stuck? not sure what's going on because the unit reports success:

Did all the logs get printed out? I wonder if it just seems like it's stuck at `journalctl --sync` because the journal stopped so it didn't collect any more output from the script afterwards. Is it reproducible?

> i just experimented manually stopping journald and running sync and it gave me:
> 
> ```
> :/# journalctl --sync
> Failed to kill journal service: No main process to kill
> ```

Huh, for me it's

```
sh-5.0# journalctl --sync
Failed to connect to /run/systemd/journal/io.systemd.journal: Connection refused
```

But either way, it seems like it's doing what we expect at least (returning early with an error).

Prashanth684:
> > hmm..i tried this change and it looks like the sync doesn't error out but is actually stuck? not sure what's going on because the unit reports success:
> 
> Did all the logs get printed out? I wonder if it just seems like it's stuck at `journalctl --sync` because the journal stopped so it didn't collect any more output from the script afterwards. Is it reproducible?
> 
> > i just experimented manually stopping journald and running sync and it gave me:
> > ```
> > :/# journalctl --sync
> > Failed to kill journal service: No main process to kill
> > ```
> 
> Huh, for me it's
> 
> ```
> sh-5.0# journalctl --sync
> Failed to connect to /run/systemd/journal/io.systemd.journal: Connection refused
> ```
> 
> But either way, it seems like it's doing what we expect at least (returning early with an error).

is there an easy way to dump logs from the virtio channel? the kola test fails so i assume it does not dump the logs. this consistently happens on ppc64le/s390x.

Prashanth684:
> > > hmm..i tried this change and it looks like the sync doesn't error out but is actually stuck? not sure what's going on because the unit reports success:
> > 
> > 
> > Did all the logs get printed out? I wonder if it just seems like it's stuck at `journalctl --sync` because the journal stopped so it didn't collect any more output from the script afterwards. Is it reproducible?
> > > i just experimented manually stopping journald and running sync and it gave me:
> > > ```
> > > :/# journalctl --sync
> > > Failed to kill journal service: No main process to kill
> > > ```
> > 
> > 
> > Huh, for me it's
> > ```
> > sh-5.0# journalctl --sync
> > Failed to connect to /run/systemd/journal/io.systemd.journal: Connection refused
> > ```
> > 
> > 
> > But either way, it seems like it's doing what we expect at least (returning early with an error).
> 
> is there an easy way to dump logs from the virtio channel? the kola test fails so i assume it does not dump the logs. this consistently happens on ppc64le/s390x.

I see this passing intermittently . Sometimes, the journal logs are dumped out, but most times, it is just EOF. Not sure why this is platform specifc.

Prashanth684:
> > > > hmm..i tried this change and it looks like the sync doesn't error out but is actually stuck? not sure what's going on because the unit reports success:
> > > 
> > > 
> > > Did all the logs get printed out? I wonder if it just seems like it's stuck at `journalctl --sync` because the journal stopped so it didn't collect any more output from the script afterwards. Is it reproducible?
> > > > i just experimented manually stopping journald and running sync and it gave me:
> > > > ```
> > > > :/# journalctl --sync
> > > > Failed to kill journal service: No main process to kill
> > > > ```
> > > 
> > > 
> > > Huh, for me it's
> > > ```
> > > sh-5.0# journalctl --sync
> > > Failed to connect to /run/systemd/journal/io.systemd.journal: Connection refused
> > > ```
> > > 
> > > 
> > > But either way, it seems like it's doing what we expect at least (returning early with an error).
> > 
> > 
> > is there an easy way to dump logs from the virtio channel? the kola test fails so i assume it does not dump the logs. this consistently happens on ppc64le/s390x.
> 
> I see this passing intermittently . Sometimes, the journal logs are dumped out, but most times, it is just EOF. Not sure why this is platform specifc.

@jlebon any thoughts? i'm not sure how to proceed to make this work for ppc64le/s390x..

jlebon:
@Prashanth684 Sorry for the delay, was AFK yesterday.

Re. https://github.com/coreos/fedora-coreos-config/pull/731#issuecomment-724858741, I think the difference there is due to el8 systemd using signals to tell the journal to sync vs Fedora systemd using varlink (so using sockets).

Though still, I'm not sure why using `|| :` wouldn't work. I wonder if there's a PID re-use issue possible there where it successfully sends the signal, but the journal exited and so another process gets it and the client just waits forever because it's not getting the notification from the journal that it expected?

Anyway, if the patch you have here fixes it, then I'm fine with it. Clearly, the script expects journald to be up today for `journalctl --sync`. It just seems funny to actually wake it up since it must've already synced entries if it had exited, but meh...

Restarted CI on this!

Prashanth684:
> @Prashanth684 Sorry for the delay, was AFK yesterday.
> 
> Re. [#731 (comment)](https://github.com/coreos/fedora-coreos-config/pull/731#issuecomment-724858741), I think the difference there is due to el8 systemd using signals to tell the journal to sync vs Fedora systemd using varlink (so using sockets).
> 
> Though still, I'm not sure why using `|| :` wouldn't work. I wonder if there's a PID re-use issue possible there where it successfully sends the signal, but the journal exited and so another process gets it and the client just waits forever because it's not getting the notification from the journal that it expected?
> 
> Anyway, if the patch you have here fixes it, then I'm fine with it. Clearly, the script expects journald to be up today for `journalctl --sync`. It just seems funny to actually wake it up since it must've already synced entries if it had exited, but meh...
> 
> Restarted CI on this!

no worries! yes the PID re-use case might be very possible as i don't see this issue when i stop journal and run the script manually. I do see the rhcos pipeline failing intermittently but also succeeding sometimes, but having this fix would be better. thanks!!

jlebon:
CI here needs https://github.com/coreos/coreos-ci-lib/pull/56.

jlebon (approved):

PR #733, opened 2020-11-10
coreos-copy-firstboot-network: order after coreos-enable-network

jlebon:
In a previous iteration, the "conditional networking" logic lived in
`ignition-fetch-offline.service`. This was later moved to its own
service in `coreos-enable-network.service`, but we never updated the
ordering here.

The end result is that sometimes `coreos-enable-network.service` would
run *after* `coreos-copy-firstboot-network.service`, which meant that we
lost any NM config which was copied from `/boot` (i.e. from
`coreos-installer install --copy-network`).

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1895979


jlebon:
A workaround for this is to boot the node with `rd.neednet=1` to disable the conditional networking logic.

bgilbert (approved):
LGTM

dustymabe (approved):
LGTM

PR #734, opened 2020-11-11
[next-devel] manifest: point to release repos

dustymabe:
The f33 repos are no longer in the `development/` path. Use the release
repos now.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -12,7 +12,7 @@ repos:
   # these repos are there to make it easier to add new packages to the OS and to
   # use `cosa fetch --update-lockfile`; but note that all package versions are
   # still pinned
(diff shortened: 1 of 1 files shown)


nasirhm (approved):
LGTM, is there a seperate repo as `fedora-next` for FCOS next stream ?

dustymabe:
> LGTM, is there a seperate repo as `fedora-next` for FCOS next stream ?

`fedora-next` just references https://github.com/coreos/fedora-coreos-config/blob/testing-devel/fedora-next.repo#L4-L16

When Fedora is in development (right before a major release) that's the location where repos are held.

miabbott (approved):
LGTM

PR #736, opened 2020-11-13
15coreos-network: Remove 15-coreos-firstboot-network.conf

kelvinfan001:
https://github.com/coreos/fedora-coreos-config/pull/659 attempts to
mount `/boot` read-only. `15-coreos-firstboot-network.conf`'s job
should be handled by `ignition-firstboot-complete.service` which
remounts writable (privately/temporarily) `/boot` from the real root.


kelvinfan001:
Hold until https://github.com/coreos/ignition/pull/1124 is merged and a new Ignition is in FCOS.

PR #738, opened 2020-11-17
tree: promote changes from testing-devel at 725a63ce406a1c868f1252ed9…

travier:
…36a46ab26febee9


lucab (approved):

travier:
This failed on:
```
error: Couldn't find locked package 'kernel-modules-5.9.8-200.fc33.x86_64' (pkgs matching NEVRA: 0; mismatched checksums: 0)
```

dustymabe:
Since we're switching over to F33 we need to update the manifest (that file doesn't get updated by the bot because it's typically the only file that carries a diff between the branches). We need to make a similar change to manifest.yaml as was done in 09a4df3 (preferably in a second commit).

dustymabe (approved):
LGTM

jlebon (approved):

jlebon:
> ```
> error: Couldn't find locked package 'kernel-modules-5.9.8-200.fc33.x86_64' (pkgs matching NEVRA: 0; mismatched checksums: 0)
> ```

There's a distRepo running right now: https://koji.fedoraproject.org/koji/taskinfo?taskID=55747500. Try restarting CI once it's done.

PR #740, opened 2020-11-18
Move buildroot image here

jlebon:
When using cosa to build OSes, we've been pretty good at shielding
ourselves from versioning differences between source and target
environments (for example, RHCOS is built by Fedora packages). But when
it comes to building things meant to run *in* that OS, we really should
match the buildroot of the target.

Right now, many of our projects use the buildroot cosa container to
build their projects in CI. But this doesn't necessarily match the
buildroot of the rest of the OS.

Since the buildroot to use is defined here (specifically via the
`releasever` key), let's move the concept of a "buildroot image" from
cosa to here. Specifically, we want a container image with all the deps
to the components we mostly hack on for use by both CI and local
developers.

This can then be used by CoreOS CI for building upstream projects, as
well as by cosa itself for building kolet.

See also: https://github.com/coreos/coreos-assembler/issues/1863

### DIFF
--- a/ci/buildroot/buildroot-buildreqs.txt
+++ b/ci/buildroot/buildroot-buildreqs.txt
@@ -1,0 +1,9 @@
+# This is what the CoreOS developers tend to actively develop/own.
+# If you want to extend this, feel free to file a PR.
+ignition
+ostree
+librepo
+kernel
+systemd
+dracut
+podman
--- a/ci/buildroot/buildroot-reqs.txt
+++ b/ci/buildroot/buildroot-reqs.txt
@@ -1,0 +1,39 @@
+# This is a list of basic buildrequires; it'd be a bit better to
+# yum -y install @buildsys-build but unfortunately that hits a bug:
+# https://fedoraproject.org/wiki/Common_F30_bugs#Conflicts_between_fedora-release_packages_when_installing_package_groups
+# So here we inline it, minus the -release package.
+bash
+bzip2
+coreutils
+cpio
+diffutils
+findutils
+gawk
+glibc-minimal-langpack
+grep
+gzip
+info
+make
+patch
+redhat-rpm-config
+rpm-build
+sed
+shadow-utils
+tar
+unzip
+util-linux
+which
+xz
+
+# For rust projects like rpm-ostree
+rustfmt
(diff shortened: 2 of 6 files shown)


jlebon:
This still needs some fleshing out around how this solves the kolet case, though let's keep the discussions about that in https://github.com/coreos/coreos-assembler/issues/1863.

cgwalters (approved):
Idea here makes a lot of sense to me!

ci/buildroot/buildroot-buildreqs.txt
```diff
@@ -0,0 +1,9 @@
+# This is what the CoreOS developers tend to actively develop/own.
+# If you want to extend this, feel free to file a PR.
+ignition
+ostree
```
cgwalters, line 4: No `rpm-ostree`?

jlebon:
OK, rebased and ready to go! Tested locally with `sudo podman build -t fcos-buildroot .`.

ci/buildroot/buildroot-buildreqs.txt
jlebon: It's covered via `buildroot-specs.txt`.

ci/buildroot/buildroot-reqs.txt
```diff
@@ -0,0 +1,39 @@
+# This is a list of basic buildrequires; it'd be a bit better to
```
cgwalters, line 1: Just to xref: https://github.com/coreos/rpm-ostree/issues/2326#issuecomment-781490950

cgwalters:
That said there is a somewhat profound thing going on here in *not* deriving from coreos-assembler because some of our CI may actually rely on tools in cosa but not in this buildroot.

But...let's try it out!

cgwalters (approved):

cgwalters:
> That said there is a somewhat profound thing going on here in not deriving from coreos-assembler because some of our CI may actually rely on tools in cosa but not in this buildroot.

(Which...is going to lead to the question of whether we should in some cases ship some parts of cosa as an RPM)

jlebon:
Skipping CI since it's separate files which don't affect the compose at all.

PR #743, opened 2020-11-19
30ignition-coreos: Add `coreos-boot-edit.{service,sh}` 

kelvinfan001:
`coreos-boot-edit.service` will run late in the initrd process after
Ignition is completed successfully and temporarily mount /boot read write
to make edits (e.g. removing firstboot networking configuration
files if necessary).

We would like to mount `/boot` read-only in the real root,
so remove the current 15-coreos-firstboot-network.conf since
it would not work once `/boot` is mounted ro. Drop a stamp
file instead so that `coreos-boot-edit.service` would notice
and perform the clean up later in the initramfs.

xref https://github.com/coreos/fedora-coreos-config/pull/659


overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Mount /boot. Note that we mount /boot but we don't unmount it because we
+# are run in a systemd unit with MountFlags=slave so it is unmounted for us.
+mount -o rw ${bootdev} ${bootmnt}
+
+if [ -n "$(ls -A ${initramfs_firstboot_network_dir} 2>/dev/null)" ]; then
+    # Clean up the files that were placed into boot (most likely by coreos-installer)
+    rm -r ${initramfs_firstboot_network_dir}
+else
+    echo "info: no firstboot networking config files to clean from /boot. skipping"
```
kelvinfan001: Is this info necessary? Otherwise, would it be cleaner to just `rm -rf` and skip the conditional?

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.service
```diff
@@ -0,0 +1,25 @@
...
+# Since we are mounting /boot/, require the device first
+Requires=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device
+# Start after Ignition has finished
+After=ignition-files.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/coreos-cleanup-firstboot-network
+RemainAfterExit=no
```
jlebon: Is there a reason for `no`? I think we should default to `yes` unless there's a reason to use `no`.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.service
```diff
@@ -0,0 +1,25 @@
...
+# networking config files if the user copied them into the installed
+# system (most likely by using `coreos-installer install --copy-network`).
+# We want to make sure the firstboot networking config files in /boot are
+# cleaned up only after Ignition completed successfully. 
+
+[Unit]
+Description=CoreOS Clean Up Firstboot Networking Config
+ConditionPathExists=/usr/lib/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
```
jlebon: How about having `coreos-copy-firstboot-network.sh` drop a stamp file in `/run` somewhere if it found configs to copy and then using `ConditionPathExists` here?

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.sh
jlebon: I think if we use a stamp file instead we can just use `rm -vrf` instead.

dustymabe:
hmm. I wonder if we could just do something similar to what `coreos-teardown-initramfs.service` is doing.. 

https://github.com/coreos/fedora-coreos-config/blob/599311b477f62b999bc923519f1920e04787ef09/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.service

i.e. there was a lot of logic put in there to make sure we ran after Ignition was complete, we may want to re-use some of it.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
```
jlebon: Really tempting to just merge this script into the main copy network one and e.g. pass `copy` or `cleanup` to it from the systemd services.

Though cool as is too. Like I alluded to in https://github.com/coreos/ignition/pull/1124, I think eventually we should rework this so that we only mount `/boot` rw once, do everything we need to do there, and then unmount. So we'd change how this logic is wired anyway at that point if we do that.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.service
kelvinfan001: I might've misunderstood how `RemainAfterExit` should be used. My understanding was based off of this. https://stackoverflow.com/a/38377069:
"Use RemainAfterExit=yes for services, which somehow change state of the system. When you want that state reverted, you just stop the service. Use RemainAfterExit=no for services, which do some action, but do not change the state of the system. An example would be a service to cleanup /tmp."

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.service
kelvinfan001: The stamp file makes sense.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.sh
kelvinfan001: Right, this would eventually be replaced or merged into the final "post-ignition" service where all the stuff we want to do to `/boot` is performed in a single rw mount, so maybe it's simpler to have this separate already?

kelvinfan001:
> hmm. I wonder if we could just do something similar to what `coreos-teardown-initramfs.service` is doing..
> 
> https://github.com/coreos/fedora-coreos-config/blob/599311b477f62b999bc923519f1920e04787ef09/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.service
> 
> i.e. there was a lot of logic put in there to make sure we ran after Ignition was complete, we may want to re-use some of it.

@dustymabe sorry I didn't really follow. Did you mean that the current `After=ignition-files.service` is insufficient to ensure that Ignition was complete?

dustymabe:
> > hmm. I wonder if we could just do something similar to what `coreos-teardown-initramfs.service` is doing..
> > https://github.com/coreos/fedora-coreos-config/blob/599311b477f62b999bc923519f1920e04787ef09/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.service
> > i.e. there was a lot of logic put in there to make sure we ran after Ignition was complete, we may want to re-use some of it.
> 
> @dustymabe sorry I didn't really follow. Did you mean that the current `After=ignition-files.service` is insufficient to ensure that Ignition was complete?

It should be fine, but to me it would be nice to only clean up these files once all stages of Ignition have run. Maybe `After=ignition-files.service` is good enough. @jlebon knows way more about this than I do, so I'll trust it if he's good with what you've proposed.

kelvinfan001:
ah got it. I guess it wouldn't hurt to be safer?

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.sh
dustymabe: @jlebon - any reason not to make the "post-ignition" service now and just start with this?

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.service
jlebon: That [second answer](https://stackoverflow.com/a/55998992/308136) there mentions a big reason for using it:

> This is particularly useful when you are adding dependency on services during bootup. wherein the service status is important to start dependent services.

If you do `After=` and `Requires=service-which-does-RemainAfterExit-no`, systemd will try to activate the service again, and if it's not a service you actually want to run multiple times, it can cause issues (see https://github.com/ostreedev/ostree/pull/1697).

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.sh
jlebon: No reason, we could do this now AFAICT. E.g. a `coreos-boot-edit.service`?

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.service
kelvinfan001: Ohh okay, understood.

jlebon:
> It should be fine, but to me it would be nice to only clean up these files once all stages of Ignition have run. Maybe After=ignition-files.service is good enough.

Meh, I'm fine either way really. Like I mentioned in IRC, I think a fundamental difference here from the network teardown one (and the `umount` stage) is that we want code which require these services not to have to think about their shutdown (and e.g. it's useful to have network/mounts still up in `rd.break`). I don't think the same argument applies here so personally I'd just keep it simple.

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-cleanup-firstboot-network.sh
kelvinfan001: Now that we're putting this inside `coreos-boot-edit.service`, there could be cases where there is nothing to clean up when this script is executed. But `rm -vrf ${initramfs_firstboot_network_dir}` should still be safe right?

jlebon: Yup, agreed! :+1:

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.service
```diff
@@ -0,0 +1,24 @@
+# This unit will run late in the initrd process after Ignition is completed
+# successfully and temporarily mount /boot read-only to make edits in /boot
```
jlebon:
```suggestion
# successfully and temporarily mount /boot read-write to make edits
```

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.service
```diff
@@ -0,0 +1,24 @@
...
+# (e.g. removing firstboot networking configuration files if necessary).
+
+[Unit]
+Description=CoreOS Boot Edit
+ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=/run/coreos-copy-firstboot-network.stamp
```
jlebon: Hmm, since this is meant to be a generic unit, maybe let's put the conditional on that stamp file in the script itself?

We could also keep it here, but make it [a triggering condition](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Conditions%20and%20Asserts) to make it clear that more things will be able to trigger this service in the future.

kelvinfan001: oops, yeah I meant for it to be a triggering condition. But actually putting the conditional in the script itself might be better to keep the service file cleaner.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/module-setup.sh
```diff
@@ -36,4 +36,11 @@ install() {
     # units only started when we have a boot disk
     # path generated by systemd-escape --path /dev/disk/by-label/root
     install_ignition_unit coreos-gpt-setup.service ignition-diskful.target
+
+    inst_script "$moddir/coreos-boot-edit.sh" \
+        "/usr/sbin/coreos-boot-edit"
```
jlebon, line 41: Hmm, this should probably be `/usr/libexec` instead, but not worth a respin! :)

jlebon (approved):
LGTM!

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.service
```diff
@@ -0,0 +1,23 @@
...
+
+[Unit]
+Description=CoreOS Boot Edit
+ConditionPathExists=/usr/lib/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+# Since we are mounting /boot, require the device first
+Requires=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device
```
travier, line 13: I don't think that's needed with `After=ignition-files.service`. But this probably does not hurt either.

travier (commented):
LGTM overall but is there any specific reason to not directly always remove the files? I don't think we need to be extra careful here.

kelvinfan001:
@travier I think you're right. The stamp file seems a bit unnecessary now. It was originally there because we could potentially not run the service that removes those files if there are no files to clean up in `/boot`. But then we decided to merge the firstboot network config clean up service into a single `coreos-boot-edit.service` that will be in charge of doing all the stuff we want to do to `/boot` after Ignition and probably needs to be run every time.
Anyways, going to merge for now. The extra carefulness shouldn't hurt, right?

travier:
> The extra carefulness shouldn't hurt, right?

Unnecessary complexity may lead to issues but things here looks fine. We will revisit later if needed.

jlebon:
I think the stamp file is useful so that we can make this whole unit conditional: #745.

PR #744, opened 2020-11-19
manifests/ignition-and-ostree: Disable systemd-repart

travier:
Ignition is responsible for partition setup on first boot and does not use systemd-repart currently.
See also https://github.com/coreos/fedora-coreos-tracker/issues/570

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -68,6 +68,12 @@ postprocess:
   - |
     #!/usr/bin/env bash
     systemctl mask dnsmasq.service
+  # Mask systemd-repart. Ignition is responsible for partition setup on first
+  # boot and does not use systemd-repart currently. See also
+  # https://github.com/coreos/fedora-coreos-tracker/issues/570
+  - |
+    #!/usr/bin/env bash
+    systemctl mask systemd-repart.service
 
 packages:
   # Security
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -135,3 +135,10 @@ if [ $(systemctl is-enabled dnsmasq.service) != 'masked' ]; then
     fatal "dnsmasq.service systemd unit should be masked"
 fi
 ok "dnsmasq.service systemd unit is masked"
+
+# make sure systemd-repart is masked
+# https://github.com/coreos/fedora-coreos-config/pull/744
+if [ $(systemctl is-enabled systemd-repart.service) != 'masked' ]; then
+    fatal "systemd-repart.service systemd unit should be masked"
+fi
+ok "systemd-repart.service systemd unit is masked"


dustymabe:
I think I'd prefer to not remove more things from the systemd package if it's not harmful to keep it in. Even for the other things in this file that we've chosen to remove we maybe should consider the alternative.

We got bit by this a bit with ripping out `systemd-networkd`. A good middle ground is trying to get the package maintainers to make it a subpackage if there's good reason to not include it in some cases.

cgwalters:
The thing is, like other systemd stuff it's designed with a vision of block device setup that conflicts with ostree; it also heavily overlaps with Ignition.

The other middle ground may be like networkd; keep the binary but add `ConditionPathExists=/etc/coreos-systemd-repart-i-know-what-im-doing` or so.

LorbusChris:
I'm with Dusty on this. I think we should push the RPM maintainers towards more finely grained subpackages in general.

travier:
I understand the concerns but in this case, having it as a sub-package does not really have advantages as it is very unlikely that anyone would want to overlay it in contrast with `systemd-networkd`. The only use case I could think of would be to call the binary with custom arguments to manually re-partition another disk.

I have pushed a new version that only removes the units as they conflict with the way Ignition works but keeps the binary to avoid the "I want to use it and I can't overlay it" issue from `systemd-networkd`.

If we decide to go the route of split packages then I would argue that we should split `systemd-firstboot` and the GPT generator out of the `systemd` and `systemd-udev` packages (see same file) and I don't think this would make sense.

cgwalters (approved):

dustymabe:
> @cgwalters 
> The thing is, like other systemd stuff it's designed with a vision of block device setup that conflicts with ostree; it also heavily overlaps with Ignition.

Then other `ostree` systems are going to have this same problem? Why not try to solve it more generically in systemd?

> @cgwalters 
> The other middle ground may be like networkd; keep the binary but add `ConditionPathExists=/etc/coreos-systemd-repart-i-know-what-im-doing` or so.

Right. My counter proposal is that we either try to solve this more generically upstream in systemd (`ConditionPathExists=! /run/ostree-booted`?) or keep the units (the size isn't large), but just neuter them and document why they are disabled and how to un-disable them. We did this [recently for dnsmasq](https://github.com/coreos/fedora-coreos-docs/pull/197). 

Looking at `systemd-repart.service`, in what scenario currently do they get run?

```
ConditionDirectoryNotEmpty=|/usr/lib/repart.d
ConditionDirectoryNotEmpty=|/usr/local/lib/repart.d
ConditionDirectoryNotEmpty=|/etc/repart.d
ConditionDirectoryNotEmpty=|/run/repart.d
```

Is there a generator that puts files in one of those directories?

jlebon:
Let's just match exactly what we do for dnsmasq? (I.e. `systemctl mask` it and document it). I think it's helpful to be consistent in how we tackle these "discouraged" services.

(On that topic, rpm-ostree could probably learn a e.g. `masked-units` treefile entry for this to make it nicer to work with.)

travier:
Note that in this case, the system package enables those by default:
```
/usr/lib/systemd/system/initrd-root-fs.target.wants/systemd-repart.service
/usr/lib/systemd/system/sysinit.target.wants/systemd-repart.service
```
Which is the part I think should go away. I don't really care about the unit or the binary. Maybe this can be moved to a preset.

jlebon:
> Note that in this case, the system package enables those by default:

Looking at https://github.com/systemd/systemd/blob/master/units/systemd-repart.service.in, it doesn't have an `[Install]` section, which means that those static enablement links you quoted are all it has. I don't *think* using presets will work for this. Maybe let's leave those alone since it makes it significantly harder to re-enable via Ignition than just doing `masked: false`?

travier:
> Maybe let's leave those alone since it makes it significantly harder to re-enable via Ignition than just doing `masked: false`?

But I don't think we want to support that when we already have Ignition doing re-partitioning.

> I don't think using presets will work for this

Indeed, as this also impacts the initramfs and thus can not be disabled via Ignition nor presets, I think that we should either do this PR or do the RPM split.

jlebon (approved):
>> Maybe let's leave those alone since it makes it significantly harder to re-enable via Ignition than just doing masked: false?
>
> But I don't think we want to support that when we already have Ignition doing re-partitioning.

I think we're all in agreement that we want to discourage using `systemd-repart`. :) I think what we're debating now is how we want to convey that to users. We've already set a precedent for this sort of situation [when masking dnsmasq and documenting it](https://github.com/coreos/fedora-coreos-docs/pull/197). I'm suggesting following that same pattern here.

IMO, masking the unit is cleaner than removing the enablement symlinks and unit because it avoids filtering the RPM and it has a clear semantic meaning.

But that said, I'm not completely against what this PR is doing if folks prefer to go that route.

jlebon:
This was discussed in [the community meeting today](https://meetbot.fedoraproject.org/fedora-meeting-1/2020-11-25/fedora_coreos_meeting.2020-11-25-16.30.html). There was agreement to use masking for now while we investigate with upstream if they're open to splitting it out.

Added hold label on this.

travier:
https://src.fedoraproject.org/rpms/systemd/pull-request/40 > Upstream does not want to split those into sub packages.

jlebon (approved):
Thanks @travier for bringing this up with upstream.
Patch LGTM!

travier:
Pushed an update that does the masking as agreed although I don't think this is a satisfying solution. As the RPM split is not likely to happen, I'm not sure masking is a good idea as it will make it more complex if we ever want to re-enable that again.

This case is different from `dnsmaq` where we need the binary but want to explicitly convey that it should not be used and relied upon as it may go away in the future. Here we don't need either now but they are probably never going away and we might need the binary and the unit in the future but only for the initramfs. I don't know if it would be possible to mask it just for the final root and not for both (I also don't know if masking it here will mask it for the initramfs).

This feels too complex compared to the minor disadvantage of having a small discrepancy in the RPM DB for only two symlinks that can easily be re-installed by an administrator via Ignition in `/etc` (https://github.com/travier/fedora-coreos-config/commit/9ad7fc16f0bb9e64cbf129f3f85f6fb90fed9297).

dustymabe (approved):
Patch LGTM

keszybz:
systemd-repart and friends do nothing if there is no configuration for them. (And there is no configuration by default.)
I'm confused as to what problem is being solved here.

Edit: also see https://src.fedoraproject.org/rpms/systemd/pull-request/40.

cgwalters:
> I'm confused as to what problem is being solved here.

See https://github.com/coreos/fedora-coreos-config/pull/744#issuecomment-7306339111
Basically we don't want people using it, we have Ignition.

keszybz:
If people should use Ignition, then that should be documented in appropriate places, so that when people try to solve a specific problem, they hit on the right answer.

Trying to remove small bits of functionality just because they are not useful in the *default* setup is futile and counterproductive. It'd be like removing sfdisk from util-linux because Workstation users are supposed to use Anaconda do setup their disk. Futile, because it's Linux and there're countless other things which "shouldn't be done" and a countless number of ways to break the machine as root, and removing one or two of them doesn't change anything. And counterproductive because command-line tools like sfdisk and systemd-repart are also used in scripts, to operate on disk images in files, testing, management of non-root devices, etc. And obviously those packaging gimmicks are creating noise and consuming valuable packager time (and making things harder for users who actually want to use those tools). I think the whole approach is fixing a non-existent problem, because systemd-repart is a generally-unknown low-level tool. If somebody is expert enough to know about it and set up the configuration for it, they will also know when not to to use it. It's not a big icon on the desktop one can click by mistake.

travier:
> Edit: also see https://src.fedoraproject.org/rpms/systemd/pull-request/40.

Thanks for the response in the PR. I agree that splitting systemd into more packages is more work for you for not much gain, which is why I would prefer we remove the default enabled symlinks instead.

> If people should use Ignition, then that should be documented in appropriate places, so that when people try to solve a specific problem, they hit on the right answer.

I would say that we are good here as it is clearly written in the documentation (and you can not avoid it if you want a working system).

> It'd be like removing sfdisk from util-linux because Workstation users are supposed to use Anaconda do setup their disk. Futile, because it's Linux and there're countless other things which "shouldn't be done" and a countless number of ways to break the machine as root, and removing one or two of them doesn't change anything.

With Fedora CoreOS, we are able to be more opinionated about what we ship in the OS and what we support. For example, we do [not ship man packages](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/fedora-coreos.yaml#L75) and we make sure that [no interpreter (beyond bash) gets into the image](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/fedora-coreos.yaml#L85). So if we do not want to support a tool being used to configure the system or in scripts, we try to remove it from the image, especially if this tool can be used from a container. In this case, we only want to support Ignition for first boot disk partitioning.

> And counterproductive because command-line tools like sfdisk and systemd-repart are also used in scripts, to operate on disk images in files, testing, management of non-root devices, etc.

This is part of the support discussion: if we ship those tools, users can start using them. If we then remove those later, we potentially break existing systems as users relied on them being available in the host system. My initial PR removed it entirely, the last update only masks the unit to make sure it does not get called by default and an alternative option linked in a previous comment only removes the default enabled symlinks.

The general goal here is to make it explicit that this is not supported for first boot partitioning on Fedora CoreOS. In the current state, as those units are enabled, one might rightfully expect that to be the case.

keszybz:
I think it's fine to remove high-level components. But trying to get rid of small components (that do nothing unless configured to do something) is another story. OK, you want to drop `systemd-timesync`, `systemd-repart`, `systemd-firstboot`, `systemd-gpt-auto-generator`. What about `systemd-mkfs`, `systemd-volatile-root`, `systemd-cryptsetup-generator` and associated tools, `systemd-growfs`, and the fifteen other toollets we might add over the next year? I find it hard to keep track and remember about all of them *as a maintainer* of the package. You are building an elaborate hack on top of the existing packaging, that will need to be adjusted and maintained over time. The gains from removing such low-level tools are small, but the cost that is created by complicated packaging and difference from mainline Fedora is growing.

dustymabe:
I agree with @keszybz here. There is value in Fedora CoreOS being opinionated, but we should try to do it without adding a lot of additional burden on existing package maintainers who are most likely either volunteering their free time or overloaded already from a full plate.

I'm fine with the current approach in this PR (along with documentation) as it's a good middle ground.

dustymabe:
Anything blocking us from hitting the big green button to merge this?

LorbusChris:
@keszybz @dustymabe I think I do disagree here, but it's probably not a popular opinion with the systemd packagers:

systemd RPM packaging already has a script to assign files to sub-packages by regexing. Would it really be that much of a maintenance burden to add more sub-packages there?
In my opinion having a more fine-grained approach to installing systemd components would be desirable, not just for CoreOS but also for e.g. minimalistic container builds.
repart is just one of quite a few things that could technically be split out since they're not strictly required.

Anyway, that's just my 2 cents :)

The changes here look good.

dustymabe:
> systemd RPM packaging already has a script to assign files to sub-packages by regexing. Would it really be that much of a maintenance burden to add more sub-packages there?

yes and no. It's theoretically easy, but for each subpackage we spin off you then need to handle migrating existing users who expect it to be in the main package and also keep up with which variants have the package installed and what side effects there are as a result.

The point is that it's not just as simple as changing a regex. It's maintenance over time. If there are more than one distro/subvariant that can benefit then we should collectively work with those variants to come up with good reasons to split things out. 

Any time packagers do stuff just for the CoreOS team it adds to a negative bias against our team. Whether it's fair or not we have to consider that and be calculated about what we ask other teams to do so when we do **really** need something we haven't spent all of our political capital already.

dustymabe:
@travier - do you want to handle the docs part (something like https://docs.fedoraproject.org/en-US/fedora-coreos/faq/#_why_is_the_dnsmasq_service_systemd_unit_masked)?

travier:
Will do the docs

travier:
https://github.com/coreos/fedora-coreos-docs/pull/226

PR #745, opened 2020-11-23
coreos-boot-edit: add triggering condition on coreos-copy-firstboot-network.stamp

jlebon:
That way we completely avoid mounting `/boot` rw if we don't need to.
Also clarify comment about the boot dependency. Additional roles for
this service ideally would follow the same pattern.

Minor follow-up to #743.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.service
@@ -8,7 +8,13 @@ ConditionPathExists=/usr/lib/initrd-release
 OnFailure=emergency.target
 OnFailureJobMode=isolate
 
-# Since we are mounting /boot, require the device first
+# These are the conditions for which we may need to modify /boot. Otherwise,
+# let's avoid mounting it entirely.
+# Currently, we only have one trigger.
+ConditionPathExists=|/run/coreos-copy-firstboot-network.stamp
+
+# Since we are mounting /boot, require the device first. This isn't strictly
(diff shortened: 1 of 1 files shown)


jlebon:
Not tested yet.

travier (approved):
LGTM

kelvinfan001 (approved):
LGTM!

jlebon:
Once we [merge coreos-inject-rootmap.service into this](https://github.com/coreos/fedora-coreos-tracker/pull/680#issuecomment-732375070), then I think the motivation for using the triggering model proposed here drops a lot because right now we *always* inject at least `root=... rw` on first boot, so the service will just always kick in. (It's conditional on `root=` not being present, but that karg being present on firstboot is a bit of a snowflake -- mostly for setting up multipath at install-time).

So closing this one. (And accordingly, we can indeed drop the stamp file as discussed in #743.)

kelvinfan001:
Just to be sure, would this mean we get rid of `ConditionKernelCommandLine=!root` too (since it's rare that it's not satisfied)? Or keep the stamp file and have 
```
ConditionKernelCommandLine=|!root
ConditionPathExists=|/run/coreos-copy-firstboot-network.stamp
```

jlebon:
I was arguing for dropping the triggering conditions entirely and just having the unit always run to be consistent, even if in some rare situations it might not have been necessary.

We'd still make the call to `rdcore rootmap` itself in the script conditional on `root=` not being present in `/proc/cmdline` though.

kelvinfan001:
>We'd still make the call to rdcore rootmap itself in the script conditional on root= not being present in /proc/cmdline though.

Ahh, got it. :+1:

PR #746, opened 2020-11-23
[rhcos-4.6] coreos-copy-firstboot-network: order after coreos-enable-network 

jlebon:
In a previous iteration, the "conditional networking" logic lived in
`ignition-fetch-offline.service`. This was later moved to its own
service in `coreos-enable-network.service`, but we never updated the
ordering here.

The end result is that sometimes `coreos-enable-network.service` would
run *after* `coreos-copy-firstboot-network.service`, which meant that we
lost any NM config which was copied from `/boot` (i.e. from
`coreos-installer install --copy-network`).

Resolves: bugzilla.redhat.com/show_bug.cgi?id=1895979
(cherry picked from commit 97bd7ff)


cgwalters (approved):

jlebon:
Inclined to ignore CI for these RHCOS cherry-picks the way we do in cosa, though interested to see whether it fails and if so on what.

PR #747, opened 2020-11-24
05core: Merge `coreos-inject-rootmap.service` into `coreos-boot-edit.service`

kelvinfan001:
So we don't mount `/boot` more times than we need to.


overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.sh
```diff
@@ -14,9 +14,10 @@ mount -o rw ${bootdev} ${bootmnt}
...
-    rm -vrf ${initramfs_firstboot_network_dir}
-else
-    echo "info: no firstboot networking config files to clean from /boot. skipping"
+rm -vrf ${initramfs_firstboot_network_dir}
+
+# If root is specified, assume rootmap is already configured. Otherwise,
+# append rootmap kargs to the BLS configs.
+if ! grep -q "root" /proc/cmdline; then
+    /usr/bin/rdcore rootmap /sysroot --boot-mount ${bootmnt}
 fi
```
travier: Should we teach rdcore to do that directly?

travier (approved):
LGTM but one minor question for @jlebon maybe.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.sh
jlebon: Are you referring to the conditional on the `root` karg?

Maybe... though in general I prefer to keep rdcore functionality lower-level and tightly-scoped and have the more business-y logic here because it's a pain to iterate on rdcore functionality right now since it lives in a separate project which needs to be released, built, etc...

This is why for example I wrote `rootmap` to take either a `--boot-device` switch or a `--boot-mount` switch; so that we can change the implementation of how that code is called like we do here without having to touch rdcore itself at all.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.sh
```diff
@@ -14,9 +14,10 @@ mount -o rw ${bootdev} ${bootmnt}
...
-copy_firstboot_network_stamp="/run/coreos-copy-firstboot-network.stamp"
-if [ -f ${copy_firstboot_network_stamp} ]; then
-    rm -vrf ${initramfs_firstboot_network_dir}
-else
-    echo "info: no firstboot networking config files to clean from /boot. skipping"
+rm -vrf ${initramfs_firstboot_network_dir}
+
+# If root is specified, assume rootmap is already configured. Otherwise,
+# append rootmap kargs to the BLS configs.
+if ! grep -q "root" /proc/cmdline; then
```
jlebon: I think we need something a little stronger here because `root` is not uncommon, so we risk matching something we didn't mean to. Maybe let's steal the `karg` function from https://github.com/coreos/coreos-installer/blob/master/systemd/coreos-installer-generator, and then do something like:

```sh
root=$(karg root)
if [ -z "${root}" ]; then
    ....
fi
```

?

overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-copy-firstboot-network.sh
```diff
@@ -10,7 +10,6 @@ firstboot_network_dir_basename="coreos-firstboot-network"
 initramfs_firstboot_network_dir="${bootmnt}/${firstboot_network_dir_basename}"
 initramfs_network_dir="/run/NetworkManager/system-connections/"
 realroot_firstboot_network_dir="/boot/${firstboot_network_dir_basename}"
-copy_firstboot_network_stamp="/run/coreos-copy-firstboot-network.stamp"
```
jlebon, line 13: Minor/optional: would be clearer to have the stamp file related stuff in a separate commit.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.sh
```diff
@@ -14,9 +14,22 @@ mount -o rw ${bootdev} ${bootmnt}
 # installed system (most likely by using `coreos-installer install --copy-network`).
 firstboot_network_dir_basename="coreos-firstboot-network"
 initramfs_firstboot_network_dir="${bootmnt}/${firstboot_network_dir_basename}"
-copy_firstboot_network_stamp="/run/coreos-copy-firstboot-network.stamp"
-if [ -f ${copy_firstboot_network_stamp} ]; then
-    rm -vrf ${initramfs_firstboot_network_dir}
-else
-    echo "info: no firstboot networking config files to clean from /boot. skipping"
+rm -vrf ${initramfs_firstboot_network_dir}
```
jlebon, line 28: This would belong better in the next commit.

jlebon (approved):
One minor comment, but not worth the respin!

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.sh
```diff
@@ -14,9 +14,22 @@ mount -o rw ${bootdev} ${bootmnt}
...
+
+# If root is specified, assume rootmap is already configured. Otherwise,
+# append rootmap kargs to the BLS configs.
+cmdline=( $(</proc/cmdline) )
+karg() {
+    local name="$1"
```
jlebon: Try this:

```suggestion
    local name="$1" value="${2:}"
```

That should fix the CI failure.

jlebon: Also, WDYT about putting that function at the top of the script to separate it out from the rest of the logic?

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.sh
kelvinfan001: Ah yes that would make a lot more sense.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-boot-edit.sh
jlebon: The reason this is failing here is because we're running with `set -u` whereas the code in coreos-installer we copied this from just has `set -e` (though we should change it to use `set -euo pipefail` like here).

kelvinfan001: +1 to putting it on top. Actually I think CI would still fail because I did `set -u`, so the script would error when there's no `root` karg.

jlebon: It should be OK with `set -u` as long as you provide a default value on variables which may not be defined. This is what `value="${2:}"` does. That *should* be all that's needed.

kelvinfan001: oh sorry, I misread. Right, `value="{2:}"` would work (didn't notice the `:` earlier).

kelvinfan001: ah, I think the correct syntax should be `"${2:-}"`

jlebon: Yes sorry, you're right, that should've been `"${2:-}"`.

jlebon (approved):
LGTM!

PR #754, opened 2020-12-01
[next] manifest: remove duplicate fedora-repos-modular entry

dustymabe:
When we switched testing-devel over to Fedora 33 in 09a4df3
we added `fedora-repos-modular` to fedora-coreos-base.yaml, which
makes this entry in manifest.yaml redundant now.


jlebon (approved):
Nice catch!

PR #757, opened 2020-12-03
tests/manual: add coreos-docs-net-testing.sh

dustymabe:
This script attempts to test the configurations in our documentation
at https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-network-configuration/.
All it does is test the various documented scenarios via both the
dracut kernel networking arguments approach, as well as the NM
keyfiles provided via Ignition approach.


dustymabe:
I killed CI since these files stand alone and have no bearing on actual function.

lucab (approved):

PR #758, opened 2020-12-03
20live: log rootfs fetch errors

bgilbert:
When rootfs fetch fails, ensure we have some useful logging for tracking down the problem.

`--show-error` has existed in curl since 1999 and I'm not sure why I didn't use it originally.


cgwalters (approved):

cgwalters:
> --show-error has existed in curl since 1999

I too suggest to my friends we party like we have modern curl!

PR #759, opened 2020-12-03
tree: promote changes from testing at 399117235c58fa1a7d22050142f214f5749d6329

cverna:
…3355ea90c


overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
```diff
@@ -3,5 +3,3 @@ enable fedora-coreos-pinger.service
 # Provide information if no ignition is provided
 enable coreos-check-ignition-config.service
 enable coreos-check-ssh-keys.service
-# Enable the migration service to systemd-resolved
```
lucab: There is something fishy going on here. This change comes from PR https://github.com/coreos/fedora-coreos-config/pull/752 which only landed very recently in `testing-devel`.

I don't expect this change to be promoted in this `stable` release cycle, but on the next one.

lucab (changes requested):

lucab (approved):

PR #761, opened 2020-12-03
dracut/15coreos-network: enable initramfs networking on Azure

### LINKED ISSUE #689: azure: boot issues due to missing networking in initramfs
I was trying booting a custom FCOS image in Azure and saw this on the serial console:

```
[   73.342818] afterburn[543]: Dec 02 15:22:14.000 INFO Fetching http://168.63.129.16/?comp=versions: Attempt #9
[   73.355681] afterburn[543]: Dec 02 15:22:14.013 INFO Failed to fetch: error sending request for url (http://168.63.129.16/?comp=versions): error trying to connect: tcp connect error: Network is unreachable (os error 101)
```

Apparently when we moved to optional networking in initramfs we broke plain Azure booting, as the Ignition configuration is coming from a config-drive but other pieces (e.g. hostname and boot check-in from Afterburn) needs to reach the metadata endpoint and the wireserver over the network.

lucab:
This enables networking in initramfs on Azure, via a platform
condition.
While the Ignition configuration may be self-contained locally
on the config-drive, other bits referenced by the provisioning
flow (e.g. Afterburn hostname and boot check-in) always require
reaching remote endpoints.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/689

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-enable-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-network/coreos-enable-network.service
@@ -4,9 +4,13 @@ ConditionPathExists=/etc/initrd-release
 DefaultDependencies=false
 After=basic.target
 
-# Triggering conditions for cases where we need network
+# Triggering conditions for cases where we need network:
+#  * when Ignition signals that it is required for provisioning.
+#  * on live systems fetching the remote rootfs in initramfs.
+#  * on Azure, for hostname fetching (metadata endpoint) and boot check-in (wireserver).
 ConditionPathExists=|/run/ignition/neednet
 ConditionKernelCommandLine=|coreos.live.rootfs_url
+ConditionKernelCommandLine=|ignition.platform.id=azure
 
 # Creates /run/ignition/neednet
 After=ignition-fetch-offline.service


dustymabe (approved):
LGTM - but will defer to jlebon

jlebon (approved):
Nice, thanks for the added detailed comment!

cgwalters (approved):
I'll reiterate here I still think at some point we should take advantage of the fact that the metadata endpoint is "link local" and make it so that NM has first-class support for something like `networkmanager-primary-link-local-up.target`.

(But bigger picture, maybe there isn't a huge amount of value in avoiding DHCP on clouds because we generally expect most people to use Ignition includes)

jlebon:
@cgwalters That used to be somewhat tracked in https://github.com/coreos/fedora-coreos-tracker/issues/460 though I've closed it now. Can you file a separate tracker issue specifically for this?

lucab:
For Azure specifically and thus the scope of this PR, the [wireserver](https://docs.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16) is not link-local (in the traditional sense of IPv4LL / RFC3927) but it's a global public IPv4 (`168.63.129.16`) part of a larger BGP-announced /15 subnet, owned by MS.

PR #765, opened 2020-12-03
Preparation for new version of console-login-helper-messages

kelvinfan001:
In order for newer versions of `console-login-helper-messages` to display issue files before login properly, the `--issue-file` option for `agetty` must be set to read from `/run/issue.d`. We also need to add `/run/motd.d` as one of the directories that `login(1)` reads from to display the MOTD after login through a console.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -96,6 +96,20 @@ postprocess:
     DNSStubListener=no
     EOF
 
+  # Edit `login.defs` to configure `login(1)` to read from `/run/motd.d` for
+  # displaying the MOTD. This is required for newer versions of
+  # `console-login-helper-messages` to function properly.
+  # This will be dropped once Fedora util-linux adds `/run/motd.d` as a default
+  # in Fedora 34.
+  # https://src.fedoraproject.org/rpms/util-linux/pull-request/8
+  # https://github.com/coreos/fedora-coreos-tracker/issues/704#issuecomment-772862174
+  - |
+    #!/usr/bin/env bash
+    source /etc/os-release
+    if [ ${VERSION_ID} -lt 34 ]; then
+      echo 'MOTD_FILE=/usr/share/misc/motd:/run/motd:/run/motd.d:/etc/motd:/etc/motd.d' >> /etc/login.defs
+    fi
+
 packages:
   # Security
   - selinux-policy-targeted


kelvinfan001:
Hold until new `console-login-helper-messages` release.

overlay.d/15fcos/etc/systemd/system/serial-getty@ttyS0.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
```
lucab: This file should be placed under `/usr`, not `/etc`.

kelvinfan001: Ah thanks! I was under the impression that `override.conf` files only go in `/etc`. I'm guessing because we're the distro, we can put it in `/usr`? In that case, what would happen if we directly edited the original systemd usr file?

manifests/fedora-coreos-base.yaml
```diff
@@ -69,6 +69,15 @@ postprocess:
     #!/usr/bin/env bash
     systemctl mask dnsmasq.service
 
+  # Edit `login.defs` to configure `login(1)` to read from `/run/motd.d` for
+  # displaying the MOTD. This is required for newer versions of
+  # `console-login-helper-messages` to function properly. 
+  # This will be dropped once Fedora util-linux adds `/run/motd.d` as a default.
+  # https://github.com/coreos/console-login-helper-messages/issues/60
```
dustymabe: should we link to the existing PR to change it in Fedora too?

overlay.d/15fcos/usr/lib/systemd/system/serial-getty@ttyS0.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=-/sbin/agetty -o '-p -- \\u' --keep-baud 115200,38400,9600 %I $TERM --issue-file /etc/issue:/run/issue.d
```
dustymabe: Do we need to carry this override indefinitely? Or will we be able to drop it at some point too when new upstream changes land (either upstream or in Fedora itself)?

manifests/fedora-coreos-base.yaml
```diff
@@ -69,6 +69,15 @@ postprocess:
     #!/usr/bin/env bash
     systemctl mask dnsmasq.service
 
+  # Edit `login.defs` to configure `login(1)` to read from `/run/motd.d` for
+  # displaying the MOTD. This is required for newer versions of
+  # `console-login-helper-messages` to function properly. 
+  # This will be dropped once Fedora util-linux adds `/run/motd.d` as a default.
```
dustymabe: > `# This will be dropped once`

I feel like we need a better way to remind ourselves to do these "go back and fix this up" type of thing. I wonder if we just made a new test within exttests that said:
```
if fedora version is != 33; then
    echo "you should go back and check if we can drop <link to code> now"
    fail test
fi
```

cc @jlebon for ideas

overlay.d/15fcos/usr/lib/systemd/system/serial-getty@ttyS0.service.d/override.conf
dustymabe: looks like this is an open question being asked over in https://github.com/coreos/console-login-helper-messages/issues/60#issuecomment-738882358

kelvinfan001: I'm trying to contact the maintainer of util-linux and ask him if he's open to giving `agetty` the same treatment as `login` (carry a patch just in Fedora to have the defaults we like) https://github.com/coreos/console-login-helper-messages/issues/60#issuecomment-738882358. 

However, the way this PR currently does the overriding seems a bit sketchy. For example, it'll override the autologin ignition config that cosa injects during `cosa run`.

dustymabe: > However, the way this PR currently does the overriding seems a bit sketchy. For example, it'll override the autologin ignition config that cosa injects during `cosa run`.

Now that it's in `/usr/` the stuff we inject on `cosa run` should win, right?

kelvinfan001: > cosa run should win, right?
 
Ah I see. Yes let me try again to make sure. But still, if `cosa run` wins, then we'd still have the issue snippets not displaying on the serial console when using `cosa run` (but maybe that's not a big deal).

kelvinfan001:
```
$ cd /
$ sudo find /etc -name 10-autologin.conf
/etc/systemd/system/getty@.service.d/10-autologin.conf
/etc/systemd/system/serial-getty@.service.d/10-autologin.conf
$ sudo find /usr -name override.conf
/usr/lib/systemd/system/serial-getty@.service.d/override.conf
```
But for some reason I'm still not automatically logged in when I do `cosa run -c`. The `autologin.conf` in `/etc` should win, right?

Edit: Ok so the systemd conf file in `serial-getty@.service.d` injected by `cosa run` would only override the conf file in the `/usr` equivalent if they had the same name. Seems kind of counter-intuitive to me; I imagined that any directive in `/etc` would just win out, regardless of filename.

dustymabe: I just ran a local test (tip: you can play around with the local system and `systemctl cat serial-getty@.service` to see what systemd is actually overriding and what it isn't). It seems to be working like I would expect.

I think the problem you were hitting is that you were initially putting a file in `/usr/lib/systemd/system/serial-getty@ttyS0.service.d/override.conf ` and not `/usr/lib/systemd/system/serial-getty@.service.d/override.conf`. It seems like your PR is right now. Does it still not work for you?

dustymabe: ok, I get it now. I just rebuilt your PR locally. The file names don't have to be the same name, but you do want the filename to sort in front of `10-autologin.conf` (so that `10-autologin.conf` will win). So let's rename it from `/usr/lib/systemd/system/serial-getty@.service.d/override.conf` to `/usr/lib/systemd/system/serial-getty@.service.d/09-override.conf`.

Another option is that we change COSA, so we'd use something like `50-override.conf` and move COSA to use something like `90-autologin.conf`.

While we're at it, can we thing of a better name than `override.conf`? Maybe `modifiedissuefile.conf` or something.

kelvinfan001: Thanks for digging in to it! I think you're right. From [`systemd.unit.html`](https://www.freedesktop.org/software/systemd/man/systemd.unit.html):
> Drop-in files in /etc/ take precedence over those in /run/ which in turn take precedence over those in /usr/lib/. Drop-in files under any of these directories take precedence over unit files wherever located. Multiple drop-in files with different names are applied in lexicographic order, regardless of which of the directories they reside in.

I'm interpreting this to mean conf files in configuration subdirectories of the same unit do not care whether they're in `/usr` or `/etc` and only care about their lexicographic order if they have different names.

dustymabe: And also, maybe we should come up with a better way. We've got a few things that modify this unit to override the defaults. Maybe we should only override the `ExecStart` in one place, but we add a variables to the line and then we can override those for our autologin bits so something like:

```
[Service]
Environment="LOGINOPTIONS=-o '-p -- \\u'"
Environment="OTHEROPTIONS=--issue-file /etc/issue:/run/issue.d"
ExecStart=
ExecStart=-/sbin/agetty $LOGINOPTIONS --keep-baud 115200,38400,9600 %I $TERM $OTHEROPTIONS
```

Then we can use the other existing overrides to just set "LOGINOPTIONS=--autologin core -o '-p -f core'" I think.

Not sure. Might be a bad idea.

kelvinfan001: > We've got a few things that modify this unit to override the defaults. 

Yeah this seems to be a problem. Even if there were only one thing (`cosa run`'s `autologin.conf`) modifying this unit, because we need to override instead of append, we kind of only have one or the other, instead of ideally being able to both autologin AND show the issue files. 

I think your idea is definitely better, at least we can display both issue files and autologin at the same time, but what if multiple overrides wanted to edit `LOGINOPTIONS`? Unfortunately I think we can't do something like
```
[Service]
Environment="LOGINOPTIONS=$LOGINOPTIONS --more-options"
```

I'm wondering if systemd has a more built-in way of appending to options... 

Anyways, I think we should wait for the maintainer of util-linux to respond whether he's open to adding what we want as defaults so we don't need to worry about this problem at all.

kelvinfan001: We've decided to use `/etc/issue.d` instead. Thus, no configuration for `agetty` is necessary now.

kelvinfan001:
We no longer need to do any additional configuration for `agetty`, but still waiting on the upstream maintainers regarding whether the `login(1)` defaults for `MOTD_FILE` can be changed to what we want (https://src.fedoraproject.org/rpms/util-linux/pull-request/8); if so, we can close this PR.

kelvinfan001:
The changes in https://src.fedoraproject.org/rpms/util-linux/pull-request/8 has been incorporated into the rawhide branch and will be available in Fedora 34. https://src.fedoraproject.org/rpms/util-linux/blob/rawhide/f/util-linux.spec#_951.
@jlebon Can we merge this for now while we wait?

manifests/fedora-coreos-base.yaml
```diff
@@ -96,6 +96,17 @@ postprocess:
...
+  # Edit `login.defs` to configure `login(1)` to read from `/run/motd.d` for
+  # displaying the MOTD. This is required for newer versions of
+  # `console-login-helper-messages` to function properly.
+  # This will be dropped once Fedora util-linux adds `/run/motd.d` as a default
+  # in Fedora 34.
+  # https://src.fedoraproject.org/rpms/util-linux/pull-request/8
+  # https://github.com/coreos/console-login-helper-messages/issues/60
+  - |
+    #!/usr/bin/env bash
+    echo 'MOTD_FILE=/usr/share/misc/motd:/run/motd:/run/motd.d:/etc/motd:/etc/motd.d' >> /etc/login.defs
```
jlebon: Following up on Dusty's comment, WDYT about making this conditional on the Fedora version? E.g. something like

```bash
source /etc/os-release
if [ ${VERSION_ID} -lt 34 ]; then
    echo ...
fi
```

Then we can test in the rawhide stream that clhm works fine without this bit in preparations for dropping it entirely once we move to f34.

jlebon: I also added a reminder in https://github.com/coreos/fedora-coreos-tracker/issues/704#issuecomment-772862174.

kelvinfan001: thanks! sounds good.

jlebon (approved):
Thanks!

PR #767, opened 2020-12-04
99emergency-timeout: fix log display from failed units

bgilbert:
`timeout.sh` was getting confused by `systemctl` output that looked like this:

    ● ignition-disks.service loaded failed failed Ignition (disks)

The result was output like this:

    Displaying logs from failed units: ●
    Invalid unit name "●" escaped as "\xe2\x97\x8f" (maybe you should use systemd-escape?).
    -- Logs begin at Fri 2020-12-04 04:09:20 UTC, end at Fri 2020-12-04 04:09:33 UTC. --
    -- No entries --
    Press Enter for emergency shell or wait 5 minutes for reboot.


jlebon (approved):
Nice, thanks. I noticed this recently too, but forgot to follow up on it.

PR #768, opened 2020-12-04
switch from tmpfs to zram

sedlund:
This reduces ram usage for switching root by about 40% with the default lzo-rle with 5.x kenel.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -20,6 +20,7 @@ saved_boot=${saved_data}/boot
 saved_esp=${saved_data}/esp
 saved_bios=${saved_data}/bios
 saved_prep=${saved_data}/prep
+zram_dev=${saved_data}/zram_dev
 partstate_root=/run/ignition-ostree-rootfs-partstate.json
 
 # Print jq query string for wiped filesystems with label $1
@@ -66,11 +67,20 @@ case "${1:-}" in
             echo "Found duplicate or missing BIOS-BOOT or PReP labels in config" >&2
             exit 1
         fi
+        modprobe zram
+        read dev < /sys/class/zram-control/hot_add
+        # disksize is set arbitrarily large, as zram is capped by mem_limit
+        echo 10G > /sys/block/zram"${dev}"/disksize
+        # Limit zram to 90% of available RAM: we want to be greedy since the
+        # boot breaks anyway, but we still want to leave room for everything
+        # else so it hits ENOSPC and doesn't invoke the OOM killer
+        echo $(( $(grep MemAvailable /proc/meminfo | awk '{print $2}') * 90 / 100 ))K > /sys/block/zram"${dev}"/mem_limit
+        mkfs.xfs /dev/zram"${dev}"
         mkdir "${saved_data}"
-        # use 80% of RAM: we want to be greedy since the boot breaks anyway, but
-        # we still want to leave room for everything else so it hits ENOSPC and
-        # doesn't invoke the OOM killer
-        mount -t tmpfs tmpfs "${saved_data}" -o size=80%
+        mount /dev/zram"${dev}" "${saved_data}"
+        # save the zram device number created for when called to cleanup
+        echo "${dev}" > "${zram_dev}"
+
         if [ "${wipes_root}" != "0" ]; then
             mkdir "${saved_root}"
         fi
@@ -122,6 +132,9 @@ case "${1:-}" in
             echo "Moving PReP partition to RAM..."
             cat "${prep_part}" > "${saved_prep}/partition"
         fi
+        echo "zram usage:"
+        read dev < "${zram_dev}"
+        cat /sys/block/zram"${dev}"/mm_stat
         ;;
     restore)
         # Mounts happen in a private mount namespace since we're not "offically" mounting
@@ -173,8 +186,10 @@ case "${1:-}" in
     cleanup)
         # Mounts are not in a private namespace so we can unmount ${saved_data}
         if [ -d "${saved_data}" ]; then
+            read dev < "${zram_dev}"
             umount "${saved_data}"
             rm -rf "${saved_data}" "${partstate_root}"
+            echo "${dev}" > /sys/class/zram-control/hot_remove
         fi
         ;;
     *)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -38,7 +50,10 @@ case "${1:-}" in
         ;;
     cleanup)
         if [ -d "${saved_sysroot}" ]; then
+            dev=$(grep "${saved_sysroot}" /proc/mounts | awk -F'[^0-9]*' '{print $2}')
             umount "${saved_sysroot}"
+            echo /dev/zram"${dev}" > /sys/class/zram-control/hot_remove
+            modprobe -r zram 2>/dev/null
```
cgwalters: Hmm...removing the module seems a bit dangerous as a general pattern because it's possible something else wants to use it.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -20,7 +20,19 @@ case "${1:-}" in
         # use 80% of RAM: we want to be greedy since the boot breaks anyway, but
         # we still want to leave room for everything else so it hits ENOSPC and
         # doesn't invoke the OOM killer
-        mount -t tmpfs tmpfs "${saved_sysroot}" -o size=80%
+        #mount -t tmpfs tmpfs "${saved_sysroot}" -o size=80%
+        if [ -d /sys/class/zram-control]; then
+            read dev < /sys/class/zram-control/hot_add
+        else
+            modprobe zram
```
cgwalters: I would do this unconditionally; it's a no-op if the module is already loaded right?  Hm well I guess this code does it this way too https://github.com/systemd/zram-generator/blob/93e88c54460c12353963e09cfaf9172b509aeab1/src/generator.rs#L68

Probably fine as is but I'd have just written:
```
modprobe zram
read dev < /sys/class/zram-control/hot_add
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -20,7 +20,19 @@ case "${1:-}" in
...
+            dev=0
+        fi
+        echo Using /dev/zram"${dev}" for copying sysroot
+        echo 4G > /sys/block/zram"${dev}"/disksize
```
cgwalters: Let's make the size here be the same as e.g. `lsblk -n -o SIZE /dev/disk/by-label/root`.

cgwalters (commented):
A few comments; thanks for working on this!

I did verify that `modprobe zram` works in current RHCOS/RHEL8.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
sedlund: modprobe -r won't force unload, it will say it's in use and exit.  I'll remove that though.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
sedlund: True.  Simpler is better.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -38,7 +50,10 @@ case "${1:-}" in
         ;;
     cleanup)
         if [ -d "${saved_sysroot}" ]; then
+            dev=$(grep "${saved_sysroot}" /proc/mounts | awk -F'[^0-9]*' '{print $2}')
             umount "${saved_sysroot}"
+            echo /dev/zram"${dev}" > /sys/class/zram-control/hot_remove
```
jlebon: Shouldn't this be

```suggestion
            echo ${dev} > /sys/class/zram-control/hot_remove
```

? Or does it work both way?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -38,7 +50,10 @@ case "${1:-}" in
         ;;
     cleanup)
         if [ -d "${saved_sysroot}" ]; then
+            dev=$(grep "${saved_sysroot}" /proc/mounts | awk -F'[^0-9]*' '{print $2}')
```
jlebon: Let's just save the device number somewhere under `${saved_data}` and read it back from there.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
jlebon: With #718 now, that should probably just be the size of the parent device. But hmm, is there a reason why we don't just simplify this by doing something obviously large here like 10G since we're covered by the memory limit anyway?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
```diff
@@ -20,7 +20,19 @@ case "${1:-}" in
...
+        if [ -d /sys/class/zram-control]; then
+            read dev < /sys/class/zram-control/hot_add
+        else
+            modprobe zram
+            dev=0
+        fi
+        echo Using /dev/zram"${dev}" for copying sysroot
+        echo 4G > /sys/block/zram"${dev}"/disksize
+	# Limit zram to 95% of available memory
+	echo $(( $(grep MemAvailable /proc/meminfo | awk '{print $2}') * 95 / 100 ))K > /sys/block/zram"${dev}"/mem_limit
```
jlebon: Ahh nice. Makes sense to make it a percentage of available memory vs total memory we were doing before with the tmpfs. 95% still seems a bit high though. Let's do 90%?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
sedlund: yes you're right

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
sedlund: `${saved_data}` or `${saved_sysroot}` ? how about in /run ?

jlebon: `${saved_data}` came in with #718, which just merged. You'll need to rebase this PR. :)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-dracut-rootfs.sh
sedlund: @jlebon
" is there a reason why we don't just simplify this by doing something obviously large here like 10G since we're covered by the memory limit anyway?"

Thats why I started with 4G, but using by-label/root seems to be ok.

I reduced from 95% available ram to 90%.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -57,11 +58,14 @@ case "${1:-}" in
             echo "Found duplicate or missing BIOS-BOOT or PReP labels in config" >&2
             exit 1
         fi
+        modprobe zram
+        dev=$(cat /sys/class/zram-control/hot_add | tee "${saved_zdev}")
+        echo $(lsblk -n -b -o SIZE "${root_part}") > /sys/block/zram"${dev}"/disksize
```
jlebon: Following up on https://github.com/coreos/fedora-coreos-config/pull/768#discussion_r536478361, I think it's a bit misleading to use the root partition size here. In a RAID1 reconfiguration, we're saving more than just the root partition. And while I think it will fit anyway since the rootfs has some cushion, I think it'd be cleaner to just rely on the RAM limit instead and use something way larger here like 10G.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -57,11 +58,14 @@ case "${1:-}" in
...
+        # Limit zram to 90% of available memory
```
jlebon: I think the comment block deleted below is still valuable and should just replace this one instead (with "80% of RAM" replaced with "90% of available RAM").

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -20,6 +20,7 @@ saved_boot=${saved_data}/boot
 saved_esp=${saved_data}/esp
 saved_bios=${saved_data}/bios
 saved_prep=${saved_data}/prep
+saved_zdev=${saved_data}/zdev
```
jlebon: WDY about calling this e.g. `zram_dev`? The `saved_...` naming is otherwise used for partitions/filesystems which we've saved from the boot disk into RAM, whereas here we're just saving a variable.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -57,11 +58,14 @@ case "${1:-}" in
             echo "Found duplicate or missing BIOS-BOOT or PReP labels in config" >&2
             exit 1
         fi
+        modprobe zram
+        dev=$(cat /sys/class/zram-control/hot_add | tee "${saved_zdev}")
```
jlebon: This requires adding `tee` to the initramfs, no? Let's just do this in two steps instead to avoid that?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: Same idea as https://github.com/coreos/fedora-coreos-config/pull/685#discussion_r504923611. We're already capping at the RAM usage level which is what we care most about. So there's little value in also capping at the "virtual RAM disk size" level.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -165,6 +184,7 @@ case "${1:-}" in
         # Mounts are not in a private namespace so we can unmount ${saved_data}
         if [ -d "${saved_data}" ]; then
             umount "${saved_data}"
+            cat "${zram_dev}" > /sys/class/zram-control/hot_remove
```
bgilbert: `${zram_dev}` is inside `${saved_data}`, which we just unmounted.

bgilbert (commented):
Looks good generally; thanks for doing this!

Please rebase.

sedlund:
force merge ftw?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -66,11 +67,18 @@ case "${1:-}" in
             echo "Found duplicate or missing BIOS-BOOT or PReP labels in config" >&2
             exit 1
         fi
+        modprobe zram
+        read dev < /sys/class/zram-control/hot_add
+        echo "${dev}" > "${zram_dev}"
```
jlebon: Shouldn't this line be after the `mount` now that we read it back before unmounting in the cleanup phase?

jlebon (approved):
One final comment, LGTM otherwise!

dustymabe:
do you mind adding a comment to the commit about the benefits of making this change? If I'm not mistaken right now the commit only has a title and no body.

dustymabe:
@sedlund - mind pushing this PR over the finish line. Just a few more small changes!

sedlund:
sorry for the delay -- appreciate the feedback.

sedlund:
Hmm.  I'm not seeing what's failing

sedlund:
Got a good build.  filesystem-only reprovision reduced from 1491MB RAM used to 907MB

jlebon (approved):
Cool, thanks a lot for this!

PR #769, opened 2020-12-07
40ignition-ostree: be verbose about what we're mounting

jlebon:
We're sometimes hitting a flake in the RHCOS CI for LUKS-related tests:

    ignition-ostree-dracut-rootfs[909]: mount: /sysroot: unknown filesystem type 'crypto_LUKS'.
    systemd[1]: Failed to start Ignition OSTree: restore rootfs.

I think what may be happening here is that the `/dev/disk/by-label/root`
symlink is still pointing at the original partition which now hosts the
LUKS device and didn't actually get updated to point to the newly
created filesystem on top of it.

The `disks` stage does do a `udevadm trigger` on exit, though I'm not
sure how foolproof that is.

What we could do eventually to work around this is get the `device` from
the Ignition config and just mount that instead. Though to confirm first
this is what's happening, let's make the mount operations explicitly
call out what devices they're mounting.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -32,6 +32,15 @@ query_parttype() {
     echo ".storage?.disks? // [] | map(.partitions?) | flatten | map(select(try .typeGuid catch \"\" | ascii_downcase == \"$1\"))"
 }
 
+
+# Mounts device to directory, with extra logging of the src device
+mount_verbose() {
+    local srcdev=$1; shift
+    local destdir=$1; shift
+    echo "Mounting ${srcdev} ($(realpath "$srcdev")) to $destdir"
+    mount "${srcdev}" "${destdir}"
+}
+
 # Print partition offset for device node $1
 get_partition_offset() {
     local devpath=$(udevadm info --query=path "$1")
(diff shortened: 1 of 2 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -117,21 +126,21 @@ case "${1:-}" in
...
         fi
         if [ -d "${saved_boot}" ]; then
             mkdir -p /sysroot/boot
-            mount "${boot_part}" /sysroot/boot
+            mount_verbose "${boot_part}" /sysroot/boot
             echo "Restoring bootfs from RAM..."
             find "${saved_boot}" -mindepth 1 -maxdepth 1 -exec mv -t /sysroot/boot {} \;
         fi
         if [ -d "${saved_esp}" ]; then
             echo "Restoring EFI System Partition from RAM..."
```
bgilbert, line 141: Somehow this case has the log message before the mount rather than after.  Can we move the log before the mount in all cases to make the logs less confusing?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -32,6 +32,15 @@ query_parttype() {
     echo ".storage?.disks? // [] | map(.partitions?) | flatten | map(select(try .typeGuid catch \"\" | ascii_downcase == \"$1\"))"
 }
 
+
+# Mounts device to directory, with extra logging of the src device
+mount_verbose() {
+    srcdev=$1; shift
```
bgilbert: `local`

bgilbert (approved):

PR #770, opened 2020-12-08
Add a udevadm settle invocations in transposefs and uuid

cgwalters:
I think this will fix a race I saw in an rpm-ostree PR that looks like
this:

```
...
Ignition finished successfully
Missed 21 kernel messages
audit: type=1130 audit(1607446019.771:30): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=ignition-disks comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=ignition-disks comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Finished Ignition (disks).
Reached target Initrd Root Device.
Restoring rootfs from RAM...
Mounting /dev/disk/by-label/root (/dev/disk/by-label/root) to /sysroot
Missed 3 kernel messages
audit: type=1130 audit(1607446019.805:31): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=ignition-ostree-uuid-root comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'
Missed 1 kernel messages
audit: type=1131 audit(1607446019.816:32): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=dracut-pre-mount comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=ignition-ostree-uuid-root comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'
SERVICE_STOP pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=dracut-pre-mount comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Starting Ignition OSTree: Restore Partitions...
mount: /sysroot: special device /dev/disk/by-label/root does not exist.
Missed 2 kernel messages
[...]

PR #771, opened 2020-12-09
40ignition-ostree: fix possible transposefs-restore race

bgilbert:
Serialize `transposefs-restore` and `firstboot-uuid`, since they both want to mutate the root filesystem.  Also have `firstboot-uuid` wait for udev to settle, since closing a device node for write can cause udevd to recreate `/dev/disk` links.

Alternative to https://github.com/coreos/fedora-coreos-config/pull/770.  Untested.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
@@ -34,6 +34,7 @@ if [ "${TYPE}" == "${orig_type}" ] && [ "${UUID}" == "${orig_uuid}" ]; then
     xfs) xfs_admin -U generate "${target}" ;;
     *) echo "unexpected filesystem type ${TYPE}" 1>&2; exit 1 ;;
   esac
+  udevadm settle
   echo "Regenerated UUID for ${target}"
 else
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
LGTM, thanks!

jlebon (approved):
Nice! This may also fix the race that motivated https://github.com/coreos/fedora-coreos-config/pull/769.

jlebon:
We're still hitting a similar issue even with this:

```
[   54.767338] ignition-ostree-transposefs[938]: Restoring rootfs from RAM...
[   54.776049] ignition-ostree-transposefs[938]: Mounting /dev/disk/by-label/root (/dev/vda4) to /sysroot
[   54.791159] systemd[1]: Starting Ignition OSTree: Restore Partitions...
[FAILED] Failed to start Ignition OSTree: Restore Partitions.
See 'systemctl status ignition-ostree-transposefs-restore.service' for details.
[DEPEND] Dependency failed for Ignition Complete.
[DEPEND] Dependency failed for Initrd Default Target.
[   54.814061] ignition-ostree-transposefs[938]: mount: /sysroot: unknown filesystem type 'crypto_LUKS'.
```

From the verbose mounting added in #769, we can tell that `/dev/disk/by-label/root` is pointing at the wrong thing (`/dev/vda4` instead of `/dev/dm-0`).

The firstboot UUID restamping service doesn't kick in in those cases because we did recreate the rootfs so there's no need to restamp.

Posted a follow-up in https://github.com/coreos/fedora-coreos-config/pull/779.

PR #773, opened 2020-12-09
coreos-teardown-initramfs: don't propagate purely default network configs

### LINKED ISSUE #696: RFE: don't propagate default initramfs networking configs
**Describe the enhancement**

In the initramfs we use `nm-initrd-generator` to generate NM keyfiles and then start up networking from that config (if networking is needed). We also will propagate (aka persist) this networking configuration from the initramfs into the real root of the system if there were no NM config files written down by Ignition.

For a purely "defaults" boot this has a subtle difference between the configuration of say NetworkManager booted on Fedora Workstation versus Fedora CoreOS. In Fedora Workstation there would be no files in `/etc/NetworkManager/system-connections` and NM will dynamically generate dhcp profiles for each device (i.e. `Wired connection 1`, `Wired connection 2`). In Fedora CoreOS right now we get a single profile `Wired connection` that matches all NICs in the machine. The end goal configuration is the same, all devices try DHCP. However we should consider trying to reach parity with the default NM behavior because:

1. `coreos-installer --copy-network`

When interactively doing an install from the live ISO you need to delete the default connection profile and then recreate individual ones for each NIC if you want to customize. This is a bit counter intuitive. See https://bugzilla.redhat.com/show_bug.cgi?id=1901517

2. maintaining parity with NM default behavior is valuable

Mostly self explanatory. It will look more like a default NM boot on any other variant.

We should possibly try to do some heuristics when we propagate the configuration forward to see if it's purely default configs and not propagate them if so.

dustymabe:
If the config we are going to copy into the real root is just the
defaults let's just not copy any configuration in because
NetworkManager's "unconfigured" behavior is good enough.

This has side benefits of making the coreos-installer --copy-network
workflow slightly more intuitive.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/696

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
@@ -33,6 +33,39 @@ selinux_relabel() {
     fi
 }
 
+# Determine if the generated NM connection profiles match the default
+# that would be given to us if the user had provided no additional
+# configuration. i.e. did the user give us any network configuration
+# other than the default? We determine this by comparing the generated
+# output of nm-initrd-generator with a new run of nm-initrd-generator.
+# If it matches then it was the default, if not then the user provided
+# something extra.
+are_default_NM_configs() {
+    # Make two dirs for storing files to use in the comparison
+    mkdir -p /run/coreos-teardown-initramfs/connections-compare-{1,2}
+    # Make another that's just a throwaway for the initrd-data-dir
+    mkdir -p /run/coreos-teardown-initramfs/initrd-data-dir
+    # Copy over the previously generated connection(s) profiles
+    cp  /run/NetworkManager/system-connections/* \
+        /run/coreos-teardown-initramfs/connections-compare-1/
+    # Do a new run with the default input
+    /usr/libexec/nm-initrd-generator \
+        -c /run/coreos-teardown-initramfs/connections-compare-2 \
+        -i /run/coreos-teardown-initramfs/initrd-data-dir -- ip=dhcp,dhcp6
+    # remove unique identifiers from the files (so our diff can work)
+    sed -i '/^uuid=/d' /run/coreos-teardown-initramfs/connections-compare-{1,2}/*
+    # currently the output will differ based on whether rd.neednet=1
+    # was part of the kargs. Let's ignore the single difference (wait-device-timeout)
+    sed -i '/^wait-device-timeout=/d' /run/coreos-teardown-initramfs/connections-compare-{1,2}/*
+    if diff -r -q /run/coreos-teardown-initramfs/connections-compare-{1,2}/; then
+        rc=0 # They are the default configs
+    else
+        rc=1 # They are not the defaults, user must have added configuration
+    fi
+    rm -rf /run/coreos-teardown-initramfs
+    return $rc
+}
+
 # Propagate initramfs networking if desired. The policy here is:
 #
 #    - If a networking configuration was provided before this point
@@ -40,7 +73,11 @@ selinux_relabel() {
 #      we do nothing and don't propagate any initramfs networking.
 #    - If a user did not provide any networking configuration
 #      then we'll propagate the initramfs networking configuration
-#      into the real root.
+#      into the real root, but only if it's different than the NM
+#      defaults (trying dhcp/dhcp6 on everything). If it's just the
+#      defaults then we want to avoid a slight behavior diff between
+#      propagating configs and just booting with no configuration. See
+#      https://github.com/coreos/fedora-coreos-tracker/issues/696
 #
 # See https://github.com/coreos/fedora-coreos-tracker/issues/394#issuecomment-599721173
 propagate_initramfs_networking() {
@@ -53,9 +90,13 @@ propagate_initramfs_networking() {
     else
         echo "info: no networking config is defined in the real root"
         if [ -n "$(ls -A /run/NetworkManager/system-connections/)" ]; then
-            echo "info: propagating initramfs networking config to the real root"
-            cp -v /run/NetworkManager/system-connections/* /sysroot/etc/NetworkManager/system-connections/
-            selinux_relabel /etc/NetworkManager/system-connections/
+            if are_default_NM_configs; then
+                echo "info: skipping propagation of default networking configs"
+            else
+                echo "info: propagating initramfs networking config to the real root"
+                cp -v /run/NetworkManager/system-connections/* /sysroot/etc/NetworkManager/system-connections/
+                selinux_relabel /etc/NetworkManager/system-connections/
+            fi
         else
             echo "info: no initramfs networking information to propagate"
         fi
--- a/tests/kola/networking/no-default-initramfs-net-propagation/test.sh
+++ b/tests/kola/networking/no-default-initramfs-net-propagation/test.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# With pure network defaults no networking should have been propagated
+# from the initramfs. This test tries to verify that is the case.
+# https://github.com/coreos/fedora-coreos-tracker/issues/696
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if ! journalctl -t coreos-teardown-initramfs | \
+       grep 'info: skipping propagation of default networking configs'; then
+    echo "no log message claiming to skip initramfs network propagation" >&2
+    fail=1
+fi
+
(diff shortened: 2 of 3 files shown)


dustymabe:
/hold

Need to do some more testing tomorrow.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
```diff
@@ -33,14 +33,50 @@ selinux_relabel() {
     fi
 }
 
+# Add a function that will determine if the generated NM connection
```
jlebon: I'd just start the sentence with:

```suggestion
# Determine if the generated NM connection ...
```

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
```diff
@@ -33,14 +33,50 @@ selinux_relabel() {
...
+# profiles match the default that would be given to us if the user had
+# provided no additional configuration. i.e. did the user give us any
+# network configuration other than the default? We determine this by
+# comparing the generated output of nm-initrd-generator with a new run
+# of nm-initrd-generator. If it matches then it was the default, if not
+# then the user provided something extra.
+are_default_NM_configs() {
+    # Make two dirs for storing files to use in the comparison
+    mkdir -p /run/initramfsteardown/connectionscompare{1,2}
```
jlebon: WDYT about e.g. `/run/coreos-network-teardown/connections-compare-{1,2}`?

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
```diff
@@ -33,14 +33,50 @@ selinux_relabel() {
...
+    # Make another that's just a throwaway for the initrd-data-dir
+    mkdir -p /run/initramfsteardown/initrd-data-dir
+    # Copy over the previously generated connection(s) profiles
+    cp  /run/NetworkManager/system-connections/* /run/initramfsteardown/connectionscompare1
```
jlebon: I think we should handle the trivial case of "no system connections" before this and return early. Right now, our default kargs do mean we always output a connection, but if the user does e.g. `ip=` then nothing gets emitted.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
```diff
@@ -33,14 +33,50 @@ selinux_relabel() {
...
+    # Do a new run with the default input
+    /usr/libexec/nm-initrd-generator -c /run/initramfsteardown/connectionscompare2  \
+                                     -i /run/initramfsteardown/initrd-data-dir -- ip=dhcp,dhcp6
```
jlebon: We could instead reuse the same drop-in we use for Afterburn for this service and here use `AFTERBURN_NETWORK_KARGS_DEFAULT`.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
```diff
@@ -33,14 +33,50 @@ selinux_relabel() {
...
+    # currently the output will differ based on whether rd.neednet=1
+    # was part of the kargs. Let's ignore the single difference (wait-device-timeout)
+    sed -i '/^wait-device-timeout=/d' /run/initramfsteardown/connectionscompare{1,2}/*
+    if diff -r -q /run/initramfsteardown/connectionscompare{1,2}/; then
+        rc=0 # They are the default configs
+    else
+        rc=1 # They are not the defaults, user must have added configuration
+    fi
+    rm -f /run/initramfsteardown/connectionscompare{1,2}/*
+    rmdir /run/initramfsteardown/connectionscompare{1,2}
```
jlebon: Why not just

```suggestion
    rm -rf /run/initramfsteardown
```

?

jlebon (approved):
Approach LGTM overall!

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
dustymabe: updated to `/run/coreos-teardown-initramfs/connections-compare-{1,2}` - matches the script name.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
dustymabe: I thought about this decided against it. If we change `AFTERBURN_NETWORK_KARGS_DEFAULT` to something other than `ip=dhcp,dhcp6` then we'll no longer match "unconfigured" NM behavior and we'll need to start propagating the configs again. I did think about adding a function that will assert `AFTERBURN_NETWORK_KARGS_DEFAULT == ip=dhcp,dhcp6` to alert us when this assumption is no longer valid, but the exttests I added will alert us.

If `AFTERBURN_NETWORK_KARGS_DEFAULT` is no longer `ip=dchp,dhcp6` then we'll start getting configs propagated again and the exttests will fail and we'll have to investigate.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
dustymabe: `rm -rf` scares me, but i'll change it.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
dustymabe: if there are "no system connections" then we'll hit `echo "info: no initramfs networking information to propagate"` from `propagate_initramfs_networking()` and this function never gets called.

dustymabe:
Initial tests on this look good. I do want to run through and do some more comprehensive testing before merge, though.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
jlebon: Right makes sense.

Related to this, I think long-term when we finally can run NM as a systemd service in the initramfs, we should just get rid of the default `ip=dhcp,dhcp6` kargs and instead rely on the default behaviour in NM being the same in the initrd and in the real root. Then, we'll naturally propagate configs only if network kargs were added (because by default `nm-initrd-generator` won't output anything).

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
```diff
@@ -53,9 +90,13 @@ propagate_initramfs_networking() {
     else
         echo "info: no networking config is defined in the real root"
         if [ -n "$(ls -A /run/NetworkManager/system-connections/)" ]; then
-            echo "info: propagating initramfs networking config to the real root"
-            cp -v /run/NetworkManager/system-connections/* /sysroot/etc/NetworkManager/system-connections/
-            selinux_relabel /etc/NetworkManager/system-connections/
+            if are_default_NM_configs; then
+                echo "info: skipping propagate of default networking configs"
```
jlebon:
```suggestion
                echo "info: skipping propagation of default networking configs"
```

?

tests/kola/networking/no-default-initramfs-net-propagation/test.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
jlebon, line 1: I personally would just merge this into `misc-ro`, but separate WFM too.

jlebon (approved):
Minor/optional comments. LGTM as is.

tests/kola/networking/no-default-initramfs-net-propagation/test.sh
dustymabe: Yeah, I'd still prefer to keep it separate for now.

overlay.d/05core/usr/lib/dracut/modules.d/30ignition-coreos/coreos-teardown-initramfs.sh
dustymabe: fixed

PR #774, opened 2020-12-11
core/dracut/ignition-ostree: add ignition-ostree-sysusers service 

lucab:
This introduces a new `ignition-ostree-sysusers.service`, which takes
care of poulating users and groups on the target sysroot before the
Ignition `files` stage.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Needed to work around the initrd `rootfs` / filesystem not being a valid
+# mount to pivot out of. For reference, see:
+#  - https://github.com/torvalds/linux/blob/26bc672134241a080a83b2ab9aa8abede8d30e1c/fs/namespace.c#L3605
+#  - https://gist.github.com/jlebon/fb6e7c6dcc3ce17d3e2a86f5938ec033
+set -euo pipefail
+
+TMP_CHROOT_DIR=""
+
+main() {
+    setup_chroot_tmpdir
+    run_chrooted_bwrap "$@"
+}
+
+setup_chroot_tmpdir() {
+    TMP_CHROOT_DIR=$(mktemp --directory --tmpdir=/mnt '.coreos-sysroot-bwrap.tmp.XXXXXXXXXX')
+    mount --bind / "${TMP_CHROOT_DIR}"
+    mount --make-private "${TMP_CHROOT_DIR}"
+    mount --bind "${TMP_CHROOT_DIR}" "${TMP_CHROOT_DIR}"
+    for mnt in proc sys dev; do
+      mount --bind /$mnt "${TMP_CHROOT_DIR}"/$mnt
+    done
+    touch "${TMP_CHROOT_DIR}"/run/ostree-booted
+    mount --bind /sysroot "${TMP_CHROOT_DIR}"/sysroot
+}
+
+run_chrooted_bwrap() {
+    chroot "${TMP_CHROOT_DIR}" \
+        /usr/bin/env --chdir /sysroot \
+            bwrap \
+            --unshare-pid --unshare-uts --unshare-ipc --unshare-net \
+            --unshare-cgroup-try --dev /dev --proc /proc --chdir / \
+            --ro-bind usr /usr --bind etc /etc --dir /tmp --tmpfs /var/tmp \
+            --tmpfs /run --ro-bind /run/ostree-booted /run/ostree-booted \
+            --symlink usr/lib /lib \
+            --symlink usr/lib64 /lib64 \
+            --symlink usr/bin /bin \
+            --symlink usr/sbin /sbin -- "$@"
+}
+
+cleanup() {
+  if test -z "${TMP_CHROOT_DIR}"; then
+    return
+  fi
+
+  umount --lazy --recursive "${TMP_CHROOT_DIR}"
+  umount --recursive "${TMP_CHROOT_DIR}"
+  rmdir "${TMP_CHROOT_DIR}"
+}
+
+trap cleanup EXIT
+main "$@"
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+# Run systemd-sysusers for the target OSTree sysroot.
+
+set -euo pipefail
+
+main() {
+    coreos-sysroot-bwrap systemd-sysusers
+    coreos-relabel \
+        /etc/group \
+        /etc/group- \
+        /etc/gshadow \
+        /etc/gshadow- \
+        /etc/passwd \
+        /etc/passwd- \
+        /etc/shadow \
+        /etc/shadow-
+}
+
+main "$@"
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Populate OSTree sysusers
+DefaultDependencies=false
+ConditionKernelCommandLine=|ostree
+
+# Need to do this with all mount points active
+After=ignition-mount.service
+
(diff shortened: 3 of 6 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
```diff
@@ -0,0 +1,53 @@
...
+            --symlink usr/sbin /sbin -- "$@"
+}
+
+cleanup() {
+  if test -z "${TMP_CHROOT_DIR}"; then
+    return
+  fi
+
+  umount --lazy --recursive "${TMP_CHROOT_DIR}"
+  umount --recursive "${TMP_CHROOT_DIR}"
```
jlebon, line 48: Interesting. What's the rationale for doing it one time first lazily? Were you hitting `EBUSY` in your tests?

jlebon (approved):
This looks good to me! With this, are you able to refer to users/groups created via systemd-sysusers in your Ignition config?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
lucab: Indeed. As we are cross-layering multiple levels of bind-mounts, they end up in a graph-like shape which can't be unmounted in a linear sweep.

lucab:
@jlebon yes, although testing this right now is skewed due to the pending `dnsmasq` fix at https://src.fedoraproject.org/rpms/dnsmasq/pull-request/6. I can add a regression test on top once that RPM is fixed.

On a custom image without `dnsmasq`, the result looks like this:
```
# grep zincati /usr/lib/passwd /etc/passwd
/etc/passwd:zincati:x:979:979:Zincati user for auto-updates:/:/usr/sbin/nologin

# jq .storage.files[0] /run/ignition.json 
{
  "group": {},
  "path": "/etc/zincati/config.d/empty.toml",
  "user": {
    "name": "zincati"
  },
  "contents": {
    "source": "data:,%23%20Empty%20placeholder%0A",
    "verification": {}
  },
  "mode": 420
}

# ls -laZ /etc/zincati/config.d/empty.toml
-rw-r--r--. 1 zincati root system_u:object_r:etc_t:s0 20 Dec 11 16:22 /etc/zincati/config.d/empty.toml
```

lucab:
Added a test, dropped RFC, this should be ready to merge.

jlebon (dismissed):

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
travier: Anything preventing us from using different mount namespace like we do for /boot RO for example?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+# Run systemd-sysusers for the target OSTree sysroot.
+
+set -euo pipefail
+
+main() {
+    coreos-sysroot-bwrap systemd-sysusers
```
travier, line 7: Any reason not to use `systemd-sysusers --root=root`?

travier:
~~As far as I understand, we could directly call `systemd-sysusers --root=/sysroot`. Any reason not to do that?~~ See: https://github.com/coreos/fedora-coreos-config/pull/774#discussion_r543491937

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers
lucab: That was indeed my first attempt. It didn't work, I think because it ended up mixing/leaking details from the initramfs NSS setup (possibly when looking up existing/duplicate ID, I didn't fully investigate).

jlebon: Yeah, likely due to our use of nss-altfiles (and non-use of nss-altfiles in the initramfs). So in theory once we don't use nss-altfiles anymore, we could probably switch to using `--root`.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
lucab: @travier no, indeed I think we can look into that direction. Are you effectively suggesting a `MountFlags=slave` on the service unit, or something different?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers
travier: Ack

travier:
> This looks good to me! With this, are you able to refer to users/groups created via systemd-sysusers in your Ignition config?

I don't think this will work as this would require splitting the Ignition stage that creates users/groups from the files stage to make a similar call like the one included here to create the users. This will only setup users/groups that are declared in sysusers files in the ostree commit but not included in passwd/shadow/group.

jlebon:
> > This looks good to me! With this, are you able to refer to users/groups created via systemd-sysusers in your Ignition config?
> 
> I don't think this will work as this would require splitting the Ignition stage that creates users/groups from the files stage to make a similar call like the one included here to create the users. This will only setup users/groups that are declared in sysusers files in the ostree commit but not included in passwd/shadow/group.

Are you talking about the case where the Ignition config adds sysusers drop-in configs and in the same Ignition config there are files which reference those users/groups as owners? Hmm, do we really need to handle this? Ignition already provides a declarative way of creating system users/groups and it creates users/groups before creating files.

travier:
> > I don't think this will work as this would require splitting the Ignition stage that creates users/groups from the files stage to make a similar call like the one included here to create the users. This will only setup users/groups that are declared in sysusers files in the ostree commit but not included in passwd/shadow/group.
> 
> Are you talking about the case where the Ignition config adds sysusers drop-in configs and in the same Ignition config there are files which reference those users/groups as owners? Hmm, do we really need to handle this? Ignition already provides a declarative way of creating system users/groups and it creates users/groups before creating files.

Yes. From https://github.com/coreos/fedora-coreos-tracker/issues/155, one of the option is to move to a fully sysusers defined list of users for the system for rpm-ostree & Ignition. This option could enable first boot removal among other things.

travier:
I think we need a larger design discussion for this topic but this has not happened yet as this has not been a priority so far. But this should probably not prevent this PR from being merged.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
```diff
@@ -0,0 +1,53 @@
...
+    mount --make-private "${TMP_CHROOT_DIR}"
+    mount --bind "${TMP_CHROOT_DIR}" "${TMP_CHROOT_DIR}"
+    for mnt in proc sys dev; do
+      mount --bind /$mnt "${TMP_CHROOT_DIR}"/$mnt
+    done
+    touch "${TMP_CHROOT_DIR}"/run/ostree-booted
+    mount --bind /sysroot "${TMP_CHROOT_DIR}"/sysroot
+}
+
+run_chrooted_bwrap() {
```
cgwalters, line 28: This would probably be a good thing to live in rdcore (and in theory in the future it could be a crate shared between rdcore and rpm-ostree).

Not a blocker though.

cgwalters (dismissed):
Did it come up to have Ignition have support for this built in?  It really feels like we're doing this special thing behind its back here that other Ignition users would want.

Ignition still carries code to use `chroot()` for getent lookups which is a "poor-man's" containerization.

cgwalters:
Also I agree strongly with https://github.com/coreos/fedora-coreos-config/pull/774#discussion_r543479676

lucab:
@cgwalters it indeed started with https://github.com/coreos/ignition/pull/1153 on Ignition side but got redirected here.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
travier: Yes, something like `MountFlags=slave` set in the unit that would apply to the whole script

lucab: Ack, added.

dustymabe:
bump

PR #777, opened 2020-12-14
[next-devel] overrides: fast-track afterburn release

sohankunkerkar:


ashcrow (approved):
Package version matches upstream release and Fedora build.

PR #778, opened 2020-12-14
overlay.d: Use `systemctl {add-wants, add-requires}`

kelvinfan001:
Use `systemctl`'s built-in way of adding `Wants=` and `Requires=`
dependencies, since this is the recommended way of adding dependencies
and we can do so with a single command.


travier (approved):

jlebon (approved):

jlebon:
Follow-up to this in https://github.com/coreos/fedora-coreos-config/pull/800.

PR #779, opened 2020-12-14
40ignition-ostree: trigger udev on by-label link mismatch

jlebon:
Sometimes, the `/dev/disk/by-label/root` symlink still points to the
original block device rather than the latest one. This then leads to
e.g. us mounting the wrong device and get errors like:

```
[   54.814061] ignition-ostree-transposefs[938]: mount: /sysroot: unknown filesystem type 'crypto_LUKS'.
```

It's not that we're not waiting long enough for udev to update them.
It's more like udev misses those events entirely. Filed:

bugzilla.redhat.com/show_bug.cgi?id=1908780

Since we know what the underlying device should be (from the Ignition
config), let's just detect those cases and do a `udevadm trigger`.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -42,6 +42,21 @@ mount_verbose() {
     mount "${srcdev}" "${destdir}"
 }
 
+# Sometimes, for some reason the by-label symlinks aren't updated. Detect these
+# cases, and explicitly `udevadm trigger`.
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1908780
+udev_trigger_on_label_mismatch() {
+    local label=$1; shift
+    local expected_dev=$1; shift
+    local actual_dev
+    expected_dev=$(realpath "${expected_dev}")
+    actual_dev=$(realpath "/dev/disk/by-label/$label")
+    if [ "$actual_dev" != "$expected_dev" ]; then
+        echo "Expected /dev/disk/by-label/$label to point to $expected_dev, but points to $actual_dev; triggering udev"
+        udevadm trigger --settle "$expected_dev"
+    fi
+}
+
 # Print partition offset for device node $1
 get_partition_offset() {
     local devpath=$(udevadm info --query=path "$1")
@@ -140,12 +155,16 @@ case "${1:-}" in
         # Mounts happen in a private mount namespace since we're not "offically" mounting
         if [ -d "${saved_root}" ]; then
             echo "Restoring rootfs from RAM..."
+            new_root_dev=$(jq -r "$(query_fslabel root) | .[0].device" "${ignition_cfg}")
+            udev_trigger_on_label_mismatch root "${new_root_dev}"
             mount_verbose "${root_part}" /sysroot
             find "${saved_root}" -mindepth 1 -maxdepth 1 -exec mv -t /sysroot {} \;
             chattr +i $(ls -d /sysroot/ostree/deploy/*/deploy/*/)
         fi
         if [ -d "${saved_boot}" ]; then
             echo "Restoring bootfs from RAM..."
+            new_boot_dev=$(jq -r "$(query_fslabel boot) | .[0].device" "${ignition_cfg}")
+            udev_trigger_on_label_mismatch boot "${new_boot_dev}"
             mkdir -p /sysroot/boot
             mount_verbose "${boot_part}" /sysroot/boot
             find "${saved_boot}" -mindepth 1 -maxdepth 1 -exec mv -t /sysroot/boot {} \;


jlebon:
This works fine locally, but need to test it in RHCOS in CI where it happens more often.

jlebon:
(Obviously, we could also just use the device path from the Ignition config directly, but there are a bunch of other services which assume that `/dev/disk/by-label/root` is up to date.)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -41,6 +41,17 @@ mount_verbose() {
     mount "${srcdev}" "${destdir}"
 }
 
+wait_for_udev() {
+    local label=$1; shift
+    local igndev igndev_real
+    igndev=$(jq -r "$(query_fslabel "$label") | .[0].device" "${ignition_cfg}")
+    igndev_real=$(realpath "$igndev")
```
cgwalters: Perhaps hoist this out as something passed to the function so we're not re-parsing the Ignition config each time?

cgwalters (commented):
Hmm.  I think the race might be if udev hasn't woken up to respond to the kernel changes yet or so?

Shouldn't we instead do something like wait for `/dev/disk/by-uuid/<uuid>` that we receive from Ignition somehow?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: Not sure I follow -- this function is called twice only, each with a different parameter.

jlebon:
> Shouldn't we instead do something like wait for `/dev/disk/by-uuid/<uuid>` that we receive from Ignition somehow?

We'd still want to make sure that the `/dev/disk/by-label/root` symlink points to the right thing too because we have a bunch of things which rely on that.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
cgwalters: OK yep I misread.

cgwalters:
> We'd still want to make sure that the /dev/disk/by-label/root symlink points to the right thing too because we have a bunch of things which rely on that.

Yeah, true.  Which...hmm, we could invent our own `/dev/coreos-rootfs` symlink too...but eh.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
cgwalters: Actually now that I look again, how about we change `wipes_root` to something like `replaces_root_dev` that if non-empty is the path to the device?

jlebon: I think that makes sense. Though I've got a full `kola run` of RHCOS going in the cluster with the current patch that's pretty far along. If it passes, let's keep that as a follow-up instead and get this in? If I need to iterate again, I can fold it in.

jlebon:
OK, more info on this. Testing it in RHCOS in the cluster known to trigger this bug demonstrates that it's worse than that; it seems like udev *never* actually updates the symlinks even if we give it all the time in the world:

```
...
[ ***  ] A start job is running for Ignition…rtitions (1h 23min 32s / no limit)
[ 5016.327922] ignition-ostree-transposefs[931]: Waiting for /dev/disk/by-label/root to point to /dev/dm-0
[***   ] A start job is running for Ignition…rtitions (1h 23min 33s / no limit)
...
```

I'm testing a `udevadm trigger && udevadm settle` version of this patch now.

jlebon:
OK, this works now! Confirmed that it kicks in on RHCOS and saves the day:

```
$ grep -nr 'Expected ' tmp/kola/qemu-unpriv-latest/
tmp/kola/qemu-unpriv-latest/luks.sss.t11/8f74db81-87f9-4a0f-8d72-4303072d79b4/journal.txt:766:Dec 17 01:49:57.954433 ignition-ostree-transposefs[1060]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-latest/luks.sss.t11/8f74db81-87f9-4a0f-8d72-4303072d79b4/console.txt:832:[   64.492239] ignition-ostree-transposefs[1060]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-latest/luks.tang3/32b4f34f-5508-4316-b5e0-571e04c74dc6/journal.txt:767:Dec 17 02:13:19.686982 ignition-ostree-transposefs[1007]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-latest/luks.tang3/32b4f34f-5508-4316-b5e0-571e04c74dc6/console.txt:832:[   67.358294] ignition-ostree-transposefs[1007]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
```

(This is from a run of `kola run luks.* --multiply 5`; got all passes!)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: > Actually now that I look again, how about we change `wipes_root` to something like `replaces_root_dev` that if non-empty is the path to the device?

I initially did this but then undid it because (1) the `wipes_root` bit is in the `detect` path only right now, so we'd need to hoist it up out of the `case` if we wanted to use it in the `restore` path too, which would actually make us parse the Ignition config more times than previously, and (2) it breaks the nice "00000" comparison that we currently do to detect no-ops.

I did tweak it though so that we pass in the paths instead of having the function do the parsing.

dustymabe:
> OK, more info on this. Testing it in RHCOS in the cluster known to trigger this bug demonstrates that it's worse than that; it seems like udev _never_ actually updates the symlinks even if we give it all the time in the world

Is there an upstream bug in here somewhere?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -42,6 +42,19 @@ mount_verbose() {
...
 
+udev_trigger_on_label_mismatch() {
+    local label=$1; shift
+    local expected_dev=$1; shift
+    local actual_dev
+    expected_dev=$(realpath "${expected_dev}")
+    actual_dev=$(realpath "/dev/disk/by-label/$label")
+    if [ "$actual_dev" != "$expected_dev" ]; then
+        echo "Expected /dev/disk/by-label/$label to point to $expected_dev, but points to $actual_dev; triggering udev"
+        udevadm trigger
```
bgilbert: Should we trigger the affected device rather than... everything?

bgilbert (commented):
LGTM generally.  I agree with @dustymabe; let's add an upstream bug link.

bgilbert:
And thanks for tracking this down!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: Done!

jlebon:
OK, updated with link to newly filed bug: https://bugzilla.redhat.com/show_bug.cgi?id=1908780.
Retested this latest version in RHCOS and got all passes again. Can see it kick in:

```
$ grep -nr 'Expected ' tmp/kola
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.sss.t16/10cfce32-8969-4577-9709-51803dd1d5ee/journal.txt:763:Dec 17 13:41:36.379838 ignition-ostree-transposefs[1061]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.sss.t16/10cfce32-8969-4577-9709-51803dd1d5ee/console.txt:832:[   67.002299] ignition-ostree-transposefs[1061]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.sss.t10/8b595796-51d5-4d9d-ad28-63269c34e87f/journal.txt:764:Dec 17 14:11:06.120711 ignition-ostree-transposefs[1064]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.sss.t10/8b595796-51d5-4d9d-ad28-63269c34e87f/console.txt:832:[   59.543255] ignition-ostree-transposefs[1064]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.tang5/01379b02-b3b4-4e9c-b2a9-999c712199b1/journal.txt:766:Dec 17 14:17:28.263910 ignition-ostree-transposefs[1007]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.tang5/01379b02-b3b4-4e9c-b2a9-999c712199b1/console.txt:832:[   56.674278] ignition-ostree-transposefs[1007]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.sss.t11/2f0bf548-68e7-4686-b7a7-afd0238bcc91/journal.txt:765:Dec 17 14:21:38.241459 ignition-ostree-transposefs[1064]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/luks.sss.t11/2f0bf548-68e7-4686-b7a7-afd0238bcc91/console.txt:832:[   60.653285] ignition-ostree-transposefs[1064]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/ext.config.luks.tpm1/46e72c73-532a-4dde-83f4-8d6a82d8c60e/journal.txt:689:Dec 17 14:23:32.308290 ignition-ostree-transposefs[955]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
tmp/kola/qemu-unpriv-2020-12-17-1338-17706/ext.config.luks.tpm1/46e72c73-532a-4dde-83f4-8d6a82d8c60e/console.txt:758:[   56.713224] ignition-ostree-transposefs[955]: Expected /dev/disk/by-label/root to point to /dev/dm-0, but points to /dev/vda4; triggering udev
...
```

I'd like to get this in and into RHCOS soon.

cgwalters (approved):
LGTM

PR #780, opened 2020-12-15
effectively revert use of systemd-resolved

dustymabe:
This one is complicated, we need to revert the systemd change
in f33 that makes it fallback to `fedora` rather than `localhost`
for the hostname and we also need to essentially render systemd-resolved
ineffective because it's internal fallbacks for reverse DNS make
it really hard to set the hostname via reverse DNS, which is
something we don't want to break and needs upstream work to
get more appropriate fixes in.

More context in https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-736104003

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -1,4 +1,20 @@
 packages:
+  # Fast-track systemd update that reverts fallback hostname change
+  # https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-739956843
+  # This is a one-off build, please don't remove it without talking
+  # to dustymabe or jlebon.
+  systemd:
+    evra: 246.7-1.fc33.aarch64
+  systemd-container:
+    evra: 246.7-1.fc33.aarch64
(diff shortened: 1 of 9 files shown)


dustymabe:
still needs a new official build of selinux policy

dustymabe:
selinux policy build is hot off the press

jlebon (approved):
Awesome, thanks for seeing this through!

LGTM overall. I think it would've been useful to have all these new units and scripts cross-reference each other so that we're sure we've removed all of it when we get there, but not a blocker.

Just as a sanity-check: were you able to test this in an environment with rDNS answers?

dustymabe:
ok just pushed up a comment change. going to skip the CI this time since we're in a time crunch.

keszybz:
I scanned this PR and I don't see anything wrong, fwiw.

BTW, systemd-247.2-1.fc34 has the selinux patch.

dustymabe:
Thanks @keszybz - let's talk more about this tomorrow or Friday if you're free.

PR #786, opened 2020-12-16
use WALinuxAgent-udev package for Azure udev rules

miabbott:
BZ#1748432 was filed while working on #159 and the udev rules were
split out into a separate package as part of F32.  Let's stop carrying
the rules in the overlay and use the provided package now.


miabbott:
I stumbled on this while doing some triage on a downstream BZ; see the comments here showing that the udev rules are not in the initrd of RHCOS:

https://bugzilla.redhat.com/show_bug.cgi?id=1756173#c4
https://bugzilla.redhat.com/show_bug.cgi?id=1756173#c6

This PR by itself builds fine (haven't actually tested in Azure yet).  I think I also need a simple dracut module to have the rules added to the initrd.  Any suggestions on the right ordering?  (i.e. `05core`, somewhere else, something entirely new?)

lucab:
> see the comments here showing that the udev rules are not in the initrd of RHCOS

It looks like we'd need the package to be updated to install them into the initramfs too, not only in the rootfs.

More generally, it looks like we'd benefit from having this [dracut shortcoming](https://github.com/dracutdevs/dracut/blob/28d0fc93476da59498af3814f137b94baf193bb8/modules.d/95udev-rules/module-setup.sh#L7-L9) fixed.

miabbott:
Never  dabbled in `dracut` modules before, but I came up with this that seems to get the rules installed:

```diff
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/25coreos-udev/module-setup.sh b/overlay.d/05core/usr/lib/dracut/modules.d/25coreos-udev/module-setup.sh
new file mode 100644
index 0000000..4ad324d
--- /dev/null
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/25coreos-udev/module-setup.sh
@@ -0,0 +1,13 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# We want to provide Azure udev rules as part of the initrd, so that Ignition
+# is able to detect disks and act on them.
+
+install() {
+    inst_simple WALinuxAgent-udev
+    inst_multiple \
+        /usr/lib/udev/rules.d/66-azure-storage.rules \
+        /usr/lib/udev/rules.d/99-azure-product-uuid.rules
+}
```

dustymabe:
would it be most appropriate to add the dracut snippet to the `WALinux-udev` RPM?

miabbott:
> would it be most appropriate to add the dracut snippet to the `WALinux-udev` RPM?

I think that is a good long-term solution; I'm proposing this because it is something that we would like to include in RHCOS in the near term.

cgwalters:
> would it be most appropriate to add the dracut snippet to the WALinux-udev RPM?

I think this is only relevant on Ignition based systems though.

dustymabe:
> > would it be most appropriate to add the dracut snippet to the WALinux-udev RPM?
> 
> I think this is only relevant on Ignition based systems though.

If it helps the argument, it's worth noting that we'd like to support Ignition as an option for the cloud base image at some point in the future.

miabbott:
> If it helps the argument, it's worth noting that we'd like to support Ignition as an option for the cloud base image at some point in the future.

Filed a BZ to discuss this further with the maintainers - https://bugzilla.redhat.com/show_bug.cgi?id=1909287

dustymabe:
Just clearing up my stance: I don't think we need to block on having the dracut bits in the RPM. The open request in the BZ is plenty.

jlebon:
WFM too to just ship the dracut glue to add the rules to the initrd for now. What you have in https://github.com/coreos/fedora-coreos-config/pull/786#issuecomment-747601499 mostly LGTM. (You should be able to drop the `inst_simple WALinuxAgent-udev` bit, and maybe name it e.g. `25coreos-azure-udev` instead and add a link to https://bugzilla.redhat.com/show_bug.cgi?id=1909287 in there?)

miabbott:
Incorporated feedback from @jlebon and successfully tested an FCOS image in Azure.  (Our docs for booting on Azure are really good and easy to follow!)

Was able to successfully boot the image like this:

`az vm create -n "${az_vm_name}" -g "${az_resource_group}" --image "${az_image_name}" --admin-username core --custom-data "$(cat ${ignition_path})" --attach-data-disks data0`

And used an Ignition config that had the following:

```json
{
  "ignition": {
    "version": "3.2.0"
  },
  "passwd": {
    "users": [
      {
        "name": "miabbott",
        "sshAuthorizedKeys": [
          "ssh-rsa AAAAB3Nz..."
        ]
      }
    ]
  },
  "storage": {
    "disks": [
      {
        "device": "/dev/disk/azure/scsi1/lun0",
        "partitions": [
          {
            "label": "varlibetcd",
            "sizeMiB": 0,
            "startMiB": 0
          }
        ]
      }
    ],
    "filesystems": [
      {
        "device": "/dev/disk/by-partlabel/varlibetcd",
        "format": "xfs",
        "path": "/var/lib/etcd"
      }
    ]
  },
  "systemd": {
    "units": [
      {
        "contents": "# Generated by FCCT\n[Unit]\nBefore=local-fs.target\nRequires=systemd-fsck@dev-disk-by\\x2dpartlabel-varlibetcd.service\nAfter=systemd-fsck@dev-disk-by\\x2dpartlabel-varlibetcd.service\n\n[Mount]\nWhere=/var/lib/etcd\nWhat=/dev/disk/by-partlabel/varlibetcd\nType=xfs\n\n[Install]\nRequiredBy=local-fs.target",
        "enabled": true,
        "name": "var-lib-etcd.mount"
      }
    ]
  }
}
```

jlebon (approved):
Typo in the commit message (`WALinux-udev` instead of `WALinuxAgent-udev`). LGTM otherwise!

So, I just realized that the el8 version of `WALinuxAgent` doesn't yet have a `-udev` subpackage. Were you planning to request that, or just keep shipping the rules there as FCOS did before this PR? (The divergence makes it slightly less appealing, though it's also nice to test out this cleanup in FCOS first.)

miabbott:
> Typo in the commit message (`WALinux-udev` instead of `WALinuxAgent-udev`). LGTM otherwise!
> 
> So, I just realized that the el8 version of `WALinuxAgent` doesn't yet have a `-udev` subpackage. Were you planning to request that, or just keep shipping the rules there as FCOS did before this PR? (The divergence makes it slightly less appealing, though it's also nice to test out this cleanup in FCOS first.)

Dang, didn't realize the `-udev` package wasn't present in el8.  I have doubts about getting that package built and shipped in time for RHCOS 4.7, so we'll probably have to carry the rules for RHCOS and install them in the initramfs separately.

miabbott:
Filed https://bugzilla.redhat.com/show_bug.cgi?id=1913074 for the el8 subpackage

miabbott:
Fixed up the commit message :arrow_up:

jlebon (approved):

PR #787, opened 2020-12-17
overlay.d/15fcos: Stop utilizing c-l-h-m private dir

kelvinfan001:
Drop issue file snippets directly into `/etc/issue.d`, which
is read by `agetty`.

Following https://github.com/coreos/console-login-helper-messages/pull/91,
new versions of console-login-helper-messages will no longer
support users dropping files into its private directories
(e.g. `/run/console-login-helper-messages/issue.d`).

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
@@ -16,7 +16,11 @@ nc='\033[0m'
 output=$(journalctl -o json-pretty MESSAGE_ID=57124006b5c94805b77ce473e92a8aeb | jq -s '.[] | select(.IGNITION_CONFIG_TYPE == "user")'| wc -l)
 
 if [[ $output -gt  0 ]];then
-   echo "Ignition: user provided config was applied" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+    echo "Ignition: user provided config was applied" > /etc/issue.d/30_ignition_config_info.issue
 else
-  echo -e "${warn}Ignition: no config provided by user${nc}" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+    echo -e "${warn}Ignition: no config provided by user${nc}" > /etc/issue.d/30_ignition_config_info.issue
 fi
+
+# Ask all running agetty instances to reload and update their
+# displayed prompts in case this script was run before agetty.
+/usr/sbin/agetty --reload
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys.sh
@@ -36,11 +36,15 @@ main() {
     fi
 
     if [ -n "$output" ]; then
-        echo "$output" > /run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue
+        echo "$output" > /etc/issue.d/30_ssh_authorized_keys.issue
     else
         echo -e "${warn}No ssh authorized keys provided by Ignition or Afterburn${nc}" \
-             > /run/console-login-helper-messages/issue.d/30_ssh_authorized_keys.issue
+            > /etc/issue.d/30_ssh_authorized_keys.issue
     fi
+
+    # Ask all running agetty instances to reload and update their
+    # displayed prompts in case this script was run before agetty.
+    /usr/sbin/agetty --reload
 }
 
 main


overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
```diff
@@ -2,6 +2,7 @@
 # no Ignition config is provided.
 [Unit]
 Description=Check if Ignition config is provided
+Before=serial-getty@.service
```
kelvinfan001: Theoretically I think we want this to run _before_ `agetty` is run, but local testing seems to show that this isn't necessary. Do we need to keep this just in case?

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
```diff
@@ -6,6 +6,7 @@ Description=Check that ssh-keys are added by Afterburn/Ignition
 # of `afterburn-sshkeys@` and not just the `core` user.
 # See https://github.com/coreos/afterburn/pull/481
 After=afterburn-sshkeys.target
+Before=serial-getty@.service
```
kelvinfan001: Same as above, not sure if this is necessary.

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
```diff
@@ -16,7 +16,7 @@ nc='\033[0m'
 output=$(journalctl -o json-pretty MESSAGE_ID=57124006b5c94805b77ce473e92a8aeb | jq -s '.[] | select(.IGNITION_CONFIG_TYPE == "user")'| wc -l)
 
 if [[ $output -gt  0 ]];then
-   echo "Ignition: user provided config was applied" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+   echo "Ignition: user provided config was applied" > /etc/issue.d/30_ignition_config_info.issue
 else
-  echo -e "${warn}Ignition: no config provided by user${nc}" > /run/console-login-helper-messages/issue.d/30_ignition_config_info.issue
+  echo -e "${warn}Ignition: no config provided by user${nc}" > /etc/issue.d/30_ignition_config_info.issue
 fi
```
dustymabe: hmm.. We want these files cleaned up on boot? If so we need to add the `*_clhm_*` identifier in the file name?

dustymabe: same comment for the auth keys files below too

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
dustymabe: If there is a dependency I think it's better to be explicit. I guess one way to "test" is to flip it around and say `After=serial-getty@.service` and see if it still works. If it does then we can probably remove it.

I also wonder if we do keep this if we should also add a `Before=getty@.service` for the VGA console case.

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
kelvinfan001: ah right, I forgot about that. Yes, we'll need `*_clhm_*`.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
kelvinfan001: Just tested, seems like `After=serial-getty@.service` works too. I'm not sure why that is, though.

dustymabe: I guess it's OK because we call `agetty --reload` in various places. I suspect if we didn't have those calls we'd need `Before=serial-getty@.service`?

lucab:
This is perhaps a naive question as I haven't been closely following this topic, but why are are we re-targeting from `/run` to `/etc`?

cgwalters:
I haven't been closely following the c-l-h-m rearchitecting; will this work on RHEL8?  I had the vague impression that all the new design required new util-linux. 

Oh but nevermind, this goes into `15fcos` which is *only* fcos.

kelvinfan001:
@lucab In newer versions of CLHM, we will be dropping files directly into the public `/run/motd.d` and `/etc/issue.d` directories so tools like `login(1)` and `agetty` and read directly from these directories and display the snippets we want. This significantly simplifies CLHM as CLHM was originally duplicating the work of concatenating snippets from a directory, monitoring its private directories, etc. This logic was the source of most of CLHM's bugs, so we decided it'd be better to let the upstream tools handle it. Because of this, CLHM no longer supports the `/run/console-login-helper-messages/issue.d` directory. 
Re: why `/etc/issue.d` instead of `/run/issue.d`, see [here](https://github.com/coreos/console-login-helper-messages/issues/60#issuecomment-740574677). TL;DR, `agetty` does not read from `/run/issue` by default if `/etc/issue` exists; but currently, systemd makes it virtually impossible to remove `/etc/issue` and we also didn't want to configure `agetty` with our custom options, so we opted for using the somewhat temporary/hacky location of `/etc/issue.d`.

kelvinfan001:
@cgwalters Yeah we'll essentially be maintaining a separate version of CLHM for RHEL until it catches up. New features and development on CLHM are pretty slow so it should be okay to maintain the two versions separately.

dustymabe:
> TL;DR, `agetty` does not read from `/run/issue` by default if `/etc/issue` exists; but currently, systemd makes it virtually impossible to remove `/etc/issue` and we also didn't want to configure `agetty` with our custom options, so we opted for using the somewhat temporary/hacky location of `/etc/issue.d`.

yeah, it was a bit of a mess and this was the best option we came up with without spending significant efforts in multiple upstreams for not a lot of gain.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
kelvinfan001: Ah yes, that would make sense. Perhaps it'd be better to `agetty --reload` here as well then, to keep things consistent.

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config.sh
dustymabe: actually, now that I think about it a little bit more deeply. It might be nice to *change the behavior* leave these messages in place for subsequent boots too. I think they'd be useful for more than just the first boot. WDYT?

kelvinfan001: Sounds to me like it would be useful in subsequent boots as well. But I also see from its corresponding .service units.
```
# Only perform checks on the first (Ignition) boot as they are
# mostly useful only on that boot.
```

kelvinfan001: But yeah, I think keeping them there would be OK. It also has the added bonus of having this service+script be completely independent of CLHM.

dustymabe: Yeah I think we only perform the checks on first boot because it's possible the logs get garbage collected at some point so the "check" would stop giving the right output eventually. But if we can leave the results of the check around then that works even better I think.

kelvinfan001: Cool, I've removed it now :)

dustymabe (approved):
LGTM

PR #791, opened 2020-12-17
40ignition-ostree/transposefs: load zram with num_devices=0

jlebon:
By default, upon inserting the zram module, it'll automatically create
one zram device. Because we don't know if we're the first ones or not,
it's not safe to make use of that device.

Instead, tell it to not do that so that if we *are* the first ones to
load the module, we don't end up with an extra zram device. If we
aren't, it's a no-op.


jlebon:
From noticing this in the logs:

```
[    6.666133] ignition-ostree-transposefs[437]: Detected partition replacement in fetched Ignition config: /run/ignition.json
[    6.763477] zram: Added device: zram0
[    6.767329] zram: Added device: zram1
[    6.776523] zram1: detected capacity change from 0 to 10737418240
```

bgilbert (approved):
LGTM :+1:

PR #793, opened 2020-12-18
[rawhide] Setup rawhide stream

jlebon:
This is forked and adapted from testing-devel:
- drop lockfiles, this is a mechanical stream
- bump releasever to 34
- switch to using the rawhide repo
- substitute `testing-devel` to `rawhide` in `manifest.yaml`


jlebon:
We can ignore CI for now. The current failure there will be fixed by https://src.fedoraproject.org/rpms/cifs-utils/pull-request/6.

I'm just looking to have all the pieces in place for now so that we get builds going (even if they fail) in the pipeline.

dustymabe (approved):
LGTM

dustymabe:
one thing we'll need to think about is how/if we want to be able to override packages when things break. Right now we haven't polluted `coreos-pool` with anything that isn't going to ship in an production stream release.

PR #794, opened 2020-12-20
40ignition-ostree: copy ESP contents as independent filesystems

### LINKED ISSUE #581: metal: Support redundant bootable disks
We're about to [gain support for many types of complex storage for the *root filesystem*](https://github.com/coreos/fedora-coreos-config/pull/503), but this is distinct from the "root disk" which includes `/boot` and the ESP for example.

We had a subthread around this starting around here: https://github.com/coreos/fedora-coreos-tracker/issues/94#issuecomment-650582087

Basically my proposal is that we support "equivalent RAID 1" by teaching `coreos-installer` how to replicate the `/boot` partition and ESP to multiple block devices.  Making the root filesystem RAID is an orthogonal thing - for example, one might choose to use a higher RAID level for the OS and data, or one could use Stratis/btrfs/LVM for `/`.  Another important variation here is to use LUKS-on-RAID (or equivalent) for the root device.  To restate, the root filesystem can be distinct from how `/boot` and the ESP works.

I *think* it would work in this scenario to make `/boot` be MD-RAID - the [bootloaders support that](https://wiki.archlinux.org/index.php/LVM_on_software_RAID#GRUB_legacy) and it would mean having ostree update kernels would be fully transparent.
We can't make the ESP be RAID - that would need to be manually sync'd.

### LINKED ISSUE #694: Do not mount /boot/efi by default
## Objectives

To avoid issues on AWS and to better convey the fact that users and administrators should not manually alter the content of the EFI partition, we want to not mount `/boot/efi` / the EFI partition by default.

The only programs on the system that should alter ESP partition content should be bootupd and fwupd and special support will be included for them to operate properly.

## References

Split from https://github.com/coreos/fedora-coreos-tracker/issues/652

From https://github.com/coreos/fedora-coreos-config/pull/407:
> Nothing in the OS touches the ESP by default, so there's no reason to mount it by default, particularly writable. This is good for avoiding wear&tear on the filesystem, but I am specifically doing this as preparation for potentially removing the ESP from AWS images, because AWS ImportImage chokes on its presence: openshift/os#396

See also: https://github.com/coreos/fedora-coreos-config/pull/356

bgilbert:
If the firmware writes to an individual ESP replica, the RAID will desynchronize.  Linux `md` will return reads from either replica, and then any dependent writes could corrupt the filesystem.  To prevent this, FCCT will no longer put the ESP on a RAID (https://github.com/coreos/fcct/pull/178); instead we create multiple independent filesystems and copy the contents to each.  This is okay because bootupd and fwupd should be the only things that care about the contents of the ESP.

Drop the mount unit for `/boot/efi`, since there's no longer going to be a "canonical ESP" to mount.

Don't worry too much about backward compatibility because we're making this change soon after the functionality landed, and before it was documented.  For the record, old configs will fail on new systems (because the partitions will be RAID members) but new configs will skip copying `/boot` on old systems (because there's no filesystem labeled `EFI-SYSTEM`).

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/694 and [RHBZ 1909453](https://bugzilla.redhat.com/show_bug.cgi?id=1909453).  Closes https://github.com/coreos/fedora-coreos-tracker/issues/581.  Obsoletes https://github.com/coreos/fedora-coreos-config/pull/407.  Discussion in https://github.com/coreos/fedora-coreos-config/pull/718 and design doc update in https://github.com/coreos/enhancements/pull/4.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -63,7 +63,7 @@ get_partition_offset() {
     cat "/sys${devpath}/start"
 }
 
-mount_and_restore_filesystem() {
+mount_and_restore_filesystem_by_label() {
     local label=$1; shift
     local mountpoint=$1; shift
     local saved_fs=$1; shift
@@ -167,23 +167,23 @@ case "${1:-}" in
         # Mounts happen in a private mount namespace since we're not "offically" mounting
         if [ -d "${saved_root}" ]; then
             echo "Restoring rootfs from RAM..."
-            mount_and_restore_filesystem root /sysroot "${saved_root}"
+            mount_and_restore_filesystem_by_label root /sysroot "${saved_root}"
             chcon -v --reference "${saved_root}" /sysroot  # the root of the fs itself
             chattr +i $(ls -d /sysroot/ostree/deploy/*/deploy/*/)
         fi
         if [ -d "${saved_boot}" ]; then
             echo "Restoring bootfs from RAM..."
-            mount_and_restore_filesystem boot /sysroot/boot "${saved_boot}"
+            mount_and_restore_filesystem_by_label boot /sysroot/boot "${saved_boot}"
             chcon -v --reference "${saved_boot}" /sysroot/boot  # the root of the fs itself
         fi
         if [ -d "${saved_esp}" ]; then
             echo "Restoring EFI System Partition from RAM..."
-            mount_and_restore_filesystem EFI-SYSTEM /sysroot/boot/efi "${saved_esp}"
+            mount_and_restore_filesystem_by_label EFI-SYSTEM /sysroot/boot/efi "${saved_esp}"
         fi
         if [ -d "${saved_bios}" ]; then
             echo "Restoring BIOS Boot partition and boot sector from RAM..."
             expected_start=$(cat "${saved_bios}/start")
-            # iterate over each new BIOS Boot partition, by label
+            # iterate over each new BIOS Boot partition, by partlabel
             jq -r "$(query_parttype ${bios_typeguid}) | .[].label" "${ignition_cfg}" | while read label; do
                 cur_part="/dev/disk/by-partlabel/${label}"
                 # boot sector hardcodes the partition start; ensure it matches
@@ -201,7 +201,7 @@ case "${1:-}" in
         fi
         if [ -d "${saved_prep}" ]; then
             echo "Restoring PReP partition from RAM..."
-            # iterate over each new PReP partition, by label
+            # iterate over each new PReP partition, by partlabel
             jq -r "$(query_parttype ${prep_typeguid}) | .[].label" "${ignition_cfg}" | while read label; do
                 cat "${saved_prep}/partition" > "/dev/disk/by-partlabel/${label}"
             done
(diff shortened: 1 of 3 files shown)


overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -72,18 +72,4 @@ EOF
 # the dependency on the underlying device unit.
 if [ ! -f /run/ostree-live ]; then
     mk_mount /boot boot
-
```
jlebon, line 75: I see it listed in the PR message, though I usually prefer the `Closes` footers to be in the commit messages, because they end up in the git history. So e.g. for this commit, `Closes: https://github.com/coreos/fedora-coreos-tracker/issues/694`?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -176,7 +178,20 @@ case "${1:-}" in
         fi
         if [ -d "${saved_esp}" ]; then
             echo "Restoring EFI System Partition from RAM..."
-            mount_and_restore_filesystem_by_label EFI-SYSTEM /sysroot/boot/efi "${saved_esp}"
+            # iterate over each new ESP partition, by partlabel
+            i=1
+            jq -r "$(query_parttype ${esp_typeguid}) | .[].label" "${ignition_cfg}" | while read label; do
```
jlebon: Optional: at this point, might be worth factoring out a e.g. `get_partlabels_for_guid()` and do:

```sh
for label in $(get_partlabels_for_guid ${esp_typeguid}); do
   ...
done
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -176,7 +178,20 @@ case "${1:-}" in
...
+                # Don't use mount_and_restore_filesystem_by_label because:
+                # 1. We're mounting by partlabel, not FS label
+                # 2. We need to copy the contents to each partition, not move
+                #    them once
+                # 3. We don't need the by-label symlink to be correct and
+                #    nothing later in boot will be mounting the filesystem
+                mountpoint="/mnt/efi-system-$i"
```
jlebon: Optional: why not just avoid the `$i` and use e.g. `/mnt/efi-system-$label`? We've already enforced uniqueness.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -2,6 +2,7 @@
 set -euo pipefail
 
 boot_sector_size=440
+esp_typeguid=c12a7328-f81f-11d2-ba4b-00a0c93ec93b
```
jlebon, line 5: And this commit, `Closes: https://github.com/coreos/fedora-coreos-tracker/issues/581`?

jlebon (approved):
Some minor comments, but LGTM overall!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
bgilbert: 581 was reopened only to track that the original implementation was going to change, so this commit doesn't really fix 581 in a fundamental way.  Closing that issue is more of a bookkeeping thing.

bgilbert:
Updated!

jlebon (approved):

PR #623, opened 2020-09-18
[testing] tree: promote changes from testing-devel at 1f24ad0266ef992b9535a6f71…

cverna:
…ee683ad60f73844

PR #795, opened 2020-12-21
Switch to sqlite rpmdb backend

jlebon:
There won't be any support for writing to the bdb backend in f34, so
e.g. pkglayering won't work (and obviously even composes wouldn't work
once we move cosa to f34).

All the production streams now have an f33 barrier release (as part of
the resolved workarounds), so it should be safe to switch to sqlite now.

For the other rpm-ostree-based variants, we'll probably just flip the
default in rpm-ostree to sqlite in f34. But at least in FCOS, we can do
this now so that we flush out any issues earlier and squash the ugly
warnings when querying the rpmdb.

Closes: #623

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -40,6 +40,9 @@ check-groups:
 
 default-target: multi-user.target
 
+# we can drop this when it's the rpm-ostree default
+rpmdb: sqlite
+
 # ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
 # See also the version of this in fedora-coreos.yaml
 postprocess:
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -114,12 +114,12 @@ if [ -n "${unlabeled}" ]; then
 fi
 ok no files with unlabeled_t SELinux label
 
-# make sure we stick with bdb until we're ready to move to sqlite
+# make sure we're using the sqlite rpmdb backend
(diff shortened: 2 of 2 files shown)


jlebon:
Let's keep this as draft until the new year and e.g. one more round of releases to be safe.

dustymabe (approved):
LGTM - love that we have a simple test and how easy it is to switch/verify the opposite behavior.

Any potential issues with upgrading systems (with or without package layering)?

cgwalters (approved):

jlebon:
> Any potential issues with upgrading systems (with or without package layering)?

The main risk is if a user tries to do pkglayering on top of an f33/sqlite commit from an f32 deployment. Now, that should be pretty rare because we do have a barrier f33 release, and they'd be skipping over that. But e.g. they could have reverted from the barrier release back to the f32 fallback, and then tried a direct `rpm-ostree upgrade` instead after a new update with the f33/sqlite change is available. The bit about

> and e.g. one more round of releases to be safe

is to widen the distance from that possibly happening.

But then again, I think at this point we really require nodes to go through the barriers, because our maintenance efforts rely on assuming that all the precise state mutations we've rolled out actually took place.

jlebon:
New year and we've had another round of releases, so I think we're ready to do this!

jlebon:
This already had two approvals, so just restarted CI and marked for auto-merge.

cgwalters:
This is causing https://github.com/coreos/rpm-ostree/issues/2435
Haven't fully debugged why though.  The code paths here are rather complex.

cgwalters:
Possibly some variant of https://github.com/openSUSE/libsolv/pull/386

jlebon:
OK, let's revert for now: https://github.com/coreos/fedora-coreos-config/pull/812.

PR #800, opened 2021-01-04
05core: re-order and rename some dracut modules

### LINKED ISSUE #799: missing `ignition-diskful` and `ignition-complete` targets when building initrd
Saw the following when doing `coreos-assembler build`; they don't appear to be fatal or even affect operation, but thought I would log the messages anyways

```
dracut: *** Including module: i18n ***                                                                                                                                                                                                                                                                                        
dracut: *** Including module: coreos-network ***                                                                                                               
Failed to add dependency on unit, unit ignition-diskful.target does not exist.                                                                                                                                                                                                                                                
dracut: *** Including module: live ***                                                                                                                                                                                                                                                                                        
Failed to add dependency on unit, unit ignition-complete.target does not exist.                                                                                
dracut: *** Including module: coreos-udev ***                                                                                                                                                                                                                                                                                 
dracut: *** Including module: afterburn ***      
```

jlebon:
Unlike `ln -sf`, `systemctl add-{requires,wants}` wants to verify that
the target and member units all exist. Because we had some dracut
modules which ran earlier than the `30ignition` module, we would lose
the service enablements which hook into e.g. `ignition-complete.target`
and related units.

Let's simply rename our modules so that they're ordered after
`30ignition`. While we're here, let's consistently prefix them by
`coreos-`. So overall, the changes are:
- `30ignition-coreos` -> `35coreos-ignition`
- `20live` -> `35coreos-live`
- `15coreos-network` -> `35coreos-network`

I've verified that there are now no "Failed to add dependency on unit"
error messages at build-time. (The next patch enforces this more
strictly.)

Fixes: #778
Closes: #799


cgwalters (approved):
LGTM

dustymabe (approved):
LGTM

I almost wonder if we should add some of this context to `overlay.d/README.md`.

jlebon:
> I almost wonder if we should add some of this context to `overlay.d/README.md`.

Having it in `overlay.d/README.md` felt too high-level, so instead I added some context in the locations where we do `|| exit 1` (which essentially includes all the `module-setup.sh` files of the dracut modules which are affected by this).

PR #803, opened 2021-01-04
[stable] 05core: re-order and rename some dracut modules

jlebon:
Unlike `ln -sf`, `systemctl add-{requires,wants}` wants to verify that
the target and member units all exist. Because we had some dracut
modules which ran earlier than the `30ignition` module, we would lose
the service enablements which hook into e.g. `ignition-complete.target`
and related units.

Let's simply rename our modules so that they're ordered after
`30ignition`. While we're here, let's consistently prefix them by
`coreos-`. So overall, the changes are:
- `30ignition-coreos` -> `35coreos-ignition`
- `20live` -> `35coreos-live`
- `15coreos-network` -> `35coreos-network`

I've verified that there are now no "Failed to add dependency on unit"
error messages at build-time. (The next patch enforces this more
strictly.)

Fixes: #778
Closes: #799
(cherry picked from commit 94c435f03ea6f6a39ab46a3528f9029d051ebfdc)


jlebon:
Respin for https://github.com/coreos/fedora-coreos-config/issues/799.

dustymabe (approved):
LGTM

PR #808, opened 2021-01-07
coreos-teardown-initramfs: propagate NM configs in /etc

jlebon:
In the case where a user is PXE booting (or booting a custom made ISO as
suggested in https://github.com/coreos/fedora-coreos-tracker/issues/661)
we want users to be able to specify their NM configurations using a
secondary initrd which overlays the config in `/etc/NetworkManager/` as
a better alternative to using networking kargs.

In that case, we do still want to propagate those configs into the real
root. This avoids users having to also specify the same config again via
Ignition.

Related: https://discussion.fedoraproject.org/t/25833
Related: https://github.com/coreos/fedora-coreos-tracker/issues/661


jlebon:
(Not tested yet.)

jlebon:
This is obsolete now that we have first class support for NM configs.

PR #812, opened 2021-01-08
Revert "Switch to sqlite rpmdb backend"

jlebon:
This reverts commit 51cd4932c7346e9f266ce6940044c7c7e8f80195.

It seems like it breaks some rpm-ostree pkglayering tests, which means
it might be subtly changing layering semantics. Let's revert for now
while we investigate to make sure it doesn't somehow get into a release.

See: https://github.com/coreos/rpm-ostree/issues/2435

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -40,9 +40,6 @@ check-groups:
 
 default-target: multi-user.target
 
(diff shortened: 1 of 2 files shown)


cgwalters:
OK I think we have a solution in https://github.com/coreos/rpm-ostree/pull/2436 - let's let that try out and avoid the double revert churn.

PR #815, opened 2021-01-13
Multiple multipath patches

jlebon:

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
@@ -18,7 +18,12 @@ karg() {
 # are run in a systemd unit with MountFlags=slave so it is unmounted for us.
 bootmnt=/mnt/boot_partition
 mkdir -p ${bootmnt}
-bootdev=/dev/disk/by-label/boot
+
+bootdev=/dev/disk/by-label/dm-mpath-boot
+if [ ! -b "${bootdev}" ]; then
+    bootdev=/dev/disk/by-label/boot
+fi
+
 mount -o rw ${bootdev} ${bootmnt}
 
 # Clean up firstboot networking config files if the user copied them into the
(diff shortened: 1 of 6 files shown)


Prashanth684:
@jlebon i see a similar error with these changes when i add `rd.multipath=default` alone (without `root=`) but for the root partition:
```
[    9.847880] mount[702]: mount: /sysroot: /dev/sda4 already mounted or mount point busy.
```

should the same change in `coreos-boot-edit` be done for the root partition as well ?

jlebon:
OK, this is no longer required actually given https://github.com/openshift/os/pull/484#issuecomment-760454177. Will split out some patches from here which I think are still valuable.

@Prashanth684 Will chat with you about the failures you're seeing.

Prashanth684:
i was actually playing with multipath support on first boot and with this patch and https://github.com/openshift/os/pull/484, it actually works , but I had to make a change to the coreos-installer to persist the `rd.multipath` and `root` kargs. for testing i hacked it by putting it here: https://github.com/coreos/coreos-installer/blob/master/scripts/coreos-installer-service#L14, but that's not the right place though. 

I'm wondering even if we should get this PR checked in any case (of course not to support OCP 4.7 but for future releases) because it helps with provisioning multipath on first boot if the requirement comes in the future. I can also open a GH issue for appending the multipath kargs to the installer if necessary.

jlebon:
> I'm wondering even if we should get this PR checked in any case (of course not to support OCP 4.7 but for future releases) because it helps with provisioning multipath on first boot if the requirement comes in the future.

I think I prefer to just leave it as is for now and narrow the multipath gap further when we get kargs-via-Ignition as described [here](https://github.com/openshift/os/pull/484#issuecomment-760501588). This matches the model of root reconfiguration via Ignition (e.g. like full-disk mirroring and LUKS is done now).

Prashanth684:
> > I'm wondering even if we should get this PR checked in any case (of course not to support OCP 4.7 but for future releases) because it helps with provisioning multipath on first boot if the requirement comes in the future.
> 
> I think I prefer to just leave it as is for now and narrow the multipath gap further when we get kargs-via-Ignition as described [here](https://github.com/openshift/os/pull/484#issuecomment-760501588). This matches the model of root reconfiguration via Ignition (e.g. like full-disk mirroring and LUKS is done now).

ah..i see..ok.. thanks @jlebon !

PR #816, opened 2021-01-14
coreos-growpart: drop support for growing multipathed disks

jlebon:
The model we're moving towards is one where multipath enablement is
configured via Ignition and so only takes effect on the next boot:

https://github.com/openshift/os/pull/484#issuecomment-760454177

Therefore, we don't need to worry here about support for growing a
multipathed root partition. This simplifies a big chunk of the script.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
@@ -45,43 +45,23 @@ case "${TYPE:-}" in
     *) echo "error: Unsupported filesystem for ${path}: '${TYPE:-}'" 1>&2; exit 1 ;;
 esac
 
-if [[ "${src}" =~ "/dev/mapper" ]]; then
-    eval $(udevadm info --query property --export "${src}")
-    # get the partition, if any, and the name for the device mapper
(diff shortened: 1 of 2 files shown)


jlebon:
(Not tested, but pretty straightforward.)
Diff best viewed with whitespace ignored: https://github.com/coreos/fedora-coreos-config/pull/816/files?w=1

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-growpart
```diff
@@ -45,43 +45,23 @@ case "${TYPE:-}" in
     *) echo "error: Unsupported filesystem for ${path}: '${TYPE:-}'" 1>&2; exit 1 ;;
 esac
 
-if [[ "${src}" =~ "/dev/mapper" ]]; then
```
jlebon, line 48: This path was actually already dead since https://github.com/coreos/fedora-coreos-config/commit/604f9eb1a1a7434ac0a3ec6420b4e5d849998075 (#503) because if the root partition was modified/moved, we now exit early at the top of the script.

jlebon:
This should be an easy one to review!

arithx (approved):
Makes sense to me. Sorry for a late review, somehow this got caught in a flurry of notifications and I missed it.

PR #820, opened 2021-01-18
[WIP]Test podman-3.0.1rc1

baude:
Do not merge this. It is only for a sniff test with newest podman
release.

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/manifest-lock.overrides.x86_64.yaml
+++ b/manifest-lock.overrides.x86_64.yaml
@@ -19,3 +19,7 @@ packages:
   # https://bodhi.fedoraproject.org/updates/FEDORA-2021-2c1659799a
   ignition:
     evra: 2.9.0-2.git1d56dc8.fc33.x86_64
+  podman:
+    evra: 2:3.0.0-0.1.rc1.fc33.x86_64
+  podman-plugins:
+    evra: 2:3.0.0-0.1.rc1.fc33.x86_64


manifest-lock.overrides.x86_64.yaml
```diff
@@ -19,3 +19,7 @@ packages:
   # https://bodhi.fedoraproject.org/updates/FEDORA-2021-2c1659799a
   ignition:
     evra: 2.9.0-2.git1d56dc8.fc33.x86_64
+  podman:
+    evra: 3.0.0-0.1.rc1.fc33.x86_64
```
lsm5: hmm, can you try 2:3.0.0-0.1.rc1.fc33.x86_64 ?

manifest-lock.overrides.x86_64.yaml
```diff
@@ -19,3 +19,7 @@ packages:
...
+  podman-plugins:
+    evra: 3.0.0-0.1.rc1.fc33.x86_64
```
lsm5: same as above ^

lsm5:
@baude btw, if it matters, the commit message might need to be changed, the version should be v3.0.0-rc1 not 3.0.1rc1. Or 3.0.0-0.1.rc1 if we're talking rpm version.

PR #822, opened 2021-01-20
40ignition-ostree: make transposefs-related errors hard

jlebon:
Noticed that `ignition-ostree-transposefs-save.service` didn't break the
boot immediately after an error while debugging logs with @mike-nguyen.
Let's cargo-cult the `OnFailure` bits into those services too.

I think we may still be suffering from
https://github.com/systemd/systemd/issues/14142 in RHEL systemd.


mike-nguyen (approved):
LGTM

cgwalters (approved):
/lgtm

PR #823, opened 2021-01-24
overlay: Disable systemd-firstboot even if installed

cgwalters:
I'm trying to hack on systemd using `cosa build-fast`, which
doesn't run through the `remove-from-packages` path which we
use to disable systemd-firstboot.

Let's also add this cheap mechanism to disable it even if it's
installed.


overlay.d/05core/usr/lib/systemd/system/systemd-firstboot.service.d/fcos-disable.conf
```diff
@@ -0,0 +1,3 @@
+# Disable systemd-firstboot becuase it conflicts with Ignition.
```
jlebon:
```suggestion
# Disable systemd-firstboot because it conflicts with Ignition.
```

Also maybe let's cross-reference this file and the `remove-from-packages` bits with each other?

travier (approved):

jlebon (approved):

PR #830, opened 2021-01-28
manifest: Enable cliwrap

### LINKED ISSUE #730: Discuss enabling rpm-ostree cliwrap
This is a pointer issue for https://github.com/coreos/fedora-coreos-config/pull/830 to raise visibility.  That PR proposes to replace the `dracut`, `grubby`, and `rpm` commands with wrappers that don't allow modifying the system.

I'm locking this issue; please discuss in the above-linked PR.

cgwalters:
This functionality has been around for quite a while in rpm-ostree
but I'm not aware of anyone using it.  I'd like to turn it
on here so we can gain the benefits, which right now are
better protection against non-ostree-aware commands run as
root (`rpm` and `dracut` namely).

A bit more in https://github.com/coreos/rpm-ostree/pull/1789

Closes https://github.com/coreos/fedora-coreos-tracker/issues/730.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -43,6 +43,9 @@ default-target: multi-user.target
 # we can drop this when it's the rpm-ostree default
 rpmdb: sqlite
 
+# See the treefile docs
+cliwrap: true
+
 # ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
 # See also the version of this in fedora-coreos.yaml
 postprocess:
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -123,6 +123,12 @@ if [ ! -f /usr/share/rpm/rpmdb.sqlite ]; then
 fi
 ok rpmdb is sqlite
 
+if dracut 2>err.txt; then
+    fatal "ran dracut"
+    grep -q 'rpm-ostree initramfs' err.txt
+fi
+echo "ok rpm-ostree cliwrap"
+
 # make sure we don't default to having swap on zram
 # https://github.com/coreos/fedora-coreos-tracker/issues/509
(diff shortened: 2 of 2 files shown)


jlebon:
This probably deserves higher visibility. Let's discuss it in Wednesday's community meeting.

bgilbert:
Filed meeting ticket at https://github.com/coreos/fedora-coreos-tracker/issues/730.

jlebon:
This was discussed in [the community meeting](https://meetbot.fedoraproject.org/teams/fedora_coreos_meeting/fedora_coreos_meeting.2021-02-17-16.30.html):

> AGREED: We will turn on cliwrap in the FCOS `next` stream to gain feedback. We will document it, and send an email to coreos-status. We will file an f35 Change before turning it on on all streams (as well as possibly other rpm-ostree variants like FSB and IoT).

cgwalters:
OK rebased :surfer: on next-devel per discussion.

jlebon:
So because of [how the branches are setup](https://github.com/coreos/fedora-coreos-config#about-this-repo), to only turn this on for `next-devel` and `next` requires is a bit awkward. Basically something like this:
- PR to `next-devel` to turn on `cliwrap: true` in `manifest.yaml`.
- PR to `next` to turn on `cliwrap: true` in `manifest.yaml`.
- PR to `testing-devel` for the test, which would be conditional on the stream being `next*`.

cgwalters:
Ahh right OK.   And we should e.g. `rpm-ostree status -b | grep -q next-devel` in the test to find the stream?  I'll look at this soon I hope.

jlebon:
> Ahh right OK. And we should e.g. `rpm-ostree status -b | grep -q next-devel` in the test to find the stream? I'll look at this soon I hope.

Yeah, exactly (maybe just `next` instead of `next-devel` for extra greppiness).

cgwalters:
OK I thought about this more, and I think a better way to go about getting small-scale real world testing for cliwrap is to make it easier to enable client side.  We can circle back to enabling it by default later.

PR #843, opened 2021-02-05
manifest: Add back `dracut-squash` and `squashfs-tools`

kelvinfan001:
`dracut-squash` and `squashfs-tools` were made weak dependencies
by `kexec-tools` in version `2.0.21-4.fc33.x86_64`. However,
without these two packages, kdump's generated initramfs will be
unable to boot.
https://bugzilla.redhat.com/show_bug.cgi?id=1925585

Add back these packages for now so kdump functionality is not
broken.

### DIFF
--- a/manifests/user-experience.yaml
+++ b/manifests/user-experience.yaml
@@ -24,7 +24,10 @@ packages:
   - console-login-helper-messages-profile
   # kdump support
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
+  ## dracut-squash - weak dependency of kexec-tools
+  ## https://github.com/coreos/fedora-coreos-tracker/issues/798
   - kexec-tools
+  - dracut-squash
   # Remote Access
   - openssh-clients openssh-server
   # Container tooling
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -237,6 +237,9 @@
     "dracut-network": {
       "evra": "050-64.git20200529.fc33.x86_64"
     },
+    "dracut-squash": {
+      "evra": "050-64.git20200529.fc33.x86_64"
+    },
     "e2fsprogs": {
       "evra": "1.45.6-4.fc33.x86_64"
     },
@@ -1100,6 +1103,9 @@
     },
     "sqlite-libs": {
       "evra": "3.34.1-1.fc33.x86_64"
+    },
+    "squashfs-tools": {
+      "evra": "4.4-2.20200513gitc570c61.fc33.x86_64"
     },
     "ssh-key-dir": {
       "evra": "0.1.2-5.fc33.x86_64"


kelvinfan001:
@jlebon I'm not sure if this PR is necessary since kdump is actually already broken on FCOS after we mounted `/boot` RO.
So right now, kdump on FCOS does not work due to two reasons:
- RO `/boot` leading to kdump initramfs generation failure
  - BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1918493
  - proposed patch: https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/thread/G3CXOZQGETNQF33BB7ZO56YHZ72NFJTI/
  - PR version of the same patch: https://src.fedoraproject.org/rpms/kexec-tools/pull-request/6
- FCOS missing `squashfs-tools` and `dracut-squash` leading to kdump's generated initramfs not being able to start
  - BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1925585

jlebon:
Hmm yeah, let's hold it for now while the read-only `/boot` gets sorted out anyway. My understanding is that the use of squashfs should be opportunistic, so let's see what the maintainer says in that RHBZ.

travier:
Is there any particular reason not to add an explicit dependency to `libzstd` instead of pulling back in `squashfs-tools` & `dracut-squash` module? Also, do you have the full boot log here as we don't know which binary is missing the libzstd dep.

kelvinfan001:
> Is there any particular reason not to add an explicit dependency to `libzstd` 

It was meant to be a quick workaround, ideally the `kdump` dracut module should pull in `libzstd` right (even when `dracut-squash` isn't present)?

> do you have the full boot log

Boot log: https://paste.ubuntu.com/p/KbcY4jJ33C/

travier:
> > Is there any particular reason not to add an explicit dependency to `libzstd`
> 
> It was meant to be a quick workaround, ideally the `kdump` dracut module should pull in `libzstd` right (even when `dracut-squash` isn't present)?

Yes, it looks like the initramfs `init` itself needs libzstd so another core dracut module should probably install it.

travier:
So `/init` is directly systemd so we should probably figure out why the systemd or systemd-initrd dracut modules do not pull in the libzstd library.

kelvinfan001:
I noticed that FCOS already has `libzstd`, so we just need some dracut module to install it in the initramfs. We get the `/init: error while loading shared libraries: libzstd.so.1: cannot open shared object file: No such file or directory` error, but I wonder why `/init` doesn't already pull in `libzstd`, since it's working fine in our own (unmodified by kdump) initramfs.
My hypothesis is that `/init` itself doesn't need `libzstd`, and perhaps kdump has some hook that is executed by `/init` that requires `libzstd`; however, I'm having some trouble finding where kdump requires `libzstd`. 
So kdump works fine when it pulls in the `dracut-squash` module (which has a package dependency on `libzstd`), but I also cannot find where `dracut-squash` installs `libzstd`. 

`99kdumpbase` dracut module (included in kdump-generated initramfs)'s `module_setup.sh`: https://src.fedoraproject.org/rpms/kexec-tools/blob/f33/f/dracut-module-setup.sh#_43

`99squash` dracut module (kdump initramfs boots normally if this module is included in it)'s `module_setup.sh`:
```
$ cat /usr/lib/dracut/modules.d/99squash/module-setup.sh 

#!/bin/bash

check() {
    return 255
}

depends() {
    echo "bash systemd systemd-initrd"
    return 0
}

installkernel() {
    hostonly="" instmods -c squashfs loop overlay
}

install() {
    if ! type -P mksquashfs >/dev/null || ! type -P unsquashfs >/dev/null ; then
        derror "squash module requires squashfs-tools to be installed."
        return 1
    fi

    inst_multiple kmod modprobe mount mkdir ln echo
    inst $moddir/setup-squash.sh /squash/setup-squash.sh
    inst $moddir/clear-squash.sh /squash/clear-squash.sh
    inst $moddir/init.sh /squash/init.sh

    inst "$moddir/squash-mnt-clear.service" "$systemdsystemunitdir/squash-mnt-clear.service"
    systemctl -q --root "$initdir" add-wants initrd-switch-root.target squash-mnt-clear.service
}
```

Edit: interestingly, when I do `lsinitrd` on the initramfs generated by kdump, I do see `libzstd.so.1` inside.
```
$ sudo lsinitrd /var/lib/kdump/initramfs-5.10.19-200.fc33.x86_64kdump.img | grep libzstd
lrwxrwxrwx   1 root     root           16 Jan  1  1970 usr/lib64/libzstd.so.1 -> libzstd.so.1.4.7
-rwxr-xr-x   1 root     root       873216 Jan  1  1970 usr/lib64/libzstd.so.1.4.7
```

travier:
You're definitely making progress! This is a really strange issue. Can you compare the full lsinitrd output on initramfs built with and without the squashfs dracut module? We might be able to pick-up what's missing from that.

In the past, I've encountered cases where we needed to manually install dependent libraries (https://github.com/clipos/src_portage_clipos/blob/master/sys-kernel/dracut-clipos-modules/files/11clipos-core-state/module-setup.sh#L27) but that appears to not be the issue as you've noticed.

kelvinfan001:
Thanks for the tip, @travier! 
kdump-generated initramfs with `99squash` dracut module included puts most files in `squashfs-root`, including `libzstd`, so the initramfs with and without `99squash` are quite different:
```
$ sudo lsinitrd /var/lib/kdump/initramfs-5.10.19-200.fc33.x86_64kdump.img | grep libzstd
lrwxrwxrwx root/root                16 2021-03-08 15:28 squashfs-root/usr/lib64/libzstd.so.1 -> libzstd.so.1.4.7
-rwxr-xr-x root/root            873216 1970-01-01 00:00 squashfs-root/usr/lib64/libzstd.so.1.4.7
```
I am still not sure why `/init` cannot find `libzstd` when the initramfs files are not stored in a squashfs image. I suspect `99kdumpbase`'s (modified?) `/init` expects to find e.g. `libzstd` in a squashfs image, and cannot find it if it is in a regular location.

kelvinfan001:
I tested on Fedora Server and the kdump-generated initramfs seems to boot up fine without the `dracut-squash` package installed (I also did a `lsinitrd` on the generated initramfs and confirmed that it is not squashed), so it looks like this might only be a problem on FCOS :(

travier:
An option here would be to generate several initrds, skipping one module each time to figure out which module is responsible for this issue on FCOS.

kelvinfan001:
https://github.com/coreos/fedora-coreos-config/pull/936 adds a drop in to get `kdump.service` to remount `/boot` rw before `kdump` generates its initramfs, so after that is merged, the only cause of `kdump` being broken would be the lack of `squashfs-tools`.

https://github.com/coreos/fedora-coreos-config/pull/936 is currently a workaround for now until https://bugzilla.redhat.com/show_bug.cgi?id=1918493 is addressed properly.

travier:
OK, it's been broken for a while and we have not had the time to figure out what the root issue is so we should probably do this one (although it looks like it is not adding the packages to a list in a manifest) and open a tracking issue for the revert + investigation for tracking.

manifests/user-experience.yaml
```diff
@@ -25,6 +25,7 @@ packages:
   # kdump support
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
   - kexec-tools
+  - dracut-squash
```
travier, line 30: Can you add a link to the tracking issue here?

kelvinfan001: yep! sorry about that. added now

manifests/user-experience.yaml
```diff
@@ -24,7 +24,10 @@ packages:
   - console-login-helper-messages-profile
   # kdump support
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
+  ## dracut-squash - weak dependency of kexec-tools
+  ## https://github.com/coreos/fedora-coreos-tracker/issues/798
```
jlebon, line 28: Should this be right above `dracut-squash` instead?

jlebon (approved):
I'm also OK getting this in for now while we debug the root issue.

jlebon:
Actually, see https://github.com/coreos/fedora-coreos-tracker/issues/798#issuecomment-823398460.

kelvinfan001:
Wow, thanks! agh, I don't know why I didn't think of trying with more reserved memory. So it looks like the issue is caused by the reserved memory not having enough space to even store all the modules that the kdump initramfs needs.

kelvinfan001:
PR to update docs to reserve more memory for crash kernel https://github.com/coreos/fedora-coreos-docs/pull/276.

PR #845, opened 2021-02-08
overlay.d/20platform-chrony: Exit early if already run

travier:
overlay.d/20platform-chrony: Minor whitespace fix

---

overlay.d/20platform-chrony: Exit early if already run

Make sure we exit early if:
  * the config is not the default one and as been modified
  * this script already generated a configuration for chrony

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1924869


travier:
Untested.
/hold

travier:
Tested and confirmed as working. You can reproduce this issue by running `systemctl daemon-reload` on an instance in AWS/GCP/Azure (tested only AWS).

overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony
```diff
@@ -26,6 +26,18 @@ case "${platform}" in
     *) exit 0 ;;
 esac
 
+# Exit early if we have already been run once
+if [[ -f "${confpath}" ]]; then
+    echo "$self: ${confpath} already exists; skipping"
+    exit 0
+fi
+
+# Exit early if chrony configuration as been changed from the image default
```
jlebon, line 35: Hmm, this seems to be changing behaviour that is somewhat separate from what the RHBZ is about. Moving it up means that we no longer auto-inject `PEERNTP=no` even if the user customized `chrony.conf` whereas before we still would. So doesn't this break users who relied on that?

travier: True. I though it would make more sense to not touch the config if it had been modified but changing this might create a regression.

jlebon: I'm not sure if that was on purpose or not. Though reading the code, at least it's idempotent.

miabbott:
This makes sense to me; I think the `PEERNTP` behavior that Jonathan pointed out is correct in this modified form.  We can release note this change as part of the release it lands in.

/approve

jlebon (approved):
Yup, fine with me as is too. I agree conceptually it's nicer if our automatic configuration is either fully enabled or disabled.

cgwalters (approved):

travier:
So all three releases failed on this issue this week so maybe I missed something here while testing it.

travier:
I just tried that fix on a previous testing image and it fixed the issue when running `systemctl daemon-reload`. So I don't know where the errors in the kola tests from the release are coming from.

travier:
This failed on update tests only in kola which is expected as the old image does not have the fix. This fix should thus be good.

miabbott:
> This failed on update tests only in kola which is expected as the old image does not have the fix. This fix should thus be good.

sweating

PR #848, opened 2021-02-11
[WIP] overlay: generate and propagate udev rules for s390x znet devices

nikita-dubrovskii:
Fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1818033

When varying off and on the chpid to simulate an outage, the network does not come up automatically afterwards.
The guest has to be restarted. Adding missing udev rules so device could be added back withot system restart.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+# Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1818033
+
+znet=$(lszdev qeth -c ID --no-headings | awk -F ":" '{print $1}')
+for dev in ${znet}; do
+    # chzdev creates udev rules to store the persistent configuration of devices in this directory. File names start with "41-".
+    rule="/sysroot/etc/udev/rules.d/41-qeth-${dev}.rules"
+    if [[ ! -f ${rule} ]]; then
+	echo "${dev}: generating ${rule}"
+	chzdev qeth -e ${dev} --no-root-update --base /etc=/sysroot/etc
+    fi
+done
--- a/overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=IBM znet udev-rules generator
+ConditionArchitecture=s390x
+ConditionPathExists=/etc/initrd-release
+DefaultDependencies=false
+Before=ignition-complete.target
+
+After=ostree-prepare-root.service
+Before=ignition-files.service
+
+ConditionKernelCommandLine=rd.znet
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/ibm-znet-rules
+RemainAfterExit=yes
--- a/overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/module-setup.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+depends() {
+       echo systemd
+}
+
+install_and_enable_unit() {
+       local unit="$1"; shift
+       local target="$1"; shift
+       inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
+       # note we `|| exit 1` here so we error out if e.g. the units are missing
+       # see https://github.com/coreos/fedora-coreos-config/issues/799
+       systemctl -q --root="$initdir" add-requires "$target" "$unit" || exit 1
+}
+
+install() {
+    inst_multiple chzdev lszdev awk
+
+    install_and_enable_unit "ibm-znet-rules.service" \
+                            "ignition-complete.target"
+
+    inst_script "$moddir/ibm-znet-rules.sh" \
+                "/usr/sbin/ibm-znet-rules"
+}


overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1818033
+
+zdevices=$(lszdev qeth -c ID --no-headings)
+for dev in ${zdevices}; do
+    echo "Generating udev rules for ${dev}"
+    chzdev -e ${dev} --no-root-update
+done
+
+# chzdev creates udev rules to store the persistent configuration of devices in this directory. File names start with "41-".
```
hbrueckner: When the udev files are present on the target system, do not forget to remove the rd.znet kernel parameter .

overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+
+# Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1818033
+
+zdevices=$(lszdev qeth -c ID --no-headings)
+for dev in ${zdevices}; do
+    echo "Generating udev rules for ${dev}"
+    chzdev -e ${dev} --no-root-update
```
hbrueckner: @oberpar Do we have a mean to create those udev directory in /sysroot/etc/udev/rules.d/... otherwise it could picked the wrong ones.

nikita-dubrovskii:
After some discussions in slack agreed on keeping this PR opened until we have a full picture of what needs to be done.

overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.sh
oberpar: You can use `chzdev`'s `--base` option to change the base directory for all file operations, or only for specific ones.

Example: This invocation generates udev rules for all existing qeth devices in /sysroot/etc:
`chzdev qeth --existing -ep --base /etc=/sysroot/etc`

jlebon:
Are the udev rules dynamic? If not, it seems like they should be shipped as part of the OSTree (by being part of the appropriate package). If yes, then that package should probably also ship the code needed to dynamically generate them on boot.

nikita-dubrovskii:
> Are the udev rules dynamic? If not, it seems like they should be shipped as part of the OSTree (by being part of the appropriate package). If yes, then that package should probably also ship the code needed to dynamically generate them on boot.

They are dynamic, so for each `rd.znet=xyz12345` we may create udev rule (`41-qeth-xyz12345.rules`) , but than it's better to remove corresponding `rd.znet` karg option, because `rd.znet` and `udev` could be conflicting. Yesterday we discussed that and found 3 possible solutions:

- create `Operator` for handling `attach/detach` of `s390x` devices (long term work)
- update this PR (and make it similar to `ignition-firstboot-complete.service`) so it can generate udev rules and update kernel args (+ ramdisk)
- move this functionality to `ignition`
- patch `coreos-installer` so it generates udev rules and skips all `rd.znet` during installation of `zipl` 

Would appreciate any suggestion from your side about written above.

Anyhow discussions are going on, and this PR is like a temporary workaround and place for further discussion.

overlay.d/05core/usr/lib/dracut/modules.d/99ibm-znet-rules/ibm-znet-rules.sh
hbrueckner: Thanks

lucab:
I'm not familiar with this s390x topic, so I may have misread something, but to my understanding there are two topics here to handle:
 1. udev rules generation in initramfs: this seems to be some custom translation logic for `rd.znet` kargs, which I think should be living [here](https://github.com/dracutdevs/dracut/tree/051/modules.d/95znet) in dracut and generate ephemeral rules under /run
 2. rules propagation from initramfs to rootfs: this one I think should be living here, but it should probably only happen once at first-boot (similar to the NM-config propagation logic we have). It should happen earlier than Ignition `files` stage, so that it is possibly to override this from user config.

>  but than it's better to remove corresponding rd.znet karg option, because rd.znet and udev could be conflicting.

I'd rather not go this route, but instead perform the propagation only on first-boot.
If something else comes up after provisioning and changes the udev rules OR the kargs, it should take care of changing the other half too.

nikita-dubrovskii:
@lucab 
>     1. udev rules generation in initramfs: this seems to be some custom translation logic for `rd.znet` kargs, which I think should be living [here](https://github.com/dracutdevs/dracut/tree/051/modules.d/95znet) in dracut and generate ephemeral rules under /run

As i know, SUSE & Ubuntu use other approaches for `z-devices + udev` , so i'm not sure about adding this fix to `dracut`.
@hbrueckner wdyt?

>     2. rules propagation from initramfs to rootfs: this one I think should be living here, but it should probably only happen once at first-boot (similar to the NM-config propagation logic we have). It should happen earlier than Ignition `files` stage, so that it is possibly to override this from user config.

done

hbrueckner:
> > ```
> > 1. udev rules generation in initramfs: this seems to be some custom translation logic for `rd.znet` kargs, which I think should be living [here](https://github.com/dracutdevs/dracut/tree/051/modules.d/95znet) in dracut and generate ephemeral rules under /run
> > ```
> 
> As i know, SUSE & Ubuntu use other approaches for `z-devices + udev` , so i'm not sure about adding this fix to `dracut`.
> @hbrueckner wdyt?
@lucab @nikita-dubrovskii 

The rd.znet writes the 3 ccw channels to a configuration file to become automatically enabled at boot. Typically, this is specified for the installer/anaconda on fedora/rhel to get the networking up and running and apply those to the final NM configuration to set up the network. The new change is on the re-naming the network interface. rd.znet came originally with dracut while lszdev/chzdev has been added later to improve Z device handling. However, chzdev can be used but the system itself still uses the /etc/ccw.conf and /etc/zfcp.conf for device enablement. NM is able to automatically group the znet CCW channels to create and configure the network device.

@oberpar Does chzdev provides an attribute to change the name of the network interface?

oberpar:
> @oberpar Does chzdev provides an attribute to change the name of the network interface?

No. `chzdev` only modifies Z-specific device attributes, in this case related to a CCWGROUP device for QETH.

julianwiedmann:
From looking at the dracut initramfs and RHCOS-proper infrastructure, I feel we would be better suited (and in harmony with RHEL) by just installing an ifcfg file (or NM keyfile). If I see things correctly then `/usr/lib/dracut/hooks/pre-pivot/85-write-ifcfg.sh` should create a full ifcfg for the rd.znet interface (with all the s390 options), and the combination of `/usr/lib/udev/rules.d/81-ccw.rules` and `/usr/lib/udev/ccw_init` should then be able to consume this when the qeth ccw devices get hotplugged later.

So let me play around with that a bit more, and see what happens if that ifcfg file is in place ...

julianwiedmann:
For the record, if we go with a NM keyfile (instead of ifcfg) then we would also need to pick up the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1885913

hbrueckner:
> For the record, if we go with a NM keyfile (instead of ifcfg) then we would also need to pick up the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1885913

So if anaconda moved to create keyfiles instead of ifcfg-style files, it would be more consistent to follow this path as well.
@sharkcz Probably need a backport of ccw_init from above bugfix for rhel-8 to be picked up?
Further, I wonder whether is an option to move to chzdev. @oberpar / @sharkcz Where do we stand with its integration to replace ccw_init?

julianwiedmann:
> So let me play around with that a bit more, and see what happens if that ifcfg file is in place ...

As expected, adding the ifcfg file is sufficient to get the qeth device fully configured on hotplug (eg. CP DETACH & CP ATTACH).

julianwiedmann:
> > For the record, if we go with a NM keyfile (instead of ifcfg) then we would also need to pick up the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1885913
> 
> So if anaconda moved to create keyfiles instead of ifcfg-style files, it would be more consistent to follow this path as well.
> @sharkcz Probably need a backport of ccw_init from above bugfix for rhel-8 to be picked up?
> Further, I wonder whether is an option to move to chzdev. @oberpar / @sharkcz Where do we stand with its integration to replace ccw_init?

I would suggest to split the RHEL vs chzdev topic into a separate bugzilla... unless we want to block the immediate RHCOS fix on resolving this topic.

julianwiedmann:
Ok, finally chased down what I think is going wrong here.

What we **want** in RHCOS right now is for initrd to build a .nmconnection with the s390-subchannels and s390-options from the rd.znet parameters [1], and copy that over into real-root [2]. Then ccw_init can react to a hotplug of a qeth ccw_device, dig for the NM keyfile with the corresponding s390-subchannels [3], group the subchannels again and set the s390-options.

Unfortunately nm-initrd-generator has stopped to create connections from rd.znet options with [4], so initrd has no .nmconnection to copy over into real-root any longer...

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/11d4412ee1554358c629f44eb41b5e8cc3546870
[2] https://github.com/coreos/fedora-coreos-config/commit/ff09e91ca0cebbc3e9907682fd35195b1ccd1545
[3] https://fedorapeople.org/cgit/sharkcz/public_git/utils.git/commit/?id=6f264c2a4279fa9616f2450f09fcc93bd4c2b7c6
[4] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/3957d40f54167bcc49dd4147bfa8d6a159304677

sharkcz:
I agree the current situation isn't ideal, but I think there is a philosophical problem how to approach the network interface initialization/configuration on s390x. In the past we had everything in the `ifcfg` files, which is not possible any more when using NetworkManager with the keyfiles as the primary network configuration tool. I am adding CCing Thomas (@thom311), he is our NM guy. But there is an overlap to the initrd environment (dracut), to the regular installer/anaconda, so I think all those people should sit together and come up with a proposal.
I think there are 2 ways - keep configuration in a single place (NM keyfiles?) or split the s390x specific part into one place (zdev) and keep only the network info in the NM keyfile, but still make them work together.

sharkcz:
> Ok, finally chased down what I think is going wrong here.
> 
> What we **want** in RHCOS right now is for initrd to build a .nmconnection with the s390-subchannels and s390-options from the rd.znet parameters [1], and copy that over into real-root [2]. Then ccw_init can react to a hotplug of a qeth ccw_device, dig for the NM keyfile with the corresponding s390-subchannels [3], group the subchannels again and set the s390-options.
> 
> Unfortunately nm-initrd-generator has stopped to create connections from rd.znet options with [4], so initrd has no .nmconnection to copy over into real-root any longer...

IIRC there was a problem when the `ip=` option created one interface in NM with all the network configuration and `rd.znet=` a second one with only the s390 specific details in it. And it didn't work together as it was dependent on the order of the `ip` and `rd.znet` options on the parameter line.

thom311:
I find it very ugly to have udev read's NetworkManager configuration (be it keyfiles or ifcfg files). Which is also what initscripts' `/usr/lib/udev/rename_device` does. I wish, .link files or udev rules would be used instead.

With NetworkManager, there is an expectation that when you modify a connection profile and activate it, that the settings take effect. That is not working with a udev rule, which only runs when the interface gets created.

But I don't have a good answer, especially as this is how this used to work for decade(s). I apologize for my nonconstructive rambling.

I wonder, the udev rule is in particular useful when a user is not using NetworkManager. But when using NetworkManager, why can it not configure that interface? Must this all happen in udev? Why?

CC @bengal

julianwiedmann:
@sharkcz fully agreed. This goes way beyond RHCOS, and making the overall situation worse just to quick-fix znet hotplug on RHCOS is not a good idea. So getting a common understanding how this should all work in a NetworkManager world, and eg. what responsibilites the installer has (be it anaconda or {first-boot,ignition}) sounds like the right way forward.

(cc @dustymabe as well)

hbrueckner:
Hi @thom311,

> With NetworkManager, there is an expectation that when you modify a connection profile and activate it, that the settings take effect. That is not working with a udev rule, which only runs when the interface gets created.

On s390x, there is a common I/O layer that does not exist in that way on x86. Because there could be thousands of devices on s390x, only required devices are made available on Linux instance running on s390x. Those udev rules typically make the network device visible. This is like hotplugging. On x86, you typically have auto-probing of network devices... on s390, you have to specifically configure which you want to use (and to become auto-probed). For this specific device management there the ccw_init or chzdev commands available.

> But I don't have a good answer, especially as this is how this used to work for decade(s). I apologize for my nonconstructive rambling.
> 
> I wonder, the udev rule is in particular useful when a user is not using NetworkManager. But when using NetworkManager, why can it not configure that interface? Must this all happen in udev? Why?

Basically, as @sharkcz mentioned there are two options. From an NM perspective, what is the requirement on the network interface? Should it be present and NM does additional link and ipam? Or (in case of s390) should NM actually "enable" the device (thru the s390 common I/O layer) and then continue with link and ipam?

julianwiedmann:
> I wonder, the udev rule is in particular useful when a user is not using NetworkManager. But when using NetworkManager, why can it not configure that interface? Must this all happen in udev? Why?
> 
In short - the udev rule lets ccw_init **create** and configure the device, which only then provides the network interface. There's no register_netdev() without some preceding sysfs magic. And storing the device config parameters inside the ifcfg / keyfile was ... convenient?

cgwalters:
> As i know, SUSE & Ubuntu use other approaches for z-devices + udev , 

Was this answered?  What do these distributions do?  AFAIK all of them here have different legacy network config formats (all support NM now too).

> Because there could be thousands of devices on s390x, only required devices are made available on Linux instance running on s390x. Those udev rules typically make the network device visible. This is like hotplugging. 

I think this was explained to me in the past but I still find it confusing - it seems like the platform is providing a (subset?) of devices to an instance, then the instance needs to say "please turn on device X, Y, Z" via some out-of-band knowledge of which devices those should be?  But perhaps an analogy is more like an IaaS (e.g. OpenStack) and one could imagine this "turn on the device" request as a bit more like an API request from inside a running VM to provision and attach e.g. a [Cinder block device](https://wiki.openstack.org/wiki/Cinder)?  The instance being able to do this is a bit like providing cloud credentials to the instance, but the provisioning request is decoupled from the boot to allow things to be more runtime elastic?  

> @sharkcz fully agreed. This goes way beyond RHCOS, and making the overall situation worse just to quick-fix znet hotplug on RHCOS is not a good idea. So getting a common understanding how this should all work in a NetworkManager world, and eg. what responsibilites the installer has (be it anaconda or {first-boot,ignition}) sounds like the right way forward.

Yes...as far as I know RHEL9 will still support keyfiles but I hope we can try to do a cutover (like RHCOS has) to "NM first and keyfile legacy" at least, so plumbing through support here to natively support NM would be great.

nikita-dubrovskii:
Closing this as non relevant anymore

PR #850, opened 2021-02-11
overlay: add warning if rootfs < 8G

jlebon:
We've recently had our first case of a "trapped" rootfs running out of
space for upgrades:

https://github.com/coreos/fedora-coreos-tracker/issues/731

Until we actually implement stronger behaviour for this, let's
explicitly check for this case and emit a warning if we detect it. In
the future, we'll look at making this a hard error by default (with an
escape hatch).

For more information, see:

https://github.com/coreos/fedora-coreos-tracker/issues/586#issuecomment-777220000

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size
@@ -1,0 +1,37 @@
+#!/bin/bash
+set -euo pipefail
+
+# See also ignition-ostree-check-rootfs-size.service
+# https://github.com/coreos/fedora-coreos-tracker/issues/586#issuecomment-777220000
+
+srcdev=$(findmnt -nvr -o SOURCE /sysroot | tail -n1)
+size=$(lsblk --nodeps --noheadings --bytes -o SIZE "${srcdev}")
+
+MINIMUM_GB=8
+MINIMUM_BYTES=$((1024 * 1024 * 1024 * MINIMUM_GB))
+
+MOTD_DROPIN=/etc/motd.d/60-coreos-rootfs-size.motd
+
+YELLOW=$(echo -e '\033[0;33m')
+RESET=$(echo -e '\033[0m')
+
+if [ "${size}" -lt "${MINIMUM_BYTES}" ]; then
+    mkdir -p "/sysroot/$(dirname "${MOTD_DROPIN}")"
+    cat > "/sysroot/${MOTD_DROPIN}" <<EOF
+${YELLOW}
+############################################################################
+WARNING: The root filesystem is too small. It is strongly recommended to
+allocate at least ${MINIMUM_GB} GiB of space to allow for upgrades. From June 2021, this
+condition will trigger a failure in some cases. For more information, see:
+https://docs.fedoraproject.org/en-US/fedora-coreos/storage/
+
+You may delete this warning using:
+sudo rm ${MOTD_DROPIN}
+############################################################################
+${RESET}
+EOF
+
+    # And also write it on stdout for the journal and console
+    cat "/sysroot/${MOTD_DROPIN}"
+    coreos-relabel "${MOTD_DROPIN}"
+fi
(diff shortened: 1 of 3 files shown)


jlebon:
Screenshot from 2021-02-11 16-27-25

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size
```diff
@@ -0,0 +1,44 @@
...
+YELLOW=$(echo -e '\033[0;33m')
+RESET=$(echo -e '\033[0m')
+
+if [ "${size}" -lt "${MINIMUM_BYTES}" ]; then
+    mkdir -p "/sysroot/$(dirname "${MOTD_DROPIN}")"
+    cat > "/sysroot/${MOTD_DROPIN}" <<EOF
+${YELLOW}
+########################################################################
+WARNING: The root filesystem is too small. It is strongly recommended to
+allocate at least ${MINIMUM_GB}G of space to allow for upgrades. In the future, this
```
lucab: Minor: `GiB`.

travier (approved):
Not tested but LGTM.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size
```diff
@@ -0,0 +1,44 @@
...
+condition will trigger a failure by default. For more information, see:
+https://docs.fedoraproject.org/en-US/fedora-coreos/storage/
+
+You may delete this warning using:
+rm ${MOTD_DROPIN}
```
bgilbert: `sudo`

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+set -euo pipefail
+
+# See also ignition-ostree-check-rootfs-size.service
+# https://github.com/coreos/fedora-coreos-tracker/issues/586#issuecomment-777220000
+
+srcdev=$(findmnt -nvr -o SOURCE /sysroot | tail -n1)
+size=$(lsblk --nodeps --noheadings --bytes -o SIZE "${srcdev}")
```
bgilbert, line 8: I was imagining we'd only produce the warning if the root filesystem is constrained by another partition, not if the user just uses the default partition layout on a small virtual disk.  A warning might be okay in any case, but do we want to ship a small disk image and then (eventually) fail if the user doesn't resize it?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size
```diff
@@ -0,0 +1,44 @@
...
+https://docs.fedoraproject.org/en-US/fedora-coreos/storage/
+
+You may delete this warning using:
+rm ${MOTD_DROPIN}
+########################################################################
+${RESET}
+EOF
+
+    # And also write it on stdout for the journal and console
+    cat "/sysroot/${MOTD_DROPIN}"
```
bgilbert, line 35: Have you checked that the embedded ANSI escapes won't prevent viewing of the journal entries?  I think I've had problems with `\r` in the past.

jlebon: It seems to show up fine in the journal. It's in black, so it's possible journald is stripped them out.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size
jlebon: I'd rather still print a warning, but yeah I like your suggestion on only making it a hard error in the trapped partition case.

jlebon:
Updated!

jlebon:
Let's hold this until we discuss specific sizing recommendations in the next community meeting (https://github.com/coreos/fedora-coreos-tracker/issues/586#issuecomment-779467905).

jlebon:
Updated to use 8 GiB now as per https://github.com/coreos/fedora-coreos-tracker/issues/586#issuecomment-785254764. Dropped `hold` label!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-check-rootfs-size.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Ignition OSTree: Check Root Filesystem Size
```
cgwalters, line 2: Perhaps also
```
Documentation=https://docs.fedoraproject.org/en-US/fedora-coreos/storage/
```
?

cgwalters (approved):
LGTM

bgilbert:
LGTM

PR #852, opened 2021-02-12
manifest: ship specific sssd subpackages instead

jlebon:
The `sssd` subpackage is a meta-package which just pulls in the default
set of packages. This default set changed in Fedora 34, which started
pulling in `python3-sssdconfig`:

https://bugzilla.redhat.com/show_bug.cgi?id=1927907

From the maintainer feedback there, it seems like it'd be better anyway
to be more explicit about the specific backends we want to ship. So
let's do that here: break out `sssd` into the client and the backends
for AD, IPA, KRB5, and LDAP.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -159,7 +159,9 @@ packages:
   # https://github.com/coreos/fedora-coreos-tracker/issues/445
   - libsss_sudo
   # Extra runtime
-  - sssd shadow-utils
+  - shadow-utils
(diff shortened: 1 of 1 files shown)


bgilbert (approved):
Superficial LGTM.

jlebon:
As a sanity-check, you can see from the CI run that the only dropped RPM from this is `sssd`:

```json
[
  [
    "sssd",
    1,
    {
      "PreviousPackage": [
        "sssd",
        "2.3.1-4.fc33",
        "x86_64"
      ]
    }
  ]
]
```

PR #853, opened 2021-02-16
[stable] tree: promote changes from testing at 1e357b015026a7a3a5ce5215f205bd0…

jlebon:
…63c4a95c5


travier:
Can you confirm that the sudo override drops are safe?

jlebon:
> Can you confirm that the sudo override drops are safe?

Yes, it's safe. You can see in the base lockfile that sudo was upgraded to the same version that was in the override. You should also be able to see in the package diff in the CI logs that we're not downgrading sudo.

travier (approved):

PR #860, opened 2021-02-18
presets: disable systemd-oomd.service

jlebon:
It requires swap, which we don't turn on by default. (It also requires
cgroups v2, which we currently don't turn on by default but are aiming
to in the near future.)


jlebon:
This should fix the f34 builds (on the new `branched` stream).

bgilbert (approved):
LGTM

PR #862, opened 2021-02-23
tests: Add basic toolbox functional testing

### LINKED ISSUE #864: kola: Add toolbox tests
# Feature Request #

Add toolbox tests to make sure we do not release an FCOS image with a broken combination of toolbox & podman.

travier:
This will make sure that builds can at least pull the image, create a
toolbox and run a basic command inside it without error.

Closes: https://github.com/coreos/fedora-coreos-config/issues/864

### DIFF
--- a/tests/kola/toolbox/test.sh
+++ b/tests/kola/toolbox/test.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# Make sure that basic toolbox functionnality is working:
+# - Creating a toolbox
+# - Running a command in a toolbox
+# - Removing all toolbox containers
+#
+# Important note: Commands are run indirectly via calls to `machinectl shell`
+# to re-create the user environment needed for unprivileged podman
+# functionality. However, machinectl shell does not propagate the exit
+# code/status of the invoked shell process thus we need additionnal checks to
+# ensure that previous commands were successful.
+
+# Only run on QEMU to reduce CI costs as nothing is platform specific here.
+# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv" }
+
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+machinectl shell core@ /bin/toolbox create --assumeyes 1>/dev/null
+if [[ $(machinectl shell core@ /bin/toolbox list --containers | grep --count fedora-toolbox-) -ne 1 ]]; then
+	fatal "Could not create toolbox"
+fi
+ok toolbox create
+
+machinectl shell core@ /bin/toolbox run touch ok_toolbox
+if [[ ! -f '/home/core/ok_toolbox' ]]; then
+	fatal "Could not run a simple command inside a toolbox"
+fi
+ok toolbox run
+
+toolbox="$(machinectl shell core@ /bin/toolbox list --containers | grep fedora-toolbox- | awk '{print $2}')"
+machinectl shell core@ /bin/podman stop "${toolbox}"
+machinectl shell core@ /bin/toolbox rm "${toolbox}"
+if [[ -n "$(machinectl shell core@ /bin/toolbox list --containers)" ]]; then
+	fatal "Could not remove the toolbox container"
+fi
+ok toolbox rm


travier:
This will fail in CI as toolbox support is currently broken in testing: https://github.com/coreos/fedora-coreos-tracker/issues/734

tests/kola/toolbox/test.sh
```diff
@@ -0,0 +1,30 @@
...
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if ! machinectl shell core@ /bin/toolbox create --assumeyes 1>/dev/null; then
```
cgwalters: I hadn't seen `machinectl shell core@` before, interesting.

cgwalters (approved):

tests/kola/toolbox/test.sh
```diff
@@ -0,0 +1,30 @@
...
+	fatal "Could not create toolbox"
+fi
+
+if ! machinectl shell core@ /bin/toolbox run touch OK 1>/dev/null; then
```
cgwalters: I think a bit longer term what would work well is to have something like `/bin/toolbox --selftest` that just runs some basic sanity tests baked into the binary.

cgwalters: Among other issues, I think a lot of the value in toolbox is testing the actual interactive (i.e. pseudoterminal) flow and you don't quite get that without jumping through some hoops to set one up.

tests/kola/toolbox/test.sh
travier: That's the easiest way I found to get a real user session up and running

tests/kola/toolbox/test.sh
martymichal: > I think a bit longer term what would work well is to have something like `/bin/toolbox --selftest` that just runs some basic sanity tests baked into the binary.

To be frank, I'm not exactly thrilled about the idea...

cgwalters: Can you elaborate why?

cgwalters: I've been struggling with the "run upstream tests as part of FCOS testing" problem for a long time, and while I just made up this `--selftest` idea in the context of toolbox, the more I think about it the more I like it for at least ostree/rpm-ostree.  And probably `podman` and perhaps `systemd/NetworkManager` etc. too.

Basically here the idea is that the (actually let's call it `--destructive-selftest`) option runs in something like our default kola exttests setup which is in turn inspired by the Debian autopkgtest destructive setup (xref [ostree blog](https://blog.verbum.org/2020/12/01/committed-to-the-integrity-of-your-root-filesystem/)).

- Fresh VM allocated for solely for this, test can be destructive and runs as root
- Has APIs to reboot the system
- Is fully distribution/OS and framework agnostic - should also make it easier for e.g. a kernel CI system to also run `podman --destructive-selftest`

cgwalters: Now obviously it's not going to scale really far to bake in *lots* of testing logic into a lot of system binaries, and it will bloat them a bit.  But, I think in many cases the additional bloat would be quite small - the goal here is mostly "basic sanity checking", leaving exotic/unusual things to separate CI systems - `foo-tests` subpackage type approaches (as Debian does), or the current Fedora dist-git approach etc.

travier: This is interesting and I like this but we might need to figure out a way to run / ship all those tests as separated RPMs that we could overlay just for testing maybe? Sounds like we need a separated design issue just for this topic?

cgwalters: OK some IRC discussion with @lucab I think is concluding this needs some sort of mini-specification - it's really an extension/reframing of the [Debian autopkgtest](https://wiki.debian.org/ContinuousIntegration/autopkgtest) spec I think.

In particular I think the `needs-internet` flag is really, really useful at scale to separate "tests that can flake because e.g. docker.io rate limited me" versus truly self-contained/isolated tests.

So the spec might be more like `$project --output-autopkgtests` and then the framework executes them - i.e. it could output multiple tests and use the metadata keys from the autopkgtest.

cgwalters:
Test is failing

```
[2021-03-04T20:32:02.467Z] Mar 04 20:31:36 qemu0 systemd[1]: Started kola-runext.service.

[2021-03-04T20:32:02.467Z] Mar 04 20:31:36 qemu0 kola-runext-test.sh[1209]: + machinectl shell core@ /bin/toolbox create --assumeyes

[2021-03-04T20:32:02.467Z] Mar 04 20:31:36 qemu0 kola-runext-test.sh[1210]: Connected to the local host. Press ^] three times within 1s to exit session.

[2021-03-04T20:32:02.467Z] Mar 04 20:31:59 qemu0 kola-runext-test.sh[1210]: Connection to the local host terminated.

[2021-03-04T20:32:02.467Z] Mar 04 20:31:59 qemu0 kola-runext-test.sh[1209]: + machinectl shell core@ /bin/toolbox run touch OK

[2021-03-04T20:32:02.467Z] Mar 04 20:31:59 qemu0 kola-runext-test.sh[1389]: Connected to the local host. Press ^] three times within 1s to exit session.

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 kola-runext-test.sh[1389]: Connection to the local host terminated.

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 kola-runext-test.sh[1209]: + [[ ! -f OK ]]

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 kola-runext-test.sh[1209]: + fatal 'Could not run a simple command inside a toolbox'

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 kola-runext-test.sh[1209]: + echo 'Could not run a simple command inside a toolbox'

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 kola-runext-test.sh[1209]: Could not run a simple command inside a toolbox

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 kola-runext-test.sh[1209]: + exit 1

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 systemd[1]: kola-runext.service: Main process exited, code=exited, status=1/FAILURE

[2021-03-04T20:32:02.467Z] Mar 04 20:32:01 qemu0 systemd[1]: kola-runext.service: Failed with result 'exit-code'.
```

but not clear to me why.

travier:
The fix for the existing toolbox issue (https://github.com/containers/toolbox/pull/712) has not been included in a release yet so the test will fail. What's not ideal right now is that this will not fail exactly at the right place as machinectl does not return errors from spawned commands so the test will always error out in the last check.

tests/kola/toolbox/test.sh
martymichal: > Can you elaborate why?

Of course. I shouldn't have left the comment without a proper explanation. I apologize.

As you mentioned, it would bloat the tools a bit. I have to admit that in the case of Toolbox that is not that much of a problem considering the size of the tool. Also the fact that every command can be executed without much problems from the code. Thinking about it now, it isn't that big of a problem...

Just paused at writing this reply and spent some time reading your [article](https://blog.verbum.org/2020/12/01/committed-to-the-integrity-of-your-root-filesystem/), going through CoreOS/rpm-ostree tests and learning about Go testing. The knowledge I gained from the tests themselves is close to none caused by my lack of insight :). The same applies to my knowledge of autpkgtest. I'd have to dive deeper.

I still feel like the testing (even though not extremely extensive) should not be part of the main binary. Instead it should be available separately. I knew that Go has built-in testing facilities but did not know it was possible export the test binaries and use them on other systems. I feel like this is a place where the testing could go in the case of Toolbox. Of course, this approach will require the use of e.g., `foo-tests` subpackage as you've already mentioned. Still, we already have a library in Toolbox for more convenient "shell-ing" out, we could use that in on of the particular tests and ship it as the "basic integration test". Now, I'm just throwing out my thoughts :).

> In particular I think the needs-internet flag is really, really useful at scale to separate "tests that can flake because e.g. docker.io rate limited me" versus truly self-contained/isolated tests.

This is something that needs to be taken into consideration when testing Toolbox. We know from experience that registries can not be fully relied on.

> So the spec might be more like $project --output-autopkgtests and then the framework executes them - i.e. it could output multiple tests and use the metadata keys from the autopkgtest.

I'm not exactly sure how that applies to Toolbox but I interpret it as `toolbox --output-autopkgtests`. That is something I'd like to avoid, if possible. But I may be, possibly, just too paranoid.

martymichal: Any more thoughts about this? Thinking back again, adding a very minimal testing facilities even to the main binary does not sound all that bad. Is that something you'd want us to provide?

travier: I think that having a couple basic tests included would help as, as you can see here, the test I made is not really great and coverage is rather small.

martymichal: Fair enough. I'll try to put something together in the near future.

travier:
Alright, I fixed my test and I have this passing now on testing which means that the previously mentioned fix was not the root cause of the issue as there hasn't been a new toolbox release yet with that fix.

Anyway, this should get us basic coverage for future breakage.

travier:
This failed on upgrade test with is untouched by this PR so seems unrelated:

```
[2021-03-19T18:31:27.920Z] --- PASS: ext.config.toolbox (58.96s)
```

travier:
Tests passed. Ready to merge.

tests/kola/toolbox/kola.json
```diff
@@ -0,0 +1,3 @@
+{
+    "tags": "needs-internet"
```
jlebon: Note you can fold this into `test.sh` using a comment [like this](https://github.com/coreos/fedora-coreos-config/blob/02c41e3012911330699626278bde3f2393c87f59/tests/kola/chrony/dhcp-propagation#L13).

tests/kola/toolbox/test.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if ! machinectl shell core@ /bin/toolbox run touch ok_toolbox 1>/dev/null; then
+	fatal "Could not run a simple command inside a toolbox"
+fi
+
+if [[ ! -f '/home/core/ok_toolbox' ]]; then
+	fatal "Could not run a simple command inside a toolbox"
+fi
+
+ok toolbox
```
jlebon: To tie it off, would be neat to `toolbox rm` the container here.

jlebon (approved):
Cool! Two comment, but LGTM as is! (Feel free to self-merge.)

travier:
Added `toolbox rm` test as suggested and improved logic to check each step instead of failing late on the last one on error.

tests/kola/toolbox/test.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# Make sure that basic toolbox functionnality is working:
+# - Creating a toolbox
+# - Running a command in a toolbox
+# - Removing all toolbox containers
+#
+# Important note: Commands are run indirectly via calls to `machinectl shell`
+# to re-create the user environment needed for unprivileged podman
+# fonctionnality. However, machinectl shell does not propagate the exit
```
martymichal: Nitpick: `fonctionnality` -> `functionality`

martymichal (approved):
Found a typo, otherwise looks great 👍!

tests/kola/toolbox/test.sh
jlebon: Fixed it for @travier to avoid waiting for another round trip! :)

PR #863, opened 2021-02-23
overlay: create 40-coreos-systemd.preset

cgwalters:
overlay: create 40-coreos-systemd.preset

Let's keep 40-coreos.preset to be only `enable` stanzas for
"our stuff" overall.

Create a separate `40-coreos-systemd.preset` which only
relates to systemd things because it really is a distinct
problem domain.

---

40-coreos-systemd.preset: Disable systemd-sysext.service

I was just reading the pre-release notes: https://lists.freedesktop.org/archives/systemd-devel/2021-February/046147.html

This hasn't been tested with ostree/rpm-ostree and heavily overlaps
with the latter.  Preemptively disable the service; it will hopefully
be subpackaged though for Fedora.

---

overlay/40-coreos-systemd:: Add systemd-firstboot disablement here too

This way we're moving closer to having our systemd overrides in one place.

---


jlebon (approved):
Hmm, we should keep an eye on this `sysext` thing. Interested to see where they're going with that.

PR #865, opened 2021-02-23
manifest: split out networking tools into separate file

miabbott:
As part of requesting `iproute-tc`[0], it was suggested to break out
the generic networking tools into a separate file so that they could
be shared with RHCOS.

[0] https://github.com/coreos/fedora-coreos-tracker/issues/742


miabbott:
Built this locally and observed no RPM changes in the ostree commit.

cgwalters (approved):
Thanks for doing this!  We need to do more of this to also pick up e.g. the stalld addition.

Looks sane to me!

PR #866, opened 2021-02-23
overrides: pin clevis to 15-2.fc33

jlebon:
The dracut module from the latest Clevis v16 release has an undeclared
dependency on `seq`. There's a patch to fix that upstream:

https://github.com/latchset/clevis/pull/295

But for now, let's just pin to the previous release. This will unblock
lockfile bumps.


ashcrow (approved):

jlebon:
I just cherry-picked this to `next-devel` as well instead of opening a separate PR since they're in the same state right now.

PR #868, opened 2021-02-24
manifest: use DEFAULT_HOSTNAME=localhost

### LINKED ISSUE #649: default hostname now is `fedora`, used to be `localhost`
Booted 33.20201006.1.0 from `next` stream on GCP and default hostname is `fedora`:
```
[    2.041577] systemd[1]: No hostname configured.
[    2.042356] systemd[1]: Set hostname to <fedora>.
```

OKD-on-FCOS expects default hostname to be `localhost` (as on AWS).

See https://github.com/openshift/machine-config-operator/pull/2160#issuecomment-710059362

jlebon:
systemd in f34 learned to read the default hostname from a
`DEFAULT_HOSTNAME` key in `/etc/os-release`.

Use that to set the hostname back to `localhost` on f34 so that once we
move over we can unpin the systemd build which had this same effect at
compilation time.

Ideally, this would be part of `fedora-release`, but I'm trying to aim
for a better fix where the compiled-in default hostname in Fedora is
back to `localhost` and only on the variants where we really want
`fedora` would we set `DEFAULT_HOSTNAME=fedora` in `fedora-release`.

So for now, let's just do this here.

Closes: coreos/fedora-coreos-tracker#649

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -1,8 +1,7 @@
 packages:
-  # Fast-track systemd update that reverts fallback hostname change
-  # https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-739956843
-  # This is a one-off build, please don't remove it without talking
-  # to dustymabe or jlebon.
+  # Keep this until we move to Fedora 34.
+  # https://github.com/coreos/fedora-coreos-tracker/issues/649
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1892235#c25
   systemd:
     evra: 246.7-1.fc33.aarch64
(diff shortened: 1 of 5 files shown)


bgilbert (approved):
LGTM.  The description seems a bit confusing though: it appears the postprocess script will only fix the problem :sparkles: in the future :sparkles:, and we're resigning ourselves to pinning systemd for 2-3 more months?

jlebon:
Thanks, I tweaked the commit message.

It does fix the problem now for `branched` and `rawhide`. My goal with this is to have everything ready so that at rebase time, there shouldn't be any extra work or surprises.

bgilbert (approved):

PR #874, opened 2021-03-03
transposefs: error out cleanly if less than 1G of RAM available

jlebon:
I commonly do `cosa run` to test something related to rootfs
reprovisioning and forget to allocate more RAM. The failure mode is a
massive spam of "Input/output error" messages from `cp` unable to copy
more files into RAM (and there doesn't seem to be a way to have it fail
on first error; yet another reason we should rewrite this into !shell).

Instead, just sanity-check that we have at least 1G of RAM and error out
with a nice and clear message if not.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -100,6 +100,16 @@ case "${1:-}" in
             echo "Found duplicate or missing ESP, BIOS-BOOT, or PReP labels in config" >&2
             exit 1
         fi
+        mem_available=$(grep MemAvailable /proc/meminfo | awk '{print $2}')
+        # Just error out early if we don't even have 1G to work with. This
+        # commonly happens if you `cosa run` but forget to add `--memory`. That
+        # way you get a nicer error instead of the spew of EIO errors from `cp`.
+        # The amount we need is really dependent on a bunch of factors, but just
+        # ballpark it at 3G.
+        if [ "${mem_available}" -lt $((1*1024*1024)) ] && [ "${wipes_root}" != 0 ]; then
+            echo "Root reprovisioning requires at least 3G of RAM" >&2
+            exit 1
+        fi
         modprobe zram num_devices=0
         read dev < /sys/class/zram-control/hot_add
         # disksize is set arbitrarily large, as zram is capped by mem_limit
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -104,10 +104,20 @@ case "${1:-}" in
...
         echo 10G > /sys/block/zram"${dev}"/disksize
+        mem_available=$(grep MemAvailable /proc/meminfo | awk '{print $2}')
+        # Just error out early if we don't even have 1G to work with. This
+        # commonly happens if you `cosa run` but forget to add `--memory`. That
+        # way you get a nicer error instead of the spew of EIO errors from `cp`.
+        # The amount we need is really dependent on a bunch of factors, but just
+        # ballpark it at 3G.
+        if [ "${mem_available}" -lt $((1*1024*1024)) ] && [ "${wipes_root}" != 0 ]; then
+            echo "Root reprovisioning requested but not enough RAM available; please allocate at least 3G" >&2
+            exit 1
```
bgilbert, line 111: Nit: leaves the zram device allocated (at the e-shell, meh).  We could move the test earlier.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -104,10 +104,20 @@ case "${1:-}" in
...
         # disksize is set arbitrarily large, as zram is capped by mem_limit
         echo 10G > /sys/block/zram"${dev}"/disksize
+        mem_available=$(grep MemAvailable /proc/meminfo | awk '{print $2}')
+        # Just error out early if we don't even have 1G to work with. This
+        # commonly happens if you `cosa run` but forget to add `--memory`. That
+        # way you get a nicer error instead of the spew of EIO errors from `cp`.
+        # The amount we need is really dependent on a bunch of factors, but just
+        # ballpark it at 3G.
+        if [ "${mem_available}" -lt $((1*1024*1024)) ] && [ "${wipes_root}" != 0 ]; then
+            echo "Root reprovisioning requested but not enough RAM available; please allocate at least 3G" >&2
```
bgilbert: This is more than 80 characters; do we care?

bgilbert (approved):
LGTM.  Thanks for fixing this papercut!

PR #875, opened 2021-03-03
manifests: improve ability to share downstream

miabbott:
This introduces two new categories of packages which exist in both
the FCOS and RHCOS manifests.

The `system-configuration` manifest includes packages related to
configuration and management of the system itself.  Think storage,
NTP, users/groups, etc.

The `user-experience` manifest includes packages that will generally
make a users time on the system more pleasant.  Think commonly used
Linux utilities, container tooling, remote access, etc.

The goal of this change is to improve the ability to share these
manifests with RHCOS and reduce the duplication of packages in the
FCOS and RHCOS manifests.

### DIFF
--- a/manifests/user-experience.yaml
+++ b/manifests/user-experience.yaml
@@ -1,0 +1,36 @@
+# These packages are either widely used utilities/services or 
+# are targeted for improving the general CoreOS user experience.
+# It is intended to be kept generic so that it may be shared downstream with
+# RHCOS.
+
+packages:
+  # Basic user tools
+  ## jq - parsing/interacting with JSON data
+  - bash-completion
+  - coreutils
+  - jq
+  - less
+  - sudo
+  - vim-minimal
+  # File compression/decompression
+  ## (bsdtar - dependency of 35coreos-live dracut module)
+  - bsdtar
+  - bzip2
+  - gzip
+  - tar
+  - xz
+  # Improved MOTD experience
+  - console-login-helper-messages-issuegen
+  - console-login-helper-messages-profile
+  # DNS/DHCP server
+  - dnsmasq
+  # kdump support
+  # https://github.com/coreos/fedora-coreos-tracker/issues/622
+  - kexec-tools
+  # Remote Access
+  - openssh-clients openssh-server
+  # Container tooling
+  - podman
+  - runc
+  - skopeo
+  - toolbox
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -6,6 +6,8 @@ include:
   - ignition-and-ostree.yaml
   - file-transfer.yaml
   - networking-tools.yaml
+  - system-configuration.yaml
+  - user-experience.yaml
 
 initramfs-args:
   - --no-hostonly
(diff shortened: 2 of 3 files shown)


miabbott:
This is a first pass at just the packages that are explicitly called out in each manifest.  Additional refinements may be desirable for packages that are not explicitly enumerated in the manifest, but are pulled in via dependencies.

This built successfully for me on my local system and no changes to the ostree content were observed.

miabbott:
Feel free to bikeshed on naming, categorization,  etc.

ashcrow:
Likely not related as I don't see anything changing for zincati in the PR....

```
[2021-03-03T23:34:40.864Z]     --- FAIL: fcos.upgrade.basic/upgrade-from-previous (245.00s)

[2021-03-03T23:34:40.864Z]             basic.go:341: failed to find 'zincati_update_agent_updates_enabled 1' on Zincati metrics.promsock
```

ashcrow:
I like this idea :+1:

cgwalters (approved):
Awesome, this looks really good to me!

manifests/user-experience.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - bsdtar
+  - bzip2
+  - gzip
+  - tar
+  - xz
+  # Improved MOTD experience
+  - console-login-helper-messages-issuegen
+  - console-login-helper-messages-profile
+  # DNS/DHCP server
+  - dnsmasq
```
jlebon, line 26: Hmm, this looks out of place. Is dnsmasq really a UX package?

jlebon:
Not strongly opposed to this, but... this feels a bit forced to me. If the intent is sharing package lists more easily with RHCOS, then let's just have an explicit e.g. `manifests/rhcos-shared.yaml` file instead?

manifests/user-experience.yaml
travier: Agree that this one is weird as we have dnsmasq only for podman support.

miabbott: Agreed, it is not a perfect fit. Most of the other packages fit more cleanly into the two categories and I made a best effort to fit it in.

I can slot it under 'containers' if that's why it is there in the first place.

miabbott:
> Not strongly opposed to this, but... this feels a bit forced to me. If the intent is sharing package lists more easily with RHCOS, then let's just have an explicit e.g. `manifests/rhcos-shared.yaml` file instead?

Sure!  I was trying to follow previous categorization attempts in breaking up the package sets.  But one larger, shared manifest file would be fine, too.

Maybe I'll file an issue to discuss the best way to break up the manifests?

jlebon:
> Maybe I'll file an issue to discuss the best way to break up the manifests?

Yeah, worth discussing I think.

I mean, doing it as a single shared manifest has disadvantages too (like re-usability for other potential FCOS-based variants), but if the classifications are too artificial, then it's not really useful for that either (and might end up prompting more fracturing). Right now at least, IMO it'd be cleaner to have a single `rhcos-shared.yaml` instead. (It also makes it more explicit when adding a package whether we mean for it to be in just FCOS or also RHCOS, regardless of its "category").

PR #876, opened 2021-03-04
README: add blurb about interacting with CI jobs

miabbott:
I keep forgetting how to interact with failed CI jobs, so I documented
now to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,3 +149,16 @@ done
 # After verifying the list looks good:
 #   - koji untag-build coreos-pool $untaglist
 ```
+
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
This is fine, though I think a better fix would be some sort of banner on the top of the Jenkins instance web UI that links to this information in a central place.

jlebon:
Thanks for writing this up Micah! I took your blurb and adapted it into https://github.com/coreos/coreos-ci/pull/26. I submitted a follow-up which just links to that in https://github.com/coreos/fedora-coreos-config/pull/879.

PR #877, opened 2021-03-04
manifest: Block plymouth

cgwalters:
For (datacenter/cloud oriented) servers, we want to see the details by default.
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HSMISZ3ETWQ4ETVLWZQJ55ARZT27AAV3/

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -96,6 +96,9 @@ exclude-packages:
   # Let's make sure initscripts doesn't get pulled back in
   # https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-611566254
   - initscripts
+  # For (datacenter/cloud oriented) servers, we want to see the details by default.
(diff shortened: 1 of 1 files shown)


manifests/fedora-coreos.yaml
```diff
@@ -96,6 +96,9 @@ exclude-packages:
   # Let's make sure initscripts doesn't get pulled back in
   # https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-611566254
   - initscripts
+  # For (datacenter/cloud oriented) servers, we want to see the details by default.
+  # https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HSMISZ3ETWQ4ETVLWZQJ55ARZT27AAV3/
+  - plymouth
```
cgwalters, line 101: And to be clear, this package is in the set of `exclude-packages`, which we hard fail if they are somehow pulled in via dependencies.

lucab (approved):

jlebon (approved):

PR #878, opened 2021-03-04
Rename coreos-growpart to ignition-ostree-growfs and teach it to grow LUKS containers

jlebon:
See individual commits.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1934174.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 
 # If root reprovisioning was triggered, this file contains state of the root
 # partition *before* ignition-disks.
-saved_partstate=/run/ignition-ostree-rootfs-partstate.json
+saved_partstate=/run/ignition-ostree-rootfs-partstate.sh
 
 # We run after the rootfs is mounted at /sysroot, but before ostree-prepare-root
 # moves it to /sysroot/sysroot.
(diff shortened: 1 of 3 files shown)


jlebon:
Works fine on FCOS. Still need to test RHCOS.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -71,6 +71,13 @@ lsblk --paths --pairs -o NAME,TYPE,PKNAME "${partition}" | while read line; do
             # XXX: ideally this'd be idempotent and we wouldn't `|| :`
             growpart "${PKNAME}" "${partnum}" || :
             ;;
+        crypt)
+            # XXX: yuck... we need to expose this sanely in clevis
+            (. /usr/bin/clevis-luks-common-functions
+             eval $(udevadm info --query=property --export "${NAME}")
+             clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
```
jlebon, line 76: @sergio-correia, am I correct there isn't a cleaner/supported way to do this today?

sergio-correia: You mean getting the passphrase? There is `clevis luks pass` (from v16) that uses `clevis_luks_unlock_device`, however right now it also requires a slot to be passed.

cgwalters:
Prep patches LGTM (maybe split them off?)

Hmmm...it feels like the cleanest is for Ignition to hand us this information.  Or...we could add an extension to Ignition that says whether or not to grow, but it defers to the OS somehow?  (I guess this would require anyone specifying luks-for-root to also specify a new `grow: true` flag?)

cgwalters:
Actually last time this came up (somewhere, not finding in search) I think I argued that Ignition should be growing the root device at the time it creates it, but I think @bgilbert had an argument why we can't do that?  

It just seems weird to me to have a flow particularly for LUKS where we create the device small, then resize it in the same run of the initramfs.

jlebon:
Right, the context for this is that we do eventually want to get rid of this script, but there's design work left on how to do that. This is being discussed in https://github.com/coreos/fedora-coreos-tracker/issues/570 and https://github.com/coreos/fedora-coreos-tracker/issues/586. One thing we've recently discussed with @bgilbert and @arithx is (1) adding a dev symlink for a stable way to reference the boot disk, and (2) enhance the fcct sugar to resize the rootfs by default. This helps address the in-place LUKS case, though fully dropping this script will require something more generic, like an Ignition drop-in (though see option 3 in https://github.com/coreos/fedora-coreos-tracker/issues/586#issuecomment-777220000).

This PR is more a short/mid-term fix because we need something for 4.7 until we get rid of it.

jlebon:
OK cool, tested this on RHCOS 4.8 and 4.7 (both manually and also running the root reprovisioning external host tests)!

@Prashanth684 As a sanity-check, would you be able to test this on RHCOS 4.7 on ppc64le and s390x?

> Prep patches LGTM (maybe split them off?)

OK, done in https://github.com/coreos/fedora-coreos-config/pull/880!

bgilbert:
> Right, the context for this is that we do eventually want to get rid of this script, but there's design work left on how to do that.

Well, sort of.  For the LUKS-root case, we want to allow the user to explicitly resize the root partition so this script doesn't need to get involved.  That's not possible right now unless the user hardcodes the device path for the boot disk.

But in the simple cases, I think we still want the script.  There are two: 1. the user doesn't touch the boot disk partitioning at all, and 2. the user adds a data partition without touching the rootfs.  If we handled those with Ignition, we'd need to do a full filesystem copyout + mkfs + copyin, which is much more expensive (and requires more RAM) than resizing in place.

bgilbert:
The root-disk symlink proposal is now in https://github.com/coreos/fedora-coreos-tracker/issues/759.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -17,24 +17,39 @@ path=/sysroot
...
-    if [ "${partstate}" != "$(cat "${saved_partstate}")" ]; then
-        echo "coreos-growpart: detected rootfs partition changes; not auto-growing"
+if [ ! -f "${saved_partstate}" ]; then
+    partition=$(realpath /dev/disk/by-label/root)
+else
+    # The rootfs was reprovisioned. Our rule in this case is: we only grow if
+    # the partition backing the rootfs is the same and its size didn't change
+    # (IOW, it was an in-place reprovisioning; e.g. xfs -> btrfs).
+    source "${saved_partstate}"
+    if [ "${TYPE}" != "part" ]; then
```
bgilbert, line 27: Are we concerned about a missing `TYPE`/`NAME`/`SIZE`?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -17,24 +17,39 @@ path=/sysroot
...
+        # this really should never happen; but play nice
+        echo "$0: original rootfs blockdev not of type 'part'; not auto-growing"
+        exit 0
+    fi
+    partition=$(realpath ${NAME})
+    PREV_SIZE=$SIZE
+    eval $(lsblk --nodeps --pairs -b --paths -o NAME,TYPE,SIZE "${partition}")
```
bgilbert: `TYPE` unused.

jlebon: Thanks, ended up just reworking this to drop the `eval`! :tada:

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: We expect them to exist and we'll error out if they don't because of `set -u`.

Prashanth684:
> OK cool, tested this on RHCOS 4.8 and 4.7 (both manually and also running the root reprovisioning external host tests)!
> 
> @Prashanth684 As a sanity-check, would you be able to test this on RHCOS 4.7 on ppc64le and s390x?
> 
> > Prep patches LGTM (maybe split them off?)
> 
> OK, done in #880!

tested on ppc64le and s390x with the root re-provisioning test and s390x baremetal. lgtm!

cgwalters (approved):
A lot going on here, but it looks good to me.

bgilbert (approved):

jlebon:
Pure rebase on top of #880, enabled auto-merge based on approvals!

PR #881, opened 2021-03-08
denylist: Add default-checks for chrony AVC

cgwalters:
Pairs with https://github.com/coreos/coreos-assembler/pull/2067
which will allow us to turn on SELinux AVC checks before/after
all tests.  But we can't do that until the chrony AVC is fixed.


cgwalters:
Draft since we're going to do something more sophisticated.

PR #882, opened 2021-03-09
overlay: temporarily ship `seq` in initrd

jlebon:
Previously we pinned clevis to v15 in order to work around
https://github.com/latchset/clevis/pull/295.

But overrides are not set up in the tagger for mechanical streams, so we
can't easily pin in `branched` and `rawhide`. Let's just unpin and
temporarily ship `seq` ourselves to work around this. Put that temporary
hack in `40ignition-ostree` since that's the closest module conceptually
to clevis.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -59,6 +59,9 @@ install() {
         sgdisk    \
         find
 
+    # TODO f34: check if we can drop this temporary workaround for https://github.com/latchset/clevis/pull/295
+    inst_multiple seq
+
     for x in mount populate; do
         install_ignition_unit ignition-ostree-${x}-var.service
         inst_script "$moddir/ignition-ostree-${x}-var.sh" "/usr/sbin/ignition-ostree-${x}-var"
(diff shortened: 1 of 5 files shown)


jlebon:
This will fix LUKS-related test failures in `branched` and `rawhide`.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
```diff
@@ -59,6 +59,9 @@ install() {
         sgdisk    \
         find
 
+    # XXX: temporary workaround for https://github.com/latchset/clevis/pull/295
```
cgwalters: What we need is some sort of periodic reminder system for this stuff.  How about a standardized tag like `# TODO F34 rebase` that we can grep for to find things that we can likely drop?

cgwalters (approved):
LGTM, one optional.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
jlebon: Yeah, good idea. (I was playing with making it a time bomb instead using e.g. `grep seq ...` but it felt too blunt.)

PR #883, opened 2021-03-09
kola-denylist.yaml: add ext.config.podman.rootless-systemd

jlebon:
This test is currently failing on f34+ due to a regression in buildah:
https://github.com/containers/buildah/issues/3071


jlebon:
With this and https://github.com/coreos/fedora-coreos-config/pull/882, we should be able to get `branched` builds going again!

cgwalters (approved):

PR #884, opened 2021-03-09
overrides: move to an arch-independent lockfile

jlebon:
Now that rpm-ostree and cosa support it, move to just using a
`manifest-lock.overrides.yaml` which uses `evr` instead of `evra` so
that we can avoid all the duplication.

The previous filenames are still supported in case we do have
arch-specific overrides (e.g. grub) in the future.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/2075

jlebon:
Actually, before we do this, we also need to teach it to the tagger and enhance [`download-overrides.py`](https://github.com/coreos/fedora-coreos-releng-automation/blob/master/scripts/download-overrides.py). So marking back as draft.

jlebon:
Requires: https://github.com/coreos/fedora-coreos-releng-automation/pull/131

jlebon:
We have new rpm-ostree in cosa now, so this should be good to go!

manifest-lock.overrides.yaml
```diff
@@ -3,28 +3,28 @@ packages:
...
-    evra: 246.7-1.fc33.s390x
+    evr: 246.7-1.fc33
   systemd-libs:
-    evra: 246.7-1.fc33.s390x
+    evr: 246.7-1.fc33
   systemd-pam:
-    evra: 246.7-1.fc33.s390x
+    evr: 246.7-1.fc33
   systemd-rpm-macros:
     evra: 246.7-1.fc33.noarch
```
jlebon, line 14: Notice here we're still using `evra` for `noarch` packages. This is because of limitations in how `coreos-koji-tagger` and `download-overrides.py` treat this data and try to match them up to Koji builds. We should be able to fix that up in the future. But anyway, having `noarch` packages here is fine since it's (of course) valid on all arches.

travier (approved):

PR #885, opened 2021-03-10
manifests: Fix split to enable sharing with RHCOS

travier:
- Move stalld back to fedora-coreos-base until it is released in RHEL
- Keep dnsmasq alongside podman-plugins as explicit dependency

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -125,6 +125,9 @@ postprocess:
       echo 'MOTD_FILE=/usr/share/misc/motd:/run/motd:/run/motd.d:/etc/motd:/etc/motd.d' >> /etc/login.defs
     fi
 
+# Packages listed here should be specific to Fedore CoreOS (as in not yet
+# available in RHCOS or not desired in RHCOS). All other packages should go
+# into one of the sub-manifests listed at the top.
 packages:
   # Security
   - polkit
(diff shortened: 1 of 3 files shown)


manifests/fedora-coreos-base.yaml
```diff
@@ -137,7 +140,7 @@ packages:
   - fuse-overlayfs slirp4netns
   # name resolution for podman containers
   # https://github.com/coreos/fedora-coreos-tracker/issues/519
-  - podman-plugins
+  - podman-plugins dnsmasq
```
miabbott, line 143: We definitely need `dnsmasq` in RHCOS.  There are a growing number of OCP customers that are using it.

manifests/fedora-coreos-base.yaml
```diff
@@ -167,15 +170,20 @@ packages:
...
   - kbd
   # nvme-cli for managing nvme disks
   - nvme-cli
   # zram-generator (but not zram-generator-defaults) for F33 change
   # https://github.com/coreos/fedora-coreos-tracker/issues/509
   - zram-generator
+  # Improved MOTD experience
+  - console-login-helper-messages-motdgen
+  - console-login-helper-messages-issuegen
+  - console-login-helper-messages-profile
```
miabbott: These are also included in RHCOS

manifests/fedora-coreos-base.yaml
travier: From what I remember from the inclusion discussion, it is only in FCOS for podman plugins support so we can not rely on it staying there. It is also explicitly listed in the RHCOS manifest. I don't have a strong opinion here.

miabbott: Ah I see; makes sense to leave it here.

jlebon (approved):
Thanks!

PR #886, opened 2021-03-10
rootless-systemd: update packages before installing httpd

vrutkovs:
Fedora container may be outdated and in order to install httpd existing packages may require an update.

Test in https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/job/fedora-coreos/job/fedora-coreos-fedora-coreos-pipeline/1168/ fails with
```
Mar 10 05:15:52 qemu0 kola-runext-rootless-systemd[1997]:   systemd-pam-246.10-1.fc33.x86_64
Mar 10 05:15:52 qemu0 kola-runext-rootless-systemd[1997]:   systemd-rpm-macros-246.10-1.fc33.noarch
Mar 10 05:15:52 qemu0 kola-runext-rootless-systemd[1997]:   xkeyboard-config-2.30-3.fc33.noarch
Mar 10 05:15:52 qemu0 kola-runext-rootless-systemd[1997]: Failed:
Mar 10 05:15:52 qemu0 kola-runext-rootless-systemd[1997]:   httpd-2.4.46-9.fc33.x86_64
Mar 10 05:15:52 qemu0 systemd[1]: kola-runext.service: Main process exited, code=exited, status=125/n/a
```

### DIFF
--- a/tests/kola/podman/rootless-systemd
+++ b/tests/kola/podman/rootless-systemd
@@ -26,7 +26,8 @@ set -euxo pipefail
 cd $(mktemp -d)
 cat <<EOF > Containerfile
 FROM registry.fedoraproject.org/fedora:33
-RUN dnf -y install systemd httpd \
+RUN dnf -y update \
+&& dnf -y install systemd httpd \
 && dnf clean all \
 && systemctl enable httpd
 ENTRYPOINT [ "/sbin/init" ]


jlebon:
Hmm, did you see https://github.com/coreos/fedora-coreos-config/pull/883? (Just started another `branched` build which has this now!). I think the error there is due to some caps regression in rootless mode (see https://github.com/containers/buildah/issues/3071).

Does the test pass on a local build of `branched`?

vrutkovs:
I didn't try it yet and I'm not entirely sure if it fixes the issue, but this PR might be useful in the long run

jlebon (approved):
Yeah, the change itself LGTM regardless.

(`dnf install` should already be auto-updating packages as needed to satisfy the transaction, but in general it's a good idea to update everything too, so this makes the test more realistic.)

PR #888, opened 2021-03-11
image.yaml: bump virtual disk images to 10G

jlebon:
Right now, our images are triggering the new 8G warning because the disk
size itself is 8G, so the rootfs is less than that. Let's go to 10G so
we meet our own standard.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -1,6 +1,7 @@
-# This replaces image.ks
-# size is the target disk size in GB.
-size: 8
+# Target disk size in GB.
+# Make it at least 10G because we want the rootfs to be at least 8G:
+# https://github.com/coreos/fedora-coreos-tracker/issues/586
+size: 10
 
 extra-kargs:
     # Disable SMT on systems vulnerable to MDS or any similar future issue.


cgwalters:
Heh, we need a kola test `journalctl --grep=warning`.

bgilbert:
That's expected, though, right?  We assume the disk will be resized before use.

jlebon:
> That's expected, though, right? We assume the disk will be resized before use.

Hmm, but IMO we should have the default disk not trigger this warning. E.g. right now the instructions in https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-qemu/#_setting_up_a_new_vm trigger it. (And so in all the kola tests too.)

jlebon:
My excuse for not noticing this earlier is that the initial size I was testing with was 5G, which comfortably fits in the default 8G. :see_no_evil:

bgilbert (approved):
Ah, right, this parameter mostly affects sparse image formats, and not the bare-metal image.

lucab:
For cross-reference: the warning is https://github.com/coreos/fedora-coreos-config/pull/850 and the umbrella ticket is https://github.com/coreos/fedora-coreos-tracker/issues/586.

jlebon:
To raise awareness: this implicitly requires anyone who previously assumed an 8G image to update their scripts/cloud provisioning manifests to increase the disk size request. We probably should've sent a heads up on the list about it.

PR #889, opened 2021-03-11
overrides: pin to kernel-5.10.19-200.fc33 for rootless podman

jlebon:
There's a regression in 5.10.20+ which breaks rootless podman:
https://github.com/containers/buildah/issues/3071

(This is the same regression which prompted
https://github.com/coreos/fedora-coreos-config/pull/883 but for the
production streams, let's try to avoid exposing the regression to users
for now while it gets sorted out.)

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -14,3 +14,11 @@ packages:
     evra: 246.7-1.fc33.noarch
   systemd-udev:
     evra: 246.7-1.fc33.aarch64
+  # There's a regression in 5.10.20+ which breaks rootless podman
+  # https://github.com/containers/buildah/issues/3071
+  kernel:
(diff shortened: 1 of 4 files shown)


jlebon:
If it takes too long to get a fix, we can unpin and live with it. But at least for now this will unblock lockfile bumps for other packages.

cgwalters:
One thing I'd highlight with us is that if we hooked up our CI to upstream kernel builds directly, we would find stuff like this much earlier.

cgwalters (approved):

bgilbert:
This is fixed by gregkh/linux@5d5422a294e6 in 5.10.24.

jlebon:
And in 5.11.7 (https://github.com/gregkh/linux/commit/7fbc077be2f3fad5d75ddfd4b598eeba66459c5d).

PR #892, opened 2021-03-12
ci: print human-readable RPM diff and error out if downgrade detected

jlebon:
commit cbe1253b8cfb741ebf0d647e4cd06bf0da12697c
Date:   Fri Mar 12 16:27:23 2021 -0500

    ci: print human-readable RPM diff

    The `.pkgdiff` element from the build metadata isn't very useful to
    humans. Let's just re-invoke `rpm-ostree db diff` to get a nice legible
    diff.

commit 0b0a10e41a3d906b0975050b93fe1bfd420a7f5e
Date:   Fri Mar 12 16:28:27 2021 -0500

    ci: error out if downgrade detected

    In the majority of cases, we don't expect new builds to downgrade RPMs.
    Let's default to erroring out and contributors can consciously override
    CI if was really expected.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -3,6 +3,8 @@
 cosaPod {
     checkoutToDir(scm, 'config')
 
+    def basearch = shwrapCapture("cosa basearch")
+
     shwrap("cd config && ci/validate")
 
     shwrap("""
@@ -17,8 +19,8 @@ cosaPod {
     // use a --parent-build arg so we can diff later and it matches prod
     def parent_arg = ""
     def parent_commit = ""
-    if (shwrapRc("test -e /srv/fcos/builds/latest/x86_64/meta.json") == 0) {
-        shwrap("cp /srv/fcos/builds/latest/x86_64/meta.json .") // readJSON wants it in the WORKSPACE
+    if (shwrapRc("test -e /srv/fcos/builds/latest/${basearch}/meta.json") == 0) {
+        shwrap("cp /srv/fcos/builds/latest/${basearch}/meta.json .") // readJSON wants it in the WORKSPACE
         def meta = readJSON file: "meta.json"
         def version = meta["buildid"]
         parent_arg = "--parent-build ${version}"
@@ -48,7 +50,7 @@ cosaPod {
         stage("RPM Diff") {
             shwrap("""
                 cd /srv/fcos
-                new_commit=\$(jq -r '.["ostree-commit"]' builds/latest/x86_64/meta.json)
+                new_commit=\$(jq -r '.["ostree-commit"]' builds/latest/${basearch}/meta.json)
                 rpm-ostree db diff --repo tmp/repo ${parent_commit} \${new_commit} | tee tmp/diff.txt
                 if grep -q Downgraded tmp/diff.txt; then
                     echo "Downgrade detected. This is likely unintentional. If not, you may safely ignore this error."
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -14,7 +14,18 @@ cosaPod {
         cosa buildprep https://builds.coreos.fedoraproject.org/prod/streams/${env.CHANGE_TARGET}/builds
     """)
 
-    fcosBuild(skipInit: true, extraFetchArgs: '--with-cosa-overrides')
+    // use a --parent-build arg so we can diff later and it matches prod
+    def parent_arg = ""
+    def parent_commit = ""
+    if (shwrapRc("test -e /srv/fcos/builds/latest/x86_64/meta.json") == 0) {
+        shwrap("cp /srv/fcos/builds/latest/x86_64/meta.json .") // readJSON wants it in the WORKSPACE
+        def meta = readJSON file: "meta.json"
+        def version = meta["buildid"]
+        parent_arg = "--parent-build ${version}"
+        parent_commit = meta["ostree-commit"]
+    }
+
+    fcosBuild(skipInit: true, extraFetchArgs: '--with-cosa-overrides', extraArgs: parent_arg)
 
     parallel metal: {
         shwrap("cd /srv/fcos && cosa buildextend-metal")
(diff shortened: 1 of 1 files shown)


jlebon:
Ahh, hitting a regression in the `download-overrides.py` script: https://github.com/coreos/fedora-coreos-releng-automation/pull/132

.cci.jenkinsfile
```diff
@@ -43,6 +43,9 @@ cosaPod {
                 previous_commit=\$(cat tmp/previous_commit)
                 new_commit=\$(jq -r '.["ostree-commit"]' builds/latest/x86_64/meta.json)
                 rpm-ostree db diff --repo tmp/repo \${previous_commit} \${new_commit} | tee tmp/diff.txt
+                if grep -q Downgraded tmp/diff.txt; then
```
cgwalters, line 55: I'd prefer:

- Add `rpm-ostree compose tree --disallow-downgrades`
- Change cosa to pass that by default
- Change the pipeline to have a declarative way to drop it

Hmm perhaps it's actually a declarative flag in the lockfiles?

jlebon: Hmm, maybe. I think we need to think this through a bit more. At least for the dev case, it's not at all clear that defaulting to disallowing downgrades is good (because in the end, the diff is with respect to whatever the last build was, which might've had e.g. overrides, or some other repos enabled, etc...).

But also, it feels odd to default to blocking downgrades when lockfiles are provided because presumably you know what you're asking for by having a lockfile for it. IOW, if we want a failure for this, it should be at the spot where the lockfile is being updated. And especially when they're updated as part of content promotion for a release, which is my primary concern for this PR. (Though it will also catch other ways that downgrades can happen, notably PRs to e.g. add overrides or add packages).

jlebon:
```
[2021-03-18T20:26:07.543Z] Downgraded:
[2021-03-18T20:26:07.543Z]   kernel 5.10.19-200.fc33 -> 5.10.18-200.fc33
[2021-03-18T20:26:07.543Z]   kernel-core 5.10.19-200.fc33 -> 5.10.18-200.fc33
[2021-03-18T20:26:07.543Z]   kernel-modules 5.10.19-200.fc33 -> 5.10.18-200.fc33
[2021-03-18T20:26:07.543Z] + grep -q Downgraded tmp/diff.txt
[2021-03-18T20:26:07.543Z] + echo 'Downgrade detected. This is likely unintentional. If not, you may ignoring this error.'
[2021-03-18T20:26:07.543Z] Downgrade detected. This is likely unintentional. If not, you may ignoring this error.
[2021-03-18T20:26:07.543Z] + exit 1
```

Cool!

Dropped the DNM commit and ready for review!

.cci.jenkinsfile
```diff
@@ -14,7 +14,18 @@ cosaPod {
         cosa buildprep https://builds.coreos.fedoraproject.org/prod/streams/${env.CHANGE_TARGET}/builds
     """)
 
-    fcosBuild(skipInit: true, extraFetchArgs: '--with-cosa-overrides')
+    // use a --parent-build arg so we can diff later and it matches prod
+    def parent_arg = ""
+    def parent_commit = ""
+    if (shwrapRc("test -e /srv/fcos/builds/latest/x86_64/meta.json") == 0) {
```
cgwalters: Let's have this use `$(arch)` to be nicer or any future multi-arch work?

.cci.jenkinsfile
```diff
@@ -33,7 +44,17 @@ cosaPod {
...
-    stage("RPM Diff") {
-        shwrap("jq .pkgdiff /srv/fcos/builds/latest/x86_64/meta.json")
+    if (parent_arg != "") {
+        stage("RPM Diff") {
+            shwrap("""
+                cd /srv/fcos
+                new_commit=\$(jq -r '.["ostree-commit"]' builds/latest/x86_64/meta.json)
+                rpm-ostree db diff --repo tmp/repo ${parent_commit} \${new_commit} | tee tmp/diff.txt
+                if grep -q Downgraded tmp/diff.txt; then
+                    echo "Downgrade detected. This is likely unintentional. If not, you may ignoring this error."
```
cgwalters: you may... something?

.cci.jenkinsfile
cgwalters: >  IOW, if we want a failure for this, it should be at the spot where the lockfile is being updated. 

OK yeah, I think I agree with that.

But what's the declarative way to override?

cgwalters:
Using this as a CI canary

.cci.jenkinsfile
jlebon: > > IOW, if we want a failure for this, it should be at the spot where the lockfile is being updated.
> 
> OK yeah, I think I agree with that.
> 
> But what's the declarative way to override?

None right now, but we can definitely do that down the line. Actually, I've just realized now that we could codify this as a kola upgrade test instead, and then you'd naturally be able to add it to `kola-denylist.yaml` if needed. Has a different set of trade-offs though.

jlebon:
Updated for comments!

.cci.jenkinsfile
```diff
@@ -33,7 +46,17 @@ cosaPod {
...
-    stage("RPM Diff") {
-        shwrap("jq .pkgdiff /srv/fcos/builds/latest/x86_64/meta.json")
+    if (parent_arg != "") {
+        stage("RPM Diff") {
+            shwrap("""
+                cd /srv/fcos
+                new_commit=\$(jq -r '.["ostree-commit"]' builds/latest/${basearch}/meta.json)
+                rpm-ostree db diff --repo tmp/repo ${parent_commit} \${new_commit} | tee tmp/diff.txt
+                if grep -q Downgraded tmp/diff.txt; then
+                    echo "Downgrade detected. This is likely unintentional. If not, you may safely ignore this error."
```
dustymabe, line 56: "you may safely ignore this error" -> i.e. manually push the merge button even though CI is showing a failure?

jlebon: Yeah, exactly.

dustymabe (approved):
LGTM

PR #900, opened 2021-03-18
overlay/growfs: handle broken lsblk in el8

jlebon:
Sometimes `lsblk` returns devices in the wrong order, so we can't rely
on it to iterate over the devices in hierarchical order. Instead, use
the `holders/` directory ourselves directly and use `--nodeps` when
calling `lsblk`.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1934174#c4
See: https://bugzilla.redhat.com/show_bug.cgi?id=1940607
See: https://github.com/coreos/coreos-installer/pull/453


jlebon:
(And yes, we want to rewrite this into a compiled language eventually. At the very least so we can use `lsblk --json` instead. See discussions in https://github.com/coreos/fedora-coreos-tracker/issues/570.)

travier (approved):
Weak LGTM from me (no testing). The eval calls make me sad 😢

mike-nguyen:
Tested the nvme and non nvme case.  Both successfully grew to fill the disk.

PR #901, opened 2021-03-18
[next-devel] lockfile: refresh after manifest changes

jlebon:
(cherry picked from commit 33cedaf13dc23521bb72be4c0c9c64cd0f84f4ef)


jlebon:
Lockfiles aren't automatically shared between `testing-devel` and `next-devel`. Alternatively, the next successful `bump-lockfile` run would also add it when it updates the lockfile.

bgilbert (approved):

Issue #805, opened 2021-01-05
Copy rootfs Ignition mountOptions to rootflags karg


bgilbert:
> Hmm, I think if we have magic for reprovisioning root, then we should also just handle `mount_options` too and translate that into `rootflags`. This is similar to how `with_mount_unit` also injects the mount options into the unit.

_Originally posted by @jlebon in https://github.com/coreos/fedora-coreos-config/pull/503#issuecomment-739967090_

PR #903, opened 2021-03-19
overlay: always use prjquota for XFS rootfs

jlebon:
By default, we use `prjquota` for the rootfs for container orchestrators
to monitor and set drive space limits.

However, with the added support for rootfs reprovisioning, we made this
conditional on the rootfs not being reprovisioned, with the assumption
that you can just set whatever options you'd like instead. Though
actually doing that is really awkward right now, and it requires
surfacing prjquota in user documentations (see
https://bugzilla.redhat.com/show_bug.cgi?id=1940704#c2).

Since AFAICT `prjquota` doesn't actually have any significant overhead,
let's just simplify the messaging to: we *always* enable prjquota on XFS
root filesystems. Users who want to override this can fallback to
`rpm-ostree kargs` (and eventually once we have
https://github.com/coreos/fedora-coreos-config/issues/805, to
`mountOptions: []`).

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1940704

### DIFF
--- a/tests/kola/root-reprovision/filesystem-only/test.sh
+++ b/tests/kola/root-reprovision/filesystem-only/test.sh
@@ -2,15 +2,24 @@
 # kola: {"platforms": "qemu", "minMemory": 4096}
 set -xeuo pipefail
 
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
 fstype=$(findmnt -nvr / -o FSTYPE)
 [[ $fstype == ext4 ]]
+ok "source is ext4"
 
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
       # check that the partition was grown
       if [ ! -e /run/ignition-ostree-growfs.stamp ]; then
-          echo "ignition-ostree-growfs did not run"
-          exit 1
+          fatal "ignition-ostree-growfs did not run"
       fi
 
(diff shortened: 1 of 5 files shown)


tests/kola/root-reprovision/filesystem-only/test.sh
```diff
@@ -15,6 +15,12 @@ fstype=$(findmnt -nvr / -o FSTYPE)
 [[ $fstype == ext4 ]]
 ok "source is ext4"
 
+rootflags=$(findmnt /sysroot -no OPTIONS)
+if ! grep prjquota <<< "${rootflags}"; then
+    fatal "missing prjquota in root mount flags: ${rootflags}"
+fi
+ok "root mounted with prjquota"
```
arithx: The root filesystem in this test is `ext4`, isn't this `prjquota` change only for `xfs` filesystems?

jlebon: Heh yeah, just saw the CI failure for this. Pushed an update!

arithx (approved):

cgwalters:
Nice work on this!  Hmm.  Definitely a bit of a philosophical issue here.  I feel like if we could go back to the start, arguably we might have not enabled `prjquota` at all by default and had higher level systems (OpenShift, etc.) explicitly enable it internally.  But, that ship has definitely sailed.

PR #904, opened 2021-03-19
overlay: propagate rootfs mountOptions to rootflags karg

### LINKED ISSUE #805: Copy rootfs Ignition mountOptions to rootflags karg

jlebon:
Right now, specifying `mountOptions` for the rootfs in the Ignition
config is ignored. To fix that, we simply need to have
`coreos-rootflags` inspect the Ignition config if the rootfs was
reprovisioned and extract the field.

This then takes effect on first boot (via
`ignition-ostree-mount-sysroot.sh` calling out to it) and subsequent
boots (via `rdcore rootmap` calling out to `coreos-rootflags`).

Closes: https://github.com/coreos/fedora-coreos-config/issues/805

### DIFF
--- a/tests/kola/root-reprovision/filesystem-only/test.sh
+++ b/tests/kola/root-reprovision/filesystem-only/test.sh
@@ -15,6 +15,12 @@ fstype=$(findmnt -nvr / -o FSTYPE)
 [[ $fstype == ext4 ]]
 ok "source is ext4"
 
+rootflags=$(findmnt /sysroot -no OPTIONS)
+if ! grep debug <<< "${rootflags}"; then
+    fatal "missing debug in root mount flags: ${rootflags}"
+fi
+ok "root mounted with debug"
+
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
       # check that the partition was grown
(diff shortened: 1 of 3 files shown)


tests/kola/root-reprovision/filesystem-only/test.sh
```diff
@@ -15,6 +15,12 @@ fstype=$(findmnt -nvr / -o FSTYPE)
 [[ $fstype == ext4 ]]
 ok "source is ext4"
 
+rootflags=$(findmnt /sysroot -no OPTIONS)
+if ! grep debug <<< "${rootflags}"; then
```
cgwalters, line 19: Unrelated, https://stackoverflow.com/a/35924143/12240197 is faster for this.

cgwalters (approved):
Nice!

bgilbert (approved):

tests/kola/root-reprovision/filesystem-only/test.sh
jlebon: Ahhh yup, that'd be a good use of bash's built-in regex support.

PR #906, opened 2021-03-19
overrides: move to locking by source RPM

jlebon:
This is just much more natural and easier to maintain.


jlebon:
This works on top of https://github.com/coreos/fedora-coreos-config/pull/884. (It doesn't strictly require it, but we can get that one in right now until this one is ready.)

jlebon:
Requires: https://github.com/coreos/rpm-ostree/pull/2680

travier (approved):

jlebon:
And of course, this will also require tweaks to `coreos-koji-tagger` and `download-overrides.py` to support this.

jlebon:
I was hacking on this and was pretty far along updating all of the GitHub Action, `download-overrides.py`, and `coreos-koji-tagger` to handle SRPMs correctly, but it's messy because SRPMs are pure strings in the RPM headers and not easy to work with in queries, etc... It also opens the door for correctly handling corner-cases like "what if an SRPM update also changes a binary RPM package name?".

Anyway, I think a good 80% of lockfile pain was addressed in https://github.com/coreos/fedora-coreos-config/pull/884. I think efforts are better spent improving other parts of the release process at this point.

PR #909, opened 2021-03-26
buildroot: Add jq

cgwalters:
This is used by rpm-ostree's CI; add it to the buildroot for
the same reason we ship it in FCOS.


bgilbert (approved):
LGTM

jlebon:
RAID reprovisioning test flaked. But for the buildroot stuff I think we can skip CI since it's not covered at all.

PR #292, opened 2020-03-10
overrides: fasttrack podman-1.8.1-0.7.rc4.fc31

dustymabe:
This release candidate contains a fix for a bug that can
leave rootless podman inoperable on a system. I've hit it
twice in the past 3 weeks: https://github.com/containers/libpod/issues/5306

PR #910, opened 2021-03-26
Move next-devel to cgroupsv2 for new nodes

jlebon:
All we need to do is to make the associated karg be specific to
`testing-devel` by uplisting it to `image.yaml`. Then `next-devel` will
naturally shed it when `image-base.yaml` gets synced over.

See: https://github.com/coreos/fedora-coreos-tracker/issues/292

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -150,3 +150,26 @@ if ! grep prjquota <<< "${rootflags}"; then
     fatal "missing prjquota in root mount flags: ${rootflags}"
 fi
 ok "root mounted with prjquota"
+
+has_cgroup_karg=1
+grep -q systemd.unified_cgroup_hierarchy /proc/cmdline || has_cgroup_karg=0
+sys_fs_cgroup_source=$(findmnt -no SOURCE /sys/fs/cgroup)
+stream=$(rpm-ostree status -b --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]')
+case "$stream" in
+    "testing-devel" | "testing" | "stable")
+        if [ $has_cgroup_karg == 0 ]; then
+            fatal "missing systemd.unified_cgroup_hierarchy=0"
+        fi
+        if [[ $sys_fs_cgroup_source != tmpfs ]]; then
+            fatal "/sys/fs/cgroup is not tmpfs"
+        fi
+        ;;
+    *)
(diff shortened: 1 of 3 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-releng-automation/pull/134

Hold until there's agreement in #292 on timing.

cgwalters:
A quick `misc-ro.sh` addition would be good here.

jlebon:
> A quick `misc-ro.sh` addition would be good here.

Yeah... it's a little tricky because the tests are *also* shared, so we need to make it conditional on the stream name, which can be confusing. Anyway, pushed something, but didn't test it. We can circle back to this when we're ready to get it in.

tests/kola/misc-ro
```diff
@@ -150,3 +150,26 @@ if ! grep prjquota <<< "${rootflags}"; then
     fatal "missing prjquota in root mount flags: ${rootflags}"
 fi
 ok "root mounted with prjquota"
+
+has_cgroup_karg=1
+grep -q systemd.unified_cgroup_hierarchy /proc/cmdline || has_cgroup_karg=0
+sys_fs_cgroup_source=$(findmnt -no SOURCE /sys/fs/cgroup)
+stream=$(rpm-ostree status -b --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]')
```
cgwalters, line 157: This is 100% completely fine by me, but we could also just
```
fmajor=$(. /usr/lib/os-release && echo ${VERSION_ID})`
if test "${fmajor}" -ge 34; then test_cgroups_v2(); else test_cgroups_v1(); fi
```
or so right?

jlebon: No, because the cgroupsv2 move won't necessarily be perfectly coupled with the f34 rebase. See https://github.com/coreos/fedora-coreos-tracker/issues/292#issuecomment-807359150.

travier (approved):

jlebon:
OK, I've verified that this passes on `next-devel` as well, so this is ready to go! (This leads to an interesting question of whether CI should include `next-devel` testing where it first promotes the changes in the PR a-la-config-bot.)

jlebon:
We probably should merge https://github.com/coreos/fedora-coreos-config/pull/931 first to be safe.

dustymabe (approved):
LGTM

PR #911, opened 2021-03-26
buildroot: Add attr

cgwalters:
ostree's tests depend on this, it's tiny and useful.


cgwalters:
(Actually I think we should split pure *build* requirements from test requirements here too...and *potentially* even consider splitting the container images.  But we can circle back to that later)

cgwalters:
/override continuous-integration/jenkins/pr-merge
not covered by this CI and can't affect the main build.

bgilbert (approved):

jlebon (approved):

PR #916, opened 2021-03-29
buildroot: add dumb-init

jlebon:
We want this, just like we had for the cosa buildroot image. At least
until this becomes built into k8s and we can just request it via the pod
manifest.


cgwalters (approved):
Looks like a smart idea.

PR #917, opened 2021-03-29
ci/buildroot: default to builder user

jlebon:
Let's match cosa's default of using an unprivileged builder user for
builds. Projects which really want to can request uid 0 via
`runAsUser:`.

Add `wheel` access to make it nicer for local development.


cgwalters (approved):
Ideally...ideally...these images would run out of the box with `toolbox`.  We don't need a static uid for our stuff.  And OpenShift already supports dynamically choosing a uid.  But, there's a `podman` shaped gap here.

It kind of argues for draining at least bits toolbox into podman.

Anyways LGTM

jlebon:
Note this was reverted in https://github.com/coreos/fedora-coreos-config/pull/932.

PR #923, opened 2021-03-31
ci/buildroot: set WORKDIR to /home/builder

jlebon:
The `builder` user doesn't have permissions to access `/root`.
Should've been part of #917.


sohankunkerkar (approved):
LGTM

arithx (approved):

PR #926, opened 2021-04-01
overlay: helper service for warning about cgroupsv1

miabbott:
This will check if a system is still using cgroupsv1 and generate a
message to be printed as part of CLHM.

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-cgroups.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-cgroups.service
@@ -1,0 +1,11 @@
+# This service is used for printing a message if
+# cgroups v1 is still being used
+[Unit]
+Description=Check if cgroupsv1 is still being used
+ConditionControlGroupController=v1
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-check-cgroups.sh
+RemainAfterExit=yes
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/bash
+# This script checks if the system is still using cgroups v1
+# and prints a message to the serial console.
+
+# Change the output color to yellow
+warn=$(echo -e '\033[0;33m')
+# No color
+nc=$(echo -e '\033[0m')
+
+motd_path=/run/motd.d/30_cgroupsv1_warning.motd
+
+cat << EOF > "${motd_path}"
+${warn}
+############################################################################
+WARNING: This system is using cgroups v1. For increased reliability
+it is strongly recommended to migrate this system and your workloads
+to use cgroups v2. For instructions on how to adjust kernel arguments
+to use cgroups v2, see:
+https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/
+
+To disable this warning, use:
+sudo systemctl disable coreos-check-cgroups.service
+############################################################################
+${nc}
+EOF
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -3,3 +3,5 @@ enable fedora-coreos-pinger.service
 # Provide information if no ignition is provided
 enable coreos-check-ignition-config.service
 enable coreos-check-ssh-keys.service
+# Check if cgroupsv1 is still being used
+enable coreos-check-cgroups.service


miabbott:
Saw the suggestion from @dustymabe in the community meeting, so decided to take a stab at something.

Bikeshedding welcome :wink:

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/bash
+# This script checks if the system is still using cgroupsv1
+# and prints a message to the serial console.
+
+# Change the output color to yellow
+warn='\033[0;33m'
+# No color
+nc='\033[0m'
+
+if grep -q systemd.unified_cgroup_hierarchy=0 /proc/cmdline; then
```
cgwalters: It'd be more efficient to do `ConditionKernelCommandLine=systemd.unified_cgroup_hierarchy=0` and have systemd do the check.

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,20 @@
...
+   echo -e "${warn}cgroups: This system is still using cgroupsv1.${nc}" > /etc/issue.d/30_cgroups_info.issue
```
cgwalters: An...issue...with this is that `/etc` is persistent - so if in response to this the admin fixes it via `rpm-ostree kargs ...`, the warning will persist.

We have a `/run/motd.d` (though that's motd vs issue, but I assume motd is fine).

Alternatively you could remove the file if the condition isn't met (though that conflicts with the above suggestion).

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,20 @@
...
+   echo -e "${warn}cgroups: Consider migrating your system + workloads to cgroupsv2.${nc}" >> /etc/issue.d/30_cgroups_info.issue
```
cgwalters: Running multiple append commands can mean that the file is only partially written; unfortunately bash doesn't expose an `O_TMPFILE` helper but it'd be a bit better to do e.g.:
```
cat > /etc/issue.d/30_cgroups_info.issue << EOF
${warn}cgroups: ...
${warn}cgroups ...
...
EOF
```

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
miabbott: Excellent suggestion, will take it!

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
miabbott: I'll try the `/run/motd.d` approach.

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
miabbott: I tried using a heredoc for this, but the text didn't get colored as expected.  I'll experiment some more with it.

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
kelvinfan001: With https://github.com/coreos/fedora-coreos-config/pull/765/commits/fa29cd6fc3e9e7eab4bc763c517e986660ee4998, MOTDs in `/run/motd.d` will be displayed upon login via console as well (in addition to via ssh), so this approach should work.

miabbott:
Pushed a new version that uses `ConditionKernelCommandLine` and a workaround for colored text via heredoc :arrow_up: 

Successfully tested locally with SSH/console login with kernel arg present/absent.

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,21 @@
...
+# See: https://stackoverflow.com/a/24701349 
+cat << EOF > /run/motd.d/30_cgroups_info.motd
+$(echo -e "${warn}cgroups: This system is still using cgroupsv1.${nc}")
+$(echo -e "${warn}cgroups: Consider migrating your system + workloads to cgroupsv2.${nc}")
+$(echo -e "${warn}cgroups: See https://github.com/coreos/fedora-coreos-tracker/issues/292 for more information.${nc}")
+EOF
+
+# Ask all running agetty instances to reload and update their
+# displayed prompts in case this script was run before agetty.
+/usr/sbin/agetty --reload
```
kelvinfan001: Now that we're using `/run/motd.d`, I think we don't need to bother `agetty` anymore, since MOTDs are taken care of by `login` (if login via console) or `pam_motd` (if login via ssh).

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-cgroups.service
```diff
@@ -0,0 +1,11 @@
+# This service is used for printing a message if
+# cgroupsv1 is still being used
+[Unit]
+Description=Check if cgroupsv1 is still being used
+ConditionKernelCommandLine=systemd.unified_cgroup_hierarchy=0
```
jlebon: Hmm, I don't think we should run this on every boot. There needs to be a way for the user to squash the warning while staying on cgroups v1. So maybe... let's use a stamp file in `/var` here + `ConditionPathExists=!...` to make sure we just run once, go back to using `/etc` for the motd, and include instructions in the motd for how to delete it (similar to [this one](https://github.com/coreos/fedora-coreos-config/blob/01c6d01cca73c87688529c3db81aadb81e830a51/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size#L28-L29)) ?

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,17 @@
...
+# No color
+nc='\033[0m'
+
+# `cat` doesn't know how to interpret escape sequences so workaround
+# it by using `echo` inside the heredoc
+# See: https://stackoverflow.com/a/24701349 
+cat << EOF > /run/motd.d/30_cgroups_info.motd
+$(echo -e "${warn}cgroups: This system is still using cgroupsv1.${nc}")
+$(echo -e "${warn}cgroups: Consider migrating your system + workloads to cgroupsv2.${nc}")
+$(echo -e "${warn}cgroups: See https://github.com/coreos/fedora-coreos-tracker/issues/292 for more information.${nc}")
```
jlebon: We have [fresh docs](https://github.com/coreos/fedora-coreos-docs/pull/269) (which will soon appear in https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/ once the site regens) for how to move to cgroups v2. Let's link to that instead?

Hmm, related to that, not sure if apart from the karg change whether there are any manual steps for e.g. migrating containers. In that case, we should probably add that to the docs too (even if it's just "you'll have to nuke all your container stores").

jlebon (commented):
Thanks for doing this!

Two comments, but idea LGTM otherwise. I think we should hold this though until we rebase to f34 because until then there's at least one good reason to still be on cgroupsv1 (moby-engine).

miabbott:
@jlebon using your suggestions, I still see the following use case:

- system boots with cgroupsv1
- service runs, writes stamp, writes `/etc` motd
- user logs in, sees warning message
- user deletes cgroupsv1 karg
- system reboots into cgroupsv2
- service doesn't run due to missing karg + stamp, but `/etc/` motd exists
- user logs in and sees warning message

This makes the move to cgroupsv2 kind of a two step process (removing karg, removing motd)

Maybe I have the logic wrong, but I wanted to understand if this is what you intended

jlebon:
> @jlebon using your suggestions, I still see the following use case:
> 
>     * system boots with cgroupsv1
> 
>     * service runs, writes stamp, writes `/etc` motd
> 
>     * user logs in, sees warning message
> 
>     * user deletes cgroupsv1 karg
> 
>     * system reboots into cgroupsv2
> 
>     * service doesn't run due to missing karg + stamp, but `/etc/` motd exists
> 
>     * user logs in and sees warning message
> 
> 
> This makes the move to cgroupsv2 kind of a two step process (removing karg, removing motd)
> 
> Maybe I have the logic wrong, but I wanted to understand if this is what you intended

Hmm yeah, I think you're correct. OK, WDYT about this: we drop the stamp file and we go back to `/run` and the instruction to disable the warning is simply `systemctl disable ...`?

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Change the output color to yellow
+warn=$(echo -e '\033[0;33m')
+# No color
+nc=$(echo -e '\033[0m')
+
+motd_path=/etc/motd.d/30_cgroupsv1_warning.motd
+
+# `cat` doesn't know how to interpret escape sequences so workaround
+# it by using `echo` inside the heredoc
```
jlebon: This is no longer true. :)

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,23 @@
...
+# See: https://stackoverflow.com/a/24701349 
+cat << EOF > "${motd_path}"
+${warn}
+cgroups: This system is still using cgroupsv1.
```
jlebon: Could we use the same warning format as in https://github.com/coreos/fedora-coreos-config/blob/01c6d01cca73c87688529c3db81aadb81e830a51/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size to be consistent? (And also makes it easier to stay within the 80 char limit).

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,23 @@
...
+cgroups: Consider migrating your system + workloads to cgroupsv2.
+cgroups: See the docs about removing kernel arguments https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/ for more information.
```
jlebon: Maybe with the URL on a newline? E.g. something like:

```suggestion
For instructions on how to modify kernel arguments for cgroupsv2, see:
https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/
```

I think we might end up having more docs for migrating. We can change the URL then.

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
miabbott: Sure, I was cribbing from `coreos-check-ignition-config.service` and `coreos-check-ssh-keys.service` before so that would explain the chosen format.

jlebon (dismissed):
LGTM, thanks a lot for this!

As mentioned, let's hold this until we actually move `testing-devel` to cgroupsv2.

miabbott:
Can we `/unhold` this now?

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-cgroups.service
dustymabe:
```suggestion
ConditionControlGroupController=v1
```

TIL about `ConditionControlGroupController`. Seems like a better check to use here. https://www.man7.org/linux/man-pages/man5/systemd.unit.5.html

overlay.d/15fcos/usr/libexec/coreos-check-cgroups.sh
```diff
@@ -0,0 +1,24 @@
...
+cat << EOF > "${motd_path}"
+${warn}
+############################################################################
+WARNING: This system is still using cgroupsv1.  It is strongly recommended
+to migrate this system and your workloads to use cgroupsv2.
+For instructions on how to remove kernel arguments for cgroupsv2, see:
+https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/
+
+To disable this warning, use:
+sudo systemctl disable coreos-check-cgroups.service
```
dustymabe:
```suggestion
WARNING: This system is using cgroups v1. For increased reliability
it is strongly recommended to migrate this system and your workloads
to use cgroups v2. For instructions on how to adjust kernel arguments
to use cgroups v2, see:
https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/

To disable this warning, use:
sudo systemctl disable coreos-check-cgroups.service
```

Optional wording enhancements. Feel free to use or ignore.

dustymabe:
> Can we `/unhold` this now?

I'd say we land it at the same time we migrate `testing` to cgroups v2 by default, which should be happening soon. Let me look into when that should be.

dustymabe:
opened a PR for moving to cgroups v2: #1033

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-cgroups.service
miabbott: Neat!  I'll give it a try.

dustymabe: I notice you did another push.. Did `ConditionControlGroupController=v1` work?

dustymabe (approved):
LGTM

PR #928, opened 2021-04-06
[next-devel] overrides: freeze dracut on 051 in f34/next

dustymabe:
Freeze dracut for now while we figure out the change in
networking causing the CI test failure.
https://github.com/coreos/fedora-coreos-tracker/issues/782


jlebon (approved):
Thanks!

jlebon (approved):

PR #929, opened 2021-04-06

dustymabe:

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
@@ -30,11 +30,24 @@ elif [[ -n "${rootfs_url}" ]]; then
         echo "Please fix your PXE configuration." >&2
         exit 1
     fi
+
+    # First, reach out to the server to verify connectivity before
+    # trying to download and pipe content through other programs.
+    # Doing this allows us to retry all errors (including transient
+    # "no route to host" errors during startup), without using the
+    # --retry-all-errors, which is problematic (see curl man page)
+    # when piping the output.
+    curl_common_args="--silent --show-error --insecure --location --retry 5"
+    if ! curl --head --retry-all-errors $curl_common_args "${rootfs_url}" >/dev/null; then
+        echo "Couldn't establish connectivity with the server specified by coreos.live.rootfs_url=" >&2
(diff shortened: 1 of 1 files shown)


dustymabe:
cc @bgilbert

cgwalters (dismissed):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
```diff
@@ -30,6 +30,18 @@ elif [[ -n "${rootfs_url}" ]]; then
...
         exit 1
     fi
+
+    # First, reach out to the server to verify connectivity before
+    # trying to download and pipe content through other programs.
+    # Doing this allows us to retry all errors (including transient
+    # "no route to host" errors during startup), without using the
+    # --retry-all-errors, which is problematic (see curl man page)
+    # when piping the output.
+    if ! curl --head --silent --show-error --insecure --location --retry 5 --retry-all-errors "${rootfs_url}"; then
```
jlebon: Would be good to be able to share the common args between the two `curl` invocations. Maybe simply:

```
curl_args="--silent --show-error --insecure --location --retry 5"
```

?

jlebon (dismissed):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
dustymabe: done

dustymabe: done

dustymabe:
This looks good in local tests.

jlebon (approved):
LGTM! ISTM though like we should just match Ignition and retry forever instead of limiting to just 5 retries. Anyway, something we can discuss and do as a follow-up.

bgilbert (commented):
LGTM

miabbott:
@dustymabe I think this broke the `pxe-install` scenario on RHCOS; the `--retry-all-errors` option doesn't appear to be available in the RHEL version of `curl`

See this internal MR - https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/1252

jlebon:
Follow-up in https://github.com/coreos/fedora-coreos-config/pull/964.

PR #930, opened 2021-04-06
buildroot: Blow out quay.io cache

cgwalters:
This is going to be problematic for the buildroot in
general.  Not finding docs on a mechanism to avoid
this on quay.


dustymabe (approved):
LGTM

dustymabe:
just for information purposes.. what issue did you hit/are we hitting?

cgwalters:
I thought we didn't have the latest ostree, but on second inspection it appears we do.

PR #931, opened 2021-04-06
[next-devel] lockfiles: bump to latest

jlebon:
Specifically, we want to make sure we have the latest runc before moving
to cgroupsv2 (see discussions in
https://github.com/coreos/fedora-coreos-tracker/issues/292).


dustymabe (approved):
LGTM

PR #932, opened 2021-04-07
buildroot: Drop USER creation, add createrepo_c+ostree

cgwalters:
The USER bit breaks the ability to use this in Github actions.
Further, it's not useful for OpenShift because the platform
allocates a dynamic UID today (Eventually Kube user namespaces
will augment this).

We added this for the Jenkins/CCI, but previously we
were running the buildroot as uid 0 (i.e. Docker default) anyways.
If we want to enhance the flow to run as non-root, I think the
right thing is to do it in the platform and not the container.

### DIFF
--- a/ci/buildroot/buildroot-reqs.txt
+++ b/ci/buildroot/buildroot-reqs.txt
@@ -28,6 +28,9 @@ xz
 # For rust projects like rpm-ostree
 rustfmt
 
+# For unit tests at least.
+ostree
+
 # A super common tool
 jq
 
@@ -36,6 +39,7 @@ attr
 rsync
 python3-pyyaml
 parallel gjs
+createrepo_c
 
 # Also, add clang since it's useful at least in CI for C/C++ projects
 clang lld
--- a/ci/buildroot/Dockerfile
+++ b/ci/buildroot/Dockerfile
@@ -6,13 +6,5 @@
 # This image is used by CoreOS CI to build software like
 # Ignition, rpm-ostree, ostree, coreos-installer, etc...
 FROM registry.fedoraproject.org/fedora:33
-USER root
-WORKDIR /root/containerbuild
-COPY . tmp
-RUN ./tmp/install-buildroot.sh && yum clean all && rm tmp -rf # nocache 20210406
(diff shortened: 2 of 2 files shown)


cgwalters:
I pushed a build of this to quay.io/cgwalters/fcos-buildroot and it successfully works in GH actions here:
https://github.com/cgwalters/ostree-ext/pull/7

zincati today for example is shipping a variant of this https://github.com/coreos/zincati/blob/master/contrib/Dockerfile.dev
and building it on each PR:
https://github.com/coreos/zincati/blob/master/.github/workflows/containers.yml

I think having the same container image work in GH actions, Jenkins and OpenShift is important, so the argument here is to change Jenkins.

cgwalters:
(And yes, I think we should probably drop the `USER` bits from cosa too)

ci/buildroot/buildroot-reqs.txt
```diff
@@ -28,6 +28,9 @@ xz
 # For rust projects like rpm-ostree
 rustfmt
 
+# For unit tests at least.
+ostree
```
jlebon, line 32: Hmm, this is OK but we should make sure to `rpm -e` in the ostree CI because having bits pre-installed can lead to really subtle bugs.

jlebon (approved):
Makes sense to me. My initial motivation for https://github.com/coreos/fedora-coreos-config/pull/917 was really just because some tests want an unprivileged user (e.g. https://github.com/coreos/ssh-key-dir/commit/1ca2ab63d8cb5933e97b6feaa80bd5e4fb6f84c1), but OTOH other tests want root. So meh, might as well default to root and let those projects set up the unpriv user themselves if needed.

Being able to use it in GHA is totally worth it.

jlebon:
CI flaked, and doesn't cover this anyway so overriding.

ci/buildroot/buildroot-reqs.txt
cgwalters: Yeah, agree.  Another alternative is split into say a "buildroot-core" and 'build" image that also covers unit tests that derives from it.

PR #935, opened 2021-04-07
35coreos-network: enable initrd networking on azurestack

jlebon:
Just like on Azure, Afterburn needs networking for hostname fetching and
boot check-in. Ideally, Afterburn would request it itself but this is
where that logic lives for now (see [1]).

[1] https://github.com/coreos/fedora-coreos-tracker/blob/master/internals/README-initramfs.md#networking

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.service
@@ -7,10 +7,11 @@ After=basic.target
 # Triggering conditions for cases where we need network:
 #  * when Ignition signals that it is required for provisioning.
 #  * on live systems fetching the remote rootfs in initramfs.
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
Hmm, another rootfs flake?

```
[   23.393492] ignition-ostree-transposefs[520]: mount: /sysroot: special device /dev/disk/by-label/root does not exist.
[[0;1;31mFAILED[0m] Failed to start [0;1;39mIgnition OSTree: Restore Partitions[0m.
```

Anyways this can't be related, and by merging this we'll all be full stack engineers so, merging!

jlebon:
Cross-posting from https://github.com/coreos/afterburn/pull/561#issuecomment-816752604:

One thing worth noting is that in the base case, ASH support also requires coreos/fedora-coreos-config#935 so it has networking in the initrd. For OKD/OCP it works without it, because we use pointer Ignition configs and so Ignition itself needs networking and requests it.

PR #936, opened 2021-04-07
12kdump: Remount /boot rw before `kdump.service`

kelvinfan001:
Right now, `kdump` fails to generate its initramfs because
it cannot place it next to our default initramfs in
`/boot/ostree`, which is read-only.
This is a workaround to get `kdump` working again until
it gains the ability to place its generated initramfs
in a different location (e.g. `/var/lib/kdump`).

xref: https://bugzilla.redhat.com/show_bug.cgi?id=1918493

### DIFF
--- a/overlay.d/12kdump/usr/lib/systemd/system/kdump.service.d/remount-boot.conf
+++ b/overlay.d/12kdump/usr/lib/systemd/system/kdump.service.d/remount-boot.conf
@@ -1,0 +1,9 @@
+# https://bugzilla.redhat.com/show_bug.cgi?id=1918493
+# `/boot` is read-only, but `kdump.service` wants to 
+# places its generated initramfs alongside the default
+# initramfs under `/boot/ostree`.
+# Until `kdump` gains the ability to place its initramfs
+# elsewhere, temporarily remount `/boot` read-write before
+# the `kdump` initramfs is generated.
+[Service]
+ExecStartPre=/usr/bin/mount -o remount,rw /boot


kelvinfan001:
I have submitted a patch to the Fedora kexec mailing list to get kdump to place its initramfs in `/var/lib/kdump` instead of `/boot`. 
After that, the default SELinux policies need to be updated to allow kexec to read from `/var/lib/kdump`, as well.

kelvinfan001:
Please let me know if this systemd drop in for `kdump.service` should be placed somewhere else in `overlay.d/`.

cgwalters (commented):
Ahhh nice!  I hadn't even thought of using `ExecStartPost=` to remount it again.

But...hmm.  This can be racy because if any other process happens to get a writable fd there we'll fail to remount it read only.

And ostree has logic today around remounting `/boot` writable in a private namespace; I am not completely sure what will happen in the combinations involved here (e.g. depending on the mount ordering).

It may be safest to leave off the `ExecStartPost=` and just leave it writable.  But, I'm also fine keeping it.

cgwalters (dismissed):

cgwalters:
> Please let me know if this systemd drop in for kdump.service should be placed somewhere else in overlay.d/.

Looks correct to me!

kelvinfan001:
>  This can be racy because if any other process happens to get a writable fd there we'll fail to remount it read only.

ooh, thanks for pointing that out.
Actually, instead of leaving `/boot` rw, could we get `kdump.service` to mount `/boot` rw at a random/not well-known location so that it is less likely that another process woud get a writable fd there?

cgwalters:
> Actually, instead of leaving /boot rw, could we get kdump.service to mount /boot rw at a random/not well-known location so that it is less likely that another process woud get a writable fd there?

Right, the pattern we use in several places basically adding e.g.  `MountFlags=slave` to the unit - then kdump could mount it writable just in its own namespace.  This approach might work equally well; presumably kdump doesn't need to persistent change the host mounts.

kelvinfan001:
Ok, I tinkered around with the second approach, but I encountered two problems. I _think_ `MountFlags=slave` applies to every process started by the unit, including the one started in `ExecStartPre=`, so seems like we can't do the mounting there, since the actual `kdump` process won't have access to mounts mounted in `ExecStartPre=`. Another thing is, even without `MountFlags=slave`, I somehow wasn't able to mount the `boot` device read-write at a different location, e.g. `/var/mnt/kdump-boot` via the `kdump` unit, as it says the source is protected and it can only be mounted ro; however, doing the same outside the unit, i.e. manually in the shell, works just fine.
If it's ok, I'll just take the original approach, but skip the `ExecStartPost=` bit and leave `/boot` rw on nodes where `kdump` is enabled.

cgwalters (approved):

PR #937, opened 2021-04-07
tests/kola: handle possible rootless-systemd flake

dustymabe:
The sleep 5 might not be long enough if CI is loaded. Let's make curl
itself do the retry, which contains logic for exponential backoff.

### DIFF
--- a/tests/kola/podman/rootless-systemd
+++ b/tests/kola/podman/rootless-systemd
@@ -53,10 +53,8 @@ main() {
     chmod +x /tmp/runascoreuserscript
     runascoreuser /tmp/runascoreuserscript
 
-    # Let it come up
-    sleep 5
-
-    if ! curl http://localhost:8080 1>/dev/null; then
+    # Try to grab the web page. Retry as it might not be up fully yet.
+    if ! curl --silent --show-error --retry 4 --retry-all-errors http://localhost:8080 >/dev/null; then
         echo TEST FAILED 1>&2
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
I bet some sort of analysis on how many git commits contain the text "race" and modify a line of code that contains `sleep` would be over 50%.

tests/kola/podman/rootless-systemd
```diff
@@ -53,10 +53,8 @@ main() {
     chmod +x /tmp/runascoreuserscript
     runascoreuser /tmp/runascoreuserscript
 
-    # Let it come up
-    sleep 5
-
-    if ! curl http://localhost:8080 1>/dev/null; then
+    # Try to grab the web page. Retry as it might not be up fully yet.
+    if ! curl --silent --show-error --retry 4 --retry-all-errors http://localhost:8080 >/dev/null; then
```
cgwalters, line 57: But shouldn't we still sleep a bit to give the container a chance to come up at all?  Otherwise we'll quickly start the exponential backoff.

dustymabe: Sorry.. I had enabled auto merge.. Perhaps I should have left some sleep in there.

jlebon: Follow-up in https://github.com/coreos/fedora-coreos-config/pull/948.

PR #938, opened 2021-04-09
35coreos-ignition: add coreos-kargs

arithx:
Marking as draft until an Ignition release & coreos-installer release have made it to `testing-devel`.

Based on content from these PRs:
https://github.com/coreos/ignition/pull/1178
https://github.com/coreos/coreos-installer/pull/498


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
```diff
@@ -0,0 +1,11 @@
...
+set -euo pipefail
+
+# Mount /boot. Note that we mount /boot but we don't unmount it because we
+# are run in a systemd unit with MountFlags=slave so it is unmounted for us.
+bootmnt=/mnt/boot_partition
+mkdir -p ${bootmnt}
+bootdev=/dev/disk/by-label/boot
+mount -o rw ${bootdev} ${bootmnt}
+
+/usr/bin/rdcore kargs --boot-mount ${bootmnt} --create-if-changed /run/ignition-modified-kargs "$@"
```
jlebon: Note `rdcore kargs` also has a `--boot-device` to let it mount the device itself. Yay less bash!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.service
```diff
@@ -0,0 +1,22 @@
+[Unit]
```
jlebon: Shouldn't this unit be part of https://github.com/coreos/ignition/pull/1178?

arithx: Because there's a required distribution binary implementation (`rdcore` for FCOS/RHCOS) to modify the kargs, Ignition is only shipping an example unit and leaves it up to the distributor to actually ship a unit running the stage.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
arithx: Ah, I had just followed what I had seen other pre-existing units doing but that makes sense as we're not actually touching anything with this script.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.service
jlebon: But this unit is agnostic of the actual binary used to handle the kargs since it just calls out to `ignition`. Or are we saying that even e.g. the fact that it runs between the `fetch` and `disks` stages is up to the distro?

arithx: So the PR itself currently recommends in the example service to put it between the `fetch` and `disks` stages but it's not a hard requirement.

Additionally the example unit in the Ignition repo isn't installed (nor do I think it makes sense to given we aren't shipping an actual distribution binary, just an example script, and the kargs stage wouldn't work without it).

jlebon: I think at a maintenance level, this service unit would belong better in the Ignition package alongside the other stage service units. I think also it's fine if Ignition is opinionated about the service running right after the `fetch` stage. Given that it requires distro tooling anyway, in the end it's still completely in the distro's control when to act on the output of that stage.

Here's an idea: we have upstream Ignition default to `kargsCmd = ""` and the `ignition-kargs.service` simply no-ops if it wasn't configured. WDYT? This makes it naturally backwards compatible.

arithx: I don't think having the `kargsCmd = ""` would actually work because we'd still be doing an exec on `["", "--should-exist", "my args"]` which is going to fail.

Additionally having the kargs stage still run and just no-op feels like it could lead to some confusion when users are reading journal logs on distributions that don't provide the kargs script.

Finally, that would mean that each distribution now has a required override if they want to support kargs. This means that if say both FCOS & Fedora IoT want to support Ignition kargs but chose a different distribution binary there'd need to be two different builds of Ignition in Fedora (with different override build flags on `kargsCmd`) whereas right now they can just place their binary / a shell wrapper at the default location.

cc @bgilbert

jlebon: > I don't think having the `kargsCmd = ""` would actually work because we'd still be doing an exec on `["", "--should-exist", "my args"]` which is going to fail.

We would explicitly detect this case and exit early.

> Additionally having the kargs stage still run and just no-op feels like it could lead to some confusion when users are reading journal logs on distributions that don't provide the kargs script.

I think we could just log something, or if we really want to, we can make the installation of the unit conditional on the build flag.

> Finally, that would mean that each distribution now has a required override if they want to support kargs. This means that if say both FCOS & Fedora IoT want to support Ignition kargs but chose a different distribution binary there'd need to be two different builds of Ignition in Fedora (with different override build flags on `kargsCmd`) whereas right now they can just place their binary / a shell wrapper at the default location.

I might be missing something here, this service unit doesn't actually need to change if the distro-specific kargs command changes, right?

Re. IoT, I think in the Fedora dist-git of Ignition, we can ship a dummy no-op command at some well-known path, and it's up to FCOS/IoT to override that path.

jlebon: We discussed this out of band and there was agreement to move this unit to be owned by Ignition and to provide a mechanism for allowing distros to fill in the missing distro-specific functionality or otherwise error out.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs-reboot.service
```diff
@@ -0,0 +1,22 @@
+[Unit]
+Description=Ignition (kargs) Reboot
```
jlebon: How about something like `CoreOS Kernel Arguments Reboot`?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -40,6 +40,13 @@ install() {
     # path generated by systemd-escape --path /dev/disk/by-label/root
     install_ignition_unit coreos-gpt-setup.service ignition-diskful.target
 
+    # dracut inst_script doesn't allow overwrites and we are replacing
+    # the default script placed by Ignition
+    binpath="/usr/bin/ignition-kargs-helper"
+    mv "$moddir/coreos-kargs.sh" \
+        "$initdir$binpath"
```
jlebon: Super minor: feels like this is short enough already that we don't need a line break. :)

jlebon (dismissed):
Some bikesheddy nits, but LGTM overall!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -40,6 +40,13 @@ install() {
     # path generated by systemd-escape --path /dev/disk/by-label/root
     install_ignition_unit coreos-gpt-setup.service ignition-diskful.target
 
+    # dracut inst_script doesn't allow overwrites and we are replacing
+    # the default script placed by Ignition
+    binpath="/usr/bin/ignition-kargs-helper"
```
bgilbert: `/usr/sbin` now.

arithx:
Rebased on top of https://github.com/coreos/fedora-coreos-config/pull/1009

jlebon (dismissed):
Awesome! :tada:

jlebon:
Were you thinking of enhancing kola to allow setting kargs from external tests? Or just have internal tests in kola for now?

arithx:
> Were you thinking of enhancing kola to allow setting kargs from external tests? Or just have internal tests in kola for now?

Are you meaning kargs set via the harness (e.x. set via the qemu cmdline args) specifically for testing this feature? If so my plan was to just have it be an internal test, any other test that wants kargs can just add it to the Ignition config (as that's the user-facing happy path we are trying to enable here).

jlebon:
> Are you meaning kargs set via the harness (e.x. set via the qemu cmdline args) specifically for testing this feature? If so my plan was to just have it be an internal test, any other test that wants kargs can just add it to the Ignition config (as that's the user-facing happy path we are trying to enable here).

Heh right sorry, I got mixed up. Obviously, since this is in the Ignition config now (that was the whole point!) it shouldn't need any more work in the kola harness for external tests. So adding a test for this here should be pretty trivial. Something like:

```diff
diff --git a/tests/kola/ignition/kargs/config.ign b/tests/kola/ignition/kargs/config.ign
new file mode 100644
index 0000000..3d6d859
--- /dev/null
+++ b/tests/kola/ignition/kargs/config.ign
@@ -0,0 +1,9 @@
+{
+  "ignition": {
+    "version": "3.3.0-experimental"
+  },
+  "kernelArguments": {
+    "shouldExist": "foobar",
+    "shouldNotExist": "mitigations=auto,nosmt"
+  }
+}
diff --git a/tests/kola/ignition/kargs/test.sh b/tests/kola/ignition/kargs/test.sh
new file mode 100755
index 0000000..a753525
--- /dev/null
+++ b/tests/kola/ignition/kargs/test.sh
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if ! grep foobar /proc/cmdline; then
+    fatal "missing foobar in kernel cmdline"
+fi
+if grep mitigations /proc/cmdline; then
+    fatal "found mitigations in kernel cmdline"
+fi
+ok "Ignition kargs"
```

WDYT?

Cool with an internal test if you prefer, though we're trying to move towards external tests when possible. A nice advantage for example is that we can then test the code added here in CI before merging. But again, feel free to self-merge if you want to do this in kola proper or as a follow-up.

arithx:
> Cool with an internal test if you prefer, though we're trying to move towards external tests when possible. A nice advantage for example is that we can then test the code added here in CI before merging. But again, feel free to self-merge if you want to do this in kola proper or as a follow-up.

My thought on an internal test was so that we could also test the `shouldNotExist` by adding a karg from the harness side and removing it w/ Ignition but the `nosmt` one would work.

arithx:
Went ahead and pushed with your patch (and added you as a co-author of the commit).

jlebon:
```
[2021-05-14T19:33:52.998Z] 2021-05-14T19:33:52Z platform/conf: invalid userdata: error at line 6 col 28: json: cannot unmarshal string into Go struct field KernelArguments.kernelArguments.shouldExist of type []types.KernelArgument
[2021-05-14T19:33:52.998Z] Error: Parsing config.ign: config is not valid
[2021-05-14T19:33:52.998Z] 2021-05-14T19:33:52Z cli: Parsing config.ign: config is not valid
```

Ahh right OK, this does still require teaching the new spec to kola [like we did here](https://github.com/coreos/coreos-assembler/pull/1592). I think we'll need that anyway eventually, though OK to not block on that.

arithx:
> ```
> [2021-05-14T19:33:52.998Z] 2021-05-14T19:33:52Z platform/conf: invalid userdata: error at line 6 col 28: json: cannot unmarshal string into Go struct field KernelArguments.kernelArguments.shouldExist of type []types.KernelArgument
> [2021-05-14T19:33:52.998Z] Error: Parsing config.ign: config is not valid
> [2021-05-14T19:33:52.998Z] 2021-05-14T19:33:52Z cli: Parsing config.ign: config is not valid
> ```
> 
> Ahh right OK, this does still require teaching the new spec to kola [like we did here](https://github.com/coreos/coreos-assembler/pull/1592). I think we'll need that anyway eventually, though OK to not block on that.

That was actually an error in the tests Ignition config (it was passing a string to `shouldExist`/`shouldNotExist` instead of a list of strings).

Ignition 2.10.1 was vendored into mantle in https://github.com/coreos/coreos-assembler/pull/2165

jlebon (approved):
> [2021-05-14T20:38:04.936Z] --- PASS: ext.config.ignition.kargs (37.22s)

Awesome! :tada:

lucab:
I'm not exactly sure what's going on, but I'm now seeing the following in FCOS build logs:
```
dracut: *** Including module: coreos-ignition ***
mv: cannot remove '/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh': Read-only file system
```

jlebon:
Thanks, follow-up in https://github.com/coreos/fedora-coreos-config/pull/1076.

PR #941, opened 2021-04-09
manifests: f34: Delete coreos-reset-stub-resolv-selinux-context.{path,service}.

dustymabe:
Not needed in Fedora 34 since https://github.com/systemd/systemd/pull/17976 has landed.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -99,6 +99,18 @@ postprocess:
     DNSStubListener=no
     EOF
 
+  # Disable and delete the coreos-reset-stub-resolv-selinux-context.{path,service}.
+  # Not needed in Fedora 34 since https://github.com/systemd/systemd/pull/17976 has landed.
+  # Can remove this and the files in the overlay once we've migrated everything to F34.
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    source /etc/os-release
+    if [ ${VERSION_ID} -ge 34 ]; then
+    rm -fv /etc/systemd/system/multi-user.target.wants/coreos-reset-stub-resolv-selinux-context.{path,service} \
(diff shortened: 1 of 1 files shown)


dustymabe:
This should allow us to unquestionably close out https://github.com/coreos/fedora-coreos-tracker/issues/777

jlebon (approved):

manifests/fedora-coreos-base.yaml
```diff
@@ -99,6 +99,18 @@ postprocess:
...
+  # Disable and delete the coreos-reset-stub-resolv-selinux-context.{path,service}.
+  # Not needed in Fedora 34 since https://github.com/systemd/systemd/pull/17976 has landed.
+  # Can remove this and the files in the overlay once we've migrated everything to F34.
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    source /etc/os-release
+    if [ ${VERSION_ID} -ge 34 ]; then
+    rm -fv /etc/systemd/system/multi-user.target.wants/coreos-reset-stub-resolv-selinux-context.{path,service} \
+           /usr/lib/systemd/system/coreos-reset-stub-resolv-selinux-context.{path,service}
```
lucab, line 111: I guess this "remove content from /usr" is spurious?

lucab: Nevermind, I didn't realize this was a quick hack, the bits in our overlay are staying a bit longer and will be properly removed after rebase.

dustymabe: I delete both the systemd enablement and the bits from /usr, just in case the presets get evaluated again (it won't get re-enabled). We're dropping these bits in a postprocess conditional just because it's easier to have it propagate to all streams (branches) that way. Once everything is on f34, we'll drop this *and* the overlay.d bits.

PR #943, opened 2021-04-10
overrides: fast-track afterburn release

sohankunkerkar:


jlebon (approved):

jlebon:
Can you fast-track it to next-devel as well?

jlebon:
In general, `next-devel` RPMs should be versioned greater or equal to `testing-devel`.

jlebon:
Ahh I see https://github.com/coreos/fedora-coreos-config/pull/944, thanks!

PR #946, opened 2021-04-12
[testing] tree: promote changes from testing-devel at 24c126f4f67c5691684216870a3b3b1c7c54ed86

coreosbot-releng:
Triggered by @lucab in https://github.com/coreos/fedora-coreos-streams/issues/288.

PR #948, opened 2021-04-12
tests/kola/rootless-systemd: add back sleep

jlebon:
Let's both sleep upfront and have `curl` do its exponential retry.
Follow-up to #937 since I saw this test failing still in #946.


dustymabe (approved):
LGTM

PR #951, opened 2021-04-12
tests/kola/rpm-ostree-countme: Update for pre-release repo count

travier:
In production, we allow counting to silently fail when partially
reporting count me status to avoid unnecessary unit failures as this not
critical. In CI however, we want to make sure this test is properly
working so we need to manually validate the output.

FCOS versions based on pre-release Fedora (branched/beta/rc) will have
an additionnal repo enabled, so this updates the check to take that into
account.

For: https://github.com/coreos/fedora-coreos-tracker/issues/790


dustymabe (approved):
passes my tests running against f34 (`next`)

PR #954, opened 2021-04-12
[stable] overrides: fast-track rpm-ostree-2021.4

jlebon:
For CVE-2021-3445 (https://bugzilla.redhat.com/show_bug.cgi?id=1932079).

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -44,3 +44,9 @@ packages:
     evr: 1:1.1.1k-1.fc33
   openssl-libs:
     evr: 1:1.1.1k-1.fc33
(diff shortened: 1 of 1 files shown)


dustymabe:
https://github.com/coreos/fedora-coreos-streams/issues/287#issuecomment-818363508

jlebon:
OK, let's not do this one to be safe. The statement in https://access.redhat.com/security/cve/cve-2021-3445 is pretty helpful in evaluating how serious this is:

> In addition, to exploit this flaw, an attacker must either compromise an RPM repository or convince an administrator to install an untrusted RPM. It is strongly recommended to only use RPMs from trusted repositories.

PR #957, opened 2021-04-13
Fix chrony misconfiguration with leap-smearing servers

mlichvar:
NTP servers provided by AWS and GCP implement a "leap smear" to
hide leap seconds from clients. When chronyd is configured with
leap-smearing servers, the leapsectz directive must be disabled to
avoid the kernel making a wrong correction of the clock at the time of
the leap second.


lucab:
@mlichvar thanks for the PR! Do you maybe have a back-reference to cross-link where this is coming from? Was this misbehavior observed in the field, reported to some issues tracker, or just casually noticed on some node at some specific FCOS version?

mlichvar:
I noticed this issue when I was looking at the chrony support in this repository to be sure it won't break with some changes I'm preparing in the Fedora chrony package.

There was no leap second since 2016, so this issue didn't have a chance to show up in real world yet.

Here are some sources:
https://aws.amazon.com/about-aws/whats-new/2017/11/introducing-the-amazon-time-sync-service/
https://developers.google.com/time/smear
https://chrony.tuxfamily.org/faq.html#_should_be_a_leap_smear_enabled_on_ntp_server

lucab:
Thanks for the proactive action then, LGTM!

lucab (approved):

PR #960, opened 2021-04-14
Enable IMA by default

cgwalters:
Co-authored-by: Colin Walters <walters@verbum.org>

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -7,3 +7,5 @@ extra-kargs:
     # https://github.com/coreos/fedora-coreos-tracker/issues/292
     # https://fedoraproject.org/wiki/Changes/CGroupsV2
     - systemd.unified_cgroup_hierarchy=0
+    # See https://sourceforge.net/p/linux-ima/wiki/Home/#enabling-ima-measurement
+    - ima_template=ima-sig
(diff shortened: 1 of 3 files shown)


cgwalters:
Derived from code here https://github.com/stefanberger/fedora-coreos-config/commit/171ff4dc0f2c1304c57af06b8e1f0fc59f1145e4 written by @stefanberger

jlebon:
Should we have an IMA tracker issue at this point where we can link all those PRs and discuss whether we want to do this specifically in FCOS and potential design/UX issues?

image.yaml
```diff
@@ -7,3 +7,6 @@ extra-kargs:
     # https://github.com/coreos/fedora-coreos-tracker/issues/292
     # https://fedoraproject.org/wiki/Changes/CGroupsV2
     - systemd.unified_cgroup_hierarchy=0
+    # See https://sourceforge.net/p/linux-ima/wiki/Home/#enabling-ima-measurement
+    - ima_template=ima-sig
+    - ima_policy=appraise_tcb
```
stefanberger: You may need to build in a key into the kernel and use a certified signing key for this to work. I would start with a simple policy first only enforcing on ping for example. Makes life easier.

cgwalters:
> Should we have an IMA tracker issue at this point where we can link all those PRs and discuss whether we want to do this specifically in FCOS and potential design/UX issues?

I'm mostly using this draft PR as a place to share discussion around a basic "demo" of this.  It's definitely not clear to me we want to do this by default in FCOS.  But rpm-ostree upstream does.

PR #962, opened 2021-04-14
Added Kubernetes-Master Package and Etcd in the distribution

Xfireball8:


lucab:
Thanks for the PR, but those are both applications/services which don't strictly belong to the OS, thus I'm turning this PR down.
For the future, we'd appreciate upfront discussion before adding things to the base OS, the proper flow for that is https://github.com/coreos/fedora-coreos-tracker#adding-packages-to-fedora-coreos.

PR #964, opened 2021-04-15

jlebon:

`--retry-all-errors`.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
@@ -34,20 +34,24 @@ elif [[ -n "${rootfs_url}" ]]; then
     # First, reach out to the server to verify connectivity before
     # trying to download and pipe content through other programs.
     # Doing this allows us to retry all errors (including transient
-    # "no route to host" errors during startup), without using the
-    # --retry-all-errors, which is problematic (see curl man page)
-    # when piping the output.
-    curl_common_args="--silent --show-error --insecure --location --retry 5"
-    if ! curl --head --retry-all-errors $curl_common_args "${rootfs_url}" >/dev/null; then
+    # "no route to host" errors during startup). Note we can't use
+    # curl's --retry-all-errors here because it's not in el8's curl yet.
+    # We retry forever, matching Ignition's semantics.
+    curl_common_args="--silent --show-error --insecure --location"
+    while ! curl --head $curl_common_args "${rootfs_url}" >/dev/null; do
         echo "Couldn't establish connectivity with the server specified by coreos.live.rootfs_url=" >&2
-        echo "Check that the URL is correct and can be reached." >&2
-        exit 1
-    fi
+        echo "Retrying in 5s..." >&2
+        sleep 5
+    done
+
     # We don't need to verify TLS certificates because we're checking the
     # image hash.
     # bsdtar can read cpio archives and we already depend on it for
     # coreos-liveiso-persist-osmet.service, so use it instead of cpio.
-    if ! curl $curl_common_args "${rootfs_url}" | \
+    # We shouldn't need a --retry here since we've just successfully HEADed the
+    # file, but let's add one just to be safe (e.g. if the connection just went
+    # online and flickers or something).
+    if ! curl $curl_common_args --retry 5 "${rootfs_url}" | \
             rdcore stream-hash /etc/coreos-live-want-rootfs | \
             bsdtar -xf - -C / ; then
         echo "Couldn't fetch, verify, and unpack image specified by coreos.live.rootfs_url=" >&2


jlebon:
Still need to test this.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
```diff
@@ -34,15 +34,17 @@ elif [[ -n "${rootfs_url}" ]]; then
...
+    # We retry forever, matching Ignition's semantics.
     curl_common_args="--silent --show-error --insecure --location --retry 5"
-    if ! curl --head --retry-all-errors $curl_common_args "${rootfs_url}" >/dev/null; then
+    while ! curl --head $curl_common_args "${rootfs_url}" >/dev/null; do
         echo "Couldn't establish connectivity with the server specified by coreos.live.rootfs_url=" >&2
         echo "Check that the URL is correct and can be reached." >&2
-        exit 1
-    fi
+        echo "Retrying..." >&2
+        sleep 1 # note this sleep is between curl's exponential backoffs
```
dustymabe: > sleep 1 # note this sleep is between curl's exponential backoffs

Yes and no. Since we can't use `--retry-all-errors` the curl retry won't retry `no route to host` errors, which is what this was put in place to check. So we'll just fail on the first try and drop directly to `sleep 1`,  before continuing. I think what we probably need to do is just retry every 10s or so and drop `--retry 5`.

jlebon:
OK, tweaked and tested now!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
jlebon: Yeah, just cleaner to drop the `--retry` and not have to think about nested retries.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
```diff
@@ -34,20 +34,25 @@ elif [[ -n "${rootfs_url}" ]]; then
...
+    # "no route to host" errors during startup). Note we can't use
+    # curl's --retry-all-errors here because it's not in el8's curl yet.
+    # We retry forever, matching Ignition's semantics.
+    curl_common_args="--silent --show-error --insecure --location"
+    while ! curl --head $curl_common_args "${rootfs_url}" >/dev/null; do
         echo "Couldn't establish connectivity with the server specified by coreos.live.rootfs_url=" >&2
         echo "Check that the URL is correct and can be reached." >&2
-        exit 1
-    fi
+        echo "Retrying in 5s..." >&2
```
dustymabe: How does this error message look on the CLI when hitting failures? Should we try to do some magic and squeeze/trim so we only output one line of error?

dustymabe (dismissed):
LGTM

I am interested in @bgilbert's thoughts on retrying forever, though.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
jlebon: Here's a sample screenshot (this is from before the last tweaks):

Screenshot from 2021-04-15 16-37-07-cropped

jlebon: > Should we try to do some magic and squeeze/trim so we only output one line of error?

Meh, seems OK to me as is, but happy to trim if folks prefer!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
```diff
@@ -34,20 +34,25 @@ elif [[ -n "${rootfs_url}" ]]; then
...
-    # when piping the output.
-    curl_common_args="--silent --show-error --insecure --location --retry 5"
-    if ! curl --head --retry-all-errors $curl_common_args "${rootfs_url}" >/dev/null; then
+    # "no route to host" errors during startup). Note we can't use
+    # curl's --retry-all-errors here because it's not in el8's curl yet.
+    # We retry forever, matching Ignition's semantics.
+    curl_common_args="--silent --show-error --insecure --location"
+    while ! curl --head $curl_common_args "${rootfs_url}" >/dev/null; do
         echo "Couldn't establish connectivity with the server specified by coreos.live.rootfs_url=" >&2
         echo "Check that the URL is correct and can be reached." >&2
```
bgilbert: Maybe we should drop this line?  A couple failures will be normal if the network is slow, and the specific call to action won't apply in that case.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
bgilbert: I'm not particularly concerned about the multiple lines myself.

bgilbert (dismissed):
I think it makes sense to retry forever.  We'd just fail outright otherwise.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
dustymabe: Yeah. Looks better than I expected. 👍

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
jlebon: Good point!

dustymabe (approved):
LGTM

PR #965, opened 2021-04-15
ci: add GitHub Action to automatically graduate overrides

jlebon:
Currently, we have to manually check if fast-tracked packages have made
it into the Fedora repos and drop them. This GitHub Action automates
this process.

It also formalizes the process of pinning packages a bit by adding
guidelines for metadata keys overrides should use to be more explicit.

### DIFF
--- a/.github/workflows/remove-graduated-overrides.yml
+++ b/.github/workflows/remove-graduated-overrides.yml
@@ -1,0 +1,45 @@
+name: remove-graduated-overrides
+
+on:
+  schedule:
+    - cron: '0 */6 * * *'
+
+jobs:
+  remove-graduated-overrides:
+    name: Remove graduated overrides
+    runs-on: ubuntu-latest
+    # TODO: use cosa directly here
+    # https://github.com/coreos/coreos-assembler/issues/2223
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
+    strategy:
+      matrix:
+        branch:
+          - testing-devel
+          - next-devel
+    steps:
+      - run: dnf install -y rpm-ostree # see related TODO above
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ matrix.branch }}
+      - name: Remove graduated overrides
+        run: |
+          git config user.name 'CoreOS Bot'
+          git config user.email coreosbot@fedoraproject.org
+          ci/remove-graduated-overrides.py
+      - name: Open pull request
+        run: |
+          if ! git diff --quiet --exit-code; then
+              git commit -am "lockfiles: drop graduated overrides 🎓" \
+                -m "Triggered by remove-graduated-overrides GitHub Action."
+          fi
+      - name: Open pull request
+        uses: peter-evans/create-pull-request@v3.8.2
+        with:
+          token: ${{ secrets.COREOSBOT_RELENG_TOKEN }}
+          branch: ${{ matrix.branch }}-graduation
+          push-to-fork: coreosbot-releng/fedora-coreos-config
+          title: "lockfiles: drop graduated overrides 🎓"
+          body: "Triggered by remove-graduated-overrides GitHub Action."
+          committer: "CoreOS Bot <coreosbot@fedoraproject.org>"
+          author: "CoreOS Bot <coreosbot@fedoraproject.org>"
--- a/ci/remove-graduated-overrides.py
+++ b/ci/remove-graduated-overrides.py
@@ -1,0 +1,125 @@
+#!/usr/bin/python3
+
+import os
+import sys
+import json
+import yaml
+import subprocess
+
+import dnf
+import hawkey
+
+ARCHES = ['s390x', 'x86_64', 'ppc64le', 'aarch64']
+
+OVERRIDES_HEADER = """
+# This lockfile should be used to pin to a package version (`type: pin`) or to
+# fast-track packages ahead of Bodhi (`type: fast-track`). Fast-tracked
+# packages will automatically be removed once they are in the stable repos.
+#
+# IMPORTANT: YAML comments *will not* be preserved. All `pin` overrides *must*
+# include a URL in the `metadata.reason` key. Overrides of type `fast-track`
+# *should* include a URL in the `metadata.reason` key, though it's acceptable to
+# omit one for FCOS-specific packages (e.g. ignition, afterburn, etc...).
+"""
+
+
+def main():
+    treefile = get_treefile()
+    base = get_dnf_base(treefile)
+    setup_repos(base, treefile)
+
+    for fn in get_lockfiles():
+        update_lockfile(base, fn)
+
+
+def get_treefile():
+    treefile = subprocess.check_output(['rpm-ostree', 'compose', 'tree',
+                                        '--print-only', 'manifest.yaml'])
+    return json.loads(treefile)
+
+
+def get_dnf_base(treefile):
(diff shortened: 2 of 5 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/2131

jlebon:
(As usual, this would also require plumbing through coreos-koji-tagger.)

manifest-lock.fast-track.yaml
```diff
@@ -0,0 +1,19 @@
+# This lockfile should be used for fast-tracking packages into FCOS ahead of
+# Bodhi. The intention is that once that package or newer makes it into the
+# stable repos, the pin should be removed. Pins are automatically removed when
+# no longer necessary. Note that YAML comments will not be preserved.
+# Rationales should be kept in the commit messages.
```
jlebon: I tried to find a way to preserve comments using [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) instead, but couldn't get it to do the right thing.

I think if we really want to we could add a `metadata` key to the lockfile spec under package name keys where you can put arbitrary data in there that rpm-ostree just ignores. Then we could have e.g. a `bodhi` key and/or a `tracker` key under there or whatever.

jlebon:
Example commit: https://github.com/jlebon/fedora-coreos-config/commit/affe019de409d6c3640e75174aa9966063bd0529

manifest-lock.fast-track.yaml
lucab: I think rpm-ostree deserializer already ignores unknown keys. So it should be possible right now to turn this (and the other per-entry remarks) into the string value of a `_comment` key.

jlebon: This top-level comment is preserved, because it's regenerated by the action.

For the per-entry bits, I thought because we do https://github.com/coreos/rpm-ostree/blob/b2b3795df126927236bbff81a560c95c25a508b1/rust/src/lockfile.rs#L104 rpm-ostree would error out, but actually we don't enable it on https://github.com/coreos/rpm-ostree/blob/b2b3795df126927236bbff81a560c95c25a508b1/rust/src/lockfile.rs#L127, so it might work as is. (But we should enable it there though, and then we also just add a generic `metadata` key.)

.github/workflows/remove-graduated-overrides.yml
```diff
@@ -0,0 +1,36 @@
...
+  schedule:
+    - cron: '0 */6 * * *'
+
+jobs:
+  remove-graduated-overrides:
+    name: Remove graduated overrides
+    runs-on: ubuntu-latest
+    # really, we just need dnf, but this automatically keeps us in sync with
+    # testing-devel's $releasever
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
```
cgwalters, line 13: Why not `coreos-assembler`?

ci/remove-graduated-overrides.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/python3
```
cgwalters, line 1: Why not put this in coreos-assembler?  I think it'd be good to keep *code* there, and this repo is just configuration.

cgwalters: There's also some nontrivial interaction with both rpm-ostree and packages...any reason not to put this in rpm-ostree as something like `rpm-ostree lockfile update`?

.github/workflows/remove-graduated-overrides.yml
jlebon: Yeah maybe easier. It's kinda silly because we really just need `dnf` and `rpm-ostree` for this, so cosa is way overkill. But meh, this is a cron job so the actual image size doesn't really matter.

ci/remove-graduated-overrides.py
jlebon: In cosa... maybe. In rpm-ostree... not so sure. This is a pretty specific use case. I think this is a pretty good use of the Python dnf API.

jlebon: Offhand it seems to me like it belongs better here because again it's very specific to how this repo operates. Not sure, maybe we can revisit this down the line once we gain more experience with GitHub Actions across repos and best practices.

jlebon:
https://github.com/coreos/rpm-ostree/pull/2818 adds official support for metadata keys.

jlebon:
Updated this now! I reworked it as discussed in https://github.com/coreos/coreos-assembler/pull/2131 to use `metadata` keys instead to know the difference between fast-tracks and pins.

Testing the script directly locally works fine. Still need to test the GitHub Action on my fork.

ci/remove-graduated-overrides.py
```diff
@@ -0,0 +1,125 @@
...
+
+FAST_TRACK_HEADER = """
+# This lockfile should be used to pin to a package version (`type: pin`) or to
+# fast-track packages ahead of Bodhi (`type: fast-track`). Fast-tracked
+# packages will automatically be removed once they are in the stable repos.
+#
+# IMPORTANT: YAML comments *will not* be preserved. All `pin` overrides *must*
+# include a URL in the `metadata.reason` key. Overrides of type `fast-track`
+# *should* include a URL in the `metadata.reason` key, though it's acceptable to
+# omit one for FCOS-specific packages (e.g. ignition, afterburn, etc...).
```
dustymabe, line 22: your example in the readme don't have `metadata.reason` but it says here that it's required.

dustymabe (dismissed):
mostly LGTM - one comment

ci/remove-graduated-overrides.py
jlebon: Thanks, I initially called it `metadata.tracker` but made it `metadata.reason` to be more generic.

jlebon:
OK, this is tested and ready to go now!
Example output: https://github.com/jlebon/fedora-coreos-config/commit/6aaf20d99b96576980e65434cb74bbefe9f3676d

.github/workflows/remove-graduated-overrides.yml
jlebon: Ahh right, I remember now why I couldn't use cosa: we need to get rid of the `builder` user first. I filed https://github.com/coreos/coreos-assembler/issues/2223 to track this and we can circle back here once/if we do that.

ci/remove-graduated-overrides.py
jlebon: Going to resolve this for now. I'm pretty happy with how this is turning out and I think it makes sense to keep this kind of repo-specific automation in the target repo itself. If the logic gets gnarly or becomes generically useful we can look at moving it to cosa.

jlebon:
After this we can look at moving over to SRPMs to streamline things even more!

dustymabe (dismissed):
LGTM

jlebon:
Discussing this with @dustymabe, I realized that it's not entirely safe to automatically drop overrides without going through CI since if `bump-lockfile` hasn't run since an override was added, the base lockfile will not necessarily have the overridden package at the same NEVRA.

So I modified this now to instead submit a PR.

I wasn't able to fully test this version because the pull request code doesn't like the fact that `coreosbot-releng/fedora-coreos-config` isn't a fork of `jlebon/fedora-coreos-config`. And I didn't want to change that just to test this nor wanted to create another bot account to test. That said, it's pretty much what we already have in the streams repo for promotion PRs. If there are issues, we can just fix them as they come up.

dustymabe (approved):
LGTM

PR #967, opened 2021-04-16
overlay.d: Add back CLHM presets

kelvinfan001:
CLHM v0.21+ used in FCOS no longer has `{issue,motd}gen`-related units, but the
`05core` overlay is shared and used as `01fcos` in RHCOS, and RHCOS is still on
older versions of CLHM that require these units. Keep presets for these units
until RHCOS catches up to CLHM v0.21+.

Note: CLHM does not have `motdgen`-related units in RHEL even in older versions,
so we only add back `issuegen.service` and `issuegen.path` here.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -1,6 +1,8 @@
 # Presets here that eventually should live in the generic fedora presets
 
 # console-login-helper-messages - https://github.com/coreos/console-login-helper-messages
+enable console-login-helper-messages-issuegen.service
+enable console-login-helper-messages-issuegen.path
 enable console-login-helper-messages-gensnippet-os-release.service
 enable console-login-helper-messages-gensnippet-ssh-keys.service
 # CA certs (probably to add to base fedora eventually)


kelvinfan001:
Or would it be better to split CLHM-related presets into their own FCOS-specific and RHCOS-specific overlays?

dustymabe:
> Or would it be better to split CLHM-related presets into their own FCOS-specific and RHCOS-specific overlays?

Is it an option to just PR https://github.com/openshift/os to add them there (with a comment about when to drop it from the presets file for RHCOS)?

kelvinfan001:
> Is it an option to just PR https://github.com/openshift/os to add them there (with a comment about when to drop it from the presets file for RHCOS)?

Yeah, but I was thinking the the presets for CLHM should eventually be unified in FCOS and RHCOS, so we'll need to unify the two overlays some time in the future again. I _think_ there's no harm in enabling non-existent units, but if it's cleaner to split them, then I'll make a PR to https://github.com/openshift/os!

dustymabe:
> Yeah, but I was thinking the the presets for CLHM should eventually be unified in FCOS and RHCOS, so we'll need to unify the two overlays some time in the future again.

What do you mean exactly? Let's say RHCOS gets CLHM v0.21+, then what does the unified overlays (in terms of CLHM) look like? 

> I _think_ there's no harm in enabling non-existent units, but if it's cleaner to split them, then I'll make a PR to https://github.com/openshift/os!

I think if it's a temporary workaround until CLHM v0.21+ gets into RHCOS then I'd just PR openshift/os.

kelvinfan001:
> What do you mean exactly? Let's say RHCOS gets CLHM v0.21+, then what does the unified overlays (in terms of CLHM) look like?

Then we'll only need
```
enable console-login-helper-messages-gensnippet-os-release.service
enable console-login-helper-messages-gensnippet-ssh-keys.service
```

OK, I see what you mean now. Yeah, just temporarily modifying the RHCOS overlays makes much more sense. For some reason, I had thought that we needed to have separate CLHM presets in a FCOS-only directory like `15fcos` 😅. Thanks!

kelvinfan001:
Moved to here: https://github.com/openshift/os/pull/540

PR #969, opened 2021-04-19
manifests: Move CLHM's `motdgen` package to `user-experience`

kelvinfan001:
So all of console-login-helper-messages' packages are in one place.


kelvinfan001:
I'm not sure if this was intentional, but I noticed that CLHM's `motdgen` package is in the `fedora-coreos-base.yaml` manifest, instead of in `user-experience.yaml`, where the rest of the CLHM packages are. If `motdgen` was intentionally placed in `fedora-coreos-base.yaml`, we can close this PR.

PR #972, opened 2021-04-20
fedora-coreos-base: skip the script to disable module repos if no files match

vrutkovs:
Setting nullglob makes the repos skip passing "/etc/yum.repos.d/*modular.repo" to sed if no matching files found.

See error message in https://cirrus-ci.com/task/6296878716289024 - bash passes `/etc/yum.repos.d/*modular.repo` as is if no files match

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -55,6 +55,7 @@ postprocess:
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
+    shopt -s nullglob
     for x in /etc/yum.repos.d/*modular.repo; do
       sed -i -e 's,enabled=[01],enabled=0,' ${x}
     done


dustymabe:
In what case do the modular repos not exist?

vrutkovs:
Seems a clean F34 next-devel is no longer including these.
Manifest I'm using: https://github.com/vrutkovs/okd-os/tree/cirrus

jlebon:
They were split out in a separate subpackage but we explicitly keep shipping it: https://github.com/coreos/fedora-coreos-config/blob/2fa5061f4b417fd347250819866b35d09d8e5007/manifests/fedora-coreos.yaml#L13-L15

I guess your fork is missing that bit. Anyway, probably cleaner approach then is to move this `postprocess` entry up to `fedora-coreos.yaml` so it's colocated with the package specification.

vrutkovs:
No longer required

PR #975, opened 2021-04-22
Move comments in systemd services to own line

andymcc:
Comments in systemd services at the end of lines cause unnecessary
warning messges:

Failed to add dependency on on, ignoring: Invalid argument
Failed to add dependency on dracut, ignoring: Invalid argument
Failed to add dependency on 053+, ignoring: Invalid argument

The services would still start so this is a cosmetic change only.


jlebon (approved):
Nice catch, thanks!

PR #976, opened 2021-04-22
[next-devel] overrides: Fast-track dracut-053-4.fc34

dustymabe:
It has some upstream changes we'd like.
https://src.fedoraproject.org/rpms/dracut/pull-request/13


dustymabe:
requires #977

jlebon (approved):
Nice, didn't realize we'd get it so fast!

PR #979, opened 2021-04-26
[next-devel] overrides: drop graduated overrides

dustymabe:
Now that Fedora 34 is unfrozen we can drop most of these.


jlebon (approved):
Nice! CI will sanity-check there's no downgrade.

PR #985, opened 2021-04-27
manifests: explicitly add fwupd

jlebon:
This was pulled in transitively when we moved to f33, and we knowingly
allowed it because we wanted it. But in f34, it looks like it gets
dropped out again, so explicitly list it now to make sure we keep
shipping it.


sohankunkerkar (dismissed):

dustymabe (dismissed):
LGTM

sohankunkerkar (approved):

PR #859, opened 2021-02-18
[rawhide] manifest.yaml: start tracking f35

jlebon:
Rawhide is now f35.

PR #988, opened 2021-04-29
Add temporary workaround for `agetty --reload` SELinux denial

jlebon:
In f34+, we're hitting an SELinux denial from c-l-h-m trying to do
`agetty --reload` and causing `/run/agetty.reload` to be created with
the wrong label. This then prevents agetty from adding an inotify watch
to know when to reload the prompt to display new information. For more
details, see:

https://github.com/coreos/fedora-coreos-config/pull/859#issuecomment-783713383

This is tracked at https://bugzilla.redhat.com/show_bug.cgi?id=1932053.

With this workaround, we create the file up front in the initrd so that
it gets relabeled by systemd on switchroot and thus will already exists
with the right label well before c-l-h-m or anything else tries to
`agetty --reload`.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/60coreos-agetty-workaround/coreos-touch-run-agetty.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/60coreos-agetty-workaround/coreos-touch-run-agetty.service
@@ -1,0 +1,12 @@
+# Temporary hack to work around agetty SELinux denials.
+# https://github.com/coreos/fedora-coreos-config/pull/859#issuecomment-783713383
+# https://bugzilla.redhat.com/show_bug.cgi?id=1932053
+[Unit]
+Description=CoreOS: Touch /run/agetty.reload
+Documentation=https://bugzilla.redhat.com/show_bug.cgi?id=1932053
+DefaultDependencies=false
+
+[Service]
+Type=oneshot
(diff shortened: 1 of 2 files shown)


jlebon:
Still need to test this.

jlebon:
Once we get this in, we can drop the way hackier workaround in rawhide that was added in https://github.com/coreos/fedora-coreos-config/pull/859.

dustymabe:
would dropping a stub ignition config that creates the file work too? I think files that get laid down by ignition get relabeled by default.

jlebon:
> would dropping a stub ignition config that creates the file work too? I think files that get laid down by ignition get relabeled by default.

Ignition configs can't write to `/run`.

dustymabe (approved):
LGTM: It would be nice if we bake in some sort of reminder to go back and remove this.

jlebon:
> LGTM: It would be nice if we bake in some sort of reminder to go back and remove this.

Heh yeah, this is what this comment is about:

> TODO f34: check if we can drop this whole module

(See https://github.com/coreos/fedora-coreos-tracker/issues/704#issuecomment-794347073). We can bump it to f35 when we're rebasing `testing` if it's still not fixed.

jlebon:
Tested working!

dustymabe (approved):
LGTM

PR #993, opened 2021-04-30
tests/countme: adapt for rawhide

jlebon:
Rawhide only has a single repo enabled with the countme flag.


cgwalters (approved):
Seems sane.  Maybe we could `cat /etc/yum.repos.d/* | grep -c enabled=1` or so though?

dustymabe (approved):

jlebon:
> Seems sane. Maybe we could `cat /etc/yum.repos.d/* | grep -c enabled=1` or so though?

It's a bit trickier, because not all enabled repos have countme (`fedora-cisco-openh264`) and not all repos with countme are enabled.

Maybe instead we should just check that (1) more than 1 repo was found, and (2) all the repos found were successful (i.e. grep for `N/N`, and check that N > 1)?

PR #1002, opened 2021-05-10
overrides: freeze podman to 3.1.2-1

jlebon:
There is a regression in the latest 3.2.0 RC in stable:
- https://github.com/containers/podman/issues/10274
- https://github.com/containers/podman/pull/10288

No Bodhi updates yet with the fix.

Also hit that bug myself locally on FSB.


jlebon:
CI failing due to downgrade, which is expected here.

dustymabe (approved):
LGTM

dustymabe:
Merging over red because we want the downgrade.

PR #1003, opened 2021-05-12
overlay.d & tests/misc-ro: Fix and test for 664 mode for some files in /etc

### LINKED ISSUE #829: Group writable files in /etc
**Describe the bug**

The following files are group writable in `/etc` when they should not be:

```
$ sudo find /etc -type f -perm /g+w
/etc/group
/etc/crypto-policies/state/current
/etc/iscsi/initiatorname.iscsi
/etc/selinux/config
/etc/systemd/dont-synthesize-nobody
/etc/group-
/etc/passwd
```

Initial fix for rpm-ostree is https://github.com/coreos/rpm-ostree/pull/2824 and this will fix most of those during an update but we need manual handling for passwd/group files as they are marked an modified (M) by ostree:

```
$ sudo ostree admin config-diff | grep -E "group|passwd"   
M    passwd
M    group
A    group-
A    passwd-
```

Corresponding test is in progress in https://github.com/coreos/fedora-coreos-config/pull/1003

**Reproduction steps**

By default

**Expected behavior**

Those files are not group writable.

**Actual behavior**

Those files are group writable.

**System details**

All Fedora CoreOS images I have tested (at least in the last month).

**Ignition config**

N/A

travier:
overlay.d/09misc: Add workaround for /etc/passwd & group mod issue

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/829

---

tests/misc-ro: Check mode for files in /etc

- Do not allow others or group writable files in /etc
- Extended ownership (root:root) & mode (644) checks for
  /etc/{passwd,group}

Test for https://github.com/coreos/fedora-coreos-tracker/issues/829

### DIFF
--- a/overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
+++ b/overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
@@ -1,0 +1,11 @@
+# Workaround for https://github.com/coreos/fedora-coreos-tracker/issues/829
+# Fix mode (chmod g-w) for existing files on the system during boot
+z /etc/crypto-policies/state/current               644 root root
+z /etc/group                                       644 root root
+z /etc/group-                                      644 root root
+z /etc/iscsi/initiatorname.iscsi                   644 root root
+z /etc/passwd                                      644 root root
+z /etc/passwd-                                     644 root root
+z /etc/selinux/config                              644 root root
+z /etc/ssh/sshd_config.d/40-disable-passwords.conf 644 root root
+z /etc/systemd/dont-synthesize-nobody              644 root root
--- a/overlay.d/README.md
+++ b/overlay.d/README.md
@@ -15,7 +15,8 @@ https://bugzilla.redhat.com/show_bug.cgi?id=1700056
 09misc
 ------
 
-Warning about `/etc/sysconfig`.
+* Warning about `/etc/sysconfig`.
+* Temporary systemd-tpmfiles.d config to fix ownership and permissions in /etc
 
 14NetworkManager-plugins
 ------------------------
--- a/overlay.d/README.md
+++ b/overlay.d/README.md
@@ -22,14 +22,6 @@ Warning about `/etc/sysconfig`.
 
 Disables the Red Hat Linux legacy `ifcfg` format.
 
-20platform-chrony
------------------
-
-Add static chrony configuration for NTP servers provided on platforms
-such as `azure`, `aws`, `gcp`. The chrony config for these NTP servers
-should override other chrony configuration (e.g. DHCP-provided)
-configuration.
-
 15fcos
 ------
 
@@ -45,4 +37,7 @@ Things that are more closely "Fedora CoreOS":
 20platform-chrony
 -----------------
 
-Platform aware timeserver setup for chrony daemon.
+Add static chrony configuration for NTP servers provided on platforms
+such as `azure`, `aws`, `gcp`. The chrony config for these NTP servers
+should override other chrony configuration (e.g. DHCP-provided)
+configuration.
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -155,3 +155,18 @@ fi
 if [[ $sys_fs_cgroup_source != cgroup2 ]]; then
     fatal "/sys/fs/cgroup is not cgroup2"
 fi
+
+for perms in 'o+w' 'g+w'; do
+	list="$(find /etc -type f -perm /${perms})"
+	if [[ -n "${list}" ]]; then
+		fatal "found files with ${perms}:\n${list}"
+	fi
+done
+ok "no files with o+w or g+w found in /etc"
+
+for f in '/etc/passwd' '/etc/group'; do
+	if [[ $(stat --format="%a %u %g" "${f}") != "644 0 0" ]]; then
+		fatal "found incorrect permissions for ${f}"
+	fi
+done
+ok "correct ownership and mode on /etc/passwd & /etc/group"


cgwalters (dismissed):
Looks fine as is.

minor: Could dedup code with a `for` loop.

Actually on that topic, any reason not to just do: `find /etc -type f -perm /g+w` and check that's empty?

travier:
> Actually on that topic, any reason not to just do: find /etc -type f -perm /g+w and check that's empty?

Good idea, looking into that

cgwalters:
We need to land a systemd unit that fixes the perms before we can ship this, right?  But maybe refactor the test to check for *other* `root` group writable files and not error for passwd/group for now?

travier:
We should make a barrier release for this fix and then we could revert it in a following one once we have released a fixed rpm-ostree for COSA.

Ping @dustymabe

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
```diff
@@ -0,0 +1,3 @@
+z /etc/passwd                                      644 root root
+z /etc/group                                       644 root root
+z /etc/ssh/sshd_config.d/40-disable-passwords.conf 644 root root
```
dustymabe, line 10: I don't see `/etc/ssh/sshd_config.d/40-disable-passwords.conf` mentioned anywhere else (this PR or the original issue) and it seems to have the right permissions on my system. What's the context for adding it here?

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
```diff
@@ -0,0 +1,3 @@
+z /etc/passwd                                      644 root root
+z /etc/group                                       644 root root
```
dustymabe: Should we do anything for `passwd-` and `group-`?

```
$ ls -l /etc/pass* /etc/group* 
-rw-rw-r--. 1 root root 162 Apr 29 03:17 /etc/group
-rw-rw-r--. 1 root root 108 Apr 29 03:17 /etc/group-
-rw-rw-r--. 1 root root 243 Apr 29 03:17 /etc/passwd
-rw-rw-r--. 1 root root  87 Apr 29 03:17 /etc/passwd-
```

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
travier: It was failing locally when running tests with kola. Will double check.

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
travier: Yes. We will probably have to add all the files to the list.

overlay.d/README.md
```diff
@@ -45,4 +38,7 @@ Things that are more closely "Fedora CoreOS":
 20platform-chrony
 -----------------
 
-Platform aware timeserver setup for chrony daemon.
+Add static chrony configuration for NTP servers provided on platforms
```
jlebon, line 41: This block move looks unrelated. Would be good to make it a separate commit so it's easier to revert the `tmpfiles.d` bits only later on.

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
```diff
@@ -0,0 +1,11 @@
+z /etc/fedora-coreos-pinger/config.d/90-disable-on-non-production-stream.toml 644 root root
+z /etc/fedora-coreos-pinger/config.d/95-disable-on-dev.toml                   644 root root
```
dustymabe: These are files from local *dev* builds.

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
```diff
@@ -0,0 +1,11 @@
...
+z /etc/group                                                                  644 root root
+z /etc/group-                                                                 644 root root
+z /etc/passwd                                                                 644 root root
+z /etc/passwd-                                                                644 root root
+z /etc/selinux/config                                                         644 root root
+z /etc/ssh/sshd_config.d/40-disable-passwords.conf                            644 root root
+z /etc/systemd/dont-synthesize-nobody                                         644 root root
+z /etc/zincati/config.d/90-disable-on-non-production-stream.toml              644 root root
+z /etc/zincati/config.d/95-disable-on-dev.toml                                644 root root
```
dustymabe: how are you coming up with this list?

travier: So far I've added everything that was triggering failures in CI. Agree that we should probably manually exclude the files that are only set for dev builds. Will update the test for that.

jlebon: Hmm, I think we should really just be fixing `/etc/{group,passwd}*`. They're the most important and the rpm-ostree patch only covers those and puts us back at parity with traditional systems.

For everything else, we should check what the intended permissions should be. If we diverge from traditional systems, then that's likely another patch in rpm-ostree. If we don't, then either we leave it alone or work with upstream to determine whether that's correct and if their Makefile/spec file needs to be fixed.

travier: I'm afraid all of those are due to the rpm-ostree issue as I don't have those modes on a classic Fedora system.

tests/kola/misc-ro
```diff
@@ -155,3 +155,27 @@ fi
 if [[ $sys_fs_cgroup_source != cgroup2 ]]; then
     fatal "/sys/fs/cgroup is not cgroup2"
 fi
+
+dev_only_excludes=(
+	'/etc/fedora-coreos-pinger/config.d/90-disable-on-non-production-stream.toml'
+	'/etc/fedora-coreos-pinger/config.d/95-disable-on-dev.toml'
+	'/etc/zincati/config.d/90-disable-on-non-production-stream.toml'
+	'/etc/zincati/config.d/95-disable-on-dev.toml'
+)
```
dustymabe: Is there any way to just make these files get created with the right permissions to begin with? Would setting the umask to something more appropriate here help?

https://github.com/coreos/fedora-coreos-config/blob/adb618327f29f113214a7769186fcb681b11186d/manifest.yaml#L31-L35

travier: Good idea, will try that

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
jlebon: So you mean both `/etc/selinux/config` and `/etc/systemd/dont-synthesize-nobody` are fixed by https://github.com/coreos/rpm-ostree/pull/2824? I'm confused how that's possible seeing as the PR only relates to passwd/group files.

Let me test this locally.

jlebon:
For `/etc/ssh/sshd_config.d/40-disable-passwords.conf`, I think we need to fix it up in a postprocessing script (this is a shortcoming of git only storing the execute bit). Or maybe better to just tweak cosa to canonicalize everything in `/etc` to 0644.

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
jlebon: Really weird, for some reason I can't reproduce the original issue for `/etc/{passwd,group}` even with https://github.com/coreos/rpm-ostree/pull/2824 reverted. Indeed `/etc/selinux/config` is 644 on traditional, but it's also 644 on recent local FCOS builds. Need to dig into where the disconnect is and whether there is a bug somewhere we need to fix.

Hmm, it's possibly yet another umask difference in the pipeline. We have workarounds like https://github.com/coreos/fedora-coreos-pipeline/pull/283, but if umask is 0002 then it won't kick in.

travier: I could not reproduce it locally either. It's something specific to the pipeline.

travier: Maybe the git clone happen with a 002 umask initially?

jlebon: I'm pretty sure now that it's privileged vs unprivileged builds. In cosa we have https://github.com/coreos/coreos-assembler/blob/961782c5de237e29fe6dedb71a00e256173cd5e1/src/supermin-init-prelude.sh#L33. I'm verifying that now.

jlebon:
https://github.com/coreos/coreos-assembler/pull/2204 addresses `/etc/ssh/sshd_config.d/40-disable-passwords.conf`.

manifest.yaml
```diff
@@ -30,6 +30,19 @@ postprocess:
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
+    # Set umask to make sure we create all folder and files with mode 644
+    umask 022
```
jlebon: I think we also need this in `fedora-coreos.yaml` for `95-disable-on-dev.toml`.

manifest.yaml
```diff
@@ -30,6 +30,19 @@ postprocess:
...
     mkdir -p /etc/fedora-coreos-pinger/config.d /etc/zincati/config.d
     echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/163\nreporting.enabled = false' > /etc/fedora-coreos-pinger/config.d/90-disable-on-non-production-stream.toml
     echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/163\nupdates.enabled = false' > /etc/zincati/config.d/90-disable-on-non-production-stream.toml
+
+  # Workaround for https://github.com/coreos/fedora-coreos-tracker/issues/829
```
dustymabe: We can at least drop `/etc/group` and `/etc/passwd` when the rpm-ostree bug is fixed? Do we know when we can we drop the other ones?

manifest.yaml
jlebon: I think we can drop this...

manifest.yaml
```diff
@@ -30,6 +30,19 @@ postprocess:
...
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    chmod 644 \
+      /etc/group \
+      /etc/passwd \
+      /etc/selinux/config \
+      /etc/ssh/sshd_config.d/40-disable-passwords.conf \
+      /etc/systemd/dont-synthesize-nobody
```
jlebon: ... and everything here minus `/etc/{passwd,group}` with https://github.com/coreos/coreos-assembler/pull/2206 and https://github.com/coreos/coreos-assembler/pull/2204.

manifest.yaml
```diff
@@ -33,3 +33,9 @@ postprocess:
     mkdir -p /etc/fedora-coreos-pinger/config.d /etc/zincati/config.d
     echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/163\nreporting.enabled = false' > /etc/fedora-coreos-pinger/config.d/90-disable-on-non-production-stream.toml
     echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/163\nupdates.enabled = false' > /etc/zincati/config.d/90-disable-on-non-production-stream.toml
+
+  # Workaround for https://github.com/coreos/fedora-coreos-tracker/issues/829
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    chmod 644 /etc/passwd /etc/group
```
jlebon: Actually, even this is fixed by https://github.com/coreos/coreos-assembler/pull/2206 so we could drop this entirely. :tada:

tests/kola/misc-ro
```diff
@@ -155,3 +155,18 @@ fi
...
+	fi
+done
+ok "no files with o+w or g+w found in /etc"
+
+for f in '/etc/passwd' '/etc/group'; do
+	if [[ $(stat --format="%a %u %g" "${f}") != "644 0 0" ]]; then
+		fatal "found incorrect permissions for ${f}"
+	fi
+done
+ok "correct ownership and mode on /etc/passwd & /etc/group"
```
jlebon, line 172: Isn't this redundant with the previous test?

jlebon (dismissed):
Two last comments, LGTM otherwise!

With the cosa patches, I think we're lined up to have the next release be the barrier release after which we can drop the `tmpfiles.d` dropin.

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
```diff
@@ -0,0 +1,7 @@
+z /etc/group                                       644 root root
```
dustymabe, line 4: are comments allowed in this file? should we give a short explainer on why these files have the wrong permissions on existing systems and why they need to be fixed?

tests/kola/misc-ro
travier: This one includes the owner/group checks but agree that it is a bit redundent.

dustymabe: I'm cool either way.

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
travier: Yes, will do that

overlay.d/09misc/usr/lib/tmpfiles.d/coreos-fix-etc-ownership.conf
```diff
@@ -0,0 +1,9 @@
+# Workaround for https://github.com/coreos/fedora-coreos-tracker/issues/829
+# Fix mode (chmod g-w) for existing files on the system during boot
+z /etc/group                                       644 root root
+z /etc/group-                                      644 root root
+z /etc/passwd                                      644 root root
+z /etc/passwd-                                     644 root root
+z /etc/selinux/config                              644 root root
+z /etc/ssh/sshd_config.d/40-disable-passwords.conf 644 root root
+z /etc/systemd/dont-synthesize-nobody              644 root root
```
dustymabe, line 11: ok I ran the detector on on one of my deployed nodes and I see one file in the list that isn't in here:

```
[core@dustymabecom ~]$ cat /tmp/foo.sh 
for perms in 'o+w' 'g+w'; do
        list="$(find /etc -type f -perm /${perms})"
        if [[ -n "${list}" ]]; then
                echo -e "found files with ${perms}:\n${list}"
        fi
done
[core@dustymabecom ~]$ sudo bash /tmp/foo.sh 
found files with g+w:
/etc/passwd-
/etc/crypto-policies/state/current
/etc/selinux/config
/etc/systemd/dont-synthesize-nobody
/etc/passwd
/etc/group-
/etc/group
```

Anybody else have nodes they can run checks on?

jlebon: Mine is covered. Hmm, it's tempting to just chmod everything in `/etc` owned by `root:root` instead. Though there's a chance that some users rely on this for new files they've created.

dustymabe: Yeah I'd prefer to keep the risk low here.

dustymabe:
What's left to do here? Add `/etc/crypto-policies/state/current` to the list?

travier:
Initial list from https://github.com/coreos/fedora-coreos-tracker/issues/829 also included `/etc/iscsi/initiatorname.iscsi` but I do not see this one anymore. Will update the list tomorrow.

tests/kola/misc-ro
travier: I'd prefer we keep both as the first check is a more general one and the second one is more security critical.

dustymabe (approved):
LGTM

overlay.d/README.md
```diff
@@ -15,21 +15,14 @@ https://bugzilla.redhat.com/show_bug.cgi?id=1700056
 09misc
 ------
 
-Warning about `/etc/sysconfig`.
+* Warning about `/etc/sysconfig`.
+* Temporary systemd-tpmfiles.d config to fix ownership and permissions in /etc
```
jlebon, line 19:
```suggestion
* Temporary systemd-tmpfiles.d config to fix ownership and permissions in /etc
```

jlebon (approved):
One typo, but not worth the respin.
Thanks!

PR #1004, opened 2021-05-12
[next-devel] overrides: freeze podman to 3.1.2-1

jlebon:
There is a regression in the latest 3.2.0 RC in stable:
- https://github.com/containers/podman/issues/10274
- https://github.com/containers/podman/pull/10288

No Bodhi updates yet with the fix.

Also hit that bug myself locally on FSB.

(cherry picked from commit b72844c5c801e7bf899ac65b4040aaecb743a7ec)


dustymabe:
I think #1006 obsoletes this.

PR #1006, opened 2021-05-12
[next-devel] overrides: Fast-track podman-3.1.2-3

dustymabe:
Fixes podman selinux labelling regression.
https://github.com/coreos/fedora-coreos-tracker/issues/818


jlebon (approved):
Would've been nice as a cherry-pick of https://github.com/coreos/fedora-coreos-config/pull/1005, but cool as is too.

PR #1009, opened 2021-05-13
[testing-devel] overrides: Fast-track ignition 2.10.1-1

arithx:


sohankunkerkar (dismissed):

jlebon (dismissed):
LGTM though can you add a link to the Bodhi update? This makes it easier for others to check when it graduated (will circle back to https://github.com/coreos/fedora-coreos-config/pull/965 soon so this is all automatic).

arithx:
Added

arithx:
Actually closing as it looks like this has already made it to stable and landed in both testing-devel & next-devel.

PR #1011, opened 2021-05-13
Add support for multipath on firstboot

jlebon:
As we've learned in https://bugzilla.redhat.com/show_bug.cgi?id=1954025,
we can't assume that we can use any individual path before multipathd
unifies them. This means that it's not correct to turn it on from the
second boot onwards only, which is the current approach.

So we need to support multipath at first boot, and further, we need to
delay all I/O to the boot disk to *after* multipathd takes ownership.

This patch does this by introducing a generator and a target. We need to
use a target because `After=multipathd.service` is not enough to ensure
that multipathd finished setting up the devices. The target explicitly
waits for the multipathed boot device to appear. (Note though that we
*can't* assume that root is also directly on top of the multipath device
because of LUKS-on-root.)

This creates an awkward UX, which is that multipath is the only root
setup option which is *not* driven by Ignition but instead set up behind
its back.

This will be somewhat better once Ignition supports kernel arguments,
because the `rd.multipath` karg can be fed that way. But long-term, we
should consider teaching Ignition to configure multipath devices. This
would fix the configuration problem (right now, we only support
first-booting with default configs), and would free users from adding
the necessary kargs, which would be done by the rootmap code as usual.

But still, even in that world, there's a gap where the Ignition config
could be on the boot device, in which case multipathd would need to be
turned on beforehand.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1954025


jlebon:
Split out prep patches in https://github.com/coreos/fedora-coreos-config/pull/1012.

Requires: https://github.com/coreos/ignition/pull/1208
For LUKS-on-multipath, requires: https://github.com/coreos/coreos-installer/pull/541

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.service
```diff
@@ -14,6 +14,8 @@ Requires=dev-disk-by\x2dlabel-boot.device
 After=dev-disk-by\x2dlabel-boot.device
 # Start after Ignition has finished
 After=ignition-files.service
+# As above, this isn't strictly necessary, but on principle.
+After=coreos-multipath-wait.target
```
cgwalters, line 18: See my comment about `blockdev@` in https://github.com/coreos/fedora-coreos-tracker/pull/832#issuecomment-841294503

I think that seems cleanest, particularly since it exists in systemd today for FCOS.  Basically everywhere instead of just depending on `Device=` we can depend on that unit and multiple things can hook off of it.

cgwalters (commented):
OK a lot going on here.   So a high level description is basically: "We support multipath setup at `coreos-installer` time", right?

We'd still support `rd.multipath.default` day 2?

Do you have tests for this yet?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
```diff
@@ -6,25 +6,37 @@ set -euo pipefail
...
-# On RHEL 8 the version of lsblk doesn't have PTUUID. Let's detect
-# if lsblk supports it. In the future we can remove the 'if' and
-# just use the 'else'.
-if ! lsblk --help | grep -q PTUUID; then
-    # Get the PKNAME
-    eval $(lsblk --output PKNAME --pairs --paths --nodeps "$1")
-    # Get the PTUUID
-    eval $(blkid -o export $PKNAME)
+# If it's on multipath, get the parent device from udev properties.
+DM_MPATH=$(eval $(udevadm info --query property --export "$1") && echo "${DM_MPATH:-}")
```
cgwalters, line 10: You changed this service to be `After=multipathd.service`, but this is going behind its back.  Why not run `Before=multipathd.service`?

Conceptually there's a bit of a "loop" here I guess either way.  Ultimately...I think we may want to have `coreos-installer` do this GPT scrambling too, and skip it on metal boots.  Or at least, skip it on multipath boots?

jlebon:
> OK a lot going on here. So a high level description is basically: "We support multipath setup at `coreos-installer` time", right?

Yup, exactly!

> We'd still support `rd.multipath.default` day 2?

Yes, if the multipath environment allows it (i.e. not like the one in the linked RHBZ).

> Do you have tests for this yet?

Not yet. I haven't fully tested it in RHCOS yet (which is partly why it's still in draft). I'd like to add a new `testiso` scenario for this so it's truly e2e. :recycle:  And the existing test will cover the day-2 path.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
jlebon: Hmm, can you expand on the "going behind its back" bit? The `After=` is required as per the "delay all I/O" bit in the commit message.

> Conceptually there's a bit of a "loop" here I guess either way. Ultimately...I think we may want to have coreos-installer do this GPT scrambling too, and skip it on metal boots. Or at least, skip it on multipath boots?

We still need at least the "move backup GPT header to the end of the disk" part I think. Not sure if moving it to `coreos-installer` is safe because you'd still end up with cloned GUIDs if you're just preparing a custom metal image.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.service
jlebon: I'll look into this some more, but it seems like it won't save us from having to ship a target unit (because it's not in el8 yet) and a multipath generator.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=CoreOS Wait For Multipathed Boot
+DefaultDependencies=false
+Before=dracut-initqueue.service
+After=dracut-cmdline.service
+Requires=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
+After=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
```
cgwalters, line 7: The more I think about this the more I feel like we should be avoiding labels in nontrivial setups.  I mean it totally makes sense as our  "stock unconfigured default", but we've long since moved to using `root=` karg for nontrivial layouts.

Or really I think what I'm settling on is: we should find `/boot` in exactly the same place we found `/`.

So instead of using a udev rule to probe any multipath-based device with a specific label, we should inspect the blockdev for `/` that we mounted in the initramfs, and find the `label=boot` partition *just on that device*.

Or perhaps even better, mount `/boot` in the initramfs using the same device and pass that through.

jlebon:
Testing this with https://github.com/coreos/coreos-assembler/pull/2172:

```
$ kola testiso -S --scenarios iso-install --output-dir tmp/kola-testiso-mpath --qemu-multipath
Testing scenarios: [iso-install]
Successfully tested scenario iso-install for 34.20210517.dev.1 on bios (metal on multipath)
```

:tada:

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
cgwalters: Sorry, I was confused at the meaning of "parent" here.  I understand now that `DM_MPATH` is just the overall mpath device, distinct from its "backing devices".

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
cgwalters: OK I ended up using the current links in https://github.com/coreos/fedora-coreos-config/pull/1022 for now.

jlebon: > Or perhaps even better, mount /boot in the initramfs using the same device and pass that through.

A general approach we've been taking which has been really helpful I think is to "turn off everything" on the way out of the initramfs (obviously except for the root mount itself). This has served us well because then e.g. it makes sure that the rootfs is configured well and that we don't get in the way of user-configured things. Obviously the boot mount is kinda special too, and I don't expect users wanting to customize it much, though conceptually it'd be nice to just leave that to the real root as well like everything else if we don't strictly need it.

> So instead of using a udev rule to probe any multipath-based device with a specific label, we should inspect the blockdev for / that we mounted in the initramfs, and find the label=boot partition just on that device.

Hmm, this approach would break if we're reprovisioning the rootfs on a different device from the initial boot disk, no?

In my mind, this always comes back to multipathd having no way of signaling when it's done taking ownership of the devices. If it did, then we could wait for that and still just use the `boot` label as usual. And in fact we wouldn't need to special-case `rd.multipath` at all. I think it's worth pursuing with multipath developers maybe to see what they think.

jlebon:
OK, updated this now! The [new tests](https://github.com/coreos/coreos-assembler/pull/2172) are passing on both FCOS and RHCOS:

```
$ kola testiso -S --scenarios iso-install,iso-offline-install --qemu-multipath
Testing scenarios: [iso-install iso-offline-install]
Successfully tested scenario iso-install for 34.20210520.dev.0 on bios (metal on multipath)
Successfully tested scenario iso-offline-install for 34.20210520.dev.0 on bios (metal on multipath)
```

```
$ kola testiso -S --scenarios iso-install,iso-offline-install --qemu-multipath
Testing scenarios: [iso-install iso-offline-install]
Successfully tested scenario iso-install for 48.84.202105202203-0 on bios (metal on multipath)
Successfully tested scenario iso-offline-install for 48.84.202105202203-0 on bios (metal on multipath)
```

That said, this is blocked on https://github.com/coreos/ignition/pull/1208. (And https://github.com/coreos/coreos-installer/pull/541 is needed for LUKS-on-multipath, but not a strict blocker otherwise.)

cgwalters:
So...I think let's land this and get some testing on real-world (non-qemu) multipath setups, particularly w/OpenShift 4.9?  Then we can debate whether to try backporting just https://github.com/coreos/fedora-coreos-config/pull/1022 to 4.8 or try this whole thing.

jlebon:
We still need the Ignition patch for multipath (in both FCOS and RHCOS) to get this actually tested in CI. I wasn't planning to backport this patchset but the workarounds we provided are hitting issues, so we may not have a choice. If so, then likely will backport the Ignition patch so we can start the backport process.

Meanwhile, feel free to stage at least #1022 (and probably also at least the `DM_ACTIVATION` patch here too) in the `rhcos-4.8` branch.

jlebon:
Ignition fast-track in https://github.com/coreos/fedora-coreos-config/pull/1034. We should be able to test this in CI after that!

jlebon:
OK, `hold` dropped and added a test in CI. This is ready to go!

jlebon:
:tada: 

```
+ kola testiso -S --scenarios iso-offline-install --qemu-multipath --output-dir tmp/kola-testiso-mpath
Testing scenarios: [iso-offline-install]
Successfully tested scenario iso-offline-install for 34.20210527.dev.0 on bios (metal on multipath)
```

cgwalters (approved):

PR #1015, opened 2021-05-17
[rhcos-4.7] backport updates for pxe rootfs fetch retry

dustymabe:
```
commit b7835c1426932a1bf6517e9c1ee6ec2754ba653a
Date:   Thu Apr 15 13:34:13 2021 -0400

    `--retry-all-errors`.
    
    (cherry picked from commit d2776ed618d7d72fc0b226b7f08e45d3182406e7)

commit fd300bece5a4163d74aaab8f548f4d1631947648
Date:   Tue Apr 6 11:11:55 2021 -0400

    (cherry picked from commit b4a65864ef3b7d05b1452ee92384266c875196df)

```


dustymabe:
xref: https://bugzilla.redhat.com/show_bug.cgi?id=1960748

jlebon (approved):

miabbott (approved):
LGTM

PR #1016, opened 2021-05-17
[rhcos-4.6] backport updates for pxe rootfs fetch retry

dustymabe:
```
commit 42298433aa9e077c9b0fe171c50f18794897d99a 
Date:   Thu Apr 15 13:34:13 2021 -0400

    `--retry-all-errors`.
    
    (cherry picked from commit d2776ed618d7d72fc0b226b7f08e45d3182406e7)

commit ca46a43fe4c1745be8aff12e4c43ca066b95a87e
Date:   Tue Apr 6 11:11:55 2021 -0400

    (cherry picked from commit b4a65864ef3b7d05b1452ee92384266c875196df)

commit 412fdc983be2d74b0f0725e11cdddb5bc6cf3300
Author: Benjamin Gilbert <bgilbert@redhat.com>
Date:   Thu Dec 3 00:04:00 2020 -0500

    20live: log rootfs fetch errors
    
    When rootfs fetch fails, ensure we have some useful logging for tracking
    down the problem.
    
    (cherry picked from commit 0e432bd7ea4839bf80304233337def06f6d7fd3d)
```

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/20live/coreos-livepxe-rootfs.sh
@@ -34,7 +34,7 @@ elif [[ -n "${rootfs_url}" ]]; then
     # image hash.
     # bsdtar can read cpio archives and we already depend on it for
     # coreos-liveiso-persist-osmet.service, so use it instead of cpio.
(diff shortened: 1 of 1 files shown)


dustymabe:
xref: https://bugzilla.redhat.com/show_bug.cgi?id=1960750

dustymabe:
I'm going to assume we can ignore that CI failure.

jlebon (approved):

PR #1021, opened 2021-05-18
live-generator: don't call `man bootup`

jlebon:
Even though it's on a comment line, because it's in a heredoc, bash does
try to execute this. This fails on FCOS thankfully because there is no
`man` on FCOS, but it still logs an error message. (And... any derived
system which does ship `man`, I think this actually would dump the
manpage into the unit.)


jlebon:
I verified that's the only instance of this in the overlays.

cgwalters (approved):
Oh man.  Wow.  :laughing: :cry:

bgilbert (approved):
😨

dustymabe (approved):

PR #1026, opened 2021-05-19
overlay/05core: Add a new /usr/lib/coreos/generator-lib.sh

cgwalters:
Hopefully in the future we'll create a nice `rdcore` like Rust
place for our generators.

For now let's factor out a little helper library.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -2,11 +2,7 @@
 export PATH="/usr/bin:/usr/sbin:${PATH}"
 set -euo pipefail
 
-# Generators don't have logging right now
-# https://github.com/systemd/systemd/issues/15638
-exec 1>/dev/kmsg; exec 2>&1
-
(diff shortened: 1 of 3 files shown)


cgwalters:
Followup to discussion in https://github.com/coreos/fedora-coreos-config/pull/1022

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -2,11 +2,7 @@
 export PATH="/usr/bin:/usr/sbin:${PATH}"
 set -euo pipefail
 
-# Generators don't have logging right now
-# https://github.com/systemd/systemd/issues/15638
-exec 1>/dev/kmsg; exec 2>&1
-
-UNIT_DIR="${1:-/tmp}"
+. /usr/lib/coreos-shell-generator/lib.sh
```
jlebon: Minor: WDYT about `/usr/lib/coreos/shell-generator/lib.sh` instead? Feels cleaner to just own a `coreos` "package" dir.

jlebon (dismissed):
Idea makes sense to me!

jlebon (approved):

PR #1032, opened 2021-05-24
[next-devel] overrides: freeze on dracut-053-5.fc34

dustymabe:
There are some NetworkManager related changes and possibly others
that are causing failures in our bump-lockfile process. We need
to investigate these issues before promoting dracut-054.

https://github.com/coreos/fedora-coreos-tracker/issues/842
(cherry picked from commit f6dab79d48ef4d82dacf04f44285a3c7eb2adfec)


darkmuggle (approved):
/lgtm

PR #1033, opened 2021-05-25
move to cgroups v2 everywhere

dustymabe:
Adjust the kernel arguments so that we're now using cgroups v2
in our testing-devel (and subsequently, testing and stable) stream(s).

Context: https://github.com/coreos/fedora-coreos-tracker/issues/292


Conan-Kudo (approved):

bgilbert (approved):
LGTM

PR #1034, opened 2021-05-26
overrides: fast-track ignition-2.10.1-3.fc34

jlebon:
For firstboot multipath support:
https://github.com/coreos/fedora-coreos-config/pull/1011

PR #1042, opened 2021-06-02
[next-devel] overrides: drop graduated overrides

bgilbert:


dustymabe (approved):
LGTM

dustymabe:
re-kicked CI

jlebon:
This is failing on the multipath test because this branch is missing the [Ignition fast-track](https://github.com/coreos/fedora-coreos-config/pull/1034).

Could cherry-pick here for the benefit of CI, though since this stream is currently disabled... let's just drop that bit from the issue template?

bgilbert:
Yeah, I wondered about that.  If we're going to let the overrides go stale in this branch, another approach is to force-merge this PR, then the checklist entry becomes a no-op.

jlebon:
> Yeah, I wondered about that. If we're going to let the overrides go stale in this branch, another approach is to force-merge this PR, then the checklist entry becomes a no-op.

Sure, that SGTM!

PR #1043, opened 2021-06-02
tests/kdump: New kdump test

cgwalters:
We should have done this from the very start.  I hit a few issues
here:

- Docs are missing the requirement for a `root` karg
- cosa needs patching to ignore the crash on the console

### DIFF
--- a/tests/kola/kdump/test.sh
+++ b/tests/kola/kdump/test.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+set -xeuo pipefail
+# https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/
+# kola: {"minMemory": 4096, "tags": "skip-base-checks"}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+  "")
+      rhelver=$(. /etc/os-release && echo ${RHEL_VERSION:-})
+      if test -n "${rhelver}"; then
+        rhelminor=$(echo "${rhelver}" | cut -f 2 -d '.')
+        if test '!' -w /boot && test "${rhelminor}" -lt "5"; then
+          mkdir -p /etc/systemd/system/kdump.service.d
+          cat > /etc/systemd/system/kdump.service.d/rw.conf << 'EOF'
+[Service]
+ExecStartPre=mount -o remount,rw /boot
+EOF
+        fi
+      fi
+      rpm-ostree kargs --append='crashkernel=300M'
+      systemctl enable kdump.service
+      /tmp/autopkgtest-reboot setcrashkernel
+      ;;
+  setcrashkernel)
+      /tmp/autopkgtest-reboot-prepare aftercrash
+      echo "Triggering sysrq"
+      sync
+      echo 1 > /proc/sys/kernel/sysrq
+      # This one will trigger kdump, which will write the kernel core, then reboot.
+      echo c > /proc/sysrq-trigger
+      # We shouldn't reach this point
+      sleep 5
+      fatal "failed to invoke sysrq"
+      ;;
+  aftercrash)
+      kcore=$(find /var/crash -type f -name vmcore)
+      if test -z "${kcore}"; then
+        fatal "No kcore found in /var/crash"
+      fi
+      info=$(file ${kcore})
+      if ! [[ "${info}" =~ 'vmcore: Kdump'.*'system Linux' ]]; then
+        fatal "vmcore does not appear to be a Kdump?"
+      fi
+      ;;
+  *) fatal "unexpected mark: ${AUTOPKGTEST_REBOOT_MARK}";;
+esac


cgwalters:
Depends https://github.com/coreos/coreos-assembler/pull/2207

cgwalters:
Lifting draft, this is passing CI now!  (And it was a nice double check that the test successfully failed with older cosa)

tests/kola/kdump/test.sh
```diff
@@ -0,0 +1,49 @@
...
+        rhelminor=$(echo "${rhelver}" | cut -f 2 -d '.')
+        if test '!' -w /boot && test "${rhelminor}" -lt "5"; then
+          mkdir -p /etc/systemd/system/kdump.service.d
+          cat > /etc/systemd/system/kdump.service.d/rw.conf << 'EOF'
+[Service]
+ExecStartPre=mount -o remount,rw /boot
+EOF
+        fi
+      fi
+      rpm-ostree kargs --append='crashkernel=300M' --append=root=/dev/disk/by-label/root
```
jlebon: Why do we need to append `root=`? That should already be in the BLS.

jlebon (dismissed):
Very cool! One comment, LGTM otherwise.

tests/kola/kdump/test.sh
cgwalters: We only append `root=` with complex partitioning, the default is to use `/dev/disk/by-label/root` in our code, right?  That's what I hit that is missing from the current kdump args.

The good thing about having this be a test case is we can just remove that bit and see whether it works or not!

tests/kola/kdump/test.sh
```diff
@@ -0,0 +1,49 @@
...
+      systemctl enable --now kdump.service
```
jlebon: Minor/optional: noticed while testing locally that we don't actually need `--now` here since there's no `crashkernel` yet in the cmdline, so it won't actually run anything.

tests/kola/kdump/test.sh
jlebon: We always append `root=`, even when there is no root reprovisioning.

Hmm, interactively testing this without the `--append=root=...` karg on a local build, the whole flow seems to work fine.

What error were you seeing?

cgwalters: > We always append root=, even when there is no root reprovisioning.

Really?  I'm not seeing that here.

I was seeing the usual kernel error for "failed to find root filesystem".  Which...hmm, now that I think about it means something went wrong in building the initramfs.

cgwalters: Anyways it does work without `root=`, presumably something went wrong generating the kdump initramfs, so that's why I saw the error.  But can't reproduce that now.

jlebon (approved):

tests/kola/kdump/test.sh
jlebon: Hmm, weird. Are you using just `cosa run`?

It's added by `coreos-boot-edit.service`, which always runs regardless of root reprovisoning. Of course, it doesn't actually show up in `/proc/cmdline` until reboot, but `rpm-ostree kargs` on the first boot should show it.

What does `journalctl -u coreos-boot-edit.service` say?

cgwalters: Yes, I was using `cosa run` on a local build.  And hmm yes, I do see it now.  OK, and also I do see it on one of my canary nodes.  

I think I'd forgotten `coreos-boot-edit.service` existed; we're probably overdue to add something about this to https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_disk_layout
or https://github.com/coreos/fedora-coreos-tracker/blob/main/internals/README-internals.md

dustymabe:
This test seems to be failing on our cloud platforms. For example with the lastest `testing-devel` (`34.20210610.20.0`) all three cloud platforms failed:

- [AWS](https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/job/kola-aws/451/)
- [OpenStack](https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/job/kola-openstack/216/)
- [GCP](https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/job/kola-gcp/435/)

bgilbert:
This is also failing CoreOS CI for Ignition.  The `console.txt` of the failed test has a variety of errors in it:

```
[    1.927308] systemd[1]: /usr/lib/systemd/system/kdump-capture.service:23: Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether.
[    1.930253] systemd[1]: /usr/lib/systemd/system/kdump-capture.service:24: Standard output type syslog+console is obsolete, automatically updating to journal+console. Please update your unit file, and consider removing the setting altogether.
[    3.260866] mount[360]: mount: /kdumproot/sysroot: unknown filesystem type 'xfs'.
[    3.324660] systemd[1]: kdumproot-sysroot.mount: Mount process exited, code=exited, status=32/n/a
[    3.330266] systemd[1]: kdumproot-sysroot.mount: Failed with result 'exit-code'.
[    3.338024] systemd[1]: Failed to mount /kdumproot/sysroot.
/lib/kdump-logger.sh: line 91: logger: command not found
[    3.807210] kdump[419] No '/dev/log' or 'logger' included for syslog logging
failed to initiate the kdump logger.
[    3.846470] systemd[1]: kdump-error-handler.service: Main process exited, code=exited, status=1/FAILURE
[    3.853033] systemd[1]: kdump-error-handler.service: Failed with result 'exit-code'.
[    3.972374] systemd[1]: Requested transaction contradicts existing jobs: Transaction for kdump-error-handler.service/start is destructive (ignition-virtio-dump-journal.service has 'start' job queued, but 'stop' is included in transaction).
```

bgilbert:
Tracked in https://github.com/coreos/fedora-coreos-tracker/issues/860 and disabled in https://github.com/coreos/fedora-coreos-config/pull/1053.

PR #1044, opened 2021-06-06
overrides: freeze on selinux-policy-34.8-1.fc34

dustymabe:
The new one causes issues and has taken too long to get a fix.
https://github.com/coreos/fedora-coreos-tracker/issues/850

Let's unblock the lockfile bumper.


jlebon (approved):
Thanks!

PR #1045, opened 2021-06-07
overlay: initramfs teardown: add support for coreos.force_persist_ip karg

### LINKED ISSUE #853: initrd: allow escape hatch to force initrd network config propagation
This would not normally be needed but our downstream (OKD/OCP) has had an issue where they are unable to add new nodes to a cluster without a configuration change to all existing nodes in the cluster. From [the bug](https://bugzilla.redhat.com/show_bug.cgi?id=1958930#c10):

```

    This turns out to be a bigger problem than we originally anticipated. The scenario
    here is that you have a cluster that was installed some time ago. After initial
    install, you realize you need some slight networking tweak (like added static route).
    You add this static route to existing nodes via the MCO (i.e. writing out
    `/etc/sysconfig/network-scripts/route-bond0`).

    Later you try to deploy a new worker node. Since the `route-bond0` file gets
    written out by Ignition during the initramfs, the initramfs network propagation
    code sees "networking configuration" that was written by Ignition and decides to
    not propagate any initramfs networking configuration to the real root (kernel args).
    So you can't deploy new nodes.
```

To get people unstuck in this scenario let's add a kernel argument that can be used to force network propagation.

dustymabe:
This karg will override the check to see if Networking Configuration is
in the real root (i.e. written by Ignition most likely) and force
propagation of initramfs networking anyway. It should only be used in
cases where you know exactly what you are doing. For example, in the
problem described in https://bugzilla.redhat.com/show_bug.cgi?id=1958930
the networking configs didn't conflict at all, thus this could be used
safely. This option will help users who get stuck like they were in that
bug to get unstuck easily and then change their strategy later.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/853

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -70,14 +70,33 @@ are_default_NM_configs() {
 #
 # See https://github.com/coreos/fedora-coreos-tracker/issues/394#issuecomment-599721173
 propagate_initramfs_networking() {
-    # Check the two locations where a user could have provided network configuration
-    # On FCOS we only support keyfiles, but on RHCOS we support keyfiles and ifcfg
+    # Check for any real root config in the two locations where a user could have
+    # provided network configuration. On FCOS we only support keyfiles, but on RHCOS
+    # we support keyfiles and ifcfg
     if [ -n "$(ls -A /sysroot/etc/NetworkManager/system-connections/)" -o \
          -n "$(ls -A /sysroot/etc/sysconfig/network-scripts/)" ]; then
         echo "info: networking config is defined in the real root"
-        echo "info: will not attempt to propagate initramfs networking"
+        realrootconfig=1
     else
         echo "info: no networking config is defined in the real root"
+        realrootconfig=0
+    fi
+
+    # Did the user tell us to force initramfs networking config
+    # propagation even if real root networking config exists?
+    # Hopefully we only need this in rare circumstances.
+    # https://github.com/coreos/fedora-coreos-tracker/issues/853
+    forcepropagate=0
+    if dracut_func getargbool 0 'coreos.force_persist_ip'; then
+        forcepropagate=1
+        echo "info: coreos.force_persist_ip detected: will force network config propagation"
+    fi
+
+    if [ $realrootconfig == 1 -a $forcepropagate == 0 ]; then
+        echo "info: will not attempt to propagate initramfs networking"
+    fi
+
+    if [ $realrootconfig == 0 -o $forcepropagate == 1 ]; then
         if [ -n "$(ls -A /run/NetworkManager/system-connections/)" ]; then
             if are_default_NM_configs; then
                 echo "info: skipping propagation of default networking configs"
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -106,43 +106,6 @@ propagate_initramfs_hostname() {
         return 0
     fi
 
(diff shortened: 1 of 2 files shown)


dustymabe:
also:

```
commit 05765910b42e5dfd79335a794989de3f7ad5d932
Date:   Mon Jun 7 11:16:47 2021 -0400

    overlay: initramfs teardown: delete compat code for hostname propagation
    
    Everything is now newer than the minimimum requirement of NetworkManager 1.26.0.
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -70,14 +70,29 @@ are_default_NM_configs() {
...
 propagate_initramfs_networking() {
-    # Check the two locations where a user could have provided network configuration
-    # On FCOS we only support keyfiles, but on RHCOS we support keyfiles and ifcfg
+    # Check for any real root config in the two locations where a user could have
+    # provided network configuration. On FCOS we only support keyfiles, but on RHCOS
+    # we support keyfiles and ifcfg
     if [ -n "$(ls -A /sysroot/etc/NetworkManager/system-connections/)" -o \
          -n "$(ls -A /sysroot/etc/sysconfig/network-scripts/)" ]; then
         echo "info: networking config is defined in the real root"
-        echo "info: will not attempt to propagate initramfs networking"
```
dustymabe, line 78: I'm sad about losing:

```
        echo "info: will not attempt to propagate initramfs networking"
```

but I couldn't find a way to keep it included without making the conditional logic hairy.

cgwalters: How about keeping it but just saying `info: no network configuration found in /sysroot`?

dustymabe: We do still have a message related to that: `info: no networking config is defined in the real root`. I'm mostly just lamenting about not being able to keep it and make it pretty. This bit of additional logic would work but mostly just extra:

```
if [ $realrootconfig == 1 -a $forcepropagate == 0 ]; then
    echo "info: will not attempt to propagate initramfs networking"
fi
```

dustymabe: ok. I added it anyway in last push. I'll mark this resolved.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -70,14 +70,33 @@ are_default_NM_configs() {
...
+    fi
+
+    # Did the user tell us to force initramfs networking config
+    # propagation even if real root networking config exists?
+    # Hopefully we only need this in rare circumstances.
+    # https://github.com/coreos/fedora-coreos-tracker/issues/853
+    forcepropagate=0
+    if dracut_func getargbool 0 'coreos.force_persist_ip'; then
+        forcepropagate=1
+        echo "info: coreos.force_persist_ip detected: will force network config propagation"
```
darkmuggle, line 92: Overall, this looks fine to me, so
/lgtm

That said, I wonder if it makes sense to put something like
```
cat <<EOM
================================================================
Kernel arg "coreos.force_persist_up=1" should be used rarely
and in exceptional cirucmstances. This is unlikely the bit you
want...
================================================================
EOM
```

darkmuggle (approved):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
dustymabe: Yeah I thought about it. I doubt anyone other than us are looking that close in the logs. I figure we can keep the only place it's documented with a fairly large disclaimer about its use.

PR #1048, opened 2021-06-08
repos: drop `failovermethod` key

jlebon:
This isn't supported by `dnf`. `libdnf` will parse it but it's not
actually hooked up to anything. Recent repo files no longer use this
key. So just nuke it.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1653831


dustymabe (approved):
LGTM

cgwalters:
Agreed, we don't want to fail over this later.

PR #1050, opened 2021-06-10
tests: add test for enabling systemd instance unit

dustymabe:


dustymabe:
I decided to write this test because I noticed the enablement bits just landed in RHEL for it to all work. Is there a way I can force this test to run on RHCOS to (in downstream CI)?

tests/kola/ignition/systemd-enable-instance-unit/test.sh
```diff
@@ -0,0 +1,21 @@
...
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# make sure the presets worked and the instantiated unit is enabled
+if [ $(systemctl is-enabled echo@foo.service) != 'enabled' ]; then
+    fatal "echo@foo.service systemd unit should be enabled"
+fi
+ok "echo@foo.service systemd unit is enabled"
```
jlebon, line 25: Shouldn't we instead/also check that the service ran? E.g. by querying the journal.

jlebon (commented):
> I decided to write this test because I noticed the enablement bits just landed in RHEL for it to all work. Is there a way I can force this test to run on RHCOS to (in downstream CI)?

Not currently, but you can symlink them in openshift/os. See e.g. https://github.com/openshift/os/pull/510.

tests/kola/ignition/systemd-enable-instance-unit/test.sh
dustymabe: added

tests/kola/ignition/systemd-enable-instance-unit/test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# make sure the unit ran and wrote 'foo' to the journal
+if [ $(journalctl -o cat -u echo@foo.service | sed -n 2p) != 'foo' ]; then
```
jlebon, line 28: We could also leverage the built-in `--grep`, e.g. `journalctl -u echo@foo.service --grep "I wrote foo"`. Obviously fine as is too!

jlebon (dismissed):
LGTM

Should we restrict this to the qemu platform? Given that it seems highly unlikely to go wrong based on a specific platform and cloud testing is much more expensive time-wise.

dustymabe:
> Should we restrict this to the qemu platform? Given that it seems highly unlikely to go wrong based on a specific platform and cloud testing is much more expensive time-wise.

done. I'll mark this to auto merge when tests pass

jlebon (approved):

PR #1053, opened 2021-06-11
tests/kdump: disable test

bgilbert:
It's a new test, and is broken on three cloud platforms plus Ignition upstream CI.

https://github.com/coreos/fedora-coreos-config/pull/1043#issuecomment-859028582


arithx (approved):
LGTM, are you planning on creating an issue to track the fix & re-enablement of this test after?

bgilbert:
https://github.com/coreos/fedora-coreos-tracker/issues/860

travier:
Is there any reason this is disabling the test inline and not using the kola denylist? @bgilbert

bgilbert:
I vaguely recall having some sort of trouble with the denylist, but I'm not sure.

PR #1054, opened 2021-06-11
overlay: ignition-ostree-firstboot-uuid: workaround tune2fs issues

### LINKED ISSUE #735: tune2fs error causes ignition-ostree-uuid-boot.service to fail
**Describe the bug**
After installing FCOS on a BHyve VM, the VM gets stuck in a boot-loop. On the emergency-console, I can see the error
messages pasted below: `Failed to start Ignition OSTree: Generate filesystem UUID (boot)`. The error occurs with VirtIO disks as well as with AHCI Disks (UEFI).
Please note that FCOS 31.20200517.3.0-live.x86_64.iso installs and runs fine, however FCOS  >=32.20201104.3.0-live.x86_64.iso fails with the described error.

```
[  253.279062] systemd[1]: Starting dracut pre-udev hook...
Starting dracut pre-udev hook...
[  OK  ] Finished dracut pre-udev hook.
[  253.324905] systemd[1]: Finished dracut pre-udev hook.
         Starting udev Kernel Device Manager...
[  253.330576] systemd[1]: Starting udev Kernel Device Manager...
[  OK  ] Started udev Kernel Device Manager.
[  253.384466] systemd[1]: Started udev Kernel Device Manager.
         Starting Generate new UUID for boot disk GPT...
[  253.389490] systemd[1]: Starting Generate new UUID for boot disk GPT...
         Starting dracut pre-trigger hook...
[  253.400210] systemd[1]: Starting dracut pre-trigger hook...
[  OK  ] Finished Generate new UUID for boot disk GPT.
[  253.420155] systemd[1]: Finished Generate new UUID for boot disk GPT.
         Starting Ignition OSTree: …rate filesystem UUID (boot)...
[  253.425985] systemd[1]: Starting Ignition OSTree: Regenerate filesystem UUID (boot)...

[  253.446333] ignition-ostree-firstboot-uuid[10686]: e2fsck 1.45.5 (07-Jan-2020)
[  253.453487] ignition-ostree-firstboot-uuid[10686]: Pass 1: Checking inodes, blocks, and sizes
[  253.461100] ignition-ostree-firstboot-uuid[10686]: Pass 2: Checking directory structure
[  253.470285] dracut-pre-trigger[10677]: rd.md=0: removing MD RAID activation
[  253.478754] ignition-ostree-firstboot-uuid[10686]: Pass 3: Checking directory connectivity
[  253.485191] ignition-ostree-firstboot-uuid[10686]: Pass 4: Checking reference counts
[FAILED] Failed to start Ignition O…nerate filesystem UUID (boot).
See 'systemctl status ignition-ostree-uuid-boot.service' for details.
[DEPEND] Dependency failed for Ignition Boot Disk Set
```

**Reproduction steps**
1. Create a VM in TrueNAS (UEFI), VirtIO Disk or AHCI
2. Boot Live-ISO as CDROM
3. Retrieve ignition-file via `curl -LO`
4. Install with `sudo coreos-installer install /dev/vda --ignition fcos.ign`

**Expected behavior**
FCOS / VM booting to shell
A clear and concise description of what actually happened.

**Actual behavior**
FCOS rebooting due to error while creating Filesystem UUID

**System details**
 - TrueNAS 12.0-U1.1 BHyve VM
 - Bare Metal live.x86_64.iso
 - FCOS  >=32.20201104.3.0

**Ignition config**
```
variant: fcos
version: 1.0.0
passwd:
  users:
    - name: core
      ssh_authorized_keys:
```

**Additional information**
Add any other information about the problem here.

dustymabe:
EDIT: see updated description in https://github.com/coreos/fedora-coreos-config/pull/1054#issuecomment-861034711

We're hitting issues in some cases where users are seeing the
`tune2fs -U random` on the /boot filesystem fail even though it's
ran directly after an `e2fsck`. To prevent needing a filesystem
check at all let's store the metadata checksum seed in the superblock
which will allow us to change the UUID without needing any filesystem
checks.

This was not possible before but we recently got a grub fix so that
having this filesystem feature enabled wouldn't cause grub2 to throw
an error.

- https://lists.gnu.org/archive/html/grub-devel/2021-06/msg00031.html

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/735

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
@@ -27,10 +27,7 @@ esac
 
 if [ "${TYPE}" == "${orig_type}" ] && [ "${UUID}" == "${orig_uuid}" ]; then
   case "${TYPE}" in
-    # For now we need to fsck first, see https://github.com/coreos/coreos-assembler/pull/1452
-    # Basically we're not passing `metadata_csum_seed` as a mkfs.ext4 option
-    # because grub2 barfs on it.
-    ext4) e2fsck -fy "${target}" && tune2fs -U random "${target}" ;;
+    ext4) tune2fs -O metadata_csum_seed -U random "${target}" ;;
     xfs) xfs_admin -U generate "${target}" ;;
     *) echo "unexpected filesystem type ${TYPE}" 1>&2; exit 1 ;;
   esac
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
@@ -27,7 +27,37 @@ esac
 
 if [ "${TYPE}" == "${orig_type}" ] && [ "${UUID}" == "${orig_uuid}" ]; then
   case "${TYPE}" in
-    ext4) tune2fs -O metadata_csum_seed -U random "${target}" ;;
+    ext4)
+          # If the filesystem supports metadata_csum_seed then the UUID is stored
+          # in the superblock and there is no need to worry with an fsck. For the
+          # boot filesystem this FS feature wasn't supported by GRUB until recently.
+          # https://lists.gnu.org/archive/html/grub-devel/2021-06/msg00031.html
+          # Once grub is updated in all systems we care about we can standardize
+          # on the metadata_csum_seed and delete the `else` code block.
+          if tune2fs -l ${target} | grep 'metadata_csum_seed'; then
+              tune2fs -U random "${target}"
+          else
+              # Run an fsck since tune2fs -U requires the FS to be clean
+              e2fsck -fy "${target}"
+              # We just ran an fsck, but there is a bug where tune2fs -U will still
+              # complain. It will still error if the last checked timestamp (just
+              # set by the e2fsck above) is older than the last mount timestamp (happens
+              # on systems with out of date or non-functioning hardware clocks).
+              # See https://github.com/coreos/fedora-coreos-tracker/issues/735#issuecomment-859605953
+              # Potentially fixed in future by: https://www.spinics.net/lists/linux-ext4/msg78012.html
+              tune2fsinfo="$(tune2fs -l ${target})"
+              lastmount=$(echo "$tune2fsinfo" | grep '^Last mount time:' | cut -d ':' -f 2,3,4)
+              lastfsck=$(echo "$tune2fsinfo" | grep '^Last checked:' | cut -d ':' -f 2,3,4)
+              lastmountsse=$(date --date="$lastmount" +%s)
+              lastfscksse=$(date --date="$lastfsck" +%s)
+              if (( lastfscksse < lastmountsse )); then
+                  echo "Detected timestamp of last fsck is older than timestamp of last mount."
+                  echo "Setting "${target}" timestamp of last fsck to same time as last mount."
+                  tune2fs -T $(date --date="$lastmount" +%Y%m%d%H%M%S) "${target}"
+              fi
+              # Finally, we can randomize the UUID
+              tune2fs -U random "${target}"
+          fi ;;
     xfs) xfs_admin -U generate "${target}" ;;
     *) echo "unexpected filesystem type ${TYPE}" 1>&2; exit 1 ;;
   esac


dustymabe:
OK I opened this PR to start some discussion. First off, once we have the grub patch in a Fedora RPM this patch should be sufficient, but...  I'm wondering if we also want something like this in coreos-assembler too:

```diff
diff --git a/src/create_disk.sh b/src/create_disk.sh
index 074db967..29583320 100755
--- a/src/create_disk.sh
+++ b/src/create_disk.sh
@@ -156,12 +156,12 @@ udevtrig
 
 root_dev="${disk}${ROOTPN}"
 
-bootargs=
+bootargs="-O metadata_csum_seed"
 case "${bootfs}" in
     ext4verity)
         # Need blocks to match host page size; TODO
         # really mkfs.ext4 should know this.
-        bootargs="-b $(getconf PAGE_SIZE) -O verity"
+        bootargs=" -b $(getconf PAGE_SIZE) -O verity"
         ;;
     ext4) ;;
     *) echo "Unhandled bootfs: ${bootfs}" 1>&2; exit 1 ;;
```


Additionally, there's another piece to this puzzle: RHCOS. When can we use this there? Does `grub2` from the `coreos-assembler` image get used when creating those images (i.e. do we use the Fedora `grub2` in RHCOS because COSA is Fedora based?).

dustymabe:
grub rpm build with the needed patch is https://bodhi.fedoraproject.org/updates/FEDORA-2021-f2d9c76e68

jlebon:
If we do it at `mkfs.ext4` time, then we don't need to enable it here too, right? So e.g. it'd just be `tune2fs -U random "${target}"` here.

> Additionally, there's another piece to this puzzle: RHCOS. When can we use this there? Does grub2 from the coreos-assembler image get used when creating those images (i.e. do we use the Fedora grub2 in RHCOS because COSA is Fedora based?).

Yes, exactly. It'll build fine, but grub will fail to boot until it gets the patch IIUC.

Hmm, have we actually hit this in RHCOS land? It sounds like it's restricted to systems which don't have clock batteries (e.g. RPi4) or hypervisors which don't initialize VMs with the host system clock? Does OCP support any platforms which have these characteristics? If not, then maybe simplest for now is to just make it an `image.yaml` knob and enable it for FCOS only. Once the GRUB patch makes it into RHEL, we can enable it unconditionally. WDYT?

dustymabe:
> If we do it at `mkfs.ext4` time, then we don't need to enable it here too, right? So e.g. it'd just be `tune2fs -U random "${target}"` here.

I think you are right, but I'm not 100% sure. Basically if the metadata_csum_seed is already enabled in the FS will `tune2fs -U random "${target}"` do the right thing or do we still need to specify `-O metadata_csum_seed` or not?

> 
> > Additionally, there's another piece to this puzzle: RHCOS. When can we use this there? Does grub2 from the coreos-assembler image get used when creating those images (i.e. do we use the Fedora grub2 in RHCOS because COSA is Fedora based?).
> 
> Yes, exactly. It'll build fine, but grub will fail to boot until it gets the patch IIUC.

But what version of grub gets installed when `grub2-install` is used  (the fedora or rhel version) when built using `coreos-assembler`? I do think it would be wrong to rely on that, though. As the bootloader might get updated later.

> 
> Hmm, have we actually hit this in RHCOS land?

I think at least [one user was reporting that](https://github.com/coreos/fedora-coreos-tracker/issues/735#issuecomment-858121932), but probably not necessarily on a prod platform.

> It sounds like it's restricted to systems which don't have clock batteries (e.g. RPi4) or hypervisors which don't initialize VMs with the host system clock? Does OCP support any platforms which have these characteristics? If not, then maybe simplest for now is to just make it an `image.yaml` knob and enable it for FCOS only. Once the GRUB patch makes it into RHEL, we can enable it unconditionally. WDYT?

Instead of an `image.yaml` knob we can just have this code (in this PR) detect RHCOS vs FCOS and run the old code vs the new code.

cgwalters:
> Additionally, there's another piece to this puzzle: RHCOS. When can we use this there? Does grub2 from the coreos-assembler image get used when creating those images (i.e. do we use the Fedora grub2 in RHCOS because COSA is Fedora based?).

It's messier than that, see
https://github.com/coreos/bootupd/issues/53#issuecomment-705183943
Basically for UEFI we use target grub, for BIOS we use Fedora grub.

cgwalters:
> Instead of an image.yaml knob we can just have this code (in this PR) detect RHCOS vs FCOS and run the old code vs the new code.

I think it'd be least fragile to not detect FCOS vs RHCOS but instead to check the target grub version.  That way e.g. building older FCOS with new assembler will still work, and it will also work if we introduce C9S CoreOS, etc etc.

cgwalters:
Also I think this is a good fix, but I also would like if we *additionally* pursued https://github.com/coreos/fedora-coreos-tracker/issues/735#issuecomment-860688381

dustymabe:
> I think it'd be least fragile to not detect FCOS vs RHCOS but instead to check the target grub version.

Yeah. The only problem is that I'm not sure we'll be able to use grub version as a good indicator. Grub's releases don't come often and this would most likely get put in as a backport if it ever made it to RHEL so the conditional logic would probably still need to have some Fedora vs RHEL logic.

> Also I think this is a good fix, but I also would like if we additionally pursued coreos/fedora-coreos-tracker#735 (comment)

I'm interested in this path, but might run out of time to fully investigate it. Unfortunately testing this requires me to physically touch hardware which has long test cycles and I'm moving everything in the next few days to a new house.

cgwalters:
> Yeah. The only problem is that I'm not sure we'll be able to use grub version as a good indicator. Grub's releases don't come often and this would most likely git put in as a backport if it ever made it to RHEL so the conditional logic would probably still need to have some Fedora vs RHEL logic.

Yes, but "Fedora vs RHEL" phrasing here seems to implicitly assume RHEL=8 which is not correct.  (Also again it doesn't account for CentOS Stream)/  In a slightly more ideal world we wouldn't have a gigantic pile of grub2 patches, and so we could compare the upstream version.  

But yes, we may end up with a distro table.  I'd just key off disttag or so, to account for the above (e.g. `el8` vs `el9`, etc.)

jlebon:
> > If we do it at `mkfs.ext4` time, then we don't need to enable it here too, right? So e.g. it'd just be `tune2fs -U random "${target}"` here.
> 
> I think you are right, but I'm not 100% sure. Basically if the metadata_csum_seed is already enabled in the FS will `tune2fs -U random "${target}"` do the right thing or do we still need to specify `-O metadata_csum_seed` or not?

Yes, that's my understanding, though one should be able to test this out pretty easily.

> > It sounds like it's restricted to systems which don't have clock batteries (e.g. RPi4) or hypervisors which don't initialize VMs with the host system clock? Does OCP support any platforms which have these characteristics? If not, then maybe simplest for now is to just make it an `image.yaml` knob and enable it for FCOS only. Once the GRUB patch makes it into RHEL, we can enable it unconditionally. WDYT?
> 
> Instead of an `image.yaml` knob we can just have this code (in this PR) detect RHCOS vs FCOS and run the old code vs the new code.

Or maybe both? Have an `image.yaml` knob which drives whether we `mkfs` with `metadata_csum_seed` at cosa build time. Then here, we just need to check e.g. `tune2fs -l /dev/disk/by-label/boot | grep metadata_csum_seed` to know which path to take.

dustymabe:
> 
> Or maybe both? Have an `image.yaml` knob which drives whether we `mkfs` with `metadata_csum_seed` at cosa build time. Then here, we just need to check e.g. `tune2fs -l /dev/disk/by-label/boot | grep metadata_csum_seed` to know which path to take.

Makes sense. Updated to do it this way.

dustymabe:
OK I pushed up another version of this PR (still needs a bit of local testing). Here are the updated commit messages:

```
commit 69538edb70e787e65ef22ea5050443009387963f
Date:   Mon Jun 14 18:19:14 2021 -0400

    overlay: ignition-ostree-firstboot-uuid: add workaround for FS without metadata_csum_seed
    
    It will be some time before we can support metadata_csum_seed in grub
    on all our target platforms. This commit adds a workaround for the case
    where grub doesn't support it yet. We specifically workaround the
    behavior of tune2fs that isn't sensitive to timing inconsistencies.
    
    See https://github.com/coreos/fedora-coreos-tracker/issues/735#issuecomment-859605953
    
    Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/735

commit cbf04c66fb4701260106ce62c1e06f3ef563cc80
Date:   Fri Jun 11 16:18:19 2021 -0400

    overlay: ignition-ostree-firstboot-uuid: use metadata_csum_seed fs feature
    
    We're hitting issues in some cases where users are seeing the
    `tune2fs -U random` on the /boot filesystem fail even though it's
    ran directly after an `e2fsck`. To prevent needing a filesystem
    check at all let's store the metadata checksum seed in the superblock
    which will allow us to change the UUID without needing any filesystem
    checks.
    
    This was not possible before but we recently got a grub fix so that
    having this filesystem feature enabled wouldn't cause grub2 to throw
    an error.
    
    - https://lists.gnu.org/archive/html/grub-devel/2021-06/msg00031.html
    
    Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/735

```

dustymabe:
With this new approach we can switch over COSA once the new grub has made it in (don't need to rush it).

dustymabe:
Example output from new approach (running on rpi4 where I originally observed the problem):

```
[core@rpi4 ~]$ journalctl -t ignition-ostree-firstboot-uuid
-- Journal begins at Tue 2021-05-25 11:15:20 UTC, ends at Tue 2021-06-15 12:16:00 UTC. --
May 25 11:16:01 localhost ignition-ostree-firstboot-uuid[487]: e2fsck 1.45.6 (20-Mar-2020)
May 25 11:16:01 localhost ignition-ostree-firstboot-uuid[487]: Pass 1: Checking inodes, blocks, and sizes
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[487]: Pass 2: Checking directory structure
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[487]: Pass 3: Checking directory connectivity
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[487]: Pass 4: Checking reference counts
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[487]: Pass 5: Checking group summary information
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[487]: boot: 381/98304 files (0.0% non-contiguous), 113828/393216 blocks
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[482]: Detected timestamp of last fsck is older than timestamp of last mount.
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[482]: Setting /dev/disk/by-label/boot timestamp of last fsck to same time as last mount.
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[500]: tune2fs 1.45.6 (20-Mar-2020)
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[500]: Setting time filesystem last checked to Tue Jun 15 11:50:17 2021
May 25 11:16:03 localhost ignition-ostree-firstboot-uuid[501]: tune2fs 1.45.6 (20-Mar-2020)
May 25 11:16:03 localhost ignition-ostree-firstboot-uuid[482]: Regenerated UUID for /dev/disk/by-label/boot
May 25 11:16:10 localhost ignition-ostree-firstboot-uuid[572]: Clearing log and setting UUID
May 25 11:16:10 localhost ignition-ostree-firstboot-uuid[572]: writing all SBs
May 25 11:16:10 localhost ignition-ostree-firstboot-uuid[572]: new UUID = 35a5f882-03df-4af0-b0c4-00827b4bc107
May 25 11:16:10 localhost ignition-ostree-firstboot-uuid[569]: Regenerated UUID for /dev/disk/by-label/root
```

The interesting line is (notice the different times reported at the beginning and end of the line):

```
May 25 11:16:02 localhost ignition-ostree-firstboot-uuid[500]: Setting time filesystem last checked to Tue Jun 15 11:50:17 2021
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
```diff
@@ -27,7 +27,37 @@ esac
...
+              tune2fsinfo="$(tune2fs -l ${target})"
+              lastmount=$(echo "$tune2fsinfo" | grep '^Last mount time:' | cut -d ':' -f 2,3,4)
+              lastfsck=$(echo "$tune2fsinfo" | grep '^Last checked:' | cut -d ':' -f 2,3,4)
+              lastmountsse=$(date --date="$lastmount" +%s)
+              lastfscksse=$(date --date="$lastfsck" +%s)
+              if (( lastfscksse < lastmountsse )); then
+                  echo "Detected timestamp of last fsck is older than timestamp of last mount."
+                  echo "Setting "${target}" timestamp of last fsck to same time as last mount."
+                  tune2fs -T $(date --date="$lastmount" +%Y%m%d%H%M%S) "${target}"
+              fi
```
jlebon, line 57: Cool!

Hmm, I guess another way we could workaround this would be to do:

```sh
mount -o ro "${target}" /mnt && umount /mnt
e2fsck -fy "${target}"
```

right? (And we use `MountFlags=slave`.) That way whatever the system time is, the last mount will always be before the last fsck.

Totally cool with this approach too though.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
```diff
@@ -27,7 +27,37 @@ esac
...
-    ext4) tune2fs -O metadata_csum_seed -U random "${target}" ;;
+    ext4)
+          # If the filesystem supports metadata_csum_seed then the UUID is stored
+          # in the superblock and there is no need to worry with an fsck. For the
+          # boot filesystem this FS feature wasn't supported by GRUB until recently.
+          # https://lists.gnu.org/archive/html/grub-devel/2021-06/msg00031.html
+          # Once grub is updated in all systems we care about we can standardize
+          # on the metadata_csum_seed and delete the `else` code block.
+          if tune2fs -l ${target} | grep 'metadata_csum_seed'; then
+              tune2fs -O metadata_csum_seed -U random "${target}"
```
jlebon: OK yup, should be able to drop `-O metadata_csum_seed` here as discussed. Here's a quick local test:

```
[root@cosa-devsh ~]# mkfs.ext4 -F /dev/vda
[root@cosa-devsh ~]# mount /dev/vda /mnt
[root@cosa-devsh ~]# umount /mnt
[root@cosa-devsh ~]# tune2fs -U random /dev/vda
tune2fs 1.45.6 (20-Mar-2020)

This operation requires a freshly checked filesystem.

Please run e2fsck -f on the filesystem.
```

With the option at `mkfs.ext4` time only:

```
[root@cosa-devsh ~]# mkfs.ext4 -F -O metadata_csum_seed /dev/vda
[root@cosa-devsh ~]# mount /dev/vda /mnt
[root@cosa-devsh ~]# umount /mnt
[root@cosa-devsh ~]# tune2fs -l /dev/vda | grep UUID
Filesystem UUID:          fc9c6058-c33e-4c38-bfe6-85521d4b808d
[root@cosa-devsh ~]# tune2fs -U random /dev/vda
tune2fs 1.45.6 (20-Mar-2020)
[root@cosa-devsh ~]# tune2fs -l /dev/vda | grep UUID
Filesystem UUID:          83a6d0e1-628b-4cf7-82bc-f7085d99aabc
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
dustymabe: Yeah that probably would work. I'm going to leave what I have for now since I've already tested it.

dustymabe:
ok pushed up a new (hopefully final) version of the code. I'm waiting for a grub build to complete and then I'll test that path (`metadata_csum_seed`) on the rpi4 too.

dustymabe:
Now ran a test with `grub2-common-2.06-2.fc34.noarch` in COSA and in the FCOS build:

```
-- Journal begins at Tue 2021-05-25 11:15:21 UTC, ends at Tue 2021-06-15 17:00:21 UTC. --
May 25 11:15:57 localhost ignition-ostree-firstboot-uuid[487]: Filesystem features:      has_journal ext_attr resize_inode dir_index filetype extent 64bit flex_bg metadata_csum_seed sparse_super large_file huge_file dir_nlink extra_isize metadata_csum
May 25 11:15:59 localhost ignition-ostree-firstboot-uuid[488]: tune2fs 1.45.6 (20-Mar-2020)
May 25 11:15:59 localhost ignition-ostree-firstboot-uuid[483]: Regenerated UUID for /dev/disk/by-label/boot
May 25 11:16:06 localhost ignition-ostree-firstboot-uuid[557]: Clearing log and setting UUID
May 25 11:16:06 localhost ignition-ostree-firstboot-uuid[557]: writing all SBs
May 25 11:16:06 localhost ignition-ostree-firstboot-uuid[557]: new UUID = 8cb4017f-8698-4ea7-be5b-54b2ed209538
May 25 11:16:06 localhost ignition-ostree-firstboot-uuid[554]: Regenerated UUID for /dev/disk/by-label/root
```

Everything seems to be working!

jlebon (approved):
Thanks a lot for digging into this and pushing it through!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
```diff
@@ -27,10 +27,37 @@ esac
...
+              lastfsck=$(echo "$tune2fsinfo" | grep '^Last checked:' | cut -d ':' -f 2,3,4)
+              lastmountsse=$(date --date="$lastmount" +%s)
+              lastfscksse=$(date --date="$lastfsck" +%s)
+              if (( lastfscksse < lastmountsse )); then
+                  echo "Detected timestamp of last fsck is older than timestamp of last mount."
+                  echo "Setting "${target}" timestamp of last fsck to same time as last mount."
+                  tune2fs -T $(date --date="$lastmount" +%Y%m%d%H%M%S) "${target}"
+              fi
+              # Finally, we can randomize the UUID
+              tune2fs -U random "${target}"
```
jlebon, line 59: I guess we could reflow this as:

```sh
if ! tune2fs -l ${target} | grep 'metadata_csum_seed'; then
  <timestamp workaround>
fi
tune2fs -U random "${target}"
```

But meh... not worth the respin.

dustymabe: Yeah I did consider it, but I had previously tested it and it would have taken some rework of the comments 🤓

PR #1055, opened 2021-06-12
[next-devel] manifest.yaml: stop using -next repos

jlebon:
Both testing and next are on the same release now, so we can go back to
the stable repos for next-devel.

That stream isn't being built right now, but remove-graduated-overrides
still wants to try to check it, and I'd rather let it than disable it
for next-devel to be consistent and reduce churn whenever we flip the
switch. It's currently failing because those repos are no longer valid
for f34.

Also, it's possible we re-enable the stream anyway for rolling out some
of the systemd-oomd and swap suggestions.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -9,11 +9,12 @@ rojig:
   summary: Fedora CoreOS next-devel
 
 repos:
-  # these repos are there to make it easier to add new packages to the OS and to
+  # These repos are there to make it easier to add new packages to the OS and to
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

jlebon:
Ahh heh, this is failing CI because the base lockfile is out of date and missing latest Ignition needed for the multipath test.

Leaning towards merging on red since the branch isn't being maintained right now anyway. Just looking to make the new GitHub Action happy.

PR #1062, opened 2021-06-15
Revert "overlay.d/09misc: Fix mode for some files in /etc"

travier:
This is now safe to revert as this has been included in a testing
release with a barrier and will thus also be in the next stable release:
https://github.com/coreos/fedora-coreos-streams/pull/328

This reverts commit 73145e08b856ec59906e2256e200d626eba75097.


travier:
Marking hold has I have not yet fully checked the status of the latest releases, etc.

jlebon (approved):

dustymabe:
@travier - what's the status here?

travier:
I have not yet had the time to check if everything is good now.

travier:
I have now checked the latest releases and all my FCOS systems and they are in order so I'd say we're good to go here.

jlebon (approved):

PR #1063, opened 2021-06-15
image-base.yaml: set bootfs_metadata_csum_seed to true

dustymabe:
This will tell COSA to set the metadata_csum_seed filesystem feature
for the boot filesystem. This helps us not have to run a filesystem
check before running tune2fs to randomize the filesystem UUID on first boot. See
https://github.com/coreos/fedora-coreos-tracker/issues/735 for more context.

Depends on https://github.com/coreos/coreos-assembler/pull/2228

### DIFF
--- a/image-base.yaml
+++ b/image-base.yaml
@@ -21,6 +21,15 @@ ostree-remote: fedora
 # https://github.com/ostreedev/ostree/issues/1265
 sysroot-readonly: true
 
+# opt in to using the `metadata_csum_seed` feature of the ext4 filesystem
+# for the /boot filesystem. Support for this was only recently added to grub
+# and isn't available everywhere yet so we'll gate it behind this image.yaml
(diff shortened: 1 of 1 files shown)


dustymabe:
Opening this PR now so I don't forget about it later.

We should be able to merge this once [the new grub2 package](https://bodhi.fedoraproject.org/updates/FEDORA-2021-4dcf124136) has landed but let's hold even longer (for a release or two) to give the fallback codepath (in https://github.com/coreos/fedora-coreos-config/pull/1054) some time to get exercised since that the path RHCOS will take for a while.

cgwalters (approved):
Seems OK to me, but I'd personally still prefer the "set the system clock if older than the filesystem on boot".  OTOH that's an uglier patch, and this filesystem flag sounds useful.

PR #1064, opened 2021-06-18
tests: add test for watching files labeled kubernetes_file_t

miabbott:
Adds a test for the ability of `systemd` to watch files                                                                                                    
labeled with `kubernetes_file_t`.                                                                                                                          
                                                                                                                                                               
See: https://github.com/coreos/fedora-coreos-tracker/issues/861                                                                                            
See: https://github.com/containers/container-selinux/issues/135

### DIFF
--- a/tests/kola/misc-ign-ro/test.sh
+++ b/tests/kola/misc-ign-ro/test.sh
@@ -36,3 +36,25 @@ if [ "$context" != "system_u:object_r:net_conf_t:s0" ]; then
     fatal "SELinux context on stub-resolv.conf is wrong"
 fi
 ok "SELinux context on stub-resolv.conf is correct"
+
+# This is for verifying that `kubernetes_file_t` labeled files can be
+# watched by systemd
+# See: https://github.com/coreos/fedora-coreos-tracker/issues/861
+# See: https://github.com/containers/container-selinux/issues/135
+if [ "$(systemctl is-active kube-watch.path)" != "active" ]; then
+    fatal "kube-watch.path did not activate successfully"
+fi
+ok "kube-watch.path successfully activated"
+
+touch /etc/kubernetes/kubeconfig
+ok "successfully created /etc/kubernetes/kubeconfig"
+
+if [ "$(systemctl is-active kube-watch.service)" != "active" ]; then
+    fatal "kube-watch.service did not successfully activate"
+fi
+ok "kube-watch.service activated successfully"
+
+if [ "$(journalctl -o cat -u kube-watch.service | sed -n 2p)" != "Found it" ]; then
+    fatal "kube-watch.service did not print message to journal"
+fi
+ok "Found message from kube-watch.service in journal"
--- a/tests/kola/misc-ign-ro/config.fcc
+++ b/tests/kola/misc-ign-ro/config.fcc
@@ -1,6 +1,12 @@
 variant: fcos
-version: 1.1.0
+version: 1.2.0
 storage:
+  directories:
+    # This is for verifying that `kubernetes_file_t` labeled files can be
+    # watched by systemd
+    # See: https://github.com/coreos/fedora-coreos-tracker/issues/861
+    # See: https://github.com/containers/container-selinux/issues/135
+    - path: /etc/kubernetes
   files:
     - path: /etc/systemd/zram-generator.conf
       mode: 0644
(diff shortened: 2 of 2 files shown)


miabbott:
This might be overkill but I was thinking we might want to expand our SELinux coverage in the future with tests like this.

Maybe even moving the existing SELinux tests from kola/mantle into the `ext.config` format

tests/kola/selinux/kubernetes_file_t/test.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+# verify the service didn't fail
+if [ $(systemctl is-failed kube-env.service) != 'active' ]; then
+    fatal "kube-env.service failed unexpectedly"
+fi
+ok "kube-env.service successfully started"
+
+# make sure the unit ran and wrote 'foo' to the journal
+if [ $(journalctl -o cat -u echo@foo.service | sed -n 2p) != 'FCOS' ]; then
```
dustymabe:
```suggestion
if [ $(journalctl -o cat -u kube-env.service | sed -n 2p) != 'FCOS' ]; then
```

dustymabe:
Seems reasonable to me.

jlebon (commented):
I think having a test for this here is totally fine, but personally think the cost of starting a whole VM for such a simple check is high. This is why I've been pushing for https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola/misc-ign-ro -- WDYT about folding that in there?

miabbott:
> I think having a test for this here is totally fine, but personally think the cost of starting a whole VM for such a simple check is high. This is why I've been pushing for https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola/misc-ign-ro -- WDYT about folding that in there?

Yup, I was concerned about the cost of a VM for a simple check.  I can fold this into that location.

dustymabe:
> I think having a test for this here is totally fine, but personally think the cost of starting a whole VM for such a simple check is high. This is why I've been pushing for https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola/misc-ign-ro -- WDYT about folding that in there?

This triggered a memory of mine and I found the old discussion and started a new feature request to bring the best of both worlds together: https://github.com/coreos/fedora-coreos-config/issues/1067

miabbott:
Moved the original test for reading `kubernetes_file_t` and pulled in another test for the issue in coreos/fedora-coreos-tracker#861

cgwalters:
Looks sane to me but the test is failing.

miabbott:
> Looks sane to me but the test is failing.

Yeah, these will both fail until we have a new `container-selinux` with an updated policy.

miabbott:
The `kube-env` part of the test is still failing with an SELinux denial.

@dustymabe if you want, I can split that test out into another PR if you would like to get the "watch" part of the test merged in.

dustymabe:
ahh - sorry for the noise. don't worry about splitting it out (unless you want to 🙂)

miabbott:
> ahh - sorry for the noise. don't worry about splitting it out (unless you want to slightly_smiling_face)

split out the `kube-env` test...this passes with latest from `testing-devel` now

tests/kola/misc-ign-ro/test.sh
```diff
@@ -36,3 +36,27 @@ if [ "$context" != "system_u:object_r:net_conf_t:s0" ]; then
...
+    fatal "kube-watch.path did not activate successfully"
+fi
+ok "kube-watch.path successfully activated"
+
+if [ "$(touch /etc/kubernetes/kubeconfig)" ]; then
+    fatal "Unable to create /etc/kubernetes/kubeconfig"
+fi
+ok "successfully created /etc/kubernetes/kubeconfig"
+
+if [ "$(systemctl is-active kube-watch.service)" != "active" ]; then
```
dustymabe, line 52: what's the point of running `systemctl is-active kube-watch.service` twice?

tests/kola/misc-ign-ro/test.sh
```diff
@@ -36,3 +36,27 @@ if [ "$context" != "system_u:object_r:net_conf_t:s0" ]; then
...
+# See: https://github.com/containers/container-selinux/issues/135
+if [ "$(systemctl is-active kube-watch.path)" != "active" ]; then
+    fatal "kube-watch.path did not activate successfully"
+fi
+ok "kube-watch.path successfully activated"
+
+if [ "$(touch /etc/kubernetes/kubeconfig)" ]; then
+    fatal "Unable to create /etc/kubernetes/kubeconfig"
+fi
+ok "successfully created /etc/kubernetes/kubeconfig"
```
dustymabe: this bit is confusing to me. I'm not sure we need an `if` statement at
all because we have `set -eux`.

```suggestion
touch /etc/kubernetes/kubeconfig
ok "successfully created /etc/kubernetes/kubeconfig"
```

tests/kola/misc-ign-ro/test.sh
miabbott: The first `systemctl is-active` is checking `kube-watch.path`

dustymabe (approved):
LGTM

PR #1065, opened 2021-06-18
Revert "overlay: add new module for installing Azure udev rules"

miabbott:
This reverts commit 8c07b7391473910ba3884ee0d3763743805ac78f.

The `WALinuxAgent-udev` package now installs the udev rules in the
initrd.


miabbott:
I verified that the udev rules are in the initrd, but didn't go as far as testing in Azure proper.

jlebon (approved):

jlebon:
Aren't we relying on this right now in RHCOS though? Or did the fixed `WALinuxAgent` make it there too already?

dustymabe:
Would be nice to have at least one person run it on azure.

dustymabe:
> Aren't we relying on this right now in RHCOS though? Or did the fixed `WALinuxAgent` make it there too already?

If true, might be a candidate for moving that RHCOS specific bit into openshift/os ?

miabbott:
> > Aren't we relying on this right now in RHCOS though? Or did the fixed `WALinuxAgent` make it there too already?
> 
> If true, might be a candidate for moving that RHCOS specific bit into openshift/os ?

Right, the `-udev` package is available in RHEL8, but it doesn't have the script to install the rules in the initrd.

We'd want to carry this deleted code in RHCOS only for now.

bgilbert:
Closing as duplicate of #1686.

PR #1077, opened 2021-06-28
coreos-propagate-multipath-conf: various minor tweaks

jlebon:
See individual commits.


dustymabe (approved):
trivial LGTM

nikita-dubrovskii:
Checked on s390x with multipath enabled at install times - it works. 
LGTM

PR #1080, opened 2021-07-01
Add extensions.yaml (subset of RHCOS)

cgwalters:
This is a subset of the `extensions.yaml` from
https://github.com/openshift/os

My immediate motivation is that day to day I develop in
a Fedora toolbox targeting FCOS, and I wanted to test
`rpm-ostree compose extensions` as `cosa buildextend-extensions`
invokes it, and it's easier if we have a file here.

Since the FCOS pipeline doesn't invoke `cosa buildextend-extensions`,
this will be a no-op.

Or stated more strongly: We might just use this as part of e.g.
rpm-ostree's CI runs.

But...one could imagine that we actually share the list of extensions
between FCOS and RHCOS in the future.

Or more obviously, perhaps we actually do something with this FCOS
extensions list and don't actually offer "all Fedora packages"
as extensions.

### DIFF
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -1,0 +1,32 @@
+# RPMs as operating system extensions, distinct from the base ostree commit/image
+# https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/extensions.md
+# and https://github.com/coreos/fedora-coreos-tracker/issues/401
+
+# No additional repositories for Fedora
+repos: []
+
+extensions:
+  # https://github.com/coreos/fedora-coreos-tracker/issues/326
+  usbguard:
+    packages:
+      - usbguard
+  # These are already in the base, so they're not OS extensions, but they're
+  # useful to have in RPM form to install in kmod build containers.
+  kernel:
+    kind: development
+    packages:
+      - kernel
+      - kernel-core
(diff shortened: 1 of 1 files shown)


extensions.yaml
```diff
@@ -0,0 +1,32 @@
+# RPMs as operating system extensions, distinct from the base ostree commit/image
```
jlebon, line 1: Maybe let's add a big comment here that we're not actually officially using this in FCOS and it's just for development?

jlebon (dismissed):
Weak LGTM. It's a no-op, though I'd argue there's still a non-zero cost to carry this here when we're not actually using it.

Will let @dustymabe or @bgilbert chime in.

jlebon:
How much of the developer pain is actually due to https://github.com/openshift/os/issues/552 and so not being able to just build RHCOS in the same toolbox on which you hack on rpm-ostree?

bgilbert:
Hmm, if we don't (yet) have a concrete use for this, I'd say I'm weakly opposed.

LorbusChris:
FWIW we do support extensions in OKD, so maybe it can be put into [okd-machine-os](https://github.com/openshift/okd-machine-os/) for now?

PR #1083, opened 2021-07-05
40ignition-ostree: explicitly add zram kmod in initrd

jlebon:
We were relying on the zram kmod already being in the initramfs so far,
though that assumption for whatever reason is now incorrect in rawhide,
causing rootfs reprovisioning related tests to fail.

Anyway, we should be more explicit here about what our requirements are.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -15,6 +15,11 @@ install_ignition_unit() {
     systemctl -q --root="$initdir" add-requires "ignition-${target}.target" "$unit" || exit 1
 }
 
+installkernel() {
+    # Used by ignition-ostree-transposefs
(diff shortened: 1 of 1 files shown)


travier:
From https://man7.org/linux/man-pages/man7/dracut.modules.7.html:
> instmods <kernelmodule> [ <kernelmodule> ... ]
>           instmods should be used only in the installkernel() function.

travier:
I think you need something like:
```
# called by dracut
installkernel() {
    instmods foo
}
```

travier:
What does `-c` mean here? I can not find docs for that option.

jlebon:
Ahh thanks, good catch!

> What does `-c` mean here? I can not find docs for that option.

See https://github.com/dracutdevs/dracut/blob/a0f12fb6a09b09f35ab28753d7c4461c10a8b562/dracut-init.sh#L1020.

travier (dismissed):
There's an extra empty line but LGTM after that

travier (approved):

PR #1086, opened 2021-07-06
15fcos: remember when Ignition ran and print on console

jlebon:
Some users sometimes may not realize that they're using a pre-booted
version of a CoreOS image. This makes things confusing because they then
don't understand why the Ignition config wasn't applied.

There's no way to consistently detect this, but at least we can print an
informational message about (1) when Ignition ran, and (2) how many
boots ago that was.

This enhances the Ignition issue we already write for whether a
user config was provided rather than creating a separate one.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1977949
Related: https://github.com/coreos/ignition/issues/1214

PR #1089, opened 2021-07-07
Move coreos-ignition-firstboot-complete logic to separate script and make it delete /boot/ignition

jlebon:


jlebon:
(Still testing this.)

jlebon:
Ahhh maybe I shouldn't have included the two prep commits here (they're from https://github.com/coreos/fedora-coreos-config/pull/1086). I can drop them, though OTOH they're trivial.

jlebon:
Tested working fine interactively:

```
$ cosa run -p qemu-iso -c --add-disk 10G
...
[root@cosa-devsh ~]# coreos-installer install /dev/vda --ignition /run/ignition.json
...
[root@cosa-devsh ~]# reboot
...
<add rd.break on GRUB menu>
sh-5.1# mount /dev/disk/by-label/boot /mnt
sh-5.1# ls -la /mnt/ignition
total 7
drwxr-xr-x. 2 root root 1024 Jul  7 20:30 .
drwxr-xr-x. 8 root root 1024 Jul  7 20:30 ..
-rw-r--r--. 1 root root 1239 Jul  7 20:30 config.ign
sh-5.1# umount /mnt
sh-5.1# exit
...
[root@cosa-devsh ~]# systemctl status coreos-ignition-firstboot-complete.service
● coreos-ignition-firstboot-complete.service - CoreOS Mark Ignition Boot Complete
     Loaded: loaded (/usr/lib/systemd/system/coreos-ignition-firstboot-complete.service; enabled; vendor preset: enabled)
     Active: active (exited) since Wed 2021-07-07 20:33:08 UTC; 3min 14s ago
       Docs: https://docs.fedoraproject.org/en-US/fedora-coreos/
    Process: 971 ExecStart=/usr/libexec/coreos-ignition-firstboot-complete (code=exited, status=0/SUCCESS)
   Main PID: 971 (code=exited, status=0/SUCCESS)
        CPU: 33ms

Jul 07 20:33:08 localhost systemd[1]: Starting CoreOS Mark Ignition Boot Complete...
Jul 07 20:33:08 localhost systemd[1]: Finished CoreOS Mark Ignition Boot Complete.

[root@cosa-devsh ~]# ls -la /boot
total 27
drwxr-xr-x.  7 root root  1024 Jul  7 20:33 .
drwxr-xr-x. 12 root root   253 Jul  7 19:05 ..
lrwxrwxrwx.  1 root root     1 Jul  7 19:05 boot -> .
-rw-r--r--.  1 root root  1472 Jul  7 19:05 bootupd-state.json
drwxr-xr-x.  2 root root  1024 Jul  7 19:04 efi
drwx------.  5 root root  1024 Jul  7 19:05 grub2
lrwxrwxrwx.  1 root root     8 Jul  7 19:05 loader -> loader.1
drwxr-xr-x.  3 root root  1024 Jul  7 19:05 loader.1
drwx------.  2 root root 12288 Jul  7 19:04 lost+found
drwxr-xr-x.  3 root root  1024 Jul  7 19:05 ostree
```

jlebon:
https://github.com/coreos/coreos-assembler/pull/2283 adds a check for this in `testiso`.

bgilbert (dismissed):
Do we want to delete `/run/ignition.json` as well?

bgilbert:
On second thought, the deletion should be handled upstream, so we don't accidentally leak the files on other OSes; `/boot/ignition/config.ign` is defined by `ignition-setup-user.service`.  Removal of `/run/ignition.json`, if we decide to do that, should happen there too.

bgilbert:
Upstream alternative in https://github.com/coreos/ignition/pull/1246.

That PR doesn't remove `/run/ignition.json`.  The sequencing is a bit awkward because the removal must happen after the `umount` stage, and that only runs during the switch-root.  It didn't seem important enough to pursue further, since the file is mode 0600 and only exists for one boot.

jlebon:
Some debating in https://github.com/coreos/ignition/pull/1246. But actually, we don't have to rush this part of the fix. Maybe let's focus on backporting https://github.com/coreos/coreos-installer/pull/571 and getting https://github.com/coreos/fedora-coreos-config/pull/1093 in for the next releases? New nodes will have it mode 600 on first boot, then deleted on the following boot.

bgilbert:
We can't 100% rely on https://github.com/coreos/coreos-installer/pull/571, since it's valid to use an older coreos-installer to install FCOS.  A packaging backport doesn't completely help either, since it doesn't affect the Quay container, so we'd need a new upstream release.  So I think it's simpler to think of this change the primary one and the `coreos-installer` one as the fallback mitigation.

bgilbert (approved):
https://github.com/coreos/ignition/pull/1246#issuecomment-876451763 convinced me that this is the right approach after all.  LGTM!

PR #1091, opened 2021-07-07
Buildroot: add golang

darkmuggle:
Adding golang means that we can start using a common build root for
Prow, Jenkins and as the seed image for COSA.


dustymabe (approved):
LGTM

PR #1093, opened 2021-07-08
overlay: remove `/boot/ignition` on upgrade if present

bgilbert:
On subsequent boots, if `/boot/ignition` is present, remove it.  This fixes up old nodes with a world-readable Ignition config in `/boot`.

For https://github.com/coreos/fedora-coreos-tracker/issues/889.

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-cleanup-ignition-config.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-cleanup-ignition-config.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=Clean Up Injected Ignition Config in /boot
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/889
+# Newer Ignition will handle this on first boot; we only want to clean up
+# leftover configs on upgrade.  Disambiguate those two code paths for tests.
+ConditionKernelCommandLine=!ignition.firstboot
+RequiresMountsFor=/boot
+ConditionPathExists=/boot/ignition
+# We ship a kdump.service dropin that remounts /boot rw; avoid conflicts
+Before=kdump.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-cleanup-ignition-config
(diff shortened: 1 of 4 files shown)


overlay.d/15fcos/usr/lib/systemd/system/coreos-cleanup-ignition-config.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Clean up injected Ignition config in /boot
```
jlebon: Minor/optional: convention is Title Case

jlebon (dismissed):
Nice and clean!

travier:
Does this work if we have another `/boot` remount happening at the same time? I'm not 100% sure about the semantics for slave mount namespace remounts.

bgilbert:
@travier I sequenced around all the parallel `/boot` remounts I could find.  Do you know of additional ones?

bgilbert:
Fixed unit description; added external test.

jlebon (approved):

dustymabe:
the `ext.config.migration.installer-cleanup` test is failing on our cloud providers

dustymabe:
> the `ext.config.migration.installer-cleanup` test is failing on our cloud providers

does the `AUTOPKGTEST_REBOOT_MARK` stuff even work on the cloud providers?

bgilbert:
Seems like it might not.  I didn't immediately find a cause, so for now, https://github.com/coreos/fedora-coreos-config/pull/1099 limits the test to QEMU.

travier:
> @travier I sequenced around all the parallel `/boot` remounts I could find. Do you know of additional ones?

Well, a user could an `rpm-ostree install` as part of a unit on first boot that would do a `/boot` remount but I'm sure if that would happen at the same time.

bgilbert:
> Well, a user could an `rpm-ostree install` as part of a unit on first boot that would do a `/boot` remount

That use case should be safe, since the unit in this PR doesn't run on the first boot.

PR #1095, opened 2021-07-08
35coreos-ignition: move `ignition-setup-user.service` here

bgilbert:
The service contains distro-specific assumptions, so move it out of the upstream Ignition repo.

https://github.com/coreos/ignition/pull/1248 removes the service upstream.  For more context, see https://github.com/coreos/ignition/pull/1246#issuecomment-876451763.


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-setup-user.sh
```diff
@@ -0,0 +1,32 @@
...
+    fi
+}
+
+destination=/usr/lib/ignition
+mkdir -p $destination
+
+if is-live-image; then
+    # Live image. If the user has supplied a config.ign via an appended
+    # initrd, put it in the right place.
+    copy_file_if_exists "/config.ign" "${destination}/user.ign"
```
jlebon, line 20: Hmm, I guess this bit could live in the live generator now. And then the service would be pulled in by `diskful.target` directly and e.g. we could hardcode the various `Requires` that are currently dynamic. OTOH, it's nice to have everything related to the Ignition config centralized, so... meh, maybe not worth it.

jlebon (approved):
Nice!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-setup-user.sh
bgilbert: Yeah, I'm on the fence too.  Leaving alone for now.

PR #1096, opened 2021-07-09
overrides: fast-track Ignition 2.11.0-2.fc34

bgilbert:
Move `ignition-firstboot-complete` and `ignition-setup-user` services to fedora-coreos-config.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -24,3 +24,7 @@ packages:
       type: fast-track
       reason: https://github.com/coreos/fedora-coreos-config/pull/1088
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2021-9d4c3ddc8a
+  ignition:
+    evr: 2.11.0-2.fc34
+    metadata:
+      type: fast-track


jlebon (approved):

manifest-lock.overrides.yaml
```diff
@@ -24,3 +24,7 @@ packages:
       type: fast-track
       reason: https://github.com/coreos/fedora-coreos-config/pull/1088
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2021-9d4c3ddc8a
+  ignition:
+    evr: 2.11.0-2.fc34
+    metadata:
+      type: fast-track
```
dustymabe, line 30: do we not need the bodhi URL?

bgilbert: I've pushed fast-tracks without the URL before, following the instructions in the comment block.  Is there a general preference here?

jlebon: The bot doesn't need it. It's purely for humans. I personally put it so that it's easy for reviewers to tell that it exists and we won't carry it forever. In this instance, I just manually checked myself.

bgilbert: Okay, cool.  If we generally want it, should we update the header comment?

dustymabe: yeah that would be nice. I do find the bodhi URL really useful especially when tracking the status manually. i.e. "we've been waiting more than a week for this thing to filter into the stable repos, what's the holdup?" then you check and see the update has negative feedback or something and we need to reconsider our stance.

bgilbert: https://github.com/coreos/fedora-coreos-config/pull/1098

PR #1097, opened 2021-07-09
[rawhide] Switch to ostree-format: "oci"

cgwalters:
Part of https://github.com/coreos/fedora-coreos-tracker/issues/812

In this initial step, we're merely switching the internal
tarball to be a different format.

A future step will change the FCOS pipeline to automatically
push this container to quay.io.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -2,3 +2,6 @@
 # similarly to manifest.yaml. Unlike image-base.yaml, which is shared by all
 # streams.
 include: image-base.yaml
+
+# https://github.com/coreos/coreos-assembler/pull/2216
+ostree-format: "oci"


kelvinfan001 (approved):
LGTM

dustymabe:
two questions:

- Does this have any implications for things that currently use it (I think our upgrade tests use it)?
- I agree that we shouldn't make this change to `testing-devel` (new releases early next week) but.. Any reason to only be doing this for rawhide?

jlebon:
> two questions:
> 
> * Does this have any implications for things that currently use it (I think our upgrade tests use it)?

Good catch! Indeed, we'll need to adapt the upgrade test for this.

> * I agree that we shouldn't make this change to `testing-devel` (new releases early next week) but.. Any reason to only be doing this for rawhide?

I think for this type of stuff it makes a lot of sense to start off on mechanical streams before propagating to development and then production streams. I don't see a reason why we wouldn't propagate once it's sanity-checked in rawhide. (I guess at some point we should just flip the cosa default?)

cgwalters:
OK this blocks on https://github.com/coreos/coreos-assembler/pull/2300

jlebon (approved):

dustymabe:
I see an issue in rawhide building - possibly related to this?

```
No apparent changes since previous commit; use --force-nocache to override
commit: cac48abcb84ddade31b949e77ba4fd7db158d5c2bba909ef0f6c4f2f89fc1fc1 image: ace11605cf595bdb025d8aeb6934f2f7d070c72397ca59513c6570a10f2964c8
Commit cac48abcb84ddade31b949e77ba4fd7db158d5c2bba909ef0f6c4f2f89fc1fc1 unchanged; reusing previous build's rpm-ostree metadata
ostree diff commit from: cac48abcb84ddade31b949e77ba4fd7db158d5c2bba909ef0f6c4f2f89fc1fc1
ostree diff commit to:   cac48abcb84ddade31b949e77ba4fd7db158d5c2bba909ef0f6c4f2f89fc1fc1
New image input checksum: ace11605cf595bdb025d8aeb6934f2f7d070c72397ca59513c6570a10f2964c8
None
New build ID: 35.20210722.91.0-1
error: Building oci: Writing ostree root to blob: Exporting commit: No such metadata object 50773817e4519629fb061cb3cfe4ddae0a996c12336d087042481fbeab1a380c.dirmeta
script returned exit code 1
```

https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/blue/organizations/jenkins/fedora-coreos%2Ffedora-coreos-fedora-coreos-pipeline/detail/fedora-coreos-fedora-coreos-pipeline/2246/pipeline

cgwalters:
I suspect this is just ye olde https://github.com/coreos/coreos-assembler/issues/972

PR #1099, opened 2021-07-10
tests/installer-cleanup: only run on QEMU

bgilbert:
coreos-installer doesn't usually run on clouds and there's nothing cloud-specific about the cleanup code.  Also, rebooting from external tests doesn't seem to work outside of QEMU right now.


dustymabe:
LGTM

bgilbert:
Merging on the basis of https://github.com/coreos/fedora-coreos-config/pull/1099#issuecomment-877520462.  Dusty wasn't logged into GitHub on his device at the time.

PR #1105, opened 2021-07-13
35coreos-live: add requirement on loop kmod

dustymabe:
For some reason in rawhide this is no longer getting pulled in to the
initramfs. Since we require the use of it here then let's be explicit
and pull it in.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/module-setup.sh
@@ -12,6 +12,11 @@ install_and_enable_unit() {
     systemctl -q --root="$initdir" add-requires "$target" "$unit" || exit 1
 }
 
+installkernel() {
+    # we do loopmounts
+    instmods -c loop
+}
+
 install() {
     inst_multiple \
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/module-setup.sh
```diff
@@ -12,6 +12,11 @@ install_and_enable_unit() {
     systemctl -q --root="$initdir" add-requires "$target" "$unit" || exit 1
 }
 
+installkernel() {
+    # we do loopmounts
+    instmods loop
```
jlebon:
```suggestion
    instmods -c loop
```

That way it's an explicit error if somehow dracut doesn't find it at compose time.

jlebon (dismissed):
One comment, LGTM otherwise.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/module-setup.sh
dustymabe: done!

bgilbert (approved):

PR #1106, opened 2021-07-13
[rawhide] lockfiles: add rawhide override lockfile, pin kernel

dustymabe:
Here we pin to kernel-5.14.0-0.rc0.20210701gitdbe69e433722.6.fc35.
It's the last good kernel before [1] and [2] so we need to go back
to it for now.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/897
[2] https://github.com/coreos/fedora-coreos-tracker/issues/898

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -1,0 +1,25 @@
+# This lockfile should be used to pin to a package version (`type: pin`) or to
+# fast-track packages ahead of Bodhi (`type: fast-track`). Fast-tracked
+# packages will automatically be removed once they are in the stable repos.
+#
+# IMPORTANT: YAML comments *will not* be preserved. All `pin` overrides *must*
+# include a URL in the `metadata.reason` key. Overrides of type `fast-track`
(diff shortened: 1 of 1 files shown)


dustymabe:
CI will need to wait on https://github.com/coreos/fedora-coreos-config/pull/1105 to propagate to the `rawhide` branch. I'll rebase/push when that happens.

bgilbert (approved):
LGTM

dustymabe:
this needs https://github.com/coreos/fedora-coreos-config/pull/1107 to land and migrate over too

dustymabe:
There is a downgrade detected so CI won't pass

```
[2021-07-14T02:50:38.905Z] Downgraded:
[2021-07-14T02:50:38.905Z]   runc 2:1.0.0-376.dev.git12644e6.fc35 -> 2:1.0.0-2.fc35
```

But I guess that's just what happened in RPMs in rawhide. So we should be able to ignore.

dustymabe:
#1107 has landed now and propagated over to `rawhide`. Pushing up a rebase now. Will skip CI since it's already run for this changeset and will "fail" anyway because of the package downgrade.

PR #1107, opened 2021-07-14
ci: don't do a strict build for mechanical streams

dustymabe:
If we are operating on a mechanical stream then we can pin packages in
lockfiles but we don't maintain a full set so we can't do a strict build.


jlebon (approved):
Makes sense to me!

PR #1108, opened 2021-07-14
manifests: move fedora-coreos-pool to fedora-coreos.yaml

jlebon:
This is common to all streams, so we can move it here. I had initially
kept it in `manifest.yaml` because of:

https://github.com/coreos/fedora-coreos-config/pull/355#issuecomment-620860762

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -16,12 +16,6 @@ repos:
   - fedora
   - fedora-updates
 
-# All Fedora CoreOS streams share the same pool for locked files.
-# This will be in fedora-coreos.yaml in the future so it can be more easily be
-# shared between all the streams
-lockfile-repos:
-  - fedora-coreos-pool
-
 add-commit-metadata:
   fedora-coreos.stream: testing-devel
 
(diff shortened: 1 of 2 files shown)


jlebon:
So after this, we should be able to drop `repos` from all the manifests for production streams.

dustymabe:
> So after this, we should be able to drop `repos` from all the manifests for production streams.

Is there a reason the production streams have `repos` in there now and not `lockfile-repos:`?

dustymabe (approved):
LGTM

jlebon:
> > So after this, we should be able to drop `repos` from all the manifests for production streams.
> 
> Is there a reason the production streams have `repos` in there now and not `lockfile-repos:`?

They just haven't been updated at all since adding support for `lockfile-repos` in rpm-ostree because it initially broke if only `lockfile-repos` was defined, and no `repos`.

jlebon:
Let's update the production branches after the current releases are clear.

jlebon:
> Let's update the production branches after the current releases are clear.

OK right, let's do this at promotion time instead so we don't update non-`manifest.yaml` files out of cycle. Added a note in https://github.com/coreos/fedora-coreos-streams/issues/339 and https://github.com/coreos/fedora-coreos-streams/issues/338.

PR #1109, opened 2021-07-14
ci: add `branched` to list of mechanical streams

jlebon:
We might as well enable overrides support on all mechanical streams
while we're at it.

Eventually, we should move our stream definitions into coreos-ci-lib to
have it in one canonical place.


dustymabe (approved):
LGTM

PR #1110, opened 2021-07-14
[rawhide] lockfiles: unpin kernel

dustymabe:
Revert of 5fe63f6. The issue has been resolved in newer kernels.


dustymabe:
> The issue has been resolved in newer kernels.

we must wait for this to be true. This PR is a reminder to check and merge when that is true.

dustymabe:
This now seems to be fixed with [kernel-5.14.0-0.rc2.23.fc35](https://bodhi.fedoraproject.org/updates/FEDORA-2021-4d98455de6) so we can unpin the kernel (merge this PR). Still might be a little delay before it hits rawhide repos (pungi run). I'll move this out of draft once that happens.

jlebon (approved):

PR #1111, opened 2021-07-14
overlay: chmod 600 /etc/sudoers.d/coreos-sudo-group

bgilbert:
Apparently there are security scanners that object to mode 644.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1981979.  Requires coreos/coreos-assembler#2293.

### DIFF
--- a/overlay.d/05core/statoverride
+++ b/overlay.d/05core/statoverride
@@ -1,0 +1,2 @@
+# Config file for overriding permission bits on overlay files/dirs
+# Format: =<file mode in decimal> <absolute path to a file or directory>
--- a/overlay.d/05core/statoverride
+++ b/overlay.d/05core/statoverride
@@ -1,2 +1,6 @@
 # Config file for overriding permission bits on overlay files/dirs
 # Format: =<file mode in decimal> <absolute path to a file or directory>
+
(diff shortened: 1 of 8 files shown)


overlay.d/05core/statoverride
```diff
@@ -0,0 +1,6 @@
+# Config file for overriding permission bits on overlay files/dirs
+# Format: =<file mode in decimal> <absolute path to a file or directory>
```
dustymabe, line 2: I suspect this is a limitation of the underlying technology, but... 

> `<file mode in decimal>`

I suppose `octal` isn't an option here? For some reason my human brain only comprehends octal when talking about file permissions.

bgilbert: Octal would be ideal, but [sadly no](https://github.com/ostreedev/ostree/blob/main/src/ostree/ot-builtin-commit.c#L162).

jlebon: https://github.com/ostreedev/ostree/issues/2400

jlebon (dismissed):
Let's add a test in `misc-ro` to verify that it works?

bgilbert:
Test added!

jlebon (approved):

PR #1112, opened 2021-07-14
manifests: make the systemd-resolved neutering F34 only

### LINKED ISSUE #834: fully enabling systemd-resolved
In the past we enabled systemd-resolved, but disabled the stub listener because of issues related to setting node hostnames using reverse DNS queries. 

context: https://github.com/coreos/fedora-coreos-tracker/issues/649

There continues to be issues with interactions between NetworkManager
and systemd-resolved that make setting the hostname via reverse DNS
unpredictable. 

current status: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/601#note_892280

We should track and work with the upstreams here to get the issues sorted out so we can fully enable `systemd-resolved` in Fedora CoreOS and match the rest of Fedora.

dustymabe:
The underlying issues are resolved in the latest NetworkManager in
Fedora 35+ so let's conditionalize it on Fedora 34.

Was fixed upstream by https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/877

Will fix https://github.com/coreos/fedora-coreos-tracker/issues/834

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -83,8 +83,6 @@ postprocess:
 
   # Neuter systemd-resolved for now.
   # https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-743219353
-  # Note: When removing this, we likely also want to remove
-  # coreos-reset-stub-resolv-selinux-context.{path,service} and their presets.
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -83,9 +83,16 @@ postprocess:
 
   # Neuter systemd-resolved for now.
   # https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-743219353
+  # Remove when on F35+ as NM now handles rdns + resolved better
+  # https://github.com/coreos/fedora-coreos-tracker/issues/834
+  # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/601
+  # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/877
   - |
     #!/usr/bin/env bash
     set -xeuo pipefail
+    # Only operate on F34 since F35+ has been fixed
+    source /etc/os-release
+    [ ${VERSION_ID} -eq 34 ] || exit 0
     # Get us back to Fedora 32's nsswitch.conf settings
     sed -i 's/^hosts:.*/hosts:      files dns myhostname/' /etc/nsswitch.conf
     mkdir -p /usr/lib/systemd/resolved.conf.d/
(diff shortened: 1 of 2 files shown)


manifests/fedora-coreos-base.yaml
```diff
@@ -83,31 +83,42 @@ postprocess:
...
   # the above. See: https://bugzilla.redhat.com/show_bug.cgi?id=1892235#c25
+  # Remove when on F35+ as NM now handles rdns + resolved better
+  # https://github.com/coreos/fedora-coreos-tracker/issues/834
+  # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/601
+  # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/877
   - |
     #!/usr/bin/env bash
     source /etc/os-release
-    if [ -z "${DEFAULT_HOSTNAME:-}" ]; then
+    if [ ${VERSION_ID} -eq 34 ] && [ -z "${DEFAULT_HOSTNAME:-}" ]; then
```
dustymabe: Note.. This does mean that an upgrading system with no legitimate hostname (not from DHCP, not from rDNS, not statically set), will go from `localhost` to `fedora`. I'd say we *can* live with that.

If we want to go to bat and try to convince the rest of fedora to switch back to `localhost` OR to only apply `fedora` to the workstation variant, I think it's going to take some extra work (if anyone wants to take up that endeavor).

dustymabe: IOW the choices here are:

- keep this delta forever only in FCOS (not ideal)
- convince a subset of Fedora editions to change (non workstation use cases)
- convince the entirety of Fedora editions to change (not likely)

manifests/fedora-coreos-base.yaml
```diff
@@ -83,31 +83,42 @@ postprocess:
...
-    sed -i 's/^hosts:.*/hosts:      files dns myhostname/' /etc/nsswitch.conf
-    mkdir -p /usr/lib/systemd/resolved.conf.d/
-    cat > /usr/lib/systemd/resolved.conf.d/fedora-coreos-stub-listener.conf <<'EOF'
-    # Fedora CoreOS is electing to not use systemd-resolved's internal
-    # logic for now because of issues with setting hostnames via reverse DNS.
-    # https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-736104003
-    [Resolve]
-    DNSStubListener=no
+    # Only operate on F34 since F35+ has been fixed
+    #
```
dustymabe:
```suggestion
```

will remove in next upload along with suggestions for other changes.

manifests/fedora-coreos-base.yaml
```diff
@@ -83,31 +83,42 @@ postprocess:
...
+    if [ ${VERSION_ID} -eq 34 ]; then
+      # Get us back to Fedora 32's nsswitch.conf settings
+      sed -i 's/^hosts:.*/hosts:      files dns myhostname/' /etc/nsswitch.conf
+      mkdir -p /usr/lib/systemd/resolved.conf.d/
+      cat > /usr/lib/systemd/resolved.conf.d/fedora-coreos-stub-listener.conf <<'EOF'
+      # Fedora CoreOS is electing to not use systemd-resolved's internal
+      # logic for now because of issues with setting hostnames via reverse DNS.
+      # https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-736104003
+      [Resolve]
+      DNSStubListener=no
```
jlebon: Shouldn't these lines be dedented two spaces?

jlebon: Another approach for a more minimal diff is to just exit early e.g.

```
source /etc/os-release
[ ${VERSION_ID} -eq 34 ] || exit
```

manifests/fedora-coreos-base.yaml
jlebon: Agreed carrying this forever isn't great. I think we should aim for the second bullet.

jlebon (dismissed):
One minor comment, but LGTM overall!
We're slowly getting to the end of this saga. :)

manifests/fedora-coreos-base.yaml
dustymabe: implemented

manifests/fedora-coreos-base.yaml
dustymabe: opened https://github.com/coreos/fedora-coreos-tracker/issues/902 and updated the comment here

manifests/fedora-coreos-base.yaml
jlebon: Hmm, not seeing it in the latest push. Forgot a `git add`?

dustymabe: Sorry.. Now really fixed!

jlebon (approved):

PR #1116, opened 2021-07-15
tests: add tests for systemd reading files labeled kubernetes_file_t

miabbott:
Adds a test for the ability of `systemd` to read files
labeled with `kubernetes_file_t`.
    
See: https://bugzilla.redhat.com/show_bug.cgi?id=1973418

### DIFF
--- a/tests/kola/kubernetes/systemd-env-read/test.sh
+++ b/tests/kola/kubernetes/systemd-env-read/test.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# In order to verify that `kubernetes_file_t` labeled files can be read by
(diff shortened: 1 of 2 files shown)


miabbott:
This builds on #1064 and will require a new `container-selinux` that has a fix for BZ#1973418

miabbott:
@dustymabe @mike-nguyen tagging you to pop this closer to the top of your queues

tests/kola/kubernetes/systemd-env-read/test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
```
dustymabe, line 13: I would add an additional check here to make sure the file at `/etc/kubernetes/envfile` has the selinux label we expect. Otherwise this test isn't really testing what we expect it to.

dustymabe (dismissed):
one recommendation, otherwise LGTM

dustymabe (approved):
LGTM

PR #1121, opened 2021-07-19
[rhcos-4.7] 35coreos-live: stop overriding NetworkManager-wait-online timeout to 5s

jlebon:
We originally did this in #326 because we wanted to support booting the
live ISO without networking. This was solved on the initramfs side by
the conditional networking work (#426). But for the real root, this was
still useful because if booting the ISO interactively on a system
without any network, or a non-DHCP network, we didn't want the user to
have to wait until the service timed out before getting a shell.

The core issue however is that we're requesting `network-online.target`
at all. It's an "active unit" which means that it's only pulled in the
transaction, possibly delaying boot, if another systemd unit needs it.
And ideally, no service would need it as per:

https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

In our case, this unit was fedora-coreos-pinger. We drop that
requirement here:

https://github.com/coreos/fedora-coreos-pinger/pull/41

With that, we no longer pull in `network-online.target` and so no longer
delay reaching the console even if NetworkManager isn't able to get an
active connection for whatever reason. This matches how it works on
traditional Fedora as well.

Having a short timeout actually also had a counterproductive effect in
the automated install case. There, `coreos-installer.service` does pull
in `network-online.target` (which with
https://github.com/coreos/coreos-installer/pull/565 we could consider
dropping as advised by systemd, though we probably should bump the
number of retries some more in that case), but because of the short
timeout, we genuinely may not yet have the network fully up before we
run (see https://bugzilla.redhat.com/show_bug.cgi?id=1967483).

(cherry picked from commit dd54e8cb1aa93420cbdd714dcd2376962c5b86a3)


jlebon:
Backport of https://github.com/coreos/fedora-coreos-config/pull/1088 for https://bugzilla.redhat.com/show_bug.cgi?id=1967483.

dustymabe (approved):
LGTM

PR #1127, opened 2021-07-21
[stable] lockfiles: fast-track kernel and systemd

bgilbert:
Fixes CVE-2021-33909 and CVE-2021-33910.

See https://github.com/coreos/fedora-coreos-tracker/issues/904.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,6 +1,6 @@
 // Documentation: https://github.com/coreos/coreos-ci/blob/main/README-upstream-ci.md
 
-cosaPod {
+cosaPod(image: 'quay.io/coreos-assembler/coreos-assembler:fcos-34.20210626.3.1') {
     checkoutToDir(scm, 'config')
 
     def basearch = shwrapCapture("cosa basearch")
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -37,3 +37,39 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/904
       type: fast-track
+  systemd:
+    evr: 248.5-1.fc34
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2021-2a6ba64260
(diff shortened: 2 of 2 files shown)


dustymabe (approved):
LGTM

dustymabe (approved):
still LGTM

jlebon:
CI failing on the classic cosa <--> FCOS binding problem (see https://github.com/coreos/fedora-coreos-streams/issues/64), in this case because of https://github.com/coreos/coreos-assembler/pull/2283 failing since stable doesn't yet have https://github.com/coreos/fedora-coreos-config/pull/1089. It'll fail in the production build too.

In hindsight, probably should've made that service only trigger if e.g. `coreos-ignition-firstboot-complete.service` exists or something. Which is really not hard to do even now, though... given that this is we already have more changes than we'd like in this `stable` respin, maybe it's better to just do the cosa branch dance instead.

/time passes

OK, pushed https://github.com/coreos/coreos-assembler/tree/fcos-34.20210626.3.1 at the same commit as the last `stable` FCOS build. Currently building in Quay: https://quay.io/repository/coreos-assembler/coreos-assembler?tab=builds. We'll need to make sure to use that when building.

For CI here, you should be able to override the `image` in the `pod()` call in `.cci.jenkinsfile` once the image is built so it uses that cosa image instead. The next regular promotion should revert it.

bgilbert:
@jlebon Thanks for tracking that down.  We would have gotten away with it, too, if not for the stable respin.

.cci.jenkinsfile
```diff
@@ -1,6 +1,6 @@
 // Documentation: https://github.com/coreos/coreos-ci/blob/main/README-upstream-ci.md
 
-cosaPod {
+cosaPod(image: 'fcos-34.20210626.3.1') {
```
jlebon: This needs to be an FQIN, i.e. `quay.io/coreos-assembler/coreos-assembler:fcos-34.20210626.3.1`.

bgilbert: Whoops, thanks.

PR #1128, opened 2021-07-21
overlay: use Ignition `result.json`

bgilbert:
As of https://github.com/coreos/ignition/pull/1250, Ignition produces its own report similar to `ignition.info.json`, so use that instead of making our own.

Hold until https://github.com/coreos/ignition/pull/1250 lands in a release.

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
@@ -1,25 +1,17 @@
 #!/usr/bin/bash
 set -euo pipefail
 
-IGNITION_INFO=/var/lib/coreos/ignition.info.json
-IGNITION_FIRSTBOOT_INFO=/run/ignition.info.json
+IGNITION_RESULT=/var/lib/ignition/result.json
 
 WARN='\033[0;33m' # yellow
 RESET='\033[0m' # reset
 
-# See coreos-ignition-firstboot-complete
-is_firstboot=0
-if [ -e "${IGNITION_FIRSTBOOT_INFO}" ]; then
-    is_firstboot=1
-    mkdir -p "$(dirname "${IGNITION_INFO}")"
-    mv "${IGNITION_FIRSTBOOT_INFO}" "${IGNITION_INFO}"
-fi
-
 mkdir -p /run/issue.d
 touch /run/issue.d/30_coreos_ignition_provisioning.issue
 
-d=$(date --date "@$(jq .date "${IGNITION_INFO}")" +"%Y/%m/%d %H:%M:%S %Z")
-if [ "${is_firstboot}" == 1 ]; then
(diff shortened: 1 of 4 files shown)


overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -1,25 +1,16 @@
...
-    mv "${IGNITION_FIRSTBOOT_INFO}" "${IGNITION_INFO}"
-fi
-
 mkdir -p /run/issue.d
 touch /run/issue.d/30_coreos_ignition_provisioning.issue
 
-d=$(date --date "@$(jq .date "${IGNITION_INFO}")" +"%Y/%m/%d %H:%M:%S %Z")
-if [ "${is_firstboot}" == 1 ]; then
+d=$(date --date "$(jq -r .provisioningDate "${IGNITION_RESULT}")" +"%Y/%m/%d %H:%M:%S %Z")
+if [ -f /run/ignition.json ]; then
```
jlebon: Hmm, should we use something else here in case we decide to delete this file in the future? Maybe let's just check `ignition.firstboot` in `/proc/cmdline` since that's the source of truth. It's not great, though `live-generator` in the real root does: https://github.com/coreos/fedora-coreos-config/blob/a75418838e2357757556ac38792ab7827339d172/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator#L9.

There's also https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/coreos/generator-lib.sh we could enhance though it's geared towards generators.

jlebon: I guess another approach is to add the boot ID to `/var/lib/ignition/result.json` and here we just compare against the current boot ID. Having the boot ID in there could be useful on its own as a way to cross-check with journal entries and anything else which records boot IDs.

jlebon (dismissed):
Nice, diff looks great!

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
cgwalters: Adding the boot ID makes sense.  Or we can check `journalctl -b -t ignition` right?

cgwalters (commented):
LGTM

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
bgilbert: I like the boot ID approach.  Do you think we should store it in UUID format with dashes (as the kernel presents it) or without dashes (as `journalctl -b` wants it)?

jlebon: I'd vote for the former since that's the canonical representation.

bgilbert:
Updated!

jlebon (approved):

PR #1129, opened 2021-07-21
lockfiles: update bodhi link for kernel-5.13.4-200.fc34

dustymabe:
The bodhi update had not been created yet when we fast-tracked it
originally.


dustymabe:
i'm going to skip CI on this one. Save some coal.

jlebon:
That bodhi update is for f33.

dustymabe:
> That bodhi update is for f33.

That was a test and... you passed!

No really, though. Thanks for catching that. Fixed now.

jlebon (approved):

PR #1130, opened 2021-07-22
lockfiles: drop graduated overrides 🎓

coreosbot-releng:
Triggered by remove-graduated-overrides GitHub Action.


dustymabe (approved):
LGTM

jlebon:
That doesn't look right. It looks like only some of the systemd subpackages graduated, which doesn't make sense.

jlebon:
https://github.com/coreos/fedora-coreos-config/pull/1131

PR #1131, opened 2021-07-22
ci/remove-graduated-overrides: query by arch too

jlebon:
The x86_64 repo also includes i686 packages so we need to filter on the
arch too when querying to make sure we're comparing packages of the same
arch.

### DIFF
--- a/ci/remove-graduated-overrides.py
+++ b/ci/remove-graduated-overrides.py
@@ -109,7 +109,8 @@ def update_lockfile(base, fn):
 
 def sack_has_nevra_greater_or_equal(base, nevra):
     nevra = hawkey.split_nevra(nevra)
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

dustymabe:
For future me

```
10:16:00     jlebon | the query returned both arches. depending which one was listed first, the
                    | comparison to know if they graduated was comparing cross-arch
10:16:26  dustymabe | got ya. and it's not consistent, meaning half of the systemd packages got
                    | cleaned and half didn't ?
10:16:34     jlebon | yeah, exactly
10:17:24     jlebon | hmm i guess there's a corner-case there with binary rpms which go from arch to
                    | noarch or the other way round

```

PR #1134, opened 2021-07-23
tests/misc-ro: Simplify permission checks for files in /etc

travier:
Make use of some GNU find extensions to make the check simpler.

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -174,16 +174,16 @@ if [[ $sys_fs_cgroup_source != cgroup2 ]]; then
     fatal "/sys/fs/cgroup is not cgroup2"
 fi
 
-for perms in 'o+w' 'g+w'; do
-	list="$(find /etc -type f -perm /${perms})"
-	if [[ -n "${list}" ]]; then
-		fatal "found files with ${perms}:\n${list}"
-	fi
-done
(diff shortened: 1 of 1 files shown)


tests/kola/misc-ro
```diff
@@ -174,13 +174,11 @@ if [[ $sys_fs_cgroup_source != cgroup2 ]]; then
...
-for perms in 'o+w' 'g+w'; do
-	list="$(find /etc -type f -perm /${perms})"
-	if [[ -n "${list}" ]]; then
-		fatal "found files with ${perms}:\n${list}"
-	fi
-done
-ok "no files with o+w or g+w found in /etc"
+list="$(find /etc -type f,d -perm /022)"
+if [[ -n "${list}" ]]; then
+	fatal "found files or directories with 'g+w' or 'o+w' permission:\n${list}"
```
jlebon: While we're here, instead of printing `${list}`, should we instead do e.g.

```suggestion
    ls -la ${list}
	fatal "found files or directories with 'g+w' or 'o+w' permission"
```

?

That way we also clearly see the mode bits that were actually set on each file.

jlebon (dismissed):
One comment, but LGTM as is too.

travier:
Updated. Thanks!

jlebon (approved):

PR #1135, opened 2021-07-23
coreos-kargs-reboot: make flag file a triggering condition

bgilbert:
RHCOS FIPS support wants to add an additional trigger.  See https://github.com/openshift/os/pull/574#discussion_r675688876.


bgilbert:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/1136.

PR #1138, opened 2021-07-26
lockfiles: initial aarch64 lockfile

dustymabe:
We'll pre-populate it here and then bump-lockfile will handle
updating it in the future.


dustymabe:
will cancel/ignore CI for this one (save coal)

jlebon (approved):
Seems reasonable to me. It'd be nice if `bump-lockfile` handled this seamlessly, though meh... it's not like we'll be adding arches every other day. :)

PR #1142, opened 2021-07-27
15fcos: Order coreos-check-ssh-keys before systemd-user-sessions

travier:
Run before user sessions to avoid reloading agetty.

See also: https://github.com/coreos/fedora-coreos-config/commit/386b6fedfdee105e35b75381e48cc46c28806fad


travier:
~~(untested yet)~~ Looks like this works 🙂

jlebon (approved):

saqibali-2k (approved):
lgtm (my knowledge here is limited so please take this with a grain of salt)

PR #1144, opened 2021-07-28
denylist: snooze podman.network-single on openstack

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/901

The problem comes and goes in the vexxhost environment. Let's wait
a month and see if it's better then.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -5,3 +5,8 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: podman.workflow
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
+- pattern: podman.network-single
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/901
+  snooze: 2021-09-01
+  platforms:
+    - openstack


dustymabe:
requires https://github.com/coreos/coreos-assembler/pull/2324

kola-denylist.yaml
```diff
@@ -5,3 +5,8 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: podman.workflow
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
+- pattern: podman.network-single
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/901
+  snooze: 2021-09-01
+  platforms:
+    - openstack
```
jlebon, line 12: I don't think `platforms` is supported, though it shouldn't be hard to add.

jlebon (commented):
Should we get in touch with our contacts at VexxHost to check if they know anything?

kola-denylist.yaml
jlebon: Sorry, totally missed your link to cosa. :)

dustymabe:
> Should we get in touch with our contacts at VexxHost to check if they know anything?

I'd love to dig in and understand the particular failure but since it's only failing on one platform and 98 or 99 of 100 of the iterations of this particular test pass, I'm inclined to say we've got more important things to work on right now.

jlebon (approved):

jlebon:
> > Should we get in touch with our contacts at VexxHost to check if they know anything?
> 
> I'd love to dig in and understand the particular failure but since it's only failing on one platform and 98 or 99 of 100 of the iterations of this particular test pass, I'm inclined to say we've got more important things to work on right now.

Not saying we need to spend a lot of time trying to fix this. Snoozing is fine. But as cloud maintainers, they might be interested to know about this.

PR #533, opened 2020-07-25
[next-devel] overrides: fast-track coreos-installer, afterburn

dustymabe:
- coreos-installer-0.4.0-1.fc32.x86_64
- rust-afterburn-4.4.2-1.fc32.x86_64

PR #1146, opened 2021-07-29
overlay.d: add zipl

nikita-dubrovskii:
We require zipl in ramdisk (only on s390x) for ignition/rdcore kargs modification feature.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -24,6 +24,12 @@ install() {
         sed \
         sgdisk
 
+    # ignition can update kargs
+    if [[ "$_arch" == "s390x" ]]; then
+        inst_multiple zipl
+        inst /lib/s390-tools/stage3.bin
+    fi
+
     inst_simple "$moddir/coreos-diskful-generator" \
         "$systemdutildir/system-generators/coreos-diskful-generator"
 


jlebon:
Hmm thinking more on this, maybe this belongs better in https://github.com/coreos/coreos-installer/blob/main/dracut/50rdcore/module-setup.sh instead since nothing in this repo is actually calling `zipl` directly from the initramfs.

With https://github.com/coreos/coreos-installer/pull/533, it will actually be the first time that happens in FCOS, via `rdcore kargs` called by the Ignition kargs stage. So... WDYT about me pushing that patch in https://github.com/coreos/coreos-installer/pull/533 instead?

nikita-dubrovskii:
> Hmm thinking more on this, maybe this belongs better in https://github.com/coreos/coreos-installer/blob/main/dracut/50rdcore/module-setup.sh instead since nothing in this repo is actually calling `zipl` directly from the initramfs.
> 
> With [coreos/coreos-installer#533](https://github.com/coreos/coreos-installer/pull/533), it will actually be the first time that happens in FCOS, via `rdcore kargs` called by the Ignition kargs stage. So... WDYT about me pushing that patch in [coreos/coreos-installer#533](https://github.com/coreos/coreos-installer/pull/533) instead?

I had the same concerns, this patch i had locally for some time for experiments with ignition.

jlebon:
Moved to [`4c43a68` (#533)](https://github.com/coreos/coreos-installer/pull/533/commits/4c43a6845df910f817af0e4806a5c1f859cc34d9)!

PR #1148, opened 2021-08-02
15fcos: print warning If Ignition is run more than once

sohankunkerkar:
This change adds a warning on the serial console if Ignition is run more than once. This is related to https://github.com/coreos/ignition/issues/1214

This needs https://github.com/coreos/fedora-coreos-config/pull/1128 and https://github.com/coreos/ignition/pull/1254

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
@@ -14,6 +14,19 @@ ignitionBoot=$(jq -r .provisioningBootID "${IGNITION_RESULT}")
 if [ $(cat /proc/sys/kernel/random/boot_id) = "${ignitionBoot}" ]; then
     echo "Ignition: ran on ${d} (this boot)" \
         > /run/issue.d/30_coreos_ignition_provisioning.issue
+
+    # checking for /run/ostree-live as the live system with persistent storage can run Ignition more than once
+    if ! test -f /run/ostree-live && jq -e .previousReport.provisioningDate "${IGNITION_RESULT}" &>/dev/null; then
+        prevdate=$(date --date "$(jq -r .previousReport.provisioningDate "${IGNITION_RESULT}")" +"%Y/%m/%d %H:%M:%S %Z")
+        cat << EOF > /etc/issue.d/30_coreos_ignition_run_more_than_once.issue
+${WARN}
+############################################################################
+WARNING: Ignition previously ran on ${prevdate}. Unexpected 
+behavior may occur. Ignition is not designed to run more than once per system.
+############################################################################
+${RESET}
+EOF
+    fi
 else
     nreboots=$(($(journalctl --list-boots | wc -l) - 1))
     [ "${nreboots}" -eq 1 ] && boot="boot" || boot="boots"


overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -29,7 +29,13 @@ else
         > /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
-if jq -e .usercfg "${IGNITION_INFO}" &>/dev/null; then
+if jq -e .previousReport.provisioningDate "${IGNITION_RESULT}" &>/dev/null; then
```
bgilbert: `$IGNITION_RESULT` isn't defined in this diff.  Please rebase this change on top of #1128.

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -21,6 +21,12 @@ else
         > /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+if jq -e .previousReport.provisioningDate "${IGNITION_RESULT}" &>/dev/null; then
+    prevdate=$(date --date "$(jq -r .previousReport.provisioningDate "${IGNITION_RESULT}")" +"%Y/%m/%d %H:%M:%S %Z")
+    echo -e "${WARN}Ignition was previously run at ${prevdate}.\nUnexpected behavior may occur. Ignition is not designed to run more than once per system.${RESET}" \
+        >> /etc/issue.d/30_coreos_ignition_run_more_than_once.issue
```
bgilbert: Won't this add another copy of the warning on every boot?

sohankunkerkar: Fixed

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -21,6 +21,13 @@ else
         > /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+FILE=/run/ostree-live
```
bgilbert: Doesn't really need a variable.  (And if it had one, `FILE` is too generic of a name.)

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -21,6 +21,13 @@ else
...
+if ! test -f "${FILE}" && jq -e .previousReport.provisioningDate "${IGNITION_RESULT}" &>/dev/null; then
+    prevdate=$(date --date "$(jq -r .previousReport.provisioningDate "${IGNITION_RESULT}")" +"%Y/%m/%d %H:%M:%S %Z")
+    echo -e "${WARN}Ignition was previously run at ${prevdate}.\nUnexpected behavior may occur. Ignition is not designed to run more than once per system.${RESET}" \
+        > /etc/issue.d/30_coreos_ignition_run_more_than_once.issue
```
bgilbert: Any reason we're rewriting a file in `/etc` on every boot?  I'd think we'd just write to `/run` as the other checks do.

jlebon: Hmm, in this case the information isn't dynamic so I think having it in `/etc` makes sense. But yes, we shouldn't keep emitting it on every boot. Let's instead only run this check *once* on the first boot? (We already have logic conditional on first boot in this script.)

That also allows users to just delete the issue file if they'd like to get rid of the warning without having to resort to disabling the systemd unit (which prints other useful dynamic things).

bgilbert:
https://github.com/coreos/fedora-coreos-config/pull/1128 landed.

sohankunkerkar:
Fixes https://github.com/coreos/ignition/issues/1214

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -21,6 +21,13 @@ else
         > /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+# checking for /run/ostree-live as the live system with persistent storage can run Ignition more than once
+if ! test -f /run/ostree-live && jq -e .previousReport.provisioningDate "${IGNITION_RESULT}" &>/dev/null; then
```
jlebon: As discussed previously, let's only do this on the first boot?

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -14,6 +14,12 @@ ignitionBoot=$(jq -r .provisioningBootID "${IGNITION_RESULT}")
 if [ $(cat /proc/sys/kernel/random/boot_id) = "${ignitionBoot}" ]; then
     echo "Ignition: ran on ${d} (this boot)" \
         > /run/issue.d/30_coreos_ignition_provisioning.issue
```
jlebon, line 16: Minor/optional: space here?

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -14,6 +14,12 @@ ignitionBoot=$(jq -r .provisioningBootID "${IGNITION_RESULT}")
...
+    # checking for /run/ostree-live as the live system with persistent storage can run Ignition more than once
+    if ! test -f /run/ostree-live && jq -e .previousReport.provisioningDate "${IGNITION_RESULT}" &>/dev/null; then
+        prevdate=$(date --date "$(jq -r .previousReport.provisioningDate "${IGNITION_RESULT}")" +"%Y/%m/%d %H:%M:%S %Z")
+        echo -e "${WARN}Ignition was previously run at ${prevdate}.\nUnexpected behavior may occur. Ignition is not designed to run more than once per system.${RESET}" \
```
jlebon:
```suggestion
        echo -e "${WARN}Ignition was previously run on ${prevdate}.\nUnexpected behavior may occur. Ignition is not designed to run more than once per system.${RESET}" \
```

to match the other issue.

jlebon: Bikeshed:

To match the style of other warning issues, should this be e.g.

```suggestion
cat << EOF > /etc/issue.d/30_coreos_ignition_run_more_than_once.issue
${WARN}
############################################################################
WARNING: Ignition previously ran on ${prevdate}. Unexpected behavior
may occur. Ignition is not designed to run more than once per system.
############################################################################
${RESET}
EOF
```

?

See `coreos-check-rootfs-size` and `coreos-check-cgroups`.

jlebon (dismissed):
Will let @bgilbert have a look as well.

overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
```diff
@@ -14,6 +14,19 @@ ignitionBoot=$(jq -r .provisioningBootID "${IGNITION_RESULT}")
...
     echo "Ignition: ran on ${d} (this boot)" \
         > /run/issue.d/30_coreos_ignition_provisioning.issue
+
+    # checking for /run/ostree-live as the live system with persistent storage can run Ignition more than once
+    if ! test -f /run/ostree-live && jq -e .previousReport.provisioningDate "${IGNITION_RESULT}" &>/dev/null; then
+        prevdate=$(date --date "$(jq -r .previousReport.provisioningDate "${IGNITION_RESULT}")" +"%Y/%m/%d %H:%M:%S %Z")
+        cat << EOF > /etc/issue.d/30_coreos_ignition_run_more_than_once.issue
+${WARN}
+############################################################################
+WARNING: Ignition previously ran on ${prevdate}. Unexpected behavior
```
bgilbert: `$prevdate` takes us right to column 80.  Let's move `behavior` to the next line.

bgilbert (dismissed):

bgilbert (approved):

PR #1149, opened 2021-08-03
manifests/fedora-coreos-base: stop disabling modular repos

jlebon:
The latest rpm-ostree release now has proper support for modules. It
will not layer modular packages unless explicitly enabled. So let's stop
disabling the repos so e.g. `rpm-ostree module install cri-o:1.20` alone
just works.


jlebon:
Hold until new rpm-ostree release with https://github.com/coreos/rpm-ostree/pull/2760.

jlebon:
Part of https://github.com/coreos/fedora-coreos-config/pull/1191.

PR #1150, opened 2021-08-05
[rawhide] lockfiles: pin shadow-utils-4.8-1.fc35 in rawhide

dustymabe:
Rootless podman is broken with it.

See https://github.com/coreos/fedora-coreos-tracker/issues/916

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -13,3 +13,8 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/913
       type: pin
+  shadow-utils:
+    evr: 2:4.8.1-20.fc35
+    metadata:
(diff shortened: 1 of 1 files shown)


manifest-lock.overrides.yaml
```diff
@@ -13,3 +13,8 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/913
       type: pin
+  shadow-utils:
+    evr: 2:4.8-1.fc35
```
jlebon: Shouldn't this be

```suggestion
    evr: 2:4.8.1-20.fc35
```

based on https://github.com/coreos/fedora-coreos-tracker/issues/916#issue-962167292?

dustymabe: fixed

jlebon (approved):

PR #1153, opened 2021-08-06
image-base: drop useless `sysroot-readonly: true`

jlebon:
It's not read by coreos-assembler, which has been turning on read-only
`/sysroot` unconditionally for a while now:

https://github.com/coreos/coreos-assembler/pull/736


dustymabe (approved):
LGTM - thanks for cleaning this up.

PR #1154, opened 2021-08-06
35coreos-ignition: allow setting kargs in live system if it's a no-op

### LINKED ISSUE #917: Ignition kargs helper should succeed on live systems if kargs are already correct
The Ignition `kernelArguments` helper currently fails on live systems, since we have no way to change the kargs there.  However, if the specified kernel arguments are already set (`shouldExist` args exist and `shouldNotExist` args don't), it should succeed.

bgilbert:
In a live system, we can't set kargs in an Ignition config because we don't control the bootloader settings.  If the config specifies kargs anyway, we currently fail with `/dev/disk/by-label/boot: not a block device`, which isn't very clear.

In the live case, notice if the `shouldExist` kargs exist and the `shouldNotExist` kargs do not, because then we don't need to do anything and can succeed anyway.  Otherwise, fail with a clearer error message.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/917.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
@@ -1,4 +1,40 @@
 #!/bin/bash
 set -euo pipefail
 
-/usr/bin/rdcore kargs --boot-device /dev/disk/by-label/boot --create-if-changed /run/coreos-kargs-reboot "$@"
+fail_live() {
+    echo "Need to $1, but cannot change kernel arguments in live system." >&2
+    exit 1
+}
+
+if is-live-image; then
+    # In the live case, we can't actually change anything.  Check the
+    # arguments and succeed anyway if there's nothing we need to do.
+
+    # add leading and trailing whitespace to allow for easy matching
+    kernelopts=" $(cat /proc/cmdline) "
+
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+        --should-exist)
+            arg="$2"
+            if [[ ! "${kernelopts}" =~ " ${arg} " ]]; then
+                fail_live "add kernel argument '${arg}'"
+            fi
+            shift 2
+            ;;
+        --should-not-exist)
+            arg="$2"
+            if [[ "${kernelopts}" =~ " ${arg} " ]]; then
+                fail_live "remove kernel argument '${arg}'"
+            fi
+            shift 2
+            ;;
+        *)
+            echo "Unknown option" >&2
+            exit 1
+            ;;
+        esac
+    done
+else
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
```diff
@@ -1,4 +1,40 @@
...
+    # arguments and succeed anyway if there's nothing we need to do.
+
+    # add leading and trailing whitespace to allow for easy matching
+    kernelopts=" $(cat /proc/cmdline) "
+
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+        --should-exist)
+            arg="$2"
+            if [[ ! "${kernelopts}" =~ " ${arg} " ]]; then
```
lump, line 20: For these tests to see if the argument is already there, it might be nice to handle the case where the argument might be at the beginning or end of the string or separated by other whitespace, like tabs. e.g.:

```
if [[ "${kernelopts}" =~ (^|[[:space:]])${arg}([[:space:]]|$) ]] ; then
```

bgilbert: The first/last argument case is handled by the leading/trailing whitespace in the value of `$kernelopts`.  The assumption that arguments are space-separated is common to a lot of our karg-processing code, including the [rdcore code](https://github.com/coreos/coreos-installer/blob/96d33974e35fe697c3f4cec2964c9906512dcc07/src/install.rs#L348-L363) that normally handles Ignition kargs and the [Ignition example kargs helper](https://github.com/coreos/ignition/blob/b0059e3e25d37078d937703634c3f8e607b2b30d/examples/ignition-kargs-helper) that this PR was adapted from.

lump: Okay, thanks!  Nice to know that was already thought about!

travier (approved):
Untested but LGTM and I like the change.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
bgilbert: I think we've generally missed the fact that kargs can be tab-separated, so it is a bug.  We should be consistent in how we process kargs in the various code paths, though.

dustymabe (approved):

jlebon (approved):
LGTM, though feels like we should just fold this into `rdcore kargs` so we keep the logic centralized. E.g. have it error out if changes are needed and no `--boot-device` or `--boot-mount` options are specified, or just make it an explicit e.g. `--no-boot-device` option.

bgilbert:
Yeah, I think that's cleaner; https://github.com/coreos/coreos-installer/pull/597 and https://github.com/coreos/fedora-coreos-config/pull/1164.

bgilbert:
Closing in favor of #1164.

PR #1158, opened 2021-08-10
tests/countme: Retry five times to avoid flakes

travier:
We don't want to fail on partial count me reporting in pratice but the
current test was designed to fail in that case to make sure that we
don't miss an error in CI. This behavior becomes an issue in case of
Fedora infra flakes so retry five times in case of partial reporting to
avoid a costly CI failure as count me reporting is cheap.

### DIFF
--- a/tests/kola/rpm-ostree-countme/test.sh
+++ b/tests/kola/rpm-ostree-countme/test.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -xeuo pipefail
 
-# No need to run an other platforms than QEMU.
+# No need to run on any other platform than QEMU.
 # kola: { "tags": "needs-internet", "platforms": "qemu-unpriv" }
 
 ok() {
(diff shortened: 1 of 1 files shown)


tests/kola/rpm-ostree-countme/test.sh
```diff
@@ -14,27 +14,39 @@ fatal() {
...
-# depending on the stream, we expect different numbers of countme-enabled repos
-if [[ "${output}" != "Successful requests: 1/1" ]] && \
-   [[ "${output}" != "Successful requests: 2/2" ]] && \
-   [[ "${output}" != "Successful requests: 3/3" ]]; then
-	fatal "rpm-ostree-countme service ouput does not match expected sucess output"
-fi
+	# Check rpm-ostree count me output
+	output="$(journalctl --output=json --boot --unit=rpm-ostree-countme.service --grep "Successful requests:" | jq --raw-output '.MESSAGE')"
+	# depending on the stream, we expect different numbers of countme-enabled repos
+	if [[ "${output}" != "Successful requests: 1/1" ]] && \
```
dustymabe, line 57: just wondering.. will `output="$(journalctl` above include all the output from the previous runs? is there any way to limit it to just the latest run of the service?

tests/kola/rpm-ostree-countme/test.sh
```diff
@@ -14,27 +14,39 @@ fatal() {
...
+	   [[ "${output}" != "Successful requests: 2/2" ]] && \
+	   [[ "${output}" != "Successful requests: 3/3" ]]; then
+		echo "rpm-ostree-countme service ouput does not match expected sucess output (try: $i)"
+		sleep 10
+		continue
```
dustymabe: In this case is it possible the service is still running (i.e. the `systemctl start` at the beginning of the loop does nothing?)

tests/kola/rpm-ostree-countme/test.sh
```diff
@@ -14,27 +14,39 @@ fatal() {
...
+	rm -f /var/lib/private/rpm-ostree-countme/countme
+	# Check that running the service manually is successful
+	systemctl start rpm-ostree-countme.service
+	if [[ $(systemctl show -p ActiveState rpm-ostree-countme.service) != "ActiveState=inactive" ]] && \
+	   [[ $(systemctl show -p SubState rpm-ostree-countme.service) != "SubState=dead" ]] && \
+	   [[ $(systemctl show -p Result rpm-ostree-countme.service) != "Result=success" ]] && \
+	   [[ $(systemctl show -p ExecMainStatus rpm-ostree-countme.service) != "ExecMainStatus=0" ]]; then
+		echo "rpm-ostree-countme exited with an error (try: $i)"
+		sleep 10
+		continue
```
darkmuggle, line 34:
```suggestion
		break
```

dustymabe, line 51: I think we want `continue` because we want to retry to gloss over infra flakes, right?

tests/kola/rpm-ostree-countme/test.sh
travier: Oh, indeed, this will not work :/

tests/kola/rpm-ostree-countme/test.sh
travier: This is a oneshot unit so I would say this is not possible

tests/kola/rpm-ostree-countme/test.sh
travier: Not sure I understand your question here.

tests/kola/rpm-ostree-countme/test.sh
jlebon: You can do something similar to what we do in rpm-ostree tests to only get the journal logs after a certain point in time:
- https://github.com/coreos/rpm-ostree/blob/e1861fef09c87000b0e3e87f628691a82aab68c5/tests/common/libvm.sh#L491
- https://github.com/coreos/rpm-ostree/blob/e1861fef09c87000b0e3e87f628691a82aab68c5/tests/common/libvm.sh#L519

cgwalters (commented):
The frequency with which we run this test will actually distort the statistics, right?

I wonder if this test should either:

1) Be disabled by default, and run explicitly at most once a week as a periodic
2) Inject an additional header to say it's the test that we can then discount on the server side
3) (likely hard, probably not worth it) use a staging server

travier:
> The frequency with which we run this test will actually distort the statistics, right?

Yes, this will artificially increase the first Count Me bucket count.

> I wonder if this test should either:
> 
> 1. Be disabled by default, and run explicitly at most once a week as a periodic

Running it only once per week would indeed be fine and would thus have a negligible impact on statistics (would that be another CI trigger?). We could also consider only running on Rawhide builds where the statistics don't really matter or on release builds.

> 2. Inject an additional header to say it's the test that we can then discount on the server side

Adding another header would require changes on the infra side. I'm not in favor of that option as this gives work to the Fedora Infra team.

> 3. (likely hard, probably not worth it) use a staging server

Using a staging server is another good option. I've asked about that.

travier:
> > The frequency with which we run this test will actually distort the statistics, right?
> 
> Yes, this will artificially increase the first Count Me bucket count.

The answer I got is that we are counting CI systems in all other cases (classic dnf/yum) so this should not be an issue. For us this is mostly true excepted that we might have people running fully stateless nodes which would never be older than 2 weeks old. But they should end up in the second (2 to 4 weeks) bucket: https://dnf.readthedocs.io/en/latest/conf_ref.html#countme-label

tests/kola/rpm-ostree-countme/test.sh
```diff
@@ -13,28 +13,56 @@ fatal() {
...
+	fi
+
+	# Check rpm-ostree count me output
+	output="$(journal_after_cursor)"
+	# depending on the stream, we expect different numbers of countme-enabled repos
+	if [[ "${output}" != "Successful requests: 1/1" ]] && \
+	   [[ "${output}" != "Successful requests: 2/2" ]] && \
+	   [[ "${output}" != "Successful requests: 3/3" ]]; then
+		echo "rpm-ostree-countme service ouput does not match expected sucess output (try: $i)"
```
jlebon: Should we print here what `output` actually was?

jlebon (dismissed):

tests/kola/rpm-ostree-countme/test.sh
travier: Good idea!

cgwalters (approved):

PR #1159, opened 2021-08-10
misc-ign-ro: change message from kube-watch to be more unique

miabbott:
We observed that the `kube-watch` test was occasionally failing
because the message from the service was being identified from `echo`
rather than the service name.  Change the service to emit a UUID that
we can check for instead of a generic log message.


dustymabe (approved):
LGTM

PR #1163, opened 2021-08-16
workaround failures in rawhide

dustymabe:
```
commit c9ccb6c377b430e595a9fb55970619123f76f714
Date:   Mon Aug 16 16:40:51 2021 -0400

    kola-denylist: snooze the toolbox test
    
    There is currently no registry.fedoraproject.org/fedora-toolbox:36
    container. Let's snooze for a few weeks in the rawhide stream and
    hopefully it will be in the registry by then.
    
    https://github.com/coreos/fedora-coreos-tracker/issues/926

commit c6f9e475bb64f703859d2f0b861763f23c2bf2c3
Date:   Mon Aug 16 16:36:30 2021 -0400

    manifests: workaround F37 GPG key issue for rawhide
    
    We'll workaround the issue by removing the F37 key from the list of
    gpgkeys to load. We'll also add a test that fails if the f37 entry
    isn't put back in place and we'll snooze that test. This gives us
    a reminder to followup on this issue if we don't get it resolved
    in upstream libdnf/rpm in the immediate future.
    
    https://github.com/coreos/fedora-coreos-tracker/issues/925
```

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -117,6 +117,17 @@ postprocess:
       echo 'DEFAULT_HOSTNAME=localhost' >> /usr/lib/os-release
     fi
 
+
+  # Workaround issue in rawhide parsing the F37 GPG key
+  # by removing it from the list of gpgkeys to load.
+  # https://github.com/coreos/fedora-coreos-tracker/issues/925
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    source /etc/os-release
+    [ ${VERSION_ID} -eq 36 ] || exit 0
+    sed -i 's|gpgkey=.*$|gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-rawhide-$basearch|' /etc/yum.repos.d/fedora-rawhide.repo
+
 # Packages listed here should be specific to Fedore CoreOS (as in not yet
 # available in RHCOS or not desired in RHCOS). All other packages should go
 # into one of the sub-manifests listed at the top.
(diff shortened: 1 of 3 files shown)


manifests/fedora-coreos-base.yaml
```diff
@@ -117,6 +117,18 @@ postprocess:
...
     fi
 
+
+  # Workaround issue in rawhide parsing the F37 GPG key
+  # by removing it from the list of gpgkeys to load.
+  # https://github.com/coreos/fedora-coreos-tracker/issues/925
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    # Only operate on F34 since F35+ has been fixed
```
jlebon: Stale comment.

tests/kola/yum-repos/test.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+source /etc/os-release
+if [ "$VERSION_ID" -eq "36" ]; then
```
jlebon, line 20: Aside: with the work in https://github.com/coreos/coreos-assembler/pull/2356, these kinds of tests will thankfully eventually not waste resources on other streams by just bringing up a VM only to have the test immediately no-op. OTOH, one can imagine a rawhide-specific test which *isn't* non-exclusive, and then https://github.com/coreos/coreos-assembler/pull/2356 doesn't help. I think it could make sense to have a mechanism in kola to have tests be filtered out based on the releasever or version string or something. (Making it stream-based is tempting, but then it's too easy to slip and lose coverage at releasever bump time if we forget to update the tests.)

jlebon (dismissed):

manifests/fedora-coreos-base.yaml
dustymabe: fixed

dustymabe:
merging since I only updated a comment

PR #1164, opened 2021-08-16
35coreos-ignition: allow setting kargs in live system if it's a no-op, via rdcore

### LINKED ISSUE #917: Ignition kargs helper should succeed on live systems if kargs are already correct

bgilbert:
In a live system, we can't set kargs in an Ignition config because we don't control the bootloader settings.  If the config specifies kargs anyway, we currently fail with `/dev/disk/by-label/boot: not a block device`, which isn't very clear.

In the live case, use the new `rdcore kargs --current` option to perform a dry run on `/proc/cmdline`, and if it doesn't create the stamp file, we don't need to do anything and can succeed anyway.  Otherwise, fail with a clearer error message.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/917.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-kargs.sh
@@ -1,4 +1,12 @@
 #!/bin/bash
 set -euo pipefail
 
-/usr/bin/rdcore kargs --boot-device /dev/disk/by-label/boot --create-if-changed /run/coreos-kargs-reboot "$@"
+if is-live-image; then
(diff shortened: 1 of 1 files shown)


bgilbert:
https://github.com/coreos/coreos-installer/pull/597 merged.

jlebon (dismissed):

cgwalters (dismissed):

dustymabe:
This just waiting on a `coreos-installer` release?

bgilbert:
Yup.

bgilbert:
Rebased and dropped WIP!

dustymabe (approved):
LGTM

PR #1167, opened 2021-08-18
overaly/preset: Enable Count Me by default

travier:
Enable rpm-ostree Count Me timer by default for existing and new
installs.

See: https://github.com/coreos/fedora-coreos-tracker/issues/717


dustymabe:
considering https://github.com/coreos/fedora-coreos-tracker/issues/717#issuecomment-901217206 do we still want this merged or will we build and fast track the rpm instead?

travier:
I think we should include this now and (partially) revert it later because the changes to the `fedora-release` package are bit more complicated and require a bit of process / review that may not be done in time for the next release.

cgwalters (approved):

PR #1171, opened 2021-08-19
ci/remove-graduated-overrides: enable for branched and rawhide

### LINKED ISSUE #1170: update graduated overrides job to handle rawhide
Sometimes the rawhide pungi runs won't succeed for many days on. Rather than wait on one of these to succeed (when we want new packages that have already been built with fixes to make it into the repos) we have started to fast-track builds. We should update our graduated overrides job to handle the rawhide branch as well to clean up the overrides.

jlebon:
In rare situations, we sometimes need to fast-track packages to these
mechanical streams, e.g. if pungi composes are failing and not updating
the repos. It's trivial to check those branches too for when the repos
are finally updated using the same GitHub Action, so let's do that.

Closes: #1170

### DIFF
--- a/.github/workflows/remove-graduated-overrides.yml
+++ b/.github/workflows/remove-graduated-overrides.yml
@@ -19,6 +19,8 @@ jobs:
         branch:
           - testing-devel
           - next-devel
+          - branched
+          - rawhide
       fail-fast: false
     steps:
       - run: dnf install -y rpm-ostree # see related TODO above


jlebon:
I've sanity-checked the script works correctly for this purpose on both streams.

sohankunkerkar (approved):

dustymabe:
> I've sanity-checked the script works correctly for this purpose on both streams.

nice.. this is what I was worried about. Was thinking we'd need to do something fancy for releasever

dustymabe (approved):
LGTM

PR #1175, opened 2021-08-23
pipeline.yaml: New file which configures the pipeline

cgwalters:
The idea is to keep the pipeline as more the mechanics/implementation,
and the config git repo is the declarative state.

One goal here is that e.g. `cosa upload-oscontainer` defaults to this value.

I also added `s3_bucket`.  Right now the pipeline writes this
as an annotation, but I think it will be easier to deal with if
e.g. `cosa buildupload` defaults to this.

### DIFF
--- a/pipeline.yaml
+++ b/pipeline.yaml
@@ -1,0 +1,8 @@
+# This file configures the pipeline, which is currently
+# https://github.com/coreos/fedora-coreos-pipeline
+
+# S3 bucket used for builds
+s3_bucket: fcos-builds
+
+# Location of the oscontainer (temporary value)
+oscontainer: quay.io/cgwalters/fcos


cgwalters:
This would help https://github.com/coreos/fedora-coreos-pipeline/pull/383 because then I could avoid the pipeline annotation bits.

jlebon:
We're definitely looking to move away from the annotation hack. Was chatting about this recently with @dustymabe and we were saying it should just be in the Jenkins configmap (which translates into: more JCASC drop-ins in https://github.com/coreos/fedora-coreos-pipeline/tree/main/jenkins/config).

Not opposed to having some knobs here, though leaving this repo scoped to generating artifacts and having releng-related configs centralized in https://github.com/coreos/fedora-coreos-pipeline/tree/main/jenkins/config feels cleaner. We just need to make it easier to configure.

On that topic, I definitely like the idea of having cosa directly pick up config values. One easy way to do this is to have JCASC dropins which define global env vars and have those read directly by cosa. E.g. `COSA_S3_BUCKET`, `COSA_OSCONTAINER`, etc...

jlebon:
I think this workflow implies pushing to your FCOS fork and pointing the pipeline to it, which is not necessarily required today if you're hacking on purely pipeline stuff (and if you're hacking on the pipeline, you're likely already operating on a branch of your fork of the pipeline repo with your modifications anyway).

dustymabe:
> Not opposed to having some knobs here, though leaving this repo scoped to generating artifacts and having releng-related configs centralized in https://github.com/coreos/fedora-coreos-pipeline/tree/main/jenkins/config feels cleaner.

👍

PR #1181, opened 2021-08-24
Add platform-specific console configuration metadata

bgilbert:
coreos-assembler `create_disk.sh` and `gf-set-platform` will read this.  `create_disk.sh` will also save a JSON version for the relevant architecture to `/boot/coreos/platforms.json` for coreos-installer.

We'll need to post advance warning to coreos-status before landing this.


dustymabe:
I know this depends on the backend implementation, but.. any chance we could or should use yaml here (supports comments), which may be quite useful in the future help add context for why a platform might need particular kernel arguments.

bgilbert:
YAML precludes the use of `jq`.  Unrecognized fields are ignored, though, so we could always add a `comment` field.

cgwalters:
I may have missed this, but it seems odd to ship things in the OS that we're not actually using at runtime.  We've been using `image.yaml` as kind of a dumping ground for "configure os-specific things for coreos-assembler".  We could also invent `bootloader.yaml` too in this repo e.g.

For `jq` a pattern used in cosa is to [convert to json](https://github.com/coreos/coreos-assembler/blob/1170a7e4b17664f0a431e2348e7ce1bd359ae164/src/cmdlib.sh#L150) in a temporary place and use `jq` on that.

bgilbert:
We don't strictly need it in `/usr`; that copy is only there for the convenience of `create_disk.sh` (and so we can use the `overlay.d` mechanism).  We do need it in `/boot`, though, for https://github.com/coreos/coreos-installer/pull/605.

You're right that we don't strictly need JSON in cosa, and coreos-installer could parse YAML if we want.  But for a machine-readable API it seems like JSON would be a better choice?

jlebon:
Not a fan of having them in *both* `/usr` and `/boot`. Also would prefer YAML for comments, but agree JSON would be better on disk.

How about having them as YAML files under a new `platforms/` dir or so in this repo (or even just a single `platforms.yaml` file I guess), and then cosa can just convert them to JSON and shove them into `/boot`?

dustymabe:
yay for YAML (at least in the most recent upload)

Any chance we can link to as much information as possible about why each platform/arch gets these values? platform documentation would be the most appropriate, but discovered knowledge (like investigation/discussion in GH issue comments) would also suffice. Basically it would be really nice in the future to know where to fish for similar information if we need to.

platforms.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - terminal_output serial
+    kernel_arguments:
+      - console=ttyS0,115200n8
+  packet:
+    grub_commands:
+      - serial --unit=1 --speed=115200
+      - terminal_input serial
+      - terminal_output serial
+    kernel_arguments:
+      - console=ttyS1,115200n8
```
dustymabe: so now on any other x86_64  platform we won't be setting a console by default (no `console=` entries) correct?

I think we probably need a announcement/time period for the "changes" that will occur to the x86_64 platforms because of this, which means we'd probably have to hardcode in this file what the current values are and then change them to what we desire them to be after the time period. WDYT?

We can certainly go ahead and apply things for `aarch64` because we have yet to ship it.

bgilbert: We won't set any console by default unless an arch/platform pair is listed in this file.

Yeah, we probably need a brief notice period.  https://github.com/coreos/coreos-assembler/pull/2400 already applies the legacy behavior if this file doesn't exist, but I can add a way to conditionalize that on a per-arch basis so we can get aarch64 switched sooner.

dustymabe: > We won't set any console by default unless an arch/platform pair is listed in this file.

+1 - that's what I thought. 

> 
> Yeah, we probably need a brief notice period. [coreos/coreos-assembler#2400](https://github.com/coreos/coreos-assembler/pull/2400) already applies the legacy behavior if this file doesn't exist, but I can add a way to conditionalize that on a per-arch basis so we can get aarch64 switched sooner.

Why not just leave https://github.com/coreos/coreos-assembler/pull/2400 the way it is and explicitly list all the platforms for x86_64 here? That way we don't have to change COSA again in the future.

Also we can insert but comment out the desired settings for the future and then after the notice period we just delete the block of current defaults and uncomment the new desired defaults.

bgilbert: Since this file only affects `gf-platformid` and `coreos-installer`, it can't affect the qemu or metal images, and so can't implement the legacy behavior.  (I missed that too, actually, and need to go fix up the cosa PR a bit.)  It doesn't seem worth fixing that, since then we'd need to have code to remove platform-specific args, not just add them.  cosa needs to special-case a missing `platforms.yaml` in any event, for ratcheting, and the additional arch test is cheap.

bgilbert:
Added comments, and a flag to disable the new behavior on x86_64 until a migration period has been announced.  Ready for review.

platforms.yaml
```diff
@@ -0,0 +1,63 @@
...
+  packet:
+    # https://metal.equinix.com/developers/docs/resilience-recovery/serial-over-ssh/#limitations
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial
+      - terminal_output serial
+    kernel_arguments:
+      - console=ttyAMA0,115200
+x86_64:
+  legacy: true  # continue using legacy behavior for now
```
dustymabe: I'd prefer the below to just be commented out to make it more obvious, but a substitute would be adding a few more words here

```suggestion
  legacy: true  # continue using legacy behavior for now, below config will not be used
```

dustymabe:
Thanks so much for adding the context links. That will be really really helpful in the future.

dustymabe (dismissed):
LGTM - will try to test later today

dustymabe:
ok one final question. I went through and built without `legacy: true`. I do feel like we still need to set `console=ttyS0` on our qemu image, no? https://github.com/coreos/fedora-coreos-tracker/issues/567 is about removing the console entry for bare metal. 

I understand that you said this file doesn't affect the `qemu` image, but I think it is worth discussing here. how we can or should behave towards qemu.

Note I'm specifically testing on `x86_64` right now.

bgilbert:
The `#agreed` in https://github.com/coreos/fedora-coreos-tracker/issues/567#issuecomment-662632315 says "platform-specific default on other platforms" (and it looks as though QEMU was not specifically discussed in that meeting).  At this point I strongly believe that on any platform that doesn't have specific requirements (e.g. AWS) we should provide no default console arguments and let users configure the system for their own needs.  We have Ignition `kernelArguments` support to help with that.  In particular with QEMU, new users will expect boot messages to go to the VGA console (e.g. in virt-manager or GNOME Boxes) and sending Ignition failures only to the serial console will continue to cause user confusion.

https://github.com/coreos/coreos-assembler/pull/2400 includes code specifically to ensure that invocation of QEMU images through kola will not be affected by this change.

platforms.yaml
bgilbert: As is, only one line needs to be changed to test the new behavior.  I've added an additional comment though.

dustymabe:
> The `#agreed` in [coreos/fedora-coreos-tracker#567 (comment)](https://github.com/coreos/fedora-coreos-tracker/issues/567#issuecomment-662632315) says "platform-specific default on other platforms" (and it looks as though QEMU was not specifically discussed in that meeting). At this point I strongly believe that on any platform that doesn't have specific requirements (e.g. AWS) we should provide no default console arguments and let users configure the system for their own needs. We have Ignition `kernelArguments` support to help with that. In particular with QEMU, new users will expect boot messages to go to the VGA console (e.g. in virt-manager or GNOME Boxes) and sending Ignition failures only to the serial console will continue to cause user confusion.

yeah - considering we have `legacy: true` set for now we can haggle over other possible desired platform defaults in new issues/discussions.

dustymabe (dismissed):
LGTM

platforms.yaml
bgilbert: With the updated cosa PR, this file can now affect `qemu` and `metal` images.  However, I've opted to continue implementing the legacy behavior by special-casing a missing config file, rather than making an interim config file that enumerates every arch and platform.  The latter seems more error-prone and wouldn't reduce the number of PRs we'll need to land this feature.

bgilbert:
Rebased and re-added the current qemu console configuration per https://github.com/coreos/fedora-coreos-tracker/issues/954.

dustymabe (dismissed):
LGTM

dustymabe:
I am going to try to look at https://github.com/coreos/fedora-coreos-tracker/issues/954 next week.

dustymabe:
When this merges can you coordinate with me. I need to unwind the [very ugly hack we did for aarch64 AWS images](https://github.com/coreos/fedora-coreos-pipeline/blob/84dc8ea4e151666e78d0e087e80e7d01db8d867c/jobs/build-arch.Jenkinsfile#L323-L325).

EDIT: actually that will probably need to be unwound when the COSA PR merges.

bgilbert:
@dustymabe I was thinking we could replace the aarch64 AWS hack with something like https://github.com/coreos/fedora-coreos-pipeline/pull/532?

bgilbert:
Added a kola test.

tests/kola/files/console-config
```diff
@@ -0,0 +1,74 @@
...
+}
+
+arch="$(uname -m)"
+platform="$(cmdline_arg ignition.platform.id)"
+
+# Check that platforms.json exists
+if [ ! -e /boot/coreos/platforms.json -a "$arch" != s390x ]; then
+    fatal "Expected platforms.json to exist on $arch"
+fi
+ok "Found platforms.json"
```
dustymabe: It's not immediately clear to me why we are excluding s390x here. I think I know why (`zipl`) but I think it's worth calling this out more clearly both in this file and in a comment at the top of `platforms.yaml`.

I also think it would make more sense if we did the s390x arch exclusion for this test at the top in the `kola: ` json fragment.

tests/kola/files/console-config
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+# kola: { "exclusive": false }
```
dustymabe: any chance we could get a description here and also an explanation of the key/value pairs in the json fragment similar to https://github.com/coreos/fedora-coreos-config/blob/a456f26a3f6fa4e77ce92a4cb83ea77c9f1b7ee8/tests/kola/systemd/sytemd-journald-fix#L2-L11

tests/kola/files/console-config
```diff
@@ -0,0 +1,74 @@
...
+    check_platforms_json aws "console=ttyS0,115200n8" "serial --speed=115200\nterminal_input serial\nterminal_output serial"
+    ;;
+aarch64)
+    check_platforms_json qemu "" "serial --speed=115200\nterminal_input serial console\nterminal_output serial console"
+    ;;
+ppc64le)
+    check_platforms_json qemu "console=hvc0 console=tty0" ""
+    ;;
+esac
+ok "platforms.json has expected contents"
```
dustymabe: I think what was important to me was the previous checks for `platforms.json matches grub.cfg and kargs`.

This one is above and beyond (thank you for putting extra effort in!), but has two drawbacks that I see:

- we need to update two places if we modify the `platforms.yaml`. I'm OK with that, but it does have this drawback, which I'm sure you're completely aware of already.
- it isn't exhaustive. For example, there is an aws aarch64 entry in `platforms.yaml`, but it's not checked here. Why not? Maybe we just enhance the comment at the top of this section and just randomly choose one platform from each arch to check? Other options: do a sha256sum and check it that way OR just remove the check altogether.

bgilbert: If we only check `platforms.json` against the kargs and grub.cfg, we're verifying that the config is consistent but not that it was correctly translated from `platforms.yaml`.  I didn't want to test every platform for the reason you said: we'd have to update the config in two places.  So I cherry-picked some platforms to test for specific problems.  I'll add some comments.

tests/kola/files/console-config
bgilbert: Ah yeah, I'll move the exclusion.  We're omitting it because `s390x` doesn't have any entries in `platforms.yaml`, so `platforms.json` is not created.  We don't have any `platforms.yaml` entries partially because of `zipl`, and partially because I don't think we have any customizations to put there.

bgilbert:
Updated!

tests/kola/files/console-config
dustymabe: I'm 👍 with your update so the below question is just me asking for my own clarity..

> we're verifying that the config is consistent but not that it was correctly translated from `platforms.yaml`

isn't it just a `yaml` -> `json` conversion. Are there risks involved here that I'm not aware of?

dustymabe (dismissed):
LGTM

tests/kola/files/console-config
bgilbert: It's a little more than that, since it copies only the section for the target architecture.  The main thing is to check that we didn't accidentally mis-extract and produce an empty JSON object.  The more detailed checks are just from an abundance of caution.

bgilbert:
Today's lesson: if a PR's head branch is renamed through the GitHub UI, GitHub closes the PR rather than following the rename.  :disappointed: 

Continuation in https://github.com/coreos/fedora-coreos-config/pull/1781.

PR #1189, opened 2021-08-26
tests/kola/authentication/passwd: Added tests for password authentication

### LINKED ISSUE #1169: Add test for password authentication
Add a test that provisions a user password via Ignition (replicating [docs](https://docs.fedoraproject.org/en-US/fedora-coreos/authentication/#_using_password_authentication)) and checks that the password works for local login.  (We should be able to do that from the command line, without adding new functionality to kola.)

Perhaps we should also test that the `passwd` command creates a `yescrypt` password hash (i.e., one starting with `$y$`).

gursewak1997:
Tested user password provisioned by ignition works and tested that passwd created yescrypt(pwd starting with $y$) password hash

### DIFF
--- a/tests/kola/authentication/passwd/config.fcc
+++ b/tests/kola/authentication/passwd/config.fcc
@@ -1,0 +1,7 @@
+variant: fcos
+version: 1.4.0
+passwd:
+  users:
+    - name: tester 
+      # encrypted version of 'foobar'. generated with `mkpasswd --method=yescrypt`
+      password_hash: $y$j9T$0HA8ReLTqRTccLKT0gzVY.$/e.OCrjePrh2tOm8CAoLqCMlZWS9q/WSAPBaZuopRs4
--- a/tests/kola/authentication/passwd/test.sh
+++ b/tests/kola/authentication/passwd/test.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+#Testing that a user password provisioned by ignition works
+OUTPUT=$(echo 'foobar' | setsid su - tester -c id)
+
+if [[ $OUTPUT != "uid=1001(tester) gid=1001(tester) groups=1001(tester) context=system_u:system_r:unconfined_service_t:s0" ]]; then
+    fatal "Failure when checking command output running with specified username and password"
+fi
+# yescrypt was changed to the default in Fedora 35
+# https://fedoraproject.org/wiki/Changes/yescrypt_as_default_hashing_method_for_shadow
+# Testing that passwd command creates a yescrypt password hash(starting with '$y$')
+source /etc/os-release 
+if [ "$VERSION_ID" -ge "35" ]; then
+    sudo useradd tester2
+    echo "42abcdef" | sudo passwd tester2 --stdin
+    PASSWD_CONFIRMATION=$(sudo grep tester2 /etc/shadow)
+    if [[ ${PASSWD_CONFIRMATION:0:11} != 'tester2:$y$' ]]; then
+        fatal "passwd did not create a yescrypt password hash"
+    fi
+fi
+ok "User-password provisioned and passwd command successfully tested"


tests/kola/authentication/passwd/config.fcc
```diff
@@ -0,0 +1,16 @@
...
+storage:
+  files:
+    - path: /etc/ssh/sshd_config.d/20-enable-passwords.conf
+      mode: 0644
+      contents:
+        inline: |
+          # Fedora CoreOS disables SSH password login by default.
+          # Enable it.
+          # This file must sort before 40-disable-passwords.conf.
+          PasswordAuthentication yes
```
dustymabe: Let's delete this part for now. If we later augment the test to do ssh password auth then we can add it back.

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,26 @@
...
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+#Testing that a user password provisioned by ignition works
+OUTPUT=$(echo '$y$j9T$lqqWvuo2n4YUmqMKefOAP0$TCSt423R7S/G86fGieA2x.3YPVcLHN1D4McfvB5fU3B' | setsid su - tester -c id)
```
dustymabe: hmm. I would assume we should be echoing in the unencrypted password here

tests/kola/authentication/passwd/config.fcc
```diff
@@ -0,0 +1,16 @@
+variant: fcos
+version: 1.4.0
+passwd:
+  users:
+    - name: tester 
+      password_hash: $y$j9T$lqqWvuo2n4YUmqMKefOAP0$TCSt423R7S/G86fGieA2x.3YPVcLHN1D4McfvB5fU3B
```
dustymabe: might be worth mentioning what the unencrypted version of the password is in a comment here.
 
```suggestion
      # encrypted version of 'foobar'. generated with `mkpasswd --method=yescrypt`
      password_hash: $y$j9T$lqqWvuo2n4YUmqMKefOAP0$TCSt423R7S/G86fGieA2x.3YPVcLHN1D4McfvB5fU3B
```

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ $OUTPUT != "uid=1001(tester) gid=1001(tester) groups=1001(tester) context=system_u:system_r:unconfined_service_t:s0"* ]]; then
```
dustymabe:
```suggestion
if [[ $OUTPUT != "uid=1001(tester) gid=1001(tester) groups=1001(tester) context=system_u:system_r:unconfined_service_t:s0" ]]; then
```

What is the value of the `*` here?

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,26 @@
...
+    fatal "Failed to login with specified username and passwd"
+fi
+#Testing that passwd command creates a yescryot password hash(starting with '$y$')
```
dustymabe: We might want to mention the change that introduced this new behavior

```suggestion
# yescrypt was changed to the default in Fedora 35
# https://fedoraproject.org/wiki/Changes/yescrypt_as_default_hashing_method_for_shadow
# Testing that passwd command creates a yescrypt password hash(starting with '$y$')

```

dustymabe: I assume also that this test will fail on < f35 so we'll need to conditionalize the test to be f35+

For an example see: https://github.com/coreos/fedora-coreos-config/blob/048b1ee32a3a3ab201926df43c366675de8ed2f4/tests/kola/misc-ro#L38-L45

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,26 @@
...
+PWD_CHANEG=$(echo -e "42abcdef\n42abcdef" | sudo passwd tester --stdin)
+PWD_CONFIRMATION=$(sudo cat /etc/shadow | grep tester)
+
+if [[ ${PWD_CONFIRMATION:0:10} != 'tester:$y$' ]]; then
```
dustymabe: Since we don't use `$PWD_CHANEG` anywhere we can just delete that variable and remove the `$()` from the command invocation.

For the varialbes let's use `PASSWD` instead of `PWD` (i.e. `PASSWD_CONFIRMATION` instead of `PWD_CONFIRMATION`). `$PWD` is a special variable in bash and my mind immediate thinks it's related to the working directory rather than password.

Also - is there a reason for the `\n` in the password could we just do `echo "42abcdef" |` instead?

Finally, let's apply this to a new user and not to the current `tester` user. so we'd add a `useradd tester2` before this and then we could `passwd tester2`.

What do you think?

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -xeuo pipefail
+
+ ok() {
```
dustymabe:
```suggestion
ok() {
```

dustymabe (commented):
Looks great @gursewak1997 - some comments

tests/kola/authentication/passwd/test.sh
gursewak1997: Agree to making a new user. Added tester2 and worked on other mentioned changes.

tests/kola/authentication/passwd/config.fcc
dustymabe: This one seems unresolved. It's optional, but helpful I think.

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,31 @@
...
+if [[ $OUTPUT != "uid=1001(tester) gid=1001(tester) groups=1001(tester) context=system_u:system_r:unconfined_service_t:s0" ]]; then
+    fatal "Failed to login with specified username and passwd"
+fi
+# yescrypt was changed to the default in Fedora 35
+# https://fedoraproject.org/wiki/Changes/yescrypt_as_default_hashing_method_for_shadow
+# Testing that passwd command creates a yescrypt password hash(starting with '$y$')
+source /etc/os-release 
+if [ "$VERSION_ID" -ge "35" ]; then
+    sudo useradd tester2
+    echo -e "42abcdef" | sudo passwd tester2 --stdin
```
dustymabe:
```suggestion
    echo "42abcdef" | sudo passwd tester2 --stdin
```
Is the `-e` needed here?

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,31 @@
...
+    PASSWD_CONFIRMATION=$(sudo cat /etc/shadow | grep tester2)
```
dustymabe: optional - we can just use grep here:

```suggestion
    PASSWD_CONFIRMATION=$(sudo grep tester2 /etc/shadow)
```

dustymabe:
Looks good.. we're almost there

tests/kola/authentication/passwd/config.fcc
gursewak1997: Ah you meant just putting the comment there. Will do it.

tests/kola/authentication/passwd/test.sh
gursewak1997: Always better off using less commands. Changed

tests/kola/authentication/passwd/test.sh
```diff
@@ -0,0 +1,31 @@
...
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+#Testing that a user password provisioned by ignition works
+OUTPUT=$(echo 'foobar' | setsid su - tester -c id)
+
+if [[ $OUTPUT != "uid=1001(tester) gid=1001(tester) groups=1001(tester) context=system_u:system_r:unconfined_service_t:s0" ]]; then
+    fatal "Failed to login with specified username and passwd"
```
dustymabe:
```suggestion
    fatal "Failure when checking command output running with specified username and password"
```

Since the failure could be a failure to log in OR a failure in the matching text let's make the error message a bit more ambiguous.

dustymabe (dismissed):
one last comment. LGTM

dustymabe (approved):
LGTM

PR #1190, opened 2021-08-26
tests: Split up misc-ro and mark tests as non-exclusive

saqibali-2k:
Moved 'z-ram is off by default' test from misc-ro and 'setup swap on zram device'
from misc-ign-ro into new files under kola/swap. Now misc-ro and misc-ign-ro
do not conflict and can be run in one VM after
coreos/coreos-assembler#2356 lands.

### DIFF
--- a/tests/kola/swap/zram-generator/test.sh
+++ b/tests/kola/swap/zram-generator/test.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# This test conflicts with swap/zram-default so we cannot set this to non-exclusive
+# kola: { "exclusive": true}
+set -xeuo pipefail
+
+ok() {
+        echo "ok" "$@"
+    }
+
+fatal() {
+        echo "$@" >&2
+            exit 1
+        }
+
+# This test makes sure that swap on zram devices can be set up
+# using the zram-generator as defined in the docs at
+# https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-configure-swaponzram/
+
+if ! grep -q 'zram0' /proc/swaps; then
+    fatal "expected zram0 to be set up"
+fi
+ok "swap on zram was set up correctly"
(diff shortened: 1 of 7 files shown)


dustymabe:
let's move the test to a different directory that more defines what it is testing. how about `tests/kola/swap/` or something? the idea here is that `misc-ro` and `misc-ign-ro` go away at some point. They were really only there to consolidate many tests into a single VM anyway and we no longer really need that.

Also let's break out the swap on zram stuff from `misc-ign-ro` into it's own test and we'll mark that one as `exclusive: true` (mainly because it requires ignition config and the other test doesn't). The `misc-ro` swap on zram stuff can be `exclusive: false`.  Though the relationship between these two tests does make me think of a new feature where tests can mention other tests they conflict with. If a set of tests doesn't include the other test they conflict with then they can be included. That's for a later time, though.

saqibali-2k:
> the idea here is that `misc-ro` and `misc-ign-ro` go away at some point. They were really only there to consolidate many tests into a single VM anyway and we no longer really need that.

Ok makes sense. Given that that the other tests in misc-ro and misc-ign-ro do not conflict, is it worthwhile to break out *all* of the tests or should we just stick to separating the swap on zram tests for now?

dustymabe:
> Ok makes sense. Given that that the other tests in misc-ro and misc-ign-ro do not conflict, is it worthwhile to break out _all_ of the tests or should we just stick to separating the swap on zram tests for now?

I think for our first PR we should just keep it scoped to a few tests like you're doing here, but I absolutely think our next step is to try to find more appropriate homes for the rest of the `misc*ro` stuff. Maybe after a few weeks of this exclusive tests model proving itself.

tests/kola/swap/zram-generator/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# kola: { "exclusive": true}
```
dustymabe, line 3: let's mention why it is exclusive here.. specifically it conflicts with the `ext.config.swap.zram-generator`. If that test didn't exist then we'd probably be fine as `exclusive: false`.

dustymabe:
Nice work. This is looking really good.

dustymabe (approved):
LGTM

PR #1194, opened 2021-08-30
s390x: add keyfile for LUKS encrypted root into initramfs

nikita-dubrovskii:
IBM s390x doesn't use/support TPM. Setting up a separate VM for Tang (or allowing local zKVM guests to access remote Tang) could be undesirable. In this case it's easier to inject `/etc/crypttab` and `/etc/luks/root` into `ramdisk`.

### DIFF
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -22,6 +22,7 @@ packages-s390x:
   # On Fedora, this is provided by s390utils-core. on RHEL, this is for now
   # provided by s390utils-base, but soon will be -core too.
   - /usr/sbin/zipl
+  - /usr/bin/genprotimg
 packages-x86_64:
   - grub2 grub2-efi-x64 efibootmgr shim
   - microcode_ctl
--- a/overlay.d/05core/usr/libexec/coreos-ignition-firstboot-complete
+++ b/overlay.d/05core/usr/libexec/coreos-ignition-firstboot-complete
@@ -4,6 +4,9 @@ set -euo pipefail
 mount -o remount,rw /boot
 
 if [[ $(uname -m) = s390x ]]; then
+    if [[ -f /etc/luks/root && -f /etc/crypttab ]]; then
+        rpm-ostree initramfs --enable --arg="-I" --arg="/etc/crypttab /etc/luks/root"
+    fi
     zipl
 fi
 


cgwalters:
The idea of TPM bound LUKS (as well as NBDE) is that if e.g. the hard drive is removed from the machine and sent to the manufacturer for analysis, it can't be decrypted.

But injecting the key into the initramfs means the key will travel with the hard drive, so the security properties of this are very small and IMO misleading.

I think we should focus on supporting something TPM *like* on s390x even if not via actual TPM protocol.  It's surprising to me that the hardware doesn't have something even roughly similar.

bgilbert (changes requested):
Hi @nikita-dubrovskii!  :wave: It looks like this PR might be part of the implementation effort for s390x Secure Execution.  If so, I still think it's important to capture the design in a design doc, and for that design doc to be widely reviewed, before any code is merged.  Without a written spec describing the threats we want to address and how the code will accomplish that, it'll be hard for code review to ensure the code is fit for purpose.  Secure Execution is a security feature with several moving parts, so let's make sure we get the implementation right.

If this PR isn't intended for Secure Execution, I agree with Colin that storing the cryptroot key in plaintext on the same disk could mislead users into thinking that their disk is securely encrypted when it is not.  Could you provide more details about the problem you're trying to solve?

nikita-dubrovskii:
Hi @bgilbert , indeed i was experimenting with SE and LUKS, but this PR could be used separately. I've pushed it to start tracking activity, because without LUKS-support SE makes not that much sense. 
I've also updated [GDoc](https://docs.google.com/document/d/1CM-U0zqAJ5iNN2oLuwXIbhfIl9en4qBEivXxqwX5Obo/edit) with some info about how SE works

cgwalters:
OK right, let's call out the integration with SE here.  The key (pun intended) bit is that SE encrypts the kernel/initramfs.

But I think we should have a higher level flag file for "SE mode enabled" and dispatch on that; something written by Ignition into `/run` say.

Also, don't we still need to call `zipl` in this mode?  Or I guess the `rpm-ostree initramfs` bit is really a placeholder for where we need to invoke the tool to encrypt the kernel/initramfs, and that would also take over the role of invoking `zipl`?

nikita-dubrovskii:
> Also, don't we still need to call `zipl` in this mode?

i was sure that `rpm-ostree / ostree` calls `zipl` automatically during `initramfs / kargs` modification. 

> Or I guess the `rpm-ostree initramfs` bit is really a placeholder for where we need to invoke the tool to encrypt the kernel/initramfs, and that would also take over the role of invoking `zipl`?

not exactly, we need an updated (with luks key within) `initramfs`, which is later encrypted , and finally we have to update BLS config before calling `zipl`

PR #1197, opened 2021-08-31
kola-denylist: extend snooze for the toolbox test

dustymabe:
registry.fedoraproject.org/fedora-toolbox:36 still doesn't exist.
See https://github.com/coreos/fedora-coreos-tracker/issues/926


dustymabe:
This one doesn't need CI.

ravanelli (approved):
LGTM

PR #1198, opened 2021-09-01
Added test case for DNS in rootless podman network

### LINKED ISSUE #1180: add test for podman network
The reproducer for https://github.com/coreos/fedora-coreos-tracker/issues/923 is simple enough we should consider adding a test for it.

```
# as non-root user
podman network create testnetwork
podman run --rm -it --network=testnetwork registry.fedoraproject.org/fedora:34 getent hosts google.com
echo $?
```

gursewak1997:
Running a test case in rootless podman container to check whether we can DNS resolve external domains.
The original issue: https://github.com/coreos/fedora-coreos-tracker/issues/923.

### DIFF
--- a/tests/kola/podman/dns/test.sh
+++ b/tests/kola/podman/dns/test.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# Tests that rootless podman containers can DNS resolve external domains.
+# https://github.com/coreos/fedora-coreos-tracker/issues/923
+# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv", "exclusive": false}
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+runascoreuserscript='
+#!/bin/bash
+set -euxo pipefail
+
+podman network create testnetwork
+podman run --rm -t --network=testnetwork registry.fedoraproject.org/fedora:34 getent hosts google.com
+podman network rm testnetwork
+'
+
+runascoreuser() {
+    # NOTE: If we don't use `| cat` the output won't get copied
+    # and won't show up in the output of the ext test.
+    sudo -u core "$@" | cat
+}
+
+main() {
+    echo "$runascoreuserscript" > /tmp/runascoreuserscript
+    chmod +x /tmp/runascoreuserscript
+    if ! runascoreuser /tmp/runascoreuserscript ; then 
+        fatal "DNS in rootless podman testnetwork failed. Test Fails" 
+    else 
+        ok "DNS in rootless podman testnetwork Suceeded. Test Passes" 
+    fi
+}
+
+main


tests/kola/podman/dns/test.sh
```diff
@@ -0,0 +1,23 @@
...
+sudo su - tester
+podman network create testnetwork
+
+OUTPUT=$(podman run --rm -t --network=testnetwork registry.fedoraproject.org/fedora:34 getent hosts google.com)
+
+if [[ ${OUTPUT} != "2607:f8b0:4005:807::200e google.com"* ]]; then
+    fatal "DNS in rootless podman testnetwork failed. Test Fails"
+else
+    ok "DNS in rootless podman testnetwork Suceeded. Test Passes"
+fi
```
dustymabe: I think checking the output of the command is not ideal here (because the IP address can change). It would be better to just check to see if the command was successful or not.

```suggestion
if !  podman run --rm -t --network=testnetwork registry.fedoraproject.org/fedora:34 getent hosts google.com; then
    fatal "DNS in rootless podman testnetwork failed. Test Fails"
else
    ok "DNS in rootless podman testnetwork Suceeded. Test Passes"
fi
```


Let's also delete the `testnetwork` after the test is done.

tests/kola/podman/dns/config.fcc
```diff
@@ -0,0 +1,7 @@
+variant: fcos
+version: 1.4.0
+passwd:
+  users:
+    - name: tester 
+      # encrypted version of 'foobar'. generated with `mkpasswd --method=yescrypt`
+      password_hash: $y$j9T$0HA8ReLTqRTccLKT0gzVY.$/e.OCrjePrh2tOm8CAoLqCMlZWS9q/WSAPBaZuopRs4
```
dustymabe: You shouldn't need this file at all I don't think. You can just use the `core` user as an unprivileged user (gets created by default).

dustymabe (commented):
Nice work.. some comments.

tests/kola/podman/dns/test.sh
```diff
@@ -0,0 +1,22 @@
...
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+podman network create testnetwork
+
+if !  podman run --rm -t --network=testnetwork registry.fedoraproject.org/fedora:34 getent hosts google.com; then
```
dustymabe:
```suggestion
if ! podman run --rm -t --network=testnetwork registry.fedoraproject.org/fedora:34 getent hosts google.com; then
```

extra space

tests/kola/podman/dns/test.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
dustymabe, line 1:
```suggestion
#!/bin/bash
# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv", "exclusive": "false"}
```

- we need the internet to grab container from the registry
- no need to run on more than one platform (if it passes here it should pass everywhere), so just run on qemu.
- let's mark it as `exclusive: false` so it can run with other tests

dustymabe (commented):
A few things..

- let's add more information to the commit message
    - link to the upstream issue that caused us to identify the need for this test
    - also add `Fixes: <link>` for the fcos configs issue

- change the `test.sh` file to executable and re-add to git `chmod +x test.sh && git add test.sh && git commit --amend`


- we got rid of the `su -` so we're now running as `root` (privileged). We need to make sure we are running unprivileged as the `core` user.
    - One way you can get a sanity check that this test is properly failing is by running this test against a version of FCOS that had [the issue](https://github.com/coreos/fedora-coreos-tracker/issues/923). According to the bug it hasn't yet been fixed in our stable release so you should be able to test against it with something like:

```
cosa kola run ext.config.podman.dns --qemu-image=fedora-coreos-34.20210808.3.0-qemu.x86_64.qcow2
```

👆 SHOULD FAIL

tests/kola/podman/dns/test.sh
```diff
@@ -0,0 +1,39 @@
...
+set -xeuo pipefail
+
+# Tests that rootless podman containers can DNS resolve external domains.
```
dustymabe, line 4: optional: link back to the original issue for context.

```suggestion
# Tests that rootless podman containers can DNS resolve external domains.
# https://github.com/coreos/fedora-coreos-tracker/issues/923
```

tests/kola/podman/dns/test.sh
```diff
@@ -0,0 +1,39 @@
...
+#!/bin/bash
+set -euxo pipefail
+
+podman network create testnetwork
+podman run --rm -t --network=testnetwork registry.fedoraproject.org/fedora:34 getent hosts google.com
+podman network rm testnetwork
+'
+
+runascoreuser() {
+    sudo -u core "$@" | cat
```
dustymabe, line 29: optional: add a comment explaining why we use `| cat` here like is done in the podman rootless systemd test we pulled this from

dustymabe (dismissed):
LGTM. I added a few comments for optional items.

Also let's change our commit message slightly:

From:

```
tests/kola/podman/dns: Added test for DNS in rootless podman network
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/923
Issue: https://github.com/coreos/fedora-coreos-config/issues/1180
```

To:

```
tests/kola/podman/dns: Added test for DNS in rootless podman network

Original Issue: https://github.com/coreos/fedora-coreos-tracker/issues/923
Fixes: https://github.com/coreos/fedora-coreos-config/issues/1180
```

So we separate the title of the commit from the commit description, link back to the original issue (for context) and also `Fixes` for the remaining open fcos configs issue (this PR will close that issue when it merges).

dustymabe:
Nice work @gursewak1997!

saqibali-2k (approved):
Verified success on latest build and failure on [fedora-coreos-34.20210808.3.0](https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/34.20210808.3.0/x86_64/fedora-coreos-34.20210808.3.0-qemu.x86_64.qcow2.xz). LGTM

dustymabe (approved):

PR #1204, opened 2021-09-07
ci: enable kola basic scenarios and many more ISO tests

bgilbert:
Basic scenarios get us UEFI boot testing with the bare-metal image, plus an NVMe boot test.  Switching to the coreos-ci-lib `kola testiso` wrapper adds `iso-live-login` and `iso-as-disk` scenarios, plus tests on metal4k, multipath, and UEFI.  It also drops the `iso-install` scenario, which isn't a typical use case for FCOS.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -45,13 +45,7 @@ cosaPod {
 
     stage("Test ISO") {
         shwrap("cd /srv/fcos && cosa buildextend-live")
-        try {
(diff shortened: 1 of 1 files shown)


jlebon (dismissed):

bgilbert:
Reworked this.

jlebon (approved):
Yeah, given that this is the integration point for everything it makes sense to go all out on testing here.

PR #1205, opened 2021-09-08
tests/misc-ro: check initrd for non-executable scripts

bgilbert:
If a script is committed to Git without `+x`, Dracut will happily install it into the initrd without `+x`.  Add a simple check for this.

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -189,3 +189,12 @@ if ! systemctl show -p ActiveState network-online.target | grep -q ActiveState=i
     fatal "Unit network-online.target shouldn't be active"
 fi
 ok "unit network-online.target inactive"
+
+# It's easy for dracut modules to accidentally ship scripts without +x set
+tmpd=$(mktemp -d)
+( cd ${tmpd} && lsinitrd --unpack /boot/ostree/*/init* )
+if find ${tmpd}/usr/{bin,sbin,libexec} ! -perm -0111 | grep -v clevis-luks-common-functions; then
+    fatal "Found non-executable scripts in initrd"
+fi
+rm -r ${tmpd}
+ok "All initrd scripts are executable"


tests/kola/misc-ro
```diff
@@ -189,3 +189,11 @@ if ! systemctl show -p ActiveState network-online.target | grep -q ActiveState=i
     fatal "Unit network-online.target shouldn't be active"
 fi
 ok "unit network-online.target inactive"
+
+# It's easy for dracut modules to accidentally ship scripts without +x set
+mkdir ~/initrd
+( cd ~/initrd && lsinitrd --unpack /boot/ostree/*/init* )
+if find ~/initrd/usr/{bin,sbin,libexec} ! -perm -0111 | grep -v clevis-luks-common-functions; then
+    fatal "Found non-executable scripts in initrd"
+fi
```
cgwalters, line 198:
```suggestion
fi
rm -rf ~/initrd
```

cgwalters (dismissed):
I think this test could also run in `ci/validate` - looping over `overlay.d/*/usr/lib/dracut/modules.d` and checking for files that start with a shebang.

That would mean we don't need to skip the clevis bit, we're only testing our own code.  And it'd be faster.  But, it is clearly useful to test the final initramfs too.

We don't need a booted VM for this test at all; this test could also be part of a post-build linting phase in coreos-assembler.

All that said, obviously fine as is!

bgilbert:
Part of the goal is to catch problems in upstream repos such as Ignition and Afterburn, where this test will run as part of CoreOS CI.  `ci/validate` wouldn't catch those.  If there's a good place to put post-build lints, I'd be happy to move it over, but I'm hoping to avoid turning this into a Project.  :slightly_smiling_face:

jlebon (approved):

PR #1207, opened 2021-09-08
35coreos-ignition: randomize partition GUIDs on first boot

bgilbert:
We've been randomizing the disk GUID but not the partition GUIDs, contrary to the GPT spec.  Partition GUIDs are already not stable identifiers, since new ones are generated by `create_disk.sh` on every image build, so ensure they're globally unique as well.

Randomizing the ESP GUID will break any existing UEFI boot variable created by `fallback.efi`, so it's important that we only do this on the first boot (to not break existing systems) and only after our bootloader setup has been taught not to create boot variables (to not break new systems).

Requires https://github.com/coreos/coreos-assembler/pull/2421 or equivalent.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
@@ -38,5 +38,5 @@ if [ "${PTUUID:-}" != "$UNINITIALIZED_GUID" ]; then
 fi
 
 echo "Randomizing disk GUID"
-sgdisk --disk-guid=R --move-second-header "$PKNAME"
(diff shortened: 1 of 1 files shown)


cgwalters:
This will make the hack "touch /boot/ignition.firstboot" for https://github.com/coreos/fedora-coreos-tracker/issues/399 more likely to break booting right?

I think I'd say we should do something like hardcode "unconfigured" GUIDs in cosa, and then have this service be idempotent by only randomizing if it sees the unconfigured GUIDs?  Or alternatively, is there a place in the GPT that we can store a bit of state?

Anyways, code seems fine as is.

bgilbert:
> I think I'd say we should do something like hardcode "unconfigured" GUIDs in cosa, and then have this service be idempotent by only randomizing if it sees the unconfigured GUIDs?

We're already doing that; see line 35.

PR #1208, opened 2021-09-10
05core/coreos-liveiso-success: fix hang with systemd 239

bgilbert:
RHEL 8 has systemd 239, which has https://github.com/systemd/systemd/issues/9374.  As a result, `journalctl -f ... | head -1` hangs because `journalctl` doesn't properly exit.  Rewrite the check to call `loginctl` once per second until it finds a session.  This approach is uglier, but the service only runs in the live ISO, on QEMU, when there's no Ignition config, and when kola's virtio port is open, so the minimal extra overhead shouldn't affect production systems.

Fixes the `kola testiso` `iso-live-login` scenario in RHCOS.


cgwalters (approved):
Ahhhh!  Nice work debugging this.

bgilbert:
RHEL BZ in https://bugzilla.redhat.com/show_bug.cgi?id=2003236.

PR #1191, opened 2021-08-26
manifests/fedora-coreos-base: stop disabling modular repos

jlebon:
The latest rpm-ostree release now has proper support for modules. It
will not layer modular packages unless explicitly enabled. So let's stop
disabling the repos so e.g. `rpm-ostree ex module install cri-o:1.20`
alone just works.

PR #1211, opened 2021-09-10
tests/misc-ro: drop `rpm-ostree ex module install` test

jlebon:
Thought this was cool to have here, but first it needs Internet, which
the rest of `misc-ro` doesn't so far which is nice. And second each
Fedora version has different stream versions so the matrix is slightly
trickier.

I think I'll put it in the rpm-ostree testsuite instead. Anyway if we
really go the modularity route for cri-o, ideally we'd have some k8s
testing going on in CI here on top of OKD or Typhoon, and then this will
implicitly be tested.

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -198,6 +198,3 @@ if find ${tmpd}/usr/{bin,sbin,libexec} ! -perm -0111 | grep -v clevis-luks-commo
 fi
 rm -r ${tmpd}
 ok "All initrd scripts are executable"
-
-rpm-ostree ex module install cri-o:1.20/default --dry-run
-ok "basic modularity support"


jlebon:
> Thought this was cool to have here, but first it needs Internet, which
> the rest of `misc-ro` doesn't so far which is nice.

I think the reason this worked originally in https://github.com/coreos/fedora-coreos-config/pull/1191 was because of https://github.com/coreos/coreos-assembler/pull/2356, which bundled misc-ro with other non-exclusive tests which do require networking.

bgilbert (approved):

dustymabe:
why can't we make this it's own separate test and just mark it as needing internet? I think the test is still valuable.

jlebon:
> why can't we make this it's own separate test and just mark it as needing internet? I think the test is still valuable.

We can... I'm just hesitant to add tests which depend on outside resources. Not planning to get rid of the test, just moving it to rpm-ostree. Edit: https://github.com/coreos/rpm-ostree/pull/3123

dustymabe:
Yeah. I think if cri-o is a part of our solution to https://github.com/coreos/fedora-coreos-tracker/issues/767 then we probably at least want to verify it layers find on each build. WDYT?

jlebon:
> Yeah. I think if cri-o is a part of our solution to [coreos/fedora-coreos-tracker#767](https://github.com/coreos/fedora-coreos-tracker/issues/767) then we probably at least want to verify it layers find on each build. WDYT?

I think that makes sense, though ideally as mentioned earlier we sanity-test OKD or Typhoon in CI instead.

PR #1212, opened 2021-09-13
overlay: Remove /boot RW workaround for kdump

travier:
Remove workaround keeping `/boot` read only when kdump support was
requested now that kdump writes the temporary initrd in
`/var/lib/kdump`.

This reverts 3167d83 12kdump: Remount /boot rw before `kdump.service`.

### DIFF
--- a/tests/kola/kdump/test.sh
+++ b/tests/kola/kdump/test.sh
@@ -1,12 +1,9 @@
 #!/bin/bash
 set -xeuo pipefail
 # https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/
-# kola: {"minMemory": 4096, "tags": "skip-base-checks"}
-
-# ===== FIXME: Disabled due to broken CI
-echo "Test disabled"
-exit 0
-# =====
+# Only run on QEMU and x86_64 & aarch64 for now:
+# https://github.com/coreos/fedora-coreos-tracker/issues/860
+# kola: {"platforms": "qemu-unpriv", "minMemory": 4096, "tags": "skip-base-checks", "architectures": "x86_64 aarch64"}
 
 fatal() {
     echo "$@" >&2
--- a/overlay.d/12kdump/statoverride
+++ b/overlay.d/12kdump/statoverride
@@ -1,2 +1,0 @@
-# Config file for overriding permission bits on overlay files/dirs
-# Format: =<file mode in decimal> <absolute path to a file or directory>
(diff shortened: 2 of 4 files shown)


travier:
I will do manual testing on FCOS (& RHCOS) to make sure this is OK as the kdump test is currently disabled. (https://github.com/coreos/fedora-coreos-tracker/issues/860)

jlebon (dismissed):
Shouldn't the commit message say "keeping `/boot` read-write"?

travier:
Marking as draft as we need a backport for F34 first: https://bugzilla.redhat.com/show_bug.cgi?id=2003780

dustymabe:
> Marking as draft as we need a backport for F34 first: https://bugzilla.redhat.com/show_bug.cgi?id=2003780

if it's a lot of effort we could just wait for f35 (which is coming pretty soon anyway).

dustymabe:
if you wanted to go ahead and apply this to only f35 now you could move the files into a post script in the manifests (conditional on f35+) and then change the test to use denylist instead and only skip it on f34. We've done this in other places.

travier:
> > Marking as draft as we need a backport for F34 first: https://bugzilla.redhat.com/show_bug.cgi?id=2003780
> 
> if it's a lot of effort we could just wait for f35 (which is coming pretty soon anyway).

I don't think it's a lot of effort but we need the kexec-tools team to make it happen.

> if you wanted to go ahead and apply this to only f35 now you could move the files into a post script in the manifests (conditional on f35+) and then change the test to use denylist instead and only skip it on f34. We've done this in other places.

I think that's more efforts for not much gain. I'll wait as this is not blocking RHCOS.

travier:
https://github.com/coreos/fedora-coreos-config/commit/6cde71e15ec923884d6cbfa83b32ff213d64b124 > This one includes the fixed ~kdump~ `kexec-tools` package.

tests/kola/kdump/test.sh
```diff
@@ -1,12 +1,9 @@
...
 set -xeuo pipefail
 # https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/
-# kola: {"minMemory": 4096, "tags": "skip-base-checks"}
-
-# ===== FIXME: Disabled due to broken CI
-echo "Test disabled"
-exit 0
-# =====
+# Only run on QEMU and x86_64 & aarch64 for now:
+# https://github.com/coreos/fedora-coreos-tracker/issues/860
```
dustymabe, line 5: Hmm. I thought the work you were doing here was to close out https://github.com/coreos/fedora-coreos-tracker/issues/860 

Is this test still failing on cloud platforms and needs further investigation there?

travier: This is the first step: get it to work on QEMU in CI.
I'll work on investigating issues in cloud platforms as a second step.

travier:
```
[2021-10-14T15:23:30.335Z] --- PASS: ext.config.kdump (128.26s)
```

tests/kola/kdump/test.sh
dustymabe: 👍

dustymabe (approved):
LGTM

PR #1215, opened 2021-09-16
live/grub.cfg: update incorrect comment

bgilbert:
This file is shipped in the ISO itself, and `efiboot.img` contains a stub config that points to it.

### DIFF
--- a/live/EFI/fedora/grub.cfg
+++ b/live/EFI/fedora/grub.cfg
@@ -4,9 +4,11 @@
 #
 # One diff to note is we use linux and initrd instead of linuxefi and
 # initrdefi. We do this because it works and allows us to use this same
-# file on other architecutres. https://github.com/coreos/fedora-coreos-config/issues/63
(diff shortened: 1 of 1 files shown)


dustymabe:
> This file is shipped in the ISO itself, and `efiboot.img` contains a stub config that points to it.

has that always been true and the comment was always wrong or did it change recently? Either way the `efiboot.img contains a stub config that points to it.` might be a good detail to replace the comment with. The knowledge in this area is limited and scarce and easy to forget. Any breadcrumbs we can leave IMO is totally worth doing.

bgilbert:
It was true until coreos/coreos-assembler@81aae8eab1fba164646336f0849f8cdf4d15d7ba.  So it turns out that the failure when booting from a USB stick was my bug in the first place. :tada:

Updated.

cgwalters (approved):

dustymabe (approved):
LGTM

PR #1219, opened 2021-09-20
Revert "overaly/preset: Enable Count Me by default"

travier:
This is now enabled via the `90-default.preset` from fedora-release
(included since commit 83f5e120101b125ad399acfc87fbf9d2744108c3).

See:
  - https://src.fedoraproject.org/rpms/fedora-release/pull-request/203
  - https://bugzilla.redhat.com/show_bug.cgi?id=1995495

This reverts commit 12ba5c2922d59fb2451ff3fbbb033913538b9460.

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -8,6 +8,3 @@ enable coreos-check-cgroups.service
 # Clean up injected Ignition config in /boot on upgrade
 # https://github.com/coreos/fedora-coreos-tracker/issues/889
 enable coreos-cleanup-ignition-config.service
-# Temporary fast track for rpm-ostree count me enablement
-# https://github.com/coreos/fedora-coreos-tracker/issues/717
-enable rpm-ostree-countme.timer


dustymabe:
Looks like the change hasn't quite made it to f35. Should we hold this until that point?
- https://bodhi.fedoraproject.org/updates/FEDORA-2021-8e7f97ebe1

travier:
I'm not sure I follow. The link above says:

```
This update has been submitted for stable by bodhi. 2 weeks ago
```

And I've made a PR for F34 which is now the right version in testing-devel.

dustymabe:
> I'm not sure I follow. The link above says:
> 
> ```
> This update has been submitted for stable by bodhi. 2 weeks ago
> ```

correct, but it hasn't reached `stable` because Fedora is in beta freeze.

> 
> And I've made a PR for F34 which is now the right version in testing-devel.

`testing-devel` is fine. It's the f35 based stream (currently only `next-devel`) that we'll hit this on. I imagine a test would start failing on that stream if we merged this without waiting OR fast-tracking (in an override) that bodhi update for f35.

travier:
> > I'm not sure I follow. The link above says:
> > ```
> > This update has been submitted for stable by bodhi. 2 weeks ago
> > ```
> 
> correct, but it hasn't reached `stable` because Fedora is in beta freeze.
> 
> > And I've made a PR for F34 which is now the right version in testing-devel.
> 
> `testing-devel` is fine. It's the f35 based stream (currently only `next-devel`) that we'll hit this on. I imagine a test would start failing on that stream if we merged this without waiting OR fast-tracking (in an override) that bodhi update for f35.

Oh indeed. Let's way for after the freeze once everything as landed.

dustymabe:
Should be good now for f35 I think

dustymabe (approved):
LGTM

PR #1220, opened 2021-09-20
tests/kola/var-mount: add a TPM-less variant for s390x

julianwiedmann:
This seems sufficient to get ext.config.var-mount working on s390x.


jlebon (approved):
Yeah makes sense. I guess even on non-s390x it's good to have at least one test which checks that a vanilla `/var` non-LUKS mount works.

PR #1233, opened 2021-09-24
overlay/dracut/multipath: remove problematic stop command

### LINKED ISSUE #803: flaky multipathd service
Sometimes, the generic kola check for failed systemd services trips on multipathd:

```
--- FAIL: coreos.tls.fetch-urls (120.75s)
        harness.go:976: Cluster failed starting machines: machine "5cba2624-8aa0-4fbb-9023-cc0287c3f1e7" failed basic checks: some systemd units failed:
● multipathd.service loaded failed failed Device-Mapper Multipath Device Controller
```

This one was hit during this week's next release:
https://github.com/coreos/fedora-coreos-streams/issues/296

[kola.zip](https://github.com/coreos/fedora-coreos-tracker/files/6393342/kola.zip)

lucab:
This is a temporary workaround to remove a problematic stop command
from `multipathd.service`, until the already-merged proper fix gets
released in dracut.

Ref: https://github.com/dracutdevs/dracut/pull/1606
Closes: https://github.com/coreos/fedora-coreos-tracker/issues/803

### DIFF
--- a/tests/kola/misc-ro
+++ b/tests/kola/misc-ro
@@ -198,3 +198,16 @@ if find ${tmpd}/usr/{bin,sbin,libexec} ! -perm -0111 | grep -v clevis-luks-commo
 fi
 rm -r ${tmpd}
 ok "All initrd scripts are executable"
+
+# We need either a fixed dracut or temporary workaround, no need for both.
+# See https://github.com/coreos/fedora-coreos-tracker/issues/803.
+has_fixed_dracut=$(grep -q 'ExecStop=/sbin/multipathd shutdown' /usr/lib/dracut/modules.d/90multipath/multipathd.service; echo $?)
+has_overlay_quickfix=$(test ! -f /usr/lib/dracut/modules.d/35coreos-multipath/90-multipathd-remove-execstop.conf; echo $?)
+if test "${has_fixed_dracut}" -eq "${has_overlay_quickfix}"; then
+    if test "${has_fixed_dracut}" -eq 1; then
+        fatal "Found fixed dracut multipath module but quickfix is present too"
+    else
+        fatal "Found buggy dracut multipath module but quickfix is missing too"
+    fi
+fi
+ok "either dracut multipath module fixed or quickfix present"
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/90-multipathd-remove-execstop.conf
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/90-multipathd-remove-execstop.conf
@@ -1,0 +1,3 @@
+# Temporary workaround for https://github.com/dracutdevs/dracut/pull/1606.
+[Service]
+ExecStop=
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/module-setup.sh
@@ -12,6 +12,11 @@ install_ignition_unit() {
 }
 
 install() {
+    # Temporary workaround for https://github.com/dracutdevs/dracut/pull/1606.
+    mkdir -p "$systemdsystemunitdir/multipathd.service.d"
+    inst_simple "$moddir/90-multipathd-remove-execstop.conf" \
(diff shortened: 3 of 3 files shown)


dustymabe (dismissed):
LGTM

tests/kola/misc-ro
```diff
@@ -198,3 +198,9 @@ if find ${tmpd}/usr/{bin,sbin,libexec} ! -perm -0111 | grep -v clevis-luks-commo
 fi
 rm -r ${tmpd}
 ok "All initrd scripts are executable"
+
+# Reminder to drop our temporary workaround (see https://github.com/coreos/fedora-coreos-config/pull/1233)
+if ! grep -q 'ExecStop=/sbin/multipathd shutdown' /usr/lib/dracut/modules.d/90multipath/multipathd.service; then
+    fatal "found dracut with fixed multipath service; 90-multipathd-remove-execstop.conf should be dropped now"
```
jlebon: Can we say

```suggestion
    fatal "found dracut with fixed multipath service; 90-multipathd-remove-execstop.conf should be moved to openshift/os"
```

instead? Otherwise, just removing it will regress RHCOS.

dustymabe: I guess the alternative is that we add it to fedora-coreos-config AND openshift/os now (in exclusive locations for each OS). That way we don't need the "move" step later.

lucab: Trying to cut the bikeshedding here, let me rewrite this test so that it only passes if fixed-dracut XOR overlay-has-quickfix.

jlebon: Yup, that works for me too!

dustymabe (approved):
LGTM

tests/kola/misc-ro
```diff
@@ -198,3 +198,16 @@ if find ${tmpd}/usr/{bin,sbin,libexec} ! -perm -0111 | grep -v clevis-luks-commo
 fi
 rm -r ${tmpd}
 ok "All initrd scripts are executable"
+
+# We need either a fixed dracut or temporary workaround, no need for both.
+# See https://github.com/coreos/fedora-coreos-tracker/issues/803.
+has_fixed_dracut=$(grep -q 'ExecStop=/sbin/multipathd shutdown' /usr/lib/dracut/modules.d/90multipath/multipathd.service; echo $?)
+has_overlay_quickfix=$(test ! -f /usr/lib/dracut/modules.d/35coreos-multipath/90-multipathd-remove-execstop.conf; echo $?)
```
jlebon, line 205: I may be wrong, but don't these have opposing semantics? `has_fixed_dracut` will be 0 if dracut is fixed, while `has_overlay_quickfix` will be 0 if the overlay quickfix isn't present.

Maybe let's switch to this format to make it easier to grok:

```bash
has_fixed_dracut=1
grep -q 'ExecStop=/sbin/multipathd shutdown' /usr/lib/dracut/modules.d/90multipath/multipathd.service || has_fixed_dracut=0
has_overlay_quickfix=1
test -f /usr/lib/dracut/modules.d/35coreos-multipath/90-multipathd-remove-execstop.conf || has_overlay_quickfix=0
```

?

lucab: We can wait for the CI to confirm, but I think the test logic is correct:
```
++ grep -q 'ExecStop=/sbin/multipathd shutdown' /usr/lib/dracut/modules.d/90multipath/multipathd.service
++ echo 0
+ has_fixed_dracut=0
```

jlebon: Right, yup. The multiple levels of semantic negation going on in these checks (when including the CLI semantics of 0/1 for true/false) makes this hard to reason about. But OK as is!

jlebon (approved):

dustymabe:
CI is kind of hung waiting for a pod to start on the testISO stuff. kola qemu tests already ran though and passed.

```
[2021-09-24T18:17:57.625Z]     --- PASS: non-exclusive-tests/ext.config.misc-ro (5.06s)
```
I'm waiting on this for our `next` promotion so I'm going to skip the rest of CI.

tests/kola/misc-ro
lucab: As it seems easier to read that way, I reworked this test logic as suggested in https://github.com/coreos/fedora-coreos-config/pull/1246.

PR #1234, opened 2021-09-24
[next-devel] overrides: fast-track frozen packages to avoid downgrades

jlebon:
We want to ship with the packages that would be in the stable repo if
updates weren't currently frozen. This avoids having downgrades during
the f34 -> f35 rebase.


jlebon:
That said, the following packages bumped in this PR genuinely are in the testing repo and have f34 stable packages with higher EVRs: conmon, dnsmasq, ethtool, libuv, skopeo.

dustymabe (approved):
LGTM

PR #1238, opened 2021-09-27
[next-devel] manifest: add iptables-legacy package

dustymabe:
In F35+ we need the `iptables-legacy` package to keep the status quo.
See https://github.com/coreos/fedora-coreos-tracker/issues/676#issuecomment-928028451

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -417,6 +417,9 @@
     "iproute-tc": {
       "evra": "5.13.0-2.fc35.x86_64"
     },
+    "iptables-legacy": {
+      "evra": "1.8.7-13.fc35.x86_64"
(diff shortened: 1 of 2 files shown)


jlebon (approved):
Nice, looks like that's all we need to preserve legacy as the default. Interestingly, I was testing this initially by just package layering `iptables-legacy`, but that still left nft as the default.

Can we add a simple subtest in misc-ro or a new shared test that just does `iptables --version | grep legacy`?

dustymabe:
> Can we add a simple subtest in misc-ro or a new shared test that just does `iptables --version | grep legacy`?

yep - will add as a followup since this is in the critical path of our `next` release and I don't want to incur the CI roundtrip penalty

PR #1239, opened 2021-09-27
[rawhide] overrides: pin kernel-5.15.0-0.rc2.18.fc36

saqibali-2k:
mdadm circular locking dependency has reappeard in rawhide. Let's pin kernel
to a working version for now.
https://github.com/coreos/fedora-coreos-tracker/issues/978


dustymabe (approved):
LGTM - thanks @saqibali-2k

PR #1240, opened 2021-09-27
[next] promote from `next-devel` at 6910ec1; add iptables-legacy

dustymabe:


dustymabe:
The promote from `next-devel` will give us newer `glib2` which [hit the repos late](https://pagure.io/releng/issue/10282#comment-753412).

jlebon (approved):

PR #1241, opened 2021-09-27
tests: add ext.config.firewall.iptables-legacy test

dustymabe:
Let's confirm for now that we're using iptables-legacy by default
until we switch to defaulting to iptables-nft.

See https://github.com/coreos/fedora-coreos-tracker/issues/676

### DIFF
--- a/tests/kola/firewall/iptables-legacy
+++ b/tests/kola/firewall/iptables-legacy
@@ -1,0 +1,20 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# Make sure we're still on legacy iptables for now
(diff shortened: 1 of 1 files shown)


tests/kola/firewall/iptables-legacy
```diff
@@ -0,0 +1,19 @@
...
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# Make sure we're still on legacy iptables for now
+# https://github.com/coreos/fedora-coreos-tracker/issues/676#issuecomment-928028451
+if ! iptables --version | grep legacy; then
```
jlebon, line 16: Minor/optional: let's rerun `iptables --version` here so that the test output has the string we tried to grep if it fails.

jlebon (dismissed):
Thanks!

tests/kola/firewall/iptables-legacy
dustymabe: done

PR #1246, opened 2021-09-28
multipathd.socket: add start conditions via dropin

lucab:
This adds a dropin for 'multipathd.socket' adding the same start
conditions that are present on the service unit. It is a temporary
workaround that can be removed once the packaged one is fixed.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=2008098


jlebon (approved):
LGTM, thanks a lot for these fixes! Will leave it open for a bit in case anyone else wants to do a review.

dustymabe:
just so I'm clear - this is independent of de62fb0 ?

jlebon:
> just so I'm clear - this is independent of [de62fb0](https://github.com/coreos/fedora-coreos-config/commit/de62fb029748440ff04476aa40882f9e4b98d271) ?

Yes. That commit fixes `multipathd.service` sometimes erroring out on switchroot because of `ExecStop=`. This PR fixes `multipathd.socket` going into a failed state when trying to access the socket on a machine where multipath is not configured.

dustymabe (approved):
LGTM

PR #1206, opened 2021-09-08
[next-devel] Move to Fedora 35

jlebon:
Prep for Fedora 35 beta release in two weeks.

PR #1247, opened 2021-09-28
[rawhide] manifest: empty out cracklib-dicts

jlebon:
We already ship a dropin to make libpwquality not use it. It became a
hard dep recently. We're hoping to revert that but for now, just empty
it out.

See thread at
https://github.com/coreos/fedora-coreos-config/pull/1206#discussion_r705425869.

Cherry-picked from part of 8ef41080bc2156f3f25ef9b0491539e5280f24ad.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -27,3 +27,9 @@ postprocess:
 packages:
   # resolved was broken out to its own package in rawhide/f35
   - systemd-resolved
+
+remove-from-packages:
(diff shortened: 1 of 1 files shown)


dustymabe (approved):

dustymabe:
rawhide CI is failing right now for a few different reasons. Will fix in a followup (which will actually be a cherry pick on top of this). skipping CI for now.

PR #1249, opened 2021-09-28
[rawhide] overrides: Unpin kernel 5.15.0-0.rc2.18.fc36

saqibali-2k:
Circular dep loop warnings seem to be gone now with the new kernel builds. Let's unpin
the kernel.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -14,21 +14,6 @@ packages:
     metadata:
       reason: 'https://github.com/coreos/fedora-coreos-tracker/issues/968'
       type: pin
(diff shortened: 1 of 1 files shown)


dustymabe (dismissed):
LGTM - thanks @saqibali-2k

dustymabe:
killed CI - it's going to fail anyway until https://github.com/coreos/fedora-coreos-config/pull/1248 and you'll have to rebase after that lands anyway because we're touching the same file.

dustymabe:
ready for rebase/merge now

dustymabe (approved):

PR #1251, opened 2021-09-30
WIP: overlay.d: abort booting if system has more than 1 'boot' partition

nikita-dubrovskii:
Inspired by: https://github.com/coreos/fedora-coreos-tracker/issues/976


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ensure-single-partition.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
lucab, line 1: From an out-of-band discussion, longer term it would be good to fold this logic into `rdcore` in order to avoid growing even more scripts in the initrd.

jlebon:
Hey Nikita, we're currently discussing in https://github.com/coreos/fedora-coreos-tracker/issues/976 how we want to tackle this issue overall. Can we hold this for a bit until we get consensus there?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ensure-single-partition.sh
nikita-dubrovskii: Have something here: https://github.com/nikita-dubrovskii/coreos-installer/commit/9a2332a2922095d92b839caa97c57b7682e84860

PR #1255, opened 2021-10-01
initramfs: ignore `udevadm settle` errors

jlebon:
We do `udevadm settle` in a few places to e.g. wait for symlinks to
update based on whatever operation we just did. But if `udevadm settle`
fails, we shouldn't fail the boot for it. It could be failing on some
completely unrelated thing (since it waits for all events).

Ideally in those scripts, we'd wait only for the specific events that we
care about. But even so, we should just opportunistically keep booting.
As a plus, even if we end up failing further down, we'll get a clearer
error.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2009662


dustymabe (approved):
LGTM. my only enhancement would be to try to convey that reasoning in a comment, but it's kind of hard to make the comment short and CI has already run. So probably just merge..

PR #1256, opened 2021-10-01
boot.mount: mount by UUID instead

jlebon:
This makes the mount more resistant to other filesystems labeled `boot`
that may get plugged in at any point.

Part of: coreos/fedora-coreos-tracker#976

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
@@ -31,5 +31,5 @@ rm -vrf ${initramfs_firstboot_network_dir}
 # append rootmap kargs to the BLS configs.
 root=$(karg root)
 if [ -z "${root}" ]; then
-    /usr/bin/rdcore rootmap /sysroot --boot-mount ${bootmnt}
+    rdcore rootmap /sysroot --boot-mount ${bootmnt}
 fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
@@ -33,3 +33,16 @@ root=$(karg root)
 if [ -z "${root}" ]; then
     rdcore rootmap /sysroot --boot-mount ${bootmnt}
 fi
+
+# And similarly, only inject boot= if it's not already present.
+boot=$(karg boot)
+if [ -z "${boot}" ]; then
+    # XXX: `rdcore rootmap --inject-boot-karg` or maybe `rdcore bootmap`
+    eval $(blkid -o export "${bootdev}")
+    if [ -z "${UUID}" ]; then
+        # This should never happen
+        echo "Boot filesystem ${bootdev} has no UUID" >&2
+        exit 1
+    fi
+    rdcore kargs --boot-mount ${bootmnt} --append boot=UUID=${UUID}
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
@@ -45,4 +45,7 @@ if [ -z "${boot}" ]; then
         exit 1
     fi
     rdcore kargs --boot-mount ${bootmnt} --append boot=UUID=${UUID}
+    # but also put it in /run for the first boot real root mount
+    mkdir -p /run/coreos
+    echo "${UUID}" > /run/coreos/bootfs_uuid
 fi
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -61,15 +61,47 @@ EOF
     add_wants "${unit_name}"
 }
 
+# Copied from
+# https://github.com/dracutdevs/dracut/blob/9491e599282d0d6bb12063eddbd192c0d2ce8acf/modules.d/99base/dracut-lib.sh#L586
+# rather than sourcing it.
+label_uuid_to_dev() {
+    local _dev
+    _dev="${1#block:}"
+    case "$_dev" in
+        LABEL=*)
+            echo "/dev/disk/by-label/$(echo "${_dev#LABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
+            ;;
+        PARTLABEL=*)
+            echo "/dev/disk/by-partlabel/$(echo "${_dev#PARTLABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
+            ;;
+        UUID=*)
+            echo "/dev/disk/by-uuid/$(echo "${_dev#UUID=}" | tr "[:upper:]" "[:lower:]")"
+            ;;
+        PARTUUID=*)
+            echo "/dev/disk/by-partuuid/$(echo "${_dev#PARTUUID=}" | tr "[:upper:]" "[:lower:]")"
+            ;;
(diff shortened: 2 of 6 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
```diff
@@ -33,3 +33,58 @@ root=$(karg root)
 if [ -z "${root}" ]; then
     /usr/bin/rdcore rootmap /sysroot --boot-mount ${bootmnt}
 fi
+
+# Generate the boot mount unit for the real root. Allow an Ignition config to
+# override it.
+mntfile=/etc/systemd/system/boot.mount
```
cgwalters: I think this should go in `/run` because it's a generated file.

Alternatively, we can keep the logic as a generator in the real root, and just pass the information for which device we found into the real root in `/run`, something like `ln -sr "${bootdev}" /run/coreos-bootdev`, and then our generator finds that?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -1,80 +0,0 @@
...
-    echo "skipping coreos-boot-mount-generator: ${UNIT_DIR} is not writable"
-    exit 0
-}
-
-# If there's already an /etc/fstab entries for /boot, then this is is a non-FCOS
-# system, likely RHCOS pre-4.3 (which still used Anaconda).  In that case, we
-# don't want to overwrite what the systemd-fstab-generator will do.
-if findmnt --fstab /boot &>/dev/null; then
-    exit 0
-fi
```
cgwalters: Are we sure we don't still need this logic?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
jlebon: > I think this should go in `/run` because it's a generated file.

It's generated, but unlike units generated from systemd generators, it's meant to be persistent. We don't need to rerender it on every boot.

> Alternatively, we can keep the logic as a generator in the real root, and just pass the information for which device we found into the real root in `/run`

It's funny, I initially did exactly that to keep the diff minimal (though using a stamp file). The (in retrospect obvious) problem is that the stamp file isn't persistent, so the generator won't be able to rely on it past the first boot. And it feels really awkward to have it live in `/etc` or `/var` when we could just as easily render it upfront.

It doesn't feel like a systemd generator is the right fit here because we're not really deriving units from a persistent source of truth config.

That said, upgrade tests are failing because it correctly points out that upgrading nodes will lose their `boot.mount` since we went from generating on every boot to generating only on first boot. So... ideally we'd get rid of the generator, but I think we need to keep it to not break updates. We'd need a barrier to move them over, but it doesn't seem worth the risk.

So some options are:
- still generate it from the initrd too; it'll take precedence over the generator version on new nodes (duplicate code)
- put the stamp file in `/var` or `/etc` (weird semantics)
- have the systemd generator write to `/etc` (really awkward)
- add `boot=UUID=...` karg and have the generator use it if available; this is closer to your original proposal in https://github.com/coreos/fedora-coreos-tracker/issues/976, though it would only be available from the second boot onward for the benefit of the real root. Then the generator would feel more akin to the `systemd-fstab-generator` (and how it reads the `root=` kargs). Yeah I think that's probably the cleanest approach.

I'll test out that last approach and then bring it back to https://github.com/coreos/fedora-coreos-tracker/issues/976.

jlebon:
Updated!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
```diff
@@ -33,3 +33,16 @@ root=$(karg root)
 if [ -z "${root}" ]; then
     rdcore rootmap /sysroot --boot-mount ${bootmnt}
 fi
+
+# And similarly, only inject boot= if it's not already present.
+boot=$(karg boot)
```
jlebon, line 38: E.g. in FIPS mode, we'll already have a `boot=` karg there. (We should change that code to use `boot=UUID` too.)

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -61,15 +61,50 @@ EOF
     add_wants "${unit_name}"
 }
 
+# Copied from
+# https://github.com/dracutdevs/dracut/blob/9491e599282d0d6bb12063eddbd192c0d2ce8acf/modules.d/99base/dracut-lib.sh#L586
+# rather than sourcing it.
+label_uuid_to_dev() {
```
cgwalters, line 67: Another thing we could put in rdcore in the future?

cgwalters (dismissed):
I didn't do a line-by-line analysis, but this looks great to me!

cgwalters:
I think this is also going to want some careful manual testing, definitely for things like upgrades etc.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
jlebon: Maybe. This runs in the real root, and so far `rdcore` has been about the initramfs only.  It'd be a bit awkward, though I guess we could just embed the full dracut path to it?

Though in this case, since we want to just implement the same API as dracut/fips, it seems simpler to just copy-paste the API implementation itself. If there are bugs in there, at least we'll have the same bugs as dracut. :)

bgilbert:
From a commit message:

> There is precedence for this in the FIPS module.

I think you mean that there's precedent in the FIPS module, and not that the FIPS module has precedence over coreos-boot-mount-generator?

bgilbert (dismissed):
LGTM!

jlebon:
> From a commit message:
> 
> > There is precedence for this in the FIPS module.
> 
> I think you mean that there's precedent in the FIPS module, and not that the FIPS module has precedence over coreos-boot-mount-generator?

Ahh yup, thanks!

I rebased this now and added a simple test. I forgot to do it in two force-pushes to make it easier to see the interdiff. But essentially the only changes are the additions in `tests/kola/root-reprovision/luks/test.sh` and fixing the precedent/precedence wording in the commit message.

jlebon:
Can I get another stamp on this?

tests/kola/root-reprovision/luks/test.sh
```diff
@@ -42,6 +49,9 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
       grep root=UUID= /proc/cmdline
       grep rd.luks.name= /proc/cmdline
       ok "found root kargs"
+
+      # while we're here, sanity-check that we have a boot=UUID karg too
+      grep boot=UUID= /proc/cmdline
```
bgilbert, line 54: Is it important that we're not doing `ok`/`fatal` here?

bgilbert (commented):
The new test is good as far as it goes, but it only tests the case where we're reprovisioning the rootfs.  Should we also have a test for the completely vanilla boot case?

jlebon:
> The new test is good as far as it goes, but it only tests the case where we're reprovisioning the rootfs. Should we also have a test for the completely vanilla boot case?

Sure, done! I initially didn't do this because we don't have a vanilla test which reboots and it felt funny to tack it onto another existing rebooting test which was testing some orthogonal thing.

But I think we should cover the vanilla case more explicitly. I added a new rebooting testcase for this. Would be interesting to somehow merge that with misc-ro and other non-exclusive tests though it'd require some thought.

bgilbert (approved):

jlebon:
> I think this is also going to want some careful manual testing, definitely for things like upgrades etc.

I added some tests here, but one test I did manually was boot up the very first stable release, bring it fully up to date and then update to this PR to confirm that the boot mount still works fine (i.e. correctly falls back to mounting by label). (Related: https://github.com/coreos/coreos-assembler/issues/2519).

PR #1257, opened 2021-10-04
[RFC] systemd-generators: perform ostree-remount early on

lucab:
**DO NOT MERGE**: this is a crude hack to experiment around performing ostree-remount
logic early on. If working, it can live directly within ostree.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-ostree-remount
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-ostree-remount
@@ -1,0 +1,31 @@
+#!/bin/bash
+export PATH="/usr/bin:/usr/sbin:${PATH}"
+set -euo pipefail
+
+. /usr/lib/coreos/generator-lib.sh
+
+/usr/lib/ostree/ostree-remount
+echo "done: ostree-remount"
+
+touch /run/ostree-sysroot-ro.stamp
+echo "done: /run/ostree-sysroot-ro.stamp"
+
+if test -f /run/ostree-live; then
+  echo "skip: /var on live loopback"
+  exit 0
+fi
+
+# NOTE(lucab): native libostree knows how to locate the booted stateroot.
+STATEROOT="fedora-coreos"
+# NOTE(lucab): native libostree knows how to locate the booted deployment.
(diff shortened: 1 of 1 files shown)


lucab:
For reference, I've been playing around with this and the following Butane fragments.

A service unit which keeps `/var` (and thus `/sysroot`) busy in RW mode, thus failing the remount:
```
systemd:
  units:
    - name: dev-ostree-race.service
      enabled: true
      contents: |
        [Unit]
        Description=Custom devtest unit
        DefaultDependencies=no
        Before=ostree-remount.service
        After=-.mount var.mount
        [Service]
        ExecStartPre=/usr/bin/rm -f /var/file-test
        ExecStartPre=/usr/bin/touch /var/file-test
        ExecStart=/usr/bin/bash -c 'while true; do sleep 5000; done > /var/file-test'
        [Install]
        WantedBy=local-fs.target ostree-remount.service
```

A mount unit with a ramdisk to mock both the "on-disk var" and the "external /var storage" cases:
```
systemd:
  units:
    - name: .var.mount
      # To activate: `sudo mv /etc/systemd/system/.var.mount /etc/systemd/system/var.mount`
      enabled: true
      contents: |
        [Unit]
        Description=Mount local ramdisk
        Before=local-fs.target
        [Mount]
        What=none
        Where=/var
        Type=tmpfs
        Options=size=2G
        [Install]
        WantedBy=local-fs.target

File: overlay.d/05core/usr/lib/systemd/system-generators/coreos-ostree-remount [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+. /usr/lib/coreos/generator-lib.sh
+
+/usr/lib/ostree/ostree-remount
+echo "done: ostree-remount"
+
+touch /run/ostree-sysroot-ro.stamp
+echo "done: /run/ostree-sysroot-ro.stamp"
+
+if test -f /run/ostree-live; then
```
lucab, line 13: This is needed for live ISO/PXE, but I find it quite problematic as the "ostree system on a live loopback FS" is a CoreOS-specific thing, and the flag file too.
However, I'm reasonably positive that this can be safely replaced with a "check whether /var is already a RW mounpoint" to make this work generically (in the real native implementation).

lucab:
We merged https://github.com/ostreedev/ostree/pull/2187 instead of this RFC, closing.

PR #1262, opened 2021-10-04
Switch to ostree-format: "oci"

cgwalters:
Part of https://github.com/coreos/fedora-coreos-tracker/issues/812

In this initial step, we're merely switching the internal
tarball to be a different format.

A future step will change the FCOS pipeline to automatically
push this container to quay.io.


cgwalters:
We

PR #1264, opened 2021-10-04
#975: bootup: ostree-remount may race with systemd-backlight and fail

smuda:
Adding a temporary fragment until ostreedev/ostree#2115 is complete.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/systemd-backlight@.service.d/45-after-ostree-remount.conf
+++ b/overlay.d/05core/usr/lib/systemd/system/systemd-backlight@.service.d/45-after-ostree-remount.conf
@@ -1,0 +1,4 @@
+# Temporary fix for https://github.com/coreos/fedora-coreos-tracker/issues/975
+# until https://github.com/ostreedev/ostree/issues/2115 is fixed.
+[Unit]
+After=ostree-remount.service


overlay.d/05core/usr/lib/systemd/system/systemd-backlight@.service.d/45-after-ostree-remount.conf
```diff
@@ -0,0 +1,2 @@
+[Unit]
```
jlebon, line 3: Can you add a comment at the top of this file referring to https://github.com/coreos/fedora-coreos-tracker/issues/975? Maybe something like:

```
# Temporary fix for https://github.com/coreos/fedora-coreos-tracker/issues/975
# until https://github.com/ostreedev/ostree/issues/2115 is fixed.
```

cgwalters:
In theory, it might be better to ship this in the fedora ostree RPM, as unfortunately right now the different ostree-using editions don't derive from FCOS, so this change won't help Fedora Silverblue etc.

Or we could put it in ostree upstream.

But, that doesn't block doing it here.

overlay.d/05core/usr/lib/systemd/system/systemd-backlight@.service.d/45-after-ostree-remount.conf
smuda: Sure!

smuda:
@cgwalters I haven't got the insights to judge what the best way forward is. If you folks tell me what to do, I'll do that. 😀

jlebon (dismissed):
LGTM! Could you squash them into a single commit?

Also good moving this in upstream or Fedora dist-git if folks prefer. Though hopefully we get it fixed properly in ostree soon anyway! :)

smuda:
Squashed! 😀

jlebon (approved):
LGTM, thanks a lot for your contribution! Will let @cgwalters or someone else review and approve in case there's a strong preference to move it higher up the stack.

lucab (approved):
Let's land this for now as a quickfix. Thanks for the patch!

smuda:
Thank you for letting me contribute!

dustymabe:
> Thank you for letting me contribute!

Thank you! Plenty of more opportunity if you stick around!

PR #1267, opened 2021-10-08
15fcos: change the location of Ignition report

### LINKED ISSUE #977: "Ignition previously ran" warning when reprovisioning using a non-CoreOS live system
**Describe the bug**

When reprovisioning a bare metal CoreOS instance with a persistent `/var` partition using a non-CoreOS live system (e.g. using `coreos-installer` from `cargo`), Ignition will issue a warning that it has run before.

**Reproduction steps**
Install CoreOS on `/dev/sda`. We now want to reprovision this existing instance.

1. Boot into some non-CoreOS live system (e.g. Debian, openSUSE, …) and make sure Rust and `cargo` are installed
2. `cargo install coreos-installer`
3. `coreos-installer --ignition-file /path/to/my/config.ign /dev/sda`
4. `reboot`

Likely the same happens with a container-based `coreos-installer`, even though I didn't test this particular case.

`coreos-installer` doesn't have to be involved at all. As an alternative one can also boot into a live system and write CoreOS' raw image to the disk while sparing the `/var` partition. This also completely wipes/overwrites the previous CoreOS instance, but CoreOS will still complain that Ignition has run before.

**Expected behavior**
No warnings.

**Actual behavior**
The following warning is shown:

```
Ignition: ran on 2021/09/27 14:43:31 UTC (this boot)
############################################################################
WARNING: Ignition previously ran on 2021/09/27 13:13:24 UTC. Unexpected
behavior may occur. Ignition is not designed to run more than once per system.
############################################################################
```

**System details**
 - Bare Metal
 - Fedora CoreOS 34.20210904.3.0

**Ignition config**
Butane config:

```yml
variant: fcos
version: 1.4.0
storage:
  disks:
    - device: /dev/sda
      wipe_table: false
      partitions:
        - number: 4
          label: root
          size_mib: 8192
          wipe_partition_entry: true
          resize: true
        - number: 5
          label: var
          size_mib: 0
  filesystems:
    - device: /dev/disk/by-partlabel/var
      format: ext4
      path: /var
      with_mount_unit: true
```

**Additional information**
Reported via IRC first.

A possible solution might be to add some sort of first boot marker to CoreOS' image (`/boot` or `/sysroot`) and to remove this marker on Ignition's first run. The warning appears if this marker doesn't exist when Ignition runs. Solely relying on the infos in `/var/lib/ignition` isn't sufficient.

sohankunkerkar:
Fixes coreos/fedora-coreos-tracker#977

This change helps to address the problem of displaying a warning about running Ignition twice while reprovisioning a bare metal
CoreOS instance with a persistent /var partition.

Needs: https://github.com/coreos/ignition/pull/1267

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-liveiso-autologin-generator
@@ -85,7 +85,7 @@ fi
 # If the user supplied an Ignition config, they have the ability to enable
 # autologin themselves.  Don't automatically render them insecure, since
 # they might be running in production and booting via e.g. IPMI.
-if jq -e .userConfigProvided /var/lib/ignition/result.json &>/dev/null; then
+if jq -e .userConfigProvided /etc/.ignition-result.json &>/dev/null; then
     exit 0
 fi
 
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ignition-config.service
@@ -3,8 +3,7 @@
 [Unit]
 Description=Check if Ignition config is provided
 Before=systemd-user-sessions.service
-RequiresMountsFor=/var/lib/ignition
-ConditionPathExists=/var/lib/ignition/result.json
+ConditionPathExists=/etc/.ignition-result.json
 
 [Service]
 Type=oneshot
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ignition-config
@@ -1,7 +1,7 @@
 #!/usr/bin/bash
 set -euo pipefail
 
-IGNITION_RESULT=/var/lib/ignition/result.json
+IGNITION_RESULT=/etc/.ignition-result.json
 
 WARN='\033[0;33m' # yellow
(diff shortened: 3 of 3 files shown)


dustymabe (dismissed):
trivial LGTM

bgilbert:
Let's s/fix/change/ in the commit message.  The code had the location correct, it just turns out that there's a reason not to put it there.

bgilbert:
`coreos-liveiso-autologin-generator` also needs an update.

jlebon (approved):
LGTM, though we do need to hold this until the new Ignition gets into FCOS, right?

Also, what do we want to do about machines that already have `/var/lib/ignition/result.json` written out? Should we detect those and automatically migrate them over?

sohankunkerkar:
> LGTM, though we do need to hold this until the new Ignition gets into FCOS, right?

Agreed!

> Also, what do we want to do about machines that already have `/var/lib/ignition/result.json` written out? Should we detect those and automatically migrate them over?
 
Yeah, it makes sense to me

bgilbert:
> Also, what do we want to do about machines that already have `/var/lib/ignition/result.json` written out? Should we detect those and automatically migrate them over?

We can, but I'm not sure it's worth doing.  The report is only informational, and we haven't been using the previous location for very long.  Most systems don't have a report on disk at all, and without migration we'd effectively be pretending that the `/var/lib/ignition` systems are in that group.

bgilbert (approved):

PR #1269, opened 2021-10-11
coreos-boot-edit: abort if system has several boot partitions

nikita-dubrovskii:
This requires https://github.com/coreos/coreos-installer/pull/658

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
@@ -14,6 +14,8 @@ karg() {
     echo "${value}"
 }
 
+rdcore verify-unique-fs-label --rereadpt boot
+
 # Mount /boot. Note that we mount /boot but we don't unmount it because we
 # are run in a systemd unit with MountFlags=slave so it is unmounted for us.
 bootmnt=/mnt/boot_partition
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-unique-boot.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-unique-boot.service
@@ -1,0 +1,22 @@
+[Unit]
+Description=Ensure filesystem labeled `boot` is unique
+ConditionPathExists=/etc/initrd-release
+DefaultDependencies=no
+Before=ignition-diskful.target
+Wants=systemd-udevd.service
+After=systemd-udevd.service
+# And since the boot device may be on multipath; optionally wait for it to
+# appear via the dynamic target.
+After=coreos-multipath-wait.target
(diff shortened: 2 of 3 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
```diff
@@ -14,6 +14,9 @@ karg() {
     echo "${value}"
 }
 
+# Ensure we have a unique filesystem labeled 'boot'
+/usr/bin/rdcore verify-unique-fs-label boot
```
bgilbert: For consistency with the rest of this file, let's not use an absolute path.

bgilbert (dismissed):
Looks good; just waiting for a coreos-installer release.

jlebon:
I think we'll also want this somewhere at the beginning of the initramfs too as per https://github.com/coreos/fedora-coreos-tracker/issues/976#issuecomment-953147348, but we can do that in a follow-up.

bgilbert (dismissed):

jlebon:
I think we still need the check in `coreos-boot-edit` (which is the one that will take the new `--rereadpt` switch from https://github.com/coreos/coreos-installer/pull/731). Or did you mean to do that in a separate PR?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-unique-boot.service
```diff
@@ -0,0 +1,24 @@
...
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/rdcore verify-unique-fs-label boot
+
+
```
jlebon: Extraneous blank lines

jlebon (dismissed):
LGTM! Need to hold until coreos-installer release with https://github.com/coreos/coreos-installer/pull/731.

Would be good to have tests that check the error conditions, though this would require some work in kola. How did you test this? A few checks I can think of not covered by CI:
- Verifying that `coreos-unique-boot.service` errors out if we boot with two bootable disks (or really just two disks each with a filesystem labeled `boot`).
- Verifying that `coreos-boot-edit.service` errors out. That one might be trickier to check. Hmm, actually it'd work to have an Ignition config which creates a filesystem labeled `/boot` on some secondary disk.

nikita-dubrovskii:
> Would be good to have tests that check the error conditions, though this would require some work in kola. How did you test this? 

I did it manually (on s390x):
 - attaching 2nd `boot` disk
 - with `multipath` 
 - `multipath` +  `boot` disk

jlebon:
We have new coreos-installer now. Restarted CI on this!

jlebon (approved):

bgilbert:
The `multipath.day1` kola test has started flaking; might be relevant to this PR: https://github.com/coreos/fedora-coreos-tracker/issues/1105

PR #1270, opened 2021-10-12
[WIP] dracut/ignition-ostree: avoid double /sysroot/var mounts

lucab:
This tweaks `ignition-ostree-mount-var.service` in order to avoid
mounting `/sysroot/var` (and unmounting it, on stop) if it is already
mounted.

Part of https://github.com/ostreedev/ostree/pull/2187

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-var.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-var.sh
@@ -40,13 +40,26 @@ do_mount() {
         fatal "${stateroot_var_path} is not a directory"
     fi
 
-    echo "Mounting $stateroot_var_path"
-    mount --bind "$stateroot_var_path" /sysroot/var
+    rm -f /tmp/ignition-ostree-mount-var-mounted
+    findmnt_exitcode="0"
+    findmnt /sysroot/var || findmnt_exitcode="1"
+    if [ "${findmnt_exitcode}" -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-var.sh
```diff
@@ -40,13 +40,26 @@ do_mount() {
         fatal "${stateroot_var_path} is not a directory"
     fi
 
-    echo "Mounting $stateroot_var_path"
-    mount --bind "$stateroot_var_path" /sysroot/var
+    rm -f /tmp/ignition-ostree-mount-var-mounted
```
cgwalters, line 43: I know this is in the initramfs where there shouldn't be any user code running, but let's avoid predictable filenames in `/tmp`.  Using `/run` is a safe bet.

cgwalters (approved):
Seems OK to me offhand.

Though honestly, I have gotten a bit lost in `/var` mount ownership.  In the FCOS (ignition+ostree) case we're always mounting `/var` in the initramfs with this unit, even if it's a subsequent boot? 

But after https://github.com/ostreedev/ostree/pull/2187 lands, ostree is doing more in the initramfs, which effectively conflicts with this?

Or in other words, is this just dead code after we have the new ostree?

lucab:
Yes, this conditional here is only required if we try to mount and leak /var, which possibly in the end we are not doing. In the best case we don't need this PR. We still need to keep this service unit in its current form (or something equivalent).

jlebon:
I think we can close this now given that we were able to avoid the leaking approach.

PR #1271, opened 2021-10-12
rework multipath fixes/tests

dustymabe:
This will enable us to apply the workarounds on specific fedora versions.

```
commit 3805d6943d3bca05ebe184cac1fd0af7708431ee
Date:   Tue Oct 12 10:03:13 2021 -0400

    move multipathd.service workaround from overlay to manifest postprocess
    
    This will allow us to conditionalize it based on the major Fedora
    version when the fix starts landing in Fedora Linux (starting with
    Rawhide). This also moves the workaround into a new dracut module
    (36coreos-multipath-fix) and out of the old module (35coreos-multipath)
    to make it easier to add from postprocess.

commit 46d06effd5718ef50c9040674e8822b102ee4a0f
Date:   Tue Oct 12 09:55:40 2021 -0400

    move multipathd.socket workaround from overlay to manifest postprocess
    
    This will allow us to conditionalize it based on the major Fedora
    version. The fix has currently landed in rawhide but not F35 and
    we want to continue to execute the check everywhere.

commit 537bb8937686f0fc646b7551544e351cb4fa0525
Date:   Tue Oct 12 09:36:21 2021 -0400

    tests: fix mulipathd.socket test
    
    I believe it should be testing for the existence of
    /usr/lib/systemd/system/multipathd.socket.d/50-start-conditions.conf.
    
    Fixup for 28d5409

```

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -127,6 +127,35 @@ postprocess:
     ConditionVirtualization=!container
     EOF
 
+  # Put in the fix for multipathd.service in dracut on releases that haven't
+  # been fixed yet.
+  # https://github.com/dracutdevs/dracut/pull/1606
+  # https://github.com/coreos/fedora-coreos-config/pull/1233
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    # This fix hasn't landed in rawhide (F36) yet, but hopefully will soon.
+    source /etc/os-release
+    [ ${VERSION_ID} -le 36 ] || exit 0
+    mkdir /usr/lib/dracut/modules.d/36coreos-multipath-fix
+    cat > /usr/lib/dracut/modules.d/36coreos-multipath-fix/90-multipathd-remove-execstop.conf <<'EOF'
+    # Temporary workaround for https://github.com/dracutdevs/dracut/pull/1606.
+    [Service]
+    ExecStop=
+    EOF
+    cat > /usr/lib/dracut/modules.d/36coreos-multipath-fix/module-setup.sh <<'EOF'
+    #!/bin/bash
+    # -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+    # ex: ts=8 sw=4 sts=4 et filetype=sh
+    install() {
+        # Temporary workaround for https://github.com/dracutdevs/dracut/pull/1606.
+        mkdir -p "$systemdsystemunitdir/multipathd.service.d"
+        inst_simple "$moddir/90-multipathd-remove-execstop.conf" \
+        "$systemdsystemunitdir/multipathd.service.d/90-multipathd-remove-execstop.conf"
+    }
+    EOF
+    chmod +x /usr/lib/dracut/modules.d/36coreos-multipath-fix/module-setup.sh
+
 # Packages listed here should be specific to Fedore CoreOS (as in not yet
 # available in RHCOS or not desired in RHCOS). All other packages should go
 # into one of the sub-manifests listed at the top.
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -107,6 +107,26 @@ postprocess:
       echo 'DEFAULT_HOSTNAME=localhost' >> /usr/lib/os-release
     fi
 
+  # Put in the fix for multipathd.socket on releases that haven't been fixed yet.
+  # https://bugzilla.redhat.com/show_bug.cgi?id=2008098
+  # https://github.com/coreos/fedora-coreos-config/pull/1246
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    # Only operate on releases before F36. The fix has landed in F36+ and there is no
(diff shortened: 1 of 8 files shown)


manifests/fedora-coreos-base.yaml
```diff
@@ -126,6 +126,35 @@ postprocess:
...
+  # Put in the fix for multipathd.service in dracut on releases that haven't
+  # been fixed yet.
+  # https://github.com/dracutdevs/dracut/pull/1606
+  # https://github.com/coreos/fedora-coreos-config/pull/1233
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    # This fix hasn't landed in rawhide (F36) yet, but hopefully will soon.
+    source /etc/os-release
+    [ ${VERSION_ID} -le 36 ] || exit 0
```
jlebon: How are we planning to track it landing so we can tweak this? Should we have it detect and error?

jlebon: Random: a less annoying approach we could take with these kinds of "remind us to do something when something happens" than erroring out would be to echo some string and then have code in the bump-lockfile job which detects this and e.g. opens a tracker issue when it sees the string.

Anyway, just a thought.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/module-setup.sh
```diff
@@ -12,11 +12,6 @@ install_ignition_unit() {
 }
 
 install() {
-    # Temporary workaround for https://github.com/dracutdevs/dracut/pull/1606.
-    mkdir -p "$systemdsystemunitdir/multipathd.service.d"
```
jlebon, line 16: Another approach for this is to have the drop-in keep living in 35coreos-multipath but renamed with a `.disabled` extension or so. Then here, we only `inst_simple` if the file isn't "disabled" and all the postprocess script would do is rename.

jlebon (dismissed):
One concrete suggestion, but LGTM overall.

This is another good example of why overlay dirs should be folded into the treefile. Then we could just have it be a separate overlay dir which gets included via `manifest.yaml` (or we could even add `releasever`-conditional `include`s).

dustymabe:
> (or we could even add `releasever`-conditional `include`s)

Yes Please! - is this RFE tracked anywhere? Also maybe base OS conditional includes (i.e. RHEL vs Fedora vs $other)

manifests/fedora-coreos-base.yaml
dustymabe: > How are we planning to track it landing so we can tweak this?

The ext kola test will complain when **both** fixes are in. When that happens we just adjust the number here.

> then have code in the bump-lockfile job 

In the case we're hitting now rawhide got a fix. `bump-lockfile` won't help us there because it doesn't run for our rawhide stream.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/module-setup.sh
dustymabe: yeah, i think I'd rather just put it in its own module. It already didn't really have much to do with `35coreos-multipath` (IIUC), just happened to be an easy place to put it without creating a new module.

dustymabe:
I think the biggest open question to me is how can we manage to make this change without the "fixes" getting yanked out of RHCOS. overlay.d will go into RHCOS, `fedora-coreos-base.yaml` won't so we need to account for that here.

manifests/fedora-coreos-base.yaml
jlebon: > `bump-lockfile` won't help us there because it doesn't run for our rawhide stream.

Right yup. Probably could just be a separate job that scans output logs from pipeline jobs or something. The main idea is that instead of breaking builds, we have CI file an issue.

dustymabe: agree - I'd like for that to exist.

jlebon:
> I think the biggest open question to me is how can we manage to make this change without the "fixes" getting yanked out of RHCOS. overlay.d will go into RHCOS, `fedora-coreos-base.yaml` won't so we need to account for that here.

Argh, right. I guess simplest would be to lower it to a shared manifest and also key off of `ID`? Yuck.

dustymabe:
ok added:

```
commit 7cecd785459c06248384b73c83554aa64ec7a1a9
Date:   Tue Oct 12 15:16:26 2021 -0400

    tests: break out multipath fix tests from misc-ro
    
    If they are their own separate tests we can symlink them in the
    openshift/os repo and re-use them there.

commit e485a7470fd224479d07d9e7a20859194bc6a93c
Date:   Tue Oct 12 14:54:44 2021 -0400

    add shared-workarounds.yaml manifest
    
    This will allow us to share postprocess workarounds with downstream
    distributions (i.e. RHCOS). In this initial run copy over the fixes
    for multipathd that were previously hoisted up in to fedora-coreos-base.yaml.

```

jlebon (approved):
Cool, thanks for untangling this! Will try to follow-up with some of the more long-term things we talked about to make this a bit less painful.

PR #1272, opened 2021-10-12
kola/var-mount: change to use by-partuuid to accomodate s390x

Prashanth684:
by-partlabel doesn't work for s390x because of https://bugzilla.redhat.com/show_bug.cgi?id=1899990.This addresses https://github.com/openshift/os/issues/648

### DIFF
--- a/tests/kola/var-mount/simple/config.bu
+++ b/tests/kola/var-mount/simple/config.bu
@@ -5,16 +5,18 @@ storage:
     - device: /dev/vda
       partitions:
         - label: var
+          guid: 63194b49-e4b7-43f9-9a8b-df0fd8279bb7
           size_mib: 1000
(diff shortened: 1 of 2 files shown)


jlebon (approved):
Thanks!

We still have the LUKS version of this test which exercises the `by-partlabel/` flow so we're not losing any coverage here.

julianwiedmann:
Hm, not entirely happy with this. The `gdisk` in fcos is actually new enough to have all the endianess fixes so that `by-partlabel` works on s390x. Hence imho the prefered solution would have been to bring those fixes into rhcos, and denylist this test until they land.

I'll send a patch so that we have `by-partlabel` coverage on s390x again. But let's also keep the `by-partuuid` test of course.

Issue #1228, opened 2021-09-22
tests: break up misc-ro and misc-ign-ro into individually scoped tests


dustymabe:
Now that we support combining non-exclusive tests into a single VM let's break up misc-ro and misc-ign-ro into their own test directories (specific focus/scope with better documentation for rationale) and mark then as `exclusive: false`.

dustymabe:
@mike-nguyen @aaradhak - did we do this one?

aaradhak:
> 

I dont think this is done yet.

dustymabe:
hmm - but those files don't exist any more in the repo. I think it was taken care of by f50f500 and 1e2e2a7

mike-nguyen:
Yes this is done.  Ash did the standardizing of the test formats after I did the break up.  I'll close this issue if there are no objections.

PR #1273, opened 2021-10-12
tests: check for FAT32 ESP

bgilbert:
For https://github.com/coreos/coreos-assembler/pull/2491.

### DIFF
--- a/tests/kola/storage/esp-fat32
+++ b/tests/kola/storage/esp-fat32
@@ -1,0 +1,20 @@
+#!/bin/bash
+# kola: { "exclusive": false, "architectures": "x86_64 aarch64" }
+# Check that the ESP is a FAT32 filesystem.
+# https://github.com/coreos/fedora-coreos-tracker/issues/993
+
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if ! file -sL /dev/disk/by-label/EFI-SYSTEM | grep "FAT (32 bit)"; then
+    fatal "EFI System Partition is not FAT32"
+fi
+ok "EFI System Partition is FAT32"


dustymabe:
The current goal is to move away from `misc-ro` towards smaller and more scoped tests (see https://github.com/coreos/fedora-coreos-config/issues/1228). The primary benefit of `misc-ro` (running tests in a single VM) is now handled additionally by marking a test as `exclusive: false`. Would you mind breaking it out to a separately defined test with `exclusive: false`? Otherwise we'll do it as part of #1228.

bgilbert:
Done!

tests/kola/storage/esp-fat32
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# kola: { "exclusive": false }
```
jlebon: Shouldn't this test only run on arches which ship an ESP?

bgilbert: Blah.  Yes.

bgilbert:
Updated.

jlebon (dismissed):

tests/kola/storage/esp-fat32
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# kola: { "exclusive": false, "architectures": "x86_64 aarch64" }
+# Check that the ESP is a FAT32 filesystem.
```
dustymabe, line 3:
```suggestion
# Check that the ESP is a FAT32 filesystem.
# https://github.com/coreos/fedora-coreos-tracker/issues/993
```

dustymabe (dismissed):
LGTM - one suggestion

dustymabe:
👍

dustymabe (approved):

dustymabe:
and it's working! CI tests failed with:

```
fatal 'EFI System Partition is not FAT32'
```

because https://github.com/coreos/coreos-assembler/pull/2491 hasn't landed yet.

bgilbert:
Closing per consensus in https://github.com/coreos/fedora-coreos-tracker/issues/993#issuecomment-943281115.

PR #1276, opened 2021-10-12
[testing] overrides: fast-track coreos-installer 0.10.1-1.fc34 for CVE-2021-20319

bgilbert:


dustymabe (approved):
LGTM

dustymabe:
We are submitting this change to `testing` so that our next promotion to `stable` will include this fix. We are not planning to do an ad-hoc `testing` release with the fix.

PR #1281, opened 2021-10-15
denylist: snooze rpmostree.install-uninstall for F35/F36 streams

dustymabe:
One of the remote yum repositories has an infra issue and is causing
the test to fail.

See https://github.com/coreos/fedora-coreos-tracker/issues/996


jlebon (approved):

dustymabe:
skipping CI on this one since this since `testing-devel` is unaffected.

PR #1284, opened 2021-10-15
denylist: snooze ostree.hotfix in rawhide on aarch64

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/942
We need someone to dig into why.


jlebon (approved):

dustymabe:
skipping CI on this one since this since testing-devel is unaffected.

PR #1285, opened 2021-10-18
tests/kola/var-mount: add back a simple by-partlabel test

julianwiedmann:
This is a follow-on to https://github.com/coreos/fedora-coreos-config/pull/1272, bringing back test coverage for by-partlabel on s390x. It actually works with the `gdisk` version that is available in Fedora.


julianwiedmann:
cc @Prashanth684

Prashanth684:
this will not work for rhcos because the gdisk fix hasn't made it there yet i believe. might want to exclude this test for rhcos and keep the partuuid test.

julianwiedmann:
> this will not work for rhcos because the gdisk fix hasn't made it there yet i believe. might want to exclude this test for rhcos and keep the partuuid test.

Yes, openshift/os will need to exclude this test. Happy to send that patch once this PR is merged.

tests/kola/var-mount/by-partlabel/test.sh
```diff
@@ -0,0 +1,44 @@
...
+
+src=$(findmnt -nvr /var/log -o SOURCE)
+[[ $(realpath "$src") == $(realpath /dev/disk/by-partlabel/varlog) ]]
+
+fstype=$(findmnt -nvr /var/log -o FSTYPE)
+[[ $fstype == ext4 ]]
+
+case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+  "")
+      ok "mounted on first boot"
```
lucab, line 34: A minimal "write a file and check that the content is still there after reboot" would be nice here.

jlebon:
Do we know if el8 will actually be patched soon-ish? (E.g. is there an outstanding RHBZ for rebasing gdisk?) If not, then we should just have this test be FCOS-only for now. Either we put it outside the `var-mount/` dir so it's not automatically inherited by RHCOS (via the symlink), or we can teach kola to parse e.g. `distros` from external tests JSON and set it to `fcos`. IMO denylist entries should be temporary.

But stepping back, I'm not sure all these variations on this var-mount test are that useful vs the added test time. Probably we can just have a check in `var-mount/simple/test.sh` that sanity-checks that the `by-label/` symlink is there and points to the right device (and we'd skip that check on RHCOS + s390x)? Alternatively, let's make it exclusive to s390x at least. On !s390x we already have coverage of `by-partlabel` in the LUKS variant.

bgilbert:
The RHBZ is the one listed in the commit message, [1899990](https://bugzilla.redhat.com/show_bug.cgi?id=1899990).  It seems to be targeted for 8.6.

jlebon:
> The RHBZ is the one listed in the commit message, [1899990](https://bugzilla.redhat.com/show_bug.cgi?id=1899990). It seems to be targeted for 8.6.

Ahh thanks, missed that!

I just realized that these tests also are closely related to the [`coreos.ignition.mount.*`](https://github.com/coreos/coreos-assembler/blob/ccf7d120e2cfd762ac6be44458882a7865f807f2/mantle/kola/tests/ignition/mount.go) tests. I think we should dedupe the coverage from these tests a bit and combine them. My suggestion is to have two tests only:
- `ext.config.var-mount.luks`: mounts `/var` from a LUKS device on a partition on the primary disk and mounts `/var/log` from a LUKS device on a partition on a secondary disk. `/var` LUKS device uses `by-partlabel`. `/var/log` LUKS device uses `by-partuuid`.
- `ext.config.var-mount.simple`: mounts `/var` from a partition on the primary disk and mounts `/var/log` from a partition on a secondary disk. `/var` filesystem uses `by-partlabel`. `/var/log` filesystem uses `by-partuuid`.

And then we delete the `coreos.ignition.mount.*` tests and denylist `ext.config.var-mount.simple` on RHCOS + s390x until we get that patch.

julianwiedmann:
> But stepping back, I'm not sure all these variations on this var-mount test are that useful vs the added test time. Probably we can just have a check in `var-mount/simple/test.sh` that sanity-checks that the `by-label/` symlink is there and points to the right device (and we'd skip that check on RHCOS + s390x)? Alternatively, let's make it exclusive to s390x at least. On !s390x we already have coverage of `by-partlabel` in the LUKS variant.

I wasn't aware that test time is a crucial factor, thanks. Having "unit tests" for the various options would be good imo, so maybe the below-mentioned `coreos.ignition.mount.*` is the way to go. Just realized that it [overrides](https://github.com/coreos/coreos-assembler/blob/ccf7d120e2cfd762ac6be44458882a7865f807f2/mantle/kola/tests/ignition/mount.go#L147) the `by-partlabel` paths for s390x.

tests/kola/var-mount/by-partlabel/test.sh
julianwiedmann: No objection to improving the test(s) - I merely copied what we did in the by-partuuid case.

julianwiedmann:
> I wasn't aware that test time is a crucial factor, thanks. Having "unit tests" for the various options would be good imo, so maybe the below-mentioned `coreos.ignition.mount.*` is the way to go. Just realized that it [overrides](https://github.com/coreos/coreos-assembler/blob/ccf7d120e2cfd762ac6be44458882a7865f807f2/mantle/kola/tests/ignition/mount.go#L147) the `by-partlabel` paths for s390x.

Here's the change to re-enable `by-partlabel` in `coreos.ignition.mount.*` instead: https://github.com/coreos/coreos-assembler/pull/2538

julianwiedmann:
The afore-mentioned cosa PR went in, closing this one.

PR #1294, opened 2021-10-19
[next-devel] overrides: fast-track frozen packages to avoid downgrades

sohankunkerkar:
related to https://github.com/coreos/fedora-coreos-streams/issues/383#issuecomment-946115687


dustymabe:
everything looks good except let's be a little more conservative with the kernel and go with `kernel-5.14.11-300.fc35` (the same version that is in f34). https://bodhi.fedoraproject.org/updates/FEDORA-2021-f377355e41

The update for kernel-5.14.11-300.fc35 was just created and hasn't had a lot of testing yet.

dustymabe (approved):
LGTM

PR #1297, opened 2021-10-19
Add Kubernetes e2e node tests

jlebon:
This patch adds new external tests which run the Kubernetes e2e node
tests. This will ensure that we don't inadvertedly break Kubernetes,
which is now itself tested on Fedora CoreOS.

The approach taken here is to model it as an external test. This makes
it easy for anyone to run it locally, which is important for
maintenance.

Note that these tests do *not* run by default: for now, I don't think we
should run these tests on every PR. They take a lot of resources and
time and it's not clear yet how flaky they'll be. Instead, we'll start
off by running them in the pipeline as a follow-on job. But as a next
step, I'd like to enable opt-in PR testing as well as testing as part of
the lockfile bumper.

One thing worth noting that makes it very different from a regular
external test is that the test binaries are built in the VM itself. This
is because (1) we may need newer golang than we have available in cosa,
and (2) cloning and building Kubernetes multiple times (per tested
release) is heavy and takes enough space that it's nicer to have it be
lifecycled to the VM itself, where it can be resource-constrained and
e.g. parallelized easily. For faster iteration, one can always just do
the same thing the test script does to build on the host and point the
k8s test harness at a pre-provisioned host.

Requires: https://github.com/coreos/coreos-assembler/pull/2505
Requires: https://github.com/kubernetes/kubernetes/pull/105764

### DIFF
--- a/tests/kola/k8s/node-e2e/config.bu
+++ b/tests/kola/k8s/node-e2e/config.bu
@@ -1,0 +1,14 @@
+variant: fcos
+version: 1.3.0
+ignition:
+  config:
+    merge:
+      # XXX: what this config does should probably be folded into the test
+      # directly instead so that the release-specific cri-o is used
+      - source: https://raw.githubusercontent.com/kubernetes/test-infra/master/jobs/e2e_node/crio/crio_cgrpv2.ign
+storage:
+  filesystems:
+    - device: /dev/disk/by-id/virtio-disk1
+      format: xfs
+      path: /var
+      with_mount_unit: true
--- a/tests/kola/k8s/node-e2e/data/node-e2e
+++ b/tests/kola/k8s/node-e2e/data/node-e2e
@@ -1,0 +1,39 @@
+#!/bin/bash
+set -xeuo pipefail
+
+SELF=$(realpath "$0")
+IMAGE=quay.io/projectquay/golang:1.17
+
+branch=$1; shift
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if [ -z "${container:-}" ]; then
+    mkdir -m 0600 ~/.ssh
+    ssh-keygen -N "" -t ed25519 -f /srv/kube
+    cat /srv/kube.pub >> ~/.ssh/authorized_keys
+    chmod 0600 ~/.ssh/*
+    exec podman run --net=host --rm -v /srv/kube:/srv/kube:z \
+        -v "${KOLA_EXT_DATA}:/srv/kola_ext_data:z" \
+        -v "${SELF}:/srv/self:z" "${IMAGE}" /srv/self "${branch}"
+    fatal "unreachable"
+fi
+
+# in container now
+dnf install -y rsync diffutils hostname
+git clone -b "${branch}" --depth 1 https://github.com/kubernetes/kubernetes
+make -C kubernetes test-e2e-node REMOTE=true REMOTE_MODE=ssh SSH_USER=root \
+    RUNTIME=remote CONTAINER_RUNTIME_ENDPOINT="unix:///var/run/crio/crio.sock" \
+    SSH_KEY=/srv/kube HOSTS=localhost FOCUS="\[NodeConformance\]" \
+    SSH_OPTIONS='-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no' \
+    TEST_ARGS='--kubelet-flags="--cgroup-driver=systemd --cgroups-per-qos=true
+                                --cgroup-root=/ --non-masquerade-cidr=0.0.0.0/0
+                                --runtime-cgroups=/system.slice/crio.service
+                                --kubelet-cgroups=/system.slice/kubelet.service"'
--- a/tests/kola/k8s/node-e2e/master
+++ b/tests/kola/k8s/node-e2e/master
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -xeuo pipefail
+# Just run on x86_64 for now. We'd need multi-arch images with go 1.17. We
+# could run on any platform, though need to replace `additionalDisks` by e.g.
+# `minDiskSpace` or something similar.
+# kola: {"minMemory": 4096, "timeoutMin": 45, "additionalDisks": ["15G"], "platforms": "qemu", "architectures": "x86_64", "requiredTag": "k8s", "tags": "needs-internet"}
+exec "${KOLA_EXT_DATA}/node-e2e" "${KOLA_TEST_EXE}"


jlebon:
I inlined patches needed for this that haven't yet merged upstream so that it's easier for anyone else to try this out if they'd like. But we'll hold merging until the patches are merged and can be dropped.

jlebon:
Cool! :tada: 

```
[2021-10-19T18:31:40.743Z] --- PASS: ext.config.k8s.node-e2e.master (2227.09s)
[2021-10-19T18:34:32.162Z] --- PASS: ext.config.k8s.node-e2e.release-1.22 (2341.54s)
```

dustymabe:
cc @harche

harche:
@jlebon This looks great. I was wondering how do you install crio on the node? Usually, in k8s CI we use [crio installer ](https://github.com/cri-o/cri-o/blob/main/scripts/node_e2e_installer) specifically written for node e2e tests.

tests/kola/k8s/node-e2e/data/node-e2e
```diff
@@ -0,0 +1,41 @@
...
+    chmod 0600 ~/.ssh/*
+    exec podman run --net=host --rm -v /srv/kube:/srv/kube:z \
+        -v "${KOLA_EXT_DATA}:/srv/kola_ext_data:z" \
+        -v "${SELF}:/srv/self:z" "${IMAGE}" /srv/self "${branch}"
+    fatal "unreachable"
+fi
+
+# in container now
+dnf install -y rsync diffutils hostname
+git clone -b "${branch}" --depth 1 https://github.com/kubernetes/kubernetes
```
cgwalters, line 31: Once our patches are upstream, in theory we can pull a container for this, right?

cgwalters (commented):
Thanks so much for working on this!

Things are going to be extra interesting when we e.g. integrate FCOS and base Fedora testing, and e.g. make it easy for the kernel maintainers to gate on this test.

jlebon:
> @jlebon This looks great. I was wondering how do you install crio on the node? Usually, in k8s CI we use [crio installer ](https://github.com/cri-o/cri-o/blob/main/scripts/node_e2e_installer?rgh-link-date=2021-10-20T05%3A53%3A01Z) specifically written for node e2e tests.

Thanks for looking @harche! For cri-o, we're using the same Ignition config as the upstream tests. See the `config.bu` file in this PR. There's a related comment there I'd like your feedback on actually. Ideally, for the 1.22 test we should be using the `node_e2e_installer` from the release-1.22 cri-o branch, right? I couldn't tell if the upstream tests did that too or not because AFAICT, these tests are not running on any release branches. (Is that right?)

tests/kola/k8s/node-e2e/data/node-e2e
jlebon: Yeah, I've toyed with this idea a lot. I think this is doable indeed, but for it to not be fragile and hacky it'd require some more work in the Kubernetes repo to allow building everything upfront, including parts of the test harness itself.

I guess there's also a tradeoff there in maintaining another container image. Ideally, that'd be maintained by upstream k8s itself.

harche:
> > @jlebon This looks great. I was wondering how do you install crio on the node? Usually, in k8s CI we use [crio installer ](https://github.com/cri-o/cri-o/blob/main/scripts/node_e2e_installer?rgh-link-date=2021-10-20T05%3A53%3A01Z) specifically written for node e2e tests.
> 
> Thanks for looking @harche! For cri-o, we're using the same Ignition config as the upstream tests. See the `config.bu` file in this PR. There's a related comment there I'd like your feedback on actually. Ideally, for the 1.22 test we should be using the `node_e2e_installer` from the release-1.22 cri-o branch, right? I couldn't tell if the upstream tests did that too or not because AFAICT, these tests are not running on any release branches. (Is that right?)

Right, in k8s CI we run node e2e only on master branch of k8s for now.

tests/kola/k8s/node-e2e/config.bu
```diff
@@ -0,0 +1,14 @@
+variant: fcos
+version: 1.3.0
+ignition:
+  config:
+    merge:
+      # XXX: what this config does should probably be folded into the test
+      # directly instead so that the release-specific cri-o is used
+      - source: https://raw.githubusercontent.com/kubernetes/test-infra/master/jobs/e2e_node/crio/crio_cgrpv2.ign
```
harche, line 8: We run node e2e only against master branch, but you are right if you are running the tests against specific k8s release then the relevant crio should be pulled from corresponding crio branch.

jlebon:
The prerequisite for https://github.com/kubernetes/kubernetes/pull/105764 was merged, but we're still waiting on that PR to merge.

tests/kola/k8s/node-e2e/config.bu
jlebon: Are there any plans to add these tests to other releases? At least 1.22 AFAICT works fine (using the master cri-o installer script, but I think we'd want to backport it to the 1.22 branch there to be correct).

jlebon:
OK this is ready to go now!

First force-push above is a rebase, second force-push are the final tweaks to drop the upstream patches that are now merged and adding a timeout of 45 minutes.

As per https://github.com/coreos/coreos-assembler/pull/2505, this won't run by default when invoking `kola run`; one has to use `--tag k8s`. As a result, CI won't test it here either. I'll look at turning it on in the pipeline first as a separate job, and then circle back to adding it to upstream CI, ideally as a separate context.

tests/kola/k8s/node-e2e/master
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -xeuo pipefail
+# Just run on x86_64 for now. We'd need multi-arch images with go 1.17. We
+# could run on any platform, though need to replace `additionalDisks` by e.g.
+# `minDiskSpace` or something similar.
+# kola: {"minMemory": 4096, "timeoutMin": 45, "additionalDisks": ["15G"], "platforms": "qemu", "architectures": "x86_64", "requiredTag": "k8s", "tags": "needs-internet"}
+exec "${KOLA_EXT_DATA}/node-e2e" "${KOLA_TEST_EXE}"
```
dustymabe, line 7: If I parse this correctly this will translate into:

```
exec tests/kola/k8s/node-e2e/data/node-e2e master
```

Is that correct?

tests/kola/k8s/node-e2e/data/node-e2e
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -xeuo pipefail
+
+SELF=$(realpath "$0")
```
dustymabe, line 4: the re-execing stuff seems a bit awkward. I'm sure you did this for good reason but I'll ask the question anyway: could we just put the `dnf install/git clone/make` commands in a separate script under `data/` and call that instead of re-execing?

dustymabe (commented):
Thanks for working on this. A few questions.

tests/kola/k8s/node-e2e/data/node-e2e
jlebon: The idea is that we can then support multiple branches by just symlinking a test named by the branch to `master`. I originally [had this for `release-1.22`](https://github.com/coreos/fedora-coreos-config/pull/1297#issuecomment-947031556) but as discussed above, 1.22 doesn't technically run these tests upstream, so it's not guaranteed to keep working. But going forward, it should be valid for `release-1.23`, etc...

dustymabe: Still a bit opaque to me. Maybe some comments/docs in the script would help (optional).

dustymabe (approved):

PR #1299, opened 2021-10-20
Bug 1990506: Add dracut module for scsi rules

ravanelli:

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
@@ -1,0 +1,13 @@
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# 1) Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1918244
+# On s390x systems with IBM 2810XIV discs multipath couldn't be configured
+# because SCSI_IDENT_* udev properties are not set at boot time
+# 2-) Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1990506
+# Missing symlinks to disk install
+install() {
+    inst_simple sg_inq
+    inst_rules 61-scsi-sg3_id.rules
+    inst_rules 63-scsi-sg3_symlink.rules
+}


overlay.d/05rhcos/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
jlebon: While we're here, let's drop this. These files are sourced not directly executed, so this comment has no effect.

jlebon (commented):
This is valid for FCOS and RHCOS, so let's put it in `overlay.d/05core/usr/lib/dracut/modules.d/01scsi-rules` instead?

overlay.d/05rhcos/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
ravanelli: You are right, that's fine, better change it ;)

jlebon: Looks like it's still here. Forgot a `git add` maybe? :)

jlebon:
WDYT about a commit message like:

> These rules will create the `/dev/disk/by-id/scsi-*` symlinks in the initramfs.

?

jlebon (approved):
LGTM!

To clarify for posterity, this is doing two things:
1. It's upstreaming [`01scsi-rules`](https://github.com/openshift/os/tree/master/overlay.d/05rhcos/usr/lib/dracut/modules.d/01scsi-rules) from RHCOS
2. It's adding `63-scsi-sg3_symlink.rules` to the list of rules for https://bugzilla.redhat.com/show_bug.cgi?id=1990506.

I've got a submodule bump going on in https://github.com/openshift/os/pull/658. Once this gets in, I can bump it again and drop the `01scsi-rules` in openshift/os.

overlay.d/05core/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
```diff
@@ -0,0 +1,13 @@
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# 1) Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1918244
+# On s390x systems with IBM 2810XIV discs multipath couldn't be configured
```
steffen-maier, line 5: This is by no means specific to any Linux architecture, especially not to s390x, so please remove that misleading part [I'm aware that my review is late].
It solely depends on the WWID designator type a storage uses [[https://www.t10.org/drafts.htm#SPC_Family](https://www.t10.org/drafts.htm#SPC_Family)].
FYI @julianwiedmann

ravanelli: @steffen-maier This piece of code came from: https://github.com/openshift/os/commit/632605a8ad0621368553534a32577c15167aa2d7
So I tried to not remove this comment.

PR #1303, opened 2021-10-21
.cci.jenkinsfile: use buildfetch instead of buildprep

dustymabe:
It's what all the cool kids are doing these days. See
https://github.com/coreos/coreos-assembler/pull/2495


jlebon (approved):
Nice! That's a good example of the improvement here.

PR #1306, opened 2021-10-25
tests/misc-ro: Look for unwanted SetUID/GID bits

travier:
tests/misc-ro: Fixup indentation

---

tests/misc-ro: Quoting variables (ShellCheck)

---

tests/misc-ro: Removing unused variable

Was added in b7e4622, and forgotten in aa16ec3.

---

tests/misc-ro: Look for unwanted SetUID/GID bits

Ensure that only known good files and directories have the SetUID/GID
bits sets.

Test for https://github.com/coreos/fedora-coreos-tracker/issues/1003


travier:
OK, so this is extra weird. This one should have failed. As it succeeded, it means that there is probably a subtle difference between the CI environment and the release environment that triggers this.

jlebon:
Was digging into this a bit, but can't reproduce locally as well trying various things. Thought maybe it was a privileged vs unprivileged thing. Funky CentOS CI umasks may be related too here (see https://github.com/coreos/coreos-ci-lib/commit/45762ba5aded40365a4777cfa462e76b737f60ad). That said, the pipeline also uses that wrapper now, so it should be consistent between CoreOS CI and the pipeline. Might also be some weird bug in the `kubernetes` Jenkins plugin.

One way to debug this further is to use your fcos-dev project in CentOS CI to bring up the pipeline and then manually poke around in the cosa build container to see where things are going wrong. If you don't have a `travier-fcos-dev` project, you can file an issue against https://pagure.io/centos-infra.

jlebon:
I think a brute-force fix for this would probably be

```diff
diff --git a/src/cmdlib.sh b/src/cmdlib.sh
index dc42d5b6a..8977bf1a6 100755
--- a/src/cmdlib.sh
+++ b/src/cmdlib.sh
@@ -273,6 +273,7 @@ commit_overlay() {
     # using ostree's --statoverride when dealing with executable files. See:
     # https://github.com/ostreedev/ostree/issues/2368
     rm -rf "${TMPDIR}/overlay" && (umask 0022 && cp -r "${respath}" "${TMPDIR}/overlay")
+    chmod -R gu-s "${TMPDIR}/overlay"
     # Apply statoverrides from a file in the root of the overlay, which may
     # or may not exist.  ostree doesn't support comments in statoverride
     # files, but we do.
```

I'm OK to do that, though ideally we'd figure out what's going on exactly first.

jlebon:
Let's see what https://github.com/coreos/fedora-coreos-pipeline/pull/420 says.

jlebon:
Ahhh got it. I think this is a result of running in the default SCC. The emptyDir we're composing on has its setgid set so we can actually write stuff there. We could request higher privs, but ideally we'd be insensitive to this, so it seems simpler to me to just strip it as in https://github.com/coreos/fedora-coreos-config/pull/1306#issuecomment-958142707.

OK, opened https://github.com/coreos/coreos-assembler/pull/2537.

travier:
Thanks @jlebon for figuring it out! Do you think we can merge this one now to have it checked for the releases?

jlebon (commented):
Nice cleanups and test! Given https://github.com/coreos/fedora-coreos-config/pull/1325, can we make it a separate non-exclusive test instead?

travier:
OK, I'll wait until #1325 is merged and will rebase on top.

tests/kola/files/setgid
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# kola: { "exclusive": false }
```
jlebon, line 2: We can restrict these new tests to `qemu-unpriv`, no?

jlebon (approved):
One comment, LGTM otherwise!

tests/kola/files/setgid
travier: Sure, but a lot of other tests would be in this case too. Any particular reason those should be `qemu-unpriv` only?

dustymabe: We haven't taken a hard-line approach, but we often will reason and say "this should be the same everywhere" and limit it to just our initial test run, which is qemu.

Though, now that non-exclusive tests are compressed into a single VM it probably doesn't hurt as much to just allow the test to run in all scenarios.

jlebon: Yup, agreed. I think non-exclusive tests change the calculus there.

jlebon: Related: https://github.com/coreos/fedora-coreos-config/pull/1340 (I was looking for a non-exclusive test which does make sense to run on clouds which then other non-exclusive tests might as well piggyback on "for free", even if their coverage is mostly redundant).

PR #1308, opened 2021-10-26
 overlay.d: add udev rule for creating stable symlink to boot disk

nikita-dubrovskii:
Creates '/dev/disk/coreos-root-disk' symlink to boot disk for
use by Ignition config, therefore 'storage.disks' section
could be defined without hardcoding /dev/sda, /dev/vda, etc.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/759

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# checks whether `disk` contains filesystem labeled `label`
+set -euo pipefail
+
+disk=$1
+label=$2
+
+# during execution of udev rules on disks 'lsblk' returns empty fields
+for pt in /sys/block/$disk/*; do
+    name=$(basename $pt)
+    if [[ "$name" =~ ${disk}p?[[:digit:]] ]] && [[ -e "/sys/block/$disk/$name/start" ]];
+    then
+        eval $(udevadm info --query=property -n /dev/$name | grep -e ID_FS_LABEL -e PARTNAME)
+        if [[ "${ID_FS_LABEL:-}" == "$label" ]] || [[ "${PARTNAME:-}" == "$label" ]]; then
+            exit 0
+        fi
+    fi
+done
+
+exit 1
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -211,6 +211,8 @@ main() {
     # clean it up so that no information from outside of the
     # real root is passed on to NetworkManager in the real root
     rm -rf /run/NetworkManager/
+
+    rm -rf /dev/disk/by-id/coreos-boot-disk
 }
 
 main
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -22,6 +22,7 @@ install() {
         diff \
         lsblk \
         sed \
+        grep \
         sgdisk
 
     inst_simple "$moddir/coreos-diskful-generator" \
@@ -29,6 +30,11 @@ install() {
 
     inst_script "$moddir/coreos-gpt-setup.sh" \
         "/usr/sbin/coreos-gpt-setup"
+
+    inst_simple "/usr/lib/udev/rules.d/80-coreos-boot-disk.rules"
+
+    inst_script "$moddir/coreos-disk-contains-fs.sh" \
+        "/usr/lib/udev/coreos-disk-contains-fs"
 
     inst_script "$moddir/coreos-ignition-setup-user.sh" \
         "/usr/sbin/coreos-ignition-setup-user"
--- a/overlay.d/05core/usr/lib/udev/rules.d/80-coreos-boot-disk.rules
+++ b/overlay.d/05core/usr/lib/udev/rules.d/80-coreos-boot-disk.rules
@@ -1,0 +1,10 @@
+# CoreOS-specific symlink for boot disk
+
+ACTION!="add|change", GOTO="stable_boot_end"
+SUBSYSTEM!="block", GOTO="stable_boot_end"
+
+ENV{DEVTYPE}=="disk" \
+  , PROGRAM=="coreos-disk-contains-fs $name boot" \
+  , SYMLINK+="disk/by-id/coreos-boot-disk"
+
+LABEL="stable_boot_end"
--- a/tests/kola/ignition/stable-boot/config.bu
+++ b/tests/kola/ignition/stable-boot/config.bu
@@ -1,0 +1,15 @@
+variant: fcos
+version: 1.3.0
+storage:
+  disks:
+  - device: /dev/disk/by-id/coreos-boot-disk
+    wipe_table: false
+    partitions:
+    - number: 5
+      size_mib: 1024
+      label: toor
+  filesystems:
+    - path: /var/lib/toor
+      device: /dev/disk/by-partlabel/toor
+      format: ext4
+      with_mount_unit: true
--- a/tests/kola/ignition/stable-boot/test.sh
+++ b/tests/kola/ignition/stable-boot/test.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# This test makes sure that ignition is able to use `coreos-boot-disk` symlink
+
+# We don't need to test this on every platform. If it passes in
+# one place it will pass everywhere.
+# kola: { "platforms": "qemu-unpriv" }
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# symlink shouldn't be propogated to real-root
+link="/dev/disk/by-id/coreos-boot-disk"
+if [[ -h "${link}" ]]; then
+    fatal "${link} still exists"
+fi
+
+# sanity-check that the root disk has all required partitions
+findmnt -nvr -o SOURCE /boot
+findmnt -nvr -o SOURCE /sysroot
+toor=$(findmnt -nvr -o SOURCE /var/lib/toor)
+if [[ ! "$toor" =~ ^/dev/[a-z0-9]+p?5$ ]]; then
+    fatal "${toor} is not 5th partition"
+fi


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-root-symlink.service
```diff
@@ -0,0 +1,25 @@
...
+Wants=systemd-udevd.service
+After=systemd-udevd.service
+
+Requires=dev-disk-by\x2dlabel-root.device
+After=dev-disk-by\x2dlabel-root.device
+# And since the boot device may be on multipath; optionally wait for it to
+# appear via the dynamic target.
+After=coreos-multipath-wait.target
+After=coreos-gpt-setup.service
+Before=coreos-ignition-setup-user.service ignition-disks.service
```
bgilbert: Does this need to be before `coreos-ignition-setup-user.service`?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-root-symlink.service
```diff
@@ -0,0 +1,25 @@
+[Unit]
+Description=Create symlink to boot disk
+ConditionPathExists=/etc/initrd-release
+DefaultDependencies=no
+Before=local-fs-pre.target systemd-fsck-root.service
+Before=ignition-diskful.target
```
bgilbert: I'm not sure how relevant these are.  The main thing is to run before `ignition-disks.service`, isn't it?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/create-root-symlink.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# force kernel to update info about disks
+for dev in $(lsblk -o NAME --noheadings --nodeps --list --paths); do
+    partx -u ${dev}
+done
+
+udevadm trigger
+udevadm settle
+
+LINK=/dev/disk/by-label/coreos-root-disk
```
bgilbert: Bikeshed: This isn't a partition, and the whole-disk device doesn't have a label, so `/dev/disk/by-label` doesn't really fit.  It's tempting to put it directly in `/dev/disk`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/create-root-symlink.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -euo pipefail
+
+# force kernel to update info about disks
+for dev in $(lsblk -o NAME --noheadings --nodeps --list --paths); do
+    partx -u ${dev}
+done
+
+udevadm trigger
+udevadm settle
```
bgilbert: What problem is this code solving?  Is there a more targeted way of doing it than retriggering every device?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -22,6 +22,10 @@ install() {
         diff \
         lsblk \
         sed \
+        mkdir \
```
bgilbert: The `99base` module installs `mkdir`, so we can skip this.

bgilbert (commented):
In general, this looks great.  Some questions about dependencies, and I'm not sure about the partx/trigger/settle part.

bgilbert:
Also, I agree with https://github.com/coreos/coreos-assembler/pull/2517#issuecomment-953956949 that the kola test should move into this repo.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/create-root-symlink.sh
nikita-dubrovskii: for some reason on `s390x`  `WWN` is not always present in `lsblk` output, so this `udev/partx` code is used to mitigate that problem

nikita-dubrovskii:
> Also, I agree with [coreos/coreos-assembler#2517 (comment)](https://github.com/coreos/coreos-assembler/pull/2517#issuecomment-953956949) that the kola test should move into this repo.

done.

nikita-dubrovskii:
https://github.com/coreos/ignition/pull/1278

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/create-root-symlink.sh
```diff
@@ -0,0 +1,32 @@
...
+mapfile -t PARTITIONS < <(lsblk -o WWN,LABEL,NAME,PKNAME --pairs --noheadings | grep 'LABEL="root"' | sed 's/WWN=\"\" //' | sort -u -k1,2)
+
+LENGTH=${#PARTITIONS[@]}
+if [[ ${LENGTH} -eq 1 ]]; then
+    eval ${PARTITIONS[0]}
+    echo "Creating symlink '${LINK}' to '${PKNAME}' (contains 'root' partition '${NAME}')"
+    mkdir -p /run/udev/rules.d/
+    echo "KERNEL==\"${PKNAME}\", SUBSYSTEM==\"block\", SYMLINK+=\"${LINK}\"" > /run/udev/rules.d/99-stable-root.rules
+    udevadm control --reload-rules
+    udev_trigger
```
bgilbert: In this case, we should only need to retrigger the boot disk, rather than the whole system.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/create-root-symlink.sh
```diff
@@ -0,0 +1,32 @@
...
+# Symlink is not propogated to the real root (removed during coreos-teardown-initramfs)
+
+set -euo pipefail
+
+udev_trigger() {
+    udevadm trigger
+    udevadm settle
+}
+
+# force kernel to update info about disks
```
bgilbert: Let's say the particular problem we're trying to solve, for future reference.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/create-root-symlink.sh
```diff
@@ -0,0 +1,32 @@
...
+for dev in $(lsblk -o NAME --noheadings --nodeps --list --paths); do
+    partx -u ${dev} || true
+done
+udev_trigger
```
bgilbert: We can limit this to storage devices, at least.  Even then, I wonder whether retriggering every storage device in the system might take a lot of time on large systems.

overlay.d/05core/usr/lib/udev/rules.d/99-zz-coreos-stable-root.rules
```diff
@@ -0,0 +1,19 @@
+# CoreOS-specific symlink for disk holdind filesystem labeled `root`
```
jlebon:
```suggestion
# CoreOS-specific symlink for disk holding filesystem labeled `root`
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -30,6 +32,11 @@ install() {
     inst_script "$moddir/coreos-gpt-setup.sh" \
         "/usr/sbin/coreos-gpt-setup"
 
+    inst_simple "/usr/lib/udev/rules.d/99-zz-coreos-stable-root.rules"
+
+    inst_script "$moddir/count-fs.sh" \
+        "/usr/bin/count-fs"
```
jlebon: Let's prefix this executable name by `coreos-`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/count-fs.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
jlebon: `set -euo pipefail`

overlay.d/05core/usr/lib/udev/rules.d/99-zz-coreos-stable-root.rules
```diff
@@ -0,0 +1,19 @@
...
+
+ACTION!="add|change", GOTO="stable_root_end"
+SUBSYSTEM!="block", GOTO="stable_root_end"
+TEST!="/usr/bin/count-fs", GOTO="stable_root_end"
+
+# First run on disks and check how many filesystems labeled `root` we have
+ENV{DEVTYPE}=="disk" \
+  , PROGRAM=="/usr/bin/count-fs root", RESULT=="1" \
```
jlebon: Hmm, shouldn't this be passing in the block device we're triggering on?
Then `count-fs` would only check for `root` filesystems on that device.

Otherwise IIUC, this could symlink whatever disk happened to be scanned last instead.

tests/kola/ignition/stable-root/test.sh
```diff
@@ -0,0 +1,24 @@
...
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
```
jlebon: Can this script verify that the Ignition config did what we expected? E.g. starting from `findmnt -nvr -o SRC /sysroot`, sanity-check that the root disk has 5 partitions.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/count-fs.sh
```diff
@@ -0,0 +1,28 @@
...
+                    id=$ID_FS_UUID_ENC
+                fi
+                if [[ -z $id ]]; then
+                    id=$(/usr/lib/udev/scsi-id -ugd /dev/$name || echo $pt)
+                fi
+                ids+=("$id")
+            fi
+        fi
+    done
+done
```
jlebon: Hmm, this is tricky code. Let's flesh out the discussions below first and then circle back to this.

overlay.d/05core/usr/lib/udev/rules.d/99-zz-coreos-stable-root.rules
```diff
@@ -0,0 +1,11 @@
+# CoreOS-specific symlink for disk holding filesystem labeled `root`
+
+ACTION!="add|change", GOTO="stable_root_end"
+SUBSYSTEM!="block", GOTO="stable_root_end"
+TEST!="/usr/bin/coreos-disk-contains-fs", GOTO="stable_root_end"
```
jlebon: Is this check necessary? Seems unnecessary to support having the rule without the corresponding script.

overlay.d/05core/usr/lib/udev/rules.d/99-zz-coreos-stable-root.rules
```diff
@@ -0,0 +1,11 @@
+# CoreOS-specific symlink for disk holding filesystem labeled `root`
```
jlebon: Re. the filename, does it need to be `99-zz`? I think the big ones worth running after are the `60-persistent-*` rules, so maybe e.g. 80 would suffice?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -0,0 +1,19 @@
...
+# checks whether `disk` contains filesystem labeled `label`
+set -euo pipefail
+
+disk=$1
+label=$2
+
+# during execution of udev rules on disks 'lsblk' returns empty fields
+for pt in /sys/block/$disk/*; do
+    name=$(basename $pt)
+    if [[ "$name" =~ "$disk" ]]; then
```
jlebon: Was curious how libblk did this, and found https://github.com/util-linux/util-linux/blob/177c9ec7c7aa0f49413f4cb5f347edf721d5a5ba/lib/sysfs.c#L203-L243, which isn't far from what you have here.

How about something like this to tighten this up:

```suggestion
    if [[ "$name" =~ ${disk}p?[[:digit:]] ]] && [ -e "/sys/block/$disk/$name/start" ]; then
```

?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -0,0 +1,19 @@
...
+        eval $(udevadm info --query=property -n /dev/$name | grep -e PARTNAME)
```
jlebon: Shouldn't this be `ID_FS_LABEL`? Otherwise we're comparing partition names to filesystem names.

overlay.d/05core/usr/lib/udev/rules.d/99-zz-coreos-stable-root.rules
jlebon: Also, maybe instead of `coreos-stable-root.rules`, how about `coreos-root-disk` to reflect the name of the symlink we actually create?

tests/kola/ignition/stable-root/test.sh
```diff
@@ -0,0 +1,32 @@
...
+# kola: { "platforms": "qemu-unpriv" }
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
```
jlebon: Check out the other tests which use `commonlib.sh` to get these functions.

jlebon (commented):
Can we use `boot` instead of `root`? It's mostly equivalent because all we care about for this is pre-`ignition-disks` in which case boot and root are on the same disk. But I like `boot` because that's what #1269 targets, so we know we'll only have a single boot filesystem.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
nikita-dubrovskii: `ID_FS_LABEL` is not always available.

tests/kola/ignition/stable-root/test.sh
nikita-dubrovskii: other `ignition` tests use copy-paste )

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -211,6 +211,8 @@ main() {
     # clean it up so that no information from outside of the
     # real root is passed on to NetworkManager in the real root
     rm -rf /run/NetworkManager/
+
+    rm -rf /dev/disk/by-id/coreos-root-disk
```
jlebon: Can we also change the symlink to `coreos-boot-disk`?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -30,6 +31,11 @@ install() {
     inst_script "$moddir/coreos-gpt-setup.sh" \
         "/usr/sbin/coreos-gpt-setup"
 
+    inst_simple "/usr/lib/udev/rules.d/80-coreos-root-disk.rules"
```
jlebon: And the name of this rules file.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -30,6 +31,11 @@ install() {
...
+
+    inst_script "$moddir/coreos-disk-contains-fs.sh" \
+        "/usr/bin/coreos-disk-contains-fs"
```
jlebon: Let's put this in `/usr/lib/udev` to be clear it's for udev only? As a bonus, the rule can then just call `coreos-disk-contains-fs` without leading dirs.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -0,0 +1,20 @@
...
+disk=$1
+label=$2
+
+# during execution of udev rules on disks 'lsblk' returns empty fields
+for pt in /sys/block/$disk/*; do
+    name=$(basename $pt)
+    if [[ "$name" =~ ${disk}p?[[:digit:]] ]] && [[ -e "/sys/block/$disk/$name/start" ]];
+    then
+        eval $(udevadm info --query=property -n /dev/$name | grep -e ID_FS_LABEL -e PARTNAME)
+        if [[ "${ID_FS_LABEL:-}" == "$label" ]] || [[ "${PARTNAME:-}" == "$label" ]]; then
```
jlebon, line 14: > `ID_FS_LABEL` is not always available.

Ahh OK. Hmm, that's unfortunate. Using the partition name rather than the filesystem label is not as good because we have expectations around the latter but not necessarily the former (e.g. I don't think anything would actually break if we have multiple partitions labeled `boot`). OTOH, it's not unreasonable to extend our requirement to partition names. So... let's roll with it!

jlebon (dismissed):

jlebon (approved):
:tada:

PR #1310, opened 2021-10-28
[testing-devel] overrides: Fast-track dracut-055-6.fc34

saqibali-2k:

PR #1311, opened 2021-10-29
Revert "manifests/fedora-coreos-base: explicitly add xen-blkfront to initrd

dustymabe:
This reverts commit 11b9f0bd1bc715972b3a70f6b2d506270788af41.

The fix has landed in all active releases of Fedora now.


dustymabe:
and... if I happen to be wrong we now have a pipeline test that will run tests on m4 AWS instances and prove me wrong.

jlebon (approved):
LGTM. Would've been nicer to make it part of https://github.com/coreos/fedora-coreos-config/pull/1310 so it's done atomically, but meh. :)

saqibali-2k (approved):

PR #1314, opened 2021-10-29
[next-devel] overrides: fast-track NetworkManager-1.32.12-2.fc35

dustymabe:
This fixes hostname fallback to `localhost` in F35+. See
https://github.com/coreos/fedora-coreos-tracker/issues/1005


jlebon (approved):
:tada:

PR #1321, opened 2021-11-02
switch default networking config to ip=auto

dustymabe:
```
commit 0411655054315530aa6587a0fdea182db08b76eb
Date:   Thu Nov 4 14:46:15 2021 -0400

    coreos-teardown-initramfs: pick up default net cfg from afterburn dropin
    
    This way we can only define it in one place.

commit ba49263d64b0834b904c6ec8ea59c6e3a66c47b7
Date:   Tue Nov 2 14:51:36 2021 -0400

    switch default networking config to ip=auto
    
    Upstream NM changed the behavior of `ip=dhcp,dhcp6` to also wait
    20 seconds for DHCPv6. We don't want to require all users by default
    to wait for DHCPv6 especially when it's doubtful most users even have
    a ipv6 DHCP server. We can stick with our current behavior by utilizing
    `ip=auto` instead.
    
    Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1000

commit d8555698bc3398fb017d0a098651bec909dca4e3
Date:   Tue Nov 2 14:31:40 2021 -0400

    tests: add test for default networking configuration
    
    Since we depend so much on the default networking configurations let's
    alert ourselves when any default networking configuration changes in
    NetworkManager. This allows us to react and adjust to the changes
    (if needed) instead of finding out later that problems were introduced.
    some context in: https://github.com/coreos/fedora-coreos-tracker/issues/1000
```

### DIFF
--- a/image-base.yaml
+++ b/image-base.yaml
@@ -11,7 +11,7 @@ extra-kargs:
     - mitigations=auto,nosmt
 
 # Disable networking by default on firstboot. We can drop this once cosa stops
-# defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
+# defaulting to `ip=auto rd.neednet=1` when it doesn't see this key.
 ignition-network-kcmdline: []
 
 # Optional remote by which to prefix the deployed OSTree ref
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -30,6 +30,9 @@ dracut_func() {
 # If it matches then it was the default, if not then the user provided
 # something extra.
 are_default_NM_configs() {
+    # pick up our CoreOS default networking kargs from the afterburn dropin
+    DEFAULT_KARGS_FILE=/usr/lib/systemd/system/afterburn-network-kargs.service.d/50-afterburn-network-kargs-default.conf
+    source <(grep -o 'AFTERBURN_NETWORK_KARGS_DEFAULT=.*' $DEFAULT_KARGS_FILE)
     # Make two dirs for storing files to use in the comparison
     mkdir -p /run/coreos-teardown-initramfs/connections-compare-{1,2}
     # Make another that's just a throwaway for the initrd-data-dir
@@ -40,7 +43,7 @@ are_default_NM_configs() {
     # Do a new run with the default input
     /usr/libexec/nm-initrd-generator \
         -c /run/coreos-teardown-initramfs/connections-compare-2 \
-        -i /run/coreos-teardown-initramfs/initrd-data-dir -- ip=auto
+        -i /run/coreos-teardown-initramfs/initrd-data-dir -- $AFTERBURN_NETWORK_KARGS_DEFAULT
     # remove unique identifiers from the files (so our diff can work)
     sed -i '/^uuid=/d' /run/coreos-teardown-initramfs/connections-compare-{1,2}/*
     # currently the output will differ based on whether rd.neednet=1
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -40,7 +40,7 @@ are_default_NM_configs() {
     # Do a new run with the default input
     /usr/libexec/nm-initrd-generator \
         -c /run/coreos-teardown-initramfs/connections-compare-2 \
-        -i /run/coreos-teardown-initramfs/initrd-data-dir -- ip=dhcp,dhcp6
+        -i /run/coreos-teardown-initramfs/initrd-data-dir -- ip=auto
     # remove unique identifiers from the files (so our diff can work)
     sed -i '/^uuid=/d' /run/coreos-teardown-initramfs/connections-compare-{1,2}/*
     # currently the output will differ based on whether rd.neednet=1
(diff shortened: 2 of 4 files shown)


miabbott:
Is this going to be usable by RHCOS?  We're gonna be stuck with RHEL 8.4 NM for a bit, so I'm concerned about the next time we decide to bump the f-c-c submodule in `openshift/os`

dustymabe:
> Is this going to be usable by RHCOS? We're gonna be stuck with RHEL 8.4 NM for a bit, so I'm concerned about the next time we decide to bump the f-c-c submodule in `openshift/os`

The test I added covers RHCOS too. We'll know any time the networking configuration changes for either FCOS or RHCOS and will be able to determine if it's the right change for us.

That's a long way of me saying, yes, I believe this is safe for RHCOS too. At least I tested on 4.9+, are we bumping f-c-c submodule for any others?

tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -0,0 +1,168 @@
...
+normalize_connection_file() {
+    sed -e s/^uuid=.*$/uuid=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/ \
+        -e s/^timestamp=.*$/timestamp=xxxxxxxxxx/                 \
+        -e s/^interface-name=.*$/interface-name=xxxx/             \
+        $1
+}
+
+source /etc/os-release
+# All current FCOS releases use the same config
+if [ "$NAME" == "Fedora Linux" -o "$NAME" == "Fedora" ]; then
```
jlebon: Can't we do

```suggestion
if [ "$ID" == "Fedora" ]; then
```

?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -40,7 +40,7 @@ are_default_NM_configs() {
     # Do a new run with the default input
     /usr/libexec/nm-initrd-generator \
         -c /run/coreos-teardown-initramfs/connections-compare-2 \
-        -i /run/coreos-teardown-initramfs/initrd-data-dir -- ip=dhcp,dhcp6
+        -i /run/coreos-teardown-initramfs/initrd-data-dir -- ip=auto
```
jlebon: Not new, though we should look at having the Afterburn drop-in be sourced in this service too, so then we can just use `AFTERBURN_KARGS_DEFAULT` here instead of duplicating.

image-base.yaml
```diff
@@ -11,7 +11,7 @@ extra-kargs:
     - mitigations=auto,nosmt
 
 # Disable networking by default on firstboot. We can drop this once cosa stops
-# defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
+# defaulting to `ip=auto rd.neednet=1` when it doesn't see this key.
 ignition-network-kcmdline: []
```
jlebon, line 15: We can drop this now. I can do that in a follow-up.

tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -0,0 +1,168 @@
...
+    if [ "$VERSION_ID" -eq "34" ]; then
+        EXPECTED_INITRD_NETWORK_CFG=$EXPECTED_INITRD_NETWORK_CFG2
+        EXPECTED_REALROOT_NETWORK_CFG=$EXPECTED_REALROOT_NETWORK_CFG1
+    elif [ "$VERSION_ID" -ge "35" ]; then
+        EXPECTED_INITRD_NETWORK_CFG=$EXPECTED_INITRD_NETWORK_CFG1
+        EXPECTED_REALROOT_NETWORK_CFG=$EXPECTED_REALROOT_NETWORK_CFG1
+    else
+        fatal "fail: not operating on expected OS version"
+    fi
+elif [ "$NAME" == "Red Hat Enterprise Linux CoreOS" ]; then
```
jlebon:
```suggestion
elif [ "$ID" == "rhcos" ]; then 
```

tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -0,0 +1,168 @@
...
+    fatal "fail: not operating on expected OS"
+fi
+
+
+# Execute nm-initrd-generator against our default kargs (defined by
+# afterburn drop in) to get the generated initrd network config.
+DEFAULT_KARGS_FILE=/usr/lib/dracut/modules.d/35coreos-network/50-afterburn-network-kargs-default.conf 
+DEFAULT_KARGS=$(grep AFTERBURN_NETWORK_KARGS_DEFAULT $DEFAULT_KARGS_FILE | \
+                sed s/Environment=AFTERBURN_NETWORK_KARGS_DEFAULT=// |
+                tr -d "'" | tr -d '"')
```
jlebon: Ahh too bad. If this weren't a non-exclusive test, the service name would be predictable, so we could have a Butane config that copies the dropin to the external test.

Hmm, here's another approach:

```suggestion
source <(grep -o 'AFTERBURN_NETWORK_KARGS_DEFAULT=.*' $DEFAULT_KARGS_FILE)
```

dustymabe: TIL about `grep -o`

tests/kola/networking/default-network-behavior-change/test.sh
jlebon: Sorry, that should be `fedora`, not `Fedora`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
dustymabe: added another commit with this

image-base.yaml
dustymabe: Will let you do in a followup with a nice commit message.

tests/kola/networking/default-network-behavior-change/test.sh
dustymabe: done

dustymabe:
OK. I think I've addressed all code review comments.

jlebon (approved):

PR #1322, opened 2021-11-02
tests: add distro tags to FCOS only tests

miabbott:
This is a first pass at categorizing the existing external tests using
the `distros` tag to indicate that the tests are only supported on
FCOS.  This should allow better sharing of tests with RHCOS.

Note: some of these tests may be able to be adapted for RHCOS; will
investigate in a follow-up PR.

See: openshift/os#661
See: coreos/coreos-assembler#2521

### DIFF
--- a/tests/kola/README.md
+++ b/tests/kola/README.md
@@ -1,0 +1,17 @@
+# Kola External Tests
+
+For more information about the `kola` external tests, see the [Integrating External Test Suites](https://coreos.github.io/coreos-assembler/kola/external-tests/#integrating-external-test-suites)
+topic in the `coreos-assembler` documentation.
+
+## Sharing tests between FCOS/RHCOS
+
+These tests are intended to be shared between FCOS and RHCOS. If you are adding
+a new test(s) to this collection, ensure that the test also run on RHCOS.
+Otherwise, use the `kola` JSON header to specify that it only runs on FCOS like
+so:
+
+```bash
+#!/bin/bash
+# kola: {"distros": "fcos"}
+...
+```
--- a/tests/kola/containers/cgroups-v2
+++ b/tests/kola/containers/cgroups-v2
@@ -1,5 +1,8 @@
 #!/bin/bash
-# kola: { "exclusive": false }
+# This test only runs on FCOS because RHCOS does not currently support cgroupsv2
+# by default.
+# TODO-RHCOS: drop "fcos" tag when cgroupsv2 lands in RHCOS
+# kola: { "distros": "fcos", "exclusive": false }
 set -xeuo pipefail
 
 ok() {
--- a/tests/kola/files/rpmdb-sqlite
+++ b/tests/kola/files/rpmdb-sqlite
@@ -1,5 +1,7 @@
 #!/bin/bash
-# kola: { "exclusive": false }
+# This test only runs on FCOS as the RHCOS rpmdb has not migrated to sqlite yet.
+# TODO-RHCOS: drop the "fcos" tag when RHCOS migrates to using sqlite
+# kola: { "distros": "fcos", "exclusive": false }
 set -xeuo pipefail
 
 ok() {
--- a/tests/kola/firewall/iptables-legacy
+++ b/tests/kola/firewall/iptables-legacy
@@ -1,5 +1,9 @@
 #!/bin/bash
-# kola: { "exclusive": false }
+# This test is currently scoped to only FCOS because the RHCOS version of `iptables`
+# is using the `nf_tables` backend.
+# TODO: modify this test to check for `nf_tables` backend when FCOS switches.
+#       See https://github.com/coreos/fedora-coreos-config/pull/1324
+# kola: { "distros": "fcos", "exclusive": false }
 set -xeuo pipefail
 
 ok() {
--- a/tests/kola/ignition/systemd-unmasking/test.sh
+++ b/tests/kola/ignition/systemd-unmasking/test.sh
@@ -7,7 +7,10 @@ set -xeuo pipefail
 
 # We don't need to test this on every platform. If it passes in
 # one place it will pass everywhere.
-# kola: { "platforms": "qemu-unpriv" }
+# This test is currently scoped to FCOS because `dnsmasq` is not masked
+# on RHCOS.
+# TODO-RHCOS: determine if any services on RHCOS are masked and adapt test
+# kola: { "distros": "fcos", "platforms": "qemu-unpriv" }
 
 ok() {
     echo "ok" "$@"
--- a/tests/kola/networking/default-network-behavior-change/test.sh
+++ b/tests/kola/networking/default-network-behavior-change/test.sh
@@ -136,7 +136,7 @@ fi
 
 # Execute nm-initrd-generator against our default kargs (defined by
 # afterburn drop in) to get the generated initrd network config.
-DEFAULT_KARGS_FILE=/usr/lib/dracut/modules.d/35coreos-network/50-afterburn-network-kargs-default.conf 
+DEFAULT_KARGS_FILE=/usr/lib/dracut/modules.d/35coreos-network/50-afterburn-network-kargs-default.conf
 source <(grep -o 'AFTERBURN_NETWORK_KARGS_DEFAULT=.*' $DEFAULT_KARGS_FILE)
 tmpdir=$(mktemp -d)
 /usr/libexec/nm-initrd-generator \
--- a/tests/kola/networking/stub-resolve.sh
+++ b/tests/kola/networking/stub-resolve.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
-# kola: { "exclusive": false }
+# This test only runs on FCOS because `systemd-resolved` is not installed on
+# RHCOS
+# kola: { "distros": "fcos", "exclusive": false }
 set -xeuo pipefail
 
 ok() {
--- a/tests/kola/root-reprovision/luks/test.sh
+++ b/tests/kola/root-reprovision/luks/test.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "minMemory": 4096, "architectures": "!s390x"}
+# This test is only running on FCOS because it is checking that /boot is
+# mounted by UUID and RHCOS is not doing that yet.
+# TODO-RHCOS: drop 'fcos' tag when RHCOS is mounting /boot by UUID
+#             See: https://github.com/openshift/os/issues/675
+# kola: { "distros": "fcos", "platforms": "qemu", "minMemory": 4096, "architectures": "!s390x" }
 set -xeuo pipefail
 
 ok() {
--- a/tests/kola/toolbox/test.sh
+++ b/tests/kola/toolbox/test.sh
@@ -13,7 +13,11 @@
 
 # Only run on QEMU to reduce CI costs as nothing is platform specific here.
 # Toolbox container is currently available only for x86_64 and aarch64 in Fedora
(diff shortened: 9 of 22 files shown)


HuijingHei:
Looks good to me

tests/kola/README.md
```diff
@@ -0,0 +1,19 @@
+# Kola External Tests
+
+For more information about the `kola` external tests, see the following topic
+in the `coreos-assembler` documentation:
+
+<https://coreos.github.io/coreos-assembler/kola/external-tests/#integrating-external-test-suites>
```
dustymabe:
```suggestion
For more information about the `kola` external tests, see the [Integrating External Test Suites](https://coreos.github.io/coreos-assembler/kola/external-tests/#integrating-external-test-suites) topic in the `coreos-assembler` documentation.
```

tests/kola/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Sharing tests between FCOS/RHCOS
+
+These tests are intended to be shared between FCOS and RHCOS. If you are adding
+a new test(s) to this collection, ensure that the tests also run on RHCOS.
```
dustymabe:
```suggestion
a new test(s) to this collection, ensure that the test also run on RHCOS.
```

tests/kola/ignition/systemd-unmasking/test.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# kola: {"distros": "fcos"}
```
dustymabe: hmm. there is another `# kola:` statement in this file. Are we supposed to combine them into one?

dustymabe (commented):
Thanks for doing this work @miabbott.

A few suggestions.. Mostly LGTM, but I'd like to advocate for adding reasons why tests only apply to FCOS even if it's blatantly obvious. I really do feel that stating the obvious is a good thing sometimes especially when new people stumble onto the code and might not know the history.

miabbott:
> Thanks for doing this work @miabbott.
> 
> A few suggestions.. Mostly LGTM, but I'd like to advocate for adding reasons why tests only apply to FCOS even if it's blatantly obvious. I really do feel that stating the obvious is a good thing sometimes especially when new people stumble onto the code and might not know the history.

Great suggestion, Dusty!  I'll work that into the next revision.

tests/kola/ignition/systemd-unmasking/test.sh
miabbott: oops!  got sloppy here.

saqibali-2k (dismissed):
I second Dusty's idea of adding a short description of why is test is limited to FCOS. Besides that changes look good overall!

tests/kola/toolbox/test.sh
```diff
@@ -13,7 +13,8 @@
 
 # Only run on QEMU to reduce CI costs as nothing is platform specific here.
 # Toolbox container is currently available only for x86_64 and aarch64 in Fedora
-# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv", "architectures": "x86_64 aarch64" }
+# This test only runs on FCOS because RHCOS is missing the `machinectl` command.
```
dustymabe, line 16: I think this might be able to be improved. While technically accurate that RHCOS might be missing the `machinectl` command I think there are other reasons that we would only want this test to run on FCOS. i.e. I imagine the default container they pull should be different and the test we want to validate is probably different. So we'd either need to adapt the test to work on both or just state that this test is targeted at FCOS and we should create a RHCOS specific toolbox test in the future.

dustymabe (dismissed):
LGTM - I just had one suggestion for improvement, but that's optional.

tests/kola/toolbox/test.sh
cgwalters: Side note: remember soon "RHCOS" will be ambiguous.  It'd be more accurate to say "RHCOS8" for example, because perhaps we do choose to pull in `machinectl` in RHCOS9.

This is actually another reason why longer term these distro tags are going to be less sustainable than doing e.g. feature detection.  IOW skipping the test if `machinectl` doesn't exist.

jlebon:
Are we planning to hold this until https://github.com/coreos/fedora-coreos-config/pull/1325? Otherwise I think this will need an additional pass for all the tests split out of misc-ro.

miabbott:
Rebased after #1325 landed, but now I see #1337, so might be another rebase in my future.

I tweaked three tests as separate commits to make them work seamlessly across FCOS/RHCOS. (@dustymabe interested in any feedback re: the networking test I changed)

There are other tests in there that could be modified for the same outcome, but require a little more investigation.

miabbott:
Rebased again on top of #1337.

Added additional comments (grep for `TODO-RHCOS`) to indicate which tests may be able to be adapted for RHCOS.

tests/kola/files/rpmdb-sqlite
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
-# kola: { "exclusive": false }
+# This test only runs on FCOS as the RHCOS rpmdb is split across multiple files
+# and cannot be queried with a single file.
+# TODO-RHCOS: determine how to test for sqlite RPM db on RHCOS
```
dustymabe: I think this is probably just because RHCOS hasn't picked up that change yet: https://fedoraproject.org/wiki/Changes/Sqlite_Rpmdb

tests/kola/firewall/iptables-legacy
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
-# kola: { "exclusive": false }
+# This test is currently scoped to only FCOS because the RHCOS version of `iptables`
+# is using the `nf_tables` backend.
+# TODO: modify this test to check for `nf_tables` backend when FCOS switches
```
dustymabe: When https://github.com/coreos/fedora-coreos-config/pull/1324 merges I think it will handle nftables and legacy iptables. we can link to that here in the TODO probably.

tests/kola/networking/stub-resolve.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
-# kola: { "exclusive": false }
+# This test only runs on FCOS because the `stub-resolv.conf` file is not present
```
dustymabe:
```suggestion
# This test only runs on FCOS because `systemd-resolved` is not installed on
# RHCOS
```

I would just use the same reasoning as the previous test..

tests/kola/root-reprovision/luks/test.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "minMemory": 4096, "architectures": "!s390x"}
+# This test is only running on FCOS because it is checking that /boot is
+# mounted by UUID and RHCOS is not doing that yet.
+# TODO-RHCOS: drop 'fcos' tag when RHCOS is mounting /boot by UUID
```
dustymabe, line 4: Probably doesn't exist, but if there were a link someone could follow here that would give a status of "when RHCOS is mounting /boot by UUID" it would make it more likely to get implemented in the future.

tests/kola/ignition/systemd-unmasking/test.sh
```diff
@@ -7,7 +7,10 @@ set -xeuo pipefail
 
 # We don't need to test this on every platform. If it passes in
 # one place it will pass everywhere.
-# kola: { "platforms": "qemu-unpriv" }
+# This test is currently scoped to FCOS only because the `dnsmasq` service on
+# RHCOS is disabled.
+# TODO-RHCOS: determine if any services on RHCOS are masked and adapt test
```
dustymabe:
```suggestion
# This test is currently scoped to FCOS because `dnsmasq` is not masked
# on RHCOS.
# TODO-RHCOS: determine if any services on RHCOS are masked and adapt test
```

tests/kola/toolbox/test.sh
dustymabe: I think the current practice is probably most efficient. Basically what we have been doing typically is using feature detection if it's easy. Sometimes this is easily done by querying a program other times not. Sometimes it can also be done by querying a version of a particular software (i.e. `systemd` or `NetworkManager`). Other times it's not so easy and kind of complex. In those cases I think we should basically switch off `RHCOS/FCOS` and then do "version detection" (i.e. F34, F35, RHEL 8.X, RHEL 9.X) within the embedded logic.

tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -65,6 +65,7 @@ mac-address-blacklist=
 dhcp-timeout=90
 dns-search=
 method=auto
+required-timeout=20000
```
dustymabe: I need to dig into this more before we merge this to understand this change. Perhaps a new version of NM got brought into RHCOS (which is exactly what this test was designed to do: highlight changes in the default configuration). If so maybe I should make a separate PR with this change to further document the change.

dustymabe:
Thank you so much for working on this @miabbott!

tests/kola/networking/default-network-behavior-change/test.sh
miabbott: On RHCOS 4.10, I see the following changes to the NetworkManager RPM:

```
$ ./pkg_report.py --release rhcos-4.10 --package NetworkManager
410.84.202109202033-0 = NetworkManager-1.30.0-10.el8_4.x86_64 on 2021-09-20 20:33:00
410.84.202111030412-0 = NetworkManager-1.30.0-13.el8_4.x86_64 on 2021-11-03 04:12:00
```

miabbott:
```
* Thu Oct 14 2021 Ana Cabral <acabral@redhat.com> - 1:1.30.0-13 
- Fixes nm-cloud-setup.service interference in network connectivity (rh#2013208)  

* Thu Oct 07 2021 Fernando Fernandez Mancera <ferferna@redhat.com> 
- 1:1.30.0-12 - initrd: set an IPv4 required-timeout of 20 seconds, to prefer IPv4 over IPv6 for default connections (rh #1961666)  

* Tue Oct 05 2021 Fernando Fernandez Mancera <ferferna@redhat.com> 
- 1:1.30.0-11 - platform: preserve IPv6 multicast route added by kernel (rh #2007264) 
- cloud-setup: don't interfere with other subnets (rh #2007341)  

* Mon Jul 12 2021 Beniamino Galvani <bgalvani@redhat.com> 
- 1:1.30.0-10 - core: prefer IPv4 to determine the hostname (rh #1970335)
```

Looks like that change in the config file came from https://bugzilla.redhat.com/show_bug.cgi?id=1961666 -> https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/907/

Ignition/RHCOS is all over the BZ, so I think this is expected.

dustymabe: You're right. This change is expected. Added more context in a commit over in https://github.com/coreos/fedora-coreos-config/pull/1339

miabbott:
Rebased one more time and dropped the commit to change the network test that was covered by #1339

tests/kola/containers/cgroups-v2
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
-# kola: { "exclusive": false }
+# This test only runs on FCOS because RHCOS does not currently support cgroupsv2
+# by defaulr.
```
dustymabe:
```suggestion
# by default.
```

dustymabe (dismissed):
LGTM - one spelling fix suggestion

dustymabe:
Hurry up and merge before anything else lands!

dustymabe (approved):

PR #1324, opened 2021-11-03
overlay.d: add 35coreos-iptables

jlebon:
From `overlay.d/README.md`:

Contains systemd service and script for remaining on iptables-nft after
the migration to nft.

Split out because (1) it will roll out to next first, and (2) it can
more easily be deleted after the barrier release.

For more details, see:
https://github.com/coreos/fedora-coreos-tracker/issues/676
https://github.com/coreos/fedora-coreos-config/pull/1324


jlebon:
Need to add tests for this. But also, we can't merge this until we actually officially start the deprecation window and wait the 3 months.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,53 @@
...
+        exit 0
+    fi
+done
+
+symlink_is_default() {
+    local symlink=$1; shift
+    if [ ! -L "/sysroot/$symlink" ] || [ ! -L "/sysroot/usr/$symlink" ]; then
+        return 1
+    fi
+    if [ "$(readlink "/sysroot/$symlink")" != "$(readlink "/sysroot/usr/$symlink")" ]; then
```
dustymabe: this always makes my head hurt.. do the nested double quotes `""` work  because of the parentheses? I usually try to not mix double or single quotes, but I think you have to here.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,53 @@
...
+        echo "Executable $path no longer present; exiting."
+        exit 0
+    fi
+done
+
+symlink_is_default() {
+    local symlink=$1; shift
+    if [ ! -L "/sysroot/$symlink" ] || [ ! -L "/sysroot/usr/$symlink" ]; then
+        return 1
+    fi
```
dustymabe: what case is this handling? Can we add a comment?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
dustymabe: also would be nice to add a comment about what it's checking just for the obvious

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+# If there are any modifications to the symlinks, do nothing. This is basically
+# like `ostree admin config-diff` but more focused and lighter/safer than doing
+# a bwrap call and grepping output.
+for symlink in $SYMLINKS; do
+    symlink=/etc/alternatives/$symlink
+    if ! symlink_is_default "$symlink"; then
+        echo "Symlink $symlink is not default; exiting."
```
dustymabe: i.e. `Symlink is not default` which means the user must have modified it so we're not going to touch it?

If true, them maybe let's add `refusing to overwrite user modifications` or similar?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+set -euo pipefail
+
+SYMLINKS="ip6tables ip6tables-restore ip6tables-save iptables iptables-restore iptables-save"
+LEGACY_TARGETS="ip6tables-legacy ip6tables-legacy-restore ip6tables-legacy-save iptables-legacy iptables-legacy-restore iptables-legacy-save"
+
+# sanity-check the stamp file is present
+if [ ! -e /sysroot/etc/fedora-coreos/iptables-legacy.stamp ]; then
```
dustymabe: Nice.. We're starting to use the `/etc/fedora-coreos` folder for things now. I really think we should create a docs page for everything we've ever put in this folder (i..e will be useful in the future) that explains or links to reasonings.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,53 @@
...
+    symlink=/etc/alternatives/$symlink
+    if ! symlink_is_default "$symlink"; then
+        echo "Symlink $symlink is not default; exiting."
+        exit 0
+    fi
+done
+
+# Do it!
+# shellcheck disable=SC2086
+set -- $LEGACY_TARGETS
```
dustymabe: I wonder if an associative array would be more elegant here.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,53 @@
...
+# a bwrap call and grepping output.
+for symlink in $SYMLINKS; do
+    symlink=/etc/alternatives/$symlink
+    if ! symlink_is_default "$symlink"; then
+        echo "Symlink $symlink is not default; exiting."
+        exit 0
+    fi
+done
+
+# Do it!
```
dustymabe: optional

```suggestion
# Perform the enablement of iptables-legacy
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=CoreOS Enable iptables-legacy
+ConditionPathExists=/etc/initrd-release
+DefaultDependencies=false
+ConditionPathExists=/sysroot/etc/fedora-coreos/iptables-legacy.stamp
```
dustymabe: should we limit this unit to running only once? It only needs to run once, right?

dustymabe (commented):
Thanks for working on this @jlebon. A few comments/questions:

dustymabe:
as part of this PR let's also update the `iptables-legacy` test (it should now create the stamp file via Ignition) and add a new test for `iptables-nft`. The `iptables-nft` test should be able to be run on both FCOS and RHCOS.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
jlebon: Sounds good to me! Let's try to find someone to tackle this bit?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
jlebon: Yup, double quotes like this is OK. If we use single quotes we won't get command/variable substitution.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
jlebon: Added a comment!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
jlebon: I like associative arrays, but their uncommonness makes me hesitate to use them in general unless it's really necessary (and usually points to scripts that probably shouldn't be bash).

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.service
jlebon: Hmm, good question. I think it would be less confusing if it's "bound" to the stamp file that already exists instead of creating another one. One case I can imagine is someone upgrading to an iptables-nft-default release, then deploying back an iptables-legacy-default release and then upgrading again. If we ran once only, we wouldn't reapply the legacy backend, but we should.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
dustymabe: It's just that I don't feel like what is being done here is very common (or maybe I'm just not recognizing it for some reason) either. Iterating through one loop and cycling through positional args at the same time inside the loop, it took me a moment to catch on to what you were doing.

I'm making the argument that declaring an associative array is at least explicit even if not as many people are familiar with them. They also have a familiar notation/intuitiveness of other languages in a `hash[key]=value` kind of way.

In the very least maybe we can get a comment in here to make it more explicit about what's being done with the positional parameters.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.service
dustymabe: Kind of torn on this one. I like the use of a systemd unit like this for "migration" or "run once" type stuff but not as much for an every boot kind of thing. For example, what if someone decides in the future they want to switch to using NFT and they log into their systems and fix the symlinks themselves (not realizing `/sysroot/etc/fedora-coreos/iptables-legacy.stamp`) exists. After a reboot they're back to legacy iptables.

This is a contrived example, but what I'm trying to argue for is not having 2nd order configuration (i.e. configuration that dynamically controls other configuration) that runs every boot. It starts to get confusing real quick.

jlebon: My main concern is that I don't want to make this conditional on another stamp file. So maybe it's simplest if we just have the script delete it once it's done. Then it's naturally one shot. It also solves my other concern of the stamp file living forever and possibly not being in sync with what's in `/etc`.

jlebon:
Updated this, but haven't sanity-checked manually myself. Let's see what CI says.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.service
dustymabe: Yeah. I actually kind of like having a second stamp file :). Mainly because there is one that was the original request (requesting for the service to do something) and there is a second one to let us know the service ran (in case the journal had been wrapped and we no longer have that information in `journalctl`). I agree it's not great the have the stamp files hang around forever, but then again it could be nice (given our new use of `/etc/fedora-coreos/`) to add a datestamp (or some other time indicator) to the stamp files (i.e. `/etc/fedora-coreos/202203-iptables-legacy.stamp`) and we can just let them collect, almost like a little log of system history. Given that we periodically have migrations like this, it could be useful to see them all in one place.

travier (dismissed):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.service
jlebon: I was thinking about this and I think we might be tackling this slightly wrong. This is introducing an API that allows new installs to default to iptables-legacy using a stamp file. Do we want to commit to that? For how long? I think instead it should be a documented Butane snippet setting the symlinks (or if we really want, some Butane sugar for it).

So then this here would be concerned strictly about upgrading nodes. In which case we can delete this script after the next barrier.

Re. keeping the stamp file in `/etc` as a "log", I'm still not convinced. :) But we could also add a stamp file in `/var/lib` as a marker that we ran if we really want something more permanent (but note also that `ostree admin config-diff` is the final source of truth for what local changes have been made anyway).

@bgilbert Any thoughts?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
bgilbert: This is in `05core` though.  Should it be in `15fcos` instead?  If not, maybe we should be using `/etc/coreos` here.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.service
bgilbert: Agree on not mutating configuration on every boot.  Also agree that this is currently introducing an API with unbounded lifetime, sort of an "anti-feature flag".  Given the lack of consensus on the feature flags proposal in https://github.com/coreos/fedora-coreos-tracker/issues/892, I think it'd be better to avoid that.  The same reasoning mitigates against Butane sugar.  A documented Butane snippet for new installs seems reasonable to me; it's consistent with other migrations we've done in the past.

As to retaining the upgrade stamp forever, I'm also not convinced.  The symlinks themselves are the source of truth for which mode we're in, and how we got there seems less relevant.  When old journal entries age out, by definition that's an indication that we no longer care about the corresponding slice of historical record, so I don't think we need to go out of our way to stash the info somewhere else.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
jlebon: This migration specifically, RHCOS has already done it. Though I like the idea of keeping the door open for things that do also affect RHCOS. So leaning towards using `/etc/coreos`, and putting the specific things that leverage the directory in either `05core` or `15fcos` depending on whether it needs to apply to RHCOS too (so this one for example, would go in `15fcos`). Thoughts?

jlebon:
Rebased and tweaked slightly based on discussions. Still need to address feedback from https://github.com/coreos/fedora-coreos-config/pull/1324#discussion_r761658999.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-iptables-legacy.sh
bgilbert: SGTM!

jlebon:
Now with https://github.com/coreos/fedora-coreos-config/pull/1324#discussion_r761658999 addressed! I've also updated [the email draft](https://github.com/coreos/fedora-coreos-tracker/issues/676#issuecomment-962070666) to reflect the latest approach. Let's try to send it this week?

overlay.d/15fcos/usr/lib/dracut/modules.d/35coreos-iptables/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# If there are any modifications to the symlinks, do nothing. This is basically
+# like `ostree admin config-diff` but more focused and lighter/safer than doing
+# a bwrap call and grepping output.
+for symlink in "${!SYMLINKS[@]}"; do
+    symlink=/etc/alternatives/$symlink
+    if ! symlink_is_default "$symlink"; then
+        echo "Symlink $symlink is not default; exiting without modifying."
+        exit 0
+    fi
```
dustymabe: I found the re-use of `$symlink` slightly harder to follow than using
a new variable. Suggestion: 

```suggestion
    symlinkpath=/etc/alternatives/$symlink
    if ! symlink_is_default "$symlinkpath"; then
        echo "Symlink $symlinkpath is not default; exiting without modifying."
        exit 0
    fi
```

If we do this we might want to update `symlink_is_default()` function too.

overlay.d/15fcos/usr/lib/dracut/modules.d/35coreos-iptables/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,62 @@
...
+        echo "Executable $path no longer present; exiting."
+        exit 0
+    fi
+done
+
+symlink_is_default() {
+    local symlink=$1; shift
+    # check that the deployment is still using the symlink (i.e. the user didn't
+    # do something funky), and that the OSTree default is still symlink-based
+    # (i.e. that we didn't change strategy and forgot to update this script)
```
dustymabe:
```
# (i.e. that we didn't change strategy and forgot to update this script)
```

Should we hard error in that case?

overlay.d/15fcos/usr/lib/dracut/modules.d/35coreos-iptables/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,62 @@
...
+    if [ ! -L "/sysroot/$symlink" ] || [ ! -L "/sysroot/usr/$symlink" ]; then
+        return 1
+    fi
+    # compare symlink targets between deployment and OSTree default
+    if [ "$(readlink "/sysroot/$symlink")" != "$(readlink "/sysroot/usr/$symlink")" ]; then
+        return 1
+    fi
```
dustymabe: Obviously not necessary but might enhance readability:

```suggestion
    fi
    return 0 # symlink matches the default
```

overlay.d/15fcos/usr/lib/dracut/modules.d/35coreos-iptables/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,62 @@
...
+    [ip6tables-restore]=ip6tables-legacy-restore
+    [ip6tables-save]=ip6tables-legacy-save
+    [iptables]=iptables-legacy
+    [iptables-restore]=iptables-legacy-restore
+    [iptables-save]=iptables-legacy-save
+)
+
+# sanity-check the stamp file is present
+if [ ! -e /sysroot/etc/coreos/iptables-legacy.stamp ]; then
+    exit 0
```
dustymabe: Since the systemd unit has `ConditionPathExists=/sysroot/etc/coreos/iptables-legacy.stamp` maybe we should at least log a statement here before we `exit 0`.

overlay.d/15fcos/usr/lib/dracut/modules.d/35coreos-iptables/coreos-enable-iptables-legacy.sh
```diff
@@ -0,0 +1,62 @@
...
+fi
+
+# if legacy doesn't exist on the host anymore, do nothing
+for legacy in "${SYMLINKS[@]}"; do
+    path=/sysroot/usr/sbin/$legacy
+    if [ ! -e "$path" ]; then
+        echo "Executable $path no longer present; exiting."
+        exit 0
```
dustymabe: feels like this should be a hard error. If this is true we failed IMO. Maybe we should add an appropriate check in our kola tests to make sure this is never true... Or I guess `iptables --version | grep legacy` (in the kola test already) is sufficient?

manifests/fedora-coreos-base.yaml
```diff
@@ -84,6 +84,17 @@ postprocess:
...
+  # remove iptables-legacy.
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    ln -sf /usr/sbin/ip6tables-nft /etc/alternatives/ip6tables
+    ln -sf /usr/sbin/ip6tables-nft-restore /etc/alternatives/ip6tables-restore
+    ln -sf /usr/sbin/ip6tables-nft-save /etc/alternatives/ip6tables-save
+    ln -sf /usr/sbin/iptables-nft /etc/alternatives/iptables
+    ln -sf /usr/sbin/iptables-nft-restore /etc/alternatives/iptables-restore
+    ln -sf /usr/sbin/iptables-nft-save /etc/alternatives/iptables-save
```
dustymabe: optional: enhance spacing for readability

```suggestion
    ln -sf /usr/sbin/iptables-nft          /etc/alternatives/iptables
    ln -sf /usr/sbin/iptables-nft-restore  /etc/alternatives/iptables-restore
    ln -sf /usr/sbin/iptables-nft-save     /etc/alternatives/iptables-save
    ln -sf /usr/sbin/ip6tables-nft         /etc/alternatives/ip6tables
    ln -sf /usr/sbin/ip6tables-nft-restore /etc/alternatives/ip6tables-restore
    ln -sf /usr/sbin/ip6tables-nft-save    /etc/alternatives/ip6tables-save
```

dustymabe:
Carrying forward the converstaion from [this thread](https://github.com/coreos/fedora-coreos-config/pull/1324#discussion_r742843554) in https://github.com/coreos/fedora-coreos-config/pull/1324#discussion_r756450497 you said *This is introducing an API that allows new installs to default to iptables-legacy using a stamp file*. Is that not still true in the current code? If not, what mechanism is preventing someone from creating the stamp file using Ignition and having the service run? `ConditionKernelCommandLine=!ignition.firstboot` means it won't run on the Ignition boot, but what about a second boot?

jlebon:
> Carrying forward the converstaion from [this thread](https://github.com/coreos/fedora-coreos-config/pull/1324#discussion_r742843554) in [#1324 (comment)](https://github.com/coreos/fedora-coreos-config/pull/1324#discussion_r756450497) you said _This is introducing an API that allows new installs to default to iptables-legacy using a stamp file_. Is that not still true in the current code? If not, what mechanism is preventing someone from creating the stamp file using Ignition and having the service run? `ConditionKernelCommandLine=!ignition.firstboot` means it won't run on the Ignition boot, but what about a second boot?

Yeah, I've thought about this and tried to find a succinct way to block it without introducing yet another stamp file, but couldn't think of one. I don't think it's a real concern though, do you? The way it's set up, it's clearly not ergonomic to use that way and we do provide the Butane snippet to use. There's always a chance someone can abuse or disable the safeguards you've put in place so it didn't seem productive to worry about it.

dustymabe:
> Yeah, I've thought about this and tried to find a succinct way to block it without introducing yet another stamp file, but couldn't think of one.

What about using the aleph-version as another knob for the script? 

```
$ jq -r .build /sysroot/.coreos-aleph-version.json
35.20211119.1.0
```

So we could allow up to a certain version.

Somewhat related: we should probably have a library or something for comparing versions of FCOS.

overlay.d/15fcos/usr/lib/dracut/modules.d/35coreos-iptables/coreos-enable-iptables-legacy.sh
jlebon: I think it'd be good to be resilient to e.g. some layered package or some customization having changed the tree (or e.g. with https://github.com/coreos/enhancements/pull/7, that'd be trivially possible though this script probably won't live that long).

> Or I guess `iptables --version | grep legacy` (in the kola test already) is sufficient?

:+1: Agreed.

overlay.d/15fcos/usr/lib/dracut/modules.d/35coreos-iptables/coreos-enable-iptables-legacy.sh
jlebon: It's not easy to tell those two cases apart though (this is related to [the other comment](https://github.com/coreos/fedora-coreos-config/pull/1324#discussion_r764238218)).

jlebon: But also in general, I think for this we should try to be really good at detecting the state we expect, but otherwise just leave everything untouched without alarming the user. These updating machines could be ancient and carry a lot of state.

dustymabe: Yeah. I think I was mostly just assuming if this hard fails we'll catch it in CI and resolve the problem. I agree, though, that if we wanted that it should be part of a CI test and not part of these units.

jlebon:
> > Yeah, I've thought about this and tried to find a succinct way to block it without introducing yet another stamp file, but couldn't think of one.
> 
> What about using the aleph-version as another knob for the script?
> 
> ```
> $ jq -r .build /sysroot/.coreos-aleph-version.json
> 35.20211119.1.0
> ```
> 
> So we could allow up to a certain version.

Hmm yeah, maybe...It'd have to be a version after or equal to the one where we switch to nft. I guess we edit this PR at the last minute when it's ready to be merged, and use the last release version numbers? Not sure. I think it'd be harder to screw up if we just add a stamp file if we really care about this (which I'm still not sure we should).

travier:
> What about using the aleph-version as another knob for the script?
> 
> ```
> $ jq -r .build /sysroot/.coreos-aleph-version.json
> 35.20211119.1.0
> ```
> 
> So we could allow up to a certain version.
> 
> Somewhat related: we should probably have a library or something for comparing versions of FCOS.

This is very similar to the discussion in https://github.com/coreos/fedora-coreos-tracker/issues/892#issuecomment-876557840.

jlebon:
Discussed with @dustymabe about this and realized to be nice we should have the Butane config we send in the email work both before and after the migration. I've updated the PR now to defend against users writing the stamp file via Ignition as an API to switch to legacy post-migration.

That said, nothing prevents them from just manually creating the stamp file after provisioning and rebooting. We'd probably need a separate stamp file to fix that, but again, I'm not convinced it's worth it. I'm increasingly thinking we should make a barrier just for this, so then we can just rip it out after one release.

Also updated email draft.

dustymabe:
> I'm increasingly thinking we should make a barrier just for this, so then we can just rip it out after one release.

+1 for a barrier.

dustymabe (dismissed):
LGTM

tests/kola/firewall/iptables-nft/test.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+ok() {
```
travier: Needs a minor update for new ok/fatal function inclusion

dustymabe:
I think the new test might need another update to conform to what is being worked on over in https://github.com/coreos/fedora-coreos-config/pull/1391. The existing test might need to be rebased once that merges.

jlebon:
OK, pushed some stuff here and in https://github.com/coreos/fedora-coreos-config/pull/1460, though still testing it. Effectively:
- we add a 35coreos-iptables overlay dir containing the opt-out logic which won't be in testing-devel, but will be in next-devel (using the new ability for cosa to not autolayer all the overlay dirs)
- we add tests which will verify iptables-legacy on `!next` and iptables-nft on `next`
- we move `next` to iptables-legacy using a postprocess script

I need to write up a checklist of how we're going to roll this out so that we don't get confused.

tests/kola/firewall/iptables-nft/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+stream=$(rpm-ostree status -b --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]')
```
jlebon: https://github.com/coreos/fedora-coreos-config/pull/1461 adds a `get_fcos_stream` function to the common lib. Will rebase to use that once it merges.

dustymabe: it's merged! 🎉

tests/kola/firewall/iptables-nft/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# kola: { "exclusive": false }
```
dustymabe: I think we need to update this to comply with the new rules: https://github.com/coreos/fedora-coreos-config#tests-layout

At least a description.

tests/kola/firewall/iptables-legacy/test.sh
```diff
@@ -1,18 +1,16 @@
 #!/bin/bash
-# kola: { "distros": "fcos", "exclusive": false }
+# kola: { "distros": "fcos", "exclusive": true }
 # This test is currently scoped to only FCOS because the RHCOS version of `iptables`
 # is using the `nf_tables` backend.
 # TODO: modify this test to check for `nf_tables` backend when FCOS switches.
 #       See https://github.com/coreos/fedora-coreos-config/pull/1324
```
dustymabe: should we drop this comment since we've got a new test testing iptables nft?

tests/kola/firewall/iptables-nft/test.sh
```diff
@@ -0,0 +1,24 @@
...
+        ok "iptables in nft mode"
+        ;;
+    *)
+        # Make sure we're on legacy iptables
+        if ! iptables --version | grep legacy; then
+            iptables --version # output for logs
+            fatal "iptables version is not legacy"
+        fi
+        ok "iptables in legacy mode"
+        ;;
```
dustymabe: since we're essentially testing: "make sure iptables version is the default we expect", maybe we just rename this test to `iptables` or `iptables-backend`?

dustymabe: once everything is defaulting to `nft` we'll just drop the conditional logic.

tests/kola/firewall/iptables-legacy/test.sh
```diff
@@ -1,18 +1,16 @@
 #!/bin/bash
-# kola: { "distros": "fcos", "exclusive": false }
+# kola: { "distros": "fcos", "exclusive": true }
 # This test is currently scoped to only FCOS because the RHCOS version of `iptables`
```
dustymabe: let's add a comment explaining why this needs to be exclusive now.

jlebon:
Updated and tested!

jlebon:
Let's write up a checklist in https://github.com/coreos/fedora-coreos-tracker/issues/676 to detail how we're going to roll this out.

dustymabe (dismissed):
LGTM

dustymabe (approved):
LGTM

PR #1325, opened 2021-11-04
tests: separate out misc-ro test

mike-nguyen:
Break up misc-ro into individual tests now that kola supports combing
non-exclusive tests into a single VM.


tests/kola/files/file-directory-permissions
```diff
@@ -0,0 +1,19 @@
...
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+list="$(find /etc -type f,d -perm /022)"
+if [[ -n "${list}" ]]; then
+>---find /etc -type f,d -perm /022 -print0 | xargs -0 ls -al
+>---fatal "found files or directories with 'g+w' or 'o+w' permission"
```
dustymabe: copy paste issue here?

dustymabe:
Note https://github.com/coreos/fedora-coreos-config/pull/1322 - we'll want to merge one of these two PRs first and make sure we rebase the other one and account for the changes properly.

jlebon:
Cool! This will give us better reporting. One suggestion is to shove all the `ok`, `fatal`, etc... functions in a shared file at e.g. `tests/kola/shared/lib.sh` and then add `data/shared` symlinks from each directory to that one. Then we'd be able to source the shared file to use the functions using [`KOLA_EXT_DATA`](https://coreos.github.io/coreos-assembler/kola/external-tests/#using-kola-run-with-externally-defined-tests). I know kola will handle `data` being a symlink, though it might not handle a subdir symlink like `data/shared` (so that we could still have test-specific data). We can fix that if it doesn't work.

mike-nguyen:
> Cool! This will give us better reporting. One suggestion is to shove all the `ok`, `fatal`, etc... functions in a shared file at e.g. `tests/kola/shared/lib.sh` and then add `data/shared` symlinks from each directory to that one. Then we'd be able to source the shared file to use the functions using [`KOLA_EXT_DATA`](https://coreos.github.io/coreos-assembler/kola/external-tests/#using-kola-run-with-externally-defined-tests). I know kola will handle `data` being a symlink, though it might not handle a subdir symlink like `data/shared` (so that we could still have test-specific data). We can fix that if it doesn't work.

Thanks jlebon! I was wondering how a shared library would work and `KOLA_EXT_DATA` looks like a good path.

jlebon:
> > Cool! This will give us better reporting. One suggestion is to shove all the `ok`, `fatal`, etc... functions in a shared file at e.g. `tests/kola/shared/lib.sh` and then add `data/shared` symlinks from each directory to that one. Then we'd be able to source the shared file to use the functions using [`KOLA_EXT_DATA`](https://coreos.github.io/coreos-assembler/kola/external-tests/#using-kola-run-with-externally-defined-tests). I know kola will handle `data` being a symlink, though it might not handle a subdir symlink like `data/shared` (so that we could still have test-specific data). We can fix that if it doesn't work.
> 
> Thanks jlebon! I was wondering how a shared library would work and `KOLA_EXT_DATA` looks like a good path.

To be clear, I think it's OK if we tackle this as a follow-up, especially if it requires additional work in kola. Let me know if you want to do that! This PR is a really nice improvement as is and will unblock a few other PRs.

mike-nguyen:
@saqibali-2k is looking into a bug right now where the `data` directory doesn't get copied when running as non-exclusive.  This may have further implications when multiple tests use a `data` directory.  The change to support the symlink to a shared directory is a one liner.  

I'll do a fixup for the indentation dusty pointed out and we can merge this to unblock the other PRs.

mike-nguyen:
^^ Pushed fixes

dustymabe (approved):
LGTM

mike-nguyen:
The shared library will need these fixes https://github.com/coreos/coreos-assembler/pull/2557 and https://github.com/coreos/coreos-assembler/pull/2556

PR #1327, opened 2021-11-04
[next-devel] overrides: unpin kernel-5.14.13-300.fc35

dustymabe:
This probably should have been classified as a fast-track so our
graduated overrides bot would have cleaned it up like the others.


dustymabe:
also added a commit to go back to the non-next repos since f35 has been released now

jmarrero (approved):
lgtm

PR #1329, opened 2021-11-05
[rawhide] manifest.yaml: remove now common manifest.yaml entries

dustymabe:
These are now in the shared `fedora-coreos-base.yaml` so we
can remove them from the toplevel manifest.yaml.


jlebon (dismissed):

dustymabe:
added a commit to pin the kernel for https://github.com/coreos/fedora-coreos-tracker/issues/1011. Let's see if CI passes now.

jlebon (approved):

saqibali-2k (approved):

PR #1331, opened 2021-11-05
Fast-track skopeo-1.5.1-1.fc35

cgwalters:
For https://github.com/coreos/enhancements/pull/7

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,10 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  skopeo:
+    evr: 1.5.1-1.fc35
+    metadata:
+      bodhi:  https://bodhi.fedoraproject.org/updates/FEDORA-2021-b8fee395ca
+      reason: https://github.com/coreos/enhancements/pull/7
+      type: fast-track


manifest-lock.overrides.yaml
```diff
@@ -8,4 +8,10 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  skopeo:
+    evr: skopeo-1.5.1-1.fc35
```
jlebon: This needs to be

```suggestion
    evr: 1.5.1-1.fc35
```

cgwalters: Done, thanks

dustymabe (approved):
LGTM

dustymabe:
Note. Because of the current timing this won't land in `testing` for another two weeks. If you want it to land in `next` this week then I suggest you open a `next-devel` PR as well.

cgwalters:
I guess I might as well just wait on the bodhi update to go stable :cry:

jlebon:
rpm-ostree's CI (and almost everything else) uses testing-devel so once this merges, we should be unblocked.

manifest-lock.overrides.yaml
```diff
@@ -8,4 +8,10 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  skopeo:
+    evr: 1.5.1-1.fc35
```
jlebon, line 13: Ouch, looks like it was also missing an epoch:

```suggestion
    evr: 1:1.5.1-1.fc35
```

dustymabe:
`skopeo-1.5.1-1.fc35` is [stable now](https://bodhi.fedoraproject.org/updates/FEDORA-2021-b8fee395ca) so this PR is obsolete.

PR #1332, opened 2021-11-08
Fedora 35 is out!

dustymabe:
Let's update all references to point to F35 now.

Also threw in a revert for bed3837 since I was in that file.

### DIFF
--- a/ci/buildroot/Dockerfile
+++ b/ci/buildroot/Dockerfile
@@ -5,6 +5,6 @@
 #
 # This image is used by CoreOS CI to build software like
 # Ignition, rpm-ostree, ostree, coreos-installer, etc...
-FROM registry.fedoraproject.org/fedora:34
+FROM registry.fedoraproject.org/fedora:35
 COPY . /src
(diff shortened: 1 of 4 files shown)


ci/buildroot/Dockerfile
```diff
@@ -5,6 +5,6 @@
 #
 # This image is used by CoreOS CI to build software like
 # Ignition, rpm-ostree, ostree, coreos-installer, etc...
-FROM registry.fedoraproject.org/fedora:34
+FROM registry.fedoraproject.org/fedora:35
 COPY . /src
-RUN ./src/install-buildroot.sh && yum clean all && rm /src -rf  # nocache 20210916
+RUN ./src/install-buildroot.sh && yum clean all && rm /src -rf  # nocache 20211108
```
jlebon, line 10: Minor: not sure this is necessary since an earlier line changed.

jlebon (approved):

ci/buildroot/Dockerfile
dustymabe: yeah, I decided to update it anyway just as an indicator of "freshness".

PR #1338, opened 2021-11-15
tests/kola/kdump: rename tests; fix denylist entry

bgilbert:
Now that we have another test in the same directory, kola is calling the broken one `ext.config.kdump.test.sh` and the denylist entry isn't matching it.  Clean up the test names and fix the denylist.

Fixes #1325 and upstream CI.


dustymabe (approved):
LGTM - cc @mike-nguyen @travier

PR #1340, opened 2021-11-16
tests/coreos-platform-chrony-generator: make non-exclusive

jlebon:
There's nothing that mutates the host in there (after removing the
unnecessary `mktemp -d`) so let's just make it shared.

### DIFF
--- a/tests/kola/chrony/coreos-platform-chrony-generator
+++ b/tests/kola/chrony/coreos-platform-chrony-generator
@@ -2,9 +2,7 @@
 set -euo pipefail
 # Test the coreos-platform-chrony generator.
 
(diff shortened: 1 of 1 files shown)


tests/kola/chrony/coreos-platform-chrony-generator
```diff
@@ -2,9 +2,7 @@
 set -euo pipefail
 # Test the coreos-platform-chrony generator.
 
-# kola: { "platforms": "aws,azure,gcp" }
-
-cd "$(mktemp -d)"
```
dustymabe, line 7: weird, wonder how that snuck in there?

dustymabe (approved):
LGTM

PR #1341, opened 2021-11-17
tests/systemd-repart-service: only run on FCOS

miabbott:
The version of `systemd-udev` on RHCOS does not include the
`systemd-repart` service.

Missed this test as part of #1322.


dustymabe (dismissed):
LGTM

mike-nguyen (approved):
LGTM

PR #1344, opened 2021-11-19
tests: check for mounting /boot by UUID on RHCOS

miabbott:
When the next git submodule update gets pulled into `openshift/os`,
RHCOS will start mounting `/boot` by UUID. (See openshift/os#678)

Let's drop the "fcos" tags on those tests that are checking for
mounting /boot by UUID, so we can change the behavior and get unit
tests enabled in one change.


jlebon (approved):
Thanks!

PR #1350, opened 2021-11-23
Break up misc-ro.sh into separated cases

HuijingHei:
Break up misc-ro.sh into separated cases

### DIFF
--- a/tests/kola/disks/tmpfs
+++ b/tests/kola/disks/tmpfs
@@ -1,0 +1,19 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# Ensure we have tmpfs on /tmp
+tmpfs=$(findmnt -n -o FSTYPE /tmp)
+if [ "${tmpfs}" != "tmpfs" ]; then
+    fatal "Error: Expected tmpfs on /tmp, found: ${tmpfs}"
+fi
+ok "tmpfs on /tmp"
--- a/tests/kola/files/kernel-headers
+++ b/tests/kola/files/kernel-headers
@@ -1,0 +1,21 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# check that we are not including the kernel headers on the host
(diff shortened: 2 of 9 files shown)


dustymabe:
I think we need more context here in the git commit message.. I *think* I know the answer to these questions but most people might not:

- What misc-ro.sh are you referring to?
- Where did these tests come from?

HuijingHei:
> I think we need more context here in the git commit message.. I _think_ I know the answer to these questions but most people might not:
> 
> * What misc-ro.sh are you referring to?
> * Where did these tests come from?

Thanks @dustymabe for your review
misc-ro.sh is from https://github.com/openshift/os/blob/master/tests/kola/misc-ro/misc-ro.sh
This PR is shared cases with fcos and rhcos, rhcos only PR is https://github.com/openshift/os/pull/680

dustymabe:
> Thanks @dustymabe for your review

No Problem! :) 

> misc-ro.sh is from https://github.com/openshift/os/blob/master/tests/kola/misc-ro/misc-ro.sh This PR is shared cases with fcos and rhcos, rhcos only PR is [openshift/os#680](https://github.com/openshift/os/pull/680)

Right, but the git commit message doesn't say any of that.. Here's an example of what the git commit message should probably look like:

```
tests: import test cases from openshift/os repo

These are tests cases that were in `misc-ro.sh` in the openshift/os repo [1].
We're importing them here to share them between FCOS and RHCOS because
they benefit both platforms.

[1] https://github.com/openshift/os/blob/43044b95eb9288162dd62b84ad7dc7b6c8b5707a/tests/kola/misc-ro/misc-ro.sh
```

tests/kola/networking/dhcpclient
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
dustymabe: I would maybe rename this file `nm-dhcp-client`

tests/kola/files/selinux
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
dustymabe: let's create a new `tests/kola/selinux` dir and put this test under `tests/kola/selinux/enforcing`

tests/kola/files/usrlocal-selinux
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
dustymabe: let's create a new `tests/kola/selinux` dir and put this test under `tests/kola/selinux/usrlocal-context`

HuijingHei:
Done, thanks!

dustymabe (approved):
LGTM

PR #1351, opened 2021-11-25
tests: Use an uniform layout for test scripts

travier:
This is a first pass to unify the test layout. Ideal we would have something like this for all tests:
```
#!/usr/bin/bash
# kola option comment
# Documentation explaining the test
set -euxo ...

fatal() function definition or common function inclusion

other_functions() definitions

actual test code

# Or should we wrap all test code in main functions? Usually it's not required.
main() {
    actual test code
}
main "${@}" <-- We need to make sure to preserve the return code here
```


miabbott:
@aaradhak is working on making a library with `fatal()` and `ok()` functions; there's probably opportunity for swarming/sharing here

travier:
Sure! @aaradhak feel free to pick up this one.

travier:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/1391

PR #1353, opened 2021-11-29
s390x: add secure-execution support

nikita-dubrovskii:
This adds `genprotimg` tool, for creating encrypted `sd-boot` image, and turns on (if required) SecureExecution during `firstboot-complete` 
Depends on: https://github.com/coreos/coreos-installer/pull/851

### DIFF
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -28,6 +28,7 @@ packages-s390x:
   # On Fedora, this is provided by s390utils-core. on RHEL, this is for now
   # provided by s390utils-base, but soon will be -core too.
   - /usr/sbin/zipl
+  - /usr/bin/genprotimg
 packages-x86_64:
   - grub2 grub2-efi-x64 efibootmgr shim
   - microcode_ctl
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
@@ -7,11 +7,15 @@ set -euo pipefail
 UNINITIALIZED_GUID='00000000-0000-4000-a000-000000000001'
 
 # If it's on multipath, get the parent device from udev properties.
-DM_MPATH=$(eval $(udevadm info --query property --export "$1") && echo "${DM_MPATH:-}")
+# If it's LUKS2 encrypted, UUID woud start with "CRYPT-LUKS2"
+eval $(udevadm info --query property --export "$1" | grep -e DM_NAME -e DM_UUID -e DM_MAPTH)
 
-if [ -n "${DM_MPATH:-}" ]; then
+if [[ -n "${DM_MPATH:-}" ]]; then
     PKNAME=/dev/mapper/$DM_MPATH
     PTUUID=$(eval $(udevadm info --query property --export "$PKNAME") && echo "${ID_PART_TABLE_UUID:-}")
+elif [[ "${DM_UUID:-}" =~ 'CRYPT-LUKS2' ]] && [[ -n "${DM_NAME:-}" ]]; then
+    eval $(lsblk --output PKNAME --pairs --paths --nodeps /dev/mapper/$DM_NAME)
+    PTUUID=$(eval $(udevadm info --query property --export /dev/mapper/$DM_NAME) && echo "${ID_PART_TABLE_UUID:-}")
 else
     # On RHEL 8 the version of lsblk doesn't have PTUUID. Let's detect
     # if lsblk supports it. In the future we can remove the 'if' and
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
@@ -17,8 +17,21 @@ path=/sysroot
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
 
+# IBM SecureExecution
+secure_execution=0
+if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+fi
+
 if [ ! -f "${saved_partstate}" ]; then
     partition=$(realpath /dev/disk/by-label/root)
+    # in the Secure Execution case, the rootfs is pre-baked on LUKS, so
+    # `partition` is actually the device mapper device; get its parent
+    if [[ "${secure_execution}" != "0" ]]; then
+        # lsblk doesn't print PKNAME of crypt devices with --nodeps
+        MAJMIN=$(echo $(lsblk -dno MAJ:MIN "${partition}"))
+        partition=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
+    fi
 else
     # The rootfs was reprovisioned. Our rule in this case is: we only grow if
     # the partition backing the rootfs is the same and its size didn't change
@@ -85,7 +98,11 @@ while true; do
              eval $(udevadm info --query=property --export "${NAME}")
              # lsblk doesn't print PKNAME of crypt devices with --nodeps
              PKNAME=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
-             clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
+             if [[ "${secure_execution}" != "0" ]]; then
+                 cryptsetup resize --key-file=/etc/luks/root "${DM_NAME}"
+             else
+                 clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
+             fi
             )
             ;;
         # already checked
--- a/overlay.d/05core/usr/libexec/coreos-ignition-firstboot-complete
+++ b/overlay.d/05core/usr/libexec/coreos-ignition-firstboot-complete
@@ -3,10 +3,6 @@ set -euo pipefail
 
 mount -o remount,rw /boot
 
-if [[ $(uname -m) = s390x ]]; then
-    zipl
-fi
-
 # We're done provisioning. Remove the whole /boot/ignition directory if present,
 # which may include a baked Ignition config. See
 # https://github.com/coreos/fedora-coreos-tracker/issues/889.
@@ -16,3 +12,8 @@ rm -rf /boot/ignition
 # this file. Fail if we are unable to remove it, rather than risking rerunning
 # Ignition at next boot.
 rm /boot/ignition.firstboot
+
+# rdcore zipl checks for /boot/ignition.firstboot
+if [[ $(uname -m) = s390x ]]; then
+    /usr/lib/dracut/modules.d/50rdcore/rdcore zipl --boot-mount=/boot
+fi


bgilbert (dismissed):
As previously agreed, hold until feature planning is complete.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -80,13 +80,26 @@ while true; do
...
             ;;
         crypt)
-            # XXX: yuck... we need to expose this sanely in clevis
-            (. /usr/bin/clevis-luks-common-functions
-             eval $(udevadm info --query=property --export "${NAME}")
-             # lsblk doesn't print PKNAME of crypt devices with --nodeps
-             PKNAME=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
-             clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
-            )
+            eval $(udevadm info --query=property --export "${NAME}")
```
travier: Might be nitpicking a bit but I think it might be easier to follow if we do the generic case/arch first and then do the specific one. Maybe another switch case?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
```diff
@@ -9,9 +9,16 @@ UNINITIALIZED_GUID='00000000-0000-4000-a000-000000000001'
 # If it's on multipath, get the parent device from udev properties.
 DM_MPATH=$(eval $(udevadm info --query property --export "$1") && echo "${DM_MPATH:-}")
 
-if [ -n "${DM_MPATH:-}" ]; then
+# If it's LUKS2 encrypted, UUID woud start with "CRYPT-LUKS2"
+DM_UUID=$(eval $(udevadm info --query property --export "$1") && echo "${DM_UUID:-}")
```
travier: Lots of eval here. Are they really needed? Can they be abused by an attacker with access to the disk?

travier (commented):
Only added some suggestions as I don't have the context to review.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
nikita-dubrovskii: hmm, attacker can only use fake-ignition config and access system with it, but that's out of scope

nikita-dubrovskii: >Lots of eval here

removed

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: done

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -80,13 +80,25 @@ while true; do
...
+            if [[ $(uname -m) != s390x ]]; then
```
jlebon: Don't we still need to support non-SE s390x instances which still want LUKS though? I.e. should this be switching on SE vs non-SE rather than s390x vs non-s390x?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -80,13 +80,25 @@ while true; do
...
+                )
+            else
+                if [[ ! -f /etc/luks/root ]]; then
+                    echo "error: /etc/luks/root doesn't exist"
+                    exit 1
+                fi
+                device=$(realpath /dev/disk/by-label/"${DM_NAME}")
+                eval $(udevadm info --query=property --export $device | grep PARTN=)
+                eval $(lsblk -o PKNAME --nodeps --noheadings --paths --pairs $device)
+                parted $PKNAME resizepart $PARTN 100%
```
jlebon: Isn't this already done by `growpart` at the earlier `part` iteration?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
```diff
@@ -7,11 +7,19 @@ set -euo pipefail
 UNINITIALIZED_GUID='00000000-0000-4000-a000-000000000001'
 
 # If it's on multipath, get the parent device from udev properties.
-DM_MPATH=$(eval $(udevadm info --query property --export "$1") && echo "${DM_MPATH:-}")
+# If it's LUKS2 encrypted, UUID woud start with "CRYPT-LUKS2"
+DM_MPATH=
+DM_UUID=
+DM_NAME=
```
jlebon: It looks like we're correctly handling the undefined case in all the bits below, so I think we could drop these upfront declarations.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: if it does, than it's broken or doesn't work wit LUKS pre-encrypted partitions

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: right. will fix

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: That `growpart` is tested by CI at least. There's actually a LUKS test which does this: https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola/root-reprovision/luks. That should be the same setup as we have here (root on LUKS on partition).

nikita-dubrovskii: Some debug output:
```
[   19.502720] ignition-ostree-growfs[1292]: ++ realpath /dev/disk/by-label/root
[   19.503058] ignition-ostree-growfs[1288]: + partition=/dev/dm-0
[   19.503216] ignition-ostree-growfs[1293]: + lsblk -no TYPE /dev/dm-0
...
[   19.504843] ignition-ostree-growfs[1288]: + current_blkdev=/dev/dm-0
[   19.504855] ignition-ostree-growfs[1288]: + true
[   19.504933] ignition-ostree-growfs[1296]: ++ lsblk --paths --nodeps --pairs -o NAME,TYPE,PKNAME /dev/dm-0
[   19.505523] ignition-ostree-growfs[1288]: + eval 'NAME="/dev/mapper/crypt_rootfs" TYPE="crypt" PKNAME=""'
[   19.505538] ignition-ostree-growfs[1288]: ++ NAME=/dev/mapper/crypt_rootfs
[   19.505550] ignition-ostree-growfs[1288]: ++ TYPE=crypt
[   19.505561] ignition-ostree-growfs[1288]: ++ PKNAME=
[   19.505892] ignition-ostree-growfs[1298]: +++ lsblk -dno MAJ:MIN /dev/mapper/crypt_rootfs
```
 So we always enter `crypt` case-switch first. And `part` is unreachable in that case

nikita-dubrovskii: >That should be the same setup as we have here (root on LUKS on partition).

pls correct me if i'm wrong, but that test creates LUKSed root with `ignition.disks` stage

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
```diff
@@ -7,11 +7,15 @@ set -euo pipefail
...
+# If it's LUKS2 encrypted, UUID woud start with "CRYPT-LUKS2"
+eval $(udevadm info --query property --export "$1" | grep -e DM_NAME -e DM_UUID -e DM_MAPTH)
 
-if [ -n "${DM_MPATH:-}" ]; then
+if [[ -n "${DM_MPATH:-}" ]]; then
     PKNAME=/dev/mapper/$DM_MPATH
     PTUUID=$(eval $(udevadm info --query property --export "$PKNAME") && echo "${ID_PART_TABLE_UUID:-}")
+elif [[ "${DM_UUID:-}" =~ 'CRYPT-LUKS2' ]] && [[ -n "${DM_NAME:-}" ]]; then
+    eval $(lsblk --output PKNAME --pairs --paths --nodeps /dev/disk/by-label/$DM_NAME)
+    PTUUID=$(eval $(udevadm info --query property --export /dev/disk/by-label/$DM_NAME) && echo "${ID_PART_TABLE_UUID:-}")
```
jlebon: `/dev/disk/by-label/$DM_NAME` looks off here. The name of the device mapper device isn't necessarily the filesystem label name on top of it. Probably this should be

```suggestion
    eval $(lsblk --output PKNAME --pairs --paths --nodeps /dev/mapper/$DM_NAME)
    PTUUID=$(eval $(udevadm info --query property --export /dev/mapper/$DM_NAME) && echo "${ID_PART_TABLE_UUID:-}")
```

instead?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: Yes indeed, that test uses Ignition to create the LUKS root. I thought the end result would be the same, but maybe I'm missing something.

What's the output of `lsblk` on a booted s390x system with your cosa patch? Maybe we should do a screen sharing session for this.

nikita-dubrovskii:
```
[core@localhost ~]$ lsblk -f
NAME             FSTYPE      FSVER LABEL        UUID                                 FSAVAIL FSUSE% MOUNTPOINTS
vda                                                                                                 
|-vda1           ext4        1.0   se           71589f1a-5698-4d22-8877-91317173ec61                
|-vda3           crypto_LUKS 2     crypt_bootfs 8dc99436-f245-470e-81c3-d2f07d49dcfa                
| `-crypt_bootfs ext4        1.0   boot         a774674a-e0d6-11ec-8ab8-0f20fe3c51e1    224M    26% /boot
`-vda4           crypto_LUKS 2     crypt_rootfs 5e9d4812-87e5-4ed7-8c68-dd1874031839                
  `-crypt_rootfs xfs               root         a8ec61ea-e0d6-11ec-a07c-525400123456    7.9G    16% /var
                                                                                                    /sysroot/ostree/deploy/fedora-coreos/var
                                                                                                    /usr
                                                                                                    /etc
                                                                                                    /
                                                                                                    /sysroot
```

manifests/bootable-rpm-ostree.yaml
```diff
@@ -28,6 +28,9 @@ packages-s390x:
   # On Fedora, this is provided by s390utils-core. on RHEL, this is for now
   # provided by s390utils-base, but soon will be -core too.
   - /usr/sbin/zipl
+  - /usr/bin/genprotimg
```
jlebon, line 31: Are you able to compose an s390x FCOS with this? I suspect it will fail because of: https://github.com/coreos/fedora-coreos-config/blob/401c56bac7b5d369554e5ba4f36a1c5b5fa94015/manifests/fedora-coreos.yaml#L114

We'll have to temporarily split that out into a separately `conditional-include`d treefile until https://github.com/coreos/fedora-coreos-tracker/issues/1217 is fixed.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: OK right, looking at this more, I think we need to adapt the top of the file to handle Secure Execution correctly. Can you try the following patch on top of this PR?

```diff
diff --git a/manifests/bootable-rpm-ostree.yaml b/manifests/bootable-rpm-ostree.yaml
index b0659834..8c41320c 100644
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -29,8 +29,6 @@ packages-s390x:
   # provided by s390utils-base, but soon will be -core too.
   - /usr/sbin/zipl
   - /usr/bin/genprotimg
-  # resize LUKS encrypted rootfs
-  - /usr/sbin/parted
 packages-x86_64:
   - grub2 grub2-efi-x64 efibootmgr shim
   - microcode_ctl
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
index 1a0d68a7..946de6e1 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
@@ -17,8 +17,21 @@ path=/sysroot
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)

+# IBM SecureExecution
+secure_execution=0
+if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+fi
+
 if [ ! -f "${saved_partstate}" ]; then
     partition=$(realpath /dev/disk/by-label/root)
+    # in the Secure Execution case, the rootfs is pre-baked on LUKS, so
+    # `partition` is actually the device mapper device; get its parent
+    if [[ "${secure_execution}" != "0" ]]; then
+        # lsblk doesn't print PKNAME of crypt devices with --nodeps
+        MAJMIN=$(echo $(lsblk -dno MAJ:MIN "${partition}"))
+        partition=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
+    fi
 else
     # The rootfs was reprovisioned. Our rule in this case is: we only grow if
     # the partition backing the rootfs is the same and its size didn't change
@@ -40,12 +53,6 @@ else
     fi
 fi

-# IBM SecureExecution
-secure_execution=0
-if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
-    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
-fi
-
 # Go through each blockdev in the hierarchy and verify we know how to grow them
 lsblk -no TYPE "${partition}" | while read dev; do
     case "${dev}" in
@@ -86,22 +93,13 @@ while true; do
             fi
             ;;
         crypt)
-            eval $(udevadm info --query=property --export "${NAME}")
-            if [[ "${secure_execution}" == "0" ]]; then
-                # XXX: yuck... we need to expose this sanely in clevis
-                (. /usr/bin/clevis-luks-common-functions
-                # lsblk doesn't print PKNAME of crypt devices with --nodeps
-                PKNAME=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
-                clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
-                )
-            else
-                device=$(realpath /dev/disk/by-label/"${DM_NAME}")
-                eval $(udevadm info --query=property --export $device | grep PARTN=)
-                eval $(lsblk -o PKNAME --nodeps --noheadings --paths --pairs $device)
-                growpart "${PKNAME}" "${PARTN}" || :
-                parted $PKNAME resizepart $PARTN 100%
-                cryptsetup resize --key-file=/etc/luks/root "${DM_NAME}"
-            fi
+            # XXX: yuck... we need to expose this sanely in clevis
+            (. /usr/bin/clevis-luks-common-functions
+             eval $(udevadm info --query=property --export "${NAME}")
+             # lsblk doesn't print PKNAME of crypt devices with --nodeps
+             PKNAME=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
+             clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
+            )
             ;;
         # already checked
         *) echo "unreachable" 1>&2; exit 1 ;;
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
index 1ce53c48..bf9a7872 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -46,7 +46,6 @@ install() {
         xfs_admin  \
         xfs_growfs \
         wc         \
-        parted     \
         wipefs

     # growpart deps
```

overlay.d/05core/usr/libexec/coreos-ignition-firstboot-complete
```diff
@@ -3,10 +3,6 @@ set -euo pipefail
 
 mount -o remount,rw /boot
 
-if [[ $(uname -m) = s390x ]]; then
```
jlebon, line 6: Might be worth mentioning in the commit message introducing this hunk that the reason we're changing ordering here is that `rdcore zipl` checks for `/boot/ignition.firstboot`.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: > OK right, looking at this more, I think we need to adapt the top of the file to handle Secure Execution correctly. Can you try the following patch on top of this PR?

yup, thx. only had to skip `clevis` part in SE. not sure, but maybe `clevis bind` could be used to automate the unlocking here. will  check that

manifests/bootable-rpm-ostree.yaml
nikita-dubrovskii: > Are you able to compose an s390x FCOS with this?

yes, no issues at all (`metal/metal4k/qemu/secex/live/dasd` images)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: looks like we cannot and shouldn't use clevis

jlebon (approved):
Looks great, thanks!

manifests/bootable-rpm-ostree.yaml
jlebon: Ahh interesting. So actually, `/usr/bin/perl` is provided by `perl-interpreter`, which is why this works.

I'll add it to the list, but will take care of doing the `include` splitting to not break s390x builds.

jlebon:
Hmm, interesting. It looks like the FCOS pipeline is failing to build s390x with this:

```
Importing rpm-md... done
rpm-md repo 'fedora-coreos-pool'; generated: 2022-06-26T16:28:46Z solvables: 17070
rpm-md repo 'fedora'; generated: 2022-05-04T21:15:53Z solvables: 57625
rpm-md repo 'fedora-updates'; generated: 2022-06-27T01:26:43Z solvables: 11906
error: Packages not found: /usr/bin/genprotimg
```

jlebon:
Ahh yes of course, we missed a step here. We need to add `s390utils-base` to the s390x lockfile [as documented in the README](https://github.com/coreos/fedora-coreos-config#adding-packages-to-the-os).

manifests/bootable-rpm-ostree.yaml
jlebon: Follow-up to this in https://github.com/coreos/fedora-coreos-config/pull/1848 now that we've moved it out of FCOS again.

PR #1354, opened 2021-11-29
manifests: move crun to user-experience

miabbott:
There is desire to include `crun` as part of RHCOS so that early
testing of OCP workloads using `crun` can be validated [1]

This moves `crun` out of an FCOS only manifest into a manifest that is
shared with RHCOS.

[1] openshift/release#23609


jlebon (approved):

cgwalters:
SGTM, let's `run` with this and `c` where it goes.

PR #1359, opened 2021-11-30
[next-devel] lockfiles: fast-track Ignition v2.13.0-1

sohankunkerkar:


jlebon:
`next-devel` isn't currently used during promotions so this isn't necessary. Check out the section "Disable `next-devel` stream" in https://github.com/coreos/fedora-coreos-tracker/issues/884.

PR #1363, opened 2021-12-04
tests/k8s: enable all platforms

jlebon:
There shouldn't be anything platform-specific about this test, and it's
opt-in anyway, so just drop the `qemu` constraint.

Motivated by wanting to run it in AWS instead to offload the cluster and
try to also test on aarch64.


dustymabe (approved):
LGTM

PR #1364, opened 2021-12-06
tests: common library for 'ok & fatal function'

aaradhak:
A separate common library is created for all the non exclusive external kola tests.
https://issues.redhat.com/browse/GRPA-4478

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+#This is a common library created for the ok & fatal function 
+#and symlinks added to the data/ in each directory
+
+ok() {
+    echo "ok" "$@"
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
--- a/tests/kola/boot/bootupd
+++ b/tests/kola/boot/bootupd
@@ -2,14 +2,7 @@
 # kola: { "exclusive": false }
 set -xeuo pipefail
 
-ok() {
-    echo "ok" "$@"
-}
-
-fatal() {
-    echo "$@" >&2
-    exit 1
-}
+. $KOLA_EXT_DATA/commonlib.sh
 
 case "$(arch)" in
     x86_64|aarch64)
--- a/tests/kola/boot/data/commonlib.sh
+++ b/tests/kola/boot/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh
--- a/tests/kola/chrony/coreos-platform-chrony-generator-permissions
+++ b/tests/kola/chrony/coreos-platform-chrony-generator-permissions
@@ -2,14 +2,7 @@
 # kola: { "exclusive": false }
 set -xeuo pipefail
 
-ok() {
-    echo "ok" "$@"
-}
-
-fatal() {
-    echo "$@" >&2
-    exit 1
-}
+. $KOLA_EXT_DATA/commonlib.sh
 
 # See https://github.com/coreos/coreos-assembler/pull/1786
 path=/usr/lib/systemd/system-generators/coreos-platform-chrony
--- a/tests/kola/chrony/data/commonlib.sh
+++ b/tests/kola/chrony/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh
--- a/tests/kola/containers/cgroups-v2
+++ b/tests/kola/containers/cgroups-v2
@@ -5,14 +5,7 @@
 # kola: { "distros": "fcos", "exclusive": false }
 set -xeuo pipefail
 
-ok() {
-    echo "ok" "$@"
-}
-
-fatal() {
-    echo "$@" >&2
-    exit 1
-}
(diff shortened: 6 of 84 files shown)


aaradhak:
@mike-nguyen Can you review this PR?

tests/common/lib.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
mike-nguyen: Maybe add a comment about what this file does and why it exists.

mike-nguyen (commented):
I noticed these only cover the tests in the files directory. There are tests in other directories that use the same `ok` and `fatal` functions.

tests/common/lib.sh
aaradhak: ok . I have added the comment

dustymabe:
The need for a symlink (like `../../../common/`) for each test is unfortunate. I wonder if we should expose a new environment variable from cosa like `$KOLA_EXT_CONFIG_TOPLEVEL` so we can drop the need to create the symlinks. @jlebon @mike-nguyen - WDYT?

dustymabe:
if a file exists at `tests/common/lib.sh` won't kola try to execute that as a test itself?

jlebon:
> I wonder if we should expose a new environment variable from cosa like `$KOLA_EXT_CONFIG_TOPLEVEL` so we can drop the need to create the symlinks.

The thing is that not all the repo is synced to the VM, just the test executable and anything in its `data/`. One thing we could do is to have it in `tests/kola/data` and teach kola to copy everything in the `data/` dirs of all its parent dirs instead of just its own dir.

> if a file exists at `tests/common/lib.sh` won't kola try to execute that as a test itself?

`kola` only looks for tests in `tests/kola`. We could also put it in `tests/kola/data` as above and mark it as non-executable (which is more correct anyway).

jlebon:
> One thing we could do is to have it in tests/kola/data and teach kola to copy everything in the data/ dirs of all its parent dirs instead of just its own dir.

I wanted to add, I don't think we should block on this. This is a really nice improvement already!

WDYT about dropping the executable bit and putting it in e.g. `tests/kola/common` for now?

dustymabe:
> WDYT about dropping the executable bit and putting it in e.g. `tests/kola/common` for now?

@jlebon Did you mean `tests/kola/data` here? In https://github.com/coreos/fedora-coreos-config/pull/1364#issuecomment-990086801 the directory you were using the `tests/kola/data` dir in your hypothetical example.

jlebon:
> > WDYT about dropping the executable bit and putting it in e.g. `tests/kola/common` for now?
> 
> @jlebon Did you mean `tests/kola/data` here? In [#1364 (comment)](https://github.com/coreos/fedora-coreos-config/pull/1364#issuecomment-990086801) the directory you were using the `tests/kola/data` dir in your hypothetical example.

Either should be fine for now until we implement the parent `data/` thing. I guess we can put it in `tests/kola/data` as prep for that.

dustymabe:
> Either should be fine for now until we implement the parent `data/` thing. I guess we can put it in `tests/kola/data` as prep for that.

That sounds good to me. Then IIUC all we'd need to do (once that change lands in COSA) is delete the symlinks.

dustymabe:
Thanks for working on this @aaradhak. Here's what I see now:

```
$ tree tests/kola/files/data/
tests/kola/files/data/
└── data -> ../../data/

1 directory, 0 files
```

Instead of having a second `data` dir (the symlink) under `$KOLA_EXT_DATA` would it be better to just symlink the file directly? Also might be good to rename the file something more obvious just if we do this. Here's a concrete suggestion:

```
$ mv tests/kola/data/lib.sh tests/kola/data/commonlib.sh 
$ rm tests/kola/files/data/data
$ ln -s ../../data/commonlib.sh tests/kola/files/data/commonlib.sh
```

Then the symlink looks like:

```
$ tree tests/kola/files/data/
tests/kola/files/data/
└── commonlib.sh -> ../../data/commonlib.sh

0 directories, 1 file
```

I think we probably also want to squash all of these commits down into a single commit before we do a final merge.

aaradhak:
Hi @dustymabe , the lib.sh have been renamed to commonlib.sh and have made the associated changes. Kindly review the changes.

tests/kola/data/commonlib.sh
```diff
@@ -0,0 +1,12 @@
+
+#!/bin/bash
+#This is a common library created for the ok and fatal function and symlinks added to the data/common in each directory
```
dustymabe:
```suggestion
#!/bin/bash

# This is a common library created for the ok and fatal function and
# symlinks added to the data/common in each directory
```

dustymabe (commented):
ok the code changes LGTM. There is one suggestion I made to delete a blank line at the beginning of commonlib.sh and break the comment into two lines.

We also need to squash all of these commits into one. Something like this:

```
tests: common library for ok() & fatal() functions

A separate common library is created for all external kola tests to use.
```

dustymabe (approved):
LGTM

PR #1367, opened 2021-12-09
delete ConditionKernelCommandLine=ostree

Rtoax:


bgilbert:
The entire `40ignition-ostree` Dracut module is currently designed for ostree systems, and those should always have that kernel argument.  The `Condition` is correct as far as I can tell.

xref https://github.com/coreos/fedora-coreos-tracker/issues/1039.  Please provide the requested information in that bug report so we can help you solve the problem you're having.

Rtoax:
You are right.

PR #1370, opened 2021-12-13
[rawhide] overrides: pin on systemd-249.7-2.fc36

dustymabe:
```
commit 391289c57d58fab8e39be688924b94ba661301e3
Date:   Mon Dec 13 16:35:43 2021 -0500

    overrides: pin on systemd-249.7-2.fc36
    
    systemd 250 breaks the ext.config.podman.dns and we need to determine
    an appropriate fix. Pin on systemd-249.7-2.fc36 for now.

commit 800f859e06f1c06da382e8396385ef9579775f8d
Date:   Mon Dec 13 16:32:19 2021 -0500

    tests: also redirect stderr in runascoreuser functions
    
    Otherwise we won't get stderr. This became obvious when investigating
    https://github.com/coreos/fedora-coreos-tracker/issues/1046 where
    running the test by hand yielded a lot more output (and the relevant
    error message).
```

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -14,3 +14,33 @@ packages:
     metadata:
       reason: 'https://github.com/coreos/fedora-coreos-tracker/issues/968'
       type: pin
+  systemd:
+    evr: 249.7-2.fc36
+    metadata:
+      reason: 'https://github.com/coreos/fedora-coreos-tracker/issues/1046'
+      type: pin
+  systemd-container:
+    evr: 249.7-2.fc36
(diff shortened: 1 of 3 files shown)


jlebon (approved):

cgwalters:
Hmm but did we file a bug or give either the systemd/podman people a heads up about this?

dustymabe:
> Hmm but did we file a bug or give either the systemd/podman people a heads up about this?

nope. https://github.com/coreos/fedora-coreos-tracker/issues/1046 is waiting on someone to investigate and do proper reporting.

PR #1372, opened 2021-12-14
manifests: add afterburn-dracut

sohankunkerkar:
The Fedora `afterburn` source package currently outputs an `afterburn` and `afterburn-dracut` subpackage. In the process of unifying spec files across Fedora/RHEL/C9S, we'd like to do the same for RHCOS. Before splitting, we need to make sure that RHCOS won't lose the dracut bits. So lower `afterburn-dracut` to a shared manifest.

### DIFF
--- a/manifests/system-configuration.yaml
+++ b/manifests/system-configuration.yaml
@@ -4,7 +4,7 @@
 
 packages:
   # Configuring SSH keys, cloud provider check-in, etc
-  - afterburn
+  - afterburn afterburn-dracut
   # NTP support
   - chrony
   # Installing CoreOS itself
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -91,8 +91,6 @@ postprocess:
 packages:
   # Security
   - polkit
-  # System setup
-  - afterburn-dracut
   # SSH
   - ssh-key-dir
   # Containers


dustymabe:
We should remove the `afterburn-dracut` entry from `fedora-coreos-base.yaml` if we do this.

manifests/system-configuration.yaml
```diff
@@ -5,6 +5,8 @@
 packages:
   # Configuring SSH keys, cloud provider check-in, etc
   - afterburn
+  # Adding this subpackage for RHCOS/RHEL
+  - afterburn-dracut
```
dustymabe: I wouldn't worry with an extra comment and just merge them onto a single line. 

```suggestion
  # Configuring SSH keys, cloud provider check-in, etc
  - afterburn afterburn-dracut
```

dustymabe:
The code changes LGTM. I'm still a bit confused about the commit message and what exact RPM packaging problem exists that this is helping solve. Can you give more detail on that?

jlebon:
For the commit message, how about something like:

> The Fedora `afterburn` source package currently outputs an `afterburn` and `afterburn-dracut` subpackage. In the process of unifying spec files across Fedora/RHEL/C9S, we'd like to do the same for RHCOS. Before splitting, we need to make sure that RHCOS won't lose the dracut bits. So lower `afterburn-dracut` to a shared manifest.
>
> An alternative is to have the `afterburn` package pull in `afterburn-dracut` which will avoid any changes for RHCOS, but in theory the `afterburn` subpackage could be useful without the dracut bits.

sohankunkerkar:
> The Fedora afterburn source package currently outputs an afterburn and afterburn-dracut subpackage. In the process of unifying spec files across Fedora/RHEL/C9S, we'd like to do the same for RHCOS. Before splitting, we need to make sure that RHCOS won't lose the dracut bits. So lower afterburn-dracut to a shared manifest.

@jlebon Thanks!  This message is an apt description as to why this change is required.

dustymabe:
Just wanted to point out that the second sentence in the suggestion didn't get added to the commit message.

dustymabe:
Either way I'm +1

dustymabe (dismissed):

dustymabe (approved):
LGTM

PR #1373, opened 2021-12-15
overlay/15fcos: coreos-check-ssh-keys: limit search to current boot

aaradhak:
fixed the problem by limiting the journal output to the most recent boot - jira #1033

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys
@@ -9,17 +9,20 @@ main() {
 
     # See https://github.com/coreos/ignition/pull/964 for the MESSAGE_ID 
     # source. It will track the authorized-ssh-keys entries in journald 
-    # provided via Ignition.
+    # provided via Ignition. Limit journal output to the most recent boot
+    # so we don't get output from re-used /var/ partitions.
     ignitionusers=$(
-            journalctl -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | \
+            journalctl -b 0 -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | \
             jq -r '.MESSAGE' | \
             xargs -I{} echo "Ignition: {}")
 
     # See https://github.com/coreos/afterburn/pull/397 for the MESSAGE_ID 
     # source. It will track the authorized-ssh-keys entries in journald  
-    # provided via Afterburn.
+    # provided via Afterburn.Limit journal output to the most recent boot
+    # so we don't get output from re-used /var/ partitions.
+
     afterburnusers=$(
-            journalctl -o json-pretty MESSAGE_ID=0f7d7a502f2d433caa1323440a6b4190 | \
+            journalctl -b 0 -o json-pretty MESSAGE_ID=0f7d7a502f2d433caa1323440a6b4190 | \
             jq -r '.MESSAGE' | \
             xargs -I{} echo "Afterburn: {}")
 


bgilbert:
Code LGTM.  This could probably use a comment explaining why we're explicitly checking only the current boot.

Similarly, let's summarize the problem being fixed in the commit message.  (Also, the correct link is https://github.com/coreos/fedora-coreos-tracker/issues/1033, and it's not a Jira card.  :slightly_smiling_face:)

dustymabe:
code LGTM as well. Agree with the points benjamin made. Also might want to change the commit title to something like:

```
overlay/15fcos: coreos-check-ssh-keys: limit search to current boot
```

aaradhak:
Hi @bgilbert @dustymabe , I have made the changed you had mentioned in the comments. Kindly provide your feedback.

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys
```diff
@@ -11,15 +11,15 @@ main() {
     # source. It will track the authorized-ssh-keys entries in journald 
     # provided via Ignition.
```
dustymabe:
```suggestion
    # source. It will track the authorized-ssh-keys entries in journald 
    # provided via Ignition. Limit journal output to the most recent boot
    # so we don't get output from re-used /var/ partitions. 
```

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys
```diff
@@ -11,15 +11,15 @@ main() {
...
     ignitionusers=$(
-            journalctl -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | \
+            journalctl -b 0 -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | \
             jq -r '.MESSAGE' | \
             xargs -I{} echo "Ignition: {}")
 
     # See https://github.com/coreos/afterburn/pull/397 for the MESSAGE_ID 
     # source. It will track the authorized-ssh-keys entries in journald  
     # provided via Afterburn.
```
dustymabe:
```suggestion
    # source. It will track the authorized-ssh-keys entries in journald  
    # provided via Afterburn. Limit journal output to the most recent boot
    # so we don't get output from re-used /var/ partitions. 
```

dustymabe (commented):
hey @aaradhak. The commit message looks great. A few improvements:

- remove the quotation marks in the commit title
- break up lines at around 70 to 80 characters
- the last line should be something like `Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1033`


Also as bgilbert said. It would be good if we added this context as a comment in the code. I made the suggestion inline in the code to show you what we mean.

overlay.d/15fcos/usr/libexec/coreos-check-ssh-keys
```diff
@@ -9,17 +9,20 @@ main() {
...
     ignitionusers=$(
-            journalctl -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | \
+            journalctl -b 0 -o json-pretty MESSAGE_ID=225067b87bbd4a0cb6ab151f82fa364b | \
             jq -r '.MESSAGE' | \
             xargs -I{} echo "Ignition: {}")
 
     # See https://github.com/coreos/afterburn/pull/397 for the MESSAGE_ID 
     # source. It will track the authorized-ssh-keys entries in journald  
-    # provided via Afterburn.
+    # provided via Afterburn.Limit journal output to the most recent boot
```
dustymabe, line 21:
```suggestion
    # provided via Afterburn. Limit journal output to the most recent boot
```

dustymabe:
LGTM - let's squash these down to a single commit.

aaradhak:
Yes, the commits are git squashed now. 

> LGTM - let's squash these down to a single commit.

dustymabe (approved):

PR #1374, opened 2021-12-15
05core: add coreos-network-initramfs-etc.service

jlebon:
A papercut right now is that when using Tang, one must use kernel
arguments to specify network configuration because network is required
on every boot.

This breaks the UX we're trying to emphasize, which is towards
specifying networking settings using NM keyfiles *once* and having it
apply everywhere.

Let's fix this by automatically detecting this case and enabling
`initramfs-etc` tracking of NM keyfiles on first boot. This will allow
users to stay on the NM keyfile path and have it transparently work even
when using Tang.

The new service does this from the real root. Long-term, I'd like to
investigate doing it earlier in the initramfs (this is related to
coreos/rpm-ostree#3006).

Modifications to NM keyfiles in the real root are automatically synced
on every new deployment, or explicitly when using
`rpm-ostree ex initramfs-etc --force-sync`.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -107,6 +107,9 @@ propagate_initramfs_networking() {
                 echo "info: propagating initramfs networking config to the real root"
                 cp -v /run/NetworkManager/system-connections/* /sysroot/etc/NetworkManager/system-connections/
                 coreos-relabel /etc/NetworkManager/system-connections/
+
+                mkdir -p /run/coreos
+                touch /run/coreos/network-propagated
             fi
         else
             echo "info: no initramfs networking information to propagate"
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.sh
@@ -21,6 +21,9 @@ copy_firstboot_network() {
     echo "info: copying files from ${src} to ${initramfs_network_dir}"
     mkdir -p ${initramfs_network_dir}
     cp -v ${src}/* ${initramfs_network_dir}/
+
+    mkdir -p /run/coreos
+    echo "${src}" > /run/coreos/firstboot-network-src
 }
 
 if ! is-live-image; then
--- a/overlay.d/05core/usr/lib/systemd/system/coreos-network-initramfs-etc.service
+++ b/overlay.d/05core/usr/lib/systemd/system/coreos-network-initramfs-etc.service
@@ -1,0 +1,26 @@
+[Unit]
+Description=CoreOS Enable Network Initramfs-Etc
+# All services which use ConditionFirstBoot=yes should use
+# Before=first-boot-complete.target, which is a target that
+# was introduced in https://github.com/systemd/systemd/issues/4511
+# and hasn't propagated everywhere yet. Once the target propagates
+# everywhere, we can drop the systemd-machine-id-commit.service
+# from the Before= line.
+Before=first-boot-complete.target systemd-machine-id-commit.service
+Wants=first-boot-complete.target
+# If there are custom NM connection files,
+ConditionDirectoryNotEmpty=/etc/NetworkManager/system-connections/
+# and they come from `--copy-network` or `iso network embed`,
+ConditionPathExists=/run/coreos/firstboot-network-src
+ConditionPathExists=/run/coreos/network-propagated
+# and it's the first boot,
+ConditionFirstBoot=true
+
+# then automatically turn on initramfs-etc tracking of the NM configs.
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-network-initramfs-etc
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/05core/usr/libexec/coreos-network-initramfs-etc
+++ b/overlay.d/05core/usr/libexec/coreos-network-initramfs-etc
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+# Would've been really nice to be able to use
+# `ConditionKernelCommandLine=rd.neednet=1` for this, but that doesn't work
+# because the current boot doesn't have it; the *next* boot will.
+
+karg() {
+    local name="$1" value="${2:-}"
+    local cmdline=( $(rpm-ostree kargs) )
+    for arg in "${cmdline[@]}"; do
+        if [[ "${arg%%=*}" == "${name}" ]]; then
+            value="${arg#*=}"
+        fi
+    done
+    echo "${value}"
+}
+
+if [[ $(karg rd.neednet) == 1 ]]; then
+    rpm-ostree ex initramfs-etc --track /etc/NetworkManager/system-connections
+fi
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -25,3 +25,5 @@ enable rtas_errd.service
 enable clevis-luks-askpass.path
 # Provide information if no ignition is provided
 enable coreos-check-ignition-config.service
+# Automatically enables `initramfs-etc` tracking of NM keyfiles
+enable coreos-network-initramfs-etc.service


jlebon:
Note this is softer than the proposal in https://github.com/coreos/fedora-coreos-tracker/issues/886, which is asking for doing this even if the network configuration comes from kernel arguments. The problem with that though is that we need to be able to distinguish between first boot kernel arguments (i.e. the auto-forwarding magic in `coreos-installer.service`) and persistent kernel arguments (i.e. `--append-karg`). Because in the latter case, I think it'd be too confusing to have *both* kernel arguments and NM keyfiles in the initramfs.

So this is optimized for the NM keyfile path, which is what we want to emphasize.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -107,6 +107,9 @@ propagate_initramfs_networking() {
                 echo "info: propagating initramfs networking config to the real root"
                 cp -v /run/NetworkManager/system-connections/* /sysroot/etc/NetworkManager/system-connections/
                 coreos-relabel /etc/NetworkManager/system-connections/
+
+                mkdir -p /run/coreos
+                touch /run/coreos/network-propagated
```
dustymabe, line 112: If we keep this I think we should add the context from the commit message in a comment

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.sh
```diff
@@ -21,6 +21,9 @@ copy_firstboot_network() {
     echo "info: copying files from ${src} to ${initramfs_network_dir}"
     mkdir -p ${initramfs_network_dir}
     cp -v ${src}/* ${initramfs_network_dir}/
+
+    mkdir -p /run/coreos
+    echo "${src}" > /run/coreos/firstboot-network-src
```
dustymabe, line 26: If we keep this I think we should add the context from the commit message in a comment

overlay.d/05core/usr/libexec/coreos-network-initramfs-etc
```diff
@@ -0,0 +1,21 @@
...
+        if [[ "${arg%%=*}" == "${name}" ]]; then
+            value="${arg#*=}"
+        fi
+    done
+    echo "${value}"
+}
+
+if [[ $(karg rd.neednet) == 1 ]]; then
+    rpm-ostree ex initramfs-etc --track /etc/NetworkManager/system-connections
+fi
```
dustymabe, line 21: I feel like where this gets ugly and we need to be a lot more careful about enabling it. For example, what if someone has `rd.neednet=1` and also `ip=xyz`? Now our initramfs `/etc/NetworkManager/system-connections` connections are battling against the ones created by `nm-initrd-generator`. Also, what if there are no files in `/etc/NetworkManager/system-connections` because the system just uses DHCP? Or what if someone haphazardly added `--append-kargs=rd.neednet` and isn't using tang?

Also I think `rd.neednet` and `rd.neednet=1` are equivalent. 

Sorry I just have a lot of open questions right now.

dustymabe: I guess the 2nd question is taken care of by `ConditionDirectoryNotEmpty=/etc/NetworkManager/system-connections/`.

overlay.d/05core/usr/lib/systemd/system/coreos-network-initramfs-etc.service
```diff
@@ -0,0 +1,26 @@
...
+# and hasn't propagated everywhere yet. Once the target propagates
+# everywhere, we can drop the systemd-machine-id-commit.service
+# from the Before= line.
+Before=first-boot-complete.target systemd-machine-id-commit.service
+Wants=first-boot-complete.target
+# If there are custom NM connection files,
+ConditionDirectoryNotEmpty=/etc/NetworkManager/system-connections/
+# and they come from `--copy-network` or `iso network embed`,
+ConditionPathExists=/run/coreos/firstboot-network-src
+ConditionPathExists=/run/coreos/network-propagated
```
dustymabe, line 15: It feels weird to require these. Does it matter where the files came from as long as they end up in `/etc/NetworkManager/system-connections/`?

dustymabe (commented):
I've given this an initial run through. Obviously I want to think critically about any additions to our the networking "house of cards" we've built so give me some time to see if I think of anything else. For now here are some questions/comments.

overlay.d/05core/usr/lib/systemd/system/coreos-network-initramfs-etc.service
bgilbert: AIUI: if they came from Ignition, then they weren't needed in the initrd during first boot, so they shouldn't be needed in the initrd in later boots.  If they came from kargs, then the disposition of those kargs should govern: either they'll exist on every boot, or they won't and we shouldn't propagate their effects forward.

I guess the `network-propagated` condition is for the corner case where the user provided some NM configs, but they looked default-y so we didn't copy them forward?  If so, let's mention that.

overlay.d/05core/usr/libexec/coreos-network-initramfs-etc
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+# Would've been really nice to be able to use
+# `ConditionKernelCommandLine=rd.neednet=1` for this, but that doesn't work
+# because the current boot doesn't have it; the *next* boot will.
+
+karg() {
+    local name="$1" value="${2:-}"
+    local cmdline=( $(rpm-ostree kargs) )
```
bgilbert, line 10: We paste this function all over the place, so it might be worth calling out that we're not reading `/proc/cmdline` here.

overlay.d/05core/usr/libexec/coreos-network-initramfs-etc
bgilbert: I think the answer to the third question is "then they want networking in the initrd, and they'll get it".

For the first question, if they didn't supply firstboot NM keyfiles, the unit conditions will ensure we'll never get here.  If they did, _and_ they specified `ip=`, ...do we need to handle that case?  Or can we just say "don't do that"?

jlebon: > For example, what if someone has `rd.neednet=1` and also `ip=xyz`? Now our initramfs `/etc/NetworkManager/system-connections` connections are battling against the ones created by `nm-initrd-generator`.

Yeah, indeed. But worth noting that any network kargs will not have taken effect on first boot at least because our current stance is that firstboot NM keyfiles win. The only use case I can think of where they'd use both network kargs *and* NM keyfiles is if they want the *same* configuration between the initrd and the real root on first boot, but not on subsequent boots... which would be weird. I guess one example is they add a NIC after install and they want to e.g. make sure that it stays off in the initrd? Anyway, high-level, yes if the kargs describe a configuration for devices that overlap with the `/etc` configuration, then there could be conflicts. I'm not sure if NM will let the `/run` one win, but I feel OK just documenting that you should stop tracking NM keyfiles if you fall in that category.

> Or what if someone haphazardly added --append-kargs=rd.neednet and isn't using tang?

Essentially, what this is proposing is setting a new default behaviour (which purposely isn't specific to Tang): if the system needs networking in the initramfs, and you specified NM keyfiles, then we assume that you want that configuration to apply on not just the first boot, but every boot. It seems like a reasonable assumption (and IMO [less surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) than not doing it). You can disable this default by switching only to kargs, or disabling the service (at install time), or using `rpm-ostree initramfs-etc --untrack` (for later boots).

Related to this, in the future I'd like to also add support to `rpm-ostree initramfs-etc` to track files that aren't necessarily in the same directory, so e.g. they could do `--track /etc/NetworkManager/system-connections-initrd:/etc/NetworkManager/system-connections`. And then there'd be even less need for network kargs.

jlebon:
Spoke to @dustymabe about this. Some concerns were raised around the conditionals in place. Some things we discussed:
- We should try to cover the magic kargs auto-forwarding case too. To do this, we could have `coreos-installer-service` add another firstboot karg to signal that forwarding occurred and we can detect that.
- We shouldn't care where the NM keyfiles came from firstboot kargs or NM keyfiles or target Ignition config proper, so we should be able to drop the conditional on `/run/coreos/network-propagated`.
- If users use both `--copy-network` and `--append-karg` and use Tang, then we'll turn on `initramfs-etc`, but the persistent kargs will also be there. It's kind of a weird scenario we don't expect users to use because the persistent kargs will have no effect on first boot. But if they do (could imagine this as a day-2 operation), they will be there on subsequent boots where they may conflict with the overlayed NM configs. We should try to see if there's an elegant way to address this (which ideally doesn't involve detecting network kargs again like we do in `coreos-installer-service`). This might also lead to less conditionals.

jlebon:
I'll close this for now. The suboptimal status quo is described in the still-open https://github.com/coreos/fedora-coreos-tracker/issues/886 but there's a real cost too to adding yet more complexity to our networking code.

PR #1375, opened 2021-12-16
tests/kola: fix `platforms` key for `coreos-platform-chrony-generator`

jlebon:
It's a space-separated list, not comma-separated.

This... may have caused us to never actually run that test. Though it's
possible also something changed in cosa, but I'm not seeing it right
now.

Fixes: b73a3757a8d9 ("20platform-chrony: Add support for AWS")

### DIFF
--- a/tests/kola/chrony/coreos-platform-chrony-generator
+++ b/tests/kola/chrony/coreos-platform-chrony-generator
@@ -2,7 +2,7 @@
 set -euo pipefail
 # Test the coreos-platform-chrony generator.
 
-# kola: { "exclusive": false, "platforms": "aws,azure,gcp" }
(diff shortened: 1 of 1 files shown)


jlebon:
I think we should have kola explicitly error out if a platform or architecture doesn't exist after splitting.

dustymabe:
> I think we should have kola explicitly error out if a platform or architecture doesn't exist after splitting.

👍

dustymabe (approved):
LGTM

dustymabe:
Now we get to see if the test passes!

dustymabe:
I feel like this is one where we should skip CI especially since this test is limited to cloud platforms. I'll test at least on Azure real quick.

PR #1376, opened 2021-12-16
[rawhide] manifest: pull in authselect

jlebon:
This is required now as per:
https://fedoraproject.org/wiki/Changes/Make_Authselect_Mandatory
https://src.fedoraproject.org/rpms/glibc/c/cadee80b1316?branch=rawhide
https://bugzilla.redhat.com/show_bug.cgi?id=2023741

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -14,6 +14,10 @@ repos:
 add-commit-metadata:
   fedora-coreos.stream: rawhide
 
+packages:
+  # for nsswitch.conf
+  - authselect
+
 postprocess:
   # Disable Zincati and fedora-coreos-pinger on non-production streams
   # https://github.com/coreos/fedora-coreos-tracker/issues/163


jlebon:
There's some history with authselect and OSTree-based systems (see https://github.com/authselect/authselect/issues/48), though I think those have been ironed out now. If not, this will be a good opportunity to find out and get things fixed while it's still in rawhide.

travier:
Weird, I already have it pulled in my builds:
```
   authselect-1.3.0-3.fc36.x86_64 (fedora-rawhide)
```

jlebon:
Hmm, how though? It looks like it conflicts with older systemd. Or did you revert https://github.com/coreos/fedora-coreos-config/pull/1370?

travier:
> Or did you revert #1370?

Yes

jlebon:
Looking at the spec file, something might be causing the logic around the "force file" handling to not kick in correctly:
- https://src.fedoraproject.org/rpms/authselect/blob/rawhide/f/authselect.spec#_267
- https://src.fedoraproject.org/rpms/authselect/blob/rawhide/f/authselect.spec#_283
- https://src.fedoraproject.org/rpms/authselect/blob/rawhide/f/authselect.spec#_326

Do you want to try digging into that?

dustymabe:
obsoleted by https://github.com/coreos/fedora-coreos-config/pull/1385

PR #1377, opened 2021-12-16
tests: networking.nic-naming: disable on Azure platform

dustymabe:
Disable on azure because of a limitation of the hv_netvsc driver
there. According to [1] and [2] the driver does not provide sufficient
information to systemd-udev for systemd's naming to work. It falls
back to basic ethX naming.

[1] https://access.redhat.com/solutions/3204751
[2] https://github.com/Azure/WALinuxAgent/issues/1877


miabbott (approved):
Great documentation for the change!

PR #1380, opened 2021-12-17
[stable] tree: promote changes from testing at 48cf602

coreosbot-releng:
Triggered by @saqibali-2k in https://github.com/coreos/fedora-coreos-streams/issues/423.


dustymabe:
Let's add another commit on top similar to 179a4c4 for the `manifest.yaml`.

dustymabe (approved):
LGTM

PR #1384, opened 2021-12-20
networking: add test scripts for force-persist-ip

HuijingHei:
This is to test with and without coreos.force_persist_ip.

Setup configuration for a single NIC with two different ways:

- karg ip=x provides static network config for ens5 and (or not)
  coreos.force_persist_ip
- ignition provides dhcp network config for ens5

Expected result:
- With coreos.force_persist_ip, ip=kargs wins, ens5 gets the static IP address via kargs
- Without coreos.force_persist_ip, Ignition networking configuration wins,
  ens5 gets ip via dhcp
See https://bugzilla.redhat.com/show_bug.cgi?id=1958930#c29
    
commonlib.sh: add func get_ip_for_nic
    This is to add func get_ip_for_nic to get ip for the specified nic

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -11,3 +11,13 @@ fatal() {
     echo "$@" >&2
     exit 1
 }
+
+get_ip_for_nic () {
+    local nic_name=$1
+    nic_ip=""
+    nic_ip=`ip -j addr show ${nic_name} | jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local'`
+    if [ -z "$nic_ip" ]; then
+        echo "Error: can not get ${nic_name} ip"
+        exit 1
+    fi
+}
--- a/tests/kola/networking/force-persist-ip/config.bu
+++ b/tests/kola/networking/force-persist-ip/config.bu
@@ -1,0 +1,16 @@
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/NetworkManager/system-connections/ens5.nmconnection
+      mode: 0600
+      contents:
+        inline: |
+          [connection]
+          id=ens5
+          type=ethernet
+          interface-name=ens5
+          [ipv4]
+          dns-search=
+          may-fail=false
+          method=auto
--- a/tests/kola/networking/force-persist-ip/data/commonlib.sh
+++ b/tests/kola/networking/force-persist-ip/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../../data/commonlib.sh
--- a/tests/kola/networking/force-persist-ip/test.sh
+++ b/tests/kola/networking/force-persist-ip/test.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# Setup configuration for a single NIC with two different ways:
+# - kargs provide static network config for ens5 and also coreos.force_persist_ip
+# - ignition provides dhcp network config for ens5
+# Expected result:
+# - with coreos.force_persist_ip ip=kargs win, verify that 
+#   ens5 has the static IP address via kargs
+
+# https://bugzilla.redhat.com/show_bug.cgi?id=1958930#c29
+# kola: { "platforms": "qemu", "additionalNics": 1, "appendKernelArgs": "ip=10.10.10.10::10.10.10.1:255.255.255.0:myhostname:ens5:none:8.8.8.8 coreos.force_persist_ip"}
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# Verify ens5 get staic ip via kargs
+nic_name="ens5"
+nic_ip=""
+get_ip_for_nic ${nic_name}
+if [ ${nic_ip} != "10.10.10.10" ]; then
+    fatal "Error: get ${nic_name} ip = ${nic_ip}, expected is 10.10.10.10"
+fi
+ok "get ${nic_name} ip is 10.10.10.10"
+
+syscon="/etc/NetworkManager/system-connections"
+if [ ! -f "${syscon}/${nic_name}.nmconnection" ]; then
+    fatal "Error: can not find ${syscon}/${nic_name}.nmconnection"
+fi
+ok "find ${syscon}/${nic_name}.nmconnection"
+
+# Verify logs
+if ! journalctl -b 0 -u coreos-teardown-initramfs | \
+     grep -q "info: coreos.force_persist_ip detected: will force network config propagation"; then
+  fatal "Error: force network config propagation not work"
+fi
+ok "force network config propagation"
--- a/tests/kola/networking/prefer-ignition-networking/config.bu
+++ b/tests/kola/networking/prefer-ignition-networking/config.bu
@@ -1,0 +1,16 @@
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/NetworkManager/system-connections/ens5.nmconnection
+      mode: 0600
+      contents:
+        inline: |
+          [connection]
+          id=ens5
+          type=ethernet
+          interface-name=ens5
+          [ipv4]
+          dns-search=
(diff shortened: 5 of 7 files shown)


tests/kola/data/commonlib.sh
```diff
@@ -11,3 +11,13 @@ fatal() {
     echo "$@" >&2
     exit 1
 }
+
+getip () {
```
dustymabe: Optional improvement. Since we are passing in a NIC name as an argument we could change the function name to be more descriptive. For example:

```suggestion
get_ip_for_nic () {
```

tests/kola/data/commonlib.sh
```diff
@@ -11,3 +11,13 @@ fatal() {
...
+    local nic=$1
+    nic_ip=""
+    nic_ip=`ip -j addr show $nic | jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local'`
+    if [ -z "$nic_ip" ]; then
+        echo "Error: can not get ${nic} ip"
+        exit 1
+    fi
+}
```
dustymabe, line 23: Maybe let's make the change to this file a separate commit.

tests/kola/networking/force-persist-ip-with-static-and-kargs/data
```diff
@@ -0,0 +1 @@
+../../data
```
dustymabe: This isn't consistent with how it is done for other tests. For the other tests the `data` is not a symlink, but a directory and there is a `commonlib.sh` symlink inside the directory. For example:

```
$ git rm tests/kola/networking/force-persist-ip-with-static-and-kargs/data
rm 'tests/kola/networking/force-persist-ip-with-static-and-kargs/data'
$ mkdir tests/kola/networking/force-persist-ip-with-static-and-kargs/data 
$ ln -s ../../data/commonlib.sh tests/kola/networking/force-persist-ip-with-static-and-kargs/data/commonlib.sh 
$ ls -l tests/kola/networking/force-persist-ip-with-static-and-kargs/data/commonlib.sh
lrwxrwxrwx. 1 dustymabe dustymabe 23 Dec 21 10:00 tests/kola/networking/force-persist-ip-with-static-and-kargs/data/commonlib.sh -> ../../data/commonlib.sh
$ git add tests/kola/networking/force-persist-ip-with-static-and-kargs/data/commonlib.sh
```

tests/kola/networking/no-persist-ip-with-static-and-kargs/data
```diff
@@ -0,0 +1 @@
+../../data
```
dustymabe: Same symlink issue here.

tests/kola/networking/force-persist-ip-with-static-and-kargs/config.bu
```diff
@@ -0,0 +1,24 @@
...
+          dhcp-hostname=myhostname
+          dns=8.8.8.8;
+          dns-search=
+          may-fail=false
+          method=manual
+
+kernel_arguments:
+  should_exist:
+    - ip=ens5:dhcp
+    - coreos.force_persist_ip
```
dustymabe: This works, but it's not in the spirit of what we want to test here. 90% of the time people are passing the kernel arguments so they can configure networking in the initramfs and then use the network to retrieve the ignition config. In order to test things more like users will use it we need to make the machine boot with the kernel arguments on first boot.

What we need to do here is plumb through exttest support for [AppendKernelArguments](https://github.com/coreos/coreos-assembler/blob/19531a6597eb4fc27beefea230002545e9a7cb11/mantle/platform/qemu.go#L406-L407).

and then have something like:

```
kola: { "platforms": "qemu", "additionalNics": 2, "appendKernelArguments": "ip=ens5:dhcp coreos.force_persist_ip"}
```

tests/kola/networking/force-persist-ip-with-static-and-kargs/test.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# https://bugzilla.redhat.com/show_bug.cgi?id=1958930#c29
+# kola: { "platforms": "qemu", "additionalNics": 2 }
```
dustymabe:
```suggestion
# kola: { "platforms": "qemu", "additionalNics": 1 }
```

Usually the way I would test `coreos.force_persist_ip` is by providing configuration for a single NIC two different ways and seeing which one wins. For example:

- kargs provide static network config for ens5 and also `coreos.force_persist_ip` 
- ignition provides dhcp network config for ens5

Now verify that `ens5` has the static IP address.

Thus we should only need one additional NIC.

tests/kola/networking/no-persist-ip-with-static-and-kargs/test.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# https://bugzilla.redhat.com/show_bug.cgi?id=1958930#c29
+# kola: { "platforms": "qemu", "additionalNics": 2 }
```
dustymabe:
```suggestion
# kola: { "platforms": "qemu", "additionalNics": 1}
```

For this test we're just testing that Ignition is the preferred vehicle for networking configuration, correct? Let's also use a single additional NIC here and provide configuration both ways for it (similar to the other test). It focuses the test and makes these tests a bit more simple.

dustymabe (commented):
Good first pass @HuijingHei! This is going to be a foundation for a much more comprehensive set of automation for network testing. 

To me it looks like with these two tests we're trying to test with and without `coreos.force_persist_ip` and make sure the behavior is what we expect, which is

- with `coreos.force_persist_ip` ip= kargs win
- without `coreos.force_persist_ip` Ignition networking configuration wins

Let's rename the tests:

- `force-persist-ip-with-static-and-kargs` -> `force-persist-ip`
- `no-persist-ip-with-static-and-kargs` -> `prefer-ignition-networking`

And maybe we can add a comment at the top of each test that explains what we expect a bit more.

tests/kola/networking/force-persist-ip-with-static-and-kargs/config.bu
dustymabe: Note that this is additional feature work and I'm sorry about bringing up something else new: https://github.com/coreos/coreos-assembler/issues/2621

dustymabe: Do you want to review: https://github.com/coreos/coreos-assembler/pull/2622 ?

tests/kola/networking/force-persist-ip-with-static-and-kargs/test.sh
HuijingHei: Thanks for your detailed reply! 
Test with `cosa kola qemuexec -U --additional-nics 1 --qemu-image test1.qcow2 --kargs ip=10.10.10.10::10.10.10.1:255.255.255.0:myhostname:ens5:none:8.8.8.8 --kargs coreos.force_persist_ip --ignition ./net.ign --ignition-direct`, net.ign includes dhcp config for ens5, get result:

- ens5 get static ip 10.10.10.10 (with `coreos.force_persist_ip` ip= kargs win)

tests/kola/networking/no-persist-ip-with-static-and-kargs/test.sh
HuijingHei: Test with `cosa kola qemuexec -U --additional-nics 1 --qemu-image test1.qcow2 --kargs ip=10.10.10.10::10.10.10.1:255.255.255.0:myhostname:ens5:none:8.8.8.8 --ignition ./net.ign --ignition-direct`, net.ign includes dhcp config for ens5, get result:

- ens5 get dhcp ip 10.0.2.31(without `coreos.force_persist_ip` Ignition networking configuration wins)

miabbott (approved):
This looks great! Awesome job, Huijing!

PR #1385, opened 2021-12-20
[rawhide] unpin systemd; run authselect on first boot

dustymabe:
```
commit 7ec35dbd959bce1664ccaec6c23525831e1df74f
Date:   Mon Dec 20 10:21:44 2021 -0500

    manifest: run authselect on first boot to workaround bug
    
    The ownership of nsswitch.conf has moved to authselect in F36+
    and right now the file is created by running authselect in a
    %posttrans, which fails. We need to work around this problem.
    This workaround is twofold:
    
      - Manually input nsswitch.conf during the build process so
        that rpm-ostree won't fail when it tries to inject altfiles.
      - Run `authselect` on first boot to create all the files it needs
        and re-add altfiles entries to newly created nsswitch.conf.
    
    Workaround for https://github.com/coreos/fedora-coreos-tracker/issues/1051

commit 5a2be4d864fa1c39fa6685fea3a38bac5c94930d
Date:   Thu Dec 16 11:44:04 2021 -0500

    manifest: pull in authselect
    
    This is required now as per:
    https://fedoraproject.org/wiki/Changes/Make_Authselect_Mandatory
    https://src.fedoraproject.org/rpms/glibc/c/cadee80b1316?branch=rawhide
    https://bugzilla.redhat.com/show_bug.cgi?id=2023741

commit 9e75411667e61c9940d567e1438ca92e28fbf144
Date:   Mon Dec 20 10:20:04 2021 -0500

    overrides: unpin systemd
    
    It turns out too many things are being built against the new
    systemd for us to effectively pin it. We'll snooze the test
    (ext.config.podman.dns) instead while we investigate the fix.
    
    See https://github.com/coreos/fedora-coreos-tracker/issues/1046
```

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -14,6 +14,10 @@ repos:
 add-commit-metadata:
   fedora-coreos.stream: rawhide
 
+packages:
+  # for nsswitch.conf
+  - authselect
+
 postprocess:
   # Disable Zincati and fedora-coreos-pinger on non-production streams
   # https://github.com/coreos/fedora-coreos-tracker/issues/163
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -27,3 +27,50 @@ postprocess:
     mkdir -p /etc/fedora-coreos-pinger/config.d /etc/zincati/config.d
     echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/163\nreporting.enabled = false' > /etc/fedora-coreos-pinger/config.d/90-disable-on-non-production-stream.toml
     echo -e '# https://github.com/coreos/fedora-coreos-tracker/issues/163\nupdates.enabled = false' > /etc/zincati/config.d/90-disable-on-non-production-stream.toml
+
+  # The ownership of nsswitch.conf has moved to authselect in F36+
+  # and right now the file is created by running authselect in a
+  # %posttrans, which fails. We need to work around this problem.
+  # This workaround is twofold:
+  #
+  #   - Manually input nsswitch.conf during the build process so
+  #     that rpm-ostree won't fail when it tries to inject altfiles.
+  #   - Run `authselect` on first boot to create all the files it needs
+  #     and re-add altfiles entries to newly created nsswitch.conf.
+  #
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1051
+  - |
+    #!/usr/bin/bash
+    set -xeuo pipefail
+    cat <<'EOF' > /etc/nsswitch.conf
+    # In order of likelihood of use to accelerate lookup.
+    passwd:     files sss systemd
+    shadow:     files
+    group:      files sss systemd
+    hosts:      files resolve [!UNAVAIL=return] myhostname dns
+    services:   files sss
+    netgroup:   files sss
+    automount:  files sss
+    
+    aliases:    files
+    ethers:     files
+    gshadow:    files
+    networks:   files dns
+    protocols:  files
+    publickey:  files
+    rpc:        files
+    EOF
+    cat <<'EOF' > /usr/lib/systemd/system/rpm-ostree-authselect.service
+      [Unit]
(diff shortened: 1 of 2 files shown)


dustymabe:
- requires https://github.com/coreos/fedora-coreos-config/pull/1386
- obsoletes https://github.com/coreos/fedora-coreos-config/pull/1376

manifest.yaml
```diff
@@ -23,3 +27,81 @@ postprocess:
...
+    # To use 'db', install the appropriate package(s) (provide 'makedb' and
+    # libnss_db.so.*), and place the 'db' in front of 'files' for entries
+    # you want to be looked up first in the databases, like this:
+    #
+    # passwd:    db files
+    # shadow:    db files
+    # group:     db files
+    
+    # In order of likelihood of use to accelerate lookup.
+    passwd:     sss files systemd
```
cgwalters: You're missing `altfiles`

dustymabe: discussed with @cgwalters in IRC. This is an intermediate file. RPM-OSTree itself will inject the `altfiles` entries.

dustymabe:
With just the nsswitch.conf I still couldn't get into the system. Running `single` on the kernel command line allowed me to see what was keeping me from logging in on a previous boot:

```
Dec 20 17:06:00 localhost login[1118]: PAM _pam_load_conf_file: unable to open config for system-auth
Dec 20 17:06:00 localhost login[1118]: PAM _pam_load_conf_file: unable to open config for postlogin
Dec 20 17:06:00 localhost login[1118]: PAM _pam_load_conf_file: unable to open config for system-auth
Dec 20 17:06:00 localhost login[1118]: PAM _pam_load_conf_file: unable to open config for system-auth
Dec 20 17:06:00 localhost login[1118]: PAM _pam_load_conf_file: unable to open config for system-auth
Dec 20 17:06:00 localhost login[1118]: PAM _pam_load_conf_file: unable to open config for postlogin
Dec 20 17:06:00 localhost audit[1118]: USER_ACCT pid=1118 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 msg='op=PAM:accounting grantors=? acct="core" exe="/usr/bin/login" hostname=localhost addr=? terminal=/dev/ttyS0 res=failed'
```

So the PAM stack is loading up more files that are created from running `authselect`. I tried figuring out which files and splicing them in the same way we're doing `nsswitch.conf` but I couldn't find them all quickly. Let's just run `authselect` on first boot to workaround this issue for now.

cgwalters (approved):
If it works I have no problem with shipping this - but just for now.  I think fixing this truly properly should actively block the Fedora 36 release; the https://bugzilla.redhat.com/show_bug.cgi?id=2034360 bug is annotated as such.  Or the authselect change should be backed out.

manifest.yaml
```diff
@@ -23,3 +27,50 @@ postprocess:
...
+    cat <<'EOF' > /usr/lib/systemd/system/rpm-ostree-authselect.service
+      [Unit]
+      # https://github.com/coreos/fedora-coreos-tracker/issues/1051
+      Description=Run Authselect on RPM-OSTree System to Workaround Issue
+      ConditionPathExists=!/etc/authselect/nsswitch.conf
+      [Service]
+      Type=oneshot
+      RemainAfterExit=yes
+      ExecStart=/usr/bin/authselect select sssd with-silent-lastlog --force
+      ExecStartPost=sed -i -E "s/^(passwd|group):(.*?)files/\1:\2files altfiles/" /etc/nsswitch.conf
```
jlebon, line 72: Did you try without this `ExecStartPost` BTW? authselect *should* know to keep altfiles on rpm-ostree systems.

dustymabe: I *think* so but have lost all memory of things that happened before the holidays. I'm hoping to rip this out soon anyway once we get a proper fix.

PR #1388, opened 2021-12-21
var-mount: verify udev rules /dev/disk/by-id/scsi-* symlinks present in initramfs

HuijingHei:
Add test script for https://bugzilla.redhat.com/show_bug.cgi?id=1990506

### DIFF
--- a/tests/kola/var-mount/scsi-id/config.bu
+++ b/tests/kola/var-mount/scsi-id/config.bu
@@ -1,0 +1,16 @@
+variant: fcos
+version: 1.4.0
+storage:
+  disks:
+    - device: /dev/disk/by-id/scsi-0NVME_VirtualMultipath_disk1
+      wipe_table: true
+      partitions:
+        - number: 1
+          label: var
+  filesystems:
+    - path: /var
+      device: /dev/disk/by-partlabel/var
+      format: xfs
+      wipe_filesystem: true
+      label: var
+      with_mount_unit: true
--- a/tests/kola/var-mount/scsi-id/test.sh
+++ b/tests/kola/var-mount/scsi-id/test.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+set -xeuo pipefail
+# This is to verify udev rules /dev/disk/by-id/scsi-* 
+# symlinks present in initramfs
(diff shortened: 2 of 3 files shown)


HuijingHei:
For scsi disk I use `"additionalDisks": ["5G:mpath"]` flag, but it also enables the multipath.

tests/kola/var-mount/scsi-id/config.bu
```diff
@@ -0,0 +1,27 @@
...
+      wipe_table: true
+      partitions:
+        - number: 1
+          label: var
+  filesystems:
+    - path: /var
+      device: /dev/disk/by-partlabel/var
+      format: xfs
+      wipe_filesystem: true
+      label: var
```
dustymabe, line 15: can we use [`with_mount_unit`](https://coreos.github.io/butane/examples/#filesystems-and-partitions) here and skip the systemd section below?

tests/kola/var-mount/scsi-id/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -xeuo pipefail
+# This is to verify udev rules /dev/disk/by-id/scsi-* 
+# symlinks present in initramfs
+# https://bugzilla.redhat.com/show_bug.cgi?id=1990506
+
+# kola: {"platforms": "qemu", "additionalDisks": ["5G:mpath"]}
+
+fstype=$(findmnt -nvr /var -o FSTYPE)
+[[ $fstype == xfs ]]
```
dustymabe: should we add another check of running `lsinitrd /path/to/initramfs` and making sure the files we expect to see are in the output?

tests/kola/var-mount/scsi-id/config.bu
HuijingHei: Good suggestion, updated

tests/kola/var-mount/scsi-id/test.sh
HuijingHei: That makes sense, add check `61-scsi-sg3_id.rules` and `63-scsi-sg3_symlink.rules` in initramfs

miabbott (approved):
LGTM

PR #1389, opened 2021-12-21
denylist: snooze ostree.hotfix podman.base tests on rawhide stream

dustymabe:
The ostree.hotfix tests rears it's ugly head a few weeks a month
and the podman.base test is surfacing kernel locking dep warnings.
Since no one is going to dig into these for the next few weeks let's
snooze them in the hopes we can finally get a new aarch64 rawhide
build.


saqibali-2k (approved):
looks good!

PR #1390, opened 2021-12-22
chrony: use /dev/ptp_hyperv on AzureStack

freedge:
As implemented on Azure (d0ad4c8d78b883b6324436d0eeb7b4597e0a4605)
/dev/ptp_hyperv should be used as well on AzureStack.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1053


cgwalters (approved):
Thanks, LGTM!

PR #1393, opened 2021-12-30
Re-enable kdump test for all platforms

gursewak1997:
Related to https://github.com/coreos/fedora-coreos-tracker/issues/860
Fix for kdump failure on online platforms: https://github.com/coreos/coreos-assembler/pull/2639

### DIFF
--- a/tests/kola/kdump/crash
+++ b/tests/kola/kdump/crash
@@ -1,8 +1,6 @@
 #!/bin/bash
-# kola: {"platforms": "qemu-unpriv", "minMemory": 4096, "tags": "skip-base-checks", "architectures": "x86_64"}
+# kola: {"minMemory": 4096, "tags": "skip-base-checks"}
 # https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/
-# Only run on QEMU x86_64 for now:
-# https://github.com/coreos/fedora-coreos-tracker/issues/860
 
 set -xeuo pipefail
 


tests/kola/kdump/crash
```diff
@@ -1,9 +1,7 @@
 #!/bin/bash
 set -xeuo pipefail
 # https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/
-# Only run on QEMU x86_64 for now:
-# https://github.com/coreos/fedora-coreos-tracker/issues/860
-# kola: {"platforms": "qemu-unpriv", "minMemory": 4096, "tags": "skip-base-checks", "architectures": "x86_64"}
+# kola: {"platforms": "qemu-unpriv aws azure gcp", "minMemory": 4096, "tags": "skip-base-checks", "architectures": "x86_64 aarch64"}
```
miabbott: Can we drop the `platforms` key altogether?  Or are there platforms where this shouldn't run?

Maybe just trim this to `qemu` only?

jlebon:
Can you add more context to the commit message? Even if just "These tests now pass.". If you know what made it work again, you could also include that in the message.

tests/kola/kdump/crash
dustymabe: If it can now run on all platforms then we can remove `platforms` from the list.

gursewak1997: @dustymabe I took it out of the list. This is an outdated change :)

tests/kola/kdump/crash
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# kola: {"platforms": "qemu-unpriv", "minMemory": 4096, "tags": "skip-base-checks", "architectures": "x86_64"}
+# kola: {"minMemory": 4096, "tags": "skip-base-checks", "architectures": "x86_64 aarch64"}
```
dustymabe: If there are no architectures we know of that this fails on we should remove the architectures key.

dustymabe:
Can you squash the commits down into a single commit?

dustymabe:
Now there are 82 changed files. Try again :)

dustymabe (approved):
LGTM

PR #1394, opened 2022-01-03
manifests/user-experience: Set nano as default editor

### LINKED ISSUE #1058: New Package Request: nano (and explicit choice for default editor)
The main goal here is to match the default with the one from Fedora as of Fedora 33: https://fedoraproject.org/wiki/Changes/UseNanoByDefault

Benefits (from the linked Change request):
> Users are exposed to the default editor when they use commands that call it. The main example here is something like git commit.
>
> Fedora does not currently have a default terminal text editor, because the $EDITOR environment variable is unset by default. But a common scenario where users wind up in a terminal text editor is when using 'git commit'. By default, git picks vi. You need to spend time learning how to use it, for even basic editing tasks. This increases the barrier to entry for those who are switching to Fedora and don't know how to use vi. It also makes things hard for those who don't particularly want to learn how to use vi. (These arguments would apply just as well if git picked Vim. vi is like hard mode for Vim, with fewer features, missing syntax highlighting, and no indication of what mode you are in. Even Vim users may feel lost and bewildered when using vi.)
>
> In contrast, Nano offers the kind of graphical text editing experience that people are used to, and therefore doesn't require specialist knowledge to use. It is already installed across most Fedora Editions and Spins. This proposal will make Nano the default editor, while continuing to install vim-minimal (which provides vi, but not Vim). People will still be able to call vi if they want to edit a file. It will also obviously be possible to change the default editor to vi or Vim, for those who want it.
>
> Why make Nano default and vi optional, rather than the other way round? Because Nano is the option that everyone can use.

----

None

From `dnf info`:

- nano: 709k
- nano-default-editor: 11k

User friendly default editor

No, the default editor must be pre-installed to be able to edit files outside of a container and before pulling containers.

Yes

No

7. Is it possible to layer the package onto the base OS as a day 2 operation?  Explain why or why not.

Yes

8. In the case of packages providing services and binaries, can the packaging be adjusted to just deliver binaries?

No service

9. Can the tool(s) provided by the package be used to do things we’d rather users not be able to do in FCOS? (e.g. can it be abused as a Turing complete interpreter?)

No

10. Does the software provided by the package have a history of CVEs?

Not relevant

travier:
Add nano and set it as default editor.

This is following the corresponding Fedora Change in Fedora 33:
https://fedoraproject.org/wiki/Changes/UseNanoByDefault

To revert to current default:
```
$ sudo rm /etc/profile.d/nano.{sh,csh}
```

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1058

### DIFF
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -870,6 +870,12 @@
     "mpfr": {
       "evra": "4.1.0-8.fc35.x86_64"
     },
+    "nano": {
+      "evra": "5.8-4.fc35.x86_64"
+    },
+    "nano-default-editor": {
+      "evra": "5.8-4.fc35.noarch"
+    },
     "ncurses": {
       "evra": "6.2-8.20210508.fc35.x86_64"
     },
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -26,6 +26,8 @@ packages:
   - fedora-coreos-pinger
   # Updates
   - zincati
+  # Include and set the default editor
+  - nano nano-default-editor
 
 etc-group-members:
   # Add the docker group to /etc/group
--- a/README.md
+++ b/README.md
@@ -129,8 +129,11 @@ the corresponding entries in the lockfiles:
 There will be better tooling to come to enable this, though
 one easy way to do this is for now:
 - add packages to the correct YAML manifest
-- run `cosa fetch --update-lockfile`
-- commit only the new package entries
+- run `cosa fetch --update-lockfile` (this will only update the lockfile for
+  the current architecture, most likely `x86_64`)
+- copy the new lines to the lockfiles for other architectures (i.e. `aarch64`)
+- commit only the new package entries (skip the timestamped changes to avoid
+  merge conflicts with the lockfile updates made by the bot)
 
 ## Moving to a new major version (N) of Fedora
 
--- a/manifest-lock.aarch64.json
+++ b/manifest-lock.aarch64.json
@@ -858,6 +858,12 @@
     "mpfr": {
       "evra": "4.1.0-8.fc35.aarch64"
     },
+    "nano": {
+      "evra": "5.8-4.fc35.aarch64"
+    },
+    "nano-default-editor": {
+      "evra": "5.8-4.fc35.noarch"
+    },
     "ncurses": {
       "evra": "6.2-8.20210508.fc35.aarch64"
     },


travier:
Let's keep the discussion in https://github.com/coreos/fedora-coreos-tracker/issues/1058

manifests/user-experience.yaml
```diff
@@ -11,7 +11,7 @@ packages:
   - jq
   - less
   - sudo
-  - vim-minimal
+  - nano-default-editor nano vim-minimal
```
miabbott: Since this manifest is shared with RHCOS, I don't think making the change here is correct.

travier: Indeed. Will move that.

miabbott (dismissed):
LGTM

manifests/fedora-coreos.yaml
```diff
@@ -26,6 +26,8 @@ packages:
   - fedora-coreos-pinger
   # Updates
   - zincati
+  # Include and set the default editor
+  - nano-default-editor nano
```
dustymabe: optional (and only because CI failed and has to run again anyway): I prefer this ordering: 

```suggestion
  - nano nano-default-editor
```

dustymabe (dismissed):
LGTM

jlebon:
See https://github.com/coreos/fedora-coreos-config#adding-packages-to-the-os re. the CI failure.

travier:
I've confirmed that removing the new profile files reverts to the previous behavior.

dustymabe (dismissed):
LGTM

dustymabe:
> I've confirmed that removing the new profile files reverts to the previous behavior.

When we add docs for "how to set your default editor" I would think we would just tell people to write a new /etc/profile.d file with a higher precedence rather than delete the file delivered by the `nano-default-editor` RPM. Do you agree?

travier:
I have a draft for that and I think the easiest way is to overwrite the existing config and add new ones unfortunately as there is no guarantee that this is read in order AFAIK.

travier:
https://github.com/coreos/fedora-coreos-docs/pull/349

dustymabe:
> I have a draft for that and I think the easiest way is to overwrite the existing config and add new ones unfortunately as there is no guarantee that this is read in order AFAIK.

It just uses a [shell glob](https://pagure.io/setup/blob/ac61b13c0483260d3fb0aa9cda81132172d93350/f/bashrc#_81) which I believe sorts reliably ([reference](https://unix.stackexchange.com/questions/368318/does-the-bash-star-wildcard-always-produce-an-ascending-sorted-list)).

dustymabe:
> coreos/fedora-coreos-docs#349

This page is good for an OS Extensions example, but I feel it would be good to have a separate section somewhere specifically targeted at the editor. For example, what if I just want to set my editor back to `vi` and not `vim`. I shouldn't have to look at the os extensions doc for that.

dustymabe:
FTR I think this particular PR is ready to merge.

travier:
> It just uses a [shell glob](https://pagure.io/setup/blob/ac61b13c0483260d3fb0aa9cda81132172d93350/f/bashrc#_81) which I believe sorts reliably ([reference](https://unix.stackexchange.com/questions/368318/does-the-bash-star-wildcard-always-produce-an-ascending-sorted-list)).

We would need something that gets loaded *after* the nano one to effectively apply. I'm afraid this will be confusing.

> This page is good for an OS Extensions example, but I feel it would be good to have a separate section somewhere specifically targeted at the editor. For example, what if I just want to set my editor back to `vi` and not `vim`. I shouldn't have to look at the os extensions doc for that.

Agree. I did not find a good place to add that yet but I can create another page. It's also a sort of *one-liner*:
```
$ sed -i 's/EDITOR=nano/EDITOR=vi/g /etc/profile.d/nano-*
```

dustymabe:
> > It just uses a [shell glob](https://pagure.io/setup/blob/ac61b13c0483260d3fb0aa9cda81132172d93350/f/bashrc#_81) which I believe sorts reliably ([reference](https://unix.stackexchange.com/questions/368318/does-the-bash-star-wildcard-always-produce-an-ascending-sorted-list)).
> 
> We would need something that gets loaded _after_ the nano one to effectively apply. I'm afraid this will be confusing.

What's confusing? It's alphabetical sorting and last one wins. Just make our file name start with `z` and add a sentence explaining why. 

> 
> > This page is good for an OS Extensions example, but I feel it would be good to have a separate section somewhere specifically targeted at the editor. For example, what if I just want to set my editor back to `vi` and not `vim`. I shouldn't have to look at the os extensions doc for that.
> 
> Agree. I did not find a good place to add that yet but I can create another page. It's also a sort of _one-liner_:
> 
> ```
> $ sed -i 's/EDITOR=nano/EDITOR=vi/g /etc/profile.d/nano-*
> ```

ehh. I would make it a butane snippet that defines a file `/etc/profile.d/z-default-editor.sh` and not a dynamically run command (though providing the dynamic command to run could be useful for people who installed first and then came to the docs).

manifest-lock.x86_64.json
```diff
@@ -870,6 +870,12 @@
     "mpfr": {
       "evra": "4.1.0-8.fc35.x86_64"
     },
+    "nano": {
+      "evra": "5.8-4.fc35.x86_64"
+    },
+    "nano-default-editor": {
+      "evra": "5.8-4.fc35.noarch"
+    },
```
jlebon, line 878: Don't we need the equivalent snippet in the `aarch64` lockfile too?

travier: We do. My bad.

travier:
https://github.com/coreos/fedora-coreos-docs/pull/349#issuecomment-1007666775 > I'm moving the doc changes discussion there to avoid splitting it between two places.

jlebon (approved):

dustymabe (approved):
LGTM

PR #1395, opened 2022-01-03
tests: Add OS extensions test

travier:
Ensure that we can install common tools that are not strictly required
for operation but do not really make sense to run from a container and
are usually useful as overlayed packages/OS extensions.

### DIFF
--- a/tests/kola/extensions/test.sh
+++ b/tests/kola/extensions/test.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# This test only runs on FCOS as OS extensions are implemented differently on RHCOS.
+# kola: { "distros": "fcos", "tags": "needs-internet" }
+#
+# This test ensures that we can install some common tools as OS extensions.
+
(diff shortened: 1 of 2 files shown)


miabbott (dismissed):
LGTM

tests/kola/extensions/test.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# No need to run on any other platform than QEMU.
```
cgwalters: I think we generally only do platform limitations when required.  Now clearly, running many of the tests on multiple platforms is wasteful.  But, you never know when e.g. somehow podman starts failing just on GCP but not AWS or qemu.

In this case, I could imagine for example that Fedora adds an AWS-internal mirror and somehow the routing setup is broken for package installs.  Perhaps farfetched but not out of the question.

Maybe what would be better is a new tag like "platform-sensitive" that we could add to tests, or a "platform-independent" tag that can be used to say e.g. "we'll only run this test on at most one or two platforms, not all".

cgwalters (approved):

PR #1401, opened 2022-01-05
networking: configure MTU on a VLAN subinterface for the bond works

HuijingHei:
Configuring MTU on a VLAN subinterface for the bond,
- verify MTU on the bond matches config
- verify MTU on the VLAN subinterface for the bond matches config
- verify ip address on the VLAN subinterface for the bond matches config
See https://bugzilla.redhat.com/show_bug.cgi?id=1932502#c9

### DIFF
--- a/tests/kola/networking/mtu-on-bond
+++ b/tests/kola/networking/mtu-on-bond
@@ -1,0 +1,35 @@
+#!/bin/bash
+# kola: { "platforms": "qemu", "additionalNics": 2, "appendKernelArgs": "bond=bond0:eth1,eth2:mode=active-backup,miimon=100:9000 ip=10.10.10.10::10.10.10.1:255.255.255.0:staticvlanbond:bond0.100:none:9000: vlan=bond0.100:bond0 net.ifnames=0"}
+# - We use net.ifnames=0 to disable consistent network naming here because on
+#   different firmwares (BIOS vs UEFI) the NIC names are different.
+#   See https://github.com/coreos/fedora-coreos-tracker/issues/1060
+
+# Configuring MTU on a VLAN subinterface for the bond,
+# - verify MTU on the bond matches config
+# - verify MTU on the VLAN subinterface for the bond matches config
+# - verify ip address on the VLAN subinterface for the bond matches config
+# See https://bugzilla.redhat.com/show_bug.cgi?id=1932502#c9
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+bond="bond0"
+vlan="bond0.100"
+for interface in $bond $vlan
+do
+    mtu=""
+    # MTU is changed to 9000 by `appendKernelArgs` in `kola` above
+    mtu=$(nmcli -g 802-3-ethernet.mtu connection show ${interface})
+    if [ "${mtu}" != "9000" ]; then
+        fatal "Error: get ${interface} mtu = ${mtu}, expected is 9000"
+    fi
+    ok "${interface} mtu is correct"
+done
+
+# Verify "bond0.100" gets ip address by `appendKernelArgs` in `kola` above
+nic_ip=$(get_ipv4_for_nic ${vlan})
+if [ "${nic_ip}" != "10.10.10.10" ]; then
+    fatal "Error: get ${vlan} ip = ${nic_ip}, expected is 10.10.10.10"
+fi
+ok "get ${vlan} ip is 10.10.10.10"


miabbott:
Can we have a similar test that uses an Ignition config to setup the bond + MTU?

tests/kola/networking/mtu-on-bond
```diff
@@ -0,0 +1,31 @@
...
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+bond="bond0"
+vlan="bond0.100"
+for interface in $bond $vlan
+do
+    mtu=""
+    mtu=`nmcli -g 802-3-ethernet.mtu connection show ${interface}`
+    if [ "${mtu}" != "9000" ]; then
```
miabbott, line 24: Might be helpful to drop a comment here saying that changing the expected MTU value should be done here and in the `appendKernelArgs` value above.

miabbott (dismissed):
This looks great, just one minor suggestion, but not blocking.

tests/kola/networking/mtu-on-bond
```diff
@@ -0,0 +1,31 @@
...
+        fatal "Error: get ${interface} mtu = ${mtu}, expected is 9000"
+    fi
+    ok "${interface} mtu"
```
ravanelli:
```suggestion
    ok "${interface} mtu is correct"
```

ravanelli (dismissed):
Just a suggestion, other LGTM

mike-nguyen (dismissed):
Not a bond expert but LGTM.  I ran the test locally also and it passed

tests/kola/networking/mtu-on-bond
```diff
@@ -0,0 +1,31 @@
...
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+bond="bond0"
+vlan="bond0.100"
+for interface in $bond $vlan
+do
+    mtu=""
+    mtu=`nmcli -g 802-3-ethernet.mtu connection show ${interface}`
```
dustymabe: I tend to prefer the `$()` form over backticks. 

```suggestion
    mtu=$(nmcli -g 802-3-ethernet.mtu connection show ${interface})
```

dustymabe:
Unfortunately we're going to run into https://github.com/coreos/fedora-coreos-tracker/issues/1060 here as well. We either need to limit this test to x86_64 only or figure out how to make things arch independent before we can merge this.

HuijingHei:
Thanks all for the review!
@dustymabe Add the limitation to x86_64 only, will open this to other arch if we have the common workaround for coreos/fedora-coreos-tracker#1060

dustymabe:
> Thanks all for the review! @dustymabe Add the limitation to x86_64 only, will open this to other arch if we have the common workaround for [coreos/fedora-coreos-tracker#1060](https://github.com/coreos/fedora-coreos-tracker/issues/1060)

You can probably just update this PR similar to how I did the other two in https://github.com/coreos/fedora-coreos-config/pull/1408. I'll merge #1408 in the morning if no one objects and then we can do this one after that.

HuijingHei:
> Can we have a similar test that uses an Ignition config to setup the bond + MTU?

That makes sense. I am not sure which way I should choose:
1) Add `kernel_arguments` to setup the bond + MTU in Ignition config
2) Use `$ /usr/libexec/nm-initrd-generator -s -- $kargs` to generate config files, then in Ignition config add them under `/etc/NetworkManager/system-connections/`
Another question, is there any difference between them?

dustymabe (dismissed):
LGTM

dustymabe:
> > Can we have a similar test that uses an Ignition config to setup the bond + MTU?
> 
> That makes sense. I am not sure which way I should choose:
> 
> 1. Add `kernel_arguments` to setup the bond + MTU in Ignition config
> 2. Use `$ /usr/libexec/nm-initrd-generator -s -- $kargs` to generate config files, then in Ignition config add them under `/etc/NetworkManager/system-connections/`
>        Another question, is there any difference between them?

Maybe we can push this off to a followup PR.

dustymabe:
I'm going to rebase this on top of latest `testing-devel` and then CI should pass hopefully.

dustymabe:
> I'm going to rebase this

Looks like I don't have permissions to push to the remote branch/repo. @HuijingHei can you rebase this on top of latest `testing-devel`?

HuijingHei:
> Looks like I don't have permissions to push to the remote branch/repo. @HuijingHei can you rebase this on top of latest `testing-devel`?

Sure, done, thanks!

miabbott (approved):
LGTM

HuijingHei:
coreos/fedora-coreos-tracker#1060 is closed and remove the hold label

PR #1405, opened 2022-01-05
tests: Rework 'find' calls with multiple file type filters

travier:
Fedora has a recent enough `find` command to be able to use the `-type
f,d` option but that is not yet supported by `find` on RHEL.

We can not use `-type f -type d` instead as this filters on entries
being both a file and a directory which never happens.

Thus rework tests using that option to be compatible on both FCOS and
RHCOS. We can drop this change once RHCOS moves to RHEL 9.

Fixes: https://github.com/coreos/fedora-coreos-config/commit/dac1f6835e1e730409a4062c6562ece32bddaee1

### DIFF
--- a/tests/kola/files/file-directory-permissions
+++ b/tests/kola/files/file-directory-permissions
@@ -5,9 +5,11 @@ set -xeuo pipefail
 
 . $KOLA_EXT_DATA/commonlib.sh
 
-list="$(find /etc -type f -type d -perm /022)"
+list="$(find /usr /etc -type f -perm /022 -o -type d -perm /022)"
+
 if [[ -n "${list}" ]]; then
-    find /etc -type f -type d -perm /022 -print0 | xargs -0 ls -al
+    find /usr /etc -type f -perm /022 -print0 -o -type d -perm /022 -print0 | xargs -0 ls -al
+    find /usr /etc -type f -perm /022 -print0 -o -type d -perm /022 -print0 | xargs -0 rpm -qf
     fatal "found files or directories with 'g+w' or 'o+w' permission"
 fi
 ok "no files with 'g+w' or 'o+w' permission found in /etc"
--- a/tests/kola/files/file-directory-permissions
+++ b/tests/kola/files/file-directory-permissions
@@ -5,9 +5,42 @@ set -xeuo pipefail
 
 . $KOLA_EXT_DATA/commonlib.sh
 
-list="$(find /usr /etc -type f -perm /022 -o -type d -perm /022)"
+# List of known files and directories with group write permission
+list_known=()
 
-if [[ -n "${list}" ]]; then
+# List of known files and directories with group write permission (RHCOS only)
+list_known_rhcos=(
+    '/usr/share/licenses/publicsuffix-list-dafsa/COPYING'
+)
+
+is_fcos="false"
+if [[ "$(source /etc/os-release && echo "${ID}")" == "fedora" ]]; then
+    is_fcos="true"
+fi
+
+unknown=""
+while IFS= read -r -d '' e; do
+    found="false"
+    for k in "${list_known[@]}"; do
+        if [[ "${k}" == "${e}" ]]; then
+            found="true"
+            break
+        fi
+    done
+    if [[ "${is_fcos}" == "false" ]]; then
+        for k in "${list_known_rhcos[@]}"; do
+            if [[ "${k}" == "${e}" ]]; then
+                found="true"
+                break
+            fi
+        done
+    fi
+    if [[ "${found}" == "false" ]]; then
+        unknown+=" ${e}"
+    fi
+done< <(find /usr /etc -type f -perm /022 -print0 -o -type d -perm /022 -print0)
+
+if [[ -n "${unknown}" ]]; then
     find /usr /etc -type f -perm /022 -print0 -o -type d -perm /022 -print0 | xargs -0 ls -al
     find /usr /etc -type f -perm /022 -print0 -o -type d -perm /022 -print0 | xargs -0 rpm -qf
     fatal "found files or directories with 'g+w' or 'o+w' permission"
--- a/tests/kola/files/setuid
+++ b/tests/kola/files/setuid
@@ -25,6 +25,20 @@ list_setuid_files=(
     '/usr/sbin/unix_chkpwd'
 )
 
+# List of known files and directories with SetUID bit set (RHCOS only)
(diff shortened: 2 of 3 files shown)


tests/kola/files/file-directory-permissions
```diff
@@ -4,9 +4,19 @@ set -xeuo pipefail
...
+)
+for t in 'f' 'd'; do
+  while IFS= read -r -d '' e; do
+    unknown_setuid[${t}]+=" ${e}"
+  done< <(find /usr /etc -type "${t}" -perm /022 -print0)
+done
+
+if [[ -n "${list['f']}" ]] || [[ -n "${list['d']}" ]]; then
+    find /etc -type f -perm /022 -print0 | xargs -0 ls -al
+    find /etc -type d -perm /022 -print0 | xargs -0 ls -al
```
miabbott: Since the first `for` loop is checking `/usr` and `/etc`, we should check each of those directories here, too.

miabbott: As an additional bit of data, could you pipe the output of the `find` command into `xargs -0 rpm -qf` too?

bgilbert:
> We can not use `-type f -type d` instead as this filters on entries being both a file and a directory which never happens.

We could use `-type f -o -type d`.

tests/kola/files/setuid
```diff
@@ -23,22 +23,28 @@ list_setuid_files=(
...
+      for k in "${list_setuid_files[@]}"; do
+          if [[ "${k}" == "${e}" ]]; then
+              found="true"
+              break
+          fi
+      done
+      if [[ "${found}" == "false" ]]; then
+          unknown_setuid[${t}]+=" ${e}"
+      fi
+  done< <(find /usr /etc -type "${t}" -perm /4000 -print0)
```
miabbott: This returns a larger list of files on a recent RHCOS 4.10 build:

```
$ sudo find /usr /etc -type f -perm /4000 -print0 | xargs -0 ls -la
-rwsr-xr-x. 2 root root  79672 Jan  1  1970 /usr/bin/chage
-rwsr-xr-x. 2 root root  38680 Jan  1  1970 /usr/bin/fusermount
-rwsr-xr-x. 2 root root  34560 Jan  1  1970 /usr/bin/fusermount3
-rwsr-xr-x. 2 root root  84344 Jan  1  1970 /usr/bin/gpasswd
-rwsr-xr-x. 2 root root  50472 Jan  1  1970 /usr/bin/mount
-rwsr-xr-x. 2 root root  43584 Jan  1  1970 /usr/bin/newgrp
-rwsr-xr-x. 2 root root  33544 Jan  1  1970 /usr/bin/passwd
-rwsr-xr-x. 2 root root  29048 Jan  1  1970 /usr/bin/pkexec
-rwsr-xr-x. 2 root root  50336 Jan  1  1970 /usr/bin/su
---s--x--x. 2 root root 165488 Jan  1  1970 /usr/bin/sudo
-rwsr-xr-x. 2 root root  33664 Jan  1  1970 /usr/bin/umount
-rwsr-xr-x. 2 root root  16912 Jan  1  1970 /usr/lib/polkit-1/polkit-agent-helper-1
-rwsr-x---. 2 root dbus  63592 Jan  1  1970 /usr/libexec/dbus-1/dbus-daemon-launch-helper
-rwsr-x---. 2 root sssd 172088 Jan  1  1970 /usr/libexec/sssd/krb5_child
-rwsr-x---. 2 root sssd  97392 Jan  1  1970 /usr/libexec/sssd/ldap_child
-rwsr-x---. 2 root sssd  29112 Jan  1  1970 /usr/libexec/sssd/proxy_child
-rwsr-x---. 2 root sssd  55344 Jan  1  1970 /usr/libexec/sssd/selinux_child
-rwsr-xr-x. 2 root root  12096 Jan  1  1970 /usr/sbin/grub2-set-bootflag
-rwsr-xr-x. 2 root root 197104 Jan  1  1970 /usr/sbin/mount.nfs
-rwsr-xr-x. 2 root root  12336 Jan  1  1970 /usr/sbin/pam_timestamp_check
-rwsr-xr-x. 2 root root  37952 Jan  1  1970 /usr/sbin/unix_chkpwd
```

miabbott:
Good catch on the use of `-type f -type d`; the changed behavior looks correct now.

travier:
> > We can not use `-type f -type d` instead as this filters on entries being both a file and a directory which never happens.
> 
> We could use `-type f -o -type d`.

Could idea. Made a new try.

tests/kola/files/setuid
travier: I'll have to make an RHCOS specific test or exclude list

travier:
This one is green but missing RHCOS testing.

travier:
Updated with RHCOS support. Not pretty. Open to suggestions.

tests/kola/files/file-directory-permissions
```diff
@@ -5,9 +5,44 @@ set -xeuo pipefail
...
 
-list="$(find /etc -type f -type d -perm /022)"
-if [[ -n "${list}" ]]; then
-    find /etc -type f -type d -perm /022 -print0 | xargs -0 ls -al
+# List of known files and directories with group write permission
+list_known=()
+
+# List of known files and directories with group write permission (RHCOS only)
+list_known_rhcos=(
+    '/usr/share/licenses/publicsuffix-list-dafsa/COPYING'
```
cgwalters, line 13: Tempting to just canonicalize this in our ostree builds too.

tests/kola/files/setuid
```diff
@@ -24,6 +24,21 @@ list_setuid_files=(
     '/usr/sbin/pam_timestamp_check'
     '/usr/sbin/unix_chkpwd'
 )
+
+# List of known files and directories with SetUID bit set (RHCOS only)
```
cgwalters, line 28: This is another case where we're saying "rhcos" but we really mean "rhel8".

cgwalters (approved):
Didn't review the `find` incantations in depth, but looks sane to me!

tests/kola/files/file-directory-permissions
travier: I've made https://gitlab.com/redhat/centos-stream/rpms/publicsuffix-list/-/merge_requests/2 for this one.

travier: But indeed, we can directly fix it in openshift/os too and drop this case from the test.

PR #1406, opened 2022-01-05
tests: Add OS module extensions test

LorbusChris:
Ensure that we can install commonly used RPM modules as overlay.

This is somewhat related to OKD's planned move to use the cri-o module, even though it is currently not added as an overlay there but part of the compose: https://github.com/openshift/okd-machine-os/pull/249

~~I'm not sure the perl overlay is actually commonly overlayed, but it seemed to make the most sense when looking at the output of `dnf module list`, other than cri-o of course.~~ replaced with nginx

/cc @jlebon
/cc @travier

### DIFF
--- a/tests/kola/extensions/module
+++ b/tests/kola/extensions/module
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# This test only runs on FCOS as OS extensions are implemented differently on RHCOS.
+# kola: { "distros": "fcos", "tags": "needs-internet" }
+#
+# This test ensures that we can install some common tools as OS extensions.
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# Note: This version must be kept up to date for now, since there currently is no 'latest' (or similar) stream
+CRIO_VERSION=1.22
+
+declare -A commands=(
+  ["cri-o:${CRIO_VERSION}/default"]="crio"
+  ["nginx:mainline"]="nginx"
+)
+
+rpm-ostree ex module install "${!commands[@]}"
+rpm-ostree ex apply-live
+
+failed=""
+
+for m in "${!commands[@]}"; do
+  if [[ -z "$(command -v "${commands[$m]}")" ]]; then
+    echo "Could not find: ${commands[$m]}"
+    failed+=" ${1}"
+  fi
+done
+
+if [[ -n "${failed}" ]]; then
+  fatal "could not install: ${failed}"
+fi
+ok "successfully installed os rpm module extensions"
--- a/tests/kola/extensions/package
+++ b/tests/kola/extensions/package
@@ -31,4 +31,4 @@ done
 if [[ -n "${failed}" ]]; then
   fatal "could not install: ${failed}"
 fi
-ok "successfully installed os extensions"
+ok "successfully installed os rpm package extensions"


tests/kola/extensions/module
```diff
@@ -0,0 +1,30 @@
...
+# This test only runs on FCOS as OS extensions are implemented differently on RHCOS.
+# kola: { "distros": "fcos", "tags": "needs-internet" }
+#
+# This test ensures that we can install some common tools as OS extensions.
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+commands=(
+  'cri-o:1.22/default'
+  'perl:5.32'
```
miabbott: Any concern about hardcoding versions?  Will these versions always be available?

Maybe just test something like `nginx:mainline` which seems less likely to disappear from the available modules?

tests/kola/extensions/module
```diff
@@ -0,0 +1,30 @@
...
+)
+
+rpm-ostree ex module install --apply-live "${commands[@]}"
```
miabbott: `--apply-live` doesn't appear to work with `ex module install`:

```bash
$ sudo rpm-ostree ex module install --apply-live perl:5.32
NOTICE: Experimental commands are subject to change.
error: Found argument '--apply-live' which wasn't expected, or isn't valid in this context
```

Will need to `rpm-ostree ex apply-live` as a secondary step

tests/kola/extensions/module
```diff
@@ -0,0 +1,30 @@
...
+
+failed=""
+
+for c in "${commands[@]}"; do
+  if [[ -z "$(command -v "${c}")" ]]; then
```
miabbott: You'll have to strip off the module version from the item in the array before passing it to `command -v`

jlebon:
Heh, we did have this test in this repo, but then migrated it to rpm-ostree:
- https://github.com/coreos/fedora-coreos-config/pull/1211
- https://github.com/coreos/rpm-ostree/pull/3123

I'm good having it here too, but as mentioned in that first PR and alluded to by Micah, it'll require some maintenance to keep the versions valid. Note these tests also run on rawhide, so might need to do something similar to what's in rpm-ostree: https://github.com/coreos/rpm-ostree/blob/23d894d80aacd4bdece1193b49bdf3be0e2625f9/tests/kolainst/destructive/layering-modules#L13-L20 (but adding f36 too, and it's likely the existing NSVs there are outdated too).

LorbusChris:
Thanks for the review @miabbott! I think I've addressed all your points.

travier:
I think it has value here as it will check things for releases and not just for inclusion in rpm-ostree, thus making this more likely to be closer to what's been used. Agree that full OKD testing would help too but this is a first step.

LorbusChris:
Some context: OKD today does actually not use the cri-o Fedora module at all but instead consumes an RPM built on Open Build Service. And since there is no (and won't be) Go 1.17 in F35, the switch to the module will land in F36 at the earliest.

tests/kola/extensions/module
```diff
@@ -0,0 +1,38 @@
...
+modules=(
+  "cri-o:${CRIO_VERSION}/default"
+  "nginx:mainline"
+)
+declare -A commands=(
+  ["cri-o:${CRIO_VERSION}/default"]="crio"
+  ["nginx:mainline"]="nginx"
+)
+
+rpm-ostree ex module install "${modules[@]}"
```
jlebon: Isn't this

```suggestion
rpm-ostree ex module install "${!commands[@]}"
```

?

And then we could get rid of `modules`.

jlebon (dismissed):
One comment, but LGTM overall.

tests/kola/extensions/module
travier: Learned a new bash trick :)

LorbusChris: I was looking for a better way to loop over the indices, and this it, thank you very much! TIL :)

jlebon (approved):

PR #1381, opened 2021-12-17
[next-devel] manifest: add microdnf to be used internally by rpm-ostree

jmarrero:
This is as proposed on: https://github.com/coreos/fedora-coreos-tracker/issues/1050
c.c. @cgwalters

PR #1411, opened 2022-01-11
[next] manifest: add microdnf to be used internally by rpm-ostree

jmarrero:
Same as: https://github.com/coreos/fedora-coreos-config/pull/1381 but for `next`

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -10,3 +10,19 @@ rojig:
 
 add-commit-metadata:
   fedora-coreos.stream: next
+
+# These packages are needed for installing packages thru microdnf
+# Currently, this is planned for use on the layering enhancement
+# enhancement: https://github.com/coreos/enhancements/pull/7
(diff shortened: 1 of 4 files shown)


manifest.yaml
```diff
@@ -10,3 +10,16 @@ rojig:
...
   fedora-coreos.stream: next
+
+# These packages are needed for installing packages thru microdnf
+# Currently, this is planned for use on the layering enhancement
+# enhancement: https://github.com/coreos/enhancements/pull/7
+# package request: https://github.com/coreos/fedora-coreos-tracker/issues/1050
+packages:
+ - microdnf
+
+postprocess:
```
dustymabe, line 21: in `next-devel` branch there is a comment here that would be good for us to also add.

jmarrero: done

dustymabe (approved):
LGTM

PR #1412, opened 2022-01-11
networking: set MTU on a VLAN subinterface for the bond using ignition config works

HuijingHei:
Set MTU on a VLAN subinterface for the bond using ignition config and check
- verify MTU on the bond matches config
- verify MTU on the VLAN subinterface for the bond matches config
- verify ip address on the VLAN subinterface for the bond matches config

The ignition config is generated using nm-initrd-generator according to
https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-network-configuration/
$ kargs="bond=bond0:eth1,eth2:mode=active-backup,miimon=100:9000 \
       ip=10.10.10.10::10.10.10.1:255.255.255.0:staticvlanbond:bond0.100:none:9000: \
       vlan=bond0.100:bond0"
$ /usr/libexec/nm-initrd-generator -s -- $kargs


HuijingHei:
This is the following PR for https://github.com/coreos/fedora-coreos-config/pull/1401

tests/kola/networking/mtu-on-bond-ignition/config.bu
```diff
@@ -0,0 +1,65 @@
...
+    - path: /etc/NetworkManager/system-connections/bond0.100.nmconnection
+      mode: 0600
+      contents:
+        inline: |
+          [connection]
+          id=bond0.100
+          type=vlan
+          interface-name=bond0.100
+          [ethernet]
+          mac-address-blacklist=
```
ravanelli: I know you used the doc to create it. A curiosity, make sense to have these items even with empty values?

HuijingHei: Thanks @ravanelli , maybe no need to keep them, I will remove all items with empty values

dustymabe: ehh - I would use exactly what was generated by `nm-initrd-generator`. This makes it easy to regenerate in the future.

dustymabe (approved):
LGTM - though I would keep the NM config files exactly as they were generated by `nm-initrd-generator`.

tests/kola/networking/mtu-on-bond-ignition/config.bu
HuijingHei: Does it need to remove line like `uuid=d1171f53-2502-45dd-9199-698a778bf3bb`?

dustymabe: no - we don't need to remove the uuid lines I don't think.

HuijingHei: Just to confirm, this means we should remove the uuid lines?

dustymabe: No. I would think here that we don't need to remove **any** lines from the output of `nm-initrd-generator`.

HuijingHei: OK, thanks for your confirmation

PR #1387, opened 2021-12-21
workflows: comment on `next-devel` PRs if the branch is closed

bgilbert:
When a next-devel PR is opened, reopened, marked ready for review, or its base branch is changed, post a comment if the branch is not currently accepting PRs.  This should help reduce confusion for occasional maintainers.

Also add a README badge indicating the current `next-devel` branch status.

PR #1414, opened 2022-01-11
workflows: allow commenting on next-devel PRs from forks

bgilbert:
This was written to be a privileged job, but wasn't actually declared as such.

Fixes #1387.


jlebon (approved):
Not familiar with the details, but seems sane.

bgilbert:
Jenkins CI isn't relevant here; force-merging.

PR #1415, opened 2022-01-11
Use new `overlay.d/cosa-no-autolayer` option

jlebon:
Tell coreos-assembler to not automatically add the overlays to the final
manifest. Add references to the auto-generated OSTree layers in our
lower-level manifests where most appropriate.

Prep for using this to make an overlay only active on the `next-devel`
stream.

Requires: https://github.com/coreos/coreos-assembler/pull/2641

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -10,6 +10,13 @@ include:
   - user-experience.yaml
   - shared-workarounds.yaml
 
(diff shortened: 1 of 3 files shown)


dustymabe:
Does the `cosa-no-autolayer` have to be empty? Might be a good place to add a comment explaining why it's there and when it can be removed.

dustymabe (approved):
LGTM

PR #1417, opened 2022-01-12
kola-denylist: snooze new karg networking test on rawhide

dustymabe:
Similar to 104ecf1 we need to snooze the new karg networking test
added in 0a577a1 on rawhide because of the same issue.

See https://github.com/coreos/fedora-coreos-tracker/issues/1059


saqibali-2k (approved):

dustymabe:
skipping CI since the denial is for the `rawhide` stream and this is `testing-devel`

PR #1419, opened 2022-01-12
[next] overrides: pin to kernel-5.15.7-200.fc35

dustymabe:
Newer kernels seem to have an issue with enhanced networking on some
AWS instance types so they don't boot. Let's pin on the older kernel
for now while we investigate and find a proper solution.

See https://github.com/coreos/fedora-coreos-tracker/issues/1066

(cherry picked from commit 467af828b69fb451aab7aaff2f6e2a2854fe64ef)


bgilbert (approved):

dustymabe:
Note these PRs won't automerge because CI will fail because of the downgrade check.

bgilbert:
CI failed on expected downgrades.

PR #1420, opened 2022-01-12
kola-denylist: remove kdump.crash from denylist

gursewak1997:
kdump.crash should run and pass after fixes in https://github.com/coreos/coreos-assembler/pull/2639
Will need to skip kdump.crash test in `coreos-installer` and `afterburn` due CI failures.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -12,8 +12,6 @@
     - rawhide
   arches:
     - aarch64
-- pattern: ext.config.kdump.crash
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/860
 - pattern: ext.config.podman.dns
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1046
   snooze: 2022-01-20


bgilbert:
From OOB discussion: this test has broken upstream CI a couple times now, so it'd be good to patch it into upstream CI in a PR and make sure it passes before proceeding.  coreos-installer might be a good candidate.

dustymabe:
Is this ready yet?

gursewak1997:
> Is this ready yet?

Unfortunately not, it looks likes the the test fails for upstream CI for [coreos-installer](https://jenkins-coreos-ci.apps.ocp.ci.centos.org/blue/organizations/jenkins/github-ci%2Fcoreos%2Fcoreos-installer/detail/PR-750/5/pipeline)

travier:
> https://github.com/coreos/coreos-installer/pull/750

OK, that's weird. Hopefully we can figure that out but we might also want to make this test "release" only or "config" repo and not block other projects on it. It's not worth it.

bgilbert (approved):
This has now been denylisted in coreos-installer and Afterburn, which are believed to be the affected repos.

PR #1421, opened 2022-01-12
[next-devel] manifest-lock: import from testing-devel

jmarrero:
splitting this from #1381 to better debug where the the CI failure is coming from.


jmarrero:
I am guessing this is what I am hitting https://github.com/coreos/coreos-assembler/pull/2643

PR #1422, opened 2022-01-13
kola-denylist: update mtu-on-bond test name

dustymabe:
The test name changed in bf37b9c.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -34,7 +34,7 @@
   snooze: 2022-01-25
   streams:
     - rawhide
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

jlebon (approved):
Do we need to also add the new networking test added in that commit as well?

saqibali-2k (approved):

dustymabe:
> Do we need to also add the new networking test added in that commit as well?

no, IIUC it doesn't use any kargs that systemd-network-generator acts on.

dustymabe:
skipping CI because it won't test this change.

PR #1424, opened 2022-01-13
overrides: update pinned kernel to last known good

dustymabe:
We shipped out updates this morning to `testing` and `next`
with a downgraded kernel that matches what is already in `stable`.
For our upcoming regularly scheduled releases let's at least get to
the latest possible known good kernel.

See https://github.com/coreos/fedora-coreos-tracker/issues/1066


bgilbert (dismissed):
We'll want this in next-devel too.

bgilbert (approved):

jlebon (approved):

PR #1436, opened 2022-01-20
manifests: move location of nvme-cli for sharing with RHCOS

miabbott:
OCP/RHCOS customers want to have `nvme-cli` included as part of the
OS, so let's move the location of the package in the manifests to one
that is shared with RHCOS.

See: https://issues.redhat.com/browse/RFE-2503


dustymabe (approved):
LGTM

travier:
Well, turns out it's not enabled for this repo 🙂

PR #1437, opened 2022-01-20
denylist: extend snooze for ext.config.podman.dns

dustymabe:
I poked upstream to see if we can further the discussion:
https://github.com/containers/podman/issues/12661#issuecomment-1017899651


dustymabe:
will skip CI because this applies to kola tests run against rawhide.

ravanelli (approved):
LGTM

PR #1440, opened 2022-01-21
denylist: extend snooze for podman.base on rawhide

dustymabe:
Still seeing it. There is some movement in the BZ, but
let's snooze again.

https://github.com/coreos/fedora-coreos-tracker/issues/1049
https://bugzilla.redhat.com/show_bug.cgi?id=2033016


dustymabe:
will skip CI for this one as it applies to the rawhide stream

jlebon (approved):

PR #1441, opened 2022-01-22
Add tool and GH workflow for adding fast-tracks and pins

bgilbert:
Rename and generalize `remove-graduated-overrides.py`.  Add subcommands for adding fast-tracks (from Bodhi update IDs or URLs) and pins (from SRPM NVRs).  Unless `--force` is specified, reject newly added overrides for the wrong Fedora major by comparing their dist tags to the `releasever` from `manifest.yaml`.

Add a wrapper GitHub workflow that creates a PR to one branch, pinning one SRPM or fast-tracking one Bodhi update.   More complex cases can use multiple workflow runs or manual PRs.

### DIFF
--- a/.github/workflows/add-override.yml
+++ b/.github/workflows/add-override.yml
@@ -1,0 +1,90 @@
+---
+name: Add package override
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: Target branch
+        default: testing-devel
+      what:
+        description: "Bodhi update (fast-track) or SRPM NVR (pin)"
+      pin:
+        description: "Pin (don't remove when stable)"
+        type: boolean
+      reason:
+        description: "Reason URL (optional for routine fast-tracks)"
+
+permissions:
+  # none at all
+  contents: none
+
+# This workflow could almost use the default GITHUB_TOKEN, if we were to
+# push the branch into this repo.  However, GitHub Actions has recursion
+# avoidance that would prevent CI from running on the PR:
+#
+# https://github.com/peter-evans/create-pull-request/blob/28fa4848947e/docs/concepts-guidelines.md#workarounds-to-trigger-further-workflow-runs
+#
+# So we create the PR using a separate Personal Access Token in
+# COREOSBOT_RELENG_TOKEN, belonging to a machine account.  That allows CI to
+# run when the PR is first created.  However, it's also possible to rerun
+# the workflow and have it force-push the branch, reusing the same PR.  In
+# that case the push also cannot come from GITHUB_TOKEN, or CI will not
+# rerun.  Thus we also do the push using COREOSBOT_RELENG_TOKEN.  Since we
+# don't want to give the machine account privileges to this repo, we push
+# to a forked repo owned by the machine account.
+
+jobs:
+  add-override:
+    name: Add package override
+    runs-on: ubuntu-latest
+    container: registry.fedoraproject.org/fedora:latest
+    steps:
+      - name: Install dependencies
+        run: dnf install -y git jq python3-bodhi-client python3-pyyaml
+      - name: Check out repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.target }}
+          # We need an unbroken commit chain when pushing to the fork.  Don't
+          # make assumptions about which commits are already available there.
+          fetch-depth: 0
+      - name: Update metadata
+        env:
+          TARGET: ${{ github.event.inputs.target }}
+          WHAT: ${{ github.event.inputs.what }}
+          PIN: ${{ github.event.inputs.pin }}
+          REASON: ${{ github.event.inputs.reason }}
+        run: |
+          set -euxo pipefail
+
+          if [ "${PIN}" = true ]; then
+              ci/overrides.py pin "${WHAT}" -r "${REASON}"
+              title="overrides: pin ${WHAT}"
+          else
+              ci/overrides.py fast-track "${WHAT}" ${REASON:+-r "${REASON}"}
+              srpms=$(ci/overrides.py srpms "${WHAT}" | paste -sd,)
+              title="overrides: fast-track ${srpms//,/, }"
+          fi
+
+          if [ "${TARGET}" = testing-devel ]; then
+              pr_title="${title}"
+          else
+              pr_title="[${TARGET}] ${title}"
+          fi
+          branch_name=override-$(echo "${TARGET}:${title}" | sha256sum | cut -c1-8)
+
+          echo "BRANCH_NAME=${branch_name}" >> ${GITHUB_ENV}
+          echo "COMMIT_TITLE=${title}" >> ${GITHUB_ENV}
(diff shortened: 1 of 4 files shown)


bgilbert:
Ah, I see we can't add keys to the override YAML without teaching rpm-ostree about them.  Will leave this as-is for feedback for now.

ci/overrides.py
```diff
@@ -41,6 +61,41 @@ def main():
...
+        update = get_bodhi_update(update)
+        for n, evr in get_binary_packages(get_source_nvrs(update)).items():
+            overrides[n] = dict(
+                evr=evr,
+                metadata=dict(
+                    type='fast-track',
+                    bodhi=update['url'],
+                )
+            )
+            if args.reason:
```
jlebon, line 108: FWIW, the policy has roughly been that any CoreOS-specific/CoreOS-team-owned packages doesn't need an explicit reason for fast-tracking, but e.g. fast-tracking the kernel should have a reason. Should we have an allowlist of packages that can be fast-tracked without reason?

ci/overrides.py
```diff
@@ -33,6 +38,21 @@ def main():
...
+            description='Fast-track Bodhi updates.')
+    fast_track.add_argument('update', nargs='+',
+            help='ID or URL of Bodhi update to fast-track')
+    fast_track.add_argument('-r', '--reason',
+            help='URL explaining the reason for the fast-track')
+    fast_track.set_defaults(func=do_fast_track)
+
+    pin = subcommands.add_parser('pin', description='Pin source RPMs.')
+    pin.add_argument('nvr', nargs='+',
+            help='NVR of SRPM to pin')
```
jlebon, line 66: Minor/optional: since we're already importing koji, could be nice to also support Koji build URLs.

ci/overrides.py
```diff
@@ -65,6 +120,50 @@ def get_dnf_base(treefile):
     return base
 
 
+@functools.cache
```
jlebon, line 161: Woah, nice. TIL.

ci/overrides.py
```diff
@@ -65,6 +120,50 @@ def get_dnf_base(treefile):
...
+def get_manifest_packages(arch):
```
jlebon, line 162: This isn't entirely foolproof, because you could e.g. be pinning/fast-tracking to a version which has a previously-removed/new subpackage.

Though yeah, I guess this falls under the "more complex cases" caveat.

ci/overrides.py
```diff
@@ -65,6 +120,50 @@ def get_dnf_base(treefile):
...
+
+def get_binary_packages(source_nvrs):
+    '''Return name => EVR dict for the specified source NVRs.  A binary
+    package is included if, for any arch, it exists and is in the manifest
+    lock.'''
+    binpkgs = {}
+    client = koji.ClientSession(KOJI_URL)
+    for source_nvr in source_nvrs:
+        for binpkg in client.listBuildRPMs(source_nvr):
+            if binpkg['name'] in get_manifest_packages(binpkg['arch']):
```
jlebon: Optional: note this will not handle arch-specific packages correctly I think. I agree this falls under the "more complex cases", though one thing we could do is check upfront that the package is in all arches before adding it, and otherwise error out with a clear message.

ci/overrides.py
```diff
@@ -44,13 +44,17 @@ def main():
             help='ID or URL of Bodhi update to fast-track')
     fast_track.add_argument('-r', '--reason',
             help='URL explaining the reason for the fast-track')
+    fast_track.add_argument('--force', action='store_true',
```
jlebon: Should this switch name be more specific? E.g. `--ignore-releasever-mismatch`? `--force` could change meaning in the future.

.github/workflows/add-override.yml
```diff
@@ -0,0 +1,123 @@
...
+          # forks" API because the target repo might be the same as the
+          # origin (when testing this workflow).
+          fork_user=$(curl -s -H "Accept: application/vnd.github.v3+json" \
+              -H "Authorization: token ${{ secrets.COREOSBOT_RELENG_TOKEN }}" \
+              "https://api.github.com/user" | jq -r .login)
+          repo=$(echo "${GITHUB_REPOSITORY}" | cut -f2 -d/)
+          git push "https://github.com/${fork_user}/${repo}" "${BRANCH_NAME}" -f
+          echo "FORK_USER=${fork_user}" >> ${GITHUB_ENV}
+      - name: Create PR
+        uses: actions/github-script@v5
```
jlebon: Any reason we don't use `create-pull-request` for this?

jlebon (dismissed):
Very cool! A few minor comments, but overall LGTM.

.github/workflows/add-override.yml
bgilbert: Ah, yeah, that saves a lot of code.

ci/overrides.py
bgilbert: Right.  Any ideas for improvements?  We'll at least catch this in CI when subpackages have versioned Requires on their main package.

ci/overrides.py
bgilbert: How are arch-specific overrides normally specified?  The current code assumes the overrides file can contain any package we're shipping on any arch.

ci/overrides.py
jlebon: Probably cleanest would be to fold some of this into rpm-ostree where it can do a proper depsolve. Not worth trying to solve this here.

ci/overrides.py
jlebon: The previous `manifest-lock.overrides.$arch.yaml` files that 02c41e30 dropped are actually still supported. So in the event we need an arch-specific override, we could still use it.

Auto-graduation isn't enabled there though, which is a gotcha.

jlebon: To clarify, not necessary to block on this.

.github/workflows/add-override.yml
```diff
@@ -0,0 +1,90 @@
...
+          echo "PR_TITLE=${pr_title}" >> ${GITHUB_ENV}
+      - name: Open pull request
+        uses: peter-evans/create-pull-request@v3.8.2
+        with:
+          token: ${{ secrets.COREOSBOT_RELENG_TOKEN }}
+          branch: ${{ env.BRANCH_NAME }}
+          commit-message: ${{ env.COMMIT_TITLE }}
+          push-to-fork: coreosbot-releng/fedora-coreos-config
+          title: ${{ env.PR_TITLE }}
+          body: "Requested by @${{ github.actor }}."
```
jlebon: Can we include the GitHub workflow name here?

jlebon (dismissed):

ci/overrides.py
bgilbert: Okay, yeah, I had misunderstood how the lockfile was used.  Updated to fail on this case.

bgilbert:
Updated!  Also fixed handling of noarch packages, which were being ignored.

jlebon (approved):
LGTM! Nice work.

PR #1450, opened 2022-01-26
kdump.crash: Disable test for aarch64

gursewak1997:
Disabling the kdump.crash for only aarch64 due to the test failure in multi-arch pipeline.
Failure: https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/multi-arch-pipeline/1/

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -12,6 +12,10 @@
     - rawhide
   arches:
     - aarch64
+- pattern: ext.config.kdump.crash
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1075
+  arches:
+    - aarch64
 - pattern: ext.config.podman.dns
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1046
   snooze: 2022-01-31


bgilbert (dismissed):
LGTM

dustymabe:
Feels like we should link to an issue documenting/tracking this issue. Alternatively we add a comment here explaining why the test is invalid on other architectures.

gursewak1997:
> Feels like we should link to an issue documenting/tracking this issue. Alternatively we add a comment here explaining why the test is invalid on other architectures.

Can I just edit and mention aarch64 failure in https://github.com/coreos/fedora-coreos-tracker/issues/1075 considering it is the same error?
Or do you think I should make a separate issue?

dustymabe:
> > Feels like we should link to an issue documenting/tracking this issue. Alternatively we add a comment here explaining why the test is invalid on other architectures.
> 
> Can I just mention aarch64 failure in [coreos/fedora-coreos-tracker#1075](https://github.com/coreos/fedora-coreos-tracker/issues/1075) considering it is the same error?

Seems reasonable - maybe we add that information to the issue and update the title?

> Or do you think I should make a separate issue?

Either way works.

Though now that I think about it more, might be a good idea to put this in the kola denylist (you can deny tests based on architecture there too) instead of marking it here (that file gets looked at a little more often). WDYT?

gursewak1997:
> Though now that I think about it more, might be a good idea to put this in the kola denylist (you can deny tests based on architecture there too) instead of marking it here (that file gets looked at a little more often). WDYT?

I see. It definitely makes more sense to put it in kola denylist and just disable it for aarch64.

dustymabe (dismissed):
LGTM

dustymabe:
can you squash this into one commit?

dustymabe (approved):
LGTM

PR #1452, opened 2022-01-27

coreosbot-releng:
Triggered by remove-graduated-overrides GitHub Action.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,6 +9,12 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  ignition:
+    evr: 2.13.0-3.fc35
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-b238d6fbbf
+      reason: https://github.com/coreos/ignition/issues/1305
+      type: fast-track
   kernel:
     evr: 5.15.10-200.fc35
     metadata:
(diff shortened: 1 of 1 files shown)


bgilbert:
```
[2022-01-27T01:44:45.158Z] Downgraded:
[2022-01-27T01:44:45.158Z]   polkit 0.120-1.fc35.1 -> 0.120-1.fc35
[2022-01-27T01:44:45.158Z]   polkit-libs 0.120-1.fc35.1 -> 0.120-1.fc35
```

dustymabe:
I can only assume this means that NVR comparison gives the answer that `0.120-1.fc35` is newer than `0.120-1.fc35.1` and thus the version in the release day repo is winning

jlebon:
I think the issue here is a little more subtle. When this workflow ran, https://bodhi.fedoraproject.org/updates/FEDORA-2022-da040e6b94 was already pushed to stable so I think it did really see `.1` and favoured it. The problem is that the base lockfiles themselves did not yet match what was in the overrides. They eventually converge via bump-lockfile: [this push](https://github.com/coreos/fedora-coreos-config/commit/449e745ce42a5ea617b0eac80514149a66623be3) bumps polkit to match the override. But CI here ran before that push and so pulled in the old polkit.

This doesn't happen often because overrides usually live long enough for bump-lockfile to run at least once. Thankfully, the CI downgrade check catches it! :tada: 

Probably simplest fix for this is to enhance the script to only consider fast-tracked overrides where the version in the base matches.

Anyway, I've restarted CI here which should pass now.

jlebon:
> Probably simplest fix for this is to enhance the script to only consider fast-tracked overrides where the version in the base matches.

Unless someone else (@bgilbert maybe?) wants to, I can tackle this after https://github.com/coreos/fedora-coreos-config/pull/1441 merges.

jlebon (approved):
Not auto-merging so we sanity-check on top of CI that polkit is no longer being downgraded (see https://github.com/coreos/fedora-coreos-config/pull/1452#issuecomment-1023426078).

PR #1453, opened 2022-01-27
tests: bump timeout for raid1 reprovisioning

jlebon:
We're timing out sometimes on this test because it takes a while for
Ignition to set up RAID1 and Ignition provisioning time is now part of
the overall test timeout[1]. Bump it to 15 mins.

[1] https://github.com/coreos/coreos-assembler/pull/2660


dustymabe (approved):
LGTM

jlebon:
Let's close this for now and re-open if we need it. It's likely not the cosa timeout causing this but heavy I/O on the aarch64 node.

PR #1461, opened 2022-01-28
tests/extensions.package: don't try vim on rawhide

jlebon:
The `vim-enhanced` package has a shared base dep (`vim-common`) so is
subject to the binding issue which the updates archiver solves. Since
the archiver isn't enabled on rawhide, don't try it there.

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -21,3 +21,7 @@ get_ipv4_for_nic() {
     fi
     echo $ip
 }
+
+get_fcos_stream() {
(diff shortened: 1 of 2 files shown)


jlebon:
Ideally we'd be conditional on the archive repo actually existing and being non-empty instead of the stream name, but that's not easy to do without spawning a container and using dnf, which it didn't feel like it was worth the complexity.

dustymabe (approved):
LGTM - thanks for working on this.

dustymabe:
Note we could have the same issue when we switch `next` over to branched, but let's tackle that if/when it happens.

PR #1475, opened 2022-02-01
tests/skopeo: Verify `experimental-image-proxy` exists

cgwalters:
Needed by the ostree-container stack.  I don't expect this test
to fail on FCOS; my main real goal is to trickle this down into
RHCOS where it would currently fail (and I have a fix pending).


miabbott (approved):
LGTM

jlebon (approved):

PR #829, opened 2021-01-27
overlay: propagate multipath config post-firstboot

jlebon:
Because multipath support now works on the second boot, the propagation
code in `coreos-teardown-network.service` didn't work. Split that code
out into a separate dracut module and have it run on subsequent boots.

Resolves: bugzilla.redhat.com/show_bug.cgi?id=1920571

PR #1476, opened 2022-02-02
35coreos-multipath: propagate config on first boot

jlebon:
In #829, we changed multipath config propagation to happen on subsequent
boots to handle upgrading nodes. Back then, we didn't yet support first
boot multipath, so not propagating on first boot didn't matter anyway.

But with #1011 we now support multipath on first boot. So we need to
adapt this so we always propagate, including on first boot.


jlebon:
No issue or RHBZ attached to this. Noticed it while working on https://github.com/coreos/coreos-assembler/pull/2673.

cgwalters (approved):
(Somewhat superficial) LGTM

jlebon:
https://github.com/coreos/coreos-assembler/pull/2690 adds a test for this.

PR #1478, opened 2022-02-03
tests/root-reprovision: bump timeout for reprovision tests to 15m

dustymabe:
All of these tests include a lot of disk I/O and can take a long time.
They all flirt with the 10 minute timeout already and some can go over.
Let's bump them all to give us some breathing room.

Also updated each test to include an explanation for the kola.json
parameters.

### DIFF
--- a/tests/kola/root-reprovision/linear/test.sh
+++ b/tests/kola/root-reprovision/linear/test.sh
@@ -1,5 +1,17 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "minMemory": 4096, "additionalDisks": ["5G", "5G"]}
+# kola: { "platforms": "qemu", "minMemory": 4096, "additionalDisks": ["5G", "5G"], "timeoutMin": 15 }
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+#   - additionalDisks is only supported on qemu.
+# - minMemory: 4096
+#   - Root reprovisioning requires at least 4GiB of memory.
+# - additionalDisks: ["5G", "5G"]
+#   - Linear RAID is setup on these disks.
+# - timeoutMin: 15
+#   - This test includes a lot of disk I/O and needs a higher
+#     timeout value than the default.
+
 set -xeuo pipefail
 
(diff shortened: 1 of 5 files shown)


tests/kola/root-reprovision/linear/test.sh
```diff
@@ -1,5 +1,17 @@
...
-# kola: {"platforms": "qemu", "minMemory": 4096, "additionalDisks": ["5G", "5G"]}
+# kola: { "platforms": "qemu", "minMemory": 4096, "additionalDisks": ["5G", "5G"], "timeoutMin": 15 }
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+#   - additionalDisks is only supported on qemu.
+# - minMemory: 4096
+#   - This test requires at least 4GiB of memory.
+# - additionalDisks: ["5G", "5G"]
+#   - Add two extra disks to use for the test
```
jlebon: Hmm, shouldn't we explain *why* instead of describing the knob itself?

E.g.

```suggestion
#   - Linear RAID is setup on these disks.
```

tests/kola/root-reprovision/linear/test.sh
```diff
@@ -1,5 +1,17 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "minMemory": 4096, "additionalDisks": ["5G", "5G"]}
+# kola: { "platforms": "qemu", "minMemory": 4096, "additionalDisks": ["5G", "5G"], "timeoutMin": 15 }
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+#   - additionalDisks is only supported on qemu.
+# - minMemory: 4096
+#   - This test requires at least 4GiB of memory.
```
jlebon:
```suggestion
#   - Root reprovisioning requires at least 4GiB of memory.
```

Edit: similarly for other tests.

dustymabe:
Made some updates. Let me know if they look good now.

jlebon (approved):
Looks great, thanks!

PR #1479, opened 2022-02-03
[rawhide] overrides: pin rpm-4.17.0-6.fc36

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,18 +9,18 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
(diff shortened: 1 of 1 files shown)


dustymabe:
Doesn't look like this is working as expected. Maybe user error. @bgilbert do you want to investigate or should I manually open a PR for this one?

dustymabe:
Oh. I bet I know. The rawhide branch doesn't have any lockfiles. Only the override file, so the tool can't consult them to know what subpackages to update in the override file.

bgilbert:
https://github.com/coreos/fedora-coreos-config/pull/1481 adds error handling for this case.

PR #1481, opened 2022-02-03
ci/overrides: fail if we didn't fast-track/pin anything

bgilbert:
If none of the detected binary packages are listed in the lockfiles, fail rather than performing a no-op rewrite of the override file and appearing to succeed.

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -107,6 +107,8 @@ def do_fast_track(args):
             )
             if args.reason:
                 overrides[n]['metadata']['reason'] = args.reason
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Should the workflow support passing in the exact binary RPMs as a workaround?

bgilbert:
In that case, the tool wouldn't perform much automation and couldn't check its work.  I think it makes more sense to stick with manual PR and review in that event?

bgilbert:
CI doesn't test this.  Merging over red.

PR #1483, opened 2022-02-04
tests/kola/root-reprovision/swap-before-root: set `allowConfigWarnings`

bgilbert:
Starting with https://github.com/coreos/coreos-assembler/pull/2686, kola fails any test that produces warnings when parsing Ignition or Butane configs unless configured otherwise.

Requires https://github.com/coreos/coreos-assembler/pull/2686.


dustymabe (approved):
LGTM

PR #1484, opened 2022-02-04
manifest: Add `container-cmd`

cgwalters:
Builds on https://github.com/coreos/rpm-ostree/pull/3402

Relates to https://github.com/coreos/coreos-assembler/issues/2685

Basically, source of truth for the CMD moves from being hardcoded
in cosa hackiliy to being part of the ostree commit, which means
it survives a full round trip of
ostree → container → ostree → container

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -11,6 +11,9 @@ rojig:
 add-commit-metadata:
   fedora-coreos.stream: ${stream}
 
+container-cmd:
+  - /usr/bin/bash
+
 include: fedora-coreos-base.yaml
 conditional-include:
   - if: prod == false


cgwalters:
OK, rebased :surfer: and this should be good to merge now that we have a new enough rpm-ostree.

cgwalters:
OK CI flaked on the journal bit, but this can't break kola tests anyways.

cheesesashimi (dismissed):

cheesesashimi:
Not sure If I have these perms for this repo, but I'll give it a try:

/lgtm
/approve

jlebon:
Seems like this would be shared across all streams, right? In that case, maybe let's put it in `manifests/fedora-coreos.yaml` instead?

cgwalters:
> Seems like this would be shared across all streams, right? In that case, maybe let's put it in manifests/fedora-coreos.yaml instead?

Ah yes, thank you!

jlebon (approved):

PR #1489, opened 2022-02-06
manifests/bootable-rpm-ostree: drop `linux-firmware-whence`

bgilbert:
On Fedora, `linux-firmware` requires `linux-firmware-whence`, so we don't need to specify it.  On RHEL 8, `linux-firmware-whence` doesn't exist.


dustymabe (approved):
LGTM

PR #1490, opened 2022-02-06
ci/overrides: correctly add noarch overrides

bgilbert:
For noarch binary packages, we need to use the `evra` key and append `.noarch` to the version.

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -187,8 +187,9 @@ def get_source_nvrs(update):
 
 
 def get_binary_packages(source_nvrs):
-    '''Return name => info dict for the specified source NVRs.  A binary
-    package is included if it is in the manifest lockfiles.'''
+    '''Return name => info dict for the specified source NVRs.  The info
+    dict contains "evr" for archful packages and "evra" for noarch ones.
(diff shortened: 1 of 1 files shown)


ci/overrides.py
```diff
@@ -197,11 +197,13 @@ def get_binary_packages(source_nvrs):
     accepted_in_arch = {}
     client = koji.ClientSession(KOJI_URL)
 
+    archful = lambda arch: arch != 'noarch'
```
cgwalters, line 197: This is an interesting idiom - I like it.  I guess one doesn't see it in statically typed languages as the extra typing required from a nested helper function hurts readability more or something?

cgwalters (approved):
(Somewhat superficial) LGTM

jlebon (approved):
Yeah, this is an awkward wart right now with how lockfiles are set up. Ideally we should be able to use `evr` consistently even with noarch packages. Thanks for this!

ci/overrides.py
bgilbert: Hmm, maybe.  I've also used it in Rust with closures.

PR #1492, opened 2022-02-07
Re-pin authselect to new version, drop firstboot hack

cgwalters:
Revert "manifest: run authselect on first boot to workaround bug"

This reverts commit bec4c57a0dbb7c28c6aa98b6d89e9a9689027587.

---

Bump authselect pin

Due to compose issues (I think) the fixed authselect isn't in
the rawhide rpm-md repo.  Pin to the fixed version for now to ensure
we get it (I hope).

---


manifest-lock.overrides.yaml
```diff
@@ -10,12 +10,12 @@
 
 packages:
   authselect:
-    evr: 1.3.0-3.fc36
+    evr: 1.3.0-9.fc36
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1057
       type: pin
```
jlebon: I think these need to be changed to `fast-track` now, no?

dustymabe: yep

cgwalters: Right, thank you!

jlebon (dismissed):
Woohoo!

cgwalters:
This needs https://github.com/coreos/coreos-assembler/pull/2693

dustymabe:
rebased to fix merge conflict - also added link to bodhi.

dustymabe (approved):

PR #1495, opened 2022-02-08
denylist: extend snooze for ostree.hotfix on rawhide/aarch64

dustymabe:
We should get some help here soon but for now this is still
failing.


dustymabe:
killed CI because this applies to rawhide.

travier (approved):

PR #1498, opened 2022-02-08
s390x: Skip incompatible kola tests

jschintag:
Skip tests that are new and currently failling on s390x. These tests need further investigation wether or not they can be adapted on s390x.
Also skip grub specific part of the reboot test, as s390x does not use grub.

### DIFF
--- a/tests/kola/kdump/crash
+++ b/tests/kola/kdump/crash
@@ -1,6 +1,8 @@
 #!/bin/bash
-# kola: {"minMemory": 4096, "tags": "skip-base-checks"}
+# kola: {"minMemory": 4096, "tags": "skip-base-checks", "architectures": "!s390x"}
 # https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/
+# https://github.com/coreos/fedora-coreos-config/issues/1500
+# - kdump.service is failling on s390x
 
 set -xeuo pipefail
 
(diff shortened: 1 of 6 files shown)


mike-nguyen (dismissed):
LGTM

dustymabe:
Hey @jschintag - can you add a reason for adding that argument to each kola.json fragment? If we need more investigation can we open an issue to track that and link to it?

dustymabe:
Here's an example of doing that:

https://github.com/coreos/fedora-coreos-config/blob/d75822c6c17d1dd0847719c603700e42f1774fe1/tests/kola/root-reprovision/luks/test.sh#L8-L9

jschintag:
> Hey @jschintag - can you add a reason for adding that argument to each kola.json fragment? If we need more investigation can we open an issue to track that and link to it?

@dustymabe I added the comments to the kola.json fragments. The Issues for this are:
https://github.com/coreos/fedora-coreos-config/issues/1499 Network Manager Issue
https://github.com/coreos/fedora-coreos-tracker/issues/1267 kdump.service not working

dustymabe (approved):
LGTM

miabbott (approved):
LGTM

jlebon (approved):

travier:
Sorry for being late here, bur are all of those tests really incompatible on s390x? If the test are only "not working" right now for some unknown reasons, they should probably end up in the kola-denylist instead. I think that at least the kdump test is in this case.

dustymabe:
I'm OK with denylist entries instead. Either way I think (hopefully) we'll find real reasoning in the investigation in coreos/fedora-coreos-config#1499 and coreos/fedora-coreos-tracker#1267

jschintag:
I put the restriction into the tests for now, as they haven't to my knowledge ever worked on s390x.
Therefore i see the 2 Issues more as a development item rather than a bug fix.

PR #1501, opened 2022-02-09
ci/overrides: support pin/fast-track on mechanical streams

bgilbert:
If there are no lockfiles in the current branch, fetch generated lockfiles from the latest successful CI builds of the branch on each architecture.

Suggested by @dustymabe.


dustymabe (approved):
LGTM

PR #1505, opened 2022-02-09
denylist: snooze ext.config.chrony.dhcp-propagation in rawhide

dustymabe:
SELinux denials are causing some issues and we need to have them looked
at. See https://github.com/coreos/fedora-coreos-tracker/issues/1097


dustymabe:
killing CI because it won't test rawhide.

jlebon (approved):

travier (approved):

PR #1508, opened 2022-02-10
denylist: apply snoozes to the branched stream as well

dustymabe:
Since we're re-enabling `branched` we need these to apply
there too.


dustymabe:
Killed CI since it won't test this change.

bgilbert (approved):
LGTM

PR #1510, opened 2022-02-10
[rawhide] Add @CoreOS/continuous COPR repo

jlebon:
Rawhide moves fast. And our packages adapt quickly. Add the
@CoreOS/continuous packages to shorten the cycle.

Part of https://github.com/coreos/fedora-coreos-tracker/issues/910.

### DIFF
--- a/copr-coreos-continuous.repo
+++ b/copr-coreos-continuous.repo
@@ -1,0 +1,8 @@
+[copr-coreos-continuous]
+name=@CoreOS/continuous COPR
+baseurl=https://download.copr.fedorainfracloud.org/results/@CoreOS/continuous/fedora-$releasever-$basearch/
+type=rpm-md
+gpgcheck=1
+gpgkey=https://download.copr.fedorainfracloud.org/results/@CoreOS/continuous/pubkey.gpg
+repo_gpgcheck=0
+enabled=1
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -10,6 +10,7 @@ rojig:
 
 repos:
   - fedora-rawhide
+  - copr-coreos-continuous
 
 add-commit-metadata:
   fedora-coreos.stream: rawhide


dustymabe:
Hmm. Shouldn't we just keep the `rawhide` stream purely rawhide (with overrides when problems arise)? I think it will get confusing otherwise.

cgwalters:
Agree we should not conflate rawhide and continuous.  (There's deeper things going on here around what rawhide is and should be, but that aside)

jlebon:
Hmm, to me this makes rawhide more valuable because we can fix bugs more quickly and we get faster feedback.

Anyway, we can discuss this as part of https://github.com/coreos/fedora-coreos-tracker/issues/910 in the next community meeting.

cgwalters:
The thing is here is it's actually combinatorial - it makes sense to apply the continuous builds on top of rawhide *and* at least next/testing.  IOW I think we want e.g. `testing-continuous` at least.

jlebon:
Right, but the problem with that is exactly the combinatorial aspect. More streams to watch, more failures to account for, more PRs, more process, etc... So I think it makes sense to combine where possible and otherwise not. What I'm arguing for in https://github.com/coreos/fedora-coreos-tracker/issues/910 is to combine it for rawhide but not for testing-devel, because there I think the cost is worth it. I'm not confident I can say the same for the cost of two rawhide streams.

jlebon:
I forgot to start collecting things which this would have help with faster:
- https://github.com/coreos/rpm-ostree/pull/3535
- https://github.com/coreos/rpm-ostree/pull/3488
- https://github.com/coreos/rpm-ostree/pull/3453
- https://github.com/coreos/rpm-ostree/pull/3406
- https://github.com/coreos/rpm-ostree/pull/3399

(This is just from looking at rpm-ostree PRs.)

cgwalters:
Yeah, fair.  But, I think basically all those issues *also* affect non-CoreOS things in Fedora...so, until we:

1) Base those things on FCOS
2) Actually just do continuous/automated builds into rawhide

I don't think we're really out of the problem space.

jlebon:
Fair. Are you suggesting e.g. hooking up Packit to have continuous builds in rawhide? I'd be OK with that, but the amount of churn these bot builds generate in Koji and Bodhi is unfortunate.

cgwalters:
If I had my way, I'd basically abolish rawhide, and actually unify at least the "core" packages into a single git repository that had PR gating.

Then, things like the glibc change could easily gate on our tests, *and* it'd be easily possible to have an atomic commit change multiple things at once.

The replacement for rawhide is feature branches on top of the stable base - the same way we develop upstream projects!

But...for now?  Dunno.  I guess I'd say for now we just manually cut git snapshots to rawhide as needed?

jlebon:
OK, let's close this.

PR #1511, opened 2022-02-10
tests/default-network-behavior-change: Adapt to RHCOS 4.11

cgwalters:
We're going to rebase RHCOS 4.11, which picks up a new NM,
and the new case is actually the same.

BTW, minor rant here: We really need to stop defaulting to writing conditionals
that do `if is_fcos() {} else if is_rhcos() {}` because about 70%
of the time, this is actually trying to test "RHEL 8" versus "Fedora".

And in this case, what we want to dispatch on is the RHEL8 minor
version.  Or even the NetworkManager version.

Anyways for now, because we haven't updated the `redhat-release-coreos`
package because doing so is painful, dispatch on the OCP version.

### DIFF
--- a/tests/kola/networking/default-network-behavior-change/test.sh
+++ b/tests/kola/networking/default-network-behavior-change/test.sh
@@ -13,7 +13,7 @@ set -xeuo pipefail
 . $KOLA_EXT_DATA/commonlib.sh
 
 # EXPECTED_INITRD_NETWORK_CFG1
-#   - used on FCOS 35+ releases
+#   - used on Fedora 35+ and RHEL 8.5+ releases
 EXPECTED_INITRD_NETWORK_CFG1="[connection]
 id=Wired Connection
 uuid=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
@@ -42,7 +42,7 @@ method=auto
 [user]
 org.freedesktop.NetworkManager.origin=nm-initrd-generator"
 # EXPECTED_INITRD_NETWORK_CFG2
-#   - used on all RHCOS releases
+#   - used on older RHCOS <= 4.10 releases
 EXPECTED_INITRD_NETWORK_CFG2="[connection]
 id=Wired Connection
 uuid=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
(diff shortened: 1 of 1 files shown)


cgwalters:
> BTW, minor rant here: We really need to stop defaulting to writing conditionals
that do if is_fcos() {} else if is_rhcos() {} because about 70%
of the time, this is actually trying to test "RHEL 8" versus "Fedora".

And to repeat myself again (from e.g. https://github.com/coreos/fedora-coreos-config/pull/1405#discussion_r800690767 ) - we'll soon have "RHCOS9" - and if these conditionals are testing the RHEL version or the sub-component (NM) version, they will Just Work.

miabbott (dismissed):
Looks sane to me

travier:
LGTM but I think we should discard all the version selection logic now that everybody is using the same config.

cgwalters:
Yeah true, in general since we fork this repo per RHCOS, there's not a lot of value in having the new code support older versions.

dustymabe:
> LGTM but I think we should discard all the version selection logic now that everybody is using the same config.

I'd prefer to leave the logic in place for when things change in the future.

dustymabe:
We need to update the comments above the definitions of the variables at the top of the script.

dustymabe:
> BTW, minor rant here: We really need to stop defaulting to writing conditionals
> that do `if is_fcos() {} else if is_rhcos() {}` because about 70%
> of the time, this is actually trying to test "RHEL 8" versus "Fedora".

Yeah. There is some gray area here. In this case I decided up front that I wanted to know anytime the networking configuration changed in a stream (i.e. 4.9, etc). When the failure occurs we can then decide if that change is OK mid stream or not.

cgwalters:
> I'd prefer to leave the logic in place for when things change in the future.

OK.

> We need to update the comments above the definitions of the variables at the top of the script.

Done

tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -13,7 +13,7 @@ set -xeuo pipefail
 . $KOLA_EXT_DATA/commonlib.sh
 
 # EXPECTED_INITRD_NETWORK_CFG1
-#   - used on FCOS 35+ releases
+#   - used on Fedora 35+ and RHEL 8.5+ releases
```
dustymabe, line 16: thanks! update line 45 too?

dustymabe:
> When the failure occurs we can then decide if that change is OK mid stream or not.

IOW - this PR is us deciding that this change is safe for 4.11. If any work needed to be done in other areas we would do it now.

tests/kola/networking/default-network-behavior-change/test.sh
cgwalters: OK, though the comments are duplicating the conditionals; I think it's better to have a single source of truth.  i.e. I'd lean towards deleting the comments instead.  But for now, updated that line too

dustymabe (approved):
LGTM

cgwalters:
Unrelated CI failure, I tested this locally with FCOS and rhcos.

PR #1512, opened 2022-02-10
skipping EFI on arch ppc64le and s390x

mrcxavier:


cgwalters (dismissed):
Makes sense.  Would be nice to have a little bit more information in the commit message, even just something like:

> I noticed this test was failing on ppc64le.

Or so?  At least for future commits.

ravanelli (changes requested):
@mrcxavier Please, keep only one commit with the changes we need.

ravanelli (approved):
LGTM

PR #1517, opened 2022-02-14
denylist: extend snoozes for rawhide, branched

dustymabe:
These are still a problem.


dustymabe:
Killed CI as it won't test this change.

jlebon (approved):

PR #1519, opened 2022-02-14
[rawhide] manifest: add aardvark-dns to package list for podman v4

dustymabe:
aardvark-dns, which provides name resolution based on container name,
is recommended by netavark but not required. We need to name is so
it gets pulled in.


dustymabe:
also dropped the `tpm2-tss` override.

dustymabe:
added another commit to drop naming `authselect` in our package lists because it's now required by `pam`.

jlebon (approved):

PR #1520, opened 2022-02-14
denylist: snooze ext.config.toolbox test in rawhide

dustymabe:
There is no registry.fedoraproject.org/fedora-toolbox:37 in the
registry. Let's snooze the test for a month and hopefully it will
exist by then.

See https://github.com/coreos/fedora-coreos-tracker/issues/1103

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -58,3 +58,8 @@
   snooze: 2022-02-21
   streams:
     - rawhide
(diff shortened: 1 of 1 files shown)


dustymabe:
Killing CI since it won't test this change.

saqibali-2k (dismissed):

saqibali-2k (approved):

PR #1521, opened 2022-02-14
denylist: snooze `ext.config.extensions.module` for now

jlebon:
See https://github.com/coreos/fedora-coreos-tracker/issues/1104.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -53,3 +53,8 @@
   streams:
     - rawhide
     - branched
+- pattern: ext.config.extensions.module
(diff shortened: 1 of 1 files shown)


dustymabe:
Want to make this a snooze so we have a reminder to revisit or do you want to track it closer with the maintainers and un-deny it then?

dustymabe:
killed CI because it won't test this.

jlebon:
Yeah OK, let's just snooze.

dustymabe (approved):

PR #1526, opened 2022-02-16
35coreos-live: support `coreos.live.rootfs_url=tftp://`

bgilbert:
As discussed in https://github.com/coreos/fedora-coreos-tracker/issues/1055, the `rootfs_url` karg is preferred over initrd concatenation due to its lower memory footprint and generally better performance.  We originally chose to disallow its use with TFTP to reduce the feature matrix, but that forces users without HTTP servers to fall back to concatenation.  Enable TFTP so `rootfs_url` can be used in all environments.


bgilbert:
This will need a docs update.

jlebon (approved):
Were you able to check manually that TFTP works as expected?

bgilbert:
Yup, it works!

PR #1527, opened 2022-02-16
[rhcos-4.9] ignition-ostree-uuid-root.service: add stricter ordering

jlebon:
In the multipath case, we're mounting
`/dev/disk/by-label/dm-mpath-root`, not `by-label/root`.

Also order against `sysroot.mount` because in the general case, there's
actually nothing that prevents the restamping to run at the same time as
the mount AFAICT if the `root` karg points to some other device path we
didn't predict.

See: https://github.com/openshift/os/issues/683 (which hopefully this
fixes).

(cherry picked from commit 11a11bcd530a4e67b778dcaccca3cbe189715b00)


miabbott (approved):

miabbott:
Not sure the CI is going to meaningfully test this, so if you have tested it locally, then I'm OK with merging.

jlebon:
Tested locally and ran `iso-install` a bunch of times to be safe.

PR #1529, opened 2022-02-17
workflows: add job to check for unexpected changes in promotion diff

bgilbert:
When we get a promotion PR in the `next`/`testing`/`stable` branches, compare the new code with a checkout of the commit hash from the PR title and annotate the differences.  For `manifest.yaml`, compare with the base branch instead, since the branch we're promoting from will have a different config.

### DIFF
--- a/.github/workflows/promotion-diff.yml
+++ b/.github/workflows/promotion-diff.yml
@@ -1,0 +1,44 @@
+---
+name: Check promotion diffs
+on:
+  pull_request:
+    branches: [next, testing, stable]
+permissions:
+  contents: read
+
+jobs:
+  promotion-diff:
+    name: Check promotion diffs
+    runs-on: ubuntu-latest
+    # Only run if this looks like a promotion PR
+    if: "startsWith(github.event.pull_request.title, 'tree: promote changes from')"
+    steps:
+      - name: Get base commit hash
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/promotion-diff.yml
```diff
@@ -0,0 +1,44 @@
...
+    name: Check promotion diffs
+    runs-on: ubuntu-latest
+    # Only run if this looks like a promotion PR
+    if: "startsWith(github.event.pull_request.title, 'tree: promote changes from')"
+    steps:
+      - name: Get base commit hash
+        env:
+          COMMIT_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          echo "ORIGIN_COMMIT=$(echo ${COMMIT_TITLE} | awk '{print $NF}')" >> $GITHUB_ENV
```
jlebon, line 20: We could also consider in the future adding some structured content in the commit message itself.

jlebon (approved):
LGTM!

.github/workflows/promotion-diff.yml
bgilbert: Yup, makes sense to me.

bgilbert:
CI doesn't test this; failures are unrelated.

PR #1530, opened 2022-02-17
denylist: snooze the coreos.boot-mirror.luks test

dustymabe:
This issue is back again.

See https://github.com/coreos/fedora-coreos-tracker/issues/1092#issuecomment-1043059521


dustymabe:
Killed CI. It won't test this change.

jlebon (approved):
Feels like the commit message should specify it's only on f36+. LGTM otherwise!

PR #1531, opened 2022-02-21
overlay: add 'partprobe' and retry verification of unique boot

nikita-dubrovskii:
`rdcore` first tries the rereading of partition table by execution of the `partprobe` tool, so add it.
After change of partitions' UUIDs kernel and userspace have different info, to mitigate that try to verify the unique-boot
up to 3 times.

Required by: https://github.com/coreos/coreos-installer/pull/785
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1105

### DIFF
--- a/manifests/system-configuration.yaml
+++ b/manifests/system-configuration.yaml
@@ -36,3 +36,4 @@ packages:
   # Boost starving threads
   # https://github.com/coreos/fedora-coreos-tracker/issues/753
   - stalld
+  - /usr/sbin/partprobe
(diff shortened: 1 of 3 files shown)


bgilbert:
In general I don't think we should be working around issues by adding retries and sleeps.  In this case especially, since we control the code we're calling, it seems like any problems should be handled inside `rdcore`.

cc @jlebon

nikita-dubrovskii:
> In this case especially, since we control the code we're calling, it seems like any problems should be handled inside `rdcore`.
 
ok, agree

PR #1533, opened 2022-02-21
tests/default-network-behavior-change: update for NM 1.36.0

dustymabe:
Upstream in NM they changed it so that entries in a keyfile
don't get written out if they were empty strings [1]. Let's
update our definitions accordingly.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/27c33d15efdf974a085590aa4077ab76862c1854


jmarrero (approved):
lgtm

dustymabe:
Tested locally against `testing-devel` and `rawhide` - both pass.

PR #1538, opened 2022-02-23
manifests: various cleanups using conditional includes and templates

jlebon:
Just a few cleanups I did to showcase https://github.com/coreos/rpm-ostree/pull/3439.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,15 +1,8 @@
-ref: fedora/${basearch}/coreos/testing-devel
-include: manifests/fedora-coreos.yaml
-
 variables:
+  stream: testing-devel
   prod: false
 
 releasever: 35
-
-rojig:
-  license: MIT
-  name: fedora-coreos
-  summary: Fedora CoreOS testing-devel
 
 repos:
   # These repos are there to make it easier to add new packages to the OS and to
@@ -19,5 +12,4 @@ repos:
   - fedora
   - fedora-updates
 
-add-commit-metadata:
-  fedora-coreos.stream: testing-devel
+include: manifests/fedora-coreos.yaml
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,7 +1,7 @@
 ref: fedora/${basearch}/coreos/testing-devel
 include: manifests/fedora-coreos.yaml
 
-releasever: "35"
+releasever: 35
 
 rojig:
   license: MIT
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,6 +1,9 @@
 ref: fedora/${basearch}/coreos/testing-devel
 include: manifests/fedora-coreos.yaml
 
+variables:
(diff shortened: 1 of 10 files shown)


manifest.yaml
```diff
@@ -1,7 +1,7 @@
 ref: fedora/${basearch}/coreos/testing-devel
 include: manifests/fedora-coreos.yaml
 
-releasever: "35"
+releasever: 35
```
bgilbert, line 5: Will this break [this line](https://github.com/coreos/fedora-coreos-config/blob/7413ad4049882ed2c873db954d6147d8d43f1af0/ci/overrides.py#L159)?

manifests/fedora-coreos.yaml
```diff
@@ -43,12 +43,6 @@ etc-group-members:
   # https://github.com/projectatomic/rpm-ostree/issues/49
```
bgilbert, line 52: commit message: `specificied`

manifest.yaml
```diff
@@ -1,16 +1,9 @@
-ref: fedora/${basearch}/coreos/testing-devel
-include: manifests/fedora-coreos.yaml
-
 variables:
+  stream: testing-devel
```
bgilbert, line 2: We'll need to update `get_stream()` in `overrides.py`.

manifest.yaml
jlebon: Good call. Trying it locally, it seems to work. Looks like the API stringifies it: https://github.com/rpm-software-management/dnf/blob/c68038ff1254c91aa90e1dcb640a933c9194aa98/dnf/conf/config.py#L430

manifests/disable-zincati-pinger.yaml
```diff
@@ -0,0 +1,2 @@
+overlay-layers:
+  - overlay.d/16disable-zincati-pinger
```
dustymabe:
```suggestion
  - overlay.d/16disable-zincati-and-pinger
```

manifests/iptables-nft.yaml
```diff
@@ -0,0 +1,16 @@
+# Scripts for opting into staying on iptables-legacy after migration. Remove
+# after the next barrier release.
+ostree-layers:
+  - overlay/35coreos-iptables
+
+# Default to iptables-nft. Otherwise, legacy wins. We can drop this once/if we
+# remove iptables-legacy.
+- |
```
dustymabe, line 9: missing `postprocess:`

dustymabe:
maybe for `manifests: add iptables-nft.yaml for f36+` let's drop the denylist entry for the firewall test.

dustymabe (dismissed):
LGTM - one suggestion.

dustymabe:
I guess we'll need to PR all the branches to update the manifest.yaml too.

dustymabe (approved):

PR #1539, opened 2022-02-23
Verify `kata-osbuilder-generate.service` as part of the default presets

HuijingHei:
See:
 - https://bugzilla.redhat.com/show_bug.cgi?id=1941342
 - https://github.com/openshift/os/pull/524

### DIFF
--- a/tests/kola/extensions/kata-osbuilder
+++ b/tests/kola/extensions/kata-osbuilder
@@ -1,0 +1,27 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+# check that `kata-osbuilder-generate.service` as part of the default presets
+# - fcos: defined in /usr/lib/systemd/system-preset/90-default.preset
+# - rhcos: defined in /usr/lib/systemd/system-preset/45-rhcos-extensions.preset
+# See:
+# - https://bugzilla.redhat.com/show_bug.cgi?id=1941342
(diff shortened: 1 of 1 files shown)


tests/kola/extensions/kata-osbuilder
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+# check that `kata-osbuilder-generate.service` as part of the default presets
+# - fcos: defined in /usr/lib/systemd/system-preset/90-default.preset
+# - rhcos: defined in /usr/lib/systemd/system-preset/45-rhcos-extensions.preset
```
miabbott, line 5: I think this test can be restricted to just RHCOS because the change was specific to that platform and is applicable for the RHCOS extension case of `sandboxed-containers`.  (I don't believe we have done anything to enable `sandboxed-containers` in FCOS)

The service is enabled in FCOS through the default presets in Fedora; it looks like it was included 2 years ago - https://src.fedoraproject.org/rpms/fedora-release/c/8d0ed8b287550b16e7a7eb35c162e7bc51284b99

Though since we don't include `sandboxed-containers` in FCOS, it is a no-op

HuijingHei: Thanks @miabbott for your suggestion, I will change it

PR #1540, opened 2022-02-23
overlay.d: add service to check for unique-boot after ignition

nikita-dubrovskii:
After ignition's stages partition table could be modified, and therefore
kernel may have an old picture of disks and partitions. During verification
we force kernel to reread the partition table for each disk and call udevsettle,
which in turn leads to modification of "/dev/disk/by-*/*" symlinks and systemd
may send SIGTERM to services with "Require=dev-disk-by\x2dlabel-boot.device".
To avoid that this checks for the unique fs labeled 'boot' after ignition's
disks stage was finished but before mounting stage.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1105

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-unique-boot.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-unique-boot.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=CoreOS Ensure Unique Boot Filesystem
+ConditionPathExists=/etc/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+# That's a weak dependency, so service won't fail if boot dissaperears
+Wants=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device
+
+# Start after ignition has finished with disks but before mounting them
+After=ignition-disks.service
+Before=ignition-mount.service
+Before=ignition-ostree-uuid-root.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/rdcore verify-unique-fs-label --rereadpt boot
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -64,6 +64,7 @@ install() {
     install_ignition_unit "coreos-boot-edit.service" \
         "ignition-diskful.target"
 
+    install_ignition_unit coreos-ignition-unique-boot.service ignition-diskful.target
     install_ignition_unit coreos-unique-boot.service ignition-diskful.target
     install_ignition_unit coreos-ignition-setup-user.service
 }


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-unique-boot.service
```diff
@@ -0,0 +1,25 @@
+[Unit]
+Description=Ensure Filesystem Labeled 'boot' Is Unique After Ignition
+ConditionPathExists=/etc/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+# On s390x zVM with multipath day1 systemd sends SIGTERM, retry the check before giving up
```
bgilbert: Does this happen because we're racing with something that causes udev events?  If so, can we explicitly sequence with that thing and remove the retries here?

nikita-dubrovskii: After recent rpms update even that doesn't work. Fix for s390x (below) breaks x86, still debugging 

```
After=ignition-disks.service
Before=ignition-mount.service
```

cgwalters (dismissed):
I didn't do an in-depth analysis here, but this seems sane to me.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
```diff
@@ -14,8 +14,6 @@ karg() {
     echo "${value}"
 }
 
-rdcore verify-unique-fs-label --rereadpt boot
```
jlebon: Can we keep this but just drop the `--rereadpt`? Also, let's move it to just after the `mount` call. (This should've been part of the first review on this; feel free to leave as a follow-up.)

This runs much later than `ignition-mount` and so could in theory catch any additional devices that show up now. It's unlikely, but you could also imagine a boot device showing up after this call, but before we mount by label. So moving it down closes that race. (It could be interesting to have `rdcore verify-unique-fs-label` actually return the device path that we can that pass to `mount` to make this even more clear.) This works in conjunction with `rdcore bind-boot` which takes the boot mount to which to bind.

Conceptually, the `mount` we do here is the "commit point": theoretically after that point, we could tolerate other `boot` filesystems being present just fine because everything is strongly bound together.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-unique-boot.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Ensure Filesystem Labeled 'boot' Is Unique After Ignition
```
jlebon: Maybe:

```suggestion
Description=CoreOS Ensure Unique Boot Filesystem
```

?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-unique-boot.service
```diff
@@ -0,0 +1,16 @@
...
+ConditionPathExists=/etc/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+ConditionKernelCommandLine=ignition.firstboot
+
+# Start after ignition has finished with disks but before mounting them
+After=ignition-disks.service
+Before=ignition-mount.service
+Before=initrd.target
```
jlebon: Shouldn't need this.

jlebon: Let's additionally do

```
Before=ignition-ostree-uuid-root.service
```

here?

This will also automatically order us before sysroot mounting and transposefs restoration.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -61,6 +61,7 @@ install() {
     install_ignition_unit "coreos-boot-edit.service" \
         "ignition-diskful.target"
 
+    install_ignition_unit coreos-ignition-unique-boot.service initrd.target
```
jlebon: We only want to do this on first boot, and only on machines with a boot device.

```suggestion
    install_ignition_unit coreos-ignition-unique-boot.service ignition-diskful.target
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-unique-boot.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Ensure Filesystem Labeled 'boot' Is Unique After Ignition
+ConditionPathExists=/etc/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+ConditionKernelCommandLine=ignition.firstboot
```
jlebon: Shouldn't need this after moving to `ignition-diskful.target` in `module-setup.sh`.

jlebon (dismissed):
Thanks @nikita-dubrovskii and @sohankunkerkar!

jlebon:
`coreos.boot-mirror` is failing with

```
Starting CoreOS Ensure Unique Boot Filesystem...
Error: System has 0 devices with a filesystem labeled 'boot': []
coreos-ignition-unique-boot.service: Main process exited, code=exited, status=1/FAILURE
```

which is interesting. This runs right after `ignition-disks` which does do `udevadm settle`. And `--rereadpt` also does a `udevadm settle`, though based on [this](https://github.com/coreos/coreos-installer/blob/3273f49ae7f6577a18964bcc878b64908de89c23/src/blockdev.rs#L1079-L1083) it's possible we didn't actually wait for anything at all.

Hmm, but actually I think we could have `verify-unique-fs-label` just keep waiting until the label shows up after re-reading all partition tables. This would be the equivalent of `Requires=dev-disk-by\x2dlabel-boot.device` and `After=dev-disk-by\x2dlabel-boot.device`, except we can't do that directly otherwise we'll hit the exact issue this PR intended to solve.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-unique-boot.service
```diff
@@ -0,0 +1,19 @@
...
+
+# Start after ignition has finished with disks but before mounting them
+After=ignition-disks.service
+Before=ignition-mount.service
+Before=ignition-ostree-uuid-root.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStartPre=/usr/sbin/udevadm trigger --settle --subsystem-match block
```
jlebon: Ouch. I'm hoping we don't resort to `udevadm trigger` for this. Does it work without it if we have `After=dev-disk-by\x2dlabel-boot.device`?

nikita-dubrovskii: No. Unfortunately that's the only working solution so far. `Wants/After` doesn't help that much

nikita-dubrovskii:
using this with https://github.com/coreos/coreos-installer/pull/813 fixes CI (locally)

jlebon:
This is pending a new coreos-installer release.

jlebon (approved):
We have new enough coreos-installer now, so this is unblocked.
LGTM!

jlebon:
I restarted CI here to make sure it tests on top of the latest. Let's rerun it multiple times before merging this to try to make sure we're not introducing flakiness again. Edit: I've now enabled auto-merge after restarting the test three times.

Ideally, we should add tests for both `coreos-unique-boot.service` and this new service. I think we already have some tests in kola which expect failure in the initrd. OK, filed https://github.com/coreos/coreos-assembler/issues/2953 for this.

PR #1541, opened 2022-02-24
tests/iptables: adapt for RHCOS and f36+

jlebon:
Don't make the test stream-based, but rather releasever-based. Handle
RHCOS correctly, which has long already switched to iptables-nft.

While we're here, add a bunch of query functions to the commonlib.
There are some tests in which we could use these in a follow-up.

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -25,3 +25,26 @@ get_ipv4_for_nic() {
 get_fcos_stream() {
     rpm-ostree status -b --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]'
 }
+
+is_fcos() (
+    source /etc/os-release
+    [ "${ID}" == "fedora" ] && [ "${VARIANT_ID}" == "coreos" ]
+)
+
+# Note when using this, you probably also want to check `get_rhel_maj_ver`.
+is_rhcos() (
+    source /etc/os-release
+    [ "${ID}" == "rhcos" ]
+)
+
+get_fedora_ver() (
+    source /etc/os-release
+    if is_fcos; then
+        echo "${VERSION_ID}"
+    fi
+)
+
+get_rhel_maj_ver() {
+    source /etc/os-release
+    echo "${RHEL_VERSION%%.*}"
+}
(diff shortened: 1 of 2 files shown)


tests/kola/data/commonlib.sh
```diff
@@ -25,3 +25,24 @@ get_ipv4_for_nic() {
...
+# Note when using this, you probably also want to check `get_rhel_maj_ver`.
+is_rhcos() (
+    source /etc/os-release
+    [ "${ID}" == "rhcos" ]
+)
+
+get_fcos_ver() (
+    source /etc/os-release
+    echo "${VERSION_ID}"
+)
```
dustymabe: Wondering if we should make this return an error if someone calls it on RHCOS.

dustymabe: also this function is `get_fcos_ver` and the function below is `get_rhel_maj_ver`. Should we rename this one `get_fedora_maj_ver` or `get_fedora_ver`?

dustymabe (dismissed):
LGTM - one suggestion

tests/kola/data/commonlib.sh
jlebon: Renamed!

We can't error out because sadly the way errexit works in conditionals (which is how often this will be used) is tricky in bash. But we can at least not return anything.

tests/kola/data/commonlib.sh
```diff
@@ -25,3 +25,26 @@ get_ipv4_for_nic() {
...
+is_rhcos() (
+    source /etc/os-release
+    [ "${ID}" == "rhcos" ]
+)
+
+get_fedora_ver() (
+    source /etc/os-release
+    if is_fcos; then
+        echo "${VERSION_ID}"
+    fi
```
dustymabe, line 44: Does this actually cause an error, though? Do we want it to?

dustymabe: oh just saw https://github.com/coreos/fedora-coreos-config/pull/1541#discussion_r813968214

tests/kola/data/commonlib.sh
dustymabe: ahh. interesting.

dustymabe (approved):

PR #1542, opened 2022-02-25
15fcos: add coreos-fix-systemd-journald.service

dustymabe:
This service will detect when systemd-journald has been leaked from the
initramfs and kill it so the system can recover. This is a workaround
for https://github.com/coreos/fedora-coreos-tracker/issues/1108

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Fix systemd-journald after switch root on Fedora CoreOS
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1108
+DefaultDependencies=no
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-fix-systemd-journald
+RemainAfterExit=yes
+# can't send output to the journal because we might kill it
+StandardOutput=file:/run/coreos-fix-systemd-journald-stdout
+StandardError=file:/run/coreos-fix-systemd-journald-stderr
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -3,3 +3,5 @@ enable fedora-coreos-pinger.service
 enable coreos-check-ssh-keys.service
 # Check if cgroupsv1 is still being used
 enable coreos-check-cgroups.service
+# Fix systemd-journald if required
+enable coreos-fix-systemd-journald.service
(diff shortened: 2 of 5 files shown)


overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+# https://github.com/coreos/fedora-coreos-tracker/issues/1108
```
jlebon: Minor: this could be a `Documentation` value instead.

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
```diff
@@ -0,0 +1,13 @@
...
+Description=Fix systemd-journald after switch root on Fedora CoreOS
+DefaultDependencies=no
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-fix-systemd-journald
+RemainAfterExit=yes
+# can't send output to the journal because we might kill it
+StandardOutput=file:/run/coreos-fix-systemd-journald-stdout
+StandardError=file:/run/coreos-fix-systemd-journald-stderr
```
jlebon, line 11: Should these be under `/run/coreos/` too like the stamp file?

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
jlebon, line 1: Can we further constrain this to `ConditionFirstBoot=true`? This lets us immediately rule out upgrading nodes from consideration.

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
dustymabe: I wasn't sure if it would cause an error if the `/run/coreos` directory didn't exist.

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
dustymabe: Since we don't know the real root cause (yet) we don't know if this is limited to first boot, right?

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
dustymabe: k - i'll look at that

dustymabe: updated!

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
jlebon: Hmm true, though it seems highly likely to be related to the fact that the non-exclusive bucket has a huge Ignition config which changes timing in the initramfs. And AFAIU we haven't seen this anywhere else. But OK as is too.

jlebon (approved):

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-systemd-journald.service
dustymabe: I agree with you that it probably doesn't happen any other time. Thanks!

PR #1548, opened 2022-02-28
[testing] tree: promote changes from testing-devel at 53098661a83b3c6b26c334540aaec72d3f9cbfc9

coreosbot-releng:
Triggered by @gursewak1997 in https://github.com/coreos/fedora-coreos-streams/issues/458.


kola-denylist.yaml
```diff
@@ -19,42 +19,50 @@
...
+  streams:
+    - rawhide
+- pattern: ext.config.toolbox
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1103
+  snooze: 2022-03-15
+  streams:
+    - rawhide
+- pattern: multipath.day1
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1105
+  snooze: 2022-03-07
```
dustymabe: should we make another commit on top here that makes this not a snooze but a full denial? That way if it happens in two weeks when we spin stable it won't be an issue.

dustymabe (dismissed):
One suggestion otherwise LGTM

kola-denylist.yaml
jlebon: Makes sense to me. (Doesn't this apply to all the other 03-07 ones too?)

dustymabe: i think all of the other ones are for `branched`/`rawhide`

dustymabe (approved):
LGTM

gursewak1997 (approved):

PR #1552, opened 2022-03-02
tests/default-network-behavior-change: update for rhel8.6

HuijingHei:
RHEL8.6 includes NetworkManager-1.36.0-1.el8.x86_64, update
scripts according to F36.
See https://github.com/coreos/fedora-coreos-config/pull/1533

### DIFF
--- a/tests/kola/networking/default-network-behavior-change/test.sh
+++ b/tests/kola/networking/default-network-behavior-change/test.sh
@@ -13,7 +13,7 @@ set -xeuo pipefail
 . $KOLA_EXT_DATA/commonlib.sh
 
 # EXPECTED_INITRD_NETWORK_CFG1
-#   - used on Fedora 35 and RHEL 8.5+ releases
+#   - used on Fedora 35 and RHEL 8.5 release
 EXPECTED_INITRD_NETWORK_CFG1="[connection]
 id=Wired Connection
 uuid=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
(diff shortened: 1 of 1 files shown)


HuijingHei:
Test locally and result is PASSED using qcow2 image with RHCOS-4.11-86, RHCOS-4.11-85, RHCOS-4.10-84

<details>
[coreos-assembler]$ kola run -E network --qemu-image ./rhcos-411.86.202203020523-0-qemu.x86_64.qcow2 ext.network.networking.default-network-behavior-change
--- PASS: non-exclusive-test-bucket-0 (65.48s)
    --- PASS: non-exclusive-test-bucket-0/ext.network.networking.default-network-behavior-change (1.77s)
PASS, output in _kola_temp/qemu-unpriv-2022-03-02-0657-18333

[coreos-assembler]$ kola run -E network --qemu-image ./rhcos-411.85.202203020439-0-qemu.x86_64.qcow2 ext.network.networking.default-network-behavior-change
--- PASS: non-exclusive-test-bucket-0 (56.21s)
    --- PASS: non-exclusive-test-bucket-0/ext.network.networking.default-network-behavior-change (1.74s)
PASS, output in _kola_temp/qemu-unpriv-2022-03-02-0658-18357

[coreos-assembler]$ kola run -E network --qemu-image ./rhcos-410.84.202202282006-0-qemu.x86_64.qcow2 ext.network.networking.default-network-behavior-change
--- PASS: non-exclusive-test-bucket-0 (54.26s)
    --- PASS: non-exclusive-test-bucket-0/ext.network.networking.default-network-behavior-change (1.67s)
PASS, output in _kola_temp/qemu-unpriv-2022-03-02-0700-18416
</details>

cgwalters (approved):
LGTM!

dustymabe:
I understand the appeal of wanting to have the logic here be based on RHEL version, but I still think this is easier for us if we do it based on the OCP version so that we know exactly when it enters an OCP stream.

The reason is because an NM change may require us to change our [detection logic](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh#L25-L59) and we'd want to bundle them together.

That being said, I don't manage the RHCOS side as much so I'm OK to leave it as is.

PR #1550, opened 2022-03-01
multipathd.socket: remove workaround for rhel8.6

HuijingHei:
The fix landed in RHEL8.6+ and there is no need for a workaround.
Add check to remove workaround if `RHEL_VERSION` > `8.5`.
See:
- https://bugzilla.redhat.com/show_bug.cgi?id=2008101
- Fixed RPM version: `device-mapper-multipath-0.8.4-18.el8`

PR #1553, opened 2022-03-02
multipathd.socket: fix `remove workaround for rhel8.6`

HuijingHei:
https://github.com/coreos/fedora-coreos-config/pull/1550 does not work when https://github.com/openshift/os/pull/713 merged,
because it will use generic redhat-release package instead of redhat-release-coreos.

Need to check RHEL specific version using `VERSION_ID`(for example
`VERSION_ID=8.6`), and clarify rhel or fedora using `ID=rhel`,
as this will be executed before override according to
https://github.com/openshift/os/blob/master/manifest.yaml#L128-L150

### DIFF
--- a/manifests/shared-workarounds.yaml
+++ b/manifests/shared-workarounds.yaml
@@ -14,12 +14,15 @@ postprocess:
         # FCOS: Only operate on releases before F36. The fix has landed
         # in F36+ and there is no need for a workaround.
         [ ${VERSION_ID} -le 35 ] || exit 0
-    elif [[ "${ID}" == "rhcos" ]]; then
+    elif [[ "${ID}" == "rhel" ]]; then
         # RHCOS: Only operate on releases before RHEL8.5. The fix has 
         # landed in RHEL8.6+ and there is no need for a workaround.
         # https://bugzilla.redhat.com/show_bug.cgi?id=2008101
         # Fixed RPM version: device-mapper-multipath-0.8.4-18.el8
-        [ ${RHEL_VERSION//.} -le 85 ] || exit 0
+        # Using generic redhat-release package according to 
+        # https://github.com/openshift/os/pull/713, should check RHEL version
+        # with VERSION_ID
+        [ ${VERSION_ID//.} -le 85 ] || exit 0
     fi
     mkdir /usr/lib/systemd/system/multipathd.socket.d
     cat > /usr/lib/systemd/system/multipathd.socket.d/50-start-conditions.conf <<'EOF'


manifests/shared-workarounds.yaml
```diff
@@ -14,12 +14,15 @@ postprocess:
         # FCOS: Only operate on releases before F36. The fix has landed
         # in F36+ and there is no need for a workaround.
         [ ${VERSION_ID} -le 35 ] || exit 0
-    elif [[ "${ID}" == "rhcos" ]]; then
+    elif [[ "${ID}" == "rhel" ]]; then
```
lucab, line 17: I'm puzzled by this.

Does this check work with the current 8.4 content?

It looks like we are stepping into an OS detection mess; can't this logic directly check the content of `multipathd.socket`?

lucab:
This is quickly getting outside of my comfort zone, I will leave the review of this to somebody more knowledgeable about FCOS and RHCOS configs.

/cc @dustymabe @miabbott

manifests/shared-workarounds.yaml
cgwalters: Indeed, something like `if ! grep -qF 'ConditionKernelCommandLine=!multipath=off' /usr/lib/systemd/system/multipath.socket`?

dustymabe:
will look in a bit

miabbott:
Using Colin's suggestion:

```diff
diff --git a/manifests/shared-workarounds.yaml b/manifests/shared-workarounds.yaml
index 5b205e8..7fe7a2b 100644
--- a/manifests/shared-workarounds.yaml
+++ b/manifests/shared-workarounds.yaml
@@ -10,19 +10,10 @@ postprocess:
     set -xeuo pipefail
     # Operate on RHCOS and FCOS. 
     source /etc/os-release
-    if [[ ${NAME} =~ "Fedora" ]]; then
+    if grep -qF 'ConditionKernelCommandLine=!multipath=off' /usr/lib/systemd/system/multipathd.socket ; then
         # FCOS: Only operate on releases before F36. The fix has landed
         # in F36+ and there is no need for a workaround.
-        [ ${VERSION_ID} -le 35 ] || exit 0
-    elif [[ "${ID}" == "rhel" ]]; then
-        # RHCOS: Only operate on releases before RHEL8.5. The fix has 
-        # landed in RHEL8.6+ and there is no need for a workaround.
-        # https://bugzilla.redhat.com/show_bug.cgi?id=2008101
-        # Fixed RPM version: device-mapper-multipath-0.8.4-18.el8
-        # Using generic redhat-release package according to 
-        # https://github.com/openshift/os/pull/713, should check RHEL version
-        # with VERSION_ID
-        [ ${VERSION_ID//.} -le 85 ] || exit 0
+        exit 0
     fi
     mkdir /usr/lib/systemd/system/multipathd.socket.d
     cat > /usr/lib/systemd/system/multipathd.socket.d/50-start-conditions.conf <<'EOF'
```

This allowed me to build + test RHCOS with RHEL 8.5; also build + test FCOS using F35

dustymabe:
My recommendation would be to just add another `elif` to check for `rhel` and handle it that way. If we change this logic to check the `multipathd.socket` directly (which makes sense to do) then [our test](https://github.com/coreos/fedora-coreos-config/blob/5a2c87639ede4f9452e3e123935e3ce954dcb017/tests/kola/multipath/multipathd-socket-fix) won't fail in the future and let us know when we can remove this workaround, which I also think is important.


```diff
diff --git a/manifests/shared-workarounds.yaml b/manifests/shared-workarounds.yaml
index efe8f04..298f9bb 100644
--- a/manifests/shared-workarounds.yaml
+++ b/manifests/shared-workarounds.yaml
@@ -14,11 +14,15 @@ postprocess:
         # FCOS: Only operate on releases before F36. The fix has landed
         # in F36+ and there is no need for a workaround.
         [ ${VERSION_ID} -le 35 ] || exit 0
-    elif [[ "${ID}" == "rhcos" ]]; then
+    elif [[ "${ID}" == "rhel" ]]; then
         # RHCOS: Only operate on releases before RHEL8.5. The fix has 
         # landed in RHEL8.6+ and there is no need for a workaround.
         # https://bugzilla.redhat.com/show_bug.cgi?id=2008101
         # Fixed RPM version: device-mapper-multipath-0.8.4-18.el8
+        [ ${VERSION_ID//.} -le 85 ] || exit 0
+    elif [[ "${ID}" == "rhcos" ]]; then
+        # RHCOS: On versions before https://github.com/openshift/os/pull/713
+        # $ID was "rhcos" and we needed to use $RHEL_VERSION
         [ ${RHEL_VERSION//.} -le 85 ] || exit 0
     fi
     mkdir /usr/lib/systemd/system/multipathd.socket.d
```

manifests/shared-workarounds.yaml
travier: ~~This is really weird, the ID should still be `rhcos`.~~

jlebon: > This is really weird, the ID should still be `rhcos`.

See https://github.com/openshift/os/pull/713#issuecomment-1057030751.

miabbott:
> then [our test](https://github.com/coreos/fedora-coreos-config/blob/5a2c87639ede4f9452e3e123935e3ce954dcb017/tests/kola/multipath/multipathd-socket-fix) won't fail in the future and let us know when we can remove this workaround, which I also think is important.

Good observation; I agree with that suggestion

manifests/shared-workarounds.yaml
travier: Indeed, I thought this was in a test.

jlebon (approved):
> then [our test](https://github.com/coreos/fedora-coreos-config/blob/5a2c87639ede4f9452e3e123935e3ce954dcb017/tests/kola/multipath/multipathd-socket-fix) won't fail in the future and let us know when we can remove this workaround, which I also think is important.

Hmm, but note the postprocess script here is explicitly exiting early on versions of Fedora and RHEL which do have the fix. So IIUC the `multipathd-socket-fix` test won't start failing once we move to 8.6 anyway, right?

Did you maybe mean to say is that making it version-based is better because then it's easier to tell when we can remove the workaround? (I.e. once we're past all the versions we're gating on.) I agree with that.

So overall, I think what @HuijingHei has here is fine. (I also don't see any need for this postprocess script to be compatible with multiple versions of RHCOS; we already branch this repo for that.)

dustymabe:
> > then [our test](https://github.com/coreos/fedora-coreos-config/blob/5a2c87639ede4f9452e3e123935e3ce954dcb017/tests/kola/multipath/multipathd-socket-fix) won't fail in the future and let us know when we can remove this workaround, which I also think is important.
> 
> Hmm, but note the postprocess script here is explicitly exiting early on versions of Fedora and RHEL which do have the fix. So IIUC the `multipathd-socket-fix` test won't start failing once we move to 8.6 anyway, right?

You are right. I was slightly confused. All the test is doing is making sure we have the workaround OR the fix, but not both. Of course, there is nothing that prevents the fix from being backported to 8.5, in which case the test would flag it.

> 
> Did you maybe mean to say is that making it version-based is better because then it's easier to tell when we can remove the workaround? (I.e. once we're past all the versions we're gating on.) I agree with that.

Yeah. That's also a benefit.

> 
> So overall, I think what @HuijingHei has here is fine. (I also don't see any need for this postprocess script to be compatible with multiple versions of RHCOS; we already branch this repo for that.)

OK. Yeah I wasn't sure if we needed to handle old versions or not.

dustymabe (approved):

HuijingHei:
Thanks all for the review and comment.

PR #1554, opened 2022-03-02
Skip kola ext.config.shared.kdump.crash test for ppc64le

ravanelli:
- This test is failing with a Kernel Panic, blocking the CI
- More info: https://github.com/coreos/coreos-assembler/issues/2725

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -17,6 +17,7 @@
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1075
   arches:
     - aarch64
+    - ppc64le
 - pattern: ext.config.podman.dns
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1046
   snooze: 2022-03-07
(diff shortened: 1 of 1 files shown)


tests/kola/kdump/crash
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# kola: {"minMemory": 4096, "tags": "skip-base-checks", "architectures": "!s390x"}
+# kola: {"minMemory": 4096, "tags": "skip-base-checks", "architectures": "!s390x ppc64le"}
```
dustymabe: does this actually do what you want? I feel like you need the `!`.

At the same time, should this ever work on ppc64le? If so I think we should probably use the [denylist](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/kola-denylist.yaml) instead.

ravanelli: @dustymabe No sure if  it should work, I think we should probably investigate it more to understand this case better, I think it should be related with `crashkernel=256M'` passed to the test. Do we really need 256? I will try to run a test with less mem to see if it works

kola-denylist.yaml
```diff
@@ -66,3 +66,15 @@
   streams:
     - rawhide
     - branched
+- pattern: ext.config.shared.kdump.crash
```
dustymabe: there's already an entry for this test above - just need to add the ppc64le arch there

ravanelli: Thanks for checking it Dusty!

dustymabe (approved):
LGTM

gursewak1997 (approved):

PR #1558, opened 2022-03-02
manifests: make `releasever` a number and add iptables-nft.yaml for f36+

jlebon:
```
commit 84dea8c5624cae0650dac6f5cf0e807d5bad6131
Date:   Tue Feb 22 21:30:07 2022 -0500

    manifests: make `releasever` a number

    This is supported by rpm-ostree now.
```
---
```
commit 617558c533f306d956dd1b93d61f109810f240c2
Date:   Tue Feb 22 21:31:37 2022 -0500

    manifests: add iptables-nft.yaml for f36+

    Add a new conditional include which migrates streams on Fedora 36 and
    higher to iptables-nft. This is magnitudes simpler than the current
    alternative rollout procedure detailed in:

    https://github.com/coreos/fedora-coreos-tracker/issues/676
```
---
```
kola-denylist: drop ext.config.firewall.iptables

This should be fixed now.
```

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -60,9 +60,3 @@
   streams:
     - rawhide
     - branched
-- pattern: ext.config.firewall.iptables
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/676
-  snooze: 2022-03-07
-  streams:
-    - rawhide
-    - branched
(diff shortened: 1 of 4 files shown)


jlebon:
We'll need to propagate the `releasever` change to all the streams, which... in that case maybe it's better to do all the other things in https://github.com/coreos/fedora-coreos-config/pull/1538 at the same time too.

jlebon:
OK, let's close this and work on getting https://github.com/coreos/fedora-coreos-config/pull/1538 in instead. Will see if there are other cleanups that affect `manifest.yaml` that should be added there before propagating to all the other branches.

PR #1559, opened 2022-03-03
coreos-teardown-initramfs: account for BOOTIF kernel argument

### LINKED ISSUE #1048: networking: consider the effects of `BOOTIF` kernel argument on `nm-initrd-generator`
When booting via PXE syslinux will inject a `BOOTIF` kernel argument to let any interested parties know what interface was used to PXE boot the machine. We were previously unaware that `nm-initrd-generator` would pick up this argument and generate a network configuration file with a MAC address in it. This behavior throws off our [code](https://github.com/coreos/fedora-coreos-config/blob/752562a30e8f8b25ec4a74618ef98d0dc00dfe97/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh#L25-L60) that attempts to determine if the user supplied any networking configuration or not.

This means a user may end up with an unexpected Network configuration with a MAC address in the configuration file. Currently they can explicitly tell the propagate network code to not propagate networking configuration (`coreos.no_persist_ip`). There also exists a `rd.bootif=0` flag we can use to tell `nm-initrd-generator` to ignore `BOOTIF`. Maybe we should consider making `rd.bootif=0` the default behavior on FCOS since that's what we were expecting anyway. Though changing the defaults should be considered carefully.

dustymabe:
`nm-initrd-generator` will take into account the `BOOTIF=` kernel
argument. Let's account for that in our `propagate_initramfs_networking()`
code so don't propagate networking config if `BOOTIF` is the only
networking config.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1048

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -22,6 +22,21 @@ dracut_func() {
     return $rc
 }
 
+# Get the BOOTIF and rd.bootif kernel arguments from
+# the kernel command line.
+get_bootif_kargs() {
+    bootif_kargs=""
+    bootif_karg=$(dracut_func getarg BOOTIF)
+    if [ ! -z "$bootif_karg" ]; then
+        bootif_kargs+="BOOTIF=${bootif_karg}"
+    fi
+    rdbootif_karg=$(dracut_func getarg rd.bootif)
+    if [ ! -z "$rdbootif_karg" ]; then
+        bootif_kargs+="rd.bootif=${rdbootif_karg}"
+    fi
+    echo $bootif_kargs
+}
+
 # Determine if the generated NM connection profiles match the default
 # that would be given to us if the user had provided no additional
 # configuration. i.e. did the user give us any network configuration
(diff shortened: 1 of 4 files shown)


miabbott (approved):
This makes sense to me.

In addition to this, I think we need to get OCP docs looped in so we can document this new behavior.

dustymabe:
> In addition to this, I think we need to get OCP docs looped in so we can document this new behavior.

cc @bobfuru @jldohmann - let me know the best way to provide the necessary information.

miabbott:
> > In addition to this, I think we need to get OCP docs looped in so we can document this new behavior.
> 
> cc @bobfuru @jldohmann - let me know the best way to provide the necessary information.

For ease of discovery, there's an associated RHBZ for this change - https://bugzilla.redhat.com/show_bug.cgi?id=2032717

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -22,6 +22,21 @@ dracut_func() {
...
+# the kernel command line.
+get_bootif_kargs() {
+    bootif_kargs=""
+    bootif_karg=$(dracut_func getarg BOOTIF)
+    if [ ! -z "$bootif_karg" ]; then
+        bootif_kargs+="BOOTIF=${bootif_karg}"
+    fi
+    rdbootif_karg=$(dracut_func getarg rd.bootif)
+    if [ ! -z "$rdbootif_karg" ]; then
+        bootif_kargs+="rd.bootif=${rdbootif_karg}"
```
jlebon, line 35: Isn't this missing a leading space?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -22,6 +22,21 @@ dracut_func() {
     return $rc
 }
 
+# Get the BOOTIF and rd.bootif kernel arguments from
+# the kernel command line.
+get_bootif_kargs() {
+    bootif_kargs=""
```
jlebon, line 28: Minor: variables in this function should be `local`. And... oh yeah there are a lot of instances of that in this script.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
dustymabe: yep: https://github.com/coreos/fedora-coreos-config/pull/1561

PR #1560, opened 2022-03-03
[rawhide] overrides: fast-track rpm-ostree-2022.4-1.fc37

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -36,3 +36,13 @@ packages:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-b212ac738a
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1046#issuecomment-1054585092
       type: fast-track
+  rpm-ostree:
+    evr: 2022.4-1.fc37
+    metadata:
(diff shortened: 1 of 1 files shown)


dustymabe (dismissed):

dustymabe:
Will have someone else review this one too since I added a second commit on top.

dustymabe:
hmm - CI is failing with:
```
[2022-03-03T02:47:24.981Z] error: Executing %transfiletriggerin for glibc-common: Package 'glibc-common' has (currently) unsupported <lua> script in '%transfiletriggerin'
```

But I thought that should be fixed with the new `rpm-ostree`.

cgwalters:
> But I thought that should be fixed with the new rpm-ostree.

I think we want to fast-track it into coreos-assembler to fix that actually.

dustymabe:
> I think we want to fast-track it into coreos-assembler to fix that actually.

yeah we did that already (yesterday). New info: https://github.com/coreos/rpm-ostree/issues/3487

dustymabe:
Considering https://github.com/coreos/rpm-ostree/issues/3487 I dropped the second commit from this PR.

dustymabe (approved):

PR #1565, opened 2022-03-03
[next-devel] manifest.yaml: update for conditional-include cleanups

jlebon:
Apply the same cleanups as in testing-devel's manifest in
https://github.com/coreos/fedora-coreos-config/pull/1538.


dustymabe (approved):
LGTM

PR #1574, opened 2022-03-04
[rawhide] overrides: drop glibc pin

jlebon:
We have a new rpm-ostree now with the lua scriptlet workaround needed
for https://github.com/coreos/fedora-coreos-tracker/issues/1113.


dustymabe (approved):
LGTM

PR #1577, opened 2022-03-07
manifests: apply podman-v4.yaml manifest on F36+

dustymabe:
This will allow us to share the package inclusion rather than carrying
it in manifest.yaml on various branches.


manifests/fedora-coreos-base.yaml
```diff
@@ -22,6 +22,11 @@ conditional-include:
   - if: releasever >= 36
     include: iptables-nft.yaml
 
+conditional-include:
```
jlebon:
```suggestion
```

jlebon (approved):

dustymabe:
CI hit the journald fix issue. Re-trying.

PR #1581, opened 2022-03-07
workflows/remove-graduated-overrides: enable continuous repo

jlebon:
Until we actually use cosa for this, we should match its input repos.
Currently that means enabling the continuous repo.

This should give us a more recent rpm-ostree which has support for
`releasever` as a number.


dustymabe (approved):
LGTM

PR #1585, opened 2022-03-08
[next] overrides: fast-track kernel-5.16.13-200.fc35

dustymabe:
This is the first kernel with the most recent revert that allows
us to not regress on some AWS instance types [1]. Because it is
newer than 5.16.11 it also allows for us to pick up the fix to
CVE-2022-0847 [2].

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1066
[2] https://github.com/coreos/fedora-coreos-tracker/issues/1118


jlebon (dismissed):

dustymabe:
CI failed on the systemd-journal-fix which is a test that let's us know when we needed the workaround for https://github.com/coreos/fedora-coreos-tracker/issues/1108

will restart CI

jlebon (approved):

PR #1586, opened 2022-03-08
denylist: extend snooze for multipath.day1

dustymabe:
I thought this one might have gone away but it seems to only present
itself on aarch64.


jlebon (approved):

dustymabe:
killed CI since it doesn't test aarch64

PR #1587, opened 2022-03-08
tests: extend timeout for podman.dns test

dustymabe:
I've seen this timeout recently. It could take some extra time to
    grab the container from the registry so let's bump the timeout from
    the default of 1 minute for a non-exclusive test to 3 minutes.


cgwalters (approved):

dustymabe:
Thanks @cgwalters

PR #1589, opened 2022-03-08
manifests: Move `container-cmd` to `user-experience.yaml`

cgwalters:
I was going to do a PR to add this to openshift/os too, but actually
let's just share it more directly instead!

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -11,8 +11,6 @@ rojig:
 add-commit-metadata:
   fedora-coreos.stream: ${stream}
 
-container-cmd:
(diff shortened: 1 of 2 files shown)


jlebon (approved):
Makes sense to me!

dustymabe:
if possible let's hold this until the current bump-lockfile run finishes: https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/bump-lockfile/66/

dustymabe:
done - this is now unblocked.

cgwalters:
/override continuous-integration/jenkins/pr-merge
Flaked on journal bits

PR #1590, opened 2022-03-09
overrides: fast-track kernel-5.16.13-200.fc35

dustymabe:
This is the first kernel with the most recent revert that allows
us to not regress on some AWS instance types [1]. Because it is
newer than 5.16.11 it also allows for us to pick up the fix to
CVE-2022-0847 [2].

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1066
[2] https://github.com/coreos/fedora-coreos-tracker/issues/1118


dustymabe:
in CI the run_upgrades test failed because the [previous build](https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/blue/organizations/jenkins/fedora-coreos-config/detail/PR-1590/1/pipeline) of `testing-devel` failed part of the way through uploading to GCP.

dustymabe:
I'll merge this if the CI for the `next-devel` PR passes since the two branches are in lockstep right now.

jlebon (approved):

PR #1596, opened 2022-03-14
[next] tree: promote changes from next-devel at c4c58b08e965f471e5a04aa931e064e80399e688

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/463.

### DIFF
--- a/.github/workflows/remove-graduated-overrides.yml
+++ b/.github/workflows/remove-graduated-overrides.yml
@@ -24,6 +24,10 @@ jobs:
           - rawhide
       fail-fast: false
     steps:
+      - name: Enable CoreOS continuous repo
+        run: |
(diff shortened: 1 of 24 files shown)


dustymabe:
Let's hold this merge. For [$reasons](https://github.com/coreos/coreos-assembler/pull/2751) our `next-devel` promotion wasn't happening like it was for `testing-devel` so the content in this stream doesn't match `testing-devel` (which it should).

dustymabe:
Now that https://github.com/coreos/fedora-coreos-config/pull/1598 is merged we should be able to re-promote and we'll have the same content IIUC

jlebon (approved):

PR #1598, opened 2022-03-14
lockfiles: sync lockfiles with `testing-devel`

dustymabe:
The lockfile bumper hasn't been working because the previous
`next-devel` doesn't have a QEMU image, which causes the upgrade
test to fail. Let's sync the lockfiles from `testing-devel` to
catch us up.


jlebon (approved):

dustymabe:
skipping tests since the contentset already passed tests when promoted for `testing-devel`.

PR #1599, opened 2022-03-14
[testing] denylist: disable multipath.day1

dustymabe:
This test is failing occasionally on aarch64 and we need to
convert it from a snooze to a full denial so that it will work
in two weeks when we build `stable`.

See https://github.com/coreos/fedora-coreos-tracker/issues/1105

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -40,7 +40,8 @@
     - rawhide
 - pattern: multipath.day1
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1105
-  snooze: 2022-03-14
+  streams:
+    - stable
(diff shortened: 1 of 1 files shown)


dustymabe:
killed CI - it won't tell us anything

kola-denylist.yaml
```diff
@@ -40,7 +40,8 @@
     - rawhide
 - pattern: multipath.day1
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1105
-  snooze: 2022-03-14
+  streams:
+    - stable
```
jlebon, line 44: Heh, clever.

jlebon (approved):

PR #1600, opened 2022-03-14
[next] denylist: disable multipath.day1

dustymabe:
This test failed in our `next` aarch64 run because the snooze
had expired. Let's convert it from a snooze to a full denial
so that we can kick off a new run and not leave the success
of the run up to chance.

See https://github.com/coreos/fedora-coreos-tracker/issues/1105


dustymabe:
killed CI

jlebon (approved):

PR #1607, opened 2022-03-17
networking: veirfy rd.net.timeout.dhcp and rd.net.dhcp.retry are supported by NetworkManager.

HuijingHei:
Append rd.net.timeout.dhcp and rd.net.dhcp.retry to kernel parameter
when boot, get total dhcp timeout is `rd.net.timeout.dhcp * rd.net.dhcp.retry`
seconds
See:
- https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/559
- https://bugzilla.redhat.com/show_bug.cgi?id=1879094#c10
- https://bugzilla.redhat.com/show_bug.cgi?id=1877740

### DIFF
--- a/tests/kola/networking/kargs-rd-net
+++ b/tests/kola/networking/kargs-rd-net
@@ -1,0 +1,31 @@
+#!/bin/bash
+# kola: { "platforms": "qemu", "appendFirstbootKernelArgs": "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8"}
+# 
+# Veirfy rd.net.timeout.dhcp and rd.net.dhcp.retry are supported
+# by NetworkManager. Append them to kernel parameter when boot, 
+# get total timeout is `timeout * retry`, 30*8(240) seconds 
+# in this test scenario
+# See:
+# - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/559
+# - https://bugzilla.redhat.com/show_bug.cgi?id=1879094#c10
+# - https://bugzilla.redhat.com/show_bug.cgi?id=1877740
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# check kernel parameter
+cmdline="/proc/cmdline"
+for rd in rd.net.timeout.dhcp rd.net.dhcp.retry;
+do
+    if ! grep -q $rd $cmdline; then
+        fatal "Error: can not find $rd in kernel parameter"
+    fi
+done
+
+log=$(journalctl -b -u NetworkManager | grep "beginning transaction")
+timeout=$(echo $log | awk -F'[( )]' '{print $(NF-2)}')
+if [ $timeout -ne 240 ]; then
+    fatal "Error: actual dhcp timeout is $timeout, expected 240"
+fi
+ok "Total dhcp timeout matches expected value"


ravanelli (dismissed):
/LGTM

tests/kola/networking/kargs-rd-net
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# kola: { "platforms": "qemu", "appendKernelArgs": "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8"}
```
dustymabe:
```suggestion
# kola: { "platforms": "qemu", "appendFirstbootKernelArgs": "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8"}
```

Let's make this apply to the first boot only to more closely mimic the way it would actually be used by users.

tests/kola/networking/kargs-rd-net
```diff
@@ -0,0 +1,31 @@
...
+# check kernel parameter
+cmdline="/proc/cmdline"
+for rd in rd.net.timeout.dhcp rd.net.dhcp.retry;
+do
+    if ! grep -q $rd $cmdline; then
+        fatal "Error: can not find $rd in kernel parameter"
+    fi
+done
+
+log=$(journalctl -b -u NetworkManager | grep timeout)
```
dustymabe: Can we make the grep more specific here.. `timeout` is pretty generic and I wouldn't be surprised if another instance of it popped up in the NetworkManager logs in the future.

tests/kola/networking/kargs-rd-net
```diff
@@ -0,0 +1,31 @@
...
+timeout=$(echo $log | awk -F'[( )]' '{print $(NF-2)}')
+if [ $timeout -ne 240 ]; then
+    fatal "Error: actual dhcp timeout is $timeout, expected 240"
+fi
+ok "Total dhcp timeout is expected"
```
dustymabe:
```suggestion
ok "Total dhcp timeout matches expected value"
```

Optional wording update for clarity.

dustymabe (changes requested):
A few suggestions. Thanks for working on this!

tests/kola/networking/kargs-rd-net
HuijingHei: Thanks @dustymabe for your suggestion. I am just curious about the difference between the 2 parameters, could you help to clarify?

tests/kola/networking/kargs-rd-net
HuijingHei: The message is like `localhost.localdomain NetworkManager[1519]: <info>  [1647568963.4738] dhcp4 (ens4): activation: beginning transaction (timeout in 240 seconds)`, update the grep condition to `beginning transaction`, any concerns about this?

tests/kola/networking/kargs-rd-net
dustymabe: One of them applies kargs only on first boot (which is typically how our users supply networking kargs) and the other applies kargs permanently (not something our users typically do).

tests/kola/networking/kargs-rd-net
dustymabe: seems sufficient. thanks

dustymabe (approved):
LGTM

tests/kola/networking/kargs-rd-net
HuijingHei: Understand, thanks!

PR #1610, opened 2022-03-17
manifests/system-configuration: add `acl`

jlebon:
This is currently getting pulled in by systemd, but won't anymore in
f36. Let's keep shipping it though since users might be relying on it to
manage permissions.


bgilbert (approved):

dustymabe (approved):
LGTM

PR #1611, opened 2022-03-17
overrides: fast-track Ignition 2.13.0-5.fc35

sohankunkerkar:
For https://github.com/coreos/ignition/issues/1092.


bgilbert (approved):
This is okay as a manually-created override, but note that we have a [GitHub Actions workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) to do these now.

miabbott (approved):

PR #1613, opened 2022-03-17
manifests: fedora-coreos-base: add containernetworking-plugins explicitly

dustymabe:
In order to continue to support CNI networks in podman, which is what
any existing Fedora <= 35 system is using, we need to continue to ship
containernetworking-plugins.

It was dropped in [1]. Traditional yum based systems won't uninstall the
package on an upgrade, but OSTrees are "baked fresh" every time and
won't retain it because it's no longer a dependency. We need to name it
explicitly if we want upgrading FCOS users to be able to start their
containers.

[1] https://src.fedoraproject.org/rpms/podman/c/06c601234feac4f5da7b6e1d92430d62f556f3b9

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -118,9 +118,11 @@ packages:
   # Containers
   - systemd-container catatonit
   - fuse-overlayfs slirp4netns
-  # name resolution for podman containers
+  # support for old style CNI networks and name resolution for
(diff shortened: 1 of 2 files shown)


jlebon (dismissed):

dustymabe:
New update with `netavark` now named explicitly.

jlebon (approved):

PR #1620, opened 2022-03-22
overlay/05core: ship by-id/coreos-boot-disk udev rule in initrd only

### LINKED ISSUE #755: RHCOS 4.11 x86_64 error about missing script for creating stable symlink to boot disk
Was going through the journal for a separate issue and saw the following:

`Mar 22 19:06:09 localhost [1488]: failed to execute '/usr/lib/udev/coreos-disk-contains-fs' 'coreos-disk-contains-fs vda boot': No such file or directory`

This happens after the switch to the real root and by my read of https://github.com/coreos/fedora-coreos-config/pull/1308 the udev rule is only available in the initrd, with the symlink being removed during initramfs teardown.

So the error looks scary, but is harmless AFAICT

jlebon:
The associated script only exists in the initrd. But also, we only need
the symlink in the initrd anyway.

Fixes 6d7b9ad5 ("overlay.d: add udev rule for creating stable symlink to
boot disk").

Fixes https://github.com/openshift/os/issues/755.


dustymabe (approved):

dustymabe:
LGTM

miabbott (approved):
LGTM

PR #1621, opened 2022-03-23
05core/*: display Ignition warnings on the console

### LINKED ISSUE #1125: Use MOTD to display Ignition warnings on the console
**Describe the enhancement**
The motivation behind this enhancement: https://github.com/coreos/ignition/issues/1301
We're thinking of displaying a warning for the above issue on the serial console via a motd, however, after some discussion with @jlebon and @bgilbert, it seems like extending this feature to all Ignition warnings makes more sense here and would be a good UX improvement for FCOS users.

**System details**
 - Bare Metal/QEMU/AWS/GCP/etc.
 - Fedora CoreOS 35.20220227.3.0

sohankunkerkar:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1125

This change displays all Ignition warnings on the console. Also, renamed the existing Ignition service to make it more generalized for handling status information about the Ignition run


sohankunkerkar:
```
Fedora CoreOS 35.20220323.dev.1
Kernel 5.16.15-201.fc35.x86_64 on an x86_64 (ttyS0)

SSH host key: SHA256:G++MZpyPpblUMxpFpRLIrnMVKOQSkd0gw7cYz0zgnIE (ECDSA)
SSH host key: SHA256:sRpzjXdiaZ23ls72glVqpFYsFLcOBl+1xub7bZptFSw (ED25519)
SSH host key: SHA256:j0JPMupDzva4j76t9MrLDlQtlB64RentiC8TW0Jio9k (RSA)
ens6: 10.0.2.15 fe80::1bdc:6145:315a:4f00
Ignition: ran on 2022/03/23 15:24:07 UTC (this boot)
Ignition: warning at $.storage.directories.0.mode, line 1 col 233: setuid/setgid/sticky bits are not supported in spec versions older than 3.4.0
Ignition: warning at $.storage.files.0.mode, line 1 col 368: setuid/setgid/sticky bits are not supported in spec versions older than 3.4.0
No SSH authorized keys provided by Ignition or Afterburn
cosa-devsh login: core (automatic login)

```

overlay.d/05core/usr/libexec/coreos-ignition-display-debug-info
```diff
@@ -42,6 +42,15 @@ else
...
+if [ -n "${IGNITION_WARNINGS}" ]; then
+    touch /run/issue.d/30_coreos_ignition_warnings.issue
+    output=$(while read line; do echo "Ignition: $line"; done <<< ${IGNITION_WARNINGS})
+    echo "$output" > /run/issue.d/30_coreos_ignition_warnings.issue
+fi
+
 # Our makeshift way of getting /run/issue.d semantics. See:
 # https://github.com/coreos/console-login-helper-messages/blob/e06fc88ae8fbcc3a422bc8c686f70c15aebb9d9a/usr/lib/console-login-helper-messages/issue.defs#L8-L17
-ln -sf /run/issue.d/30_coreos_ignition_provisioning.issue /etc/issue.d/
+ln -sf /run/issue.d/* /etc/issue.d/
```
bgilbert: Hmm, why are we wildcarding here?  We don't own that directory, so there might be other files in it that we didn't write.

overlay.d/05core/usr/libexec/coreos-ignition-display-debug-info
```diff
@@ -42,6 +42,15 @@ else
         >> /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+# TODO: Find a way to query journald logs before the system switches to real root
+# This is for displaying Ignition warnings on the serial console
```
bgilbert: nit: swap these two comment lines, so we say what we're doing before noting its flaws.

overlay.d/05core/usr/libexec/coreos-ignition-display-debug-info
```diff
@@ -42,6 +42,15 @@ else
...
+IGNITION_WARNINGS=$(journalctl -t ignition -o json-pretty -b 0 -p 4 | jq -r '.MESSAGE')
```
bgilbert: For clarity, let's spell it `warning` rather than `4`.

overlay.d/05core/usr/libexec/coreos-ignition-display-debug-info
```diff
@@ -42,6 +42,15 @@ else
...
+if [ -n "${IGNITION_WARNINGS}" ]; then
+    touch /run/issue.d/30_coreos_ignition_warnings.issue
```
bgilbert: Should be redundant?

overlay.d/05core/usr/libexec/coreos-ignition-display-debug-info
```diff
@@ -42,6 +42,15 @@ else
...
+    output=$(while read line; do echo "Ignition: $line"; done <<< ${IGNITION_WARNINGS})
```
bgilbert: This is okay, but it might be a bit simpler to just write a loop that echoes each line `>>` the issue file.

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-display-debug-info.service
```diff
@@ -0,0 +1,14 @@
+# This service displays Ignition warnings and provide
+# information if no Ignition config is provided.
+[Unit]
+Description=Display debug information for Ignition
```
bgilbert: The use of the word "display" throughout the PR is a bit confusing, I think, since we're not actually displaying anything - we're adding it to the MOTD to be displayed later.

(And it's not really debug information; it's status information about the Ignition run.)

bgilbert (commented):
The commit message mentions the serial console but this is misleading; the MOTD is shown on the graphical console too.

overlay.d/05core/usr/libexec/coreos-ignition-display-debug-info
sohankunkerkar: Yeah, makes sense. Fixed!

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-status-info.service
```diff
@@ -0,0 +1,15 @@
+# This service provides status information about the 
+# Ignition run, which includes Ignition warnings and 
+# information if no Ignition config is provided.
+[Unit]
+Description=Provide status information about the Ignition run
```
jlebon: The convention for systemd unit descriptions is Title Case. Maybe something like:

```suggestion
Description=Create Ignition Status Issue Files
```

?

overlay.d/05core/usr/libexec/coreos-ignition-status-info
```diff
@@ -42,6 +42,15 @@ else
...
+IGNITION_WARNINGS=$(journalctl -t ignition -o json-pretty -b 0 -p warning | jq -r '.MESSAGE')
+if [ -n "${IGNITION_WARNINGS}" ]; then
+    echo "$(while read line; do echo "Ignition: $line"; done <<< ${IGNITION_WARNINGS})" \
+        >> /run/issue.d/30_coreos_ignition_warnings.issue
+fi
+
 # Our makeshift way of getting /run/issue.d semantics. See:
 # https://github.com/coreos/console-login-helper-messages/blob/e06fc88ae8fbcc3a422bc8c686f70c15aebb9d9a/usr/lib/console-login-helper-messages/issue.defs#L8-L17
 ln -sf /run/issue.d/30_coreos_ignition_provisioning.issue /etc/issue.d/
+ln -sf /run/issue.d/30_coreos_ignition_warnings.issue /etc/issue.d/
```
jlebon: Note issues are displayed before login. @bgilbert Are there any concerns here about Ignition warnings leaking sensitive information?

bgilbert: At present there don't appear to be any warnings which clearly contain sensitive info, and most warnings don't have interpolated strings.  But there is a case where we can leak the name of a systemd unit, and we could always add more leaks in future.

I'm inclined to proceed anyway, and fix any such issues as they arise.  Thoughts?

overlay.d/05core/usr/libexec/coreos-ignition-status-info
```diff
@@ -42,6 +42,15 @@ else
         >> /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+# This is for displaying Ignition warnings on the console
+# TODO: Find a way to query journald logs before the system switches to real root
+IGNITION_WARNINGS=$(journalctl -t ignition -o json-pretty -b 0 -p warning | jq -r '.MESSAGE')
+if [ -n "${IGNITION_WARNINGS}" ]; then
+    echo "$(while read line; do echo "Ignition: $line"; done <<< ${IGNITION_WARNINGS})" \
+        >> /run/issue.d/30_coreos_ignition_warnings.issue
+fi
```
bgilbert: I was thinking something like:

```suggestion
    journalctl -t ignition -o json-pretty -b 0 -p warning | jq -r '.MESSAGE' | while read line; do
        echo "Ignition: $line" >> /run/issue.d/30_coreos_ignition_warnings.issue
    done
```

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-status-info.service
```diff
@@ -0,0 +1,15 @@
+# This service provides status information about the
```
bgilbert: "provides" is kinda vague.  Let's mention something about the issue files.

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-status-info.service
```diff
@@ -0,0 +1,15 @@
...
+# Ignition run, which includes Ignition warnings and 
+# information if no Ignition config is provided.
+[Unit]
+Description=Create Ignition Status Issue Files
+Before=systemd-user-sessions.service
+ConditionPathExists=/etc/.ignition-result.json
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-ignition-status-info
```
bgilbert: Maybe something like `coreos-ignition-write-issue`?  The service name will be user-visible, and users may want to mask the unit in their configs.

overlay.d/05core/usr/libexec/coreos-ignition-status-info
jlebon: WFM!

overlay.d/05core/usr/libexec/coreos-ignition-status-info
sohankunkerkar: I think we need to handle a condition if the line is empty

bgilbert: Ignition shouldn't be logging empty warning lines.  Are you seeing cases where that occurs?

sohankunkerkar: I haven't seen it anywhere but wanted to cover that boundary case (just in case)

sohankunkerkar: Fixed

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
```diff
@@ -23,5 +23,5 @@ enable bootupd.socket
 # Ideally it should have been added as part of base Fedora - but since it was arch specific, it was not added: https://bugzilla.redhat.com/show_bug.cgi?id=1433859
 enable rtas_errd.service
 enable clevis-luks-askpass.path
-# Provide information if no ignition is provided
-enable coreos-check-ignition-config.service
+# Provide status information about the Ignition run
+enable coreos-ignition-status-info.service
```
bgilbert: Stale.

overlay.d/05core/usr/libexec/coreos-ignition-write-issue
```diff
@@ -42,6 +42,14 @@ else
         >> /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+# This is for displaying Ignition warnings on the console
+# TODO: Find a way to query journald logs before the system switches to real root
```
bgilbert: Wording nit: this makes it sound like the query would happen before switching root.  Maybe "find a way to query only journal entries recorded before the system..."?

overlay.d/05core/usr/libexec/coreos-ignition-write-issue
```diff
@@ -42,6 +42,14 @@ else
...
+journalctl -t ignition -o json-pretty -b 0 -p warning | jq -r '.MESSAGE' | while read 
+line; do
```
bgilbert: Stray line break.

overlay.d/05core/usr/libexec/coreos-ignition-write-issue
```diff
@@ -42,6 +42,14 @@ else
...
+        echo "Ignition: $line" >> /run/issue.d/30_coreos_ignition_warnings.issue
+    done
+
 # Our makeshift way of getting /run/issue.d semantics. See:
 # https://github.com/coreos/console-login-helper-messages/blob/e06fc88ae8fbcc3a422bc8c686f70c15aebb9d9a/usr/lib/console-login-helper-messages/issue.defs#L8-L17
 ln -sf /run/issue.d/30_coreos_ignition_provisioning.issue /etc/issue.d/
+ln -sf /run/issue.d/30_coreos_ignition_warnings.issue /etc/issue.d/
```
bgilbert: If we don't create the file, we still create the symlink.  I suspect things will still work correctly, but it's a bit confusing; maybe we should conditionalize on the existence of the file?

bgilbert (commented):
Looks good generally.  A couple bugs, and a couple things I hadn't noticed earlier.

overlay.d/05core/usr/libexec/coreos-ignition-write-issue
```diff
@@ -42,6 +42,16 @@ else
         >> /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+# This is for displaying Ignition warnings on the console
+# TODO: find a way to query journal entries recorded before the system switches to real root
+journalctl -t ignition -o json-pretty -b 0 -p warning | jq -r '.MESSAGE' | while read line; do
```
jlebon: Minor/optional: just `json` would work fine here since we're parsing it anyway.
But actually... I think `-o cat` is closer to what we want? Then we can drop the `jq` part of the pipeline entirely.

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
```diff
@@ -23,5 +23,5 @@ enable bootupd.socket
 # Ideally it should have been added as part of base Fedora - but since it was arch specific, it was not added: https://bugzilla.redhat.com/show_bug.cgi?id=1433859
 enable rtas_errd.service
 enable clevis-luks-askpass.path
-# Provide information if no ignition is provided
-enable coreos-check-ignition-config.service
+# Provide status information about the Ignition run
+enable coreos-ignition-write-issue.service
```
jlebon: Minor/optional: should the name of this unit and its associated script be `...-write-issues` since it's more than one? :)

jlebon (dismissed):
Some comments but LGTM as is too. Nice work!

I think it wouldn't be too hard to write an external test for this, right? E.g. a test with an Ignition config that we know generates a warning and verify that an issue file with the expected content is created.

sohankunkerkar:
> Some comments but LGTM as is too. Nice work!
> 
> I think it wouldn't be too hard to write an external test for this, right? E.g. a test with an Ignition config that we know generates a warning and verify that an issue file with the expected content is created.

Yeah, I will do this as a follow-up PR, sounds good?

jlebon (dismissed):
Sure, SGTM!

overlay.d/05core/usr/libexec/coreos-ignition-write-issues
```diff
@@ -42,6 +42,16 @@ else
         >> /run/issue.d/30_coreos_ignition_provisioning.issue
 fi
 
+# This is for displaying Ignition warnings on the console
+# TODO: find a way to query journal entries recorded before the system switches to real root
+journalctl -t ignition -o cat -b 0 -p warning | while read line; do
+    echo "Ignition: $line" >> /run/issue.d/30_coreos_ignition_warnings.issue
+done
```
jlebon: Actually sorry, there's an issue here (pun intended): this service runs on every boot, but we want to do this only on the first boot.

Also, the use of `/run` for the other issue is because it's dynamic data that changes on every boot. But the data here isn't dynamic.

So I think we need to:
1. make this block conditional on the first boot only
2. output to `/etc/issue.d` instead of `/run/issue.d`

sohankunkerkar: Fixed. Just one follow-up question, I think we don't need `-b 0`, right?

jlebon (approved):

overlay.d/05core/usr/libexec/coreos-ignition-write-issues
jlebon: Yup, correct!

dustymabe:
Nice work @sohankunkerkar

Issue #1499, opened 2022-02-08
s390x/NetworkManager: Connections do not work with inconsistent names e.g. eth1


jschintag:
When running the current ext.config.networking.* tests i see that the connections in the butane configs are picked up by Network Manager, but they are not linked to devices. This means that when the config calls for `eth1`, it creates the connection `eth1`, but then it goes on and creates `enc4` as a device, resulting in `ip addr show` showing `enc4`, instead of as expected `eth1`.

miabbott:
@HuijingHei since you worked on some of these tests, could you have a look at how they work on the multi-arch platforms?

HuijingHei:
Sure, I will try, but I do not have s390x platform, I will borrow one from beaker

HuijingHei:
On s390x platform, run with CoreOS 410.84.202202090402-0 using `cosa kola qemuexec --qemu-image test.qcow2 --kargs net.ifnames=0 -c`, seems some kernel args are missed in the first boot

- After first booted and login, check `/proc/cmdline` kernel arg does not include `net.ifnames=0`, get `enc2` device

```
[core@cosa-devsh ~]$ cat /proc/cmdline | more
random.trust_cpu=on ignition.platform.id=qemu $ignition_firstboot ostree=/ostree/boot.1/rhcos/16a78463454c349f8a44c08964f1500e808c0dec8f8af24bf4fd2b0c0d58e583/0 ignition.firstboot
[core@cosa-devsh ~]$ rpm-ostree kargs list --deploy-index=0       
random.trust_cpu=on ignition.platform.id=qemu $ignition_firstboot ostree=/ostree/boot.1/rhcos/16a78463454c349f8a44c08964f1500e808c0dec8f8af24bf4fd2b0c0d58e583/0 net.ifnames=0 root=UUID=e17848c2-98a2-4779-bcf1-6c8927a9f5ef rw rootflags=prjquota boot=UUID=d3d02f95-d665-48dd-9689-1ff7881948ad
```

- Exec `systemctl reboot`, and check `/proc/cmdline` kernel arg include `net.ifnames=0`, get `eth0` device

```
[core@cosa-devsh ~]$ cat /proc/cmdline 
[...]

PR #1623, opened 2022-03-24
s390x: Remove fwupd package

jschintag:
On s390x, we do not need the fwupd package and newer versions of it cause problems in RHCOS, see https://github.com/openshift/os/issues/752.

I also put in a commit to not run the `ext.config.networking.kargs-rd-net` test on s390x, as it suffers from the same root problem of not correctly applying extra kernel arguments as the other new networking tests. See https://github.com/coreos/fedora-coreos-config/issues/1499#issuecomment-1076338932

### DIFF
--- a/tests/kola/networking/kargs-rd-net
+++ b/tests/kola/networking/kargs-rd-net
@@ -1,14 +1,22 @@
 #!/bin/bash
-# kola: { "platforms": "qemu", "appendFirstbootKernelArgs": "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8"}
-# 
+# kola: { "platforms": "qemu", "appendFirstbootKernelArgs": "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8", "architectures": "!s390x"}
+#
 # Veirfy rd.net.timeout.dhcp and rd.net.dhcp.retry are supported
-# by NetworkManager. Append them to kernel parameter when boot, 
-# get total timeout is `timeout * retry`, 30*8(240) seconds 
+# by NetworkManager. Append them to kernel parameter when boot,
+# get total timeout is `timeout * retry`, 30*8(240) seconds
 # in this test scenario
 # See:
 # - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/559
 # - https://bugzilla.redhat.com/show_bug.cgi?id=1879094#c10
 # - https://bugzilla.redhat.com/show_bug.cgi?id=1877740
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+# - appendFirstbootKernelArgs: "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8"
+#   - The functionality we're testing here.
+# - architectures: !s390x
+#   - appendFirstbootKernelArgs doesn't work on s390x
+#   - https://github.com/coreos/coreos-assembler/issues/2776
 
 set -xeuo pipefail
 
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -14,14 +14,16 @@ packages:
  - linux-firmware
  # rpm-ostree
  - rpm-ostree nss-altfiles
- # firmware updates
- - fwupd
 
 # bootloader
 packages-aarch64:
   - grub2-efi-aa64 efibootmgr shim
+  # firmware updates
+  - fwupd
 packages-ppc64le:
   - grub2 ostree-grub2
+  # firmware updates
+  - fwupd
 packages-s390x:
   # On Fedora, this is provided by s390utils-core. on RHEL, this is for now
   # provided by s390utils-base, but soon will be -core too.
@@ -29,3 +31,5 @@ packages-s390x:
 packages-x86_64:
   - grub2 grub2-efi-x64 efibootmgr shim
   - microcode_ctl
+  # firmware updates
+  - fwupd
--- a/tests/kola/networking/force-persist-ip/test.sh
+++ b/tests/kola/networking/force-persist-ip/test.sh
@@ -9,13 +9,21 @@ set -xeuo pipefail
 #   eth1 has the static IP address via kargs
 
 # https://bugzilla.redhat.com/show_bug.cgi?id=1958930#c29
-# - We use net.ifnames=0 to disable consistent network naming here because on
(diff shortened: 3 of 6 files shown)


travier (dismissed):

travier (commented):
Do we need the dependencies that gets pulled-in on s390x? dbx tool, etc.? Probably not but we should check.

dustymabe:
hold this briefly - i've got some suggestions but need to drop kid off at school first.

tests/kola/networking/kargs-rd-net
```diff
@@ -1,9 +1,9 @@
...
+# kola: { "platforms": "qemu", "appendFirstbootKernelArgs": "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8", "architectures": "!s390x"}
+#
 # Veirfy rd.net.timeout.dhcp and rd.net.dhcp.retry are supported
-# by NetworkManager. Append them to kernel parameter when boot, 
-# get total timeout is `timeout * retry`, 30*8(240) seconds 
+# by NetworkManager. Append them to kernel parameter when boot,
+# get total timeout is `timeout * retry`, 30*8(240) seconds
 # in this test scenario
 # See:
 # - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/559
```
dustymabe, line 9: below this let's add some explanation of the arguments in the kola json:

```diff
diff --git a/tests/kola/networking/kargs-rd-net b/tests/kola/networking/kargs-rd-net
index d2e7781..3d14e7f 100755
--- a/tests/kola/networking/kargs-rd-net
+++ b/tests/kola/networking/kargs-rd-net
@@ -9,6 +9,14 @@
 # - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/559
 # - https://bugzilla.redhat.com/show_bug.cgi?id=1879094#c10
 # - https://bugzilla.redhat.com/show_bug.cgi?id=1877740
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+# - appendFirstbootKernelArgs: "rd.net.timeout.dhcp=30 rd.net.dhcp.retry=8"
+#   - The functionality we're testing here.
+# - architectures: !s390x
+#   - appendFirstbootKernelArgs doesn't work on s390x
+#   - https://link/to/newissue
 
 set -xeuo pipefail
 
```
Update the link with one describing the feature request for kargs and firstbootkargs to work on s390x.

dustymabe:
Is fwupd supposed to work at all on s390x? If not then should we try to get the maintainers to stop building it on that platform (that just leads to confusion)?

jschintag:
@dustymabe I added you description to the new test, as well as created a new commit to update the description on the other network tests.

dustymabe (approved):
looks great! - thanks!

dustymabe:
> Do we need the dependencies that gets pulled-in on s390x? dbx tool, etc.? Probably not but we should check.

@travier - good question - maybe we can handle that as a followup

dustymabe:
> Is fwupd supposed to work at all on s390x? If not then should we try to get the maintainers to stop building it on that platform (that just leads to confusion)?

Still interested in the answer to this question ^^

travier:
> > Do we need the dependencies that gets pulled-in on s390x? dbx tool, etc.? Probably not but we should check.
> 
> @travier - good question - maybe we can handle that as a followup

I would prefer we confirm that before we break things on s390x. The workaround for fwupd that triggered this removal is in https://github.com/openshift/os/pull/757

dustymabe:
> > > Do we need the dependencies that gets pulled-in on s390x? dbx tool, etc.? Probably not but we should check.
> > 
> > 
> > @travier - good question - maybe we can handle that as a followup
> 
> I would prefer we confirm that before we break things on s390x. The workaround for fwupd that triggered this removal is in [openshift/os#757](https://github.com/openshift/os/pull/757)

ahh I see what you are saying. I thought you were asking if there were other dependencies we should remove from our manifest list as part of this, but you were saying that there are unnamed deps that will drop out and do we really need those?

jschintag:
> > > Do we need the dependencies that gets pulled-in on s390x? dbx tool, etc.? Probably not but we should check.
> > 
> > 
> > @travier - good question - maybe we can handle that as a followup
> 
> I would prefer we confirm that before we break things on s390x. The workaround for fwupd that triggered this removal is in [openshift/os#757](https://github.com/openshift/os/pull/757)

@travier Actually that is a separate issue for x86. On s390x, it has neither `ipmi_si` nor `ipmi_devintf`. On x86 it has/needs `ipmi_devintf`. That is why i split the 2 issues and why i'm removing the whole package on s390x as it is not needed.

jschintag:
> > Is fwupd supposed to work at all on s390x? If not then should we try to get the maintainers to stop building it on that platform (that just leads to confusion)?
> 
> Still interested in the answer to this question ^^

Sadly, i don't really know the answer. IPMI might be used for vBMC with libvirt/openstack to remote provision kvm guests. Otherwise i'm afraid firmware updates is not something i ever got to do on a s390x machine. Largely because it is to my knowledge done on the machine, not on lpar level.

jschintag:
> Do we need the dependencies that gets pulled-in on s390x? dbx tool, etc.? Probably not but we should check.

I looked through the dependencies of the fwupd on rhel8.6 and compared it with what i could find on the RHCOS 4.11 image without fwupd. Here is what is now not installed without fwupd:
```
config(fwupd) = 1.5.9-1.el8_4 # provided by fwupd
libfwupd.so.2()(64bit)        # provided by fwupd
libfwupdplugin.so.1()(64bit)  # provided by fwupd
libgcab-1.0.so.0()(64bit)
libgudev-1.0.so.0()(64bit)    # libudev is still installed
libgusb(s390-64) >= 0.2.11    # libusb is still installed
libjcat.so.1()(64bit)         # provided by fwupd
libmm-glib.so.0()(64bit)
libqmi-glib.so.5()(64bit)
libxmlb(s390-64) >= 0.1.3
```
Since these dependencies are only librarys, it's safe to say that we don't need them.

cgwalters (approved):
Makes sense to me!

> I also put in a commit to not run the ext.config.networking.kargs-rd-net test on s390x, as it suffers from the same root problem of not correctly applying extra kernel arguments as the other new networking tests. See https://github.com/coreos/fedora-coreos-config/issues/1499#issuecomment-1076338932

This is fine by me, but since CI wouldn't gate on it here, I think it would have been cleaner as a distinct PR.  (You've got separate commits, so again that's fine)

PR #1626, opened 2022-03-24
[stable] overrides: fast-track rust-coreos-installer-0.13.1-3.fc35

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,6 +9,18 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  coreos-installer:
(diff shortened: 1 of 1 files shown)


dustymabe:
I don't understand this. I feel pretty strongly that we shouldn't go around our normal release process on this one.

dustymabe:
are you suggesting we do an ad-hoc release?

jlebon:
Discussed this with @dustymabe OOB. While the risk is fairly low, it's not critical to have this fixed ASAP. Users that hit https://github.com/coreos/fedora-coreos-tracker/issues/1116 can use the next `testing` release until it hits `stable`.

PR #1627, opened 2022-03-24
[testing] overrides: fast-track rust-coreos-installer-0.13.1-3.fc35

coreosbot-releng:


sohankunkerkar (approved):

jlebon:
This will be part of promotion so isn't necessary.

PR #1634, opened 2022-03-25
README.md: update section about repo

jlebon:
Tweak the section talking about how this repo works to reflect the
latest state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,16 +22,19 @@ and [tooling](https://github.com/coreos/fedora-coreos-tracker/blob/main/stream-t
 docs for more information about streams.
 
 All file changes in `testing-devel` are propagated to other
-branches (to `bodhi-updates` through
+branches (to `next-devel`, `branched`, and `rawhide` through
 [config-bot](https://github.com/coreos/fedora-coreos-releng-automation/tree/main/config-bot),
-and to `testing` through usual promotion), with the
-following exceptions:
-- `manifest.yaml`: contains the stream "identity", such as
-  the ref, additional commit metadata, and yum input repos.
-- lockfiles (`manifest-lock.*` files): lockfiles are
-  imported from `bodhi-updates` to `testing-devel`.
-  Overrides (`manifest-lock.overrides.*`) are manually
-  curated.
+and to `testing` and eventually `stable` through usual
+promotion), with the following exceptions:
+- `manifest.yaml`: contains the stream's name, yum repos
+  used during composes, and the `releasever`.
+- lockfiles (`manifest-lock.*` files): on `testing-devel`
+  and `next-devel`, lockfiles are pushed by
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,16 +22,18 @@ and [tooling](https://github.com/coreos/fedora-coreos-tracker/blob/main/stream-t
...
-  imported from `bodhi-updates` to `testing-devel`.
-  Overrides (`manifest-lock.overrides.*`) are manually
-  curated.
+and to `testing` and eventually `stable` through usual
+promotion), with the following exceptions:
+- `manifest.yaml`: contains the stream's name, yum repos
+  used during composes, and the `releasever`.
+- lockfiles (`manifest-lock.*` files): lockfiles are pushed
+  to `testing-devel` and `next-devel` by
+  [the `bump-lockfile` job](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile).
```
dustymabe: There's some nuance here. They get pushed to `testing-devel` and `next-devel` through bump-lockfile, but then they get promoted to the other branches through normal promotion.

README.md
```diff
@@ -22,16 +22,19 @@ and [tooling](https://github.com/coreos/fedora-coreos-tracker/blob/main/stream-t
...
+promotion), with the following exceptions:
+- `manifest.yaml`: contains the stream's name, yum repos
+  used during composes, and the `releasever`.
+- lockfiles (`manifest-lock.*` files): on `testing-devel`
+  and `next-devel`, lockfiles are pushed by
+  [the `bump-lockfile` job](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile).
+  Production streams receive them as part of usual
+  promotion. Overrides (`manifest-lock.overrides.*`) are
+  managed independently with the help of some GitHub Actions
+  (see sections below).
```
dustymabe, line 37: Still a bit more nuance.. Here's my stab at adding some more clarity:

```suggestion
- lockfiles (`manifest-lock.$ARCH.yaml` files): on `testing-devel`
  and `next-devel`, lockfiles are pushed by
  [the `bump-lockfile` job](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile).
  Production streams receive them as part of usual
  promotion.
- overrides lockfile (`manifest-lock.overrides.yaml`):
  On `testing-devel` and `next-devel` this file is managed
  by hand or updated by some helpful GitHub Actions (see
  sections below). For other branches this file is promoted
  with normal promotion, but can also be updated independently
  if a need arises for a change in the package set for an ad-hoc
  release.
```

Let me know if you think it helps.

jlebon: Hmm maybe. I wonder if we should fold some of those details into lower sections instead and keep this bit lighter. Maybe let's just get this in and iterate there? Seems like a good improvement overall to what we currently have.

dustymabe: sounds good

dustymabe (approved):
LGTM

PR #1641, opened 2022-03-28
[branched] lockfiles: drop graduated overrides 🎓

coreosbot-releng:


dustymabe (approved):
LGTM

dustymabe:
This branch is no longer being built so we'll ignore CI.

dustymabe:
> This branch is no longer being built so we'll ignore CI.

Though I guess that begs the question of if there is some way to disable this GitHub action on it while it's not being used.

PR #1638, opened 2022-03-25
workflows: allow magic promotion text anywhere in PR title

bgilbert:
The PR title might be prefixed with a branch name.

PR #1644, opened 2022-03-28
[stable] tree: promote changes from testing at 7f891fa7e9574ca95c1218d344a5671f461152b1

coreosbot-releng:
Triggered by @ravanelli in https://github.com/coreos/fedora-coreos-streams/issues/473.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,6 +1,6 @@
 // Documentation: https://github.com/coreos/coreos-ci/blob/main/README-upstream-ci.md
 
-cosaPod(image: 'quay.io/coreos-assembler/coreos-assembler:v0.13.0') {
+cosaPod {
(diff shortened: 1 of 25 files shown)


dustymabe:
We forgot to mention this earlier but we need an extra commit here. See https://github.com/coreos/fedora-coreos-streams/issues/473#issuecomment-1080630962

jlebon (approved):
Manually verifying until https://github.com/coreos/fedora-coreos-config/pull/1638 gets to stable:

```
$ git diff 7f891fa7e9574ca95c1218d344a5671f461152b1
index a3ae0341..04a15ff3 100644
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,5 +1,5 @@
 variables:
-  stream: testing
+  stream: stable
   prod: true

 releasever: 35
```

PR #1650, opened 2022-03-31
Use bigtime=1 for XFS by default

cgwalters:
I was randomly looking at kernel changes recently and noticed
that XFS today has the [Year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem)
by default.

While one would hope that most FCOS systems are deployed in a
reprovisionable environment, and are reprovisioned...let's still
be forward looking here and enable this by default - but only
for the testing-devel stream for now.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -2,3 +2,10 @@
 # similarly to manifest.yaml. Unlike image-base.yaml, which is shared by all
 # streams.
 include: image-base.yaml
+
+# See https://github.com/coreos/fedora-coreos-tracker/pull/51
+rootfs: "xfs"
+# See https://man7.org/linux/man-pages/man8/xfs_admin.8.html
+# Let's opt-in to supporting bigger timestamps as a forward looking
+# system.
+rootfs-args: "-m bigtime=1"
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -12,6 +12,22 @@ fatal() {
     exit 1
 }
 
+_TEST_TEMPDIR=""
+ensure_tmpdir() {
+    local tmpdir
+    tmpdir=$(mktemp -d)
+    cd "${tmpdir}"
+    touch .tmpdir
+}
+
+cleanup_tmpdir() {
+    tmpdir=$(pwd)
+    cd /
+    if test -f "${tmpdir}"/.tmpdir; then
+        rm -rf "${tmpdir}"
+    fi
+}
+
 get_ipv4_for_nic() {
     local nic_name=$1
     local ip=$(ip -j addr show ${nic_name} | jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local')
--- a/tests/kola/disks/root-xfs-bigtime
+++ b/tests/kola/disks/root-xfs-bigtime
@@ -1,0 +1,21 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+# Test that we have bigtime=1 for XFS
+# xref https://man7.org/linux/man-pages/man8/xfs_admin.8.html
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+tmpd=$(mktemp -d) && trap 'rm -rf ${tmpd}' EXIT
+cd "${tmpd}"
+
+if stat -f / | grep 'Type: xfs'; then
+    xfs_info / > xfsinfo.txt
+    if ! grep -q bigtime=1 xfsinfo.txt; then
+        fatal "failed to find bigtime=1 for /"
+    fi
+else
+    echo "notice: rootfs was not xfs"
+fi
+ok xfs bigtime


cgwalters:
Requires https://github.com/coreos/coreos-assembler/pull/2789

bgilbert:
Is the intention to promote this through the streams at the usual rate, or to bake in testing-devel for a while?  If the former, we could potentially build the next stable with an older cosa and avoid adding the `image.yaml` setting.

dustymabe:
> Is the intention to promote this through the streams at the usual rate, or to bake in testing-devel for a while? If the former, we could potentially build the next stable with an older cosa and avoid adding the `image.yaml` setting.

Along that same vein, do we see any potential risk here? If we do anything experimental that we don't plan to promote to `stable` in two weeks then I'd prefer to do it in `next-devel`/`next`.

IOW I'd like to keep `testing` exactly the same as the next `stable`.

jlebon:
It looks like our boot filesystem is also subject to the 2038 problem:

```
[root@cosa-devsh ~]# dumpe2fs /dev/disk/by-label/boot | grep 'Inode size'
Inode size:               128
```

`mkfs.ext4(8)` says:

> File systems with an inode size of 128 bytes do not support timestamps beyond January 19, 2038. Inodes which are 256 bytes or larger will support extended timestamps, project id's, and the ability to store some extended attributes in the inode table for improved performance.
>
> The default inode size is controlled by the mke2fs.conf(5) file. In the mke2fs.conf file shipped with e2fsprogs, the default inode size is 256 bytes for most file systems, except for small file systems where the inode size will be 128 bytes.

I guess our bootfs' size is categorized as small. We could tweak it (`-I 256`), but the manpage also says:

> The larger the inode-size the more space the inode table will consume, and this reduces the usable space in the filesystem and can also negatively impact performance.

So doesn't seem worth doing anything just yet.

Agreed with @dustymabe on letting it bake in `next-devel` first.

cgwalters:
https://github.com/coreos/coreos-assembler/pull/2793

dustymabe:
I'm wondering if this isn't a change we should push Fedora to make at a project level rather than one-offing it here.

IOW what are the reasons (if any) Fedora isn't doing this already?

cgwalters:
> I'm wondering if this isn't a change we should push Fedora to make at a project level rather than one-offing it here.

Agree, it likely makes sense to expand this beyond FCOS.  However - are we leaders or followers?  I think it depends.   But on things like this, it can make some sense for us to lead.

> IOW what are the reasons (if any) Fedora isn't doing this already?

We're Fedora too, so I'd say "other editions/builds" or so.  I suspect the answer is just that "no one thought of it until now".  But I can ask around!

cgwalters:
On the other hand, Workstation defaulting to btrfs (which does not have this problem by default at least today, and that may have been true for years) means it is a bit more on us.

dustymabe:
> Agree, it likely makes sense to expand this beyond FCOS. However - are we leaders or followers? I think it depends. But on things like this, it can make some sense for us to lead.

I think we *can* lead, but I view this a lot like Red Hat's "upstream first" mentality. In general we don't want to just make changes ourselves and not get them back "upstream" such that we diverge from other editions. If we do that we're adding to our maintenance burden and it's unsustainable (FCOS will suffer in one way or another). 

So typically what happens if there is a change we'd want I'd say let's propose it as a change in all of Fedora and hopefully it gets accepted. Once accepted we don't have to wait for the next major version of Fedora, we can "backport" that change to FCOS now if we want.

However, if the change doesn't get accepted we need to take a long hard look at why. Are there good reasons that apply to us too? Ultimately we can decide to make the change anyway, but we're much wiser for going through the process.

sandeen:
As always, please feel free to reach out to the filesystem experts who are close at hand for questions like this.

bigtime will be enabled on xfs by default as of the xfsprogs-5.15.0 release which should happen very soon. (I'm late thanks to PTO.) I'll be happy to push that releases to existing Fedora releases as needed.

As for ext4, I'll need to dig a little more, but I believe it's automatically choosing the "small" profile for small filesystems, which leads to the small inodes and smaller timestamps.  If typical boot filesystems encounter this problem it might be worth bringing this issue up on the ext devel list.

Thanks,
-Eric

sandeen:
xfsprogs-5.15.0 with bigtime on by default is now built in rawhide. (It also provides an upgrade path for older filesystems.)

And - ext4 chooses the "small" profile and smaller inodes for anything under 512MB. 512MB and above should get larger inodes, along with extended timestamps.

-Eric

dustymabe:
Thanks @sandeen!

@cgwalters - For XFS should we just let this passively come to us?

For ext4, should we consider forcing it to use a different profile even though we're under 512M for the boot FS? Alternatively we could consider bumping our /boot FS to 512M.

bgilbert:
Resizing a default filesystem would have implications for the Butane RAID template, and for our minimum recommended disk size such that the root filesystem is kept &ge; 8 GiB.

dustymabe:
I'm going to close this as it seems like the change is coming to us anyway. If we want to do anything for ext4 let's open a tracker issue and discuss what to do there.

PR #1652, opened 2022-04-01
[rawhide] overrides: fast-track iproute-5.17.0-2.fc37

coreosbot-releng:

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1142


jlebon (approved):

dustymabe:
I need to address the kernel circular locking dep warnings. Will do that and address this in https://github.com/coreos/fedora-coreos-config/pull/1654

PR #1657, opened 2022-04-05
manifest: Remove libvarlink-util in next stream

gursewak1997:
Remove libvarlink-util package for `next` stream.


jlebon:
This file will be overwritten on the next `next` promotion. See https://github.com/coreos/fedora-coreos-config#about-this-repo for more info.

This could be a good use case for conditional includes. Something like:

```yaml
conditional-include:
  - if: stream == "testing-devel"
    include: varlink.yaml
  - if: stream == "testing"
    include: varlink.yaml
  - if: stream == "stable"
    include: varlink.yaml
```

And then a new `varlink.yaml` which is just:

```yaml
packages:
  - libvarlink-util
```

And then we can delete those bits once the one month period is over.

dustymabe:
and commit those changes in a PR to `testing-devel`.

dustymabe:
closing in favor of #1660

PR #1660, opened 2022-04-05
manifest: Drop libvarlink-util for next stream

gursewak1997:
Include the libvarlink package for selected streams conditionally. Removed the package for next stream as an initial step to fully drop it.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1130

### DIFF
--- a/manifests/varlink.yaml
+++ b/manifests/varlink.yaml
@@ -1,0 +1,5 @@
+# Include libvarlink-util package selectively
+# before we drop it from the remaining streams 
+# https://github.com/coreos/fedora-coreos-tracker/issues/1130
+packages:
+  - libvarlink-util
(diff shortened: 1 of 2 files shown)


manifests/varlink.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
```
dustymabe, line 4: Let's add a comment at the top of this file giving some context.

dustymabe (commented):
LGTM one suggestion.

manifests/varlink.yaml
gursewak1997: Should I just add the previous comment `# Remote IPC for podman` to this file or should I add something about why we made a separate file and adding a link to package removal issue.

dustymabe: I would add a comment about why we made it a separate file and the link to the package removal issue.

dustymabe (approved):
LGTM

PR #1662, opened 2022-04-06
tests/kola/kubernetes/kube-watch: rework journal grepping

### LINKED ISSUE #1637: There are some flakes happening with ext.config.kubernetes.kube-watch
kube-watch log
```
Mar 25 18:05:34 qemu0 kola-runext-test.sh[6178]: + fatal ‘kube-watch.service did not successfully activate’
Mar 25 18:05:34 qemu0 kola-runext-test.sh[6178]: + echo ‘kube-watch.service did not successfully activate’
Mar 25 18:05:34 qemu0 kola-runext-test.sh[6178]: kube-watch.service did not successfully activate
Mar 25 18:05:34 qemu0 kola-runext-test.sh[6178]: + exit 1
```
Journal log
```
Mar 25 18:05:34.050784 unknown[6173]: Awaiting events
Mar 25 18:05:34.051438 kola-runext-test.sh[6178]: + '[' inactive == activating ']'
Mar 25 18:05:34.051734 kola-runext-test.sh[6182]: ++ systemctl is-active kube-watch.service
Mar 25 18:05:34.056913 systemd[1]: Starting kube-watch.service...
Mar 25 18:05:34.061208 kola-runext-test.sh[6178]: + '[' activating '!=' active ']'
Mar 25 18:05:34.061278 kola-runext-test.sh[6178]: + fatal 'kube-watch.service did not successfully activate'
Mar 25 18:05:34.061335 kola-runext-test.sh[6178]: + echo 'kube-watch.service did not successfully activate'
Mar 25 18:05:34.061370 kola-runext-test.sh[6178]: kube-watch.service did not successfully activate
Mar 25 18:05:34.061405 kola-runext-test.sh[6178]: + exit 1
Mar 25 18:05:34.061659 systemd[1]: kola-runext-28.service: Main process exited, code=exited, status=1/FAILURE
Mar 25 18:05:34.061831 systemd[1]: kola-runext-28.service: Failed with result 'exit-code'.
Mar 25 18:05:34.062933 echo[6183]: This is the kube-watch unique id: 27a259a8-7f2d-4144-8b8f-23dd201b630c
Mar 25 18:05:34.064000 audit[1]: SERVICE_STOP pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=kola-runext-28 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'
Mar 25 18:05:34.066276 systemd[1]: Finished kube-watch.service
```

[ext.config.kubernetes.kube-watch.txt](https://github.com/coreos/fedora-coreos-config/files/8353024/ext.config.kubernetes.kube-watch.txt)
[journal.txt](https://github.com/coreos/fedora-coreos-config/files/8353026/journal.txt)

jlebon:
The way we check the systemd unit status is racy. Instead, wait up to 30
seconds for the journal message to appear. And then verify that the
service is active.

While we're here, use `journalctl`'s built-in `--grep`.

Also while we're here, use `-u kube-watch.service` to be more
constrained, which allows us to drop the unique UUID.

Closes: https://github.com/coreos/fedora-coreos-config/issues/1637

### DIFF
--- a/tests/kola/kubernetes/kube-watch/config.bu
+++ b/tests/kola/kubernetes/kube-watch/config.bu
@@ -12,14 +12,11 @@ systemd:
     - name: kube-watch.service
       # This is for verifying that `kubernetes_file_t` labeled files can be
       # watched by systemd
-      # NOTE: we've observed a race where the journal message shows up as
-      # coming from `echo` rather than `kube-watch`, so we're embedding
-      # a UUID in the message to make it easier to find.
       # See: https://github.com/coreos/fedora-coreos-tracker/issues/861
       # See: https://github.com/containers/container-selinux/issues/135
       contents: |
         [Service]
-        ExecStart=/usr/bin/echo "This is the kube-watch unique id: 27a259a8-7f2d-4144-8b8f-23dd201b630c"
+        ExecStart=/usr/bin/touch /var/tmp/kube-watched
         RemainAfterExit=yes
         Type=oneshot
         [Install]
--- a/tests/kola/kubernetes/kube-watch/test.sh
+++ b/tests/kola/kubernetes/kube-watch/test.sh
@@ -17,21 +17,22 @@ ok "kube-watch.path successfully activated"
 touch /etc/kubernetes/kubeconfig
 ok "successfully created /etc/kubernetes/kubeconfig"
 
-# If we check the status too soon it could still be activating..
-# Sleep in a loop until it's done "activating"
-while [ "$(systemctl is-active kube-watch.service)" == "activating" ]; do
-    echo "kube-watch is activating. sleeping for 1 second"
+# Give the service 30 seconds to activate
+active=0
+for i in {1..30}; do
+    if systemctl is-active kube-watch.service; then
+        active=1
+        break
+    fi
     sleep 1
 done
-if [ "$(systemctl is-active kube-watch.service)" != "active" ]; then
+if [[ $active != 1 ]]; then
+    systemctl status kube-watch.service
     fatal "kube-watch.service did not successfully activate"
 fi
 ok "kube-watch.service activated successfully"
 
-# NOTE: we've observed a race where the journal message shows up as
-# coming from `echo` rather than `kube-watch`, so we're embedding
-# a UUID in the message to make it easier to find.
-if ! journalctl -o cat -b | grep 27a259a8-7f2d-4144-8b8f-23dd201b630c; then
-    fatal "kube-watch.service did not print message to journal"
+if ! test -e /var/tmp/kube-watched; then
+    fatal "/var/tmp/kube-watched does not exist"
 fi
-ok "Found message from kube-watch.service in journal"
+ok "/var/tmp/kube-watched exists"


dustymabe:
> Also while we're here, use `-u kube-watch.service` to be more constrained, which allows us to drop the unique UUID.

I think we did this UUID thing for a reason. See 2913472

jlebon:
> > Also while we're here, use `-u kube-watch.service` to be more constrained, which allows us to drop the unique UUID.
> 
> I think we did this UUID thing for a reason. See [2913472](https://github.com/coreos/fedora-coreos-config/commit/2913472e6e17d98659a4da2391747f270130723f)

Hmm, I'm unsure about that commit. @miabbott Do you have a link with more information re. the failure you saw that brought you to https://github.com/coreos/fedora-coreos-config/pull/1159? I'm interested to see the exact error we hit there.

jlebon:
But I guess I'm fine keeping the UUID too, even if I don't quite follow why it's necessary. So updated! :arrow_up:

miabbott:
> Hmm, I'm unsure about that commit. @miabbott Do you have a link with more information re. the failure you saw that brought you to #1159? I'm interested to see the exact error we hit there.

I don't have an issue that captures the problem and a search of my Slack history comes up empty. I think it was a conversation with Dusty via PM on IRC/Matrix.

I'm afraid my comments don't properly explain the error.  :cry:

dustymabe:
From what I remember basically when we were running `journalctl -u kube-watch` in the past we were finding that we wouldn't get the echo message, but if we searched the whole log we would find it, but it wouldn't be associated with the unit itself. Obviously some sort of systemd/journald issue/bug but we worked around it this way. I'm comfortable reverting back to the old way `with -u` to monitor if the original problem is reproducible (our tests will tell us).

dustymabe:
The way you have it now:

```
journalctl -u kube-watch.service --grep 27a259a8-7f2d-4144-8b8f-23dd201b630
```

isn't necessary. From what I remember (see https://github.com/coreos/fedora-coreos-config/pull/1662#issuecomment-1090776808) we either need to use `-u kube-watch.service` (which means we can drop the UUID) or we need to drop `-u kube-watch.service` and only use the UUID.

jlebon:
We could also not grep the journal at all and just give the unit 30 seconds to become active.

dustymabe:
> We could also not grep the journal at all and just give the unit 30 seconds to become active.

actually there's an idea. I do kind of like checking if the unit became "active" and one other measure of work. i.e. systemd tells us it ran and we also can see evidence that it ran. However, if the journal has been unreliable for us in the past then maybe we change our `echo` to a `touch` and just check if a file exists rather than querying journald

jlebon:
> > We could also not grep the journal at all and just give the unit 30 seconds to become active.
> 
> actually there's an idea. I do kind of like checking if the unit became "active" and one other measure of work. i.e. systemd tells us it ran and we also can see evidence that it ran.

Meh, I see where you're coming from but that kinda feels like overkill/out of scope for what we're testing here. But it's also super trivial to do, so done!

dustymabe (approved):
LGTM

PR #1663, opened 2022-04-06
manifests: mask systemd-network-generator.service for only fedora 35

gursewak1997:
Mask systemd-network-generator.service for only fedora 35.
See: https://github.com/coreos/fedora-coreos-tracker/issues/1059#issuecomment-1090602396


dustymabe (approved):
LGTM

PR #1664, opened 2022-04-06
manifest-lock.overrides: fast-track selinux-policy

gursewak1997:
Fast-track selinux-policy package
See: https://github.com/coreos/fedora-coreos-tracker/issues/1059#issuecomment-1090602396

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,16 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  selinux-policy:
+    evra: 36.6-1.fc36.noarch
+    metadata:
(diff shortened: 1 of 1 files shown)


dustymabe:
This isn't quite complete for a few reasons but don't worry. We have some helpers with this kind of routine behavior.

First we have [a GitHub Action](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) you can use to insert data into a form and it will handle the lockfile update for you. This is what we typically use if the update is simple (i.e. one override).

If it gets more complicated (i.e. you need to apply two overrides at once, or you need to remove an override and apply another etc.) you can run the same scripts behind the GH action locally. The script is in `ci/overrides.py` in this repo. You can run it from within a cosa container (think `cosa shell`). For syntax run `ci/overrides.py --help`.

For this one it would be something along the lines of:

```
./ci/overrides.py fast-track -r https://github.com/coreos/fedora-coreos-tracker/issues/1059#issuecomment-1090602396 https://bodhi.fedoraproject.org/updates/FEDORA-2022-690770081a 
```

dustymabe (approved):
LGTM

PR #1665, opened 2022-04-06
05core: remove automatic reboot after timeout

prestist:
Removed the timeout for reboot on ftb if in error state. 

https://github.com/coreos/fedora-coreos-tracker/issues/928


jlebon:
So for https://github.com/coreos/fedora-coreos-tracker/issues/928, the idea is that we only remove the timeout. But we still want e.g. the Ignition virtio journal dump bits (this is used by cosa to make it easier to surface initrd failures) and printing of the failed units.

jlebon:
So for https://github.com/coreos/fedora-coreos-tracker/issues/928, the idea is that we only remove the timeout. But we still want e.g. the Ignition virtio journal dump bits (this is used by cosa to make it easier to surface initrd failures) and printing of the failed units.

prestist:
@jlebon, sorry I misunderstood what the goal was. I went ahead and made the correction so the journal dump still happens.

I was wondering about the naming convention of the folder in modules.d. 

Where does the '99' come from, and would it make sense to just drop the '-timeout'?

bgilbert:
This is still removing the printing of failed units.  You'll need to go through the code and delete only the relevant parts, not just delete the whole file.

It would make sense to drop the `-timeout`.  `99` is a Dracut module priority, and AIUI only matters for resolving conflicts where two Dracut modules ship different versions of the same file.  I think the value is mostly arbitrary here.

prestist:
> This is still removing the printing of failed units. You'll need to go through the code and delete only the relevant parts, not just delete the whole file.
> 
> It would make sense to drop the `-timeout`. `99` is a Dracut module priority, and AIUI only matters for resolving conflicts where two Dracut modules ship different versions of the same file. I think the value is mostly arbitrary here.

prestist:
Closing this really quick, I am going through a workflow that updates pr, and I dont want to cause un-nessary reviews.

prestist:
> This is still removing the printing of failed units. You'll need to go through the code and delete only the relevant parts, not just delete the whole file.

Ah, yes @jlebon talked to me about that yesterday. I haven't yet gotten back to fix that.

> It would make sense to drop the `-timeout`. `99` is a Dracut module priority, and AIUI only matters for resolving conflicts where two Dracut modules ship different versions of the same file. I think the value is mostly arbitrary here.

Ah thank you! I was wondering about that.

bgilbert:
> Closing this really quick, I am going through a workflow that updates pr, and I dont want to cause un-nessary reviews.

We usually just leave them open, and post a comment when the PR is ready for re-review.

prestist:
> > Closing this really quick, I am going through a workflow that updates pr, and I dont want to cause un-nessary reviews.
> 
> We usually just leave them open, and post a comment when the PR is ready for re-review.

Ah okay, sorry, since I am going to be doing cosa - inits I was worried that I would be bugging everyone.

jlebon (commented):
This needs a rebase but it LGTM overall! For the commit message, how about something like `05core: stop automatic reboot after timeout`? And then in the commit message, you can include a short rationale and a link to the tracker issue.

Also the comment at the top of the `emergency-shell.sh` needs to be updated.

bgilbert:
LGTM generally. :+1: The module should be renamed from `99emergency-timeout`.

prestist:
> This needs a rebase but it LGTM overall! For the commit message, how about something like `05core: stop automatic reboot after timeout`? And then in the commit message, you can include a short rationale and a link to the tracker issue.
> 
> Also the comment at the top of the `emergency-shell.sh` needs to be updated.



> This needs a rebase but it LGTM overall! For the commit message, how about something like `05core: stop automatic reboot after timeout`? And then in the commit message, you can include a short rationale and a link to the tracker issue.
> 
> Also the comment at the top of the `emergency-shell.sh` needs to be updated.

First, thank you for the feedback! , small question about your pr comment.
 "... and then in the commit message, you can include a short rationale and a link to the tracker issue."
I was thinking something more like "05core: enter emergency shell on error during ftb"
In the past I have had smaller commit messages, do I want to add the bit about the rationale and the tracker issue to the message like this?
The result would be something like this =>
"05core: enter emergency shell on error during ftb. The reboot and consequently the timeout where removed to prevent the suppression of errors during the first time boot. fix:https://github.com/coreos/fedora-coreos-tracker/issues/928"

prestist:
> LGTM generally. 👍 The module should be renamed from `99emergency-timeout`.

Thank you for the feedback!

I took a swing at the folder name, let me know if it makes sense.

jlebon (dismissed):
LGTM.

For the commit message,

> 05core: enter emergency shell on error during FTB

(OOC, what does FTB stand for?)

I think it'd be better to focus the message more on what changed. In this case, it's the fact that we're no longer automatically rebooting after a timeout.

For the rationale and link, you would include it in the message body, not as part of the title (check out https://cbea.ms/git-commit/#separate).

jlebon (approved):
Great commit message, thanks!

PR #1674, opened 2022-04-13
tests: ensure that Ignition warnings are displayed on the console

sohankunkerkar:
Follow up to https://github.com/coreos/fedora-coreos-config/pull/1621#pullrequestreview-930860551
This adds kola test coverage for displaying Ignition warnings on the console.

### DIFF
--- a/tests/kola/clhm/ignition-warnings/config.bu
+++ b/tests/kola/clhm/ignition-warnings/config.bu
@@ -1,0 +1,16 @@
+variant: fcos
+version: 1.1.0
+systemd:
+  # This systemd unit doesn't have the Install
+  # section in it, so as part of the validation 
+  # step, Ignition will throw the following warning:
+  # 'warning at $.systemd.units.0.contents: unit "echo@.service" is enabled, but has no install section so enable does nothing'
+  units:
+    - name: echo.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=echo service template
+        [Service]
+        Type=oneshot
+        ExecStart=/bin/echo %i
--- a/tests/kola/clhm/ignition-warnings/data/commonlib.sh
+++ b/tests/kola/clhm/ignition-warnings/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../../data/commonlib.sh
--- a/tests/kola/clhm/ignition-warnings/test.sh
+++ b/tests/kola/clhm/ignition-warnings/test.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# kola: { "platforms": "qemu-unpriv", "allowConfigWarnings": true }
+# - platforms: qemu-unpriv 
+#   - This test should pass everywhere if it passes anywhere. 
+# - allowConfigWarnings: true 
+#   - We intentionally exclude the Install section from a systemd unit.  
+#     This is valid but not ideal, so Butane warns about it.  
+# This test ensures that Ignition warnings are displayed on the console.
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+WARN='\033[0;33m' # yellow
+RESET='\033[0m' # reset
+
+warningsfile="/etc/issue.d/30_coreos_ignition_warnings.issue"
+tmpd=$(mktemp -d)
+expectedwarningsfile="${tmpd}"/expectedwarningsfile
+
+# Check for the motd file
+if ! test -f ${warningsfile}; then
+    fatal "not found Ignition warnings issue file"
+fi
+
+echo -e "${WARN}Ignition: warning at $.systemd.units.0.contents: unit \"echo.service\" is enabled, but has no install section so enable does nothing${RESET}" > $expectedwarningsfile
+
+if ! diff $expectedwarningsfile $warningsfile; then
+    fatal "Ignition warning did not show up as expected in issue.d"
+fi
+ok "Successfully displayed Ignition warning on the console"


sohankunkerkar:
@dustymabe 
re. https://github.com/coreos/fedora-coreos-config/pull/1672#issuecomment-1098493711
I can use this test to cover the duplicated Ignition warnings

tests/kola/ignition/warnings/config.ign
```diff
@@ -0,0 +1,14 @@
...
+  "systemd": {
+    "units": [
+      {
+        "contents": "[Unit]\nDescription=echo service template\n[Service]\nType=oneshot\nExecStart=/bin/echo %i\n",
+        "enabled": true,
+        "name": "echo.service"
+      }
+    ]
+  }
+}
```
dustymabe: It's not immediately obvious to me why this Ignition config should generate a warning. Can you explain?

sohankunkerkar: Ignition will throw a [warning](https://github.com/coreos/ignition/blob/56eccf5c05c4c5dd7c06446d038acb1fcbe3f474/config/shared/validations/unit.go#L26) if the given systemd unit doesn't have an `Install` section.

dustymabe: ok cool. I have two questions now.

1. is there a technical reason why we couldn't convert this into a butane config? I suspect maybe the answer is yes so we can pass through the problematic config, which is why we're using Ignition directly here.
2. if we could use butane, could we convert it to that and embed comments about what is problematic about the config and maybe the warning output as an example?

tests/kola/ignition/warnings/test.sh
```diff
@@ -0,0 +1,19 @@
...
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+warningsfile="/etc/issue.d/30_coreos_ignition_warnings.issue"
+
+# Check for the motd file
+if ! test -f ${warningsfile}; then
+    fatal "not found Ignition warnings issue file"
+fi
+if ! grep -e 'unit "echo.service" is enabled' ${warningsfile} ; then
```
dustymabe: Just reading the code the text `unit "echo.service" is enabled` doesn't really seem much like a warning to me. What is the full text? Should we use the full text to make it more clear?

sohankunkerkar: >What is the full text? 

`warning at $.systemd.units.0.contents: unit "echo@.service" is enabled, but has no install section so enable does nothing`
I'm fine with using the entire text

tests/kola/ignition/warnings/config.ign
sohankunkerkar: > is there a technical reason why we couldn't convert this into a butane config? I suspect maybe the answer is yes so we can pass through the problematic config, which is why we're using Ignition directly here.

Yeah, this is possible. Earlier, I had a different example in mind while writing this Ignition config.

dustymabe: Thanks!

sohankunkerkar: > > is there a technical reason why we couldn't convert this into a butane config? I suspect maybe the answer is yes so we can pass through the problematic config, which is why we're using Ignition directly here.

> Yeah, this is possible. Earlier, I had a different example in mind while writing this Ignition config.

I think it's not possible. Butane treats any config warning as fatal and fails validation.
```
$ cosa kola run -p qemu --build latest 'ext.config.ignition.warnings'
COREOS_ASSEMBLER_CONFIG_GIT=/srv/fedora-coreos/src/github.com/coreos/fedora-coreos-config
+ podman run --rm -ti --security-opt label=disable --privileged --uidmap=1000:0:1 --uidmap=0:1:1000 --uidmap 1001:1001:64536 -v /home/skunkerk/fedora-coreos/tempfcos:/srv/ --device /dev/kvm --device /dev/fuse --tmpfs /tmp -v /var/tmp:/var/tmp --name cosa -v /srv/fedora-coreos/src/github.com/coreos/fedora-coreos-config:/srv/src/config/:ro quay.io/coreos-assembler/coreos-assembler:latest kola run -p qemu --build latest ext.config.ignition.warnings
kola --build latest --output-dir tmp/kola run -p qemu ext.config.ignition.warnings
2022-04-18T19:42:21Z platform/conf: warnings parsing config: warning at $.systemd.units.0.contents, line 11 col 17: unit "echo.service" is enabled, but has no install section so enable does nothing
Error: Parsing config.ign: configured to treate config warnings as fatal
2022-04-18T19:42:21Z cli: Parsing config.ign: configured to treate config warnings as fatal
+ rc=1
+ set +x

```

dustymabe: I think maybe if we set `allowConfigWarnings` to `true` then we should be OK. See 60c082c.

Do you mind trying?

tests/kola/ignition/warnings/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# kola: { "platforms": "qemu-unpriv", "allowConfigWarnings": true }
+# This test ensures that Ignition warnings 
+# are displayed on the console.
```
dustymabe: mind added some explanations here of the kola json options above similar to what is done here: https://github.com/coreos/fedora-coreos-config/blob/8425ff7cafcd82547e3273b279d215b5b74e4026/tests/kola/root-reprovision/swap-before-root/test.sh#L4-L16

tests/kola/ignition/warnings/test.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Check for the motd file
+if ! test -f ${warningsfile}; then
+    fatal "not found Ignition warnings issue file"
+fi
+# Check for the warning and ensure that duplicate content
+# gets filtered out properly.
+if grep -F 'Ignition: warning at $.systemd.units.0.contents: unit "echo.service" is enabled, but has no install section so enable does nothing' ${warningsfile} | wc -l > 1 ; then
+    fatal "either failed to display Ignition warning or seen duplicate Ignition warning"
+fi
```
dustymabe: Thinking about this a little bit more it might be better if we just did a exact comparison of the file contents we expect to the file contents in `${warningsfile}`. For example the line we are grepping for is complete (I asked for this), but in the [duplicate lines case](https://github.com/coreos/fedora-coreos-config/pull/1672#issue-1203467420) the lines were slightly different.

We should be able to do both checks at the same time and it an easier to comprehend way by comparing exactly what we expect to the actual file. 


How about something along the lines of:

```
WARN='\033[0;33m' # yellow
RESET='\033[0m' # reset

expectedwarningsfile=/tmp/expectedwarningsfile
echo -e "${WARN}Ignition: warning at $.systemd.units.0.contents: unit \"echo.service\" is enabled, but has no install section so enable does nothing${RESET}" > $expectedwarningsfile

if ! diff $expectedwarningsfile $warningsfile; then
    fatal "Ignition warning did not show up as expected in issue.d"
fi
```

dustymabe:
Since we're not really testing Ignition here I wonder if it would be more appropriate to put this test under a new `clhm` subheading. i.e. `tests/kola/clhm/ignition-warnings` or so.

Thoughts?

sohankunkerkar:
> Since we're not really testing Ignition here I wonder if it would be more appropriate to put this test under a new `clhm` subheading. i.e. `tests/kola/clhm/ignition-warnings` or so.
> 
> Thoughts?

Yeah, it makes sense to me

dustymabe (approved):
LGTM

PR #1676, opened 2022-04-14
tests/memory: check kernel messages for oom-killer

mike-nguyen:
Verify that oom-killer does not show up in dmesg.

https://bugzilla.redhat.com/show_bug.cgi?id=1931467

### DIFF
--- a/tests/kola/memory/data/commonlib.sh
+++ b/tests/kola/memory/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh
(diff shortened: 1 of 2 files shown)


bgilbert:
It might be better to add a [console check](https://github.com/coreos/coreos-assembler/blob/35c71e35b87089c298de34ea944483692e36614f/mantle/kola/harness.go#L112-L183) to kola so we'll notice if _any_ test triggers the OOM killer.

dustymabe:
> It might be better to add a [console check](https://github.com/coreos/coreos-assembler/blob/35c71e35b87089c298de34ea944483692e36614f/mantle/kola/harness.go#L112-L183) to kola so we'll notice if _any_ test triggers the OOM killer.

👍

mike-nguyen:
Closing in favor of https://github.com/coreos/coreos-assembler/pull/2818

PR #1681, opened 2022-04-19
Re-enable kdump.crash for aarch64

gursewak1997:
Re-enabling kdump.crash for aarch64 for testing-devel.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1075#issuecomment-1098060867

The test looks to be passing for `37.20220419.91.0` , `35.20220418.20.0` and `36.20220418.10.0` .


dustymabe (approved):
LGTM

PR #1684, opened 2022-04-21
05core: report `ENOSPC` in initrd when entering e-shell

### LINKED ISSUE #1055: Large and growing PXE RAM requirement
We document [here](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/#_installing_from_pxe) and [here](https://docs.fedoraproject.org/en-US/fedora-coreos/live-booting-ipxe/#_prerequisite) that PXE booting requires 2 GiB RAM for `coreos.live.rootfs_url` and 3 GiB for bootloader-fetched initrd+rootfs.  In recent tests, 2 GiB is still sufficient for `rootfs_url` but 3 GiB is no longer sufficient for bootloader-fetched rootfs.  (I'll call the latter case "concatenated rootfs", since that's how it appears to the kernel, whether the user manually concatenates the files or whether they're specified as separate initrds and concatenated by the bootloader.)

One issue is that the root filesystem in the initramfs is a tmpfs, and tmpfs instances are limited to 50% of RAM by default.  I don't think that's the only problem, though, since in the concatenated case, 1500 MiB should be sufficient even for the two memory-resident copies of the initramfs + rootfs.  It would be useful to fully track down the cause.

Some options for improving things:

1. Increase the tmpfs quota.  In the `rootfs_url` case, it might help to `mount -o remount,size=90% /` before extracting.  But in the concatenated case, we need to increase the quota before the kernel unpacks the initrd.  `rootflags=` _looks_ useful for this, but isn't; the argument [is ignored](https://github.com/torvalds/linux/blob/7a29b11da9651ef6a970e2f6bfd276f053aeb06a/fs/namespace.c#L4282) when mounting the tmpfs.  ([This patch](https://lore.kernel.org/all/20210630014744.88735-1-chenweilong@huawei.com/) should fix it, but apparently hasn't gotten any traction.)

   An alternative is to use `rootfstype=ramfs` to switch the rootfs to the old, simpler ramfs, which doesn't support quotas and which can successfully unpack in 2500 MiB RAM, but I'm not sure what side effects that might have.  If that's viable, we could recommend that karg in the concatenated case.

1. Discourage the concatenated case.  It takes nearly twice as much RAM during initialization, it's slower (because the firmware's network stack fetches the rootfs blob), and it seems to be less commonly used.  If we pursue this, we should probably [allowlist](https://github.com/coreos/fedora-coreos-config/blob/0eda5eb17adeff88a4ee60c3884c50c0d6a96a10/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh#L26) TFTP for `rootfs_url` so users don't need an HTTP server.

1. Error out if concatenated initrd unpack fails.  We report obscure failures in this case because we see a partially unpacked initramfs, but we could actually detect this case: `/root.squashfs` will exist, but some of the later files in the rootfs cpio won't.  The last member is currently an osmet file, but we could add a flag file or move `/etc/coreos-live-rootfs` to the end of the archive.

bgilbert:
When a user tries to boot a live system with an appended rootfs without enough RAM, the kernel will unpack the initramfs and rootfs until it runs out of memory, and then continue booting with a partial OS image.  This causes random confusing boot errors, such as a service failing because its binary is missing, or the rootfs squashfs failing to mount with some error that looks like filesystem corruption.

Make this failure less confusing by having `timeout.sh` try to write a small file, and if that fails, tell the user that they need more RAM.  Suppress reporting of unit failures in this case, since the unit logs will just muddy the issue.

This works reliably when the rootfs image can't be unpacked, i.e. when the system has a moderate amount of RAM.  It doesn't work reliably when we're so low on RAM that the initramfs image can't be unpacked (because the compressed initramfs and rootfs images have used up our RAM); in that case we might be missing important binaries/libraries and boot can fail at any point.  It would be possible to construct a more robust detection mechanism, e.g. by running a statically-linked program that's unpacked first.  But this can't be completely robust (`/sbin/init` might not have been unpacked, so we might not boot at all) and doesn't seem worth it for now.

This code will also trigger if we run out of RAM in the `coreos.live.rootfs_url` case, which is nice but less important, since that
case produces a clear error during rootfs fetch.

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1055.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/module-setup.sh
@@ -15,7 +15,8 @@ install_unit_wants() {
 install() {
     inst_multiple \
         cut \
-        date
+        date \
+        dd
 
     inst_hook emergency 99 "${moddir}/timeout.sh"
 
--- a/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/timeout.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/timeout.sh
@@ -53,11 +53,29 @@ downloaded from GitHub:
 
 EOF
         fi
-        echo "Displaying logs from failed units: ${failed}"
-        for unit in ${failed}; do
-            # 10 lines should be enough for everyone
-            journalctl -b --no-pager --no-hostname -u ${unit} -n 10
-        done
+
+        # If this is a live boot, check for ENOSPC in initramfs filesystem
+        # Try creating a 64 KiB file, in case a small file was deleted on
+        # service failure
+        # https://github.com/coreos/fedora-coreos-tracker/issues/1055
+        if [ -f /etc/coreos-live-initramfs ] && \
+            ! dd if=/dev/zero of=/tmp/check-space bs=4K count=16 2>/dev/null; then
+            cat <<EOF
+------
+Ran out of memory when unpacking initrd filesystem.  Ensure your system has
+at least 2 GiB RAM if booting with coreos.live.rootfs_url, or 4 GiB otherwise.
+------
+
+EOF
+            # Don't show logs from failed units, since they'll just be
+            # random misleading errors.
+        else
+            echo "Displaying logs from failed units: ${failed}"
+            for unit in ${failed}; do
+                # 10 lines should be enough for everyone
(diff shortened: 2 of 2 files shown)


dustymabe (approved):
LGTM

jlebon:
> 05user

Did you mean 05core here?

bgilbert:
Argh, yes.

PR #1686, opened 2022-04-22
Revert "overlay: add new module for installing Azure udev rules"

dustymabe:
The WALinuxAgent-udev rpm now installs the udev rules into the
initramfs [1] so we can drop this glue code.

This reverts commit 8c07b7391473910ba3884ee0d3763743805ac78f.

[1] https://src.fedoraproject.org/rpms/WALinuxAgent/c/521b67bc8575f53a30b4b2c4e63292e67483a4e1?branch=rawhide


jlebon:
Will this break RHCOS? See https://github.com/openshift/os/blob/master/overlay.d/README.md#25rhcos-azure-udev-rules.

dustymabe:
> Will this break RHCOS? See https://github.com/openshift/os/blob/master/overlay.d/README.md#25rhcos-azure-udev-rules.

Looks like the patch is in el9 but not el8.

https://github.com/openshift/os/pull/785

jlebon (approved):

bgilbert:
> Will this break RHCOS?

It did: https://bugzilla.redhat.com/show_bug.cgi?id=2092966

PR #1688, opened 2022-04-22
overrides: fast-track downgraded packages

gursewak1997:
Fast-track the packages when they are older in `next-devel` than in `testing-devel`


dustymabe:
We talked and `tzdata` needs to be built for F36. What about `libsolv`?

dustymabe (approved):

PR #1691, opened 2022-04-25
overrides: fast-track downgraded package

gursewak1997:
Fast-track the downgraded package as it was older in `next-devel` than in `testing-devel`

Follow up on https://github.com/coreos/fedora-coreos-config/pull/1688.


dustymabe (dismissed):
LGTM

dustymabe (approved):
LGTM

PR #1692, opened 2022-04-26
Disable kdump.crash test for aws

gursewak1997:
Disable kdump.crash for aws platform due to CI [failure](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/kola-aws/251/) on `kola-aws`

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -8,6 +8,12 @@
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1075
   arches:
+    - aarch64
+  platforms:
+    - aws
+- pattern: ext.config.kdump.crash
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1075
+  arches:
     - ppc64le
 - pattern: coreos.boot-mirror.luks
   tracker: https://github.com/coreos/coreos-assembler/issues/2725


kola-denylist.yaml
```diff
@@ -9,6 +9,7 @@
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1075
   arches:
     - ppc64le
+    - s390x
```
dustymabe: I wouldn't touch the s390x bits as part of this change.

tests/kola/kdump/crash
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# kola: {"minMemory": 4096, "tags": "skip-base-checks", "architectures": "!s390x"}
+# kola: {"minMemory": 4096, "tags": "skip-base-checks", "platforms": "!aws"}
```
dustymabe: I would put this in the denylist too. I think maybe we need a brand new denylist entry for it:

```
  tracker: https://new.bug.may.need.to.be.opened/
  arches:
    - aarch64
  platforms:
    - aws
```

I think that should work but it will need to be tested.

dustymabe: this goal is to allow for running the test on aws x86_64 (and any other platform), just not aws aarch64.

saqibali-2k: > I think that should work but it will need to be tested.

Yep, this should work.

dustymabe (approved):

gursewak1997:
This entry doesn't disable `kdump.crash` for `x86_64` on aws

PR #1695, opened 2022-04-26
overrides: fast-track selinux-policy-36.7-1.fc36

dustymabe:
Let's match what has been requested for stable but hasn't made it into
the latest compose yet. The compose had trouble.

See https://pagure.io/releng/issue/10670#comment-794029

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -147,6 +147,18 @@ packages:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-b547780983
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1101#issuecomment-1085987923
       type: fast-track
+  selinux-policy:
+    evra: 36.7-1.fc36.noarch
+    metadata:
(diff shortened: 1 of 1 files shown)


jlebon (approved):

manifest-lock.overrides.yaml
```diff
@@ -147,6 +147,18 @@ packages:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-b547780983
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1101#issuecomment-1085987923
       type: fast-track
+  selinux-policy:
+    evra: 36.7-1.fc36.noarch
```
gursewak1997, line 151: Is `.noarch` required?
Just wondering because I don't see it on any other entries in this file

gursewak1997: Ah, nvm. I see the difference.

gursewak1997 (approved):

manifest-lock.overrides.yaml
dustymabe: Either way. Thanks for asking! That's how we learn.

PR #1698, opened 2022-04-27
image.yaml: Switch to `ostree-format: "oci-chunked"`

cgwalters:
OK, it's time to start trying this out in "production".

https://bodhi.fedoraproject.org/updates/FEDORA-2022-4813f1daaf
is queued (and already in coreos-assembler).

I've pushed an updated demo of this to `quay.io/cgwalters/fcos-chunked`
which you can try today.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -3,5 +3,5 @@
 # streams.
 include: image-base.yaml
 
-# https://github.com/coreos/coreos-assembler/pull/2216
-ostree-format: "oci"
+# Plan to move this to all streams soon
+ostree-format: "oci-chunked"


cgwalters:
OK and to consume this on the host we do want https://bodhi.fedoraproject.org/updates/FEDORA-2022-4813f1daaf in our streams, but hopefully that will happen soon.

dustymabe:
A few questions:

- can we try this on `next-devel`/`next` first?
- can we wait until Fedora 36 is released and we've transitioned `testing` to Fedora 36 content safely?

cgwalters:
> can we try this on next-devel/next first?

Sure, done

dustymabe:
Thanks!

Marking this as draft so it doesn't get merged by accident. We'll merge (assuming no objections) once we switch our `testing-devel` and `testing` streams over to F36.

dustymabe (dismissed):

cgwalters:
OK a few other things to keep in mind here is that this conceptually can only really break two things:

- How we build the disk image - but that's tested in CI!
- Anyone pulling the container directly - but that's not a production path yet, and if it *did* break something I'd want to know *and* we can easily revert

cgwalters:
Hmm actually anything blocking this going to rawhide first?

dustymabe:
> Hmm actually anything blocking this going to rawhide first?

Nope. That would work 👍

cgwalters:
OK, done!

cgwalters:
(I discovered when rebasing that the rawhide `image.yaml` still had a now-defunct override for `ostree-format: oci` which was the legacy of the *last* attempt to do OCI stuff in rawhide first)

dustymabe (approved):
LGTM

cgwalters:
OK this is getting us a chunked image in S3, but the prod copy is being re-squashed back to a single layer because https://github.com/coreos/coreos-assembler/blob/c29b6f7d526090395f5248588eead17d2d36e464/src/cmd-sign#L185

We need to teach that code to read `image.yaml` too...which really I think wants the cleanups from https://github.com/coreos/coreos-assembler/pull/2839

cgwalters:
Eh this is going to get messy to implement here or in cosa, let's do https://github.com/ostreedev/ostree-rs-ext/issues/295

cgwalters:
OK I should have done this before, but here's a tracker issue for this work: https://github.com/coreos/fedora-coreos-tracker/issues/1226

PR #1699, opened 2022-04-27
overrides: fast-track rust-coreos-installer-0.14.0-1.fc35

coreosbot-releng:


jlebon:
With any luck, this one will actually be moot if we GA next week. But not if it slips.

bgilbert (approved):

dustymabe (approved):

PR #1702, opened 2022-04-28
tests: check for avc denials around daemon-reload

mike-nguyen:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1924869

### DIFF
--- a/tests/kola/systemd/systemd-daemon-reload-avc
+++ b/tests/kola/systemd/systemd-daemon-reload-avc
@@ -1,0 +1,10 @@
+#!/bin/bash
+# kola: { "exclusive": true }
+# Verify there are no AVC denials and check for avc denials after reloading the systemd manager configuration
+#
+# https://bugzilla.redhat.com/show_bug.cgi?id=1924869
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+systemctl daemon-reload


tests/kola/systemd/systemd-daemon-reload-avc
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# kola: { "exclusive": true }
+# Verify there are no AVC denial before and after reloading the systemd manager configuration
+#
+# https://bugzilla.redhat.com/show_bug.cgi?id=1924869
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+if ausearch -m avc | grep 'avc:  denied'; then
```
dustymabe: FCOS doesn't include `audit` - see https://github.com/coreos/fedora-coreos-tracker/issues/461

dustymabe:
Maybe instead of doing this we should pursue https://github.com/coreos/coreos-assembler/pull/2067

mike-nguyen:
There's good discussion on that PR about whether it makes sense to have it run on every test since some tests may purposely invoke errors.  I'm inclined to keep this for now unless there are objections.

dustymabe:
> There's good discussion on that PR about whether it makes sense to have it run on every test since some tests may purposely invoke errors. I'm inclined to keep this for now unless there are objections.

I think that's fair especially since that PR hasn't merged.

dustymabe (dismissed):

dustymabe:
Note that this test essentially becomes a catchall for any SELinux issue that happens by default on boot. I think that's fine, but the test name and description are probably more narrowly focused that they should be because I suspect it will fail for other reasons a lot more than it will fail for the reason this test was written.

Can you try running this agains a local build of `next-devel` and `rawhide` to make sure it passes before we merge?

mike-nguyen:
> Note that this test essentially becomes a catchall for any SELinux issue that happens by default on boot. I think that's fine, but the test name and description are probably more narrowly focused that they should be because I suspect it will fail for other reasons a lot more than it will fail for the reason this test was written.
> 
> Can you try running this agains a local build of `next-devel` and `rawhide` to make sure it passes before we merge?

Great point! My intention was to make sure there were no denials before the test and that `systemd daemon-reload` did not make cause any denials.  I'll update the test description.  I think I will also move the function into the common library so any test can use it if they choose to.

cgwalters:
> Note that this test essentially becomes a catchall for any SELinux issue that happens by default on boot. I think that's fine, but the test name and description are probably more narrowly focused that they should be because I suspect it will fail for other reasons a lot more than it will fail for the reason this test was written.

I *strongly* agree with this.  

> There's good discussion on that PR about whether it makes sense to have it run on every test since some tests may purposely invoke errors. 

Yeah, but I think what we can do on the coreos-assembler side is an an opt-in that lets us ignore errors.  Perhaps by logging a special journal message.

There is actually another approach - we could ship a systemd unit that trips into failure mode when a SELinux denial is detected.

The key advantage of this is: it means our existing failure detection for systemd unit Just Works - including things like the console login.

The OpenShift MCO uses `systemctl --failed` but doesn't know about SELinux either, etc.

We could actually discuss taking such a unit to some selinux-policy package.

I wouldn't *object* to merging this PR as is but the above has a lot more appeal to me.

bgilbert:
> Yeah, but I think what we can do on the coreos-assembler side is an an opt-in that lets us ignore errors. Perhaps by logging a special journal message.

We have the existing test flags mechanism for this.

PR #1704, opened 2022-04-29
overrides: fast-track systemd-249.12-3.fc35

dustymabe:
```
commit c30d541a420fe9447a83265f6e7742edc990e62d
Date:   Fri Apr 29 14:13:37 2022 -0400

    Revert "15fcos: add coreos-fix-systemd-journald.service"
    
    This is no longer needed because the fix from upstream [1]
    has made it into systemd-249.12-3.fc35. See [2]
    
    This reverts commit a456f26a3f6fa4e77ce92a4cb83ea77c9f1b7ee8.
    
    [1] https://github.com/systemd/systemd/pull/22861
    [2] https://github.com/coreos/fedora-coreos-tracker/issues/1108

commit 4aea3446ec4e982c0ce4043b2ea6e43f02fff9d7
Date:   Fri Apr 29 14:11:59 2022 -0400

    overrides: fast-track systemd-249.12-3.fc35
    
    This fixes the journald issues we've been seeing because it includes
    the upstream fix from https://github.com/systemd/systemd/pull/22861
    
    Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1108

```


jlebon (approved):
:tada: 

I presume the fix is already in the f36 systemd in next-devel?

dustymabe:
> I presume the fix is already in the f36 systemd in next-devel?

No. It looks like it's only in v250.5 where `next-devel` has `systemd-250.3-8.fc36`. However we never saw this problem in Fedora 36 (it's very much a race condition) so it should be fine that the code isn't in `next-devel` right now.

PR #1707, opened 2022-04-30
manifests: add curl to file-transfer manifest

dustymabe:
This is used in coreos-livepxe-rootfs.sh [1], although I'm sure there
are probably more places too.

[1] https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh

### DIFF
--- a/manifests/file-transfer.yaml
+++ b/manifests/file-transfer.yaml
@@ -3,3 +3,6 @@ packages:
   - git-core
   - gnupg2
   - rsync
+  # Explicit dependency on curl because we use it in coreos-livepxe-rootfs.sh
+  # We need curl and not curl-minimal because we support TFTP.
+  - curl


dustymabe:
I thought about reorganizing all of these manifests while I was in here. Basically we need to put the package specifications into two files and nowhere else IMO.

- `fedora-coreos-packages.yaml` - for fcos only packages
- `shared-packages.yaml` - for ones common to RHCOS

Some other manifests can stay split out because they are used for conditional includes. But other than for that purpose we should try to collapse these as much as possible and use comments in the yaml to achieve the same goal.

bgilbert (approved):

cgwalters (approved):

jlebon (approved):

PR #1448, opened 2022-01-26
kdump-ignition test added to verify kdump works via ignition config

aaradhak:

PR #1708, opened 2022-05-02
updates for `ext.config.kdump.crash` test

dustymabe:
```
commit 0be55468cac2a70f3f943909202579765a2bef67
Date:   Wed May 4 14:50:06 2022 -0400

    manifests/fedora-coreos: switch grub2-removals to conditional-include
    
    This move the `arch-include` statments to `conditional-include`, which
    allows us to use `!=`. In this case what we were really trying to target
    was s390x IIUC.

commit 03156146da01544ba434360e91a0042a2879078b
Date:   Sun May 1 23:01:42 2022 -0400

    manifests/fedora-coreos: fix kdump.crash test for aarch64 AWS
    
    The irqpoll` karg that gets added when the crash kernel is
    started causes issues on AWS aarch64 instances. Let's workaround
    by removing that from `KDUMP_COMMANDLINE_APPEND` in
    `/etc/sysconfig/kdump` for now.
    
    See https://github.com/coreos/fedora-coreos-tracker/issues/1187

commit ccb5447a1afd454ec7d143fec187960f588fadb4
Date:   Tue May 3 10:24:06 2022 -0400

    tests: kdump.crash: update crashkernel size based on our docs
    
    In our documentation [1] we recommend setting it to 300M as a
    starting baseline. Let's update from 256M to 300M here.
    
    [1] https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/

commit d69f0256f667e2351d747cd9523511138bf8f6db
Date:   Sun May 1 23:13:29 2022 -0400

    tests: kdump.crash: extend timeout
    
    The kdump.crash test on the rawhide stream for aarch64 has started
    taking just longer than 10 minutes. Let's extend the timeout so those
    tests don't timeout and fail the entire run.
    
    Also added comments for the test's kola.json arguments.

commit 7eeb6b738ef3e48b0fdbc0243da1e7f805ca28e6
Date:   Sun May 1 22:55:45 2022 -0400

    make kdump work on first boot
    
    Up until this point people have had to reboot once after setting up
    kdump so that if a crash happens the crash kernel doesn't get started
    with the ignition.firstboot kernel argument. If we add `ignition.firstboot`
    to the list of arguments in `KDUMP_COMMANDLINE_REMOVE=` in
    `/etc/sysconfig/kdump` then we can set it up on first boot without issues.
    
    Here we also switch the kdump.crash test to set up kdump via Ignition
    to prove out the change works as desired.

```

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -17,6 +17,9 @@ conditional-include:
   - if: prod == false
     # long-term, would be good to support specifying a nested TreeComposeConfig
     include: disable-zincati-and-pinger.yaml
+  - if: basearch == "aarch64"
+    # Fixup for kdump on aarch64 AWS instances
+    include: kdump-aarch64-aws-workaround.yaml
 
 ostree-layers:
   - overlay/15fcos
@@ -85,6 +88,7 @@ postprocess:
     cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
     cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
     rm /tmp/rpm-ostreed.conf
+
 
 remove-from-packages:
   # Drop NetworkManager support for ifcfg files, see also corresponding
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -20,6 +20,9 @@ conditional-include:
   - if: basearch == "aarch64"
     # Fixup for kdump on aarch64 AWS instances
     include: kdump-aarch64-aws-workaround.yaml
+  - if: basearch != "s390x"
+    # And remove some cruft from grub2
+    include: grub2-removals.yaml
 
 ostree-layers:
   - overlay/15fcos
@@ -122,9 +125,3 @@ exclude-packages:
   # For (datacenter/cloud oriented) servers, we want to see the details by default.
   # https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HSMISZ3ETWQ4ETVLWZQJ55ARZT27AAV3/
   - plymouth
-
-# And remove some cruft from grub2
-arch-include:
-  x86_64: grub2-removals.yaml
-  aarch64: grub2-removals.yaml
-  ppc64le: grub2-removals.yaml
--- a/manifests/system-configuration.yaml
+++ b/manifests/system-configuration.yaml
@@ -37,3 +37,14 @@ packages:
   # Boost starving threads
   # https://github.com/coreos/fedora-coreos-tracker/issues/753
   - stalld
+
+postprocess:
+  # Make kdump work on firstboot
+  - |
+    #!/usr/bin/env bash
+    # Make kdump ignore `ignition.firstboot` when copying kargs from
+    # the running kernel to the kdump kernel when passing to be kexec.
+    # This makes it so kdump can be set up on the very first boot.
+    # Upstream request to have this upstream so we can stop carrying it here:
+    # https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/thread/5P4WIJLW2TSGF4PZGRZGOXYML4RXZU23/
+    sed -i -e 's/KDUMP_COMMANDLINE_REMOVE="/KDUMP_COMMANDLINE_REMOVE="ignition.firstboot /' /etc/sysconfig/kdump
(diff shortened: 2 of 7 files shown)


manifests/system-configuration.yaml
```diff
@@ -37,3 +37,12 @@ packages:
...
   - stalld
+
+postprocess:
+  # Make kdump work on firstboot
+  - |
+    #!/usr/bin/env bash
+    # Make kdump ignore `ignition.firstboot` when copying kargs from
+    # the running kernel to the kdump kernel when passing to be kexec.
+    # This makes it so kdump can be set up on the very first boot.
+    sed -i -e 's/KDUMP_COMMANDLINE_REMOVE="/KDUMP_COMMANDLINE_REMOVE="ignition.firstboot /' /etc/sysconfig/kdump
```
dustymabe, line 50: wondering if we should make a PR upstream for this. Wouldn't hurt other platforms to have `ignition.firstboot` mentioned in this list.

manifests/system-configuration.yaml
```diff
@@ -37,3 +37,12 @@ packages:
   # Boost starving threads
   # https://github.com/coreos/fedora-coreos-tracker/issues/753
   - stalld
+
+postprocess:
```
dustymabe, line 41: added here so we can share with RHCOS.

manifests/fedora-coreos.yaml
```diff
@@ -85,6 +85,12 @@ postprocess:
     cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
     cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
     rm /tmp/rpm-ostreed.conf
+  # Fixups for kdump
+  - |
+    #!/usr/bin/env bash
+    # Remove irqpoll from the list of KDUMP_COMMANDLINE_APPEND. This
+    # causes issues on aarch64 AWS instances.
+    sed -i -e 's/irqpoll //' /etc/sysconfig/kdump
```
miabbott: Should this have a condition to only run on `aarch64`? What's the implication of having this run on all arches?

tests/kola/kdump/crash/config.bu
```diff
@@ -0,0 +1,9 @@
+variant: fcos
+version: 1.4.0
+kernel_arguments:
+  should_exist:
+    - crashkernel=256M
```
mike-nguyen: Lets make this `300M` to match our docs https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/

mike-nguyen (changes requested):
I ran this locally and the symlink to `commonlib.sh` is broken. It's a relative symlink and the test is one level deeper.

manifests/fedora-coreos.yaml
bgilbert: I believe removing this option could cause the crash kernel not to boot on systems where interrupts get into a bad state.

dustymabe: Yeah @miabbott - I'm not entirely sure. @bgilbert is it your opinion that we shouldn't do this at the risk of breaking other platforms/hardware?

tests/kola/kdump/crash/config.bu
dustymabe: thanks, done

mike-nguyen:
These issues can be closed once this merges

mike-nguyen:
Related PRs/Issues that can be closed if this merges

- https://github.com/coreos/fedora-coreos-tracker/issues/1079
- https://github.com/coreos/fedora-coreos-config/pull/1448

manifests/system-configuration.yaml
dustymabe: I started an upstream [mailing list thread](https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/thread/5P4WIJLW2TSGF4PZGRZGOXYML4RXZU23/) to discuss adding `ignition.firstboot` to the `KDUMP_COMMANDLINE_REMOVE` list.

manifests/fedora-coreos.yaml
bgilbert: I'm okay with doing it for all of `aarch64` for now, since our installed base there is relatively small, but I'd avoid making the change on `x86_64`.

dustymabe: this should now only apply to `aarch64`.

mike-nguyen (approved):
LGTM

dustymabe:
CI flaked on https://github.com/coreos/fedora-coreos-tracker/issues/1190

PR #1710, opened 2022-05-03
ci/continuous: New container that pulls from continuous COPR

cgwalters:
History goes in cycles.  One of the key original raison d'être of
ostree was having people consume git main/master of projects.  But...
because RPM doesn't make it easy to revert things, we really
end up going only at the same speed as the rest of Fedora sadly.

Anyways, we've talked about this a few times in FCOS; xref
https://github.com/coreos/fedora-coreos-tracker/issues/910

This adds a container image that is built *as* a container layering
on top of the base image, using the recently landed
https://github.com/coreos/rpm-ostree/pull/3605/

My plan is to add build automation (in the pipeline? or OCP Prow?)
that sticks this at e.g. `quay.io/coreos-assembler/fcos:testing-devel-continous`.

### DIFF
--- a/ci/continuous/Dockerfile
+++ b/ci/continuous/Dockerfile
@@ -1,0 +1,4 @@
+FROM quay.io/coreos-assembler/fcos:testing-devel
+ADD fcos-continuous.repo /etc/yum.repos.d
+ADD overrides.yaml /etc/rpm-ostree/origin.d/overrides.yaml
+RUN rpm-ostree ex rebuild
--- a/ci/continuous/overrides.yaml
+++ b/ci/continuous/overrides.yaml
@@ -1,0 +1,11 @@
+ex-override-replace:
+  - from:
+      repo: copr:copr.fedorainfracloud.org:group_CoreOS:continuous
+    packages:
+      - ostree
+      - ostree-libs
+      - rpm-ostree
+      - rpm-ostree-libs
+      - ignition
+      - coreos-installer
+      - coreos-installer-bootinfra
--- a/ci/continuous/fcos-continuous.repo
+++ b/ci/continuous/fcos-continuous.repo
@@ -1,0 +1,10 @@
+[copr:copr.fedorainfracloud.org:group_CoreOS:continuous]
+name=Copr repo for continuous owned by @CoreOS
+baseurl=https://download.copr.fedorainfracloud.org/results/@CoreOS/continuous/fedora-$releasever-$basearch/
+type=rpm-md
+skip_if_unavailable=True
+gpgcheck=1
+gpgkey=https://download.copr.fedorainfracloud.org/results/@CoreOS/continuous/pubkey.gpg
+repo_gpgcheck=0
+enabled=1
+enabled_metadata=1


ci/continuous/overrides.yaml
```diff
@@ -0,0 +1,11 @@
+ex-override-replace:
+  - from:
+      repo: copr:copr.fedorainfracloud.org:group_CoreOS:continuous
+    packages:
```
cgwalters, line 4: We should add `packages-all: true` or so to auto-accept *everything* in the repo since that matches how people actually use most COPRs.

dustymabe:
> Anyways, we've talked about this a few times in FCOS; xref [coreos/fedora-coreos-tracker#910](https://github.com/coreos/fedora-coreos-tracker/issues/910)

It seems like we [agreed](https://github.com/coreos/fedora-coreos-tracker/issues/910#issuecomment-1049947637) there that we didn't feel adding the continuous stream was worth the extra effort. Has something changed since then?

> 
> This adds a container image that is built _as_ a container layering on top of the base image, using the recently landed [coreos/rpm-ostree#3605](https://github.com/coreos/rpm-ostree/pull/3605)
> 
> My plan is to add build automation (in the pipeline? or OCP Prow?) that sticks this at e.g. `quay.io/coreos-assembler/fcos:testing-devel-continous`.

Who would consume that image? What happens when automation around it or testing breaks? Who would look at the test failures?

cgwalters:
> It seems like we https://github.com/coreos/fedora-coreos-tracker/issues/910#issuecomment-1049947637 there that we didn't feel adding the continuous stream was worth the extra effort. Has something changed since then?

That was for a full kola-coreos-assembler-s3 stuff.  This is just a container image build, which is *much* less heavyweight.

> Who would consume that image? What happens when automation around it or testing breaks? Who would look at the test failures?

I would, main to test out our code from git main.

I do plan to add tests, but I think "run all the kola tests, fail if anything fail" is a general anti-pattern we are going to need to move to avoid longer term.

But anyways, this is just a container build, not a "stream" in the current sense.

dustymabe:
> > It seems like we [coreos/fedora-coreos-tracker#910 (comment)](https://github.com/coreos/fedora-coreos-tracker/issues/910#issuecomment-1049947637) there that we didn't feel adding the continuous stream was worth the extra effort. Has something changed since then?
> 
> That was for a full kola-coreos-assembler-s3 stuff. This is just a container image build, which is _much_ less heavyweight.

That context isn't in the comment [you posted](https://github.com/coreos/fedora-coreos-tracker/issues/910#issuecomment-1116031471). Might be good to add that?

> 
> > Who would consume that image? What happens when automation around it or testing breaks? Who would look at the test failures?
> 
> I would, main to test out our code from git main.

Any other consumers? I think what you've got here is fine (dropping some files in the FCOS config repo), but if we're not going to have many other people using it maybe not worth adding automation around it to auto build/push the container to quay. i.e. a local container build might suffice?

That could be a conversation for a different thread since this is a PR against the config repo.

cgwalters:
> I think what you've got here is fine (dropping some files in the FCOS config repo), but if we're not going to have many other people using it maybe not worth adding automation around it to auto build/push the container to quay. i.e. a local container build might suffice?

I don't see building a container as a heavyweight thing, really the point is to have this readily accessible and available.

Another way to say it is: the UX for "I want to test something in rpm-ostree git main" switches from "pull RPMs" to "pull a container" - which is a big deal.

ci/continuous/overrides.yaml
jlebon: This comment is related: https://github.com/coreos/rpm-ostree/blob/da4aa26f5e11bbc19d04adf6b76321c4eb6f56fd/rust/src/treefile.rs#L2452-L2453.

So e.g. we could have

```yaml
from:
  copr: @CoreOS/continuous
```

which knows how to convert that to a yumrepo and by default if no packages are provided, just takes everything applicable.

jlebon (approved):
I'm cool with this overall. If the goal is to make it easier for devs to play with the latest, then this is useful (that was my overall intention with https://github.com/coreos/fedora-coreos-tracker/issues/910). (As opposed to having similar test coverage to what we currently have for all streams).

Note this does not help with testing Ignition at least (related: https://github.com/coreos/fedora-coreos-tracker/issues/1151).

If we do this in https://github.com/coreos/fedora-coreos-pipeline, I wouldn't add it to the main pipeline jobs. I could see it as a separate job that builds it and ideally lightly tests it before pushing. Or otherwise a BuildConfig triggering on pushes to `:testing-devel` (that seems to be closer to the model we're heading towards for OCP CoreOS layering).

dustymabe (approved):

ci/continuous/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM quay.io/coreos-assembler/fcos:testing-devel
```
dustymabe, line 1:
```suggestion
FROM quay.io/coreos-assembler/fcos:testing-devel
```

I assume config bot is going to copy these files to the other branches. We might want to think about how this line translates to the the other branches.

cgwalters:
OK let's try this!

cgwalters:
Build job in https://github.com/openshift/release/pull/28831

PR #1712, opened 2022-05-04
tests/kola/ignition: port resource fetching tests from mantle; move to new bucket; test `s3://` with versioned resources

bgilbert:
Reimplement tests from `mantle/kola/tests/ignition/resource.go`, using a bucket in a non-deprecated AWS account:

- `coreos.ignition.resource.remote`
- `coreos.ignition.resource.s3`
- `coreos.ignition.resource.s3.versioned` (merged into `.remote` test, since it's only separate for historical reasons)

Also test fetching versioned resources via `s3://`.

### DIFF
--- a/tests/kola/ignition/resource/s3/config.bu
+++ b/tests/kola/ignition/resource/s3/config.bu
@@ -1,0 +1,14 @@
+# Objects accessible by any authenticated S3 user, such as the IAM role
+# associated with the instance
+
+variant: fcos
+version: 1.0.0
+ignition:
+  config:
+    merge:
+      - source: "s3://rh-kola-fixtures/resources/authenticated-var-v3.ign"
+storage:
+  files:
+    - path: /var/resource/s3-auth
+      contents:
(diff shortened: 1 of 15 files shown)


tests/kola/ignition/resource/s3/config.bu
```diff
@@ -0,0 +1,14 @@
+# Objects accessible by any authenticated S3 user, such as the IAM role
+# associated with the instance
```
dustymabe, line 2: dumb question.. In this external test how do we make sure the instance gets started with the appropriate role such that it can access the authenticated resources?

dustymabe (approved):
LGTM

tests/kola/ignition/resource/s3/config.bu
bgilbert: The test fixtures are configured to allow access by any authenticated user, and kola ensures that the instance always has _some_ role, which is enough.

PR #1715, opened 2022-05-06
clhm.ignition-warnings: Improve test to ignore additional warnings

jschintag:


bgilbert:
If it's possible, I think there's value in continuing to check that the message is surrounded by the correct ANSI control characters.

jschintag:
> If it's possible, I think there's value in continuing to check that the message is surrounded by the correct ANSI control characters.

@bgilbert That should do it, but i would like to state that regex is always a pain.

bgilbert (approved):
Thanks!

PR #1717, opened 2022-05-06
tests/kola/ignition: add S3 ARN tests; test intra-cloud anonymous S3/GCS access

bgilbert:
Test the new ARN support in Ignition 2.14.0.  S3 access points can't be accessed anonymously, so even the public objects need to be tested from inside an EC2 instance.  Also test unauthenticated S3->EC2 and GCS->GCE fetches; the latter fails on Ignition < 2.14.0.

### DIFF
--- a/tests/kola/ignition/resource/authenticated-s3/config.bu
+++ b/tests/kola/ignition/resource/authenticated-s3/config.bu
@@ -9,6 +9,10 @@ ignition:
       - source: "s3://ignition-test-fixtures/resources/authenticated-var-v3.ign"
 storage:
   files:
+    # Check that anonymous access works with credentials
+    - path: /var/resource/s3-anon
+      contents:
+        source: "s3://ignition-test-fixtures/resources/anonymous"
     - path: /var/resource/s3-auth
       contents:
         source: "s3://ignition-test-fixtures/resources/authenticated"
--- a/tests/kola/ignition/resource/authenticated-s3/config.bu
+++ b/tests/kola/ignition/resource/authenticated-s3/config.bu
@@ -2,7 +2,7 @@
 # associated with the instance
 
 variant: fcos
-version: 1.0.0
+version: 1.5.0-experimental
 ignition:
   config:
     merge:
@@ -12,3 +12,20 @@ storage:
     - path: /var/resource/s3-auth
       contents:
         source: "s3://ignition-test-fixtures/resources/authenticated"
+    - path: /var/resource/arn-auth
+      contents:
+        source: "arn:aws:s3:::ignition-test-fixtures/resources/authenticated"
+    # Publicly-readable object, fetched via an access point.  Access points
+    # don't allow anonymous access.
+    - path: /var/resource/arn-ap-anon
+      contents:
+        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/ignition-test-fixtures-ap/object/resources/anonymous"
+    - path: /var/resource/arn-ap-auth
+      contents:
+        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/ignition-test-fixtures-ap/object/resources/authenticated"
+    - path: /var/resource/arn-ap-versioned-original
+      contents:
+        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/ignition-test-fixtures-ap/object/resources/versioned?versionId=Y9YqVujoLyHHSHJ4DslyXoaLvcilQJnU"
+    - path: /var/resource/arn-ap-versioned-latest
+      contents:
+        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/ignition-test-fixtures-ap/object/resources/versioned"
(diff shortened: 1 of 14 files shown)


dustymabe:
With these additions it feels a little like `tests/kola/ignition/resource/s3` is misnamed. At first when looking I couldn't understand the real difference between the two tests.

The difference is that one of them is limited to just AWS (i.e. for testing authenticated access). Should we rename the test `aws-authenticated-s3` (or anything else that's more explicit) to help explain? Or maybe at least add a description to the test and explanations for kola json values like:

https://github.com/coreos/fedora-coreos-config/blob/0856e21a48d389409f55eb2da8fa6ca9715a0991/tests/kola/root-reprovision/linear/test.sh#L4-L13

bgilbert:
Yes, you're right, we should.

bgilbert:
Updated!

bgilbert:
Split out the two prep commits into #1723.

tests/kola/ignition/resource/authenticated-s3/config.bu
```diff
@@ -12,3 +12,18 @@ storage:
     - path: /var/resource/s3-auth
       contents:
         source: "s3://ignition-test-fixtures/resources/authenticated"
+    - path: /var/resource/arn-auth
+      contents:
+        source: "arn:aws:s3:::ignition-test-fixtures/resources/authenticated"
+    - path: /var/resource/arn-ap-anon
+      contents:
+        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/ignition-test-fixtures-ap/object/resources/anonymous"
```
dustymabe, line 26: looking at the table in https://github.com/coreos/ignition/pull/1358 access point's don't support anonymous access, but this is authenticated access, just to an object with key `resources/anonymous` ?

dustymabe (dismissed):
LGTM

dustymabe (dismissed):

tests/kola/ignition/resource/authenticated-s3/config.bu
```diff
@@ -2,13 +2,46 @@
...
+#        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/myhsd4oog7zwk.mrap/object/resources/anonymous"
+#    - path: /var/resource/arn-mrap-auth
+#      contents:
+#        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/myhsd4oog7zwk.mrap/object/resources/authenticated"
+#    - path: /var/resource/arn-mrap-versioned-original
+#      contents:
+#        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/myhsd4oog7zwk.mrap/object/resources/versioned?versionId=Y9YqVujoLyHHSHJ4DslyXoaLvcilQJnU"
+#    - path: /var/resource/arn-mrap-versioned-latest
+#      contents:
+#        source: "arn:aws:s3:us-east-1:460538899914:accesspoint/myhsd4oog7zwk.mrap/object/resources/versioned"
```
dustymabe: Should we go ahead and create the files we want to do the diff against too? Feels like we could just put this in a separate PR to merge later.

tests/kola/ignition/resource/remote/test.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
-# kola: { "tags": "needs-internet" }
+# kola: { "tags": "needs-internet", "noInstanceCreds": true }
 # - tags: needs-internet
 #   - We fetch resources from S3.
```
dustymabe: minor

```suggestion
#   - We fetch resources from S3 and GCS.
```

tests/kola/ignition/resource/authenticated-s3/config.bu
dustymabe: This one is still confusing to me. Did you answer it somewhere and I missed it?

dustymabe (dismissed):
LGTM. A few more questions, though.

tests/kola/ignition/resource/authenticated-s3/config.bu
bgilbert: Whoops, sorry, I missed this.  The object in the underlying bucket is available for public read (and we read it anonymously in other tests), but we perform authenticated access to it here because that's required when using access points.

dustymabe: Since this is yaml maybe we could get a comment with that context? Or we could create a new object more appropriately named `resources/authenticated-public-read`?

bgilbert: "Authenticated public read" isn't a thing.  There's a public object and an authenticated object and we're testing them both.

tests/kola/ignition/resource/authenticated-s3/config.bu
bgilbert: Sure, moved to https://github.com/coreos/fedora-coreos-config/pull/1745.

bgilbert:
Updated!

dustymabe (approved):
LGTM

PR #1721, opened 2022-05-06
manifests: drop libvarlink conditional includes

### LINKED ISSUE #1130: Drop `libvarlink-util` from package set
This was [originally added](https://github.com/coreos/fedora-coreos-tracker/issues/231) for `podman-remote` functionality but the podman team ripped out varlink code in v3. We should be safe to remove it but there may be some users using it for other things. I feel like if there was ever an opportunity for us to actually remove a package, this would be it.

- [the current definition](https://github.com/coreos/fedora-coreos-config/blob/224be5d072125445633dfd509b9d6f28ba6fda5d/manifests/fedora-coreos-base.yaml#L124-L125)

dustymabe:
Now that we have waited an amount of time let's drop libvarlink-util
altogether.

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1130

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -18,13 +18,6 @@ ostree-layers:
   - overlay/20platform-chrony
 
 conditional-include:
-  # https://github.com/coreos/fedora-coreos-tracker/issues/1130
-  - if: stream == "testing-devel"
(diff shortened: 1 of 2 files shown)


ravanelli (approved):
LGTM

PR #1723, opened 2022-05-07
tests/kola/ignition: rename `resource.s3` to `.authenticated-s3`; document kola options

bgilbert:
Make it clearer that the purpose of the test is to check authenticated S3 access.


dustymabe (dismissed):
LGTM

dustymabe (approved):
still LGTM

PR #1729, opened 2022-05-09
tests/kola/chrony: grep metadata server IP on GCP

jlebon:
The output of `chronyc sources` on GCP contains the resolved IP address,
even though the config we write uses the hostname. GCP documents[1] the
IP address of the metadata server as 169.254.169.254 so we can just
hardcode it here like we do for AWS.

[1] https://cloud.google.com/compute/docs/internal-dns


dustymabe (approved):
LGTM

bgilbert (approved):

PR #1730, opened 2022-05-09
denylist: skip non-exclusive-test-bucket-0 for next release

ravanelli:
- It is alredy fixed for testing-devel:
https://github.com/coreos/fedora-coreos-config/pull/1729

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -17,3 +17,8 @@
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:
   - ppc64le
+- pattern: non-exclusive-test-bucket-0
+  # Already fixed in https://github.com/coreos/fedora-coreos-config/pull/1729
+  snooze: 2022-05-11
+  streams:
+    - next


kola-denylist.yaml
```diff
@@ -17,3 +17,8 @@
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:
   - ppc64le
+- pattern: non-exclusive-test-bucket-0
```
gursewak1997, line 20:
```suggestion
- pattern: ext.config.chrony.coreos-platform-chrony-generator
```
`non-exclusive-test-bucket` is just a wrapperName for various test buckets for all the non-exclusive tests:)

dustymabe:
correct. @gursewak1997 is right, but we don't really need this because the only cloud provider test that failed was on GCP and it was only the `ext.config.chrony.coreos-platform-chrony-generator` test, correct?

If that's the case we can just ignore that failure and go on. We know the fixed test is coming. We don't need to merge this PR or run any more tests.

For the streams issue we don't have to check that check box. We'll just note in a comment that the test failed for this reason and we can ignore it.

PR #1734, opened 2022-05-11
udev: don't create 'coreos-boot-disk' symlink on subsequent boots

nikita-dubrovskii:


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -2,6 +2,11 @@
 # checks whether `disk` contains filesystem labeled `label`
 set -euo pipefail
 
+cmdline=$(</proc/cmdline)
+if [[ ! "$cmdline" =~ "ignition.firstboot" ]]; then
```
jlebon: This is OK, but feels like we should solve it at a higher level. Hmm, maybe something like:
- install rules in `/usr/lib/coreos/` instead of `/usr/lib/udev/rules.d`
- in `coreos-diskful-generator`, add a symlink to it in `/run/udev/rules.d` (to be checked whether a symlink works or if we have to copy)

That way, the rule itself isn't even active on non-firstboot. Hmm, we could put it [here](https://github.com/coreos/fedora-coreos-config/blob/3b52493f2af02017137856445d61a8ec0fe2e6d8/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator#L59) which would also ensure the symlink only exists on diskful boots.

nikita-dubrovskii: hmm, i saw `coreos-diskful-generator` , but didn't want to put anything into it, cause that hides the file(s) and makes logic a bit complicated imho.

jlebon: I don't think it's out of scope for `coreos-diskful-generator` given that it only makes sense on diskful boots. We can add a comment in the rule file that mentions that it's only used during firstboot and cross-reference `coreos-diskful-generator`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -58,3 +58,8 @@ Requires=coreos-gpt-setup.service
 After=coreos-gpt-setup.service
 EOF
 fi
+
+# create symlink fir udev rule
+mkdir -p /run/udev/rules.d/
+ln -s /usr/lib/coreos/80-coreos-boot-disk.rules \
+      /run/udev/rules.d/80-coreos-boot-disk.rules
```
jlebon: Let's put this within the `if ! is-live-image; then` block instead so that it only takes effect on diskful boots.

jlebon (approved):

jlebon:
Thanks!

PR #1738, opened 2022-05-13
05core: add `coreos-ignition-delete-config.service` for upgrade boots

bgilbert:
Ignition 2.14.0 adds `ignition-delete-config.service`, which deletes Ignition configs from VMware and VirtualBox on first boot.  Add `coreos-ignition-delete-config.service` to do the same thing on existing machines on upgrade, using a stamp file in `/var/lib` to avoid multiple runs.

Add a drop-in for `ignition-delete-config.service` that creates a stamp file in `/run`, and then chain from that stamp to the long-term stamp in `/var/lib`, ensuring that we don't delete configs twice on newly-provisioned machines.  The upstream service can't create the stamp directly in `/var/lib` because it runs before `/var` is mounted.

Prevent `coreos-ignition-delete-config.service` from running if `ignition-delete-config.service` is masked, ensuring that the mask operation [documented upstream](https://coreos.github.io/ignition/operator-notes/#automatic-config-deletion) prevents the config from ever being deleted, as intended.

We can remove this after the next barrier release in FCOS and barrier equivalent in RHCOS.

The plan is to do a coreos-status post documenting `systemctl mask ignition-delete-config.service` before shipping this.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/coreos-ignition-delete-config.service
+++ b/overlay.d/05core/usr/lib/systemd/system/coreos-ignition-delete-config.service
@@ -1,0 +1,28 @@
+# Can be removed from FCOS in Fedora 37 or after the next barrier release,
+# whichever comes first.  Can be removed from RHCOS in the first release
+# after every node is guaranteed to have booted at least once with 4.11 or
+# higher.
+
+[Unit]
+Description=CoreOS Delete Ignition Config From Hypervisor
+Documentation=https://coreos.github.io/ignition/
+
+ConditionKernelCommandLine=|ignition.platform.id=virtualbox
+ConditionKernelCommandLine=|ignition.platform.id=vmware
+ConditionPathExists=!/var/lib/coreos-ignition-delete-config.stamp
+# Hack: if the user masked ignition-delete-config.service, we shouldn't run
+# either.
+ConditionPathIsSymbolicLink=!/etc/systemd/system/ignition-delete-config.service
+
+# We check a stamp file written by ignition-delete-config.service.  That
+# service runs Before=sysinit.target, on which we have a default dependency,
+# so this is really just documentation.
+After=ignition-delete-config.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-ignition-delete-config
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/05core/usr/lib/systemd/system/ignition-delete-config.service.d/10-flag-file.conf
+++ b/overlay.d/05core/usr/lib/systemd/system/ignition-delete-config.service.d/10-flag-file.conf
@@ -1,0 +1,7 @@
+# Create a flag file to notify coreos-ignition-delete-config.service that
+# we've run, and put it in /run because /var isn't mounted yet.
+# coreos-ignition-delete-config.service will then avoid trying to delete
+# the config again, and will create a persistent stamp file in /var/lib.
+
+[Service]
+ExecStart=/bin/touch /run/coreos-ignition-delete-config.stamp
--- a/overlay.d/05core/usr/libexec/coreos-ignition-delete-config
+++ b/overlay.d/05core/usr/libexec/coreos-ignition-delete-config
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cmdline=( $(</proc/cmdline) )
+cmdline_arg() {
+    local name="$1" value
+    for arg in "${cmdline[@]}"; do
+        if [[ "${arg%%=*}" == "${name}" ]]; then
+            value="${arg#*=}"
+        fi
+    done
+    echo "${value}"
+}
+
+# Avoid running again if ignition-delete-config.service has run, but still
+# create our own stamp file now that /var is mounted.
+if [ ! -e /run/coreos-ignition-delete-config.stamp ]; then
+    PLATFORM_ID=$(cmdline_arg ignition.platform.id)
+    /usr/libexec/ignition-rmcfg --platform=${PLATFORM_ID}
+fi
+
+touch /var/lib/coreos-ignition-delete-config.stamp
--- a/tests/kola/ignition/delete-config/config.bu
+++ b/tests/kola/ignition/delete-config/config.bu
@@ -1,0 +1,36 @@
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/fake-ignition-rmcfg
+      mode: 0755
+      contents:
+        inline: |
+          #!/bin/bash
+          # Mocked replacement for ignition-rmcfg that just records that
+          # we ran it, and fails if we run it twice.  We can't run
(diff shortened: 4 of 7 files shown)


overlay.d/05core/usr/lib/systemd/system/coreos-ignition-delete-config.service
```diff
@@ -0,0 +1,23 @@
...
+[Unit]
+Description=CoreOS Delete Ignition Config From Hypervisor
+Documentation=https://coreos.github.io/ignition/
+
+ConditionKernelCommandLine=|ignition.platform.id=virtualbox
+ConditionKernelCommandLine=|ignition.platform.id=vmware
+ConditionPathExists=!/var/lib/coreos-ignition-delete-config.stamp
+# Hack: if the user masked ignition-delete-config.service, we shouldn't run
+# either.
+ConditionPathIsSymbolicLink=!/etc/systemd/system/ignition-delete-config.service
```
jlebon, line 15: Minor/optional: I know it's redundant, but it'd be nice to add a `After=ignition-delete-config.service` here to "document" it since the approach here depends on it.

tests/kola/ignition/delete-config/test.sh
```diff
@@ -0,0 +1,88 @@
...
+
+case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+"")
+    # Ignition boot
+
+    if [ ! -e /run/ignition-rmcfg-ran ]; then
+        fatal "mocked ignition-rmcfg did not run on first boot"
+    fi
+
+    if [ $(systemctl is-active ignition-delete-config ||:) != active ]; then
```
jlebon, line 19: Minor:

```suggestion
    if ! systemctl is-active ignition-delete-config; then
```

And below.

jlebon (dismissed):
Nice work!
Some optional nits; LGTM as is.

tests/kola/ignition/delete-config/test.sh
bgilbert: Yeah, I know, but it seems more readable with the symmetry.

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-delete-config.service
bgilbert: Heh, I originally had that and then removed it as redundant.  Re-added.

tests/kola/ignition/delete-config/config.bu
```diff
@@ -0,0 +1,32 @@
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/fake-ignition-rmcfg
+      mode: 0755
+      contents:
+        inline: |
+          #!/bin/bash
```
dustymabe, line 9: Might be nice to add a line in here documenting why this is useful (i.e. the fakeout/mocking).

overlay.d/05core/usr/lib/systemd/system/ignition-delete-config.service.d/10-flag-file.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStart=/bin/touch /run/coreos-ignition-delete-config.stamp
```
dustymabe, line 7: So this stampfile that gets created is specific to Fedora CoreOS? And even more so, is it only useful/checked in `coreos-ignition-delete-config`? At some point we'll remove `coreos-ignition-delete-config`, right? Would we remove this at the same time?

Could we get more comments in here with context and whatnot?

overlay.d/05core/usr/libexec/coreos-ignition-delete-config
```diff
@@ -0,0 +1,23 @@
...
+}
+
+# Avoid running again if ignition-delete-config.service has run, but still
+# create our own stamp file now that /var is mounted.
+if [ ! -e /run/coreos-ignition-delete-config.stamp ]; then
+    PLATFORM_ID=$(cmdline_arg ignition.platform.id)
+    /usr/libexec/ignition-rmcfg --platform=${PLATFORM_ID}
+fi
+
+touch /var/lib/coreos-ignition-delete-config.stamp
```
dustymabe, line 23: Why can't we just create `/var/lib/coreos-ignition-delete-config.stamp` via `ignition-delete-config.service.d/10-flag-file.conf` and avoid this dance?

tests/kola/ignition/delete-config/test.sh
```diff
@@ -0,0 +1,88 @@
...
+        fatal "ignition-delete-config didn't succeed on first boot"
+    fi
+    if [ $(systemctl is-active coreos-ignition-delete-config ||:) != active ]; then
+        fatal "coreos-ignition-delete-config didn't succeed on first boot"
```
dustymabe, line 23: do we really want them both running?

dustymabe (commented):
A few comments/questions.

overlay.d/05core/usr/libexec/coreos-ignition-delete-config
bgilbert: That was the first thing I tried, but `/var` isn't mounted yet.

tests/kola/ignition/delete-config/test.sh
bgilbert: Yeah, we need the second service to convert the flag file in `/run` into a flag file in `/var/lib`.

overlay.d/05core/usr/lib/systemd/system/ignition-delete-config.service.d/10-flag-file.conf
bgilbert: The first three are correct, yes.  I don't see a point in trying to clean up the stamp file; we'd need to ship a new service for that when removing this one.  It doesn't really hurt anything.

bgilbert:
Added requested comments.

overlay.d/05core/usr/lib/systemd/system/ignition-delete-config.service.d/10-flag-file.conf
dustymabe: Sorry. When I said "Would we remove this at the same time?" I was referring to `10-flag-file.conf`.

dustymabe: IOW - should we add to the comment and say "remove this when `coreos-ignition-delete-config.service` no longer exists"?

dustymabe: I guess it's kind of implied with the existing comment you added.

dustymabe (approved):
LGTM

overlay.d/05core/usr/lib/systemd/system/ignition-delete-config.service.d/10-flag-file.conf
bgilbert: Yeah, I expect that we'll revert the entire commit at that time.

PR #1739, opened 2022-05-13
workflows: fix override-handling jobs with current Git

bgilbert:
When running Git in a containerized job, we need to tell it that repos not owned by the current user are safe.

Sample failing jobs: [add-override](https://github.com/coreos/fedora-coreos-config/actions/runs/2309622296), [remove-graduated-overrides](https://github.com/coreos/fedora-coreos-config/actions/runs/2317724942).


dustymabe (approved):
LGTM

PR #1742, opened 2022-05-14
tests/kdump.crash: add in sleep to workaround XFS race condition

dustymabe:
This seems to make the test no longer fail. We still need to investigate
the root cause. See https://github.com/coreos/fedora-coreos-tracker/issues/1195

Also add in a check to make sure the kdump.service comes up OK.

### DIFF
--- a/tests/kola/kdump/crash/test.sh
+++ b/tests/kola/kdump/crash/test.sh
@@ -23,6 +23,10 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
           fatal "kdump.service failed to start"
       fi
       /tmp/autopkgtest-reboot-prepare aftercrash
+      # Add in a sleep to workaround race condition where XFS/kernel errors happen
+      # during crash kernel boot.
+      # https://github.com/coreos/fedora-coreos-tracker/issues/1195
+      sleep 5
       echo "Triggering sysrq"
       sync
       echo 1 > /proc/sys/kernel/sysrq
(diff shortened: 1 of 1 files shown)


tests/kola/kdump/crash/test.sh
```diff
@@ -19,6 +19,9 @@ set -xeuo pipefail
 
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
+      if [ $(systemctl show -p Result kdump.service) != "Result=success" ]; then
```
bgilbert, line 22: This seems to show `Result=success` even for nonexistent units, but maybe we don't care here.

bgilbert (approved):
This isn't ideal, but I can live with it.  How did you find it?

tests/kola/kdump/crash/test.sh
dustymabe: hmm. yeah I didn't realize that. I lifted this from another test so I didn't give it too much scrutiny.

dustymabe:
> This isn't ideal, but I can live with it. How did you find it?

I was under the assumption it was a race condition because it made it into the `testing-devel` stream at all, which means it passed tests at least a few times. I also had a false positive in my local testing, which further points to race condition.

tests/kola/kdump/crash/test.sh
travier: It's "safe" in the other test as the checks that happen before fail if the unit does not exist. But I agree that this is not ideal.

PR #1741, opened 2022-05-14
ci/overrides: srpms: report only SRPMs relevant to lockfiles

bgilbert:
The `srpms` subcommand was reporting all SRPMs attached to the specified Bodhi update, even if some of those didn't have any binary RPMs included in our lockfiles.  This was causing `add-override.yml` to put spurious SRPM names in its commit titles.  Report only SRPMs that are relevant to us.

PR #1747, opened 2022-05-18
[rawhide] overrides: fast-track ignition-2.14.0-1.fc37

coreosbot-releng:


dustymabe:
@bgilbert - looks like https://github.com/coreos/fedora-coreos-config/pull/1741 needs a fixup?

jlebon:
I opened https://github.com/coreos/fedora-coreos-config/pull/1748 which should fix this, but I couldn't test it because Bodhi is currently down.

PR #1748, opened 2022-05-18
ci/overrides.py: use stderr for logging

jlebon:
Some commands are designed to output parseable data on stdout. Let's
consistently use stderr for informational output.


bgilbert (approved):
Whoops, good catch!

PR #1757, opened 2022-05-24
tests/kdump.crash: bump crashkernel size, add check for future

dustymabe:
It turns out kdump has a tool that can be used to estimate the
appropriate size for the crashkernel reserved memory known as
`kdumpctl estimate`. On FCOS/RHCOS here are the numbers I saw
for the architectures in our universe:

```
- 161M aarch64 FCOS
- 448M aarch64 RHCOS

- 134M ppc64le FCOS
- 384M ppc64le RHCOS

- 192M s390x FCOS
- 160M s390x RHCOS

- 157M x86_64 FCOS
- 165M x86_64 RHCOS
```

Bump the crashkernel here to 500M to handle our worst case and also
add a check to make sure that if we ever are under the minimum
recommended size we'll get notified.

### DIFF
--- a/tests/kola/kdump/crash/config.bu
+++ b/tests/kola/kdump/crash/config.bu
@@ -2,7 +2,10 @@ variant: fcos
 version: 1.4.0
 kernel_arguments:
   should_exist:
-    - crashkernel=300M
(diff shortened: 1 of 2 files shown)


bgilbert (approved):
:+1:

PR #1758, opened 2022-05-25
lockfiles: add ppc64le and s390x lockfiles

dustymabe:
Prep for adding these architectures in the future.

### DIFF
--- a/manifest-lock.ppc64le.json
+++ b/manifest-lock.ppc64le.json
@@ -1,0 +1,1251 @@
+{
+  "packages": {
+    "NetworkManager": {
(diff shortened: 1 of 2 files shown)


jlebon:
Should we do this one at a time, starting with the one we're closest to enabling first?

dustymabe:
I don't think so. These files are harmless in the repo, but I think they need to exist before we can start the bumpfile locker.

Are you hesitant to add one that might get a little out of date?

dustymabe:
Closing this in favor of code that will auto add new lockfiles for new arches when they get added. https://github.com/coreos/fedora-coreos-pipeline/pull/529

PR #1759, opened 2022-05-25
ci/overrides: don't load sack if overrides empty

jlebon:
Right now, `next-devel` and `branched` are inactive. But to reduce
churn, the `remove-graduated-overrides` GitHub Action is left running on
those branches since it's harmless. However, the repos on inactive
branches might not be valid, causing workflow failures.

But anyway, there are no active overrides on these branches, so we can
work around this by exiting earlier.

I think a more appropriate fix for this is to query whether `next-devel`
is open or closed (though we'd need something similar for `branched`
too), but we do also want this optimization anyway.

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -277,9 +277,6 @@ def setup_repos(base, treefile):
         base.repos[repo].enable()
         eprint(f"- {repo}")
 
-    eprint("Downloading metadata")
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

bgilbert (approved):
LGTM

PR #1764, opened 2022-05-30
tests: fix trust-cpu test

### LINKED ISSUE #1209: rawhide: ext.config.files.trust-cpu test is failing due to new kernel
We have found the `ext.config.files.trust-cpu` to be failing due to the new
kernel update (`kernel-5.19.0-0.rc0.20220525gitfdaf9a5840ac.2.fc37`). [Build](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/build/692/)

From further investigation, it looks like the line 
`pr_notice("crng init done (trusting CPU's manufacturer)\n");` was removed
in this [commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/char/random.c?id=e85c0fc1d94c52483a603651748d4c76d6aa1c6b). We will need to change how we validate if trust cpu is on in that test.

saqibali-2k:
As mentioned in https://github.com/coreos/fedora-coreos-tracker/issues/1209,
due to upstream changes, our current mechanism for testing
if RANDOM_TRUST_CPU is enabled is no longer valid. Lets instead
check the kernel config file to get the same information.

fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1209

### DIFF
--- a/tests/kola/files/trust-cpu
+++ b/tests/kola/files/trust-cpu
@@ -6,16 +6,12 @@ set -xeuo pipefail
 . $KOLA_EXT_DATA/commonlib.sh
 
 # https://bugzilla.redhat.com/show_bug.cgi?id=1830280
-tmpd=$(mktemp -d)
-cd "${tmpd}"
 case "$(arch)" in
     x86_64)
-        dmesg | grep ' random:' > random.txt
-        if ! grep -qe 'crng.*done.*trust.*CPU' <random.txt; then
-            sed -e 's/^/# /' < random.txt
+        config=$(grep ^CONFIG_RANDOM_TRUST_CPU /lib/modules/$(uname -r)/config)
+        if [ "$config" != "CONFIG_RANDOM_TRUST_CPU=y" ]; then
             fatal "Error: Failed to find crng trusting CPU"
         fi
         ok "random trust cpu" ;;
     *) echo "Don't know how to test hardware RNG state on arch=$(arch)" ;;
 esac
-rm -r "${tmpd}"
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -33,7 +33,3 @@
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1206
   arches:
   - s390x
-- pattern: ext.config.files.trust-cpu
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1209
-  streams:
-  - rawhide


cgwalters (dismissed):
We could probably just remove the test now too, it's extremely unlikely that the config change here gets reverted.

Alternatively, we could generalize the test into "kernel configuration options we expect to be set".

dustymabe:
> Alternatively, we could generalize the test into "kernel configuration options we expect to be set".

Agree.. That could be useful. Maybe we can open an issue for that and mark it as a good first issue.

tests/kola/files/trust-cpu
```diff
@@ -6,16 +6,13 @@ set -xeuo pipefail
...
 # https://bugzilla.redhat.com/show_bug.cgi?id=1830280
-tmpd=$(mktemp -d)
-cd "${tmpd}"
 case "$(arch)" in
     x86_64)
-        dmesg | grep ' random:' > random.txt
-        if ! grep -qe 'crng.*done.*trust.*CPU' <random.txt; then
-            sed -e 's/^/# /' < random.txt
+        kernel=$(rpm -q kernel | tail -c +8)
+        enabled=$(grep CONFIG_RANDOM_TRUST_CPU /lib/modules/"${kernel}"/config | tail -c 2)
```
cgwalters: BTW also, we can significantly strengthen this check by prefixing the regexp with `^` - non-enabled options have a leading `#` and so then the regexp won't see them.

tests/kola/files/trust-cpu
```diff
@@ -6,16 +6,12 @@ set -xeuo pipefail
...
 # https://bugzilla.redhat.com/show_bug.cgi?id=1830280
-tmpd=$(mktemp -d)
-cd "${tmpd}"
 case "$(arch)" in
     x86_64)
-        dmesg | grep ' random:' > random.txt
-        if ! grep -qe 'crng.*done.*trust.*CPU' <random.txt; then
-            sed -e 's/^/# /' < random.txt
+        config=$(grep ^CONFIG_RANDOM_TRUST_CPU /lib/modules/$(uname -r)/config)
+        if ![ "$config" = "CONFIG_RANDOM_TRUST_CPU=y" ]; then
```
dustymabe:
```suggestion
        if [ "$config" != "CONFIG_RANDOM_TRUST_CPU=y" ]; then
```

I don't often see the negating `!` on the outside of the brackets. Is there a preference here for that over just doing the appropriate comparison inside?

dustymabe (approved):
LGTM

PR #1765, opened 2022-05-31
rawhide: Enable cliwrap

cgwalters:
The cliwrap code has existed for a long, long time.  However,
things change now with the advent of ostree-native containers.

I think it Just Makes Sense to enable `yum install foo` inside
an ostree container, which the code now does when this flag is enabled.

This will *also* cause the cliwrap code to appear on the host, but
I think that's less confusing than not doing so.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -8,3 +8,6 @@ repos:
   - fedora-rawhide
 
 include: manifests/fedora-coreos.yaml
+
+# Only enabled on rawhide right now
+cliwrap: true


cgwalters:
After this lands I plan to change some of https://github.com/coreos/coreos-layering-examples to use `yum install` instead.

dustymabe:
I'm still hesitant on this change. I know we've talked about this before so I understand the drive to make the change here. Obviously some of the hesitancy depends on the underlying implementation, which I need to look into. i.e. Does it warn the user it's doing some $magic?, etc.

Another part of my hesitancy here is gaining some sort of understanding with the team that develops the software we are hijacking. I think in the very least we should have a conversation with them about this change and try to make sure they aren't strongly opposed. Inevitably they are going to get bug reports and finding out after the fact that another stack hijacked your software won't be fun. 

EDIT: I shouldn't assume you haven't talked to them already about this. Maybe you have?

cgwalters:
Fair, I will set up something for that.

dustymabe:
> Fair, I will set up something for that.

Let me know if you'd like to coordinate.. We could invite them to a video version of our community meeting. Or we could do an ad-hoc session.

cgwalters:
> Obviously some of the hesitancy depends on the underlying implementation, which I need to look into. i.e. Does it warn the user it's doing some $magic?, etc.

For the record, it is really easy to try out now manually:
https://coreos.github.io/rpm-ostree/cliwrap/

But this is kind of the point of just doing this on rawhide, is exactly that it changes to just do it by default so people can try it *even more* easily and give feedback.

(That said, it's not clear to me that only having this on rawhide would succeed at that)

cgwalters:
(Though now that I look, I didn't make clear originally that this was targeting the rawhide branch only; fixed)

dustymabe:
> (That said, it's not clear to me that only having this on rawhide would succeed at that)

yeah. I don't know if just trying it out on rawhide would be successful at that goal. `rawhide` is a great place to prove out cosa changes or pipeline code, but not so much user facing stuff.

I do appreciate trying it on rawhide first just so we don't break other things, so I don't have a super strong opinion on blocking this based on the above requests. I just don't want to signal that because it went in `rawhide` it should be able to go into the other branches without additional scrutiny.

jlebon:
Also cool with turning this on for rawhide first as a way to sanity-check it doesn't break things at least. To do this though, it must be in `manifest.yaml` instead.

cgwalters:
> Also cool with turning this on for rawhide first as a way to sanity-check it doesn't break things at least. To do this though, it must be in manifest.yaml instead.

Done

jlebon (approved):

cgwalters:
kdump failure is unrelated, merging over red

cgwalters:
Oh I see, the kdump failure was not actually unrelated.

```
Jun  7 14:00:33.210115 kdumpctl[1482]: This system is rpm-ostree based; initramfs handling is                                                                                                
Jun  7 14:00:33.210115 kdumpctl[1482]: integrated with the underlying ostree transaction mechanism.                                                                                          
Jun  7 14:00:33.210115 kdumpctl[1482]: Use `rpm-ostree initramfs` to control client-side initramfs generation.                                                                               
Jun  7 14:00:33.212718 kdumpctl[1482]: rpm-ostree: Note: This system is image (rpm-ostree) based.                                                                                            
Jun  7 14:00:33.212718 kdumpctl[1482]: rpm-ostree: Dropping privileges as `dracut` was executed with not "known safe" arguments.                                                             
Jun  7 14:00:33.213000 kdumpctl[1482]: rpm-ostree: You may invoke the real `dracut` binary in `/usr/libexec/rpm-ostree/wrapped/dracut`.                                                      
Jun  7 14:00:33.213000 kdumpctl[1482]: rpm-ostree: Continuing execution in 5 seconds.
```

dustymabe:
This is the second time we've merged over a CI failure and it caused real issues. We should probably start listening more 😂

cgwalters:
Revert in https://github.com/coreos/fedora-coreos-config/pull/1773

dustymabe:
> This is the second time we've merged over a CI failure and it caused real issues. We should probably start listening more joy

Actually, in the other case it wasn't your fault at all. It was an external dependency that got updated and CI just happened to be the first one that warned us about it. It had nothing to do with your PR though.

dustymabe:
Also, thank you for putting this into rawhide first. It's a great way to shake out issues without affecting our more important dev pipelines.

PR #1766, opened 2022-06-02
denylist: deny `ext.config.ignition.resource.remote` on s390x

dustymabe:
We're seeing a weird failure there. Let's denylist it for now while
it is investigated.

See https://github.com/coreos/fedora-coreos-tracker/issues/1215


gursewak1997 (approved):
LGTM

PR #1769, opened 2022-06-02
05core: print warnings to secondary consoles

### LINKED ISSUE #954: console defaults for x86_64 qemu platform
With the work in https://github.com/coreos/coreos-assembler/pull/2400 we'll soon be able to configure console defaults on each platform individually. After a notice period, the x86_64 qemu images will revert to no `console=` entry on the kernel command line, which means it will default to VGA console by default, which is what some users expect and will welcome the change.

For me (I suspect I'm not alone, but maybe), I use a lot of scripting to run various tests and I typically use the qemu qcow to pull that off. In a lot of cases I use the serial console output to debug, go through scrollback, etc.. In my cases I typically don't offer a graphical console to the machine at all (i.e. `virt-install --graphics=none`).

It would be nice if either the kernel would choose the serial console when there was no VGA console OR we set console defaults for QEMU that would **fallback** to serial console if VGA **isn't available**. From preliminary testing setting `console=ttyS0 console=tty0` does that.

dustymabe:
This allows us to print the helpful messages that we have added
for our users before they enter dracut's emergency shell to all
configured consoles on the machine.

Closes https://github.com/coreos/fedora-coreos-tracker/issues/954

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-shell-setup/emergency-shell.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-shell-setup/emergency-shell.sh
@@ -64,13 +64,15 @@ EOF
             echo "Displaying logs from failed units: ${failed}"
             for unit in ${failed}; do
                 # 10 lines should be enough for everyone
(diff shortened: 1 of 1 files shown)


bgilbert (approved):
LGTM

PR #1774, opened 2022-06-08
workflows: add scheduled job to sync `openshift/os` submodule

### LINKED ISSUE #709: Set up Dependabot to bump fedora-coreos-config submodule regularly
We should automate these bumps rather than rely on humans doing it.

bgilbert:
We've been hand-maintaining the `fedora-coreos-config` submodule in `openshift/os`, but updating it is usually completely mechanical.  Add a workflow that maintains an `openshift/os` PR with the latest changes, updating it daily or on demand.  We can merge the PR whenever we'd like, and the job will open a new one once there are additional changes.

Omit CoreOS Bot changes from the log summary, since those involve lockfiles and overrides that don't apply to RHCOS.  Likewise, if the only changes are from CoreOS Bot, skip opening a PR.

When the workflow is manually invoked, allow overriding the PR branch.  This can be used to create a one-off PR that won't be maintained by the bot afterward, for cases where a sync needs complementary changes elsewhere in `openshift/os`.

Ideally the job would run inside the `openshift/os` repo, but GitHub Actions isn't available there.

Fixes https://github.com/openshift/os/issues/709.

### DIFF
--- a/.github/workflows/openshift-os.yml
+++ b/.github/workflows/openshift-os.yml
@@ -1,0 +1,83 @@
+name: Sync to openshift/os
+on:
+  # We could do push: branches: [testing-devel] but that would restart
+  # downstream CI a lot
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+    inputs:
+      branch:
+        # Allow overriding branch for syncs that need manual fixups
+        description: PR branch
+        required: true
+        default: fcc-sync
+
+permissions:
+  # none at all
+  contents: none
+
+jobs:
+  update-submodule:
+    name: Update fedora-coreos-config submodule
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+        with:
+          repository: openshift/os
+          # We need an unbroken commit chain when pushing to the fork.  Don't
+          # make assumptions about which commits are already available there.
+          fetch-depth: 0
+
+      - name: Update submodule
+        env:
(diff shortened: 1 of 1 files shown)


travier (dismissed):

.github/workflows/openshift-os.yml
```diff
@@ -0,0 +1,83 @@
...
+
+          git submodule init
+          git submodule update
+
+          cd fedora-coreos-config
+          # Omit CoreOS Bot commits from the log message, since they generally
+          # only affect FCOS
+          git log "HEAD..testing-devel" --perl-regexp \
+              --author='^((?!CoreOS Bot <coreosbot@fedoraproject.org>).*)$' |
+              git shortlog > $RUNNER_TEMP/shortlog
```
jlebon: Minor: why not just invoke `git shortlog` directly? The same switches are supported, and ranges too.

jlebon (dismissed):
Nice! One minor comment, but LGTM as is.

.github/workflows/openshift-os.yml
bgilbert: ...ah.  I learned `git shortlog` in the stone age, when it was only a postprocessor.  Updated.

jlebon (approved):

dustymabe:
> Ideally the job would run inside the `openshift/os` repo, but GitHub Actions isn't available there.

Indeed. I think in the future if our abilities change there we should definitely move this.

dustymabe (approved):
I have no objections here. Thank you for working on it!

bgilbert:
CI doesn't test this.  Merging.

PR #1777, opened 2022-06-09
tests: verify /etc release symlinks are valid

HuijingHei:
See:
- https://bugzilla.redhat.com/show_bug.cgi?id=2068148
- https://github.com/openshift/os/pull/815

### DIFF
--- a/tests/kola/files/check-symlink
+++ b/tests/kola/files/check-symlink
@@ -1,0 +1,19 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+# Verify /etc release symlinks are valid
+# See:
+# - https://bugzilla.redhat.com/show_bug.cgi?id=2068148
+# - https://github.com/openshift/os/pull/815
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+for file_name in /etc/system-release-cpe /etc/system-release /etc/redhat-release /etc/os-release
+do
+    real_path=$(readlink -f ${file_name})
+    if ! test -f "${real_path}"; then
+        fatal "Error: ${file_name} symlink to ${real_path} which not exists"
+    fi
+    ok "${file_name} symlink to valid file ${real_path}"
+done


ravanelli (dismissed):
lgtm

mike-nguyen:
LGTM

tests/kola/files/check-symlink
```diff
@@ -0,0 +1,19 @@
...
+# Verify the symlink pointing to a valid file
+# See:
+# - https://bugzilla.redhat.com/show_bug.cgi?id=2068148
+# - https://github.com/openshift/os/pull/815
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+for file_name in /etc/system-release-cpe /etc/system-release /etc/redhat-release
```
jlebon: While we're here, we might as well add `/etc/os-release` too.

jlebon (commented):
This mostly LGTM, but can I suggest a more descriptive commit title? E.g. maybe something like `tests: verify /etc release symlinks are valid`?

jlebon (approved):
Thanks!

travier:
I'm late to the party but is there any reason preventing us from testing basically all symlinks in the image?

HuijingHei:
> is there any reason preventing us from testing basically all symlinks in the image?

Good question, do we need to fix all broken symlinks if found?

Just check /etc and find there are 4 symlinks that are broken:
```
[core@cosa-devsh ~]$ sudo find /etc -xtype l | xargs ls -l
lrwxrwxrwx. 1 root root 29 Jun  9 17:03 /etc/alternatives/ncman -> /usr/share/man/man1/ncat.1.gz
lrwxrwxrwx. 1 root root 43 Jun  9 17:03 /etc/alternatives/unversioned-python-man -> /usr/share/man/man1/unversioned-python.1.gz
lrwxrwxrwx. 1 root root 31 Jun  9 17:03 /etc/grub2-efi.cfg -> ../boot/efi/EFI/redhat/grub.cfg
lrwxrwxrwx. 1 root root 15 Jun  9 17:03 /etc/sysconfig/grub -> ../default/grub
[core@cosa-devsh ~]$ ls -l /usr/share/man/man1/ncat.1.gz /usr/share/man/man1/unversioned-python.1.gz ../boot/efi/EFI/redhat/grub.cfg ../default/grub
ls: cannot access '/usr/share/man/man1/ncat.1.gz': No such file or directory
ls: cannot access '/usr/share/man/man1/unversioned-python.1.gz': No such file or directory
ls: cannot access '../boot/efi/EFI/redhat/grub.cfg': No such file or directory
ls: cannot access '../default/grub': No such file or directory
```

travier:
I think we should. If they are broken we should fix them or remove them.

travier:
And I would check both etc & usr for broken symlinks

travier:
Made https://github.com/coreos/fedora-coreos-config/issues/1782 for tracking

PR #1779, opened 2022-06-09
denylist: drop denials for some s390x tests

LakshmiRavichandran1:
The tests 
- https://github.com/coreos/fedora-coreos-tracker/issues/1204 , 
- https://github.com/coreos/fedora-coreos-tracker/issues/1205, 
- https://github.com/coreos/fedora-coreos-tracker/issues/1207 

no longer fail because of an `ostree-2022.4-2.fc36 `update.

The `ext.config.ignition.resource.remote` (https://github.com/coreos/fedora-coreos-tracker/issues/1215) seems to be passing now as well (Not sure what changed there, but happy to see it passing.)

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -17,23 +17,7 @@
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:
   - ppc64le
-- pattern: rpmostree.install-uninstall
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1204
-  arches:
-  - s390x
-- pattern: rpmostree.upgrade-rollback
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1205
-  arches:
-  - s390x
-- pattern: ostree.hotfix
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1207
-  arches:
-  - s390x
 - pattern: multipath.day2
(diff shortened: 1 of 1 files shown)


dustymabe:
Let's update the commit message here to have more information.

Something like:

```
denylist: drop denials for some s390x tests

The first three tests no longer fail because of an ostree update.
The `ext.config.ignition.resource.remote` seems to be passing now
as well. Not sure what changed there, but happy to see it passing.
```

We should also put a comment in https://github.com/coreos/fedora-coreos-tracker/issues/1215 with the new information about the test now passing in local tests.

LakshmiRavichandran1:
@dustymabe updated as per your comments

dustymabe:
Still need to update the git commit (not just the PR description).

LakshmiRavichandran1:
> Still need to update the git commit (not just the PR description).

should look good now

dustymabe:
> > 
> 
> should look good now

almost!

There's still no body in the commit message. Just a title. It should be in format like:

```
this is the title line

This is line 1 of the body.
This is line 2 of the body.
```

dustymabe:
Thanks @LakshmiRavichandran1

dustymabe:
I'll merge this since it's already passed CI before the last push that just updated the git message and also because the change wouldonly be tested on s390x.

dustymabe (approved):

PR #1780, opened 2022-06-09
[rawhide] overrides: remove systemd-251~rc2-2.fc37 pin

gursewak1997:
Based on the fixes in referenced systemd PR in https://github.com/coreos/fedora-coreos-tracker/issues/1200 we can remove the systemd-251~rc2-2.fc37 pin from the overrides.
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1200


dustymabe (approved):
LGTM

jlebon (approved):

PR #1781, opened 2022-06-09
[next-devel] Add platform-specific console configuration metadata

bgilbert:
coreos-assembler `create_disk.sh` and `gf-set-platform` will read this.  `create_disk.sh` will also save a JSON version for the relevant architecture to `/boot/coreos/platforms.json` for coreos-installer.

Continuation of #1181.

### DIFF
--- a/platforms.yaml
+++ b/platforms.yaml
@@ -1,0 +1,115 @@
+# This file specifies image customizations that are specific to particular
+# architecture/platform pairs.  It is applied (indirectly, via
+# /boot/coreos/platforms.json) by create_disk.sh and gf-set-platform at
+# build time and coreos-installer at install time (if --platform is
+# specified).
+#
+# Currently this is used to configure the default console.  For any
+# arch/platform pairs not specified, GRUB and the kernel will apply their
+# own defaults.  Note that coreos-installer install --console will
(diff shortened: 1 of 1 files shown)


dustymabe (approved):

dustymabe:
We could go ahead and get this in `rawhide` now if we wanted some advance testing.

bgilbert:
Rebased.  Moved tests to https://github.com/coreos/fedora-coreos-config/pull/1974.  Re-added graphical consoles to AWS, Azure, and GCP on x86_64, since those platforms allow taking screenshots of the graphical console to diagnose the VM.

dustymabe:
We've got the major cloud platforms covered here. Any other ones we think we need to add?

bgilbert:
I've rechecked all of the FCOS platforms against any documentation provided by the platform provider.  The `ibmcloud` serial console apparently slipped through last time, so I've now added that.

dustymabe:
any sane defaults we could use for openstack?

bgilbert:
We're using them.  OpenStack can support both graphical and serial console, but graphical console appears to be the one that's emphasized by the UI.

dustymabe (approved):

PR #1783, opened 2022-06-10
overrides: fast-track ostree-2022.4-2.fc37

LakshmiRavichandran1:
To fix the tests
- kola: s390x: test rpmostree.install-uninstall fails fedora-coreos-tracker#1204 ,
- kola: s390x: test rpmostree.upgrade-rollback fails fedora-coreos-tracker#1205,
- kola: s390x: ostree.hotfix test fails fedora-coreos-tracker#1207

that fail in `rawhide` branch which can be bumped if ostree-2022.4-2.fc37 is used.


dustymabe (approved):
LGTM

PR #1785, opened 2022-06-13
denylist: deny ext.config.ignition.resource.remote on s390x

dustymabe:
Turns out the issue is a race condition and we need to denylist it
again while we investigate it more.

See https://github.com/coreos/fedora-coreos-tracker/issues/1215


LakshmiRavichandran1:
lgtm

LakshmiRavichandran1 (approved):

PR #1789, opened 2022-06-14
kola-denylist.yaml: denylist ext.config.extensions.module

saqibali-2k:
The test is failing in rawhide so let's denylist it there
for now.

see: https://github.com/coreos/fedora-coreos-tracker/issues/1229


jlebon (approved):
Thanks!

PR #1795, opened 2022-06-16
Revert "overrides: fast-track kernel-5.18.4-201.fc36"

dustymabe:
This reverts commit a9a8e2a4423c52f96ab4407574a8738ce463e79e.

It was only a test to see if the 5.18 kernel would pass all pipeline
tests on various platforms. It passed! Now we can go back to tracking
what is in bodhi stable for the kernel.


bgilbert (approved):

dustymabe:
From CI:

```
```

The downgrade is intentional.

PR #1796, opened 2022-06-16
denylist: deny rpmostree.install-uninstall on s390x on F36 branches

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/1215#issuecomment-1157595900


LakshmiRavichandran1 (approved):

dustymabe:
From CI: 

```
```

This is going to happen because of #1795 until the next build comes through.

PR #1797, opened 2022-06-16
overlay.d: make generators run once

dustymabe:
Generators get re-executed each time a systemctl daemon-reload happens.
Let's update our generators so that they only run once.

I was led down this path because I noticed this failure in the console logs:

```
[    5.733962] ln:
[    5.733968] failed to create symbolic link '/run/udev/rules.d/80-coreos-boot-disk.rules'
[    5.734511] : File exists
[    5.737931] systemd[1023]: /usr/lib/systemd/system-generators/coreos-diskful-generator failed with exit status 1.
```

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -12,6 +12,16 @@ set -e
 # https://github.com/systemd/systemd/issues/15638
 exec 1>/dev/kmsg; exec 2>&1
 
+# Generators get re-executed each time a daemon-reload happens.
+# We only need to run once.
+self=$(basename $0)
+if [ -e "/run/coreos-${self}" ]; then
+    echo "$self: generator has already run; skipping"
+    exit 0
+else
+    > "/run/coreos-${self}"
+fi
+
 UNIT_DIR="${1:-/tmp}"
 
 cmdline=( $(</proc/cmdline) )
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
@@ -6,6 +6,16 @@
 # https://github.com/systemd/systemd/issues/15638
 exec 1>/dev/kmsg; exec 2>&1
 
+# Generators get re-executed each time a daemon-reload happens.
+# We only need to run once.
+self=$(basename $0)
+if [ -e "/run/coreos-${self}" ]; then
+    echo "$self: generator has already run; skipping"
+    exit 0
+else
+    > "/run/coreos-${self}"
+fi
+
 command -v getarg >/dev/null || . /usr/lib/dracut-lib.sh
 
 set -e
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-generator
@@ -4,6 +4,16 @@
 # https://github.com/systemd/systemd/issues/15638
 exec 1>/dev/kmsg; exec 2>&1
 
+# Generators get re-executed each time a daemon-reload happens.
+# We only need to run once.
+self=$(basename $0)
+if [ -e "/run/coreos-${self}" ]; then
+    echo "$self: generator has already run; skipping"
+    exit 0
+else
+    > "/run/coreos-${self}"
+fi
+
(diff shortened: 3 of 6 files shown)


dustymabe:
There's also a second commit in there which modernizes `coreos-platform-chrony` generator.

dustymabe:
With this code we now see this in the console output:

```
[    4.954773] systemd[1]: Starting initrd-parse-etc.service - Reload Configuration from the Real Root...^M                                                                                                              
[    4.956767] systemd[1]: Reloading.^M                                                                                                                                                                                  
[    4.967762] EXT4-fs (vda3): mounted filesystem with ordered data mode. Quota mode: none.^M                                                                                                                            
[    4.978397] coreos-boot-edit[879]: Injected kernel arguments into BLS: root=UUID=3d67418e-3045-493c-9ea6-0eabe271f694 rw rootflags=prjquota^M                                                                         
[    5.004877] coreos-multipath-generator: generator has already run; skipping^M                                                                                                                                         
[    5.008180] coreos-diskful-generator: generator has already run; skipping^M                                                                                                                                           
[    5.008892] FAT-fs (vda2): Volume was not properly unmounted. Some data may be corrupt. Please run fsck.^M                                                                                                            
[    5.011309] live-generator: generator has already run; skipping
```

dustymabe:
From CI: 

```
```

This is going to happen because of #1795 until the next build comes through.

dustymabe:
Tagging in @bgilbert and/or @jlebon here for some perspective/expertise around how systemd-generators are handled and if this is a good idea or not.

jlebon:
I think the reason `systemctl daemon-reload` reruns generators is because if a generator has changed, it wants to make sure it has the latest output. Note it also knows to nuke everything created by generators first so it starts fresh. From `systemd.generator(7)`:

> After installing new generators or updating the configuration, `systemctl daemon-reload` may be executed. This will delete the previous configuration created by generators, re-run all generators, and cause `systemd` to reload units from disk.

The reason we're getting that symlink error is because we're abusing a bit systemd generators by writing things that fall outside designated generator-targeted outputs. So I think the right thing to do instead is to either move those out of generators or make sure those few bits are idempotent.

Edit: to make this really clear: I think the issue with this PR is that we'd lose everything generated by our generators on `systemctl daemon-reload` because they'll get nuked and not get recreated.

Second commit looks sane though!

jlebon:
Alternative in https://github.com/coreos/fedora-coreos-config/pull/1801.

dustymabe:
> Note it also knows to nuke everything created by generators first so it starts fresh.

I didn't know that.. Hmm. Wonder how that works under the hood?

> The reason we're getting that symlink error is because we're abusing a bit systemd generators by writing things that fall outside designated generator-targeted outputs. So I think the right thing to do instead is to either move those out of generators or make sure those few bits are idempotent.

Agree. We should probably stop abusing generators in this way.

cgwalters:
> I think the reason systemctl daemon-reload reruns generators is because if a generator has changed, it wants to make sure it has the latest output.

I think it's probably the case today that most of our generators don't operate on any data which could logically change once the system is booted.  Also our generator logic doesn't change often.  And we're not very oriented towards live updates for FCOS.

So...I think these things together weigh against making generators across-the-board idempotent via the "simple stamp file" approach.  Making them idempotent in the actual logic (as in https://github.com/coreos/fedora-coreos-config/pull/1801 ) seems good to me, but that needs case by case analysis.

All of this said, I think let's get https://github.com/coreos/fedora-coreos-config/pull/1797/commits/1a204418d187834453dea01d991b3c3353937b98 merged from this PR!

jlebon:
Agreed the second commit is a nice cleanup.
@dustymabe Can you split that out into a separate PR (or just repurpose this PR)?

> I didn't know that.. Hmm. Wonder how that works under the hood?

Generators are only supposed to write to [a specific set of directories passed on the cmdline](https://github.com/systemd/systemd/blob/89f119b6549e6b92360945968195b4aaf7d20280/man/systemd.generator.xml#L27-L30), but it's not strictly enforced. systemd just [empties those directories](https://github.com/systemd/systemd/blob/89f119b6549e6b92360945968195b4aaf7d20280/src/core/generator-setup.c#L49-L57) when reloading.

cgwalters:
Tangentially we should have a CI test that does `systemctl daemon-reload` though (which will re-exec our generators) and verifies no generators fail, since dynamically reloading systemd (e.g. adding units not via ignition) is a valid use case even for us I think.

dustymabe:
> @dustymabe Can you split that out into a separate PR (or just repurpose this PR)?

yep - will do.

dustymabe:
> > @dustymabe Can you split that out into a separate PR (or just repurpose this PR)?
> 
> yep - will do.

Broke out that second commit into https://github.com/coreos/fedora-coreos-config/pull/1962

I'm going to close this now since the discussion exposed that generators are intended to be run more than once.

PR #1799, opened 2022-06-17
ci/overrides: fix Bodhi lookup with bodhi-client 6.x

bgilbert:
We were assuming that `bodhi.client` re-exported `bodhi.client.bindings`, but apparently that wasn't contractual.  Import the module we actually need.

Fixes error:

    AttributeError: module 'bodhi.client' has no attribute 'bindings'

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -10,7 +10,7 @@ from urllib.parse import urlparse
 import yaml
 import subprocess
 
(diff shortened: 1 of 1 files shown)


marmijo (approved):
LGTM

bgilbert:
CI doesn't test this.  Merging.

PR #1801, opened 2022-06-17
overlay/coreos-diskful-generator: make symlink creation idempotent

jlebon:
A generator can be run multiple times. This symlink is a bit of an abuse
of systemd generators because it writes things outside of the expected
output directories. So we need to make sure that we're idempotent.


dustymabe:
yeah I had this in my local checkout first before I decided to try to fix all the scripts. Reading the man page I didn't realize the `This will delete the previous configuration created by generators` part.

dustymabe (approved):
LGTM

PR #1805, opened 2022-06-20
manifests/group: drop well-known groups with static IDs

lucab:
This drops some groups that we do not strictly need to override.
All these groups are created by scriptlets in packages, and they
have been allocated a static group ID (GID) in the `uidgid` table.
There is no need to pin those GIDs through manifest files, as the
groupname <-> GID mapping does not change across builds.

Ref: https://pagure.io/setup/blob/master/f/uidgid

### DIFF
--- a/manifests/group
+++ b/manifests/group
@@ -23,8 +23,6 @@ lock:x:54:
 audio:x:63:
 nobody:x:99:
 users:x:100:
-utmp:x:22:
-utempter:x:35:
 ssh_keys:x:999:
 systemd-journal:x:190:
 dbus:x:81:
@@ -34,13 +32,11 @@ dip:x:40:
 cgred:x:996:
 tss:x:59:
 avahi-autoipd:x:170:
-rpc:x:32:
 sssd:x:993:
 dockerroot:x:986:
 rpcuser:x:29:
 nfsnobody:x:65534:
 kube:x:994:
-sshd:x:74:
 chrony:x:992:
 tcpdump:x:72:
 ceph:x:167:
--- a/tests/kola/files/fcos_groups
+++ b/tests/kola/files/fcos_groups
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+# kola: { "distros": "fcos", "exclusive": false }
+#
+# This performs a sanity check on system groups that are shipped
+# as part of the base OS. Those groups come in different shapes
+# (with static or dynamic GIDs, from plain files or from scriptlets)
+# and each case is covered by a corresponding check here below.
+
+set -euo pipefail
+
+. "${KOLA_EXT_DATA}/commonlib.sh"
+
+# Check base system groups (from `setup` package).
+declare -A setup_groups=( \
+    ["root"]="0" \
+    ["bin"]="1" \
+    ["daemon"]="2" \
+    ["sys"]="3" \
+    ["adm"]="4" \
+    ["tty"]="5" \
+    ["disk"]="6" \
+    ["lp"]="7" \
+    ["mem"]="8" \
+    ["kmem"]="9" \
+    ["wheel"]="10" \
+    ["cdrom"]="11" \
+    ["mail"]="12" \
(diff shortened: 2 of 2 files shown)


lucab:
There are two more outliers that I left in the list (`dbus` and `tss`), because their parent packages ship a mismatched `sysusers.d` fragment.
I was thus not fully confident about dropping those two, and preferred leaving them untouched. I'll explore them separately.

travier:
LGTM but we might want to create a quick test for this one until we've completed the sysusers transition.

lucab:
CI is failing on `ext.config.extensions.package` with:
```
Jun 20 15:57:06.911595 rpm-ostree[2189]: /usr/lib/tmpfiles.d/pkg-nfs-utils-coreos.conf:2: Failed to resolve group 'rpcuser'.
Jun 20 15:57:06.912947 rpm-ostree[2189]: /usr/lib/tmpfiles.d/pkg-nfs-utils-coreos.conf:3: Failed to resolve group 'rpcuser'.
Jun 20 15:57:06.913943 rpm-ostree[2189]: /usr/lib/tmpfiles.d/pkg-nfs-utils-coreos.conf:4: Failed to resolve group 'rpcuser'.
```

I think this is an unrelated bug in `nfs-utils-coreos` package, which is using `rpcuser` as the owner of some content but it does not actually create that user/group.
I'll leave the `rpcuser` group in our overrides for now.

lucab:
Added a `ext.config.files.groups` test which covers the following cases:
 * base groups, provided by files content from the `setup` RPM
 * extra groups with static GIDs, created by RPM scriptlets
 * extra groups with dynamic GIDs, created by RPM scriplets

This should be ready to land now.

dustymabe:
> CI is failing on `ext.config.extensions.package` with:

yay for CI

Thank you @lucab for working on this.

tests/kola/files/groups
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+# kola: { "exclusive": false }
```
dustymabe: any chance we can add up here a description of this test?

lucab: Yup, test amended with a short summary/preface.

lucab:
Amended, PTAL.

travier:
I know it would be painful but should we consider having an exhaustive test for all users and groups here? If we test only a subset then we may miss some changes. We will also be more confident about not breaking existing systems during the transition.

lucab:
I'd rather not ossify the current status quo via full-coverage tests.
Several of those group entries are either out of sync with Fedora `uidgid` allocation, or related to softwares that we aren't actually shipping in the base OS.
I'd rather try to completely drain this table from here, and add specific tests for the few odd cases that we encounter as we go.

jlebon (dismissed):
Nice!

Re. the `nfs-utils-coreos` issue, should we file an RHBZ?

travier:
> I'd rather not ossify the current status quo via full-coverage tests. Several of those group entries are either out of sync with Fedora `uidgid` allocation, or related to softwares that we aren't actually shipping in the base OS. I'd rather try to completely drain this table from here, and add specific tests for the few odd cases that we encounter as we go.

Well, right now if we remove a group or user that someone was relying on, then we will break their system on upgrade as the UID or GID may change / disappear (unless I'm missing something).

Thus I'm not sure about this.

travier:
In other words, until we have full conversion from the status quo to per node `passwd` & `group` entries without nss-altfiles, we can not remove a single group or user entry.

Thus I think we need to be exhaustive in our testing here.

lucab:
The `nfs-utils-coreos` issue is tracked at https://bugzilla.redhat.com/show_bug.cgi?id=2103114.

I've reworked the test, it now checks all the static groups we are currently shipping. This also ensures that all the groups that are dropped in this PR are actually still present in the OS (this time properly created by the relevant packages).

travier (approved):

Issue #1775, opened 2022-06-08
Add kola test to check for initrd udev rules


bgilbert:
The initrd includes cloud-specific udev rules for at least AWS and Azure that, if omitted from the image, will cause certain Ignition configs to fail ([example](https://bugzilla.redhat.com/show_bug.cgi?id=2092966)).  We don't have kola CI on some of those clouds, and even when we do, may not have kola tests that use the specific device aliases created by these rules.

Since the rules files might move between packages/repos or disappear from a package/repo entirely, let's add a kola test that checks whether we're shipping the rules in the image.  It won't check the functionality of the rules, but we don't maintain them ourselves, so that seems reasonable.

The existing `ext.config.files.dracut-executable` test might be a good starting point.  `lsinitrd --unpack` or `coreos-installer dev extract initrd` can be used to unpack the initrd contents.

The test should be run on both FCOS and RHCOS.

prestist:
This has been closed. The test was added [here](https://github.com/coreos/fedora-coreos-config/pull/1807).

PR #1807, opened 2022-06-21
add kola test to assert initrd files exist

prestist:
https://github.com/coreos/fedora-coreos-config/issues/1775

Added a kola test to ensure that initrd files exist.

### DIFF
--- a/tests/kola/files/expected-initrd-files
+++ b/tests/kola/files/expected-initrd-files
@@ -1,0 +1,29 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+
+# This test runs on both, FCOS&RHCOS. The initrd includes specific files which,  
+# if omitted from the image will cause some failures with certain ingnition
+# configurations. This test doesnt assert the functionality of any files, it
+# simply gives a high level check to see if the files are available. 
+# See https://github.com/coreos/fedora-coreos-config/issues/1775
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+ 
+required_initrd_files=(
+    "/usr/lib/udev/rules.d/66-azure-storage.rules"
+    "/usr/lib/udev/rules.d/99-azure-product-uuid.rules"
+)
+
+tmpd=$(mktemp -d)
+trap "rm -r ${tmpd}" EXIT
+( cd "${tmpd}" && lsinitrd --unpack /boot/ostree/*/init* )
+
+for file in ${required_initrd_files[@]}; do
+    if [ ! -e "${tmpd}/${file}" ]; then 
+        fatal "${file} was not found in initrd"
+    fi
+done
+
+ok "Found expected initrd files"


tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# initrd includes cloud-specific udev rules for some cloud providers that , if omitted from the 
+# image will cause some ignition configs to fail. While the udev rules and there explicit 
+# functionality is not mantained by fcos/rhcos we can ensure that 66-azure-storage.rules, and 
+# 99-azure-product-uuid.rules are pressent.
+
+declare -a required_udev_rules=("66-azure-storage.rules" "99-azure-product-uuid.rules")
```
bgilbert: I believe the `declare -a` is unnecessary?

Let's split this into one filename per line.

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+
+tmpd=$(mktemp -d)
+( cd "${tmpd}" && lsinitrd --unpack /boot/ostree/*/init* )
+
+for rule in ${required_udev_rules[@]};
```
bgilbert: Let's either move the `do` up from the next line, or drop the semicolon.

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+# 99-azure-product-uuid.rules are pressent.
+
+declare -a required_udev_rules=("66-azure-storage.rules" "99-azure-product-uuid.rules")
+
+tmpd=$(mktemp -d)
+( cd "${tmpd}" && lsinitrd --unpack /boot/ostree/*/init* )
+
+for rule in ${required_udev_rules[@]}; 
+do
+    if find "${tmpd}"{/usr/local/lib,/usr/lib}"/udev/rules.d/${rule}" | wc -c == 0; then
```
bgilbert: As a distro, we shouldn't be shipping anything in `/usr/local`.  So `[ -e ... ]` should be sufficient.

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+for rule in ${required_udev_rules[@]}; 
+do
+    if find "${tmpd}"{/usr/local/lib,/usr/lib}"/udev/rules.d/${rule}" | wc -c == 0; then 
+        fatal "${rule} is not found"
+    fi
+done
+
+rm -r "${tmpd}"
+
+ok "Found expected udev rules"
```
bgilbert: Missing trailing newline.

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+( cd "${tmpd}" && lsinitrd --unpack /boot/ostree/*/init* )
+
+for rule in ${required_udev_rules[@]}; 
+do
+    if find "${tmpd}"{/usr/local/lib,/usr/lib}"/udev/rules.d/${rule}" | wc -c == 0; then 
+        fatal "${rule} is not found"
+    fi
+done
+
+rm -r "${tmpd}"
```
bgilbert: Idiomatically we put this right after the `mktemp` call, in the form `trap "rm -r ${tmpd}" EXIT`.

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+
+# initrd includes cloud-specific udev rules for some cloud providers that , if omitted from the 
+# image will cause some ignition configs to fail. While the udev rules and there explicit 
+# functionality is not mantained by fcos/rhcos we can ensure that 66-azure-storage.rules, and 
+# 99-azure-product-uuid.rules are pressent.
+
+declare -a required_udev_rules=("66-azure-storage.rules" "99-azure-product-uuid.rules")
+
+tmpd=$(mktemp -d)
+( cd "${tmpd}" && lsinitrd --unpack /boot/ostree/*/init* )
```
bgilbert: This will explode in interesting ways if the glob matches more than one file.  Maybe we should check for that?

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+
+for rule in ${required_udev_rules[@]}; 
+do
+    if find "${tmpd}"{/usr/local/lib,/usr/lib}"/udev/rules.d/${rule}" | wc -c == 0; then 
+        fatal "${rule} is not found"
```
bgilbert: ...not found in the initrd, maybe?

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,27 @@
...
+# kola: { "exclusive": false }
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# initrd includes cloud-specific udev rules for some cloud providers that , if omitted from the 
+# image will cause some ignition configs to fail. While the udev rules and there explicit 
+# functionality is not mantained by fcos/rhcos we can ensure that 66-azure-storage.rules, and 
+# 99-azure-product-uuid.rules are pressent.
```
bgilbert:
```suggestion
# initrd includes cloud-specific udev rules for some cloud providers which, if omitted from the 
# image, will cause some ignition configs to fail. While the udev rules and their explicit 
# functionality is not mantained by FCOS/RHCOS we can ensure that the rules files
# are present.
```

There's no point in maintaining a parallel list of rules files in the header comment, I think.

tests/kola/files/expected-udev-rules
prestist: (⌐■_■)

( •_•)>⌐■-■

(•_•)
Now thats cool

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,28 @@
...
+. $KOLA_EXT_DATA/commonlib.sh
+
+# initrd includes cloud-specific udev rules for some cloud providers which, if omitted from the 
+# image, will cause some ignition configs to fail. While the udev rules and their explicit 
+# functionality is not mantained by FCOS/RHCOS we can ensure that files are pressent.
+ 
+required_udev_rules=(
+    "66-azure-storage.rules"
+    "99-azure-product-uuid.rules"
+    )
```
bgilbert:
```suggestion
)
```

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+
```
bgilbert: We normally include a comment block here that explains what the test does, and why we've enabled each of the specified `kola` properties.

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,28 @@
...
+ 
+required_udev_rules=(
+    "66-azure-storage.rules"
+    "99-azure-product-uuid.rules"
+    )
+
+tmpd=$(mktemp -d)
+( cd "${tmpd}" && lsinitrd --unpack /boot/ostree/*/init* )
+
+trap "rm -r ${tmpd}" EXIT
```
bgilbert: Should be immediately after `tmpd` is assigned.

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,28 @@
...
+
+for rule in ${required_udev_rules[@]}; do
+    if [-e "${tmpd}/usr/lib/udev/rules.d/${rule}"]; then
```
bgilbert: The sense of this is inverted, and also it doesn't work.  `[` is a program but `[-e` isn't.  :sparkles: bash scripting. :sparkles:

```suggestion
    if [ ! -e "${tmpd}/usr/lib/udev/rules.d/${rule}" ]; then 
```

tests/kola/files/expected-udev-rules
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
bgilbert: Maybe the test's filename and log messages should be broadened to include all expected initrd files?  In principle there's no reason we couldn't add non-udev files to the list.

tests/kola/files/expected-udev-rules
dustymabe: for example: https://github.com/coreos/fedora-coreos-config/blob/c048af0980da12cf6d339f5d8b27e62eb1259c4d/tests/kola/podman/dns/test.sh#L4-L18

tests/kola/files/expected-udev-rules
prestist: After our conversation I think for now, I am going to keep this as is. There is already precedent for this form.

prestist:
I think this is ready for review again!

tests/kola/files/expected-initrd-files
```diff
@@ -0,0 +1,29 @@
...
+. $KOLA_EXT_DATA/commonlib.sh
+ 
+required_initrd_files=(
+    "/usr/lib/udev/rules.d/66-azure-storage.rules"
+    "/usr/lib/udev/rules.d/99-azure-product-uuid.rules"
+)
+
+tmpd=$(mktemp -d)
+trap "rm -r ${tmpd}" EXIT
+( cd "${tmpd}" && lsinitrd --unpack /boot/ostree/*/init* )
```
cgwalters, line 21: This is totally fine as is.  I just want to note an interesting intersection here with https://github.com/coreos/coreos-assembler/issues/2952

If you used instead `/usr/lib/modules/<kver>/initramfs.img` then this test would *also* work in our container images.

(But OTOH, since the initramfs only matters for host systems, it does make sense to test the final value in `/boot`)

cgwalters (approved):
I didn't follow previous review on this PR but LGTM!

PR #1808, opened 2022-06-21
ci/buildroot: install rpm-ostree test deps

jlebon:
Right now, the buildroot image only includes build deps. It's natural to
want to include test deps in there too so that devs can both build and
test.

Add a new file which lists URLs to upstream project files containing
test dependencies. Switch rpm-ostree to use that.

Delete entries for packages that are covered by the rpm-ostree test
deps.

### DIFF
--- a/ci/buildroot/install-buildroot.sh
+++ b/ci/buildroot/install-buildroot.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
-dn=$(dirname "$0")
-
 # This is invoked by Dockerfile
+
+dn=$(dirname "$0")
+tmpd=$(mktemp -d) && trap 'rm -rf ${tmpd}' EXIT
 
 echo "Installing base build requirements"
 dnf -y install /usr/bin/xargs 'dnf-command(builddep)'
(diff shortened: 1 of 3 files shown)


travier (approved):

ci/buildroot/install-buildroot.sh
```diff
@@ -16,8 +17,14 @@ echo "${brs}" | xargs dnf -y builddep
 
 echo "Installing build dependencies from canonical spec files"
 specs=$(grep -v '^#' "${dn}"/buildroot-specs.txt)
-tmpd=$(mktemp -d) && trap 'rm -rf ${tmpd}' EXIT
 (cd "${tmpd}" && echo "${specs}" | xargs curl -L --remote-name-all)
 (cd "${tmpd}" && find . -type f -print0 | xargs -0 dnf -y builddep --spec)
+rm -rf "${tmpd:?}"/*
```
travier, line 22: Why do we need this one?

jlebon: My intention was to move the `rm -rf` from being a `trap` to being done manually after each use of it, but realized I forgot to delete the `trap` when I moved all of line 19 to line 7. Anyway, it doesn't matter enough I think to waste CI on this (which doesn't know that this file isn't tested at all). Will try to remember to clean this up in a future PR when we have to do additional modifications!

PR #1810, opened 2022-06-22
35coreos-ignition: verify grub password directives

saqibali-2k:
GRUB password directives should not be present in
/boot/grub2/user.cfg if the system is not using GRUB.
Let's add a check to verify this.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
@@ -1,0 +1,17 @@
+# This unit will run late in the initrd process after the Ignition files
+# stage has completed successfully so that we may validate ignition changes
+
+[Unit]
+Description=CoreOS Post Ignition Checks
+ConditionPathExists=/usr/lib/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+# Start after Ignition has finished creating files and before ignition umount
+After=ignition-files.service
+Before=ignition-complete.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/coreos-post-ignition-checks
+RemainAfterExit=yes
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+# See coreos-post-ignition-checks.service for more information about this script
+set -euo pipefail
+
+# Verify that GRUB password directives are only used when GRUB is being used
+arch=$(uname -p)
+# Butane sugar will tell ignition to mount /boot to /sysroot/boot. We can simply check if
+# the file exists to see whether the check needs to be performed.
+# It is possible that the user creates a config, which will mount /boot at a different path
+# but that case is not officially supported. 
+if [ -f /sysroot/boot/grub2/user.cfg ]; then
+    # s390x does not use GRUB, ppcle64 uses petitboot with a GRUB config parser which does not support passwords
+    # So in both these cases, GRUB password is not supported
+    if grep -q password_pbkdf2 /sysroot/boot/grub2/user.cfg && [[ "$arch" =~ ^(s390x|ppc64le)$ ]]; then
+        echo "Ignition config provisioned a GRUB password, which is not supported on $arch"
+        exit 1
+    fi
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -23,7 +23,8 @@ install() {
         lsblk \
         sed \
         grep \
-        sgdisk
+        sgdisk \
+        uname
 
     inst_simple "$moddir/coreos-diskful-generator" \
         "$systemdutildir/system-generators/coreos-diskful-generator"
@@ -41,6 +42,11 @@ install() {
 
     inst_script "$moddir/coreos-ignition-setup-user.sh" \
         "/usr/sbin/coreos-ignition-setup-user"
+
+    inst_script "$moddir/coreos-post-ignition-checks.sh" \
+        "/usr/sbin/coreos-post-ignition-checks"
+    
+    install_ignition_unit coreos-post-ignition-checks.service
 
     # For consistency tear down the network and persist multipath between the initramfs and
     # real root. See https://github.com/coreos/fedora-coreos-tracker/issues/394#issuecomment-599721763


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# See coreos-post-ignition-checks.service for more information about this script
+set -euo pipefail
+
+bootmnt=/mnt/boot_partition
+mkdir -p $bootmnt
+# We need to mount as rw because ignition may have also mounted
+# this to write user.cfg for GRUB
+mount -o rw /dev/disk/by-label/boot $bootmnt
```
dustymabe: interesting.. shouldn't ignition have unmounted it, though?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Verify that GRUB password directives are only used when GRUB is being used
+arch=$(uname -p)
+if [ -f $bootmnt/grub2/user.cfg ]; then
+    if grep -q password_pbkdf2 $bootmnt/grub2/user.cfg && [ "$arch" = "s390x" ]; then
```
dustymabe: does the grub fragment feature work on ppc64le?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
bgilbert: Not if this runs before the umount stage.  But I don't understand the logic here.  If Ignition didn't mount `/boot`, definitionally there's nothing to check, and if it did, why not reuse the existing mount?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
```diff
@@ -0,0 +1,20 @@
...
+# ignition changes are valid
+
+[Unit]
+Description=CoreOS Post Ignition Check
+ConditionPathExists=/usr/lib/initrd-release
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+# Start after Ignition has finished creating files
+After=ignition-files.service
```
bgilbert, line 11: The implication is that you want to run before the Ignition umount stage, but you don't actually say that, so this is racy.  The syntax for that is a bit awkward, because the umount stage actually runs when `ignition-mount.service` is stopped: `Before=ignition-complete.target`

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
bgilbert: The feature does work on ppc64le, but passwords do not.  We should check for ppc64le here too.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,18 @@
...
+        echo "$arch does not use GRUB, but GRUB password directive(s) were found: Incorrect ignition config"
```
bgilbert: Maybe something like "Ignition config provisioned a GRUB password, which is not supported on $arch"?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
dustymabe: Is it a bug that it doesn't work or was it never designed to work? A comment or a link to a statement about more information could be useful.

s390x doesn't use grub so it makes sense that it doesn't work there.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
dustymabe: +1 - that handles the nuance well.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
bgilbert: It was never designed to work.  petitboot has a GRUB config parser which only understands certain commands, and optimistically ignores the rest.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
saqibali-2k: > If Ignition didn't mount /boot, definitionally there's nothing to check, and if it did, why not reuse the existing mount?

I'll try that approach.  Aside: `coreos-boot-edit.service` also mounts `/boot` and runs after the ignition mount stage. We haven't seen any conflicts yet, but should we also update how we mount `/boot` there (add a check to see if `/boot` is already mounted)?

jlebon: Strong agreement with @bgilbert. The less things mount the bootfs, the better. :)

`coreos-boot-edit.service` is different because it does things (like adding the `root` karg) which need to happen regardless of whether or not the Ignition config wants to write to the bootfs.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
```diff
@@ -0,0 +1,17 @@
+# This unit will run late in the initrd process after the Ignition files
+# stage has completed successfully and will temporarily mount /boot to verify 
+# ignition changes are valid
```
jlebon: This comment needs updating.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# See coreos-post-ignition-checks.service for more information about this script
+set -euo pipefail
+
+# Get mountpoint of boot device
+bootmnt=$(findmnt --source /dev/disk/by-label/boot -no TARGET; 2>/dev/null)
```
jlebon: This looks odd. Did you mean

```suggestion
bootmnt=$(findmnt --source /dev/disk/by-label/boot -no TARGET 2>/dev/null)
```

?

Though we need to handle the error case. I guess we can do

```suggestion
bootmnt=$(findmnt --source /dev/disk/by-label/boot -no TARGET 2>/dev/null || :)
```

We usually should avoid this idiom IMO because it can mask real issues, but I don't have a good alternative. I guess we could iterate over all mounts looking for the matching `SOURCE` source, but that would be better done in !bash.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ "$bootmnt" = "" ]; then
```
dustymabe:
```suggestion
if [ -z "$bootmnt" ]; then
```

Another way to do this..

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
dustymabe: how about something like 

```
if [ ! -e /dev/disk/by-label/boot ]; then
    should we error here if the device doesn't exist?
fi

bootdev=$(realpath /dev/disk/by-label/boot)
mountinfo=$(findmnt --json)
bootmnt=$(jq -r ".filesystems[].children[] | select(.source == \"$bootdev\").target" <<< $mountinfo)
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,20 @@
...
+    exit 0
+fi
+bootdev=$(realpath /dev/disk/by-label/boot)
+mountinfo=$(findmnt --json)
+bootmnt=$(jq -r ".filesystems[].children[] | select(.source == \"$bootdev\").target" <<< $mountinfo)
+
+# Verify that GRUB password directives are only used when GRUB is being used
+arch=$(uname -p)
+if [ -f $bootmnt/grub2/user.cfg ]; then
+    if grep -q password_pbkdf2 $bootmnt/grub2/user.cfg && ([ "$arch" = "s390x" ] || [ "$arch" = "ppc64le" ]); then
```
bgilbert: Optional: you could trim this down with something like
```suggestion
    if grep -q password_pbkdf2 $bootmnt/grub2/user.cfg && [[ "$arch" =~ ^(s390x|ppc64le)$ ]]; then
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,20 @@
...
+# See coreos-post-ignition-checks.service for more information about this script
+set -euo pipefail
+
+# Get mountpoint of boot device
+if [ ! -e /dev/disk/by-label/boot ]; then
+    exit 0
+fi
+bootdev=$(realpath /dev/disk/by-label/boot)
+mountinfo=$(findmnt --json)
+bootmnt=$(jq -r ".filesystems[].children[] | select(.source == \"$bootdev\").target" <<< $mountinfo)
```
bgilbert: This seems like more work than necessary.  AFAICT the cases are:

1. The user hasn't written a `user.cfg`, so we don't need to care what's in it.
2. The user used Butane sugar to write `user.cfg`, or hand-rolled the Ignition themselves in the natural way (by specifying `storage.filesystems.path: /boot` for the bootfs).  `/boot` will be mounted at `/sysroot/boot` by Ignition mount stage.
3. The user hand-rolled a config to write `user.cfg` but set the `storage.filesystems.path` for the bootfs to something arbitrary.  That is technically allowed, but since the purpose of this PR is to prevent surprises when using documented functionality, this case is out of scope.

Net result: I think we should just check `/sysroot/boot/grub2/user.cfg`.  If it doesn't exist, we don't need to care why; the filesystem may not even be mounted.

saqibali-2k: +1 this makes a lot of sense and simplifies the situation.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
dustymabe: Maybe we can make this change?

Also we need to no longer use `$bootmnt` because that isn't a variable that exists any longer?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# See coreos-post-ignition-checks.service for more information about this script
+set -euo pipefail
+
+# Verify that GRUB password directives are only used when GRUB is being used
+arch=$(uname -p)
```
dustymabe, line 6: does anyone have any preference between `uname p` and just using the `arch` command here?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
dustymabe: still needs to happen IIUC

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
dustymabe: @bgilbert @saqibali-2k - is this addressed now?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
dustymabe: also let's add a comment above this noting the architecture support (i.e. the context in https://github.com/coreos/fedora-coreos-config/pull/1810#discussion_r905224049).

dustymabe:
are we calling this `coreos-post-ignition-checks.service` for a reason rather than something more specific like `grub-password-arch-check.service`. i.e. do we think this could/should be used for something else in the future?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.sh
jlebon: `uname` at least is definitely pulled in already into the initrd since we use it in `ignition-ostree-growfs.sh`. I wouldn't be surprised if `arch` was also already pulled in, but at least in this repo we don't currently use it in the initrd so I'd just stick with `uname`.

Regardless, we should add the utility we use here to the list in `module-setup.sh`.

jlebon:
For the commit title: `s/35-coreos-ignition/35coreos-ignition/`.

> are we calling this `coreos-post-ignition-checks.service` for a reason rather than something more specific like `grub-password-arch-check.service`. i.e. do we think this could/should be used for something else in the future?

No strong opinion on this, but let's try to keep `coreos-` as a prefix for consistency.

saqibali-2k:
> are we calling this `coreos-post-ignition-checks.service` for a reason rather than something more specific like `grub-password-arch-check.service`. i.e. do we think this could/should be used for something else in the future?

We didn't have a service that would perform checks after the ignition files stage and the ignition umount stage. I've kept the name generic because I thought this might be a useful service for checks of a similar nature (i.e we need to verify ignition changes that may be invalid depending on hardware, in this case the architecture tells us that GRUB is not being used).

dustymabe:
> We didn't have a service that would perform checks after the ignition files stage and the ignition umount stage.

do we have services that perform checks at other Ignition stages?

saqibali-2k:
> > We didn't have a service that would perform checks after the ignition files stage and the ignition umount stage.
> 
> do we have services that perform checks at other Ignition stages?

I think this one might fall under that category: https://github.com/coreos/fedora-coreos-config/commit/b4c81c4b6e95bec52b2c5cda750df63f3ff7be6f

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
saqibali-2k: Now its addressed :)

dustymabe (dismissed):
LGTM

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-post-ignition-checks.service
```diff
@@ -0,0 +1,17 @@
+# This unit will run late in the initrd process after the Ignition files
+# stage has completed successfully tso that we may validate ignition changes
```
bgilbert:
```suggestion
# stage has completed successfully so that we may validate ignition changes
```

bgilbert (dismissed):
Nice!

Nit: the first line of https://github.com/coreos/fedora-coreos-config/pull/1810#issuecomment-1180702193 hasn't been addressed.

dustymabe (approved):
LGTM

PR #1811, opened 2022-06-22
Drop fedora-coreos-pinger

### LINKED ISSUE #770: Remove fedora-coreos-pinger
fedora-coreos-pinger was intended to serve two purposes.  In the medium term, it was intended to provide a mechanism for collecting anonymized aggregate statistics about the Fedora CoreOS installed base.  In the short term, the plan was to ship only the parser for the config file.  That way we could ship configuration docs alongside the first release of FCOS, allowing users to opt out of statistics gathering on day 1 and avoid surprises when the rest of the statistics system eventually shipped.

We did ship the config parser, but we've never implemented the rest of the statistics system and it hasn't been a priority to do so.  Also, the only place we ever documented pinger configuration in FCOS is in the [CL migration notes](https://docs.fedoraproject.org/en-US/fedora-coreos/migrate-cl/#_operator_notes).

We're currently planning to do some basic population counting via #717, which addresses part of the problem pinger was meant to solve.  If we want to gather additional statistics down the road, we'll need to provide a lot of advance notice, since most users probably aren't aware of pinger's opt-out mechanism.  Given that, it doesn't seem important to continue shipping a tool for validating an unused config file.

Consider removing fedora-coreos-pinger from the distro.  We can always add it back later if we end up needing it.

bgilbert:
It's a stub service that only parses a config file, and we never adequately documented the config file.  We have no current plans to finish implementing the service, so let's drop it.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/770.


jlebon (approved):

dustymabe (approved):
LGTM

PR #1812, opened 2022-06-23
Enable cliwrap (rawhide only for now)

cgwalters:
The cliwrap code has existed for a long, long time.  However,
things change now with the advent of ostree-native containers.

I think it Just Makes Sense to enable `yum install foo` inside
an ostree container, which the code now does when this flag is enabled.

This will *also* cause the cliwrap code to appear on the host, but
I think that's less confusing than not doing so.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -8,3 +8,6 @@ repos:
   - fedora-rawhide
 
 include: manifests/fedora-coreos.yaml
+
+# Only enabled on rawhide right now
+cliwrap: true


cgwalters:
This is a retry now that https://github.com/coreos/rpm-ostree/pull/3737 landed in rawhide.

cgwalters:
I verified
```
=== RUN   ext.fedora-coreos-config.rawhide.kdump.crash
--- PASS: ext.fedora-coreos-config.rawhide.kdump.crash (220.02s)
PASS, output in tmp/kola/qemu-unpriv-2022-06-23-1126-280272
```

here

jlebon (approved):
Let's try it!

PR #1813, opened 2022-06-23
Add and test x86_64 emulator support on s390x and aarch64

### LINKED ISSUE #1088: New Package Request: qemu-user-static

qemu

Not sure

This package allows running x86 programs on aarch64 for docker and podman. The use case is in podman's vm for Mac: https://github.com/containers/podman/discussions/12899#discussioncomment-2068854

No. This software allows for running containers.

Not likely.

Not likely.

7. Is it possible to layer the package onto the base OS as a day 2 operation?  Explain why or why not.

Not sure

8. In the case of packages providing services and binaries, can the packaging be adjusted to just deliver binaries?

This is just binaries.

9. Can the tool(s) provided by the package be used to do things we’d rather users not be able to do in FCOS? (e.g. can it be abused as a Turing complete interpreter?)

Not sure

10. Does the software provided by the package have a history of CVEs?

Not sure. CVEs for QEMU itself are here: https://www.cvedetails.com/vulnerability-list/vendor_id-7506/Qemu.html

gursewak1997:
Fixes: [https://github.com/coreos/fedora-coreos-tracker/issues/1237](https://github.com/coreos/fedora-coreos-tracker/issues/1088)

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -195,6 +195,10 @@ packages:
 # Anyways, it was requested by the Red Hat perf team for RHCOS, so we have it here.
 # https://serverfault.com/questions/513807/is-there-still-a-use-for-irqbalance-on-modern-hardware
 # https://access.redhat.com/solutions/41535
+# 
+# Include the qemu-user-static-x86 package on aarch64 and s390x FCOS images
+# to allow access to the large inventory of containers only built for x86_64.
+# https://github.com/coreos/fedora-coreos-tracker/issues/1237
 packages-x86_64:
   - irqbalance
 packages-ppc64le:
@@ -204,6 +208,9 @@ packages-ppc64le:
   - ppc64-diag-rtas
 packages-aarch64:
   - irqbalance
+  - qemu-user-static-x86
+packages-s390x:
+  - qemu-user-static-x86
 
 # See https://github.com/coreos/bootupd
 arch-include:
--- a/tests/kola/binfmt/qemu
+++ b/tests/kola/binfmt/qemu
@@ -1,0 +1,31 @@
+#!/bin/bash
+# kola: {"platforms": "qemu", "exclusive": false, "tags": "needs-internet", "architectures": "aarch64 s390x"}
+# Test the x86_64 emulator on aarch64 and s390x images for now
+# https://github.com/coreos/fedora-coreos-tracker/issues/1237
+# 
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+# - exclusive: false
+#   - This test doesn't make meaningful changes to the system and
+#     should be able to be combined with other tests.
+# - tags: needs-internet
+#   - This test pulls a container from a registry.
+# - architectures: aarch64 s390x
+#   - We decided to ship x86_64 emulator to only these arches for now.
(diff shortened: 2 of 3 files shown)


tests/kola/binfmt/qemu
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# kola: {"exclusive": false, "tags": "needs-internet"}
+# https://github.com/coreos/fedora-coreos-tracker/issues/1237
+# Test the x86_64 emulator on aarch64 and s390x machines for now
```
dustymabe: Let's add rationale for each kola.json entry like https://github.com/coreos/fedora-coreos-config/blob/c048af0980da12cf6d339f5d8b27e62eb1259c4d/tests/kola/podman/dns/test.sh#L4-L18

We can probably limit this test to qemu (i.e. if it works anywhere it should work anywhere).

Also I wonder if we should limit this here to just s390x and aarch64 with kola.json entries for `architectures`, then we'd made the `*)` in the case statement be `fatal` (as in should not get there).

manifests/fedora-coreos-base.yaml
```diff
@@ -204,6 +204,9 @@ packages-ppc64le:
   - ppc64-diag-rtas
 packages-aarch64:
   - irqbalance
+  - qemu-user-static-x86
+packages-s390x:
+  - qemu-user-static-x86
```
dustymabe, line 213: let's add a linebreak to the comment block above and add a new comment explaining why we 're adding these.

dustymabe (commented):
Code changes mostly look good. A few suggestions.

tests/kola/binfmt/qemu
gursewak1997: > Also I wonder if we should limit this here to just s390x and aarch64 with kola.json entries for `architectures`, then we'd made the `*)` in the case statement be `fatal` (as in should not get there).

I initially added that but thought we might have new changes for other architectures soon. Nonetheless, I'll add the `architectures` entry to kola.json

manifests/fedora-coreos-base.yaml
```diff
@@ -195,6 +195,10 @@ packages:
 # Anyways, it was requested by the Red Hat perf team for RHCOS, so we have it here.
 # https://serverfault.com/questions/513807/is-there-still-a-use-for-irqbalance-on-modern-hardware
 # https://access.redhat.com/solutions/41535
+# Include the qemu-user-static-x86 package on aarch64 and s390x FCOS images
+# to allow access to the large inventory of containers only built for x86_64.
+# https://github.com/coreos/fedora-coreos-tracker/issues/1237
+
```
dustymabe: optional: move line break to break up comments

```suggestion
# https://access.redhat.com/solutions/41535
#
# Include the qemu-user-static-x86 package on aarch64 and s390x FCOS images
# to allow access to the large inventory of containers only built for x86_64.
# https://github.com/coreos/fedora-coreos-tracker/issues/1237
```

dustymabe (approved):
LGTM

dustymabe (changes requested):
Actually we need to fix the commit message.. We should be linking to/fixing https://github.com/coreos/fedora-coreos-tracker/issues/1088

dustymabe (approved):
LGTM

PR #1816, opened 2022-06-28
s390x: update manifest-lock

nikita-dubrovskii:


jlebon (approved):
Thanks! This looks sane to me. Hopefully we can drop the perl dependency soon.

PR #1817, opened 2022-06-28
denylist: ext.config.files.file-directory-permissions on s390x

LakshmiRavichandran1:
some files added by s390utils base have write permission by default.
we need to denylist now and later the package be skipped once we have https://github.com/coreos/fedora-coreos-tracker/issues/1217.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -52,3 +52,7 @@
   - aarch64
   streams:
   - rawhide
+- pattern: ext.config.files.file-directory-permissions
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1243
+  arches:
+  - s390x


jschintag:
LGTM
Thank you for doing this @LakshmiRavichandran1

kola-denylist.yaml
```diff
@@ -52,3 +52,13 @@
...
+- pattern: ext.config.files.file-directory-permissions
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1243
+  arches:
+  - s390x
+  streams:
+  - next
+  - next-devel
+  - testing
+  - testing-devel
+  - stable
```
jlebon: This is missing `rawhide`. But really, we can just remove `streams` I think so it applies to all of them.

jlebon (approved):
LGTM, thanks!
Let's get this in to unblock the pipelines.
Skipping CI since it had passed already and I just changed it so it runs on every stream which won't affect the CI result.

PR #1819, opened 2022-07-01
s390x: reencrypt boot- and rootfs if SecureExecution is enabled

nikita-dubrovskii:
Only for systems running with Secure Execution. During firstboot this reencryptes filesystems
labeled 'boot' and 'root'. Because growing of crypt-integrity device is not supported by cryptsetup,
this also grows rootfs before reencryption.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -62,4 +62,17 @@ EOF
     mkdir -p /run/udev/rules.d/
     ln -sf /usr/lib/coreos/80-coreos-boot-disk.rules \
           /run/udev/rules.d/80-coreos-boot-disk.rules
+
+    # IBM Secure Execution case
+    # During firstboot we have to reencrypt '/boot' and '/', to do that an Ignition config
+    # is injected. 'coreos-boot-disk' is required for this
+    secure_execution=0
+    if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+        secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+    fi
+    if [[ "${secure_execution}" = "1" ]]; then
+        mkdir -p /run/coreos/
+        touch /run/coreos/secure-execution
+        mv /usr/lib/coreos/01-secex.ign /usr/lib/ignition/base.d/01-secex.ign
+    fi
 fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -73,6 +73,5 @@ EOF
     if [[ "${secure_execution}" = "1" ]]; then
         mkdir -p /run/coreos/
         touch /run/coreos/secure-execution
-        mv /usr/lib/coreos/01-secex.ign /usr/lib/ignition/base.d/01-secex.ign
     fi
 fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -67,4 +67,7 @@ install() {
     install_ignition_unit coreos-ignition-unique-boot.service ignition-diskful.target
     install_ignition_unit coreos-unique-boot.service ignition-diskful.target
     install_ignition_unit coreos-ignition-setup-user.service
+
+    # IBM Secure Execution. Ignition config for reencryption of / and /boot
+    inst_simple "$moddir/01-secex.ign" /usr/lib/coreos/01-secex.ign
 }
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-close-luks.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-close-luks.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=Ignition OSTree: Close Secure Execution LUKS
+DefaultDependencies=false
+After=ignition-ostree-transposefs-save.service
+Before=ignition-ostree-transposefs-restore.service
+Before=ignition-disks.service
+ConditionArchitecture=s390x
+ConditionKernelCommandLine=ostree
+ConditionPathExists=/run/coreos/secure-execution
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/cryptsetup luksClose crypt_bootfs
+ExecStart=/usr/sbin/cryptsetup luksClose crypt_rootfs
+
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-drop-luks.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-drop-luks.service
@@ -1,0 +1,19 @@
+# This is to be removed once we switch to 'keyring' and custom 'init'.
+[Unit]
+Description=Ignition OSTree: Delete Secure Execution LUKS Keys
+ConditionArchitecture=s390x
+ConditionKernelCommandLine=ostree
+ConditionPathExists=/run/coreos/secure-execution
+DefaultDependencies=no
+
+# We should drop the keys once cryptsetup had opened both rootfs and bootfs
+After=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-root.device
+
+# Run as early as possible
+Before=cryptsetup.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/rm /etc/luks/boot /etc/luks/root /etc/crypttab
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
@@ -17,21 +17,14 @@ path=/sysroot
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
 
-# IBM SecureExecution
-secure_execution=0
-if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
-    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+# In the IBM Secure Execution case we use Ignition to grow and reencrypt rootfs
+# see overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+if [[ -f /run/coreos/secure-execution ]]; then
+    exit 0
 fi
 
 if [ ! -f "${saved_partstate}" ]; then
     partition=$(realpath /dev/disk/by-label/root)
-    # in the Secure Execution case, the rootfs is pre-baked on LUKS, so
-    # `partition` is actually the device mapper device; get its parent
-    if [[ "${secure_execution}" != "0" ]]; then
-        # lsblk doesn't print PKNAME of crypt devices with --nodeps
-        MAJMIN=$(echo $(lsblk -dno MAJ:MIN "${partition}"))
-        partition=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
-    fi
 else
     # The rootfs was reprovisioned. Our rule in this case is: we only grow if
     # the partition backing the rootfs is the same and its size didn't change
@@ -98,11 +91,7 @@ while true; do
              eval $(udevadm info --query=property --export "${NAME}")
              # lsblk doesn't print PKNAME of crypt devices with --nodeps
              PKNAME=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
-             if [[ "${secure_execution}" != "0" ]]; then
-                 cryptsetup resize --key-file=/etc/luks/root "${DM_NAME}"
-             else
-                 clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
-             fi
+             clevis_luks_unlock_device "${PKNAME}" | cryptsetup resize -d- "${DM_NAME}"
             )
             ;;
         # already checked
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-secex-config.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-secex-config.service
@@ -1,0 +1,18 @@
+# RHOCS 4.12.s390x has an old kernel with a known issue: https://bugzilla.redhat.com/show_bug.cgi?id=2075085
+# Once we have kernel >= 4.18.0-387.el8.s390x we should drop this unit and copy config in coreos-diskful-generator
+[Unit]
+Description=Ignition OSTree: Inject secex config
+DefaultDependencies=false
+ConditionArchitecture=s390x
+ConditionKernelCommandLine=ostree
+ConditionPathExists=/run/coreos/secure-execution
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+After=cryptsetup.target
+Before=ignition-fetch-offline.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/ignition-ostree-secex-config
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-secex-config.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-secex-config.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+if [[ -f /run/coreos/secure-execution ]]; then
+    bootdev=$(blkid --list-one --output device --match-token PARTLABEL=boot | sed 's,[0-9]\+$,,')
+    sed "s,\${BOOTDEV},$bootdev," < /usr/lib/coreos/01-secex.ign > /usr/lib/ignition/base.d/01-secex.ign
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -83,6 +83,24 @@ mount_and_restore_filesystem_by_label() {
     find "${saved_fs}" -mindepth 1 -maxdepth 1 -exec mv -t "${mountpoint}" {} \;
 }
 
+# In Secure Execution case user is not allowed to modify partition table
+check_and_set_secex_config() {
+    if [[ -f /run/coreos/secure-execution ]]; then
+        local wr=$(jq "$(query_fslabel root) | length" "${ignition_cfg}")
+        local wb=$(jq "$(query_fslabel boot) | length" "${ignition_cfg}")
+        if [ "${wr}${wb}" != "00" ]; then
+            echo "Modifying bootfs and rootfs is not supported in Secure Execution mode"
+            exit 1
+        fi
+        # Cached config isn't merged, so reset it and recheck again, just to make sure
+        ignition_cfg=/usr/lib/ignition/base.d/01-secex.ign
+    fi
+}
+
+# We could have done this during 'detect' below, but other cases also request
+# info from config, so just check cached one and reset to secex.ign now
+check_and_set_secex_config
+
 case "${1:-}" in
     detect)
         # Mounts are not in a private namespace so we can mount ${saved_data}
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
@@ -1,0 +1,60 @@
+{
+  "ignition": {
+    "version": "3.2.0"
+  },
+  "storage": {
+    "disks": [
+      {
+        "device": "/dev/disk/by-id/coreos-boot-disk",
+        "partitions": [
+          {
+            "label": "boot",
+            "number": 3
+          },
+          {
+            "label": "root",
+            "number": 4,
+            "resize": true,
+            "sizeMiB": 0
+          }
+        ]
+      }
+    ],
+    "filesystems": [
+      {
+        "device": "/dev/mapper/boot",
+        "format": "ext4",
+        "label": "boot",
+        "wipeFilesystem": true
+      },
+      {
+        "device": "/dev/mapper/root",
+        "format": "xfs",
+        "label": "root",
+        "wipeFilesystem": true
+      }
+    ],
+    "luks": [
+      {
+        "device": "/dev/disk/by-partlabel/boot",
+        "label": "crypt_bootfs",
+        "name": "boot",
+        "options": [
+          "--integrity",
+          "hmac-sha256"
+        ],
+        "wipeVolume": true
+      },
+      {
+        "device": "/dev/disk/by-partlabel/root",
+        "label": "crypt_rootfs",
+        "name": "root",
+        "options": [
+          "--integrity",
+          "hmac-sha256"
+        ],
+        "wipeVolume": true
+      }
+    ]
+  }
+}
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
@@ -5,7 +5,7 @@
   "storage": {
     "disks": [
       {
-        "device": "/dev/disk/by-id/coreos-boot-disk",
+        "device": "${BOOTDEV}",
         "partitions": [
           {
             "label": "boot",
@@ -36,7 +36,7 @@
     ],
     "luks": [
       {
-        "device": "/dev/disk/by-partlabel/boot",
+        "device": "${BOOTDEV}3",
         "label": "crypt_bootfs",
         "name": "boot",
         "options": [
@@ -46,7 +46,7 @@
         "wipeVolume": true
       },
       {
-        "device": "/dev/disk/by-partlabel/root",
+        "device": "${BOOTDEV}4",
         "label": "crypt_rootfs",
         "name": "root",
         "options": [
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -104,4 +104,9 @@ install() {
 
     install_ignition_unit ignition-ostree-close-luks.service
     install_ignition_unit ignition-ostree-drop-luks.service
+
+    # Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2075085
+    install_ignition_unit ignition-ostree-secex-config.service
+    inst_script "$moddir/ignition-ostree-secex-config.sh" \
+        /usr/libexec/ignition-ostree-secex-config
 }
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -101,4 +101,7 @@ install() {
         /usr/libexec/coreos-check-rootfs-size
 
     inst_script "$moddir/coreos-relabel" /usr/bin/coreos-relabel
+
+    install_ignition_unit ignition-ostree-close-luks.service
+    install_ignition_unit ignition-ostree-drop-luks.service
 }


nikita-dubrovskii:
This requires https://github.com/coreos/coreos-installer/pull/909

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -22,16 +22,13 @@ secure_execution=0
 if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
     secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
 fi
+# it's not possible to grow with luks-integrity
```
cgwalters: Won't this be kind of a showstopper?  Users will end up with the default...16GB rootfs which is going to be barely usable by default unless they e.g. provide an Ignition config that creates a `/var` partition or so?  (Hmm, do we also need to ensure we support the same LUKS+dm-integrity for that?)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -22,16 +22,13 @@ secure_execution=0
...
+if [[ "${secure_execution}" = "1" ]]; then
```
cgwalters: Not a blocker/optional but: I think it'd be a bit cleaner to instead detect that the system is dm-integrity, not that it's SE.  (LUKS + dm-integrity can make sense outside of SE too)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-reencryptfs.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
cgwalters: Obviously optional/nonblocking: some of this stuff could be in rdcore in Rust right?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -77,17 +83,44 @@ mount_and_restore_filesystem_by_label() {
...
+    fi
+
     mount_verbose "/dev/disk/by-label/${label}" "${mountpoint}"
     find "${saved_fs}" -mindepth 1 -maxdepth 1 -exec mv -t "${mountpoint}" {} \;
 }
 
+copy_luks_keys() {
+    local mountpoint=$1; shift
+    if [ "${secure_execution}" = "1" ]; then
+        root_key=$(find "${mountpoint}/ostree/deploy/" -path "*/etc/luks/root" 2>/dev/null)
```
cgwalters: This would be a bit cleaner as:
```
deploymentdir=$(ls -d /ostree/deploy/*/deploy/*.0)
root_key="${deploymentdir}/etc/luks/root"
```
or so (and same for below).

But, we can avoid the first bit by having `ostree-prepare-root.service` run which will transform `/sysroot` into the target deployment - this is what happens on a normal boot.

I think that's best implemented by running `copy_luks_keys` from a separate systemd unit  which is `After=ostree-prepare-root.service`.  See e.g. `ignition-ostree-check-rootfs-size.service` for an example.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -77,17 +83,44 @@ mount_and_restore_filesystem_by_label() {
...
+    if [ "${secure_execution}" = "0" ]; then
+        new_dev=$(jq -r "$(query_fslabel "${label}") | .[0].device" "${ignition_cfg}")
+        udev_trigger_on_label_mismatch "${label}" "${new_dev}"
+    fi
+
     mount_verbose "/dev/disk/by-label/${label}" "${mountpoint}"
     find "${saved_fs}" -mindepth 1 -maxdepth 1 -exec mv -t "${mountpoint}" {} \;
 }
 
+copy_luks_keys() {
```
cgwalters: This could use at least a brief comment describing what it's doing.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -77,17 +83,44 @@ mount_and_restore_filesystem_by_label() {
...
+    local mountpoint=$1; shift
+    if [ "${secure_execution}" = "1" ]; then
+        root_key=$(find "${mountpoint}/ostree/deploy/" -path "*/etc/luks/root" 2>/dev/null)
+        cp -f /etc/luks/root ${root_key}
+        chmod 0400 ${root_key}
```
cgwalters: Better to use e.g. `cp -m` which will preserve the mode to start.

cgwalters: (I meant `cp -p`)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: By design user could only `qemu-img resize img.qcow2 +$SIZE` and any re-partitioning is not supported. All OCP nodes on s390x rely on external storage, so even 16GB is enough

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-reencryptfs.sh
nikita-dubrovskii: Hmm, maybe. But shell code is also compact (and easier to debug with `set -x`). not sure `rdcore` should contain a lot of `s390x` specific code

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
nikita-dubrovskii: Hmm, my goal was to populate `rootfs` with new keys before `chattr +i $(ls -d /sysroot/ostree/deploy/*/deploy/*/)` . it's not a problem to make `copy_luks_keys` as a `unit`, but please confirm it's needed

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: But note we *are* growing the root partition here. This is the main reason we're reprovisioning the rootfs on first boot instead of shipping with integrity enabled at compose time, right?

So maybe let's change this comment to be e.g.

```sh
# we've already grown the root partition in the SE case (see `ignition-ostree-reencryptfs.sh`)
```

The limitation is that users won't be able to grow it again in the future without reprovisioning, which seems reasonable.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -17,28 +17,32 @@ path=/sysroot
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
 
-# IBM SecureExecution
-secure_execution=0
-if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
-    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
-fi
```
jlebon, line 24: Can't we just keep this hunk and if `secure_execution` is on, we exit early?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-reencryptfs.service
```diff
@@ -0,0 +1,25 @@
+[Unit]
+Description=Ignition OSTree: Reencrypt Partitions
```
jlebon: Maybe

```suggestion
Description=Ignition OSTree: Reencrypt Partitions (Secure Execution)
```

?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: Now `cosa` builds an image with `integrity` as well, but `/sysroot` cannot be grown (not supported by luks-integrity).

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-reencryptfs.service
```diff
@@ -0,0 +1,25 @@
...
+DefaultDependencies=false
+After=ignition-ostree-transposefs-save.service
+Before=ignition-ostree-transposefs-restore.service
+Before=ignition-disks.service
+ConditionArchitecture=s390x
+ConditionKernelCommandLine=ostree
+ConditionPathIsDirectory=/run/ignition-ostree-transposefs
+# Any services looking at mounts need to order after this
+# because it causes device re-probing.
+After=coreos-gpt-setup.service
```
jlebon: This isn't really necessary. `coreos-gpt-setup.service` runs much earlier than this service.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: Ahh yup, I see https://github.com/coreos/coreos-assembler/commit/67a3c9a9151def5157497acd31d533d58275d4eb now. Let's still change that comment though to indicate that we handled rootfs growing already.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-reencryptfs.service
```diff
@@ -0,0 +1,25 @@
...
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+# This stage requires udevd to detect disk partitioning changes.
+Requires=systemd-udevd.service
+After=systemd-udevd.service
```
jlebon: This isn't necessary either (udevd is brought up much earlier too) but meh... good to keep it as a way to declare our external dependencies.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -77,17 +83,41 @@ mount_and_restore_filesystem_by_label() {
...
+        cp -p /etc/luks/boot "${deploymentdir}/etc/luks/boot"
+        cp -p /etc/crypttab  "${deploymentdir}/etc/crypttab"
+    fi
+}
+
 case "${1:-}" in
     detect)
         # Mounts are not in a private namespace so we can mount ${saved_data}
-        wipes_root=$(jq "$(query_fslabel root) | length" "${ignition_cfg}")
-        wipes_boot=$(jq "$(query_fslabel boot) | length" "${ignition_cfg}")
```
jlebon: We could keep these and then error out in the Secure Execution case if the user requested reprovisioning themselves.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-reencryptfs.sh
```diff
@@ -0,0 +1,57 @@
...
+        ext4verity)
+            mkfs.ext4 -b 4096 -O metadata_csum_seed -O verity -L ${label} ${device} -U random
+            ;;
+        ext4)
+            mkfs.ext4 -b 4096 -O metadata_csum_seed -L ${label} ${device} -U random
+            ;;
+        xfs|"")
+            mkfs.xfs ${device} -L ${label} -m reflink=1
+            ;;
+    esac
```
jlebon: This is bound to diverge from cosa. Hmm, maybe we should actually have cosa write the options it used into the rootfs. That way it's possible to recreate the filesystem exactly as it was done. I'll look into that.

jlebon (commented):
Discussed this with @nikita-dubrovskii out of band. One suggestion I had was instead of manually growing and re-encrypting the rootfs and bootfs, let's see if we can get Ignition to do it for us by generating a base config into `/usr/lib/ignition/base.d/`. If we can pull that off, I think things here become simpler because we're reusing more of the existing reprovisioning flow.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -17,21 +17,34 @@ path=/sysroot
 # this shouldn't happen for us but we're being conservative.
 src=$(findmnt -nvr -o SOURCE "$path" | tail -n1)
 
-# IBM SecureExecution
+# In the IBM Secure Execution case we use ignition to grow and reencrypt rootfs
```
jlebon: Minor: Ignition :)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -17,21 +17,34 @@ path=/sysroot
...
+        fi
+        # skip other types of dm-devices, such as mpath
+        if [[ ! "${uuid}" =~ "CRYPT-LUKS2-" ]]; then
+            break
+        fi
         # lsblk doesn't print PKNAME of crypt devices with --nodeps
         MAJMIN=$(echo $(lsblk -dno MAJ:MIN "${partition}"))
         partition=/dev/$(ls "/sys/dev/block/${MAJMIN}/slaves")
-    fi
+    done
```
jlebon: I think we can nuke all this. In the Secure Execution case, we always reprovision root, so we'll never take this branch.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -62,4 +62,76 @@ EOF
...
+
+    # IBM Secure Execution case
+    # During firstboot we have to reenctypt '/boot' and '/', to do that an ignition config
+    # is injected. 'coreos-boot-disk' is required for this
+    secure_execution=0
+    if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+        secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+    fi
+    if [[ "${secure_execution}" = "1" ]]; then
+        cat > /usr/lib/ignition/base.d/01-secex.ign <<EOF
```
jlebon: Awesome that this works!

Let's move that file to its own Ignition file in the dracut module. Then in `module-setup.sh` we install it at e.g. `/usr/lib/coreos/secex.ign` and then here we just do a copy into `/usr/lib/ignition/base.d`.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -24,6 +24,18 @@ saved_prep=${saved_data}/prep
...
 partstate_root=/run/ignition-ostree-rootfs-partstate.sh
 
+# IBM SecureExecution with its ignition config for reencryption
+secure_execution=0
+if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+fi
+
+# It's safe to reset, beacuse on s390x KVM we use a virtio-block device to pass the config,
+# so /run/ignition.json doesn't exist.
```
jlebon: `/run/ignition.json` is the cached Ignition config. Ignition always creates it, even if the source is a local virtio block device.

I think the problem you're hitting here is that Ignition caches the provider-sourced config, not the final merged result. So base config bits don't show up there.

Hmm, I think we'll need to figure out a way to operate on the merged config. Otherwise, users could be doing other modifications via Ignition and we don't detect it. This is somewhat related to some other requests we've had 

Maybe for now... before we reset `ignition_cfg`, let's do the same `wipes_root` and `wipes_boot` checks in the `detect` case and error out if they're not 0. (This means that we won't support users overriding parts of our embedded Ignition config for the time being.)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -83,6 +95,33 @@ mount_and_restore_filesystem_by_label() {
...
+# keys generated during cosa-build, which are the same for all pristine images from given
+# build. We also remove the keys from initramfs, so no one can read them in case firstboot
+# fails in some later steps
+drop_cosa_luks_keys() {
+    local mountpoint=$1; shift
+    if [ "${secure_execution}" = "1" ]; then
+        deploymentdir=$(ls -d ${mountpoint}/ostree/deploy/*/deploy/*.0)
+        rm /etc/luks/boot "${deploymentdir}/etc/luks/boot"
+        rm /etc/luks/root "${deploymentdir}/etc/luks/root"
+        rm /etc/crypttab  "${deploymentdir}/etc/crypttab"
```
jlebon: It's going to be hard to be 100% sure users won't be able to access these keys. An additional failsafe is to also ensure it's deleted when entering the emergency initrd shell. But there might be other ways to get to a shell that this wouldn't cover (dracut can provide shells too -- they can be requested via kernel args, which we don't support changing, but it's possible it could provide some in some error paths too).

We should dig into this. If it's really not good if those per-build random keys can be read by the user, we should evaluate better options.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -83,6 +95,33 @@ mount_and_restore_filesystem_by_label() {
...
+# unmount and close / and /boot. By that time luks keys are no longer here
+close_luks_partitions() {
+    local mountpoint=$1; shift
+    if [ "${secure_execution}" = "1" ]; then
+        umount $mountpoint/boot
+        umount $mountpoint
+        cryptsetup luksClose "crypt_bootfs"
+        cryptsetup luksClose "crypt_rootfs"
+    fi
+}
```
jlebon: Hmm, this feels like it belongs in a separate systemd unit instead. It feels odd to have this here when it's not us that mounted it. Also, note this code runs from a systemd unit with `MountFlags=slave`. So I think the `umount` will take effect in our mount namespace only, but the `cryptsetup luksClose` will be global (which isn't great because they'll still in fact be mounted in the global namespace).

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
nikita-dubrovskii: Colin mentioned above, that in future we may face `dm-integrity` cases. That's why i left this here

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
nikita-dubrovskii: we can try disabling emergency shell in Step4. For now anyhow we should remove them here, otherwise Ignition would fail with `EEXIST` (or similar, cannot recall exactly)

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -67,4 +67,7 @@ install() {
     install_ignition_unit coreos-ignition-unique-boot.service ignition-diskful.target
     install_ignition_unit coreos-unique-boot.service ignition-diskful.target
     install_ignition_unit coreos-ignition-setup-user.service
+
+    # IBM Secure Execution. Ignition config for reenctyption of / and /boot
```
jlebon: Minor:

```suggestion
    # IBM Secure Execution. Ignition config for reencryption of / and /boot
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -62,4 +62,15 @@ EOF
     mkdir -p /run/udev/rules.d/
     ln -sf /usr/lib/coreos/80-coreos-boot-disk.rules \
           /run/udev/rules.d/80-coreos-boot-disk.rules
+
+    # IBM Secure Execution case
+    # During firstboot we have to reenctypt '/boot' and '/', to do that an ignition config
```
jlebon: Minor:

```suggestion
    # During firstboot we have to reencrypt '/boot' and '/', to do that an Ignition config
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -62,4 +62,15 @@ EOF
...
+    # is injected. 'coreos-boot-disk' is required for this
+    secure_execution=0
+    if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+        secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+    fi
+    if [[ "${secure_execution}" = "1" ]]; then
+        mv /usr/lib/coreos/01-secex.ign /usr/lib/ignition/base.d/01-secex.ign
+    fi
```
jlebon, line 76: Let's `touch` e.g. `/run/coreos/secure-execution` here.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-close-luks.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Ignition OSTree: Close Secure Execution LUKS
+DefaultDependencies=false
+After=ignition-ostree-transposefs-save.service
+Before=ignition-ostree-transposefs-restore.service
+Before=ignition-disks.service
+ConditionArchitecture=s390x
+ConditionKernelCommandLine=ostree
+OnFailure=emergency.target
+OnFailureJobMode=isolate
```
jlebon, line 11: Then add `ConditionPathExists=/run/coreos/secure-execution`.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-close-luks.sh
```diff
@@ -0,0 +1,14 @@
...
+# IBM Secure Execution. As we saved all, we are ready for ignition.disks, so
+# close lukes / and /boot
+secure_execution=0
+if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+fi
+
+if [[ "${secure_execution}" != "0" ]]; then
+    cryptsetup luksClose "crypt_bootfs"
+    cryptsetup luksClose "crypt_rootfs"
```
jlebon: And then fold this into the unit file directly. Multiple `ExecStart`s are supported.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -17,21 +17,34 @@ path=/sysroot
...
-# IBM SecureExecution
+# In the IBM Secure Execution case we use Ignition to grow and reencrypt rootfs
+# see overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
 secure_execution=0
 if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
     secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
 fi
+if [ "${secure_execution}" = "1" ]; then
+    exit 0
+fi
```
jlebon: Then here we can just check for `/run/coreos/secure-execution`.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: Unless we're talking about shipping images with it baked, this code would still not be used in the root reprovisioning case where a user sets up LUKS + `dm-integrity`. So let's just nuke it for now to reduce complexity.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -83,6 +89,39 @@ mount_and_restore_filesystem_by_label() {
...
+    local mountpoint=$1; shift
+    if [ "${secure_execution}" = "1" ]; then
+        deploymentdir=$(ls -d ${mountpoint}/ostree/deploy/*/deploy/*.0)
+        rm /etc/luks/boot "${deploymentdir}/etc/luks/boot"
+        rm /etc/luks/root "${deploymentdir}/etc/luks/root"
+        rm /etc/crypttab  "${deploymentdir}/etc/crypttab"
+    fi
+}
+
+# In Secre Execution case user is not allowed to modificate partition table
```
jlebon:
```suggestion
# In Secure Execution case user is not allowed to modify partition table
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -83,6 +89,39 @@ mount_and_restore_filesystem_by_label() {
...
+check_and_set_secex_config() {
+    local wr=$(jq "$(query_fslabel root) | length" "${ignition_cfg}")
+    local wb=$(jq "$(query_fslabel boot) | length" "${ignition_cfg}")
+    if [[ "${secure_execution}" != "0" ]]; then
```
jlebon: Let's hoist this out to where the function is called instead, so that we only enter it and query the Ignition config if Secure Execution is active.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -24,6 +24,12 @@ saved_prep=${saved_data}/prep
 zram_dev=${saved_data}/zram_dev
 partstate_root=/run/ignition-ostree-rootfs-partstate.sh
 
+# IBM SecureExecution with its ignition config for reencryption
+secure_execution=0
+if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+    secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+fi
```
jlebon: (Here too, we can use `/run/coreos/secure-execution` where this variable is used and drop this.)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -83,6 +89,39 @@ mount_and_restore_filesystem_by_label() {
...
+    fi
+}
+
+# In Secre Execution case user is not allowed to modificate partition table
+check_and_set_secex_config() {
+    local wr=$(jq "$(query_fslabel root) | length" "${ignition_cfg}")
+    local wb=$(jq "$(query_fslabel boot) | length" "${ignition_cfg}")
+    if [[ "${secure_execution}" != "0" ]]; then
+        if [ "${wr}${wb}" != "00" ]; then
+            echo "User is not allowed to modify partition table via Ignition"
```
jlebon: Maybe

```suggestion
            echo "Modifying bootfs and rootfs is not supported in Secure Execution mode"
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: To follow-up on this, we had a long chat about this OOB and found a more foolproof approach to making it harder using the kernel keyring. See details in [the enhancement](https://github.com/coreos/enhancements/pull/6).

jlebon: But that said, regarding the copy of the keys in the deployment, ISTM we should be able to ship without baking that in at all. I.e. in `create_disk.sh` after we've regenerated the sdboot blob, let's just delete them from the deployment?

nikita-dubrovskii: Done: https://github.com/coreos/coreos-assembler/pull/2992

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -62,4 +62,20 @@ EOF
...
+    secure_execution=0
+    if [[ $(uname -m) == s390x ]] && [[ -e /sys/firmware/uv/prot_virt_guest ]]; then
+        secure_execution=$(cat /sys/firmware/uv/prot_virt_guest)
+    fi
+    if [[ "${secure_execution}" = "1" ]]; then
+        mkdir -p /run/coreos/
+        touch /run/coreos/secure-execution
+        # FIXME:
+        # RHOCS 4.12.s390x has an old kernel with a known issue: https://bugzilla.redhat.com/show_bug.cgi?id=2075085
+        # Once we have kernel >= 4.18.0-387.el8.s390x we should revert hardcoded '/dev/vda' from the json
```
jlebon: Oh right, you mentioned this this morning OOB, but I forgot this was why we were hardcoding `/dev/vda` here. Ughhh... I do think we should clean this up because it's really easy to imagine `/dev/vda` not being the correct device. As a workaround without https://bugzilla.redhat.com/show_bug.cgi?id=2075085, I think it should work to use `blkid` from here to get the parent device of the partition labeled `boot` and then substitute that into the Ignition config.

OK, this is hacky, but something like this should work:

```
bootdev=$(blkid --list-one --output device --match-token PARTLABEL=boot | sed 's,[0-9]\+$,,')
sed "s,\${BOOTDEV},$bootdev," < /usr/lib/coreos/01-secex.ign > /usr/lib/ignition/base.d/01-secex.ign
```

(You'd replace all instances of `/dev/vda` with `${BOOTDEV}` in the Ignition config.)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-drop-luks.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
```
jlebon, line 2: OK, and we'll drop this once we switch to the kernel keyring proposal, right? Let's add a comment about that here.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-drop-luks.service
```diff
@@ -0,0 +1,20 @@
...
+Description=Ignition OSTree: Drop SE LUKS keys from ramdisk
```
jlebon: Maybe

```suggestion
Description=Ignition OSTree: Delete Secure Execution LUKS Keys
```

for consistency.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-drop-luks.service
```diff
@@ -0,0 +1,20 @@
...
+DefaultDependencies=no
+
+# We should drop the keys once cryptsetup had opened both rootfs and bootfs
+Requires=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device
+Requires=dev-disk-by\x2dlabel-root.device
+After=dev-disk-by\x2dlabel-root.device
+
+# Run as early as possible
+Before=coreos-unique-boot.service
```
jlebon: Earliest would be as soon as the devices are unlocked. All auto-generated crypt devices have `Before=cryptsetup.target`. Using that here should work too since we also have `After` on the specific devices we want.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-drop-luks.service
```diff
@@ -0,0 +1,20 @@
...
+ConditionArchitecture=s390x
+ConditionKernelCommandLine=ostree
+ConditionPathExists=/run/coreos/secure-execution
+DefaultDependencies=no
+
+# We should drop the keys once cryptsetup had opened both rootfs and bootfs
+Requires=dev-disk-by\x2dlabel-boot.device
+After=dev-disk-by\x2dlabel-boot.device
+Requires=dev-disk-by\x2dlabel-root.device
+After=dev-disk-by\x2dlabel-root.device
```
jlebon: Let's drop the `Requires` directives here. There's a tricky detail with systemd conditionals where even if you have e.g. `ConditionPathExists`, systemd will still try to bring up units specified in `Requires`. But e.g. `/dev/disk/by-label/boot` does not exist in live boots. CI still works though, so I suspect there's something subtle going on here (or maybe `ConditionArchitecture` doesn't fall in that category since it's not dynamic, though I don't see it being treated differently in a quick perusal of the systemd code).

But we shouldn't need them anyway so let's just simplify things and drop them.

jlebon:
Let's also squash everything down back to one commit.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-drop-luks.service
nikita-dubrovskii: `Before=cryptsetup.target` works, but `coreos-unique-boot.service` (on FCOS) starts first in that case

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
nikita-dubrovskii: this doesn't work at `systemd-generator` step, looks like `udev` is not yet ready, so `blkid` returns nothing

nikita-dubrovskii: added a unit as a workaround

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-secex-config.service
```diff
@@ -0,0 +1,18 @@
+# RHOCS 4.12.s390x has an old kernel with a known issue: https://bugzilla.redhat.com/show_bug.cgi?id=2075085
+# Once we have kernel >= 4.18.0-387.el8.s390x we should drop this unit and copy config in coreos-diskful-generator
+[Unit]
+Description=Ignition OSTree: Inject secex config
```
jlebon, line 4: For consistency:

```suggestion
Description=Ignition OSTree: Inject Secure Execution Config
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-secex-config.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+if [[ -f /run/coreos/secure-execution ]]; then
```
jlebon, line 4: This is already verified by systemd.

jlebon (approved):
LGTM! Two minor comments, but not worth the respin. Will address them in a follow-up.

PR #1820, opened 2022-07-04
manifests/group: drop well-known static ID for "dbus" group

lucab:
This drops a redundant override for the "dbus" group.
The group is created by the "dbus-common" package and has a static
group ID (GID) allocated in the `uidgid` table.
Out of an abundance of caution, it also temporarily removes a
buggy sysusers fragment which does not match Fedora allocation.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=2090397


travier (approved):
LGTM. Can you file https://bugzilla.redhat.com/show_bug.cgi?id=2090397 in Fedora as it is more likely to be fixed there first? Thanks!

lucab:
The F36 ticket is at https://bugzilla.redhat.com/show_bug.cgi?id=2105177.

PR #1824, opened 2022-07-04
overrides: pin NetworkManager

saqibali-2k:
We need to pin NetworkManager to silence a failure in rawhide:
https://github.com/coreos/fedora-coreos-tracker/issues/1245.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,29 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  NetworkManager:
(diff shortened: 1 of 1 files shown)


dustymabe:
```
[2022-07-04T17:41:07.714Z] Downgraded:
[2022-07-04T17:41:07.714Z]   zlib 1.2.12-2.fc37 -> 1.2.11-31.fc36
[2022-07-04T17:41:07.714Z] + grep -q Downgraded tmp/diff.txt
[2022-07-04T17:41:07.714Z] + echo 'Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.'
[2022-07-04T17:41:07.714Z] Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.
```

This is not expected but it is unrelated to this change. I opened a
ticket to track it separately: https://pagure.io/releng/issue/10875

PR #1825, opened 2022-07-06
35coreos-multipath: improve config propagation logging

jlebon:
We were saying "no initramfs automatic multipath configuration to
propagate" even when there was an initramfs config but the file already
existed in the real root. Let's split out those no op conditions more
and log clearer explicit messages.

This should be a purely cosmetic change. Prompted by looking at logging
from this service to debug a multipath-related issue.


miabbott (approved):

dustymabe (approved):
LGTM

PR #1828, opened 2022-07-07
Move from `initramfs-args` in manifest to `dracut.conf.d` files

cgwalters:
See https://github.com/coreos/rpm-ostree/issues/3799

This way when we run dracut in a container build, we naturally
pick up the same config we used for server side builds.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
+++ b/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
@@ -1,0 +1,7 @@
+# We don't support root on NFS, so we don't need it in the initramfs. It also
+# conflicts with /var mount support in ignition because NFS tries to mount stuff
+# in /var/ and then ignition can't cleanly unmount it. For example:
+# https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfs-start-rpc.sh#L7
+# See also discussion in https://github.com/coreos/fedora-coreos-config/pull/60
+# Further, we currently do not use LVM or iSCSI
+omit_dracutmodules+=" nfs lvm iscsi "
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -17,18 +17,6 @@ ostree-layers:
   - overlay/14NetworkManager-plugins
   - overlay/20platform-chrony
 
-initramfs-args:
-  - --no-hostonly
-  # We don't support root on NFS, so we don't need it in the initramfs. It also
-  # conflicts with /var mount support in ignition because NFS tries to mount stuff
-  # in /var/ and then ignition can't cleanly unmount it. For example:
-  # https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfs-start-rpc.sh#L7
-  # See also discussion in https://github.com/coreos/fedora-coreos-config/pull/60
-  - --omit=nfs
-  # Omit these since we don't use them
-  - --omit=lvm
-  - --omit=iscsi
-
 # Be minimal
 recommends: false
 
--- a/manifests/ignition-and-ostree.yaml
+++ b/manifests/ignition-and-ostree.yaml
@@ -8,10 +8,6 @@
 # Include rpm-ostree + kernel + bootloader
 include: bootable-rpm-ostree.yaml
 
-initramfs-args:
-  # make it a hard error if Ignition can't be included
-  - --add=ignition
-
 # Modern defaults we want
 boot-location: modules
 tmp-is-dir: true
--- a/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-nohostonly.conf
+++ b/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-nohostonly.conf
@@ -1,0 +1,2 @@
+# Default rpm-ostree model is server-side generated initramfs
+hostonly=no


jmarrero (dismissed):
lgtm

overlay.d/05core/usr/lib/dracut/dracut.conf.d/ignition.conf
```diff
@@ -0,0 +1 @@
+add_dracutmodules+=" ignition "
```
jlebon: Aside: hmm, thinking this should be shipped as part of Ignition instead.

jlebon (dismissed):
Just to make sure, did you sanity-check locally that this works as expected? At least the `ignition` module would've been installed anyway as a dependency of other dracut modules we ship. Would be nice if dracut actually printed the dropins it read in.

overlay.d/05core/usr/lib/dracut/dracut.conf.d/ignition.conf
cgwalters: One thing I discovered that *astounded* me is that there are some (traditional) RHEL customers who literally do `yum -y install '*'` - i.e. install *all* packages.

We don't want to have "package installed" = "system enabled" for some things specifically like ignition.

Not enabling the ostree dracut module simply because ostree was installed is also why we have rpm-ostree force it on today only for the systems it builds.

But...it would actually probably make sense to deprecate *that* too and have `add_dracutmodules+=" ostree "` in this repository too.

jlebon: > One thing I discovered that _astounded_ me is that there are some (traditional) RHEL customers who literally do `yum -y install '*'` - i.e. install _all_ packages.

Yuck! Not sure how much sympathy I have there. systemd presets means that lots of services would still be enabled by default upon install.

> We don't want to have "package installed" = "system enabled" for some things specifically like ignition.

Are you thinking of `ignition-validate`? The primary `ignition` subpackage really only has the dracut module. (Edit: there's also `ignition-apply` now, which could've made sense to want on its own, but it isn't a user-facing API.)

cgwalters:
> Just to make sure, did you sanity-check locally that this works as expected? 

Yep!  Also CI covers everything we care about for this, AFAIK.  (Right?)

> At least the ignition module would've been installed anyway as a dependency of other dracut modules we ship.

Yep, I dropped the bit to enable it directly, and I see it still being enabled as expected since we have
```
depends() {
    echo ignition rdcore
}
```
in `40ignition-ostree`.

>  Would be nice if dracut actually printed the dropins it read in.

Well, I did mess up things by not having a leading space, and dracut *does* warn about that.  (I love how awesome the trap there is from "let's just make our config files executable shell script")

cgwalters:
BTW as a followup we can add a change to add a lot more explicit excludes to quiet this error spew:
```
dracut: dracut module 'systemd-networkd' will not be installed, because command 'networkctl' could not be found!
dracut: dracut module 'systemd-networkd' will not be installed, because command '/usr/lib/systemd/systemd-networkd' could not be found!
dracut: dracut module 'systemd-networkd' will not be installed, because command '/usr/lib/systemd/systemd-networkd-wait-online' could not be found!
dracut: dracut module 'busybox' will not be installed, because command 'busybox' could not be found!
dracut: dracut module 'dbus-daemon' will not be installed, because command 'dbus-daemon' could not be found!
dracut: dracut module 'rngd' will not be installed, because command 'rngd' could not be found!
dracut: dracut module 'network-legacy' will not be installed, because command 'dhclient' could not be found!
dracut: dracut module 'network-wicked' will not be installed, because command 'wicked' could not be found!
dracut: 62bluetooth: Could not find any command of '/usr/lib/bluetooth/bluetoothd /usr/libexec/bluetooth/bluetoothd'!
dracut: dracut module 'dmraid' will not be installed, because command 'dmraid' could not be found!
dracut: dracut module 'pcsc' will not be installed, because command 'pcscd' could not be found!
dracut: dracut module 'fcoe' will not be installed, because command 'dcbtool' could not be found!
dracut: dracut module 'fcoe' will not be installed, because command 'fipvlan' could not be found!
dracut: dracut module 'fcoe' will not be installed, because command 'lldpad' could not be found!
dracut: dracut module 'fcoe' will not be installed, because command 'fcoemon' could not be found!
dracut: dracut module 'fcoe' will not be installed, because command 'fcoeadm' could not be found!
dracut: dracut module 'fcoe-uefi' will not be installed, because command 'dcbtool' could not be found!
dracut: dracut module 'fcoe-uefi' will not be installed, because command 'fipvlan' could not be found!
dracut: dracut module 'fcoe-uefi' will not be installed, because command 'lldpad' could not be found!
dracut: dracut module 'nbd' will not be installed, because command 'nbd-client' could not be found!
dracut: dracut module 'biosdevname' will not be installed, because command 'biosdevname' could not be found!
dracut: dracut module 'memstrack' will not be installed, because command 'memstrack' could not be found!
```

overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
```diff
@@ -0,0 +1,7 @@
+# We don't support root on NFS, so we don't need it in the initramfs. It also
```
jlebon, line 1: Worth noting these dropins are going to be inherited by RHCOS, while previously the manifest bits in `fedora-coreos-base.yaml` were FCOS-specific. We do want those for RHCOS too, but strangely enough it looks like in https://github.com/openshift/os/blob/01f98943688b5b61a9276127a1b7f06aafd0b42a/common.yaml#L32-L53 we're explicitly pulling in `iscsi` in RHCOS but explicitly omitting it here.

I think dracut will give priority to the cmdline arg in RHCOS, but worth digging into the history there. And of course we should delete that hunk in favour of `dracut.conf.d` there too.

jlebon (approved):
LGTM!

PR #1829, opened 2022-07-07
35coreos-multipath: enter emergency.target if multipathd bootfs times out

jlebon:
This should be the default in almost all our initramfs systemd units.
Otherwise we'll keep limping on and other spurious errors might mislead
users.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
@@ -8,6 +8,9 @@ After=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
 Requires=multipathd.service
 After=multipathd.service
 
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
 # This is already enforced transitively by coreos-gpt-setup.service, but
 # let's be more explicit and list it directly here too.
 Before=coreos-ignition-setup-user.service


bgilbert:
Do we know that `OnFailure` currently works correctly in a `.target` unit?  We've had problems with that in the past.

jlebon:
> Do we know that `OnFailure` currently works correctly in a `.target` unit? We've had problems with that in the past.

Tested this. Can confirm it works correctly (at least in FCOS; it can't make things worse in RHCOS8). I think that's been fixed for a while now. Ahh yup, see https://github.com/systemd/systemd/issues/14142#issuecomment-675710064.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
```diff
@@ -8,6 +8,9 @@ After=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
 Requires=multipathd.service
 After=multipathd.service
 
+OnFailure=emergency.target
```
dustymabe, line 11: any need for comments/context here?

dustymabe (approved):
LGTM

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
jlebon: We're not consistent about this, but in the majority of the cases where we do this, we don't have a comment explaining it. I guess the underlying assumption is that it's self-explanatory?

dustymabe: WFM

PR #1831, opened 2022-07-08
s390x: drop 's390utils-base'

nikita-dubrovskii:
This package provides the `genprotimg` tool, but also depends on `perl-*` rpms, which we don't want 
to be part of the image. As long as IBM Secure Execution is targeted for RHCOS we can drop its support
in FCOS for now.
RHCOS comes with a lot of `perl-*` rpms, so there `s390utils-base` wouldn't bring new unwanted dependencies. 
For custom FCOS builds it's easier to:
```
$ mkdir -p overrides/rootfs/usr/bin/
$ mkdir -p overrides/rootfs/usr/share/s390-tools/genprotimg
$ cp /path/to/genprotimg overrides/rootfs/usr/bin/
$ cp /path/to/stage3a.bin overrides/rootfs/usr/share/s390-tools/genprotimg/
$ cp /path/to/stage3b_reloc.bin overrides/rootfs/usr/share/s390-tools/genprotimg/
```

https://github.com/coreos/fedora-coreos-tracker/issues/1217

### DIFF
--- a/manifest-lock.s390x.json
+++ b/manifest-lock.s390x.json
@@ -285,9 +285,6 @@
     "fedora-repos-ostree": {
       "evra": "36-1.noarch"
     },
-    "file": {
-      "evra": "5.41-4.fc36.s390x"
-    },
     "file-libs": {
       "evra": "5.41-4.fc36.s390x"
     },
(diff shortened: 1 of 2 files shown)


dustymabe:
LGTM

dustymabe:
cc @jschintag

dustymabe (approved):

PR #1835, opened 2022-07-11
35coreos-multipath: propagate after Ignition `files` stage

jlebon:
We didn't have a strong ordering between the propagation service and the
Ignition `files` stage. We need to run after it so that we only
propagate the multipath config if the Ignition config didn't create one.

In practice currently the propagation service runs before, which didn't
cause any issues because Ignition can still just clobber what we
propagated. But it could become racier in the future, and it's confusing
because the logic we run and messages we log assume the `files` stage
ran.


jlebon:
Related: https://github.com/dracutdevs/dracut/pull/1870

Noticed this stuff while writing https://github.com/openshift/os/pull/885.

saqibali-2k (approved):

PR #1837, opened 2022-07-12
image-base.yaml: add vmware-secure-boot

mike-nguyen:
This will set the secure boot option to true in the vmware OVF through
coreos-assembler.


travier:
You should add it here instead: https://github.com/coreos/coreos-assembler/blob/main/src/image-default.yaml

mike-nguyen:
Closing this base on @travier's comment.

PR #1838, opened 2022-07-12
overlay: Omit many more dracut modules explicitly

cgwalters:
We're not actually shipping any of these today; the way dracut
works is it looks for them and emits a loud warning if they're not
found.

This is purely to quiet build time warnings.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
+++ b/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
@@ -3,5 +3,13 @@
 # in /var/ and then ignition can't cleanly unmount it. For example:
 # https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfs-start-rpc.sh#L7
 # See also discussion in https://github.com/coreos/fedora-coreos-config/pull/60
-# Further, we currently do not use LVM or iSCSI
-omit_dracutmodules+=" nfs lvm iscsi "
+# Further, we currently do not use LVM, iSCSI or dmraid
+omit_dracutmodules+=" nfs lvm iscsi dmraid "
+# More storage modules we don't use
+omit_dracutmodules+=" fcoe fcoe-uefi nbd "
+# We use NetworkManager
+omit_dracutmodules+=" systemd-networkd network-legacy network-wicked "
+# We use systemd network naming
+omit_dracutmodules+=" biosdevname "
+# Random stuff we don't want
+omit_dracutmodules+=" rngd busybox dbus-daemon memstrack pcsc "


cgwalters:
Promised cleanup followup after https://github.com/coreos/fedora-coreos-config/pull/1828#issuecomment-1179386645

overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
```diff
@@ -3,5 +3,13 @@
 # in /var/ and then ignition can't cleanly unmount it. For example:
 # https://github.com/dracutdevs/dracut/blob/1856ae95c873a6fe855b3dccd0144f1a96b9e71c/modules.d/95nfs/nfs-start-rpc.sh#L7
 # See also discussion in https://github.com/coreos/fedora-coreos-config/pull/60
-# Further, we currently do not use LVM or iSCSI
-omit_dracutmodules+=" nfs lvm iscsi "
+# Further, we currently do not use LVM, iSCSI or dmraid
+omit_dracutmodules+=" nfs lvm iscsi dmraid "
```
cgwalters, line 7: That said, there is one intersection jlebon pointed out here https://github.com/openshift/os/blob/01f98943688b5b61a9276127a1b7f06aafd0b42a/common.yaml#L41
i.e. rhcos is explicitly adding iscsi.  The rationale for this dates back to before the public copy of openshift/os; I will dig some to see if I can find it.

But perhaps we should move this omission to a separate file that isn't inherited by rhcos out of conservatism.

cgwalters: OTOH, that's not new in this PR.  But it might become an issue if we try to migrate openshift/os away from `initramfs-args`.

So this is fine to merge I think, we can do the `iscsi` thing as a followup.

overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
```diff
@@ -3,5 +3,13 @@
...
+# More storage modules we don't use
+omit_dracutmodules+=" fcoe fcoe-uefi nbd "
+# We use NetworkManager
+omit_dracutmodules+=" systemd-networkd network-legacy network-wicked "
+# We use systemd network naming
+omit_dracutmodules+=" biosdevname "
+# Random stuff we don't want
+omit_dracutmodules+=" rngd busybox dbus-daemon memstrack pcsc "
```
dustymabe, line 15: hmm. I was pretty sure we were using dbus in the initrd now (i.e. NetworkManager being started via systemd) or maybe I was mistaken and we needed `dbus` for something else but it didn't exist.

cgwalters: You're right, but...there are actually *three* dbus modules:

```
$ ls -ld /usr/lib/dracut/modules.d/*dbus*
drwxr-xr-x. 2 root root 29 Jan  1  1970 /usr/lib/dracut/modules.d/06dbus-broker
drwxr-xr-x. 2 root root 29 Jan  1  1970 /usr/lib/dracut/modules.d/06dbus-daemon
drwxr-xr-x. 2 root root 29 Jan  1  1970 /usr/lib/dracut/modules.d/09dbus
```

This is omitting the one of those we're not currently using:

```
[root@cosa-devsh ~]# rpm-ostree status
State: idle
  DriverState: active; periodically polling for updates (last checked Tue 2022-07-19 14:19:03 UTC)
Deployments:
* fedora:fedora/x86_64/coreos/stable
                   Version: 36.20220618.3.1 (2022-07-05T23:09:41Z)
                    Commit: 474557e51b1013d4e737e0fd41f4e3d482546e3615a2480a3b34bb186a8ada94
              GPGSignature: Valid signature by 53DED2CB922D8B8D9E63FD18999F7CBF38AB71F4
[root@cosa-devsh ~]# lsinitrd -m /usr/lib/modules/5.18.5-200.fc36.x86_64/initramfs.img |grep dbus
dbus-broker
dbus
[root@cosa-devsh ~]#
```

cgwalters: (More generally, dbus-daemon has been deprecated by dbus-broker, which is what we are including although I'm not totally sure it's actually used)

dustymabe: ahh didn't know that.

dustymabe (approved):
Trivial LGTM.

I don't think this could break anything but am interested to get other reviews too.

jlebon (approved):
Nice! Output looks much cleaner. I've verified everything omitted here was already being omitted.

I do still see:

> [2022-07-12T21:04:06.316Z] dracut: 62bluetooth: Could not find any command of '/usr/lib/bluetooth/bluetoothd /usr/libexec/bluetooth/bluetoothd'!

which seems like another one we could omit, but cool to do as a followup too.

cgwalters:
> which seems like another one we could omit, but cool to do as a followup too.

OK yeah, let's do followup to save green CI/reviews!

overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-omits.conf
cgwalters: Honestly I didn't either until you asked, so glad you did :smile: 

(edit: I knew dbus-broker deprecated dbus-daemon, but not that there were distinct dracut modules)

jlebon:
> > which seems like another one we could omit, but cool to do as a followup too.
> 
> OK yeah, let's do followup to save green CI/reviews!

Done in https://github.com/coreos/fedora-coreos-config/pull/1855!

PR #1840, opened 2022-07-13
`coreos-printk-quiet.service`: Lower kernel log level to 4

### LINKED ISSUE #1244: Change default to be equivalent of `quiet`
Most other Fedora editions use anaconda, which for a long time has injected the `quiet` kernel command line argument because it has a history of use on desktop systems where we don't want users to see kernel output spew by default.

We do want server administrators to see kernel output by default.

Fedora CoreOS has never injected the `quiet` karg, and as a consequence too this means that for later runtime the kernel is much more verbose.

This is particularly bad in concert with https://github.com/coreos/fedora-coreos-tracker/issues/567
 which causes this verbose output on slow serial consoles.

I think #567 is the primary fix *but* we should also consider at least quieting kmsg *after* switching to the real root successfully.

cgwalters:
Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1244

A lot going on for this simple service.  See the tracker issue above
for more info, but briefly:

Anaconda has historically injected `quiet` into the kernel command
line in many cases, and this suppresses *both* kernel and systemd
output.  On computers in general, but particularly many bare metal
servers, the serial console can be slow.  This can cause reliability
issues.

However for servers, we usually *do* want to see informational
output when they boot.  For example, today the kernel "mitigations"
information for hardware vulnerabilities is output.

This change is a compromise; we boot up at the kernel's default
verbosity level (which for Fedora and derivatives is the upstream 7),
but switch to 4 very early on in the real root.  At that point, we've
gotten most of the boot time output, and our initramfs is not
extremely performance sensitive right now.

Also, we explicitly only *lower* the output log level, and only if
there isn't explicitly `debug` on the kernel command line.

---

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -5,6 +5,8 @@ enable console-login-helper-messages-gensnippet-os-release.service
 enable console-login-helper-messages-gensnippet-ssh-keys.service
 # CA certs (probably to add to base fedora eventually)
 enable coreos-update-ca-trust.service
+# Set kernel console log level
+enable coreos-printk-quiet.service
 # https://github.com/coreos/ignition/issues/1125
 enable coreos-ignition-firstboot-complete.service
 # Delete Ignition config from provider on platforms where it's possible
--- a/overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
+++ b/overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
@@ -1,0 +1,27 @@
+[Unit]
+Description=CoreOS: Set printk To Level 4 (warn)
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1244
+# We can run right after `/proc` being mounted at least
+DefaultDependencies=no
+# We run as early as possible; the only dependency we have really
+# is the implicit After=systemd-journald.socket injected by the
+# default of our stdout writing to the journal.
+Conflicts=shutdown.target
+Before=sysinit.target shutdown.target
+# We want this service to read what we wrote
+Before=systemd-sysctl.service
+# Relatedly, we don't want to override an explicitly specified kernel argument
+ConditionKernelCommandLine=!debug
+ConditionKernelCommandLine=!quiet
+ConditionKernelCommandLine=!loglevel
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+# We need to make /run/sysctl.d if it doesn't exist and also
+# ensure it has a SELinux label that works for systemd-sysctl.service.
+# Then we just generate a sysctl file which is read by systemd-sysctl.service.
+ExecStart=/bin/bash -euo pipefail -c 'mkdir -p /run/sysctl.d && chcon --reference=/etc/sysctl.d /run/sysctl.d && echo "kernel.printk = 4" > /run/sysctl.d/01-coreos-printk.conf'
+
+[Install]
+WantedBy=sysinit.target
--- a/tests/kola/logging/data/commonlib.sh
+++ b/tests/kola/logging/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh
--- a/tests/kola/logging/printk
+++ b/tests/kola/logging/printk
@@ -1,0 +1,15 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+# Verify default console log level; xref https://github.com/coreos/fedora-coreos-tracker/issues/1244
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+printk=$(cat /proc/sys/kernel/printk)
+
+if ! [[ "$printk" =~ ^4 ]]; then
+    fatal "printk: expected console log level 4, found ${printk}"
+fi
+
+ok "Found expected printk value "


cgwalters:
I do think this change though is most likely something that should actually happen Fedora wide.  Upstreaming this module into dracut might be the easiest.

overlay.d/05core/usr/lib/dracut/modules.d/10coreos-printk/coreos-printk-quiet.service
```diff
@@ -0,0 +1,11 @@
...
+Description=CoreOS: Quiet printk after initramfs setup
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1244
+ConditionPathExists=/etc/initrd-release
+After=initrd.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+# Change the kernel to default to warning level
+ExecStart=/bin/sh -c 'echo 4 > /proc/sys/kernel/printk'
```
cgwalters: self-review: We should probably check the *current* log level and avoid raising it.  IOW, if someone explicitly used e.g. `quiet` on the kernel cmdline we shouldn't override it.

lucab:
Out of curiosity, why did you decide against a `sysctl.d` entry in the real rootfs?

cgwalters:
> I do think this change though is most likely something that should actually happen Fedora wide. Upstreaming this module into dracut might be the easiest.

Alternatively...changing https://src.fedoraproject.org/rpms/kernel/blob/rawhide/f/kernel-x86_64-fedora.config#_983 to 4 is likely the better way to do this.

jlebon:
> Out of curiosity, why did you decide against a `sysctl.d` entry in the real rootfs?

Makes sense to me. It's also easier to disable via Ignition that way. And we'll want to do that for the kola devshell code since I think the state machine there relies on those to know e.g. when to SSH. (But also personally for `--devshell-console`, I'd just prefer to see everything.)

overlay.d/05core/usr/lib/dracut/modules.d/10coreos-printk/coreos-printk-quiet.service
miabbott: Could do a `ConditionKernelCommandLine=!quiet` perhaps?

cgwalters:
OK :new: updated

- Now runs very early in the real root, not initramfs
- No-op if we detect kernel cmdline arguments `quiet` or `debug`
- Change logic to always only lower debug output; so if someone e.g. explicitly configures the level to `2`, we don't make things more verbose

cgwalters:
Also now updated to explicitly run `Before=systemd-sysctl.service` so that configuring printk via Ignition dropping a file in `/etc/sysctl.d` should still work.

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
```diff
@@ -0,0 +1,23 @@
...
+Conflicts=shutdown.target
+# Cargo culting this dependency from systemd-sysctl.service which should
+# hopefully ensure /proc is mounted at least.
+After=systemd-modules-load.service
+Before=sysinit.target shutdown.target
+# We don't want to overwrite what a user may have configured via sysctl.
+Before=systemd-sysctl.service
+# Relatedly, we don't want to override an explicitly specified kernel argument
+ConditionKernelCommandLine=!debug
+ConditionKernelCommandLine=!quiet
```
jlebon, line 15: There's also `loglevel`: https://github.com/torvalds/linux/blob/b047602d579b4fb028128a525f056bbdc890e7f0/init/main.c#L249-L266.

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
```diff
@@ -0,0 +1,23 @@
+[Unit]
+Description=CoreOS: Set printk to level 4 (warn)
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1244
+# We can run right after `/proc` being mounted at least
+DefaultDependencies=no
+Conflicts=shutdown.target
+# Cargo culting this dependency from systemd-sysctl.service which should
+# hopefully ensure /proc is mounted at least.
+After=systemd-modules-load.service
```
jlebon: I think we can safely drop this. `/proc` is mounted by systemd on startup. E.g. `systemd.generator(7)` says:

> Generators can however rely on the most basic kernel functionality to be available, as well as mounted /sys/, /proc/, /dev/, /usr/ and /run/ file systems.

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
cgwalters: Good catch!

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
cgwalters: OK yeah, pretty sure you're right.

jlebon (dismissed):
LGTM!

tests/kola/logging/printk.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
bgilbert: kola will call this test `ext.config.logging.printk.sh`, which is probably not what you intended. :shell:

tests/kola/logging/printk.sh
```diff
@@ -0,0 +1,16 @@
...
+# kola: { "exclusive": false }
+
+# Verify fix for https://github.com/coreos/fedora-coreos-tracker/issues/1244
```
bgilbert: Maybe "Verify the kernel console log level"?

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
```diff
@@ -5,6 +5,8 @@ enable console-login-helper-messages-gensnippet-os-release.service
 enable console-login-helper-messages-gensnippet-ssh-keys.service
 # CA certs (probably to add to base fedora eventually)
 enable coreos-update-ca-trust.service
+# Our printk bits
```
bgilbert:
```suggestion
# Set kernel console log level
```

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=CoreOS: Set printk to level 4 (warn)
```
bgilbert: Should be title case.

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
```diff
@@ -0,0 +1,24 @@
...
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1244
```
bgilbert, line 3: Should we have proper FCOS docs for overriding this?

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
```diff
@@ -0,0 +1,24 @@
...
+Before=systemd-sysctl.service
+# Relatedly, we don't want to override an explicitly specified kernel argument
+ConditionKernelCommandLine=!debug
+ConditionKernelCommandLine=!quiet
+ConditionKernelCommandLine=!loglevel
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/bash -euo pipefail -c 'target=4; current=$(cat /proc/sys/kernel/printk | cut -f 1); if test $current -gt $target; then echo "Updating printk to $target"; echo $target > /proc/sys/kernel/printk; fi'
```
bgilbert: Hmm, I was imagining we'd write a file to e.g. `/run/sysctl.d/01-coreos-printk.conf` instead.  Currently, if the user enables verbose logging via systemd-sysctl, we'll turn the verbosity down and then later systemd-sysctl will turn it back up again.

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
cgwalters: Oh wow.  I didn't know that.  Looking at other systemd units, I'm not the only one.

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
cgwalters: From a read of `man sysctl.d` under `CONFIGURATION DIRECTORIES AND PRECEDENCE` it seems like this would rely on the admin injecting a config that is lexicographically after `01-`...which...OK, is certainly likely.

cgwalters:
Addressed all comments!

overlay.d/05core/usr/lib/systemd/system/coreos-printk-quiet.service
cgwalters: Obviously not opposed to documentation but...in the current design we honor all the standard Linux APIs for this in kernel arguments and injecting a sysctl.d file.  The change of *default* is notable but it's close to a bugfix.

I think I'd say let's tackle documenting the printk level when we push to do this Fedora wide?

miabbott (approved):

PR #1841, opened 2022-07-13
manifests: Add postprocess script to remove broken symlinks

### LINKED ISSUE #1782: Make sure that we do not ship broken symlinks
Based on https://github.com/coreos/fedora-coreos-config/pull/1777, make sure that we don't ship any broken symlink in the image by default.

Broken symlinks should either be removed or fixed. The test should also be update to look for all symlinks in `/etc` and `/usr`.

mohelt:
Add postprocess script to remove specific broken symlinks
from /etc/alternatives/ and /usr/lib/.build-id/.

Fix: https://github.com/coreos/fedora-coreos-config/issues/1782

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -91,6 +91,47 @@ postprocess:
     cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
     cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
     rm /tmp/rpm-ostreed.conf
+  # Make sure that we do not ship broken symlinks:
+  # https://github.com/coreos/fedora-coreos-config/issues/1782
+  # Remove known broken symlinks that point to non-existing files or directories:
+  # - Remove `.build-id` for binaries that we remove in other parts of the FCOS manifest
+  # - Remove links to man pages that we remove in FCOS
+  # Man pages are removed in FCOS thus the links in alternatives pointing to those are left there broken.
+  # Docs removal comes from manifests/fedora-coreos.yaml
+  # - systemd-firstboot comes from manifests/ignition-and-ostree.yaml
+  # - systemd-gpt-auto-generator comes from ignition-and-ostree.yaml
+  # - dotlockfile comes from bootable-rpm-ostree.yaml
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+  
+    list_broken_symlinks_folders=(
+      '/etc/alternatives/'
+      '/usr/lib/.build-id/'
+    )
+
+    # It is not possible to remove files from usr after first boot so that is 
+    # why we are removing them in the postprocess scripts here.
+    # The .build-id links are pointing to binaries that we remove in other parts of the FCOS manifest.
+    list_known_removed_folders=(
+      '/usr/bin/dotlockfile'
+      '/usr/bin/systemd-firstboot'
+      '/usr/lib/systemd/system-generators/systemd-gpt-auto-generator'
+      '/usr/share/doc/'
+      '/usr/share/info/'
+      '/usr/share/man/'
+      )
+    for folder in "${list_broken_symlinks_folders[@]}"
+    do
+        find "${folder}" -type l | while read -r file_name; do
+            real_path=$(realpath -m "${file_name}");
+            for element in "${list_known_removed_folders[@]}"; do
+              if [[ ! -e "${real_path}" ]] && [[ "${real_path}" == "${element}"* ]]; then
+                  rm -r "${file_name}"
+              fi
+            done
+        done
+    done
 
 
 remove-from-packages:
--- a/tests/kola/files/validate-symlinks
+++ b/tests/kola/files/validate-symlinks
@@ -1,0 +1,35 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+# Checking if there are no broken symlinks from from /etc/ and /usr/ .
+# https://github.com/coreos/fedora-coreos-config/issues/1782
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# Here is the list of known broken symlinks that need to be investigated further:
+list_broken_symlinks_skip=(
+    '/etc/mtab'
+    '/etc/ssl/'
+    '/etc/swid/swidtags.d/fedoraproject.org'
+    '/etc/xdg/systemd/user'
+    '/usr/lib/.build-id/'
+    '/usr/lib/firmware/brcm/brcmfmac43455-sdio.AW-CM256SM.txt.xz'
+    '/usr/lib/modules/'
+    '/usr/share/rhel/secrets/etc-pki-entitlement'
+    '/usr/share/rhel/secrets/redhat.repo'
+    '/usr/share/rhel/secrets/rhsm'
+)
+find /usr/ /etc/ -type l -not -path "/usr/etc*"| while read -r file_name; do
+    found="false"
+    for search_element in "${list_broken_symlinks_skip[@]}"; do
+        if [[ "${file_name}" == "${search_element}"* || "${file_name}" == "${search_element}" ]]; then
+            found="true"
+            break
+        fi
+    done
+    real_path=$(realpath -m "${file_name}");
+    if [[ ! -e "${real_path}" ]] && [[ "${found}" == "false" ]];  then
+        fatal "Error: ${file_name} symlink to ${real_path} which does not exist"
+    fi
+done


manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,19 @@ postprocess:
     cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
     cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
     rm /tmp/rpm-ostreed.conf
+  - |
```
travier, line 104: Let's add a short description above this one to explain why we're doing that here and why we need it.

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,19 @@ postprocess:
...
+    #!/bin/bash
```
travier: Let's add some safety related options that like to have in all bash scripts (see script above):
```
set -xeuo pipefail
```

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,19 @@ postprocess:
...
+    find /etc/ /usr/ -type l | while read file_name; do
```
travier: White space at the end of lines is sometimes significant in Bash script. Let's try not to add some wen it's not needed:

```suggestion
    find /etc/ /usr/ -type l | while read file_name; do
```

This also applies to the line below

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,19 @@ postprocess:
     cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
     cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
     rm /tmp/rpm-ostreed.conf
+  - |
+    #!/bin/bash
+    find /etc/ /usr/ -type l | while read file_name; do     
+        real_path=$(readlink -f "${file_name}");    
+        if [[ ! -f "${real_path}" ]] && [[ ! -d "${real_path}" ]]; then
+            if [[ $file_name != /etc/xdg/systemd/user* ]] && [[ $file_name != /etc/systemd* ]] && [[ $file_name != /usr/etc/* ]] ; then
```
travier: All variables should be properly quoted to make sure that spaces don't create issues: `"${filename}"`

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,19 @@ postprocess:
...
+                if [[ $file_name == /etc/alternatives/* ]] || [[ $file_name == /usr/lib/.build-id/* ]]; then
```
travier: For this script, let's match only the files that we want to remove and ignore all the others, thus make this condition the only one.

travier (commented):
Good start. Some minor suggestions.

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,23 @@ postprocess:
...
     rm /tmp/rpm-ostreed.conf
+  # Adding a postprocess script to make sure we don't ship 
+  # any broken symlinks, ignoring a few specific folders.
+  # It is not possible to remove files from usr after first boot so that is 
+  # why we are removing them in the postprocess scripts here
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    find /etc/alternatives/ /usr/lib/.build-id/ -type l | while read -r file_name; do
+    echo "${file_name}"
```
travier: Let's keep the indentation here even though we're in-lining the script to keep it readable.

tests/kola/files/validate-symlinks.sh
```diff
@@ -0,0 +1,15 @@
...
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+find /etc/alternatives/ /usr/lib/.build-id/ -type l | while read file_name; do
+real_path=$(readlink -f "${file_name}");    
+    if [[ ! -f "${real_path}" ]] && [[ ! -d "${real_path}" ]] && [[ ( $file_name == /etc/alternatives/* ) || ( $file_name == /usr/lib/.build-id/* ) ]]; then
+        fatal "Error: ${file_name} symlink to ${real_path} which does not exist"
+    fi
+    ok "${file_name} symlink to valid file ${real_path}"
+done
```
travier: Needs an empty line at the end of file.

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,23 @@ postprocess:
...
+    ls -l "${file_name}"
+    real_path=$(readlink -f "${file_name}"); 
+    echo "${real_path}"
+    if [[ ! -d "${real_path}" ]] && [[ ! -f "${real_path}" ]]; then
```
travier: We probably want to check if the link points to something that exists independently of its type so `-e` would be good here: https://tldp.org/LDP/abs/html/fto.html

tests/kola/files/validate-symlinks.sh
```diff
@@ -0,0 +1,15 @@
...
+# from /etc/alternatives and /usr/lib/.build-id/
+# https://github.com/coreos/fedora-coreos-config/issues/1782
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+find /etc/alternatives/ /usr/lib/.build-id/ -type l | while read file_name; do
+real_path=$(readlink -f "${file_name}");    
+    if [[ ! -f "${real_path}" ]] && [[ ! -d "${real_path}" ]] && [[ ( $file_name == /etc/alternatives/* ) || ( $file_name == /usr/lib/.build-id/* ) ]]; then
```
travier: We should not exclude those paths now that we remove them in the post-process script

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,23 @@ postprocess:
     cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
     cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
     rm /tmp/rpm-ostreed.conf
+  # Adding a postprocess script to make sure we don't ship
```
travier:
```suggestion
  # Remove known broken symlinks that point to non-existing files or directories:
  # - Remove `.build-id` for binaries that we remove in other parts of the FCOS manifest
  # - Remove links to man pages that we remove in FCOS
```

travier (changes requested):

travier:
If I try:
```
 #!/usr/bin/env bash
set -xeuo pipefail
find /etc/alternatives/ /usr/lib/.build-id/ -type l | while read -r file_name; do
real_path=$(readlink -f "${file_name}");    
if [[ ! -d "${real_path}" ]] && [[ ! -f "${real_path}" ]]; then  
    rm -r "${file_name}"
fi
done
```
I get:
```
[core@cosa-devsh ~]$ ls -ahl /etc/alternatives/arptables-man
lrwxrwxrwx. 1 root root 38 Jun 20 12:11 /etc/alternatives/arptables-man -> /usr/share/man/man8/arptables-nft.8.gz
[core@cosa-devsh ~]$ readlink -f /etc/alternatives/arptables-man ; echo $?
1
```
Which is the source of your issue.

If you read the readlink man page, you should find something that help here: https://man7.org/linux/man-pages/man1/readlink.1.html

jlebon:
Not to discourage the effort here, but I wouldn't delete broken symlinks without first consulting with the associated package maintainers. Broken symlinks can be by design if they're intended to be copied literally somewhere else where they're not broken (libostree is one such example). Some applications also are tolerant of broken symlinks (see e.g. https://github.com/coreos/ignition/pull/1416#pullrequestreview-1034867908). Many of the broken buildid symlinks at least seem like packaging issues we should report back instead.

travier:
Completely agree that the symlinks should be fixed directly in packages as much as possible. The initial set we're focusing on contains only symlinks that are broken by changes that we make in FCOS:
- The `.build-id` links are pointing to binaries that we remove in other parts of the FCOS manifest
- Man pages are removed in FCOS thus the links in alternatives pointing to those are left there broken.

The first step is to clean this up and add the test with a "large" allowlist to be able to tackle the remaining ones progressively upstream for each package.

We should also double check that all `.build-id` links are pointing to binaries that we remove and that those are not packaging errors.

cgwalters:
> The initial set we're focusing on contains only symlinks that are broken by changes that we make in FCOS:

This makes sense to me!

travier:
The logic for the removal should be as follow:

1. Look for broken symlinks in a few know places:
	- `/etc/alternatives/`
    - `/usr/lib/.build-id/`
2. Make sure that they point to places we know we remove:
    - documentation, man pages, etc. (`/usr/share/docs`, `/usr/share/man`, ...)
    - explicit list of binaries that we remove in other manifests
3. Remove the symlink if those conditions matched, otherwise, leave it here

---

The logic for the test should be as follow:

1. Looks for broken symlinks everywhere
2. Skip symlinks that we currently know are broken
3. Fail if we still find broken symlinks

---

In both cases, you can re-use the logic from the following test as an example for the list of binaries to skip/ignore: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/files/setuid

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,41 @@ postprocess:
...
+  # It is not possible to remove files from usr after first boot so that is 
+  # why we are removing them in the postprocess scripts here
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    list_broken_symlinks_folders=(
+      '/etc/alternatives/'
+      '/usr/lib/.build-id/'
+    )
+    list_known_removed_folders=(
```
travier, line 116: Let's add:
- `/usr/bin/dotlockfile` from https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/bootable-rpm-ostree.yaml

As comments, let's mention that
- docs removal come from `https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/fedora-coreos.yaml`
- `/usr/bin/systemd-firstboot` (not a directory) comes from `https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/ignition-and-ostree.yaml`
- add `/usr/lib/systemd/system-generators/systemd-gpt-auto-generator` from the same file

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,41 @@ postprocess:
...
+      '/usr/bin/systemd-firstboot/'
+      '/usr/share/doc/'
+      '/usr/share/info/'
+      '/usr/share/man/'
+      )
+    for folder in "${list_broken_symlinks_folders[@]}"
+    do
+        find "${folder}" -type l | while read -r file_name; do
+            echo "${file_name}"
```
travier: Probably don't need those echo/ls commands now that we're not debugging this anymore

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,41 @@ postprocess:
...
+            ls -l "${file_name}"
+            real_path=$(realpath -m "${file_name}");
+            echo "${real_path}"
+            for element in "${list_known_removed_folders[@]}"
```
travier:
```suggestion
            for element in "${list_known_removed_folders[@]}"; do
```

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,45 @@
...
+# https://github.com/coreos/fedora-coreos-config/issues/1782
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+list_broken_symlinks_skip=(
+    '/usr/etc/grub2.cfg'
+    '/usr/lib/firmware/brcm/brcmfmac43455-sdio.AW-CM256SM.txt.xz'
+    '/usr/lib/modules/5.18.6-200.fc36.x86_64/build'
```
travier: We can not hard code the kernel version here. This should be more generic. Matching `/usr/lib/modules/` should be enough for now.

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,45 @@
...
+# Checking if there are no broken symlinks from
+# from /etc/alternatives and /usr/lib/.build-id/
+# https://github.com/coreos/fedora-coreos-config/issues/1782
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+list_broken_symlinks_skip=(
+    '/usr/etc/grub2.cfg'
```
travier: the find command below ignores this path

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,45 @@
...
+    '/usr/share/rhel/secrets/etc-pki-entitlement'
+    '/usr/share/rhel/secrets/redhat.repo'
+    '/usr/share/rhel/secrets/rhsm'
+    '/etc/mtab'
+    '/etc/ssl/'
+    '/etc/swid/swidtags.d/fedoraproject.org'
+    '/etc/systemd/system/dnsmasq.service'
+    '/etc/systemd/system/systemd-repart.service'
+    '/etc/xdg/systemd/user'
+    '/usr/local'
```
travier: This was not part of the list before. Any reason to add them now?

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,45 @@
...
+        if [[ "${file_name}" == "${search_element}"* || "${file_name}" == "${search_element}" ]]; then
+            found="true"
+            break
+        fi
+    done
+    real_path=$(realpath -m "${file_name}");
+    if [[ ! -e "${real_path}" ]] && [[ "${found}" == "false" ]];  then
+        fatal "Error: ${file_name} symlink to ${real_path} which does not exist"
+    fi
+    ok "${file_name} symlink to valid file ${real_path}"
```
travier: Let's not log all valid symlinks as there could be a lot and this would take log space for not much use.

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
+  # Remove known broken symlinks that point to non-existing files or directories:
+  # - Remove `.build-id` for binaries that we remove in other parts of the FCOS manifest
+  # - Remove links to man pages that we remove in FCOS
+  # any broken symlinks, ignoring a few specific folders.
+  # It is not possible to remove files from usr after first boot so that is 
+  # why we are removing them in the postprocess scripts here.
+  # The .build-id links are pointing to binaries that we remove in other parts of the FCOS manifest.
+  # Man pages are removed in FCOS thus the links in alternatives pointing to those are left there broken.
+  # Docs removal comes from https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/fedora-coreos.yaml
+  # /usr/bin/systemd-firstboot (not a directory) comes from https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/ignition-and-ostree.yaml
```
travier:
```suggestion
  # - systemd-firstboot comes from manifests/ignition-and-ostree.yaml
```

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
+  # /usr/lib/systemd/system-generators/systemd-gpt-auto-generator comes from https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/ignition-and-ostree.yaml
```
travier:
```suggestion
  # - systemd-gpt-auto-generator comes from ignition-and-ostree.yaml
```

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
+  # /usr/bin/dotlockfile comes from https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/bootable-rpm-ostree.yaml
```
travier:
```suggestion
  # - dotlockfile comes from bootable-rpm-ostree.yaml
```

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
+  # https://github.com/coreos/fedora-coreos-config/issues/1782
+  # Remove known broken symlinks that point to non-existing files or directories:
+  # - Remove `.build-id` for binaries that we remove in other parts of the FCOS manifest
+  # - Remove links to man pages that we remove in FCOS
+  # any broken symlinks, ignoring a few specific folders.
+  # It is not possible to remove files from usr after first boot so that is 
+  # why we are removing them in the postprocess scripts here.
+  # The .build-id links are pointing to binaries that we remove in other parts of the FCOS manifest.
+  # Man pages are removed in FCOS thus the links in alternatives pointing to those are left there broken.
+  # Docs removal comes from https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/fedora-coreos.yaml
```
travier:
```suggestion
  # Docs removal comes from manifests/fedora-coreos.yaml
```

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
     rm /tmp/rpm-ostreed.conf
+  # Make sure that we do not ship broken symlinks:
+  # https://github.com/coreos/fedora-coreos-config/issues/1782
+  # Remove known broken symlinks that point to non-existing files or directories:
+  # - Remove `.build-id` for binaries that we remove in other parts of the FCOS manifest
+  # - Remove links to man pages that we remove in FCOS
+  # any broken symlinks, ignoring a few specific folders.
+  # It is not possible to remove files from usr after first boot so that is 
+  # why we are removing them in the postprocess scripts here.
+  # The .build-id links are pointing to binaries that we remove in other parts of the FCOS manifest.
```
travier: Let's move that right before the list of removed binaries below

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+  
+    list_broken_symlinks_folders=(
+      '/etc/alternatives/'
+      '/usr/lib/.build-id/'
+    )
+    list_known_removed_folders=(
+      '/usr/bin/systemd-firstboot/'
```
travier:
```suggestion
      '/usr/bin/systemd-firstboot'
```

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
+      '/usr/share/doc/'
+      '/usr/share/info/'
+      '/usr/share/man/'
+      '/usr/lib/systemd/system-generators/systemd-gpt-auto-generator'
+      '/usr/bin/dotlockfile'
```
travier: Let's sort this list alphabetically

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+# Checking if there are no broken symlinks from
+# from /etc/alternatives and /usr/lib/.build-id/
```
travier: We're not checking only those folders now

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,42 @@
...
+# https://github.com/coreos/fedora-coreos-config/issues/1782
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
```
travier, line 9: Let's add a comment here that says that this is the list of currently known broken symlinks that need to be investigated

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,42 @@
...
+list_broken_symlinks_skip=(
```
travier, line 11: Let's sort this list alphabetically too

tests/kola/files/validate-symlinks
travier: not resolved yet

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,42 @@
...
+    '/usr/share/rhel/secrets/redhat.repo'
+    '/usr/share/rhel/secrets/rhsm'
+    '/etc/mtab'
+    '/etc/ssl/'
+    '/etc/swid/swidtags.d/fedoraproject.org'
+    '/etc/systemd/system/dnsmasq.service'
+    '/etc/systemd/system/systemd-repart.service'
+    '/etc/xdg/systemd/user'
+    '/usr/local'
+    '/usr/tmp'
```
travier: idem

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,42 @@
...
+    '/usr/lib/.build-id/'
+    '/usr/lib/sysimage/rpm'
```
travier: This one and the two others below works for me, not sure why we need them here.

travier (approved):
LGTM. Could someone else than me take a quick look? @dustymabe or @lucab maybe?

lucab (approved):

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +91,47 @@ postprocess:
...
+      '/usr/share/doc/'
+      '/usr/share/info/'
+      '/usr/share/man/'
+      )
+    for folder in "${list_broken_symlinks_folders[@]}"
+    do
+        find "${folder}" -type l | while read -r file_name; do
+            real_path=$(realpath -m "${file_name}");
+            for element in "${list_known_removed_folders[@]}"; do
+              if [[ ! -e "${real_path}" ]] && [[ "${real_path}" == "${element}"* ]]; then
```
travier, line 129: We should check that before iterating over the loop of know failures here.

tests/kola/files/validate-symlinks
```diff
@@ -0,0 +1,35 @@
...
+find /usr/ /etc/ -type l -not -path "/usr/etc*"| while read -r file_name; do
+    found="false"
+    for search_element in "${list_broken_symlinks_skip[@]}"; do
+        if [[ "${file_name}" == "${search_element}"* || "${file_name}" == "${search_element}" ]]; then
+            found="true"
+            break
+        fi
+    done
+    real_path=$(realpath -m "${file_name}");
+    if [[ ! -e "${real_path}" ]] && [[ "${found}" == "false" ]];  then
```
travier, line 32: Same here

travier:
Giving this a try in https://github.com/coreos/fedora-coreos-config/pull/1938

PR #1843, opened 2022-07-13
manifests: add zstd

### LINKED ISSUE #1251: New Package Request: zstd

None.

```
$ rpm -q --queryformat '%{SIZE}\n' zstd-1.5.2-2.fc36.x86_64
1608976
```

Create initrd images with zstd compression, which would [improve boot time and reduce initrd size](https://github.com/coreos/fedora-coreos-tracker/issues/1247#issuecomment-1179490347).

In general, yes, but not for initrd creation.

No.

No.

8. Is it possible to layer the package onto the base OS as a day 2 operation?  Explain why or why not.

No, `zstd` needs to be available during `cosa build`.

9. In the case of packages providing services and binaries, can the packaging be adjusted to just deliver binaries?

There are no services.

10. Can the tool(s) provided by the package be used to do things we’d rather users not be able to do in FCOS? (e.g. can it be abused as a Turing complete interpreter?)

No.

12. Does the software provided by the package have a history of CVEs?

Yes: CVE-2019-11922, CVE-2021-24031, CVE-2021-24032.

bgilbert:
We'll soon use this to compress our initrds.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1251.

### DIFF
--- a/manifest-lock.aarch64.json
+++ b/manifest-lock.aarch64.json
@@ -1223,6 +1223,9 @@
     },
     "zram-generator": {
       "evra": "1.1.2-1.fc36.aarch64"
+    },
+    "zstd": {
+      "evra": "1.5.2-2.fc36.aarch64"
     }
   },
   "metadata": {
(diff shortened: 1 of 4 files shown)


cgwalters (dismissed):
:shipit:

dustymabe (approved):
LGTM

dustymabe (approved):
LGTM

PR #1845, opened 2022-07-14
denylist: extend snooze for ext.config.binfmt.qemu on rawhide/aarch64

gursewak1997:
See coreos/fedora-coreos-tracker#1241


dustymabe (approved):
LGTM

PR #1847, opened 2022-07-15
test luks with custom clevis configuration

freedge:
we bind the luks decryption key to pcr 0 and 7 using custom options.

This configuration should be relatively common, while still quite insecure
and no guarantee that the decryption will happen after reboot
(eg following a shim update). It is still a pretty common one.
(see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/configuring-automated-unlocking-of-encrypted-volumes-using-policy-based-decryption_security-hardening)

tests https://github.com/coreos/fedora-coreos-tracker/issues/1255

### DIFF
--- a/tests/kola/root-reprovision/luks-pcr/config.ign
+++ b/tests/kola/root-reprovision/luks-pcr/config.ign
@@ -1,0 +1,30 @@
+{
+  "ignition": {
+    "version": "3.2.0"
+  },
+  "storage": {
+    "luks": [
+      {
+        "name": "myluksdev",
+        "device": "/dev/disk/by-partlabel/root",
+        "clevis": {
+          "custom": {
+            "config": "{\"pcr_bank\":\"sha1\",\"pcr_ids\":\"0,7\"}",
+            "needsNetwork": false,
+            "pin": "tpm2"
+          }
+        },
+        "label": "root",
+        "wipeVolume": true
+      }
+    ],
+    "filesystems": [
+      {
+        "device": "/dev/mapper/myluksdev",
+        "format": "xfs",
+        "wipeFilesystem": true,
+        "label": "root"
+      }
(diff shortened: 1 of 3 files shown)


tests/kola/root-reprovision/luks-pcr/config.ign
```diff
@@ -0,0 +1,30 @@
+{
+  "ignition": {
```
bgilbert, line 2: Optional: we've generally been writing new tests as Butane configs (in `config.bu`) for legibility.

tests/kola/root-reprovision/luks-pcr/test.sh
```diff
@@ -0,0 +1,62 @@
...
+fi
+ok "root mounted with prjquota"
+
+# while we're here, sanity-check that boot is mounted by UUID
+if ! systemctl cat boot.mount | grep -q What=/dev/disk/by-uuid; then
+  systemctl cat boot.mount
+  fatal "boot mounted not by UUID"
+fi
+ok "boot mounted by UUID"
+
```
bgilbert, line 40: Is there a straightforward query we can do to confirm that we sealed to the requested PCRs?

bgilbert (commented):
Thanks for adding a test!

tests/kola/root-reprovision/luks-pcr/test.sh
freedge: we can run ```clevis luks list -d /dev/vda4  -s 1``` but even better would be to change the PCR (tpm2_pcrextend ..) and validate we cannot automatically decrypt anymore. To make development easier the test can be made on a data volume and not the root volume. Since all the files will move, I am cancelling this PR, will do another one later.

PR #1848, opened 2022-07-18
manifests: exclude `perl-interpreter`

jlebon:
We were excluding `perl`, but that's just a metapackage. The actual
interpreter is in `perl-interpreter`. Let's add that to the list to make
sure no package can pull it in.

Related: https://github.com/coreos/fedora-coreos-tracker/issues/1217


dustymabe (approved):
LGTM

cgwalters:
Ah yeah, looks like this was a bit ambiguous and open to interpretation.

PR #1852, opened 2022-07-19
test luks with custom clevis configuration

freedge:
we bind the luks decryption key to pcr 0 and 7 using custom options

tests coreos/fedora-coreos-tracker#1255

### DIFF
--- a/tests/kola/var-mount/lukspcr/config.bu
+++ b/tests/kola/var-mount/lukspcr/config.bu
@@ -4,10 +4,8 @@ storage:
   disks:
     - device: /dev/vda
       partitions:
-        - label: var
-          size_mib: 1000
-          start_mib: 5000
-        - label: publicdata
+        - start_mib: 6000
+          label: publicdata
       wipe_table: false
   luks:
     - name: data
@@ -18,10 +16,6 @@ storage:
           pin: tpm2
           config: '{"pcr_bank":"sha1","pcr_ids":"0,7"}'
   filesystems:
-    - device: /dev/disk/by-partlabel/var
-      format: xfs
-      path: /var
-      with_mount_unit: true
     - device: /dev/mapper/data
       format: ext4
       path: /var/publicdata
--- a/tests/kola/var-mount/lukspcr/config.bu
+++ b/tests/kola/var-mount/lukspcr/config.bu
@@ -1,0 +1,29 @@
+variant: fcos
+version: 1.3.0
+storage:
+  disks:
+    - device: /dev/vda
+      partitions:
+        - label: var
+          size_mib: 1000
+          start_mib: 5000
+        - label: publicdata
+      wipe_table: false
+  luks:
+    - name: data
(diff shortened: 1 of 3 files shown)


tests/kola/var-mount/lukspcr/config.bu
```diff
@@ -0,0 +1,29 @@
+variant: fcos
+version: 1.3.0
+storage:
+  disks:
+    - device: /dev/vda
+      partitions:
+        - label: var
+          size_mib: 1000
+          start_mib: 5000
```
jlebon: Is having `/var` as a mountpoint in this test of importance?

tests/kola/var-mount/lukspcr/test.sh
```diff
@@ -0,0 +1,45 @@
...
+
+src=$(findmnt -nvr /var/publicdata -o SOURCE)
+[[ $(realpath "$src") == $(realpath /dev/mapper/data) ]]
+
+# we close the drive
+umount /var/publicdata
+cryptsetup close data
+
+# we validate we can still unlock the drive
+clevis luks unlock -d /dev/vda6 -n data
```
jlebon: Using `/dev/disk/by-partlabel/publicdata` would be preferable to `/dev/vda6`.

jlebon (dismissed):
Thanks a lot for writing a test for this!

jlebon:
Added `hold` label. This requires https://github.com/coreos/ignition/pull/1418 landing in FCOS proper.

tests/kola/var-mount/lukspcr/config.bu
jlebon: Alternatively, would be interesting to also bind `/var` to PCRs, which is likely going to be more common.

freedge: one objective was to umount and mount again, and as the journal is streaming to /var/log it's a pain to use /var or /var/log. And of course having /var mountpoint is useless, it's just copy pasted. Maybe it can be removed just like that

tests/kola/var-mount/lukspcr/test.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# restrict to qemu for now because the primary disk path is platform-dependent
+# kola: {"platforms": "qemu", "architectures": "!s390x"}
```
bgilbert, line 5: Let's explicitly document why s390x is excluded.

bgilbert (commented):
Looks good, thanks!

tests/kola/var-mount/lukspcr/test.sh
cgwalters: Only tangentially related to this, I recently learned about https://github.com/opencryptoki/opencryptoki which abstracts over TPM and the s390x crypto stuff; I think in theory we could use it for things like clevis there.

PR #1855, opened 2022-07-19
overlay: omit bluetooth dracut module

jlebon:
We don't ship Bluetooth support. Should squash:

    dracut: 62bluetooth: Could not find any command of
    '/usr/lib/bluetooth/bluetoothd /usr/libexec/bluetooth/bluetoothd'!

Follow-up to #1838.


jlebon:
(Untested; we can verify that line is no longer showing up in CI logs.)

cgwalters (approved):

cgwalters:
BTW one thing we should bear in mind is the intersection here with layering - anyone who e.g. wants to include a dracut module explicitly will need to remove/override our default config to do so.

PR #1856, opened 2022-07-20
overlay: disable dracut stripping

jlebon:
While looking at the boot space issue[1], I tested whether stripping
binaries saved anything. Currently, dracut tries to strip binaries if
`strip` (or `eu-strip`) are available, which isn't the case in FCOS.

Adding it as a test, it turns out stripping barely saves anything
because we've already split out debug symbols into separate RPMs, and
the remaining symbols don't take much space.

So let's just tell dracut to stop trying to opportunistically strip
anything to be consistent. This then squashes a message emitted about
`strip` not being found.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1247

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-nostrip.conf
+++ b/overlay.d/05core/usr/lib/dracut/dracut.conf.d/fcos-nostrip.conf
@@ -1,0 +1,3 @@
+# We don't ship `strip` or `eu-strip` today, and even if we did, it doesn't
+# save much space. So let's disable it to avoid the error-looking message.
+do_strip=no


dustymabe:
do we want to share this with rhcos.. maybe we rename to file to something more generic? `coreos_nostrip`?

dustymabe (approved):

jlebon:
The file will be shared with RHCOS. The naming I used follows the current naming convention for the other dropins, though I agree we should probably rename them `s/fcos/coreos/`.

dustymabe:
> though I agree we should probably rename them `s/fcos/coreos/`.

any chance you want to do that in this pass?

jlebon:
> > though I agree we should probably rename them `s/fcos/coreos/`.
> 
> any chance you want to do that in this pass?

Will do it separately to avoid wasting good CI. That also gives a better place for Colin to chime in in case there's a reason we used `fcos_` there.

jlebon:
> Will do it separately to avoid wasting good CI.

https://github.com/coreos/fedora-coreos-config/pull/1858

PR #1858, opened 2022-07-20
overlay: rename dracut.conf.d dropins

jlebon:
These will be shared with RHCOS since they're in `05core`. Let's give
them more generic names.

See discussions in #1856.


cgwalters (approved):
Fine by me, but I also don't think it's a problem to have files named `fcos-` honestly.

jlebon:
Yeah agreed this is pretty minor. Though I think it'd be good for optics to keep "fcos" wording out of RHCOS.

PR #1860, opened 2022-07-21
s390x: add custom 'secex-init' for SE case

nikita-dubrovskii:


manifests/bootable-rpm-ostree.yaml
```diff
@@ -28,6 +28,7 @@ packages-s390x:
   # On Fedora, this is provided by s390utils-core. on RHEL, this is for now
   # provided by s390utils-base, but soon will be -core too.
   - /usr/sbin/zipl
+  - /usr/bin/keyctl
```
jlebon, line 32: Note we already ship keyutils even on x86_64, but agreed it's good to declare our explicit dependencies upfront. Maybe worth a comment to that effect?

overlay.d/05core/usr/lib/dracut/modules.d/99secex-init/secex-init
```diff
@@ -0,0 +1,21 @@
...
+
+# load the rootfs and bootfs keys into the kernel keyring
+cat /etc/luks/boot | keyctl padd user secex:boot @us
+cat /etc/luks/root | keyctl padd user secex:root @us
+
+# delete the keys from the initramfs
+rm /etc/luks/root /etc/luks/boot
+
+# exec systemd
+exec /sbin/init
```
jlebon: After https://github.com/coreos/coreos-assembler/pull/3002#discussion_r931545793, this would be

```suggestion
exec -a /init /usr/lib/systemd/systemd
```

manifests/bootable-rpm-ostree.yaml
nikita-dubrovskii: not sure about x86, but on s390x it's not there

nikita-dubrovskii:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/1922

PR #1862, opened 2022-07-22
denylist: extend snooze for ext.config.binfmt.qemu on rawhide/aarch64

dustymabe:
See coreos/fedora-coreos-tracker#1241


dustymabe:
Will skip CI since it won't test this change.

cgwalters (approved):

PR #1864, opened 2022-07-22
[rhcos-4.10] tests: set min memory for chrony dhcp propagation

mike-nguyen:
RHCOS 4.10 builds were failing because the qemu instance was running out
of memory. The test was failing as exit code 125 from the dockerfile
build but with exit code 137 when run manually. It was only running out
of memory downloading the repodata on the Fedora 35 container. The
Fedora 36 container did not have any issues.

Fixes: https://github.com/openshift/os/issues/899

### DIFF
--- a/tests/kola/chrony/dhcp-propagation
+++ b/tests/kola/chrony/dhcp-propagation
@@ -10,7 +10,7 @@ set -xeuo pipefail
 
 # Just run on QEMU; it should work theoretically on cloud platforms, but many
 # of those have platform-specific sources which take precedence.
-# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv" }
+# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv", "minMemory": 2048 }
 
 test_setup() {
 


cgwalters (dismissed):

gursewak1997 (dismissed):

mike-nguyen:
Sorry, didn't test because infra was down.  Fixed memory up to be an int instead of a string.

gursewak1997 (approved):

mike-nguyen:
I'm not sure what's going on with CI. It may be because this is for the `rhcos-4.10` branch.  I tested locally and it passed

```
[coreos-assembler]$ kola run ext.config.chrony.dhcp-propagation -v
2022-07-22T20:54:37Z cli: Started logging at level INFO
2022-07-22T20:54:37Z cli: Started logging at level INFO
⚠️  Skipping kola test pattern "fcos.internet":
  👉 https://github.com/coreos/coreos-assembler/pull/1478
⚠️  Skipping kola test pattern "podman.workflow":
  👉 https://github.com/coreos/coreos-assembler/pull/1478
=== RUN   ext.config.chrony.dhcp-propagation
--- PASS: ext.config.chrony.dhcp-propagation (115.14s)
PASS, output in tmp/kola/qemu-unpriv-2022-07-22-2054-18900

[coreos-assembler]$ head -20 src/config/tests/kola/chrony/dhcp-propagation 
#!/bin/bash

# This script creates two veth interfaces i.e. one for the host machine 
# and other for the container(dnsmasq server). This setup will be helpful
# to verify the DHCP propagation of NTP servers. This will also avoid any 
# regression that might cause in RHCOS or FCOS when the upstream changes 
# come down and obsolete the temporary work (https://github.com/coreos/fedora-coreos-config/pull/412)

set -xeuo pipefail

# Just run on QEMU; it should work theoretically on cloud platforms, but many
# of those have platform-specific sources which take precedence.
# kola: { "tags": "needs-internet", "platforms": "qemu-unpriv", "minMemory": 2048 }

test_setup() {

    # This is needed to run a container with systemd
    setsebool container_manage_cgroup 1

    # create a network namespace
```

PR #1866, opened 2022-07-25
manifests/group: fix ID for "input" group

lucab:
This fixes the static GID for the `input` group, properly using the
Fedora-wide reserved ID `104`.
This group is mostly used as the owner of ephemeral entries under
`/dev`, thus it should be safe to update (upon booting into the new
deployment, devices will get created from scratch with the new GID).

Ref: https://pagure.io/setup/blob/108f62338e57652bb3a37e9028572704f3f28c29/f/uidgid#_113
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1201#issuecomment-1192522908

### DIFF
--- a/manifests/group
+++ b/manifests/group
@@ -38,7 +38,7 @@ kube:x:994:
 chrony:x:992:
 tcpdump:x:72:
 ceph:x:167:
-input:x:995:
+input:x:104:
 systemd-timesync:x:991:
 systemd-network:x:990:
 systemd-resolve:x:989:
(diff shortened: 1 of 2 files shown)


travier:
This is mostly like https://github.com/coreos/fedora-coreos-config/pull/1832. Should be good. Same mixed feelings. Hopefully we don't break someone's workflow.

travier:
LGTM, but needs someone else to double check that we're not going to break things.

jlebon (approved):
Since it's used to create device nodes on boot, I agree this is safe to change.

> LGTM, but needs someone else to double check that we're not going to break things.

I think one way to see this is that given that the group is actively used by the system, it's owned by us. And the GID we're moving it *to* is below `SYS_GID_MIN` so shouldn't be presently used.

PR #1874, opened 2022-07-27
05core: fix permissions of `/etc/sudoers.d/coreos-sudo-group`

### LINKED ISSUE #1872: bad permissions on /etc/sudoers.d/coreos-sudo-group
HI this file is still throwing an exception due to permissions

[manifest.yaml](https://github.com/bgilbert/os/blob/fea5b2a7671695ca2bfe93a9056fbe88b595ed40/manifest.yaml)
``` 
# Make security scanners happy
  # https://bugzilla.redhat.com/show_bug.cgi?id=1981979
  - |
    #!/usr/bin/env bash
    set -xeuo pipefail
    chmod 600 /etc/sudoers.d/*
```

Set it to 0600 but perms expected to 0440

# visudo -c
/etc/sudoers: parsed OK
/etc/sudoers.d/coreos-sudo-group: bad permissions, should be mode 0440

bgilbert:
sudo wants them to be 440, not 600.

Also update the kola test to run `visudo -c` rather than checking the perms ourselves.  This not only checks perms, but also ensures that the files are syntactically valid.

Fixes https://github.com/coreos/fedora-coreos-config/issues/1872.

### DIFF
--- a/overlay.d/05core/statoverride
+++ b/overlay.d/05core/statoverride
@@ -1,6 +1,7 @@
 # Config file for overriding permission bits on overlay files/dirs
 # Format: =<file mode in decimal> <absolute path to a file or directory>
 
-# Some security scanners complain if /etc/sudoers.d files have 0044 mode bits
+# sudo prefers its config files to be mode 440, and some security scanners
+# complain if /etc/sudoers.d files are world-readable.
 # https://bugzilla.redhat.com/show_bug.cgi?id=1981979
-=384 /etc/sudoers.d/coreos-sudo-group
+=288 /etc/sudoers.d/coreos-sudo-group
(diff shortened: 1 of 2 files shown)


lucab:
Doesn't this conflict with the initial bugreport/fix? The existing comment says `Some security scanners complain if /etc/sudoers.d files have 0044 mode bits` so it's unclear to me whether setting the group-read bit is fine or not.

bgilbert:
I wrote that line, and I think it was an extrapolation on my part.  Going back through the BZ and my email, the only concrete objection was to the world-readable bit.

Since it turns out that `sudo` has an opinion here, I'd consider that opinion authoritative.  Security scanners shouldn't be surprised to see the mode bits recommended by upstream.  Notably, the original BZ didn't object to the mode of `/etc/sudoers`, which is also 440.

lucab:
Thanks for the clarification. Then let's adjust the comment too so that we won't incur future confusion. Other than, it looks good.

bgilbert:
Whoops, updated.

lucab (approved):

PR #1877, opened 2022-07-29
modify ext.config.files.setgid test as per upstream change to liblockfile

### LINKED ISSUE #1253: s390x: fcos.filesystem/sgid fails because of liblockfile
**Describe the bug**
```
15:40:05      --- PASS: fcos.filesystem/suid (0.35s)
15:40:05  
15:40:05      --- FAIL: fcos.filesystem/sgid (0.24s)
15:40:05  
15:40:05              files.go:69: Unknown SUID or SGID files found: [/usr/bin/dotlockfile]
```
**Reproduction steps**
1.  Build fcos rawhide on s390x
2. Run kola
3. 

**Additional information**

Initially it was thought this came from adding s390utils-base, but reverting to s390utils-core again did not change the liblockfile dependency. We need to investigate where the dependency was added and if it was intentional.

[ext.config.files.setgid.txt](https://github.com/coreos/fedora-coreos-tracker/files/9102768/ext.config.files.setgid.txt)

LakshmiRavichandran1:
fix: https://github.com/coreos/fedora-coreos-tracker/issues/1253

- modify `ext.config.files.setgid` test as per upstream change to liblockfile https://github.com/miquels/liblockfile/blob/9389848f70d5b1f242ffffe1ebf7865011f9026f/Makefile.in#L80-L83 which tries too set a special groupId.

- since `fcos.filesystem` test is deleted by https://github.com/coreos/coreos-assembler/pull/2986 thereby removing it from the denylist

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -48,15 +48,3 @@
   - aarch64
   streams:
   - rawhide
-- pattern: ext.config.files.setgid
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1253
-  arches:
-  - s390x
-  streams:
-  - rawhide
-- pattern: fcos.filesystem
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1253
-  arches:
-  - s390x
-  streams:
-  - rawhide
--- a/tests/kola/files/setgid
+++ b/tests/kola/files/setgid
@@ -9,6 +9,7 @@ list_setgid_files=(
     '/usr/bin/write'
     '/usr/libexec/openssh/ssh-keysign'
     '/usr/libexec/utempter/utempter'
+    '/usr/bin/dotlockfile'
 )
 
 unknown_setgid_files=""


dustymabe (approved):
LGTM

lucab:
> modify ext.config.files.setgid test as per upstream change to liblockfile

This part does not look good to me, see https://github.com/coreos/fedora-coreos-tracker/issues/1253#issuecomment-1195416915.
How did you reach the conclusion that we'd need to carry a set-gid `/usr/bin/dotlockfile` in FCOS? For which functionality is that required?

dustymabe:
IIUC she hadn't received any feedback from upstream or the package maintainers so I told her to go ahead and modify the test. I'm happy to revert if we get some info back from the package maintainers but I'd really prefer not to change `/usr/bin/dotlockfile` in postprocess. i.e. if the package doesn't change then there's nothing we *should* do about it.

@LakshmiRavichandran1 - if there isn't one already created maybe we should create a BugZilla against the s390utils package to request the package split so we can document the conversation in public there.

lucab:
To my understanding, we are not waiting for feedback from anyone.
We'd need to 1) have a look s390utils to see whether they actually use this mechanism of privilege-elevation to `mail` (sounds surprising), and 2) if not, trying to re-arrange the packaging to avoid the useless (if confirmed so) set-gid binary.

LakshmiRavichandran1:
requesting package split - https://bugzilla.redhat.com/show_bug.cgi?id=2112857

travier:
This is not OK. We can not add random SetUID/SetGID files in FCOS without carefully considering the security implication.

travier:
The dependency has been added in https://src.fedoraproject.org/rpms/s390utils/c/6f524eab5a2960a72983d87c446ea857c933490d?branch=rawhide

travier:
This one introduced the binary https://src.fedoraproject.org/rpms/liblockfile/c/1a50b71d6a2ac317205263823b48f9df9b400eba?branch=rawhide

travier:
https://github.com/coreos/fedora-coreos-config/pull/1887

PR #1879, opened 2022-07-29
05core: create top-level user symlinks on first boot

jlebon:
With this, users can specify an Ignition config like

```yaml
variant: fcos
version: 1.4.0
storage:
  files:
    - path: /etc/ostree/config.d/01-foobar.conf
      contents:
        inline: |
          [toplevel-links]
          foobar=/var/foobar
```

and on first boot, a `/foobar` symlink will be created.

Requires: https://github.com/ostreedev/ostree/pull/2681
Related: https://github.com/coreos/rpm-ostree/issues/337


jlebon:
> With this, users can specify an Ignition config like

You can see a more complete demo of this in https://github.com/ostreedev/ostree/pull/2681.

bgilbert:
Should we have Butane sugar for the `[toplevel-links]` section?

jlebon:
https://github.com/ostreedev/ostree/pull/2681#issuecomment-2138106566

PR #1881, opened 2022-08-01
tests: var-mount.scsi-id and default-network-behavior-change add support for scos

HuijingHei:
Add scos support for var-mount.scsi-id

```
[coreos-assembler]$ kola run ext.config.shared.var-mount.scsi-id
⚠️  Skipping kola test pattern "coreos.ignition.journald-log":
  👉 https://github.com/coreos/coreos-assembler/issues/1173
=== RUN   ext.config.shared.var-mount.scsi-id
--- PASS: ext.config.shared.var-mount.scsi-id (76.63s)
PASS, output in tmp/kola/qemu-unpriv-2022-08-01-0731-187
```

Also add scos support for tests/default-network-behavior-change

### DIFF
--- a/tests/kola/networking/default-network-behavior-change/test.sh
+++ b/tests/kola/networking/default-network-behavior-change/test.sh
@@ -93,7 +93,7 @@ method=auto
 [user]
 org.freedesktop.NetworkManager.origin=nm-initrd-generator"
 # EXPECTED_INITRD_NETWORK_CFG4
-#   - used on Fedora 37+
+#   - used on Fedora 37+ and scos
 EXPECTED_INITRD_NETWORK_CFG4="# Created by nm-initrd-generator
 
 [connection]
@@ -170,7 +170,7 @@ method=auto
 [.nmmeta]
 nm-generated=true"
 # EXPECTED_REALROOT_NETWORK_CFG3:
-#   - used on all Fedora 37+
+#   - used on all Fedora 37+ and scos
 EXPECTED_REALROOT_NETWORK_CFG3="[connection]
 id=Wired connection 1
 uuid=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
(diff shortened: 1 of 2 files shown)


lucab (dismissed):

travier:
Looks weird. Do we have a newer NM in C9S compared to F37? WDYT @dustymabe

HuijingHei:
C9S is using `NetworkManager-1.39.10-1.el9.x86_64` from http://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/

tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -231,6 +231,9 @@ elif [ "$ID" == "rhcos" ]; then
     else
         fatal "fail: not operating on expected OS version"
     fi
+elif [ "$ID" == "scos" ]; then
```
cgwalters: I know this problem isn't new here but my opinion here is that we should really stop testing distribution IDs like this.

In particular, in many cases what we really want to check is e.g. RHEL >= 9.  

(Almost every single place we reference `rhcos` is generally broken by the creation of RHCOS9 today)

One alternative approach here is to check for `ID_LIKE=rhel` - that will cover both rhel and centos stream.

cgwalters (dismissed):
LGTM as is but I would like to see a later cleanup here that tests for e.g. `ID_LIKE=rhel` not `ID`.

tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -93,7 +93,7 @@ method=auto
 [user]
 org.freedesktop.NetworkManager.origin=nm-initrd-generator"
 # EXPECTED_INITRD_NETWORK_CFG4
-#   - used on Fedora 37+
+#   - used on Fedora 37+ and scos
```
dustymabe, line 96: how about adding a similar comment to line 174?

tests/kola/networking/default-network-behavior-change/test.sh
dustymabe: yeah. I'd prefer not to have to check CentOS and RHEL independently and do something like what you suggest to essentially collapse the logic for the two into one.

dustymabe:
> Looks weird. Do we have a newer NM in C9S compared to F37? WDYT @dustymabe

looks like it: https://bodhi.fedoraproject.org/updates/?search=&packages=NetworkManager&releases=F37

lucab (approved):

PR #1884, opened 2022-08-02
05core: add `Before=` to more Dracut services, and add CI check

bgilbert:
If initrd services don't sequence themselves `Before` anything, it's possible that they won't be started before `initrd-cleanup.service` isolates to `initrd-switch-root.target` and kills them off.  Add at least `Before=initrd.target` to avoid this, plus a CI check to prevent a reoccurrence.

### DIFF
--- a/ci/validate
+++ b/ci/validate
@@ -6,12 +6,36 @@ import stat
 import subprocess
 import yaml
 
+INITRD_SERVICES_WITHOUT_BEFORE = {
+    # Depended on by other services
+    'coreos-livepxe-rootfs.service',
+}
+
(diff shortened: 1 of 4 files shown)


bgilbert:
Ready for review!

travier (approved):
Quick LGTM

cgwalters (approved):

cgwalters:
(Why'd you re-mark it as draft?  Did you find a bug?)

bgilbert:
I was thinking of bolting something else onto this PR, but let's do that as a followup.

PR #1885, opened 2022-08-02
tests/files: checks for FCOS system users

lucab:
This performs extensive checks on default FCOS system users.


cgwalters (approved):
Didn't check in detail, but looks sane to me!

PR #1886, opened 2022-08-02
manifests/passwd: align 'nologin' shell path with systemd defaults

lucab:
This updates all relevant shell fields to use `/usr/sbin/nologin`.
After the `/usr` hierarchy merge that happened in Fedora 17, the
`nologin` binary is now effectively located under `/usr/sbin/`.
For this reason, `systemd-sysusers` directly uses the latter path
as the default shell for system users.
While semantically equivalent, updating all entries to the new
location ensures that defaults are matching across all components.
In turn, that allows packages to avoid specifying a custom `nologin`
shell in sysusers.d fragment, and just rely on systemd defaults.

Refs:
 * https://fedoraproject.org/wiki/Features/UsrMove
 * https://github.com/systemd/systemd/blob/v251/meson.build#L633
 * https://pagure.io/setup/pull-request/40

### DIFF
--- a/manifests/passwd
+++ b/manifests/passwd
@@ -1,33 +1,33 @@
 root:x:0:0:root:/root:/bin/bash
-bin:x:1:1:bin:/bin:/sbin/nologin
-daemon:x:2:2:daemon:/sbin:/sbin/nologin
-adm:x:3:4:adm:/var/adm:/sbin/nologin
-lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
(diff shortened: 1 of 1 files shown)


travier (approved):

lucab:
@dustymabe just highlighting that we merged this changes to the `passwd` file, and I'm not sure if this will cause weird issues similar to https://github.com/coreos/fedora-coreos-tracker/issues/1250. In case, let me know and update the ticket there with any new findings.

PR #1887, opened 2022-08-02
s390x: Remove SetGID dotlockfile from liblockfile

travier:
s390x: Remove SetGID dotlockfile from liblockfile

Adding a new SetUID/SetGID binary to FCOS must come with a security
review as well as a detailed and explicit purpose.

Remove the offending binaries until we confirm that this is really
needed.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1253
See: https://bugzilla.redhat.com/show_bug.cgi?id=2112857
Fixes: 3505072d modify ext.config.files.setgid test as per upstream change to liblockfile ; fcos.filesystem test is deleted by #2986

### DIFF
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -33,3 +33,10 @@ packages-x86_64:
   - microcode_ctl
   # firmware updates
   - fwupd
+
+# See: https://github.com/coreos/fedora-coreos-tracker/issues/1253
+#      https://bugzilla.redhat.com/show_bug.cgi?id=2112857
+# Temporary workaround to remove the SetGID binary from liblockfile that is
+# pulled by the s390utils but not needed for /usr/sbin/zipl.
+remove-from-packages:
+  - ["liblockfile", "/usr/bin/dotlockfile"]
--- a/tests/kola/files/setgid
+++ b/tests/kola/files/setgid
@@ -9,7 +9,6 @@ list_setgid_files=(
     '/usr/bin/write'
     '/usr/libexec/openssh/ssh-keysign'
     '/usr/libexec/utempter/utempter'
-    '/usr/bin/dotlockfile'
 )
 
 unknown_setgid_files=""


cgwalters (dismissed):

dustymabe:
> Adding a new SetUID/SetGID binary to FCOS must come with a security
> review as well as a detailed and explicit purpose.

Hmm. I think we *can* make policy decisions but I don't think we should be in the business of dictating individualized (i.e. FCOS or RHCOS only) policy for dependent packages here. If FCOS is Fedora (and RHCOS is RHEL) then why is this binary allowed there and not here? If it legitimately applies there and not here then that is fine, but if not then we need to take a uniform approach and follow whatever the distro is doing IMO. 

This is me pushing back and saying that we need to get the dependent packages fixed if this is a true policy violation as it equally affects Fedora/RHEL.

travier:
Note that this is a temporary solution until we find a better fix (package split, code changes in s390utils, etc.)

dustymabe:
Also note that this is only affecting `rawhide` and no build for s390utils have been done for F36 since February, so we have time before it hits any production s390x boxes, of which there probably aren't many.

travier:
I don't think there is any procedure today in Fedora to ask packagers to review new SUID binaries. I think there is one in RHEL but I'm not 100% sure.

I've added this check here so that we can be notified when this situation comes up and then address it. We're **part** of the Fedora community and we will be even more moving forward with Fedora CoreOS becoming an official edition. We should thus feel comfortable driving changes. They are not imposed on us.

I think that controlling which SetUID/SetGID binaries get into the official editions is a good thing and we should keep doing it.

I also understand that it takes time to do that so please do ping me when such changes come in and I'll work on them and review them.

dustymabe:
> I don't think there is any procedure today in Fedora to ask packagers to review new SUID binaries. I think there is one in RHEL but I'm not 100% sure.
> 
> I've added this check here so that we can be notified when this situation comes up and then address it. We're **part** of the Fedora community and we will be even more moving forward with Fedora CoreOS becoming an official edition. We should thus feel comfortable driving changes. They are not imposed on us.
> 
> I think that controlling which SetUID/SetGID binaries get into the official editions is a good thing and we should keep doing it.

I 100% agree with your statements above. What I disagree with is adding complexity/restrictions to FCOS only. If we take this position (reviewing and ack/nack setgid/setuid files) then I'd like a plan for removing that burden from FCOS (i.e. a plan to promote it to a toplevel Fedora policy). Once it's promoted then FCOS passively picks up whatever Fedora puts out (and FCOS also participates in the toplevel process because we have vested interest). 

That's the only way we keep FCOS sustainable over time. 

> 
> I also understand that it takes time to do that so please do ping me when such changes come in and I'll work on them and review them.

Thank you for offering.

cgwalters:
FWIW there *is* such a process for RHEL that is currently implemented in Errata Tool - adding setuid binaries requires approval from the security team.  The same is true actually of adding new polkit policies.  I think such a thing should clearly be moved to CentOS Stream and Fedora (or at least, the subset of Fedora that is in CentOS Stream too).

travier:
I've made https://github.com/coreos/fedora-coreos-tracker/issues/1270 and we are tracking investigation and further work in https://github.com/coreos/fedora-coreos-tracker/issues/1253 so that this PRs only stays here as a temporary workaround.

dustymabe (dismissed):
LGTM

manifests/bootable-rpm-ostree.yaml
```diff
@@ -33,3 +33,8 @@ packages-x86_64:
   - microcode_ctl
   # firmware updates
   - fwupd
+
+# Remove SetGID files from liblockfile that is pulled by the s390utils package
+# for /usr/sbin/zipl
```
jlebon: Can we add links here to the tracker issue and bugzilla?

jlebon: OK, I see this was stamped already and has green CI. Fine to do in a follow-up if you prefer! But `remove-from-packages` should be used sparingly for some of the reasons mentioned above, so it's good to signal more strongly that it should be temporary and add links with more justification.

travier:
Updated to address comments

lucab (approved):

LakshmiRavichandran1:
Looks like this change causes the rawhide/s390x CI runs to fail in the post script phase?

```
...
15:05:45  Running post scripts...done
15:05:46  [0m[31merror: [0mWhile applying overrides for pkg liblockfile: fstatat(usr/bin/dotlockfile): No such file or directory
...
15:05:46  error: failed to execute cmd-build: exit status 1
```
https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/build-arch/10/consoleFull

lucab:
It looks like this is hitting an unexpected rpm-ostree bug, see https://github.com/coreos/rpm-ostree/issues/3918. For the moment, it would be better to move this removal to a postprocessing step.

PR #1888, opened 2022-08-02
[rawhide] overrides: pin cryptsetup

saqibali-2k:
`coreos.boot-mirror.luks` is failing in rawhide due to `cryptsetup`.

related: https://github.com/coreos/fedora-coreos-tracker/issues/1268

PR #1890, opened 2022-08-03
[rawhide] Revert "Enable cliwrap (rawhide only for now)"

cgwalters:
This reverts commit 35661f93f2ffc441da4a66f353d5c6b460c77aa8.  We
will work to gain more consensus before enabling this by default
anywhere.

Having it merged for a while was useful to shake out issues, of which we only found one (the kdump one), but I don't think we'll get any more real testing just in rawhide.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -8,6 +8,3 @@ repos:
   - fedora-rawhide
 
 include: manifests/fedora-coreos.yaml
-
-# Only enabled on rawhide right now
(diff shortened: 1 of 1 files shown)


jlebon (approved):
CI hitting the same thing as https://github.com/coreos/fedora-coreos-config/pull/1888. Let's try to get that fixed, but this should be pretty safe to just get in anyway.

PR #1891, opened 2022-08-03
[rhcos-4.11] Revert "05core: fix coreos-boot-edit.service race with switch-root"

mike-nguyen:
This reverts commit 539f342b78478360c1939c2547a343107bb79849.

The test change and the race condition should be split so we can
independently evaluate the changes and avoid unnecessary changes near
release time.


mike-nguyen:
Some discussion on whether we need this or not is ongoing in https://github.com/openshift/os/pull/933

mike-nguyen:
Lets move forward with this

mike-nguyen:
Closing this as we will pick up both in 4.11.1

PR #1896, opened 2022-08-05
tests/kola/root-reprovision: add `reprovision` tag

jlebon:
This can be used to conveniently skip those tests by tag.

### DIFF
--- a/tests/kola/root-reprovision/swap-before-root/test.sh
+++ b/tests/kola/root-reprovision/swap-before-root/test.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
-# kola: { "distros": "fcos", "platforms": "qemu", "minMemory": 4096, "timeoutMin": 15, "allowConfigWarnings": true }
+# kola: { "distros": "fcos", "platforms": "qemu", "minMemory": 4096, "timeoutMin": 15, "allowConfigWarnings": true, "tags": "reprovision" }
 #
 # - distros: fcos
 #   - This test only runs on FCOS due to a problem enabling a swap partition on
 #     RHCOS. See: https://github.com/openshift/os/issues/665
 # - platforms: qemu
 #   - This test should pass everywhere if it passes anywhere.
+# - tags: reprovision
+#   - This test reprovisions the rootfs.
 # - minMemory: 4096
 #   - Root reprovisioning requires at least 4GiB of memory.
 # - timeoutMin: 15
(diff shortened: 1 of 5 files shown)


bgilbert (dismissed):
LGTM.  Dusty might want the new option added to the comment blocks.

bgilbert (dismissed):

tests/kola/root-reprovision/swap-before-root/test.sh
```diff
@@ -1,11 +1,13 @@
 #!/bin/bash
-# kola: { "distros": "fcos", "platforms": "qemu", "minMemory": 4096, "timeoutMin": 15, "allowConfigWarnings": true }
+# kola: { "distros": "fcos", "platforms": "qemu", "minMemory": 4096, "timeoutMin": 15, "allowConfigWarnings": true, "tags": "reprovision" }
 #
 # - distros: fcos
 #   - This test only runs on FCOS due to a problem enabling a swap partition on
 #     RHCOS. See: https://github.com/openshift/os/issues/665
 # - platforms: qemu
 #   - This test should pass everywhere if it passes anywhere.
+# - tag: reprovision
```
dustymabe:
```suggestion
# - tags: reprovision
```

and applied to all the other places?

jlebon: The reason I did `tag` was because multiple tags can be specified. This way makes it clear it's the rationale for the `reprovision` tag only. Each tag would have their own entry.

dustymabe: we don't currently have any examples of tests with multiple tags, so this is up for debate. If we had a test with multiple tags I would do something like this:

```
- tags: "reprovision needs-internet"
  - "needs-internet": We fetch resources from the net
  - "reprovision": This test reprovisions the rootfs
```

Other instances that define tags in the comments use `tags` and not `tag`.

dustymabe: I'll leave it up to you.

dustymabe (dismissed):
LGTM

tests/kola/root-reprovision/swap-before-root/test.sh
jlebon: > Other instances that define tags in the comments use `tags` and not `tag`.

Ahh indeed. I think keeping it consistent is worth the respin.

dustymabe (approved):

PR #1898, opened 2022-08-05
[rawhide] overrides: pin grub2 and filesystem

gursewak1997:
kola basic tests aren't coming up and the machines aren't getting out of grub.
Also filesystem.posttrans fails with unsupported lua script
Relates to:
https://github.com/coreos/fedora-coreos-tracker/issues/1273

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -9,18 +9,23 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
-  cryptsetup:
-    evr: 2.5.0~rc1-2.fc37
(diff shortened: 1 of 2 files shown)


dustymabe (dismissed):
LGTM

dustymabe:
looks like now `coreos.boot-mirror` is failing consistently

gursewak1997:
The CI passed so should be good to go. Apparently the `coreos.boot-mirror` is still failing inconsistently

dustymabe (approved):
LGTM

jlebon:
Minor note for the future: the two overrides should ideally be done in separate commits because they concern separate issues. That way it's easier to revert just one of them.

PR #1903, opened 2022-08-08
[rawhide] overrides: Drop filesystem pin

gursewak1997:
Partially reverts 9e6227328388915d68cee6d2c7e17eb48647c160
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1273

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -19,8 +19,3 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1268
       type: pin
-  filesystem:
-    evr: 3.16-4.fc37
-    metadata:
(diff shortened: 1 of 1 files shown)


jlebon:
I think we can mark it as `Closes:` instead, no?
Also, can you also include more information in the commit message? E.g. "The rpm-ostree workaround for this has now landed in cosa.".

gursewak1997:
I'll update the commit message but I see that we are still getting `error: Package 'filesystem' has (currently) unsupported <lua> script in '%posttrans'` in the CI build.

jlebon:
Ahh yup, we need a new cosa build. Digging.

dustymabe:
This issue is obviously fixed because we get past the "build ostree" phase. Merging so we can continue to methodically stomp out the rawhide issues we are seeing.

PR #1906, opened 2022-08-09
overrides: fast-track grub2-2.06-47.fc36

dustymabe:
This should fix


jlebon (approved):
Bodhi update has good karma so far. We've just cut releases, so cool to get this in and we can report back any possible new fallout.

dustymabe:
marking as ready since it passes tests and the update is marked for `stable`.

PR #1908, opened 2022-08-09
s390x: Remove SetGID dotlockfile from liblockfile (postprocess)

travier:
Moved from `remove-from-packages` to post-process due to a bug in
rpm-ostree.

See: https://github.com/coreos/rpm-ostree/issues/3918
See: https://github.com/coreos/fedora-coreos-tracker/issues/1253
See: https://bugzilla.redhat.com/show_bug.cgi?id=2112857
Fixes: 5053bc7d s390x: Remove SetGID dotlockfile from liblockfile
Fixes: 3505072d modify ext.config.files.setgid test as per upstream change to liblockfile ; fcos.filesystem test is deleted by #2986

### DIFF
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -34,9 +34,13 @@ packages-x86_64:
   # firmware updates
   - fwupd
 
-# See: https://github.com/coreos/fedora-coreos-tracker/issues/1253
-#      https://bugzilla.redhat.com/show_bug.cgi?id=2112857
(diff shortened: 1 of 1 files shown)


dustymabe:
would be nice if we could get someone to test this one first. Maybe @LakshmiRavichandran1

LakshmiRavichandran1:
cosa build and the kola suite passes fine with the change in 👍  on my s390x localhost.
lgtm

LakshmiRavichandran1 (approved):

PR #1909, opened 2022-08-09
[rawhide] overrides: fast-track grub2-2.06-47.fc37

dustymabe:
This should fix

### DIFF
--- a/manifest-lock.overrides.aarch64.yaml
+++ b/manifest-lock.overrides.aarch64.yaml
@@ -10,22 +10,26 @@
 
 packages:
   grub2-common:
-    evra: 1:2.06-43.fc37.noarch
+    evra: 1:2.06-47.fc37.noarch
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1271
-      type: pin
+      type: fast-track
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-1199601116
   grub2-tools-minimal:
-    evra: 1:2.06-43.fc37.aarch64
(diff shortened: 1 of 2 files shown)


manifest-lock.overrides.aarch64.yaml
```diff
@@ -10,22 +10,26 @@
 
 packages:
   grub2-common:
-    evra: 1:2.06-43.fc37.noarch
+    evra: 1:2.06-47.fc37.noarch
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1271
-      type: pin
+      type: fast-track
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-d0c322d15a
```
jlebon: The link should be https://bodhi.fedoraproject.org/updates/FEDORA-2022-1199601116, but actually it looks like it's already pushed to stable. So maybe let's just try dropping the pin and see what CI says?

dustymabe: fixed the link..

The problem with `rawhide` specifically is that bodhi is spreading misinformation and lies...

I kid.. basically for `rawhide` the bodhi machinery itself doesn't create rawhide composes. Updates get marked "stable" based on the tag it's in. It's currently in `stable`, but there hasn't been a successful pungi rawhide compose since then so we can't pull it from the repos.

In the past I've adopted the "well we'll just wait til it lands", but then 4 or 5 days go by without a successful compose. Better to just fix it up front and graduate-overrides take care of cleanup.

jlebon (approved):

PR #1844, opened 2022-07-14
overlay: compress initrd with `zstd -19` on FCOS and RHCOS/SCOS 9

bgilbert:
On my system, relative to `gzip -9`, it roughly halves decompression time at boot while also [reducing initrd size](https://github.com/coreos/fedora-coreos-tracker/issues/1247#issuecomment-1179490347).

RHCOS 8 can't inherit this because the RHEL 8 kernel doesn't enable `CONFIG_RD_ZSTD`.

For https://github.com/coreos/fedora-coreos-tracker/issues/1247.

PR #1910, opened 2022-08-09
Move ssh-key-dir to shared manifest for EL9

LorbusChris:
The new `shared-el9.yaml` manifest is intended for packages that are
shared between FCOS and RHCOS/SCOS 9+.

Related to https://issues.redhat.com/browse/COS-1589

/cc @bgilbert
/cc @travier

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -9,6 +9,7 @@ include:
   - system-configuration.yaml
   - user-experience.yaml
   - shared-workarounds.yaml
+  - shared-el9.yaml
 
 ostree-layers:
   - overlay/05core
@@ -114,8 +115,6 @@ postprocess:
 packages:
   # Security
   - polkit
-  # SSH
-  - ssh-key-dir
   # Containers
   - systemd-container catatonit
   - fuse-overlayfs slirp4netns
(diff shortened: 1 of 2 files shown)


bgilbert (approved):
LGTM, but I'll defer to @travier.

Once this lands, the bot will incorporate it into https://github.com/openshift/os/pull/944 (or its successor) within 24 hours.  Once _that_ lands, we'll want to land an openshift/os PR to include the shared manifest.

travier:
I don't think it's worth making this change here as I don't expect more packages to be in this case. We can just directly add the package to the c9s specific manifest in openshift/os and we will clean things up after we switch RCHOS to RHEL 9.

travier:
I'm looking for other parts of the manifests that we could share that could make this more interesting.

travier:
And whoops again, I'm saying the reverse of https://github.com/openshift/os/pull/946. Let me think about it.

travier:
We have https://github.com/coreos/fedora-coreos-config/pull/1844 that needs this too so let's do it.

travier (approved):

bgilbert:
https://github.com/openshift/os/pull/948 adds this manifest to EL9 builds.

PR #1912, opened 2022-08-10
manifests/passwd: re-sync user entries with Fedora and sort them

lucab:
This updates users details in order to re-sync with the ones from Fedora packages.
It also sort all users by name, so that it is easier to diff against systemd-sysusers
output.

### DIFF
--- a/manifests/passwd
+++ b/manifests/passwd
@@ -10,11 +10,11 @@ mail:x:8:12:mail:/var/spool/mail:/usr/sbin/nologin
 operator:x:11:0:operator:/root:/usr/sbin/nologin
 games:x:12:100:games:/usr/games:/usr/sbin/nologin
 ftp:x:14:50:FTP User:/var/ftp:/usr/sbin/nologin
-nobody:x:99:99:Nobody:/:/usr/sbin/nologin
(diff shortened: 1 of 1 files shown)


dustymabe:
not sure if this is possible, but.. should we leave some sort of comment in here about how this syncing should be done in the future and also should we do it on some interval (like once a major Fedora release?)

lucab:
In the future I hope that this file completely goes away, or at least it get mostly emptied out.

For the moment this syncing is a manual process, mostly based on other upstream&packaging fixes I've been doing recently. These are the sources for the three updates here:
 * https://src.fedoraproject.org/rpms/tpm2-tss/blob/rawhide/f/tpm2-tss-systemd-sysusers.conf
 * https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.14/bus/sysusers.d/dbus.conf.in#L5
 * https://github.com/systemd/systemd/blob/7238fd51d407043de1479211156e5fb412165416/sysusers.d/basic.conf.in#L13

lucab:
There was one more missing, the GECOS field of the `root` user: https://pagure.io/setup/pull-request/43

dustymabe (approved):
LGTM

jlebon (approved):

PR #1916, opened 2022-08-19
overlay/sysusers: add user entries

lucab:
This completes the previous sysusers.d work, by adding all user entries corresponding to the content of `manifests/passwd`.
As minor tweaks, this also adds a minimal description to each fragment and renames them to accommodate both users and groups entries.


jlebon (approved):
Thanks for adding the header comments!

PR #1920, opened 2022-08-22
update fallback hostname handling; add test

dustymabe:
In Fedora 37+ the fallback hostname provided at compile time to systemd
has been changed back to `localhost`. Let's update our modification of
/etc/os-release to insert DEFAULT_HOSTNAME to be F36 only and also add
a test to verify the fallback hostname is `localhost` everywhere.

### DIFF
--- a/tests/kola/networking/hostname/fallback-hostname/test.sh
+++ b/tests/kola/networking/hostname/fallback-hostname/test.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# kola: { "platforms": "qemu" }
+#
+# Test that the fallback hostname is `localhost`. This test
+# validates that the fallback hostname is set to `localhost`
+# by first disabling NetworkManager from setting the hostname
+# via DHCP or DNS (see config.bu) and then verifying that the
+# hostname is set from the fallback hostname and is `localhost`.
+# https://github.com/coreos/fedora-coreos-tracker/issues/902
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# Use the output of hostnamectl to gather information about how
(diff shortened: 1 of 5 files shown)


tests/kola/networking/hostname/fallback-hostname/test.sh
```diff
@@ -0,0 +1,63 @@
...
+# Test that the fallback hostname is `localhost`. This test
+# validates that the fallback hostname is set to `localhost`
+# by first disabling NetworkManager from setting the hostname
+# via DHCP or DNS (see config.bu) and then verifying that the
+# hostname is set from the fallback hostname and is `localhost`.
+# https://github.com/coreos/fedora-coreos-tracker/issues/902
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+
```
cgwalters, line 15: (Feel free to ignore me but again note if we didn't run *all* the tests everywhere, then we wouldn't need to hardcode "run on qemu to avoid wasting resources running this all the time".  It's certainly not inconceivable that somehow e.g. the actual functionality could somehow be broken on just one cloud (aws/gcp/etc))

cgwalters (approved):
Looks great to me, nice work!

tests/kola/networking/hostname/fallback-hostname/test.sh
dustymabe: > (Feel free to ignore me but again note if we didn't run _all_ the tests everywhere, then we wouldn't need to hardcode "run on qemu to avoid wasting resources running this all the time". It's certainly not inconceivable that somehow e.g. the actual functionality could somehow be broken on just one cloud (aws/gcp/etc))

I think if you want to effect change here you're probably better off opening a ticket and driving concentrated discussion on the topic.

dustymabe:
> Looks great to me, nice work!

Thanks :)

PR #1931, opened 2022-08-26
[rawhide] overrides: pin on popt-1.19~rc1-3.fc37

dustymabe:
With the new popt-1.19~rc1-4.fc38 our systems won't even boot.

See https://github.com/coreos/fedora-coreos-tracker/issues/1283


jlebon (approved):
Looks sane to me, though let's add more details in the tracker about how popt is actually breaking us.

PR #1932, opened 2022-08-26
Add fedora-rawhide-nodebug-kernel repo to fedora-rawhide.repo

dustymabe:
We are choosing to use only nodebug kernels in Fedora CoreOS
for our testing. We've seen too many issues where an issue either
shows up only on the debug kernels OR a debug kernel gives us a
false positive or negative when trying to root cause a failure.
Thus we add the fedora-rawhide-nodebug-kernel repo here so that
we can use it in the rawhide manfest. This uses the nodebug kernel
repo [1].

[1] https://fedoraproject.org/wiki/RawhideKernelNodebug

### DIFF
--- a/fedora-rawhide.repo
+++ b/fedora-rawhide.repo
@@ -14,3 +14,25 @@ type=rpm
 gpgcheck=1
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
 skip_if_unavailable=False
+excludepkgs=kernel kernel-core kernel-modules
+
+# We are choosing to use only nodebug kernels in Fedora CoreOS
+# for our testing. We've seen too many issues where an issue either
+# shows up only on the debug kernels OR a debug kernel gives us a
+# false positive or negative when trying to root cause a failure.
+# Thus we use the nodebug kernel repo [1] and includepkgs=kernel
+# here and excludepkgs=kernel above.
+#
+# [1] https://fedoraproject.org/wiki/RawhideKernelNodebug
+[fedora-rawhide-nodebug-kernel]
+name=nodebug kernels for Rawhide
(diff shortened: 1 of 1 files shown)


dustymabe:
This needs to wait until https://github.com/coreos/fedora-coreos-tracker/issues/1282 is fixed (right now we are relying on the fact that the debug kernels don't show the problem).

fedora-rawhide.repo
```diff
@@ -14,3 +14,25 @@ type=rpm
 gpgcheck=1
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
 skip_if_unavailable=False
+excludepkgs=kernel kernel-core kernel-modules
```
jlebon, line 17: Note rpm-ostree supports repo-pinned packages. So a more structured way to do this is something like:

```yaml
repo-packages:
  - repo: fedora-rawhide-nodebug-kernel
    packages:
      - kernel
      - kernel-core
      - kernel-modules
```

I do appreciate the simplicity of this approach too though.

travier:
+1

jlebon (approved):
Makes sense to me!

fedora-rawhide.repo
dustymabe: Interesting. So that means the packages can only come from that repo?

jlebon: Yup, exactly! We're already using it today in RHCOS for example: https://github.com/openshift/os/blob/dbf289a8c1b3e7743c53741cd37009260a4eeac5/manifest-rhel-8.6.yaml#L156.

dustymabe:
ok we're going to [do a denylist entry](https://github.com/coreos/fedora-coreos-tracker/issues/1282#issuecomment-1231819414) so this is unblocked now.

PR #1934, opened 2022-08-26
denylist: snooze coreos.boot-mirror* tests

dustymabe:
They are failing because of a kernel regression and we are actively
working on getting a fix upstream and backported. Let's snooze these
tests for now.

See https://github.com/coreos/fedora-coreos-tracker/issues/1282


bgilbert (approved):

dustymabe:

PR #1938, opened 2022-08-29
manifest & tests: Test if a symlink is broken first

travier:
When looking for broken symlinks, test first if it is broken before
looking at the list of known broken links to skip.

Fixes: https://github.com/coreos/fedora-coreos-config/pull/1841


dustymabe:
Can we also remove the `set -x` on the postprocess script? It destroys my scrollback in my terminal when doing local builds because it prints out a bazillion lines.

dustymabe (approved):
LGTM

mohelt (approved):
LGTM.

PR #1939, opened 2022-08-30
s390x: add Ignition protection for SE

nikita-dubrovskii:
This is a proof-of-concept for encrypting/decrypting the Ignition config in SE case.

Requires: https://github.com/coreos/coreos-assembler/pull/3055

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
@@ -66,5 +66,17 @@
         "wipeFilesystem": true
       }
     ]
+  },
+  "systemd": {
+    "units": [
+      {
+        "mask": true,
+        "name": "serial-getty@.service"
+      },
+      {
+        "mask": true,
+        "name": "autovt@.service"
+      }
+    ]
   }
 }
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -74,5 +74,17 @@ EOF
         mkdir -p /run/coreos/
         touch /run/coreos/secure-execution
         cp /usr/lib/coreos/01-secex.ign /usr/lib/ignition/base.d/01-secex.ign
+        # Add dropins to disable Ignition logging for all stages
+        stages=("fetch-offline" "fetch" "kargs" "disks" "mount" "files")
+        for s in "${stages[@]}"; do
+            dropin="${UNIT_DIR}/ignition-${s}.service.d"
+            mkdir -p "${dropin}"
+            cat > "${dropin}/10-secex.conf" <<EOF
+[Service]
+Environment=IGNITION_ARGS=-log-to-stdout
+StandardOutput=null
+StandardError=null
+EOF
+        done
     fi
 fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=CoreOS Secex Ignition Config Decryptor
+ConditionPathExists=/etc/initrd-release
+ConditionPathExists=/run/coreos/secure-execution
+DefaultDependencies=false
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+# Run after virtio_blk and before Ignition
+After=coreos-gpt-setup.service
+Before=ignition-fetch-offline.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/coreos-secex-ignition-decrypt
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+set -euo pipefail
+
+disk=/dev/disk/by-id/virtio-ignition_crypted
+conf=/usr/lib/ignition/user.ign
+pkey=/usr/lib/coreos/ignition.asc
+tmpd=
+
+cleanup() {
+    rm -f "${pkey}"
+    if [[ -n "${tmpd}" ]]; then
+        rm -rf "${tmpd}"
+    fi
+}
+
+trap cleanup EXIT
+
+tmpd=$(mktemp -d)
+
+if [ ! -e "${disk}" ]; then
+    echo "Ignition config must be encrypted"
+    exit 1
+fi
+
+gpg --homedir "${tmpd}" --import "${pkey}" && rm "${pkey}"
+gpg --homedir "${tmpd}" --skip-verify --output "${conf}" --decrypt "${disk}"
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -26,6 +26,13 @@ install() {
         sgdisk \
         uname
 
+    # For IBM SecureExecution
+    if [[ $(uname -m) = s390x ]]; then
+        inst_multiple \
+            gpg \
+            gpg-agent
+    fi
+
     inst_simple "$moddir/coreos-diskful-generator" \
         "$systemdutildir/system-generators/coreos-diskful-generator"
 
@@ -76,4 +83,8 @@ install() {
 
     # IBM Secure Execution. Ignition config for reencryption of / and /boot
     inst_simple "$moddir/01-secex.ign" /usr/lib/coreos/01-secex.ign
+    install_ignition_unit "coreos-secex-ignition-decrypt.service"
+    inst_script "$moddir/coreos-secex-ignition-decrypt.sh" \
+        "/usr/sbin/coreos-secex-ignition-decrypt"
+
 }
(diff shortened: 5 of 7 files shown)


cgwalters:
Hmm, and the raw disk image containing the private key will itself be protected?

nikita-dubrovskii:
> Hmm, and the raw disk image containing the private key will itself be protected?

Yes, `private` key gets appended to the Secure Execution `sdboot` image only and never lands to the `rootfs`.

overlay.d/05core/usr/lib/dracut/modules.d/99emergency-shell-setup/emergency-shell.sh
```diff
@@ -70,9 +70,9 @@ EOF
     fi
 }
 
-# in SE case drop config before entering shell
+# in SE case drop everything before entering shell
 if [ -f /run/coreos/secure-execution ]; then
-    rm -f /run/ignition.json
+    rm -f /run/ignition.json /etc/ignition.asc
```
bgilbert: Should also delete `/usr/lib/ignition/user.ign`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
```diff
@@ -66,5 +66,27 @@
         "wipeFilesystem": true
       }
     ]
+  },
+  "systemd": {
+    "units": [
+      {
+        "dropins": [
+          {
+            "contents": "[Unit]\nConditionArchitecture=!s390x\n",
```
jlebon: Instead of this, can't we just `mask` the services?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -26,6 +26,13 @@ install() {
         sgdisk \
         uname
 
+    # For IBM SecureExecution
+    if [[ $(uname -m) = s390x ]]; then
+        inst_multiple \
+            gpg \
+            gpg-agent
```
jlebon, line 33: Hmm, how come we need `gpg-agent`?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=CoreOS Secex Ignition Config Decryptor
+ConditionPathExists=/etc/initrd-release
+ConditionPathExists=/run/coreos/secure-execution
```
jlebon, line 4: Can we add `ConditionPathExists=/dev/disk/by-id/virtio-ignition_crypted` here?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
```diff
@@ -0,0 +1,20 @@
...
+}
+tmpd=$(mktemp -d) && trap cleanup EXIT
+
+if [ ! -e "${disk}" ]; then
+    echo "Ignition config must be encrypted"
+    exit 1
+fi
+
+gpg --homedir "${tmpd}" --import "${pkey}"
+gpg --homedir "${tmpd}" --skip-verify --output "${conf}" --decrypt "${disk}"
```
jlebon, line 26: Shouldn't we delete the private key here?

jlebon (commented):
Did you check if the `rd.break` shells are covered?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
nikita-dubrovskii:
```
gpg: import from '/tmp/ignition.asc' failed: No agent running
```

nikita-dubrovskii:
> Did you check if the `rd.break` shells are covered?

This shell is only available when user adds `rd.break` to kernel args, and that's out of SecEx scope. Or do i miss smth?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
jlebon: This doesn't look resolved. ISTM like as soon as we import the key, we should be able to delete the key, no?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
jlebon: Yuck. In that case, can you confirm that once `coreos-secex-ignition-decrypt.service` finishes, the `gpg-agent` process is killed? I think systemd cgroups semantics will ensure of that, but the agent is known to sometimes hang around longer than one wants.

Once we have https://github.com/coreos/coreos-installer/pull/1033, another option would be to move this code to `rdcore` and shed the `gpg` dep.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.service
jlebon: Hmm actually... sorry to flip on this, but I wonder if it's better to drop the conditional so that the script always runs? If an encrypted Ignition config is found, we decrypt it and then delete the key. If no encrypted Ignition config is found, we still delete the key.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -74,5 +74,17 @@ EOF
         mkdir -p /run/coreos/
         touch /run/coreos/secure-execution
         cp /usr/lib/coreos/01-secex.ign /usr/lib/ignition/base.d/01-secex.ign
+        # Add dropins to disable Ignition logging for all stages
+        stages=("fetch-offline" "fetch" "kargs" "disks" "mount" "files" )
```
jlebon: Minor/optional:

```suggestion
        stages=("fetch-offline" "fetch" "kargs" "disks" "mount" "files")
```

jlebon: This list is subject to get of date. We don't have plans to add new stages, though if we do I can easily imagine forgetting to update this list. I think ideally Ignition supports some API for what we want. One idea is a kernel argument like `ignition.silent` which would be like passing `-no-log` (which doesn't exist yet). Another idea is doing this at build time, where we have a dracut module which runs before `30ignition` and drops some file in `/usr` which `30ignition` then cues off of to e.g. apply an override to all its units like we do here.

Anyway, let's not block on this, but it might be worth opening an Ignition issue (cc @bgilbert).

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+
+disk=/dev/disk/by-id/virtio-ignition_crypted
+conf=/usr/lib/ignition/user.ign
+pkey=/tmp/ignition.asc
```
jlebon: I know this comes from the cosa side, but looking at it from this side, I find `/tmp` to be a weird path for this. Let's put it in e.g. `/usr/lib/coreos/` instead?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
nikita-dubrovskii: `gpg-agent` daemon doesn't run after this unit:
```
Starting ignition-fetch-offline.service - Ignition (fetch-offline)...
[    3.019671] sh[780]: root         778  0.0  0.0   4076  3184 ?        Ss   13:56   0:00 /bin/sh -c ps aux | grep gpg
[    3.019694] sh[780]: root         780  0.0  0.0   2900  1680 ?        S    13:56   0:00 grep gpg
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
nikita-dubrovskii: there is `trap` which removes it, but also added here

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
nikita-dubrovskii: Yep, one minor downside here - empty key (from `cosa build`) than would be part of `/sysroot`. Key shouldn't outlive the firstboot, so i guess `/tmp` looks better + we have the same `/sysroot` as non-SE builds.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.service
nikita-dubrovskii: Anyhow we delete key later before entering emergency shell. But here it'd be better to drop it

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-secex-ignition-decrypt.sh
jlebon: Let's discuss this bit on the cosa side instead.

jlebon (dismissed):
LGTM!

jlebon (approved):

jschintag:
LGTM

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
bgilbert: https://github.com/coreos/ignition/issues/1684

PR #1940, opened 2022-08-30
denylist: extend denial/snooze for coreos.boot-mirror* to rawhide

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/1282#issuecomment-1231819414


dustymabe:
will skip CI because it won't test this change.

gursewak1997 (approved):

PR #1941, opened 2022-08-30
tests/validate-symlinks: QEMU & FCOS only

travier:
Let's resttrict this test to:
- QEMU only as we don't expect the result to be different on another
  platform
- FCOS only until we've tested it for RHCOS & SCOS

### DIFF
--- a/tests/kola/files/validate-symlinks
+++ b/tests/kola/files/validate-symlinks
@@ -1,13 +1,19 @@
 #!/bin/bash
-# kola: { "exclusive": false }
-# Checking if there are no broken symlinks from from /etc/ and /usr/ .
-# https://github.com/coreos/fedora-coreos-config/issues/1782
+# kola: { "exclusive": false, "distros": "fcos", "platforms": "qemu" }
+#
+# Check if there are broken symlinks in /etc/ and /usr/.
+# See: https://github.com/coreos/fedora-coreos-config/issues/1782
+#
+# - distros: fcos
+#   - Only run on FCOS until we've tested it on RHCOS.
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
 
 set -xeuo pipefail
 
 . $KOLA_EXT_DATA/commonlib.sh
 
-# Here is the list of known broken symlinks that need to be investigated further:
+# List of known broken symlinks that require further investigation:
 list_broken_symlinks_skip=(
     '/etc/mtab'
     '/etc/ssl/'
--- a/README.md
+++ b/README.md
@@ -160,16 +160,28 @@ Tests should follow the following format:
 
 ```bash
 #!/bin/bash
-# kola: { "exclusive": false }    <-- kola option comment. See all options in <https://coreos.github.io/coreos-assembler/kola/external-tests/#kolajson>
+# kola: { "exclusive": false, ... }
+# ^^----- kola option comment
+# ^^----- See all options in <https://coreos.github.io/coreos-assembler/kola/external-tests/#kolajson>
+#
 # Short summary of what the test does, why we need it, etc.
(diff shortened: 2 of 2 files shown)


tests/kola/files/validate-symlinks
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
-# kola: { "exclusive": false }
+# kola: { "exclusive": false, "distros": "fcos", "platforms": "qemu-unpriv" }
 # Checking if there are no broken symlinks from from /etc/ and /usr/ .
 # https://github.com/coreos/fedora-coreos-config/issues/1782
+#
+# No need to run on any other platform than QEMU.
+# TODO: Check if this works for RHCOS too.
```
dustymabe: Can we make this conform more to how we document our parameters in other tests.. Like:

https://github.com/coreos/fedora-coreos-config/blob/9fd3ef47a13c605b778a291d4cd9ef71afa0283e/tests/kola/root-reprovision/swap-before-root/test.sh#L4-L18

jlebon:
While we're here, can we remove the `-x` flag on the associated postprocess script? The loops are very spammy otherwise. (See e.g. the logs in the CI run of this PR itself: https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/blue/rest/organizations/jenkins/pipelines/fedora-coreos-config/branches/PR-1941/runs/2/nodes/42/steps/55/log/?start=0.)

Edit: ahhh, I see https://github.com/coreos/fedora-coreos-config/pull/1943 now. :)

tests/kola/files/validate-symlinks
```diff
@@ -1,13 +1,19 @@
...
-# kola: { "exclusive": false }
-# Checking if there are no broken symlinks from from /etc/ and /usr/ .
-# https://github.com/coreos/fedora-coreos-config/issues/1782
+# kola: { "exclusive": false, "distros": "fcos", "platforms": "qemu" }
+#
+# Check if there are broken symlinks in /etc/ and /usr/.
+# See: https://github.com/coreos/fedora-coreos-config/issues/1782
+#
+# - distros: fcos
+#   - Only run on FCOS until we've tested it on RHCOS.
```
cgwalters, line 8: The use of `-` on this like makes it look like nested YAML lists or something.  How about just:

```
# Only tested on FCOS, will test RHCOS later.  qemu only because this is the same on all platforms
# kola: { "exclusive": false, "distros": "fcos", "platforms": "qemu" }
```
?

dustymabe: I'd prefer we keep some sort of standard if possible. Those lines are just comments. If we change the preferred format it would be nice if we go through and update all tests to the new one.

dustymabe (approved):
LGTM

tests/kola/files/validate-symlinks
cgwalters: Well, it is a bit confusing here because we're parsing JSON from inside a comment, so having something that looks like YAML in a comment leads me to think it's being parsed.

So on that topic, this issue is arising because JSON doesn't have comments, but we can start using YAML instead of json here, perhaps using leading `##` to show where the yaml is...and then we can use `#` to comment the yaml:

```
#!/bin/bash
## kola:
##   exclusive: false
##   distros: "fcos"  # Not tested on rhcos yet
##   platforms: "qemu"  # test content is the same everywhere
# 
# Check if there are broken symlinks in /etc/ and /usr/.
# See: https://github.com/coreos/fedora-coreos-config/issues/1782
``

dustymabe: I'm not opposed to that idea but I'm not sure it's worth the effort to change what we have today. Either way this isn't something that should block this PR IMO. Maybe open a separate issue?

PR #1942, opened 2022-08-30
Make fedora-rawhide-nodebug-kernel repo gpgcheck=0

dustymabe:
Apparently the kernels that get in this repo can be unsigned depending
on the time that they were submitted to the repo. Let's just disable gpg
checking on the repo.

This shouldn't be an issue since we don't ship rawhide images to users
anyway and this repo only applies to the rawhide stream.


travier (approved):

dustymabe:
Merging since CI won't test this change.

PR #1943, opened 2022-08-30
manifests/fedora-coreos: fix verbosity on postprocess script

dustymabe:
In 75833da the intention was to silence the many messages printed
out by the symlink checking code but the wrong `set -` invocation
was updated. Let's update the correct one and re-set the one that
was erroneously updated.


travier:
Ouch, my bad.

travier (approved):

cgwalters (approved):

PR #1944, opened 2022-08-30
denylist: extend denials to `branched` stream, set snooze

dustymabe:
The denials related to [1] are now showing up in `branched` so
let's add it to the denylist and also convert to a snooze so we'll
follow up on this issue more regularly.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1215


gursewak1997 (approved):

dustymabe:
skipping CI again

PR #1946, opened 2022-08-31
tests/hostname: Limit to fcos only

cgwalters:
It never affected RHEL.

### DIFF
--- a/tests/kola/networking/hostname/fallback-hostname/test.sh
+++ b/tests/kola/networking/hostname/fallback-hostname/test.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -xeuo pipefail
 
-# kola: { "platforms": "qemu" }
+# kola: { "distros": "fcos", "platforms": "qemu" }
 #
 # Test that the fallback hostname is `localhost`. This test
 # validates that the fallback hostname is set to `localhost`
@@ -10,6 +10,8 @@ set -xeuo pipefail
 # hostname is set from the fallback hostname and is `localhost`.
 # https://github.com/coreos/fedora-coreos-tracker/issues/902
 #
+# - distros: fcos
+#   - The change only landed in fedora
 # - platforms: qemu
 #   - This test should pass everywhere if it passes anywhere.
 


tests/kola/networking/hostname/fallback-hostname/test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -xeuo pipefail
 
-# kola: { "platforms": "qemu" }
+# kola: { "distros": "fcos", "platforms": "qemu" }
```
travier, line 4: Can you update the list below to mention the new fcos only option? Thanks

cgwalters: Done; though I definitely disagree with the duplication inherent in this comment format

cgwalters: OK yeah I disagree so much with the copy-pasta inherent in this that this made me do https://github.com/coreos/coreos-assembler/pull/3059

dustymabe:
hmm. so yes, the change at https://fedoraproject.org/wiki/Releases/37/ChangeSet#Fallback_Hostname only landed in Fedora/FCOS, but the root problem never affected RHEL/RHCOS, so this test should be valid there too.

So the real question: was this test failing? If it was failing it was probably because of the way the test is written and could be modified to run on RHCOS too. i.e. maybe we modify the test to only check one or two things rather than 4.

HuijingHei:
Agree with you @dustymabe , maybe can merge this to let the CI run smoothly for rhcos, and do some tweaks to make it work for RHCOS in the follow PR.

dustymabe:
> Agree with you @dustymabe , maybe can merge this to let the CI run smoothly for rhcos, and do some tweaks to make it work for RHCOS in the follow PR.

Sounds good, but let's update the comment to be accurate (i.e. including the information from my previous comment).

travier:
We're not failing in RHCOS CI yet for this test, only when bumping the FCOS submodule: https://github.com/openshift/os/pull/965

travier:
And this does not fail for SCOS/RHCOS 9 apparently so maybe we need to skip it on RHCOS 8.

travier:
Did that in https://github.com/openshift/os/pull/965 so I think we can close this one.

cgwalters:
> And this does not fail for SCOS/RHCOS 9 apparently so maybe we need to skip it on RHCOS 8.

Yeah, that's fair.  This points to the obvious issue of our "distro" tag being too weak, what we really want is more like conditions `"osver": "rhel >= 9"` or so.  But doing that would require some work in kola.

PR #1947, opened 2022-08-31
tests/content-origins: Verify expected vendor for all RPM packages

cgwalters:
As we move to support multiple streams (particularly for C9S versus RHEL)
we should ensure that we're not accidentally including content
from CentOS in official RHEL builds, and vice versa.  The RPM vendor
field is useful for this.

For now, this test just verifies Fedora and RHEL vendors, and I
tested on both.

### DIFF
--- a/tests/kola/content-origins/test.sh
+++ b/tests/kola/content-origins/test.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# kola: {"platforms": "qemu", "exclusive": false, "distros": "fcos rhcos" }
+# Verify the RPM %{vendor} flag for everything installed matches what we expect.
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
+# - distros: This only handles Fedora and RHEL today.
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+. /usr/lib/os-release
+
+case "${ID}" in
+    fedora) vendor='Fedora Project';;
+    rhel|rhcos) vendor='Red Hat, Inc.';;
+    *) echo "Unknown operating system ID=${ID}; skipping this test"; exit 0;;
+esac
+
+cd $(mktemp -d)
+rpm -qa --queryformat='%{name},%{vendor}\n' > rpmvendors.txt
+if grep -vF ",${vendor}" rpmvendors.txt > unmatched.txt; then
+    cat unmatched.txt
+    fatal "Expected only vendor ${vendor} for all packages"
+fi
+echo "ok all RPMs produced by Vendor: ${vendor}"
+
--- a/tests/kola/content-origins/data/commonlib.sh
+++ b/tests/kola/content-origins/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh


tests/kola/content-origins/test.sh
```diff
@@ -0,0 +1,26 @@
...
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+. /usr/lib/os-release
+
+case "${ID}" in
+    fedora) vendor='Fedora Project';;
+    rhel|rhcos) vendor='Red Hat, Inc.';;
+    *) echo "Unknown operating system ID=${ID}; skipping this test"; exit 0;;
```
dustymabe, line 18: I wonder if we should actually error here if an unknown vendor pops up?

tests/kola/content-origins/test.sh
```diff
@@ -0,0 +1,26 @@
...
+esac
+
+cd $(mktemp -d)
+rpm -qa --queryformat='%{name},%{vendor}\n' > rpmvendors.txt
+if grep -vF ",${vendor}" rpmvendors.txt > unmatched.txt; then
+    echo 1>&2 "Expected only vendor ${vendor} for all packages"
+    cat unmatched.txt
```
dustymabe, line 24: seems like we need a `fatal` call here to notify the test failed.

tests/kola/content-origins/test.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
```
dustymabe, line 28: we usually have an `ok "no RPMs from unexpected vendor"` here.

tests/kola/content-origins/test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# kola: {"platforms": "qemu", "exclusive": false }
+# Verify the RPM %{vendor} flag for everything installed matches what we expect.
+#
+# - platforms: qemu
+#   - This test should pass everywhere if it passes anywhere.
```
dustymabe, line 6: I know you have some opposition to this so I'll say:

optional: mention exclusive=false here

tests/kola/content-origins/test.sh
cgwalters: You mean an unknown `ID` right?  We could, but it seems unnecessarily painful for e.g. someone who is creating a *coreos derivative and wants to run our tests.

I didn't want to scope in looking at c9s, partially because based on active discussions it seems highly likely we may end up e.g. using a COPR or a SIG repo at some point, and we don't *need* to be strict.

tests/kola/content-origins/test.sh
cgwalters: Good catch!

tests/kola/content-origins/test.sh
cgwalters: Right, my rationale against the duplicate comments is covered in https://github.com/coreos/coreos-assembler/pull/3060 and https://github.com/coreos/coreos-assembler/pull/3059

tests/kola/content-origins/test.sh
dustymabe: I think that's fair.. one counter is to make this run on `distros` of `fcos` and `rhcos` so this test won't even run on others?

still this is fine as is if we don't want to do that.

dustymabe (dismissed):
LGTM

optional: maybe we should rename the directory so we have something like `tests/kola/rpms/content-origins.sh` (nest under `rpms`?

cgwalters:
> optional: maybe we should rename the directory so we have something like tests/kola/rpms/content-origins.sh (nest under rpms?

I don't have a strong opinion here.  The idea of naming it "content-origins" was to be a bit independent of RPM, even though 98.2% of our content comes from RPMs.  (Perhaps in the future we somehow verify the systemd units we pull from src/config/overlay*?).  But yes OTOH it does only look at rpms today.

tests/kola/content-origins/test.sh
cgwalters: The tricky thing is today kola makes distros=`rhcos` run on centos stream https://github.com/coreos/coreos-assembler/blob/32e6d9f6664cfd05703d217e04f2c6f517e35f70/mantle/cmd/kola/options.go#L312

dustymabe (approved):
LGTM

travier (approved):

PR #1953, opened 2022-09-02
manifests: drop workaround for 'tss' user/group

lucab:
The `tss` user/group entries have been fixed in `tpm2-tss-3.2.0-3`,
so this drops all existing workarounds.
Both are static entries and are already covered by kola tests.


cgwalters (approved):

dustymabe:
Thank you for keeping up with this @lucab

PR #1956, opened 2022-09-06
tests: Switch one test to use platform-independent + YAML

cgwalters:
Since these two PRs

- https://github.com/coreos/coreos-assembler/pull/3064/commits/242e7149b8d57b5837238366bfcd70bd99b455a5
- https://github.com/coreos/coreos-assembler/pull/3059/commits/302130ec171d3d501310c40bd28db49c21459519

we support YAML for kola metadata, and there's a new `platform-independent`
tag.

Use both in a test I added recently to demonstrate.

Motivation: I think semantic tags and inline comments are far
superior to duplicating the test metadata and I don't want to
have to do that again in a test I write.

### DIFF
--- a/tests/kola/content-origins/test.sh
+++ b/tests/kola/content-origins/test.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "exclusive": false, "distros": "fcos rhcos" }
 # Verify the RPM %{vendor} flag for everything installed matches what we expect.
-#
-# - platforms: qemu
-#   - This test should pass everywhere if it passes anywhere.
-# - distros: This only handles Fedora and RHEL today.
-
+## kola:
+##   tags: "platform-independent"
+##   # This is a read-only, nondestructive test.
+##   exclusive: false
+##   # May support e.g. centos in the future
+##   distros: "fcos rhcos"
 set -xeuo pipefail
 
 . $KOLA_EXT_DATA/commonlib.sh


README.md
```diff
@@ -160,20 +160,13 @@ Tests should follow the following format:
...
 # Short summary of what the test does, why we need it, etc.
 #
 # Recommended: Link to corresponding issue or PR
-#
-# Explain the reasons behind all the kola options:
-# - distros: fcos
-#   - This test only runs on FCOS due to ...
-# - platforms: qemu
-#   - This test should ...
-# - etc.
```
bgilbert: "Use inline comments where applicable" is a weakening of the current guidance.  Since there isn't presently a consensus for that, let's keep the existing guidance wording, and convert this to an example YAML block that includes some directives and corresponding comments.

tests/kola/content-origins/test.sh
```diff
@@ -1,10 +1,9 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "exclusive": false, "distros": "fcos rhcos" }
+## kola:
+##   tags: "platform-independent"   # Just verifying content
```
bgilbert: We'll need more than ~40 characters for some of these comments, so let's set a convention of putting the YAML comment on a separate line?

README.md
bgilbert: (I do think it's a good idea to change the default to YAML though.)

cgwalters: OK there's likely to be many rounds of discussion around the new test style, I don't actually have a really strong opinion on exactly what it is as long as it doesn't involve duplicative copy/paste.

For now I dropped all changes to `README.md` out of this PR - I am just changing the test I recently added since I want to cross-verify the recent PRs work.

tests/kola/content-origins/test.sh
```diff
@@ -1,11 +1,11 @@
...
-# - platforms: qemu
-#   - This test should pass everywhere if it passes anywhere.
-# - distros: This only handles Fedora and RHEL today.
-
+# This is a read-only, nondestructive test.
+## kola:
+##   tags: "platform-independent"
+##   exclusive: false
+##   # May support e.g. centos in the future
+##   distros: "fcos rhcos"
```
dustymabe: I don't see why it's so bad to add the comment in the yaml to the `exclusive: false` option.

```suggestion
## kola:
##   tags: "platform-independent"
##   # This is a read-only, nondestructive test.
##   exclusive: false
##   # May support e.g. centos in the future
##   distros: "fcos rhcos"
```

cgwalters: Sure OK

dustymabe (approved):
LGTM

PR #1962, opened 2022-09-07
overlay.d/20platform-chrony: use generator-lib.sh for generator

dustymabe:
In this case the generator-lib.sh handles the logging setup and also offers a karg helper that we can use instead of the grep.


HuijingHei (approved):
/LGTM

PR #1967, opened 2022-09-09
[next-devel] Move to Fedora Linux 37

dustymabe:
The Fedora Linux 37 beta is shipping next week. Let's switch next-devel so we can ship it next week along with beta.


dustymabe:
We'll have to ignore the CI upgrade failure here because the image from may (last `next-devel`) is too old to support the new chunked oci-archive format.

cgwalters (approved):

PR #1971, opened 2022-09-12
[next] tree: promote changes from next-devel at 1cf759f

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/558.


dustymabe:
The "Check Promotion diffs" job seems to be complaining about me changing the `releasever: 37` in the `manifest.yaml`. In this case it's something we need so we can ignore the failure.

Note that the job did pass on the first commit so we know it was only the second commit that it's complaining about.

ravanelli (approved):
lgtm

PR #1972, opened 2022-09-13
kola test to check the network device info

aaradhak:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1153
BZ : https://bugzilla.redhat.com/show_bug.cgi?id=2080043

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -37,3 +37,10 @@
   streams:
   - branched
   - rawhide
+- pattern: ext.config.clhm.network-device-info
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1153
+  snooze: 2022-10-04
+  streams:
+  - testing-devel
+  - testing
+  - stable
--- a/tests/kola/clhm/network-device-info
+++ b/tests/kola/clhm/network-device-info
@@ -1,0 +1,15 @@
+#!/bin/bash
+## kola:
+##   # This is a read-only test that can be run with other tests.
+##   exclusive: false
+# This test makes sure CLHM wrote a snippet for the NIC IP info.
+# See - https://github.com/coreos/fedora-coreos-tracker/issues/1153
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+if [ ! -f /etc/issue.d/22*.issue ]; then
+    fatal "Network Device info does not exist"
+fi
+ok "Network Device info exists"
--- a/tests/kola/clhm/data/commonlib.sh
+++ b/tests/kola/clhm/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh


tests/kola/networking/network-device-info
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# See - https://github.com/coreos/fedora-coreos-tracker/issues/1153
+
+if [ ! -f "/etc/issue.d/22*.issue" ]; then
```
dustymabe:
```suggestion
if [ ! -f /etc/issue.d/22*.issue ]; then
```

If you quote this it won't expand the `*`.

tests/kola/networking/network-device-info
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# See - https://github.com/coreos/fedora-coreos-tracker/issues/1153
```
dustymabe: Let's move this to the top and add a comment explaining what the test does.

```
This test makes sure CLHM wrote a snippet for the NIC IP info. See https://github.com/coreos/fedora-coreos-tracker/issues/1153
```

tests/kola/networking/network-device-info
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# kola: { "exclusive": false }
```
dustymabe: We can now do this in yaml (and add comments!)

```suggestion
## kola:
##   # This is a read-only test that can be run with other tests.
##   exclusive: false
```

dustymabe (commented):
Since this is testing CLHM how about we move the test into `tests/kola/clhm/network-device-info ?

dustymabe:
we'll need to add a denylist entry for this test for `testing-devel`, `testing`, and `stable`. until the selinux fix makes it in. Just set the test to snooze for 3 weeks and we'll re-evaluate after that point in time.

aaradhak:
> we'll need to add a denylist entry for this test for `testing-devel`, `testing`, and `stable`. until the selinux fix makes it in. Just set the test to snooze for 3 weeks and we'll re-evaluate after that point in time.

is a separate PR required for the denylist entry?

dustymabe:
> is a separate PR required for the denylist entry?

no, but I would do it in a separate commit

kola-denylist.yaml
```diff
@@ -37,3 +37,10 @@
   streams:
   - branched
   - rawhide
+pattern: network-device-info
```
dustymabe:
```suggestion
- pattern: ext.config.clhm.network-device-info
```

tests/kola/clhm/network-device-info
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+## kola:
+	##   # This is a read-only test that can be run with other tests.
+	##   exclusive: false
```
dustymabe:
```suggestion
## kola:
##   # This is a read-only test that can be run with other tests.
##   exclusive: false
```

kola-denylist.yaml
```diff
@@ -37,3 +37,10 @@
   streams:
   - branched
   - rawhide
+pattern: ext.config.clhm.network-device-info
```
dustymabe:
```suggestion
- pattern: ext.config.clhm.network-device-info
```

dustymabe:
I think you also need to create the `data/commonlib.sh` symlink.

```
$ mkdir tests/kola/clhm/data
$ ln -s ../../data/commonlib.sh tests/kola/clhm/data/commonlib.sh
$ git add tests/kola/clhm/data/commonlib.sh
```

aaradhak:
> ln -s ../../data/commonlib.sh tests/kola/clhm/data/commonlib.sh

I forgot about this part. I ll add it.

dustymabe (approved):
LGTM

PR #1974, opened 2022-09-13
tests: check `platforms.json` against kargs/`grub.cfg` and ground truth

bgilbert:
Hardcode the expected cutover dates for each FCOS stream, to ensure we don't cut over too early or too late.

Split out of https://github.com/coreos/fedora-coreos-config/pull/1781.

### DIFF
--- a/tests/kola/files/console-config
+++ b/tests/kola/files/console-config
@@ -1,0 +1,99 @@
+#!/bin/bash
+## kola:
+##   # We don't need a special Ignition config and we don't modify the VM
+##   exclusive: false
+##   # s390x doesn't have any configuration in platforms.yaml, so
+##   # platforms.json is not included in the image
+##   architectures: "!s390x"
+#
+# Verify that the kargs and grub.cfg commands specified in platforms.json
+# have been properly applied to the image.  Also check that cosa correctly
+# translated platforms.yaml to platforms.json, by spot-checking certain
+# expected platforms.json values.
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+platform_json_kargs() {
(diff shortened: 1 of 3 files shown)


tests/kola/files/console-config
```diff
@@ -0,0 +1,109 @@
...
+# have been properly applied to the image.  Also check that cosa correctly
+# translated platforms.yaml to platforms.json, by spot-checking certain
+# expected platforms.json values.
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+cmdline=( $(</proc/cmdline) )
+cmdline_arg() {
```
dustymabe: I'm wondering if we should put a common function like this in `commonlib.sh`.

tests/kola/files/console-config
```diff
@@ -0,0 +1,109 @@
...
+
+# Check whether platforms.json exists
+. /usr/lib/os-release
+if [ "$ID" = fedora ]; then
+    case "$(get_fcos_stream)" in
+    next-devel) threshold=20220923 ;;
+    next) threshold=20220930 ;;
+    testing-devel|rawhide|branched) threshold=20221118 ;;
+    testing) threshold=20221125 ;;
+    stable) threshold=20221209 ;;
```
dustymabe, line 48: Should we link to the source of where these dates were decided?

dustymabe (approved):
LGTM

tests/kola/files/console-config
bgilbert: The coreos-status post hasn't gone out yet, so in that case, this PR will need to wait a bit.

dustymabe (approved):

tests/kola/files/console-config
bgilbert: Rebased and added the schedule link.

dustymabe (approved):
LGTM

PR #1983, opened 2022-09-19
tests/kola: update threshold for `next-devel` console switch

bgilbert:
The console change has landed now, so update the threshold to keep tests green.


dustymabe (approved):
LGTM

PR #1986, opened 2022-09-23
live-generator: Use a loop to work around potential races with udev

cgwalters:
This is a bigger hammer for the same issue as
https://github.com/coreos/fedora-coreos-config/pull/411/commits/170df10014e853bdfb5e82e484583d8c70db628c

`mount: /run/media/iso: /dev/sr0 already mounted or mount point busy`

While I haven't been able to reproduce the problem in synthetic testing with qemu, I am increasingly confident that there *are* other udev rules running concurrently with our mounting attempt.

This was motivated by https://issues.redhat.com/browse/OCPBUGS-1505 where I specifically noticed that the ISO appears as a USB device (which is really itself virtual and provided by the iLO), and appears much later system bootup than I see with qemu, even with emulated USB CD-ROM.


travier (dismissed):

travier:
Failed in CI on test ISO so failure seems legitimate (I have not taken a deeper look yet).

dustymabe:
> 170df10

Can you update this link to be https://github.com/coreos/fedora-coreos-config/commit/250cba0e30d6378f38dc80e0a791494fb0c27ab9

`170df10` is not a commit in the history of this repo.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -107,6 +107,7 @@ EOF
 [Unit]
 DefaultDependencies=false
 Before=initrd-root-fs.target
+Requires=run-media-iso-mount.service
```
dustymabe: Does this need an `After=run-media-iso-mount.service` too or does the `RequiresMountsFor=/run/media/iso` satisfy that?

dustymabe:
I'm not the biggest fan of changing this up (at least for FCOS) because I just spent a bunch of time investigating what I think is a systemd bug in this area around mounts (https://github.com/coreos/fedora-coreos-tracker/issues/1233), but if it's the only way then I'm +1 for the change.

I guess we can see what the original reporter of the issue says about if they can reproduce with Fedora CoreOS itself or not.

dustymabe:
Not sure what's going on with CI, but for some reason `ignition-files.service` appears to run twice and fails because it won't overwrite files that already exist (because it already ran):

```
[   11.946991] ignition[869]: CRITICAL : Ignition failed: failed to create files: failed to create files: error creating /sysroot/var/opt/pointer.ign: error creating file "/sysroot/var/opt/pointer.ign": A file exists there already and overwrite is false^M      

```

cgwalters:
> I'm not the biggest fan of changing this up (at least for FCOS) because I just spent a bunch of time investigating what I think is a systemd bug in this area around mounts (https://github.com/coreos/fedora-coreos-tracker/issues/1233), but if it's the only way then I'm +1 for the change.

Can you be more concrete with your concern?  Is it that you're worried about introducing new regressions?  Is the "at least for FCOS" bit saying that you think we should just override the live generator for RHCOS8?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
jlebon: `After=` makes sense to me. There's also `coreos-liveiso-persist-osmet.servive` which uses `RequiresMountsFor=/run/media/iso` where we'll want this.

jlebon (commented):
I've been digging into this as well and trying to reproduce it locally in various ways without success.

However, I did find at least one potential suspect: https://github.com/systemd/systemd/blob/fc709443016b0da2da97809828a45ba000f358df/rules.d/60-cdrom_id.rules#L20. systemd runs `cdrom_id --lock-media` to lock the tray and enable handling of the eject button.

Previously, `cdrom_id` would open the block device using `O_EXCL`, but this was later on changed to drop `O_EXCL`: https://github.com/systemd/systemd/commit/14cd12b3d50f21cb9be7e27785142c6c321fac39

RHEL 8.4 and 8.6 do not have this patch, so still use `O_EXCL`. This means that if the `mount` call runs at the same time as `cdrom_id`, it can get `EBUSY`. I think ideally we'd try to get that systemd patch backported, but the paperwork involved (especially to backport all the way to 8.4) is considerable.

If the customer confirms that the latest FCOS works (which has the systemd patch to drop `O_EXCL`), that would reinforce the theory.

I do share some of @dustymabe's hesitations. Hopefully we can all agree that a service unit is less elegant than a mount unit. Doing a retry loop like this feels like papering over the underlying issue, but it seems like the easiest path forward for now at least for RHCOS8.

Relatedly, `cdrom_id` *also* [has a retry loop](https://github.com/systemd/systemd/blob/fc709443016b0da2da97809828a45ba000f358df/src/udev/cdrom_id/cdrom_id.c#L745-L753) when hitting `EBUSY`. This seems to predate the udev and systemd merger, but one can see the original addition here: https://git.kernel.org/pub/scm/linux/hotplug/udev.git/commit/extras/cdrom_id/cdrom_id.c?id=cccfffbe04c01be12fb42cb12f3f7aa5e2a22dd4. However, AFAICT the original motivation for the retry loop was *because* `cdrom_id` was changed to use `O_EXCL` (see the original thread which I think motivated that patch: https://www.spinics.net/lists/hotplug/msg03593.html). So I think `cdrom_id` shouldn't really need the retry loop anymore now that `O_EXCL` was removed.

I know it's not great, but what do people think about conditionalizing this on el8? And then we can eventually drop it in the future.

Let's also try to have the customer confirming this fixes the issue before merging.

jlebon:
> I know it's not great, but what do people think about conditionalizing this on el8? And then we can eventually drop it in the future.

If there's agreement on this, I can pick up this PR and do that.

cgwalters:
Good thinking to dig specifically into `cdrom_id` changes.  Yeah, conditionalizing the hack on rhel8 seems sane to me.

travier:
+1 for RHEL 8 only

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -24,6 +24,12 @@ if ! is-live-image; then
     exit 0
 fi
 
+# copied from tests/kola/data/commonlib.sh
+is_rhcos8() {
+    source /etc/os-release
```
cgwalters: Does this leak into the outer shell environment?  Perhaps a bit cleaner as e.g.

```
(. /usr/lib/os-release && [ "${ID}" == "rhcos" ] && [ "${RHEL_VERSION%%.*}" -eq 8 ])
```
to explicitly scope to a subshell?

jlebon: It turns out this doesn't work anyway on 4.10, so went with a different approach.

jlebon:
OK, this is now tested working on RHCOS 4.10!

```
$ cosa kola testiso -S
kola -p qemu-unpriv --output-dir tmp/kola testiso -S
Testing scenarios: [miniso-install miniso-install-nm pxe-install iso-offline-install pxe-offline-install iso-live-login iso-as-disk]
Ignoring verification of signature on metal image
Could not open option rom '/usr/share/qemu/pxe-rtl8139.rom': No such file or directory
Successfully tested scenario pxe-install for 410.84.202209272044-0 on bios (metal)
Could not open option rom '/usr/share/qemu/pxe-rtl8139.rom': No such file or directory
Successfully tested scenario pxe-offline-install for 410.84.202209272044-0 on bios (metal)
Successfully tested scenario iso-offline-install for 410.84.202209272044-0 on bios (metal)
Successfully tested scenario iso-live-login for 410.84.202209272044-0 on bios (metal)
Successfully tested scenario iso-as-disk for 410.84.202209272044-0 on bios (metal)
Successfully tested scenario miniso-install for 410.84.202209272044-0 on bios (metal)
Successfully tested scenario miniso-install-nm for 410.84.202209272044-0 on bios (metal with NM keyfile)

$ grep run-media-iso-mount tmp/kola/iso-live-login/console.txt
         Starting run-media-iso-mount.service...
[    4.054623] systemd[1]: Starting run-media-iso-mount.service...
[  OK  ] Started run-media-iso-mount.service.
[    4.144065] systemd[1]: Started run-media-iso-mount.service.
[    8.880136] systemd[1]: run-media-iso-mount.service: Succeeded.
[    8.880735] systemd[1]: Stopped run-media-iso-mount.service.
[  OK  ] Stopped run-media-iso-mount.service.
```

It should have no impact elsewhere.

Will upload an ISO for the customer to test.

cgwalters:
LGTM - I can't actually approve this since it's my own PR.  But you can!  (insert obama awarding self medal meme)

Out of curiosity do you know why my initial code failed one of the testiso scenarios?

jlebon:
> Out of curiosity do you know why my initial code failed one of the testiso scenarios?

I think it was because it was missing the `add_requires` bit to actually have it run.

dustymabe (approved):
LGTM

jlebon:
Hmm, I was planning to hold this until we have a confirmation that it actually fixes something. I'm pretty confident it won't break anything at least.

dustymabe:
> Hmm, I was planning to hold this until we have a confirmation that it actually fixes something. I'm pretty confident it won't break anything at least.

Sorry!

jlebon:
Customer confirmed this fixes their issue. :tada: 
Will start the backporting process.

PR #1988, opened 2022-09-26
Drop everything in `/etc/dnf/protected.d`

cgwalters:
We don't need "protected packages"; rpm-ostree inherently provides a versioned filesystem and hence you can't actually delete something that's required.

This fixes `rpm-ostree override replace https://koji.fedoraproject.org/koji/buildinfo?buildID=1976204` with a systemd build.

### DIFF
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -45,6 +45,13 @@ postprocess:
     #!/usr/bin/env bash
     set -xeuo pipefail
     rm -f /usr/bin/dotlockfile
+  # We don't need "protected packages"; rpm-ostree inherently provides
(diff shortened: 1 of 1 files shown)


dustymabe:
Thinking about alternatives.. I'm wondering if we could/should just have rpm-ostree do the right thing here. i.e. this fixes the problem for FCOS, but if we know this won't apply for any `rpm-ostree` system could we somehow set API flags for DNF that would fix it everywhere?

cgwalters:
Yeah, it's a valid point.  Hmm, actually now that I debug further it seems like this case was potentially legitimately hit in that I was trying to use `override replace` incompletely and the solver walked off into actually trying to fully remove systemd.

Issue #1743, opened 2022-05-16
Find a safer alternative to check unit status


travier:
The following code in tests will always return true if the unit does not exists:
```
if [ $(systemctl show -p Result kdump.service) != "Result=success" ]; then
```

Thus we need a safer / smarter way to check this value.

See https://github.com/coreos/fedora-coreos-config/pull/1742#discussion_r873326178

HuijingHei:
Thanks @dustymabe for the review, close this as https://github.com/coreos/fedora-coreos-config/pull/1989 merged

PR #1989, opened 2022-09-27
commonlib: add func to check unit is started

HuijingHei:
The code in tests will always return true if the unit does not exist: 
`if [ $(systemctl show -p Result kdump.service) != "Result=success" ]; then` 
See https://github.com/coreos/fedora-coreos-config/issues/1743

Make sure service is active, and wait for 20s if service is activating

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -71,3 +71,14 @@ cmdline_arg() {
     done
     echo "${value}"
 }
+
+# wait for 20s when in activating status
+is_service_active() {
+    local service="$1"
+    for x in {0..20}; do
+        [ $(systemctl is-active "${service}") != "activating" ] && break
+        sleep 1
+    done
+    # return actual result
+    systemctl is-active "${service}"
+}
--- a/tests/kola/kdump/crash/test.sh
+++ b/tests/kola/kdump/crash/test.sh
@@ -16,7 +16,7 @@ set -xeuo pipefail
 
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
-      if [ $(systemctl show -p Result kdump.service) != "Result=success" ]; then
+      if ! is_service_active kdump.service; then
           fatal "kdump.service failed to start"
       fi
       # Verify that the crashkernel reserved memory is large enough


tests/kola/data/commonlib.sh
```diff
@@ -71,3 +71,9 @@ cmdline_arg() {
     done
     echo "${value}"
 }
+
+# verify service exists and active
+is_service_active() {
+    local service_name="$1"
+    systemctl cat -- "${service_name}.service" &> /dev/null && [ $(systemctl is-active "${service_name}.service") == "active" ]
```
dustymabe: i'm not too confident in `is-active` here. What if the unit is still running and hasn't exited yet?

I think I'd rather add the `systemctl cat` and then keep what we had before verifying `Result=success`.

dustymabe: of course maybe my concerns aren't valid

HuijingHei: Thanks Dusty for the review.
`systemctl cat` is to check service exists
`systemctl is-active` is to check service started

if service not exists, `systemctl show -p Result ttt.service` always get `Result=success`, so it is not a safe way
if the unit is running and not exited yet, `systemctl is-active` will get `activating`, so here we check `active`

Does this make sense?

dustymabe: Since we added `systemctl cat` we addressed the unsafe part of `systemctl show -p Result`. I just think checking `Result=success` is more straightforward to someone reading the code.

HuijingHei: IMU, `systemctl cat` just can check service existed/installed, add `systemctl show -p Result` can not ensure service started because always get `Result=success`. As you asked what if the unit is still running and hasn't exited yet?

dustymabe: > IMU, `systemctl cat` just can check service existed/installed,

Correct. That's the thing that was missing before.
    
> add `systemctl show -p Result` can not ensure service started because always get `Result=success`.

If the unit exists and hasn't completed it won't get `Result=success`

> As you asked what if the unit is still running and hasn't exited yet?

In the case of the kdump test we want to see a failure if at the time we check the unit hasn't completed. If that starts to happen we'll need to adjust the test. Basically what I'm doing in the kdump test is:

1. verify the unit was pulled in
2. verify the unit has completed successfully before moving on to the rest of the test.

dustymabe: ok after going through this a bit with @HuijingHei - we found that `systemctl show -p Result` doesn't imply the service has run when it shows success. We did verify that when `kdump.service` has run and exited it will still show as `active` since `RemainAfterExit=yes`.

dustymabe: and.. now here I'm thinking that we don't need the `systemctl cat` any longer here:

```
$ systemctl is-active foo.service
inactive
```

If the service doesn't exist then `systemctl is-active foo.service` will return `inactive`.

And.. we can probably just make this line:

```
return systemctl is-active kdump.service
```

In which case it's so simple we probably don't need a function for this. Just replace the original line in `test.sh` with:

```
if ! systemctl is-active kdump.service; then
```

dustymabe: actually, while the script is still running `systemctl is-active kdump.service` will show `activating`.

tests/kola/data/commonlib.sh
```diff
@@ -71,3 +71,19 @@ cmdline_arg() {
...
+    for x in {0..20}; do
+        status=$(systemctl is-active ${service_name}.service)
+        if [ $status == "active" ] || [ $status == "inactive" ]; then
+            break
+        elif [ $status == "activating" ]; then
+            sleep 1
+        fi
+    done
+
+    # return acutal result
```
dustymabe:
```suggestion
    # return actual result
```

tests/kola/data/commonlib.sh
```diff
@@ -71,3 +71,19 @@ cmdline_arg() {
...
+# wait for 20s when in activating status
+is_service_active() {
+    local service_name="$1"
+    for x in {0..20}; do
+        status=$(systemctl is-active ${service_name}.service)
+        if [ $status == "active" ] || [ $status == "inactive" ]; then
+            break
+        elif [ $status == "activating" ]; then
+            sleep 1
+        fi
```
dustymabe: Is this equivalent to what you've got?

```suggestion
        [ $status != "activating" ] && break
        sleep 1
```

tests/kola/kdump/crash/test.sh
```diff
@@ -16,7 +16,7 @@ set -xeuo pipefail
 
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
-      if [ $(systemctl show -p Result kdump.service) != "Result=success" ]; then
+      if ! is_service_active kdump; then
```
dustymabe: optional:

```suggestion
      if ! is_service_active kdump.service; then
```

I think this might more clear we're talking about systemd service by  adding the `.service` here (and modifying the function).

dustymabe (dismissed):
LGTM - a few comments

tests/kola/data/commonlib.sh
HuijingHei: Thanks! updated

tests/kola/data/commonlib.sh
HuijingHei: Nice suggestion :+1:

dustymabe (approved):
LGTM

PR #1991, opened 2022-09-28
overlay.d/14NetworkManager-plugins: Explicitly disable ifcfg-rh plugin

LorbusChris:
The NM issue [1] where compiled-in default plugins could not be disabled with the `-=`  syntax has been fixed upstream, so move to explicitly disabling the ifcfg-rh plugin.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/491

### DIFF
--- a/overlay.d/14NetworkManager-plugins/usr/lib/NetworkManager/conf.d/10-disable-default-plugins.conf
+++ b/overlay.d/14NetworkManager-plugins/usr/lib/NetworkManager/conf.d/10-disable-default-plugins.conf
@@ -1,9 +1,0 @@
-# Stop NetworkManager from trying to load the ifcfg-rh plugin by default,
-# which we don't ship.  This actually disables all default plugins, of which
-# ifcfg-rh is currently the only one.
-#
-# Note that we must do this for now because `-=` syntax doesn't work
-# with compiled-in defaults. Proposed upstream fix:
-# https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/491
-[main]
-plugins=
--- a/overlay.d/14NetworkManager-plugins/usr/lib/NetworkManager/conf.d/10-disable-ifcfg-rh-plugin.conf
+++ b/overlay.d/14NetworkManager-plugins/usr/lib/NetworkManager/conf.d/10-disable-ifcfg-rh-plugin.conf
@@ -1,0 +1,4 @@
+# Stop NetworkManager from trying to load the ifcfg-rh plugin by default,
+# which we don't ship.
+[main]
+plugins-=ifcfg-rh


dustymabe:
Aside: Just wondering.. How did you happen upon this?

Now.. on to the strategy here...

It turns out (with some encouragement from us) the NM team [moved the `ifcfg-rh` plugin to a subpackage](https://fedoraproject.org/wiki/Changes/NoIfcfgFiles) in [Fedora 36](https://bodhi.fedoraproject.org/updates/FEDORA-2022-e675404750) (i.e. no longer included by default). Instead of what you propose here I propose we:

- remove this `14NetworkManager-plugins` overlay completely
- add `NetworkManager-initscripts-ifcfg-rh` to the list of packages to make sure we never include (for FCOS)
- add a test that verifies the plugin exists or not based on the OS (i.e. RHCOS 8 would want to verify it exists and Fedora* would want to verify it doesn't exist). For RHCOS 9, not sure cc @travier.

travier:
ifcfg files are no longer the default in RHEL 9: https://www.redhat.com/en/blog/rhel-9-networking-say-goodbye-ifcfg-files-and-hello-keyfiles

travier:
We currently have this: https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/lib/NetworkManager/conf.d/10-disable-default-plugins.conf

Maybe we could remove it for RHEL 9 but not sure it's urgent.

travier:
(so overall, changes here do not impact RHCOS)

dustymabe:
> We currently have this: https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/lib/NetworkManager/conf.d/10-disable-default-plugins.conf
> 
> Maybe we could remove it for RHEL 9 but not sure it's urgent.

I think you'll want to make whatever changes we make here (i.e. making sure the `NetworkManager-initscripts-ifcfg-rh` is in the exclude list) for RHCOS 9 as soon as possible so any early testing doesn't yield false-positive feedback.

travier:
Filed https://github.com/openshift/os/issues/1022 to discuss that.

LorbusChris:
@dustymabe we stumbled over this when adding the `05rhcos` overlay to okd-machine-os (there was an error because the file now exists twice, in two different overlays), and I thought doing the cleanup couldn't hurt. I think your proposal makes sense, though.
Closing this so the discussion can be focused in https://github.com/openshift/os/issues/1022

/close

dustymabe:
any volunteers to implement https://github.com/coreos/fedora-coreos-config/pull/1991#issuecomment-1261092107 for FCOS?

travier:
Did your suggestion in https://github.com/coreos/fedora-coreos-config/pull/1993

PR #1993, opened 2022-09-30
manifest: Exclude support for NM legacy ifcfg config format

travier:
Since Fedora 36, the ifcfg-rh plugin implementing support for reading network configuration files in the legacy ifcfg format has been split out into its own RPM subpackage.

We can now remove our configuration workaround and instead ensure that we do not ship this plugin at all.

See: https://github.com/coreos/fedora-coreos-config/pull/1991

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -19,7 +19,6 @@ ostree-layers:
   - overlay/05core
   - overlay/08nouveau
   - overlay/09misc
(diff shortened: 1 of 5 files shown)


travier:
Untested as I'm on a train 🚋

dustymabe (approved):
LGTM

dustymabe:
followup in https://github.com/coreos/fedora-coreos-config/pull/1994

PR #1994, opened 2022-09-30
tests/kola: add ext.config.networking.nm-ifcfg-rh-plugin test

dustymabe:
This test checks that the ifcfg-rh NetworkManager plugin is in the state we expect it to be in. Some more context in:

- https://github.com/coreos/fedora-coreos-config/pull/1993
- https://github.com/coreos/fedora-coreos-config/pull/1991
- https://github.com/openshift/os/issues/1022

### DIFF
--- a/tests/kola/networking/nm-ifcfg-rh-plugin
+++ b/tests/kola/networking/nm-ifcfg-rh-plugin
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+# We want to do a check here to make sure the ifcfg-rh plugin
+# is not loaded on FCOS (it can't be because it should never
+# even be included in the filesystem). We also want to make
+# sure that at least for RHCOS8 it is included.
+#
+## kola:
+##   # This test should behave the same on every platform.
+##   tags: "platform-independent"
+##   # This is a read-only, nondestructive test.
+##   exclusive: false
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+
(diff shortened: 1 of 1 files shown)


mike-nguyen (dismissed):
LGTM

dustymabe:
I'll wait for a review from @travier - so he can comment on what we should do for RHCOS9 and SCOS here.

travier:
Well, I'm waiting on feedback on the RHCOS issue 😅. Let's wait for next week to see if we have objections.

travier:
Can we add that check for RHCOS 9 too? We'll keep it for now. Thanks

tests/kola/networking/nm-ifcfg-rh-plugin
```diff
@@ -0,0 +1,32 @@
...
+
+
+# Check if it exists or not. The plugin provides a dbus interface
+# so if it is loaded there will be something listening at that name
+exists=0
+busctl status com.redhat.ifcfgrh1 && exists=1
+
+if is_fcos; then
+    [ "$exists" == "0" ] || fatal "ifcfg-rh plugin detected on FCOS"
+elif is_rhcos8; then
```
travier: Maybe:
```suggestion
elif is_rhcos8 || is_rhcos9; then
```

travier: and C9S too

dustymabe: done

travier (approved):

PR #1995, opened 2022-09-30
[rhcos-4.11] live-generator: Use a loop to work around potential races with udev on el8

jlebon:
This is a bigger hammer for the same issue as
https://github.com/coreos/fedora-coreos-config/pull/411/commits/170df10014e853bdfb5e82e484583d8c70db628c

While I haven't been able to reproduce the problem in synthetic testing with qemu, I am increasingly confident that there *are* other udev rules running concurrently with our mounting attempt.

This was motivated by https://issues.redhat.com/browse/OCPBUGS-1505 where I specifically noticed that the ISO appears as a USB device (which is really itself virtual and provided by the iLO), and appears much later system bootup than I see with qemu, even with emulated USB CD-ROM.

A possibly concrete culprit is `cdrom_id`, which on RHEL8 does `open(O_EXCL)` on the CD-ROM device. For more information, see:

https://github.com/coreos/fedora-coreos-config/pull/1986#pullrequestreview-1120840529

Co-authored-by: Jonathan Lebon <jonathan@jlebon.com>
(cherry picked from commit 3481a8c70b3ef16f6e31b84e9d5fd9313c2a7d87)


dustymabe (approved):
LGTM

PR #1997, opened 2022-09-30
[rawhide] overrides: pin linux-firmware 20220815-139.fc38

gursewak1997:
pin `linux-firmware 20220815-139.fc38` till we find a solution for https://github.com/coreos/fedora-coreos-tracker/issues/1310

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,14 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  linux-firmware:
+    evra: 20220815-139.fc38.noarch
+    metadata:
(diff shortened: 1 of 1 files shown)


manifest-lock.overrides.yaml
```diff
@@ -8,4 +8,16 @@
...
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  linux-firmware:
+    evra: 20220815-139.fc38.noarch
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-a3e93132cc
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1310
+      type: fast-track
```
dustymabe: This should probably be `pin`:

```suggestion
      type: pin
```

Since it is a pin I don't think we need the bodhi link included.

dustymabe (approved):
LGTM

PR #1999, opened 2022-10-03
[stable] tree: promote changes from testing at 0510450ecb14737a664763caea5ca097ceda93be

coreosbot-releng:
Triggered by @marmijo in https://github.com/coreos/fedora-coreos-streams/issues/567.


kola-denylist.yaml
```diff
@@ -23,17 +23,23 @@
   - s390x
 - pattern: ext.config.ignition.resource.remote
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1215
-  snooze: 2022-09-10
+  snooze: 2022-09-27
```
dustymabe: let's add a second commit here to drop the `snooze` from these 2 test denials related to `#1215`. This will just ensure we don't get a flake in our `s390x` run.

marmijo: done!

dustymabe (approved):
LGTM

PR #2002, opened 2022-10-03
overrides: pin to linux-firmware 20220815-139.fc37

gursewak1997:
pin linux-firmware `20220815-139.fc37` till we find a solution for https://github.com/coreos/fedora-coreos-tracker/issues/1310


travier:
This test is less important than a kernel update. Let's do https://github.com/coreos/fedora-coreos-tracker/issues/1310#issuecomment-1265726836 instead

gursewak1997:
Closed in favour of https://github.com/coreos/fedora-coreos-config/pull/2003

PR #2003, opened 2022-10-03
kola/validate-symlinks: Add qcom/LENOVO/21BX.xz to broken symlink list

gursewak1997:
Add `qcom/LENOVO/21BX.xz` to the broken symlinks list.
Ref issue: https://github.com/coreos/fedora-coreos-tracker/issues/1310


dustymabe (approved):
LGTM

PR #2004, opened 2022-10-04
overrides: Drop linux-firmware package pin

gursewak1997:
Ref Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1310#issuecomment-1265899786


dustymabe (approved):
LGTM

PR #2005, opened 2022-10-04
kola/tests: add ext.config.platforms.aws.{nvme,assert-xen} tests

dustymabe:
```
commit 199a66ebe5e46e17e097aa742a873cd647d4697a
Date:   Tue Oct 4 10:57:35 2022 -0400

    tests/kola: add ext.config.platforms.aws.assert-xen test
    
    This is just an extra layer to make sure that when we run an AWS
    Xen test in our pipeline that the instance is legitimately a Xen
    instance. It will only be run if the caller requested the test
    by name or via the aws-xen-test tag.

commit c9decffd79a50c89b15e6dbd99b42caef78f4db5
Date:   Tue Oct 4 10:56:06 2022 -0400

    tests/kola: add ext.config.platforms.aws.nvme test
    
    This test ensures that if an nvme device exists it is accessible.
    See https://github.com/coreos/fedora-coreos-tracker/issues/1306
    
    This commit also denylists the test with a snooze for the next few
    weeks. The hope is that Amazon does the firmware rollout soon.

```

### DIFF
--- a/tests/kola/platforms/aws/assert-xen
+++ b/tests/kola/platforms/aws/assert-xen
@@ -1,0 +1,25 @@
+#!/bin/bash
+# Test to make sure the booted AWS instance is XEN based
+#
+## kola:
+##   # This is a read-only test and can be run with other tests
+##   exclusive: false
+##   # This test is targeted at AWS
+##   platforms: aws
+##   # Force this test to not run by default unless named specifically
+##   # or `--tag aws-xen-test` is passed to `kola run`. i.e. this test
+##   # should only run on Xen instances and the caller should request
+##   # the test.
+##   requiredTag: aws-xen-test
+#
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+hypervisor=$(curl http://169.254.169.254/2022-09-24/meta-data/system || true)
+if [ "${hypervisor}" != "xen" ]; then
+    fatal "expected xen instance type"
+fi
+
+ok xen instance type
--- a/tests/kola/platforms/aws/nvme
+++ b/tests/kola/platforms/aws/nvme
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Test to make sure AWS instances with nvme storage get a device
+# properly created for it. See https://github.com/coreos/fedora-coreos-tracker/issues/1306
+#
+## kola:
+##   # This is a read-only test and can be run with other tests
+##   exclusive: false
+##   # This test is targeted at AWS
+##   platforms: aws
+
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# We'll check to see if anything is listed in `nvme list-subsys`
+# and then run the test based on that. Example output:
+# 
+# [core@ip-10-0-1-155 ~]$ nvme list-subsys -o json
+# [
+#   {
+#     "HostNQN":"nqn.2014-08.org.nvmexpress:uuid:c2c16640-0d5c-472a-846c-2f981c5914db",
+#     "HostID":"c37c43ca-31a3-4015-8704-1f493081da36",
+#     "Subsystems":[
+#       {
+#         "Name":"nvme-subsys0",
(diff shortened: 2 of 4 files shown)


tests/kola/platforms/aws/nvme
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# Test to make sure AWS instances with nvme storage get a device
+# properly created for it. See https://github.com/coreos/fedora-coreos-tracker/issues/1306
+#
+## kola:
+##   # This is a read-only test and can be run with other tests
+##   exclusive: false
```
jlebon: Minor: trailing whitespace

tests/kola/platforms/aws/assert-xen
```diff
@@ -0,0 +1,25 @@
...
+##   # the test.
+##   requiredTag: aws-xen-test
+#
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+hypervisor=$(curl http://169.254.169.254/2022-09-24/meta-data/system || true)
+if [ "${hypervisor}" != "xen" ]; then
```
jlebon, line 21: Do you have more background on this test? Would this have caught a recent issue.

This feels more like we're testing AWS than testing our own artifacts.

tests/kola/platforms/aws/assert-xen
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# Test to make sure the booted AWS instance is XEN based
+#
+## kola:
+##   # This is a read-only test and can be run with other tests
+##   exclusive: false
```
jlebon: Minor: trailing whitespace

tests/kola/platforms/aws/nvme
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# Test to make sure AWS instances with nvme storage get a device
+# properly created for it. See https://github.com/coreos/fedora-coreos-tracker/issues/1306
```
jlebon, line 3: Although, isn't the fact that we were able to boot at all to run any test already covering that tracker issue? Should we instead be running e.g. `basic` on on of those instance types in `kola-aws`?

dustymabe: > Although, isn't the fact that we were able to boot at all to run any test already covering that tracker issue?

No. In this case the instance storage isn't the boot disk. It's just an extra disk.

> Should we instead be running e.g. `basic` on on of those instance types in `kola-aws`?

PR to switch the instance type used for `Xen` tests in https://github.com/coreos/fedora-coreos-pipeline/pull/669

tests/kola/platforms/aws/assert-xen
dustymabe: It is a bit of testing AWS. TBH I can't find easily what instances types are what and I don't think having an `exlusive: false` test just to make sure someone didn't make a mistake is OK.

I also wouldn't be surprised if AWS didn't try to start to migrate some instance types away from Xen in the future.

tests/kola/platforms/aws/nvme
dustymabe: fixed

tests/kola/platforms/aws/assert-xen
dustymabe: fixed

tests/kola/platforms/aws/nvme
jlebon: > No. In this case the instance storage isn't the boot disk. It's just an extra disk.

Ahh gotcha. Should we extend this test to more generally check that every NVMe storage device found has a device node?

dustymabe: I think this is where my knowledge might be limited. Are all devices on an NVMe controller storage devices? Is there a way in that json output in I pasted in the file to see an entry is for storage and go based on that?

I admittedly was focused minimal on a test that would check https://github.com/coreos/fedora-coreos-tracker/issues/1306

jlebon: > I think this is where my knowledge might be limited. Are all devices on an NVMe controller storage devices?

AFAIK that's the case, yes. But I'm good going with what we have for now and expanding it in the future.

jlebon (approved):

PR #2006, opened 2022-10-07
[rawhide] overrides: pin on bash 5.1.16-4.fc38

gursewak1997:
Ref Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1313

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,9 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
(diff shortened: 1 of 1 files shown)


manifest-lock.overrides.yaml
```diff
@@ -8,4 +8,9 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  bash:
+    evr: 5.1.16-3.fc36
```
jlebon: This is an f36 RPM. It should be an f38 RPM for rawhide.

jlebon (approved):
Thanks!

PR #2008, opened 2022-10-10
[next-devel] overrides: fast-track kernel-5.19.14-300.fc37

coreosbot-releng:


lucab (approved):

dustymabe:
The upgrade tests are going to fail because of some issue with the pipeline that should be resolved with the next `next-devel` build. I hacked up the CI run here to not run upgrade tests.

PR #2015, opened 2022-10-17
overrides: drop bash pin

gursewak1997:
Unpin bash now that the issue https://github.com/coreos/fedora-coreos-tracker/issues/1313 is fixed in the latest package.


bgilbert (approved):
LGTM

PR #2019, opened 2022-10-17
image-base: bump VMware hardware version to 17

### LINKED ISSUE #1141: Update hardware version in VMware OVA after 6.5/6.7 EOL date
(Split off of https://github.com/coreos/fedora-coreos-tracker/issues/1119; please see that issue for relevant discussion)

VMware vSphere 6.5/6.7 is scheduled to be EOL on Oct 15, 2022.  This provides an opportunity to change the hardware version that we use by default to something newer, which would effectively block users on those older versions of vSphere from booting FCOS as a guest VM.

We should evaluate changing the hardware version to something newer than 13 (the current default) after the EOL date of vSphere 6.5/6.7

A reasonable new version would be ~16~ 17; see the KB article for more details about hw versions - https://kb.vmware.com/s/article/1003746

bgilbert:
The last VMware products requiring older hardware versions are now EOL.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1141.

### DIFF
--- a/image-base.yaml
+++ b/image-base.yaml
@@ -30,4 +30,4 @@ vmware-os-type: fedora64Guest
 # VMware hardware versions: https://kb.vmware.com/s/article/1003746
 # We use the newest version allowed by the oldest non-EOL VMware
 # Workstation/Player/Fusion/ESXi release: https://lifecycle.vmware.com/
-vmware-hw-version: 13
+vmware-hw-version: 17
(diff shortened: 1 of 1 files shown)


bgilbert:
Boot tested on VMware Workstation 16.2.4.

dustymabe (approved):
LGTM

PR #2024, opened 2022-10-17
ci: add a cpu and memory request

jlebon:
We're not currently requesting anything, which means we're at the mercy of any default limit set on the cluster we're running on. Let's add an explicit request.

Also bump the CPU request to 4 from the default of 2.


jlebon:
I'm still debating whether to add a default memory request to `cosaPod`... It's tempting, but at the same time, not all upstreams run `kolaTestIso` and 8.5G requests quickly add up. Open to other opinions.

bgilbert (approved):
LGTM

dustymabe (approved):
LGTM

PR #2030, opened 2022-10-23
[next-devel] overrides: fast-track kernel-5.19.16-301.fc37

coreosbot-releng:


dustymabe:
added a second commit for fast tracking samba subpackages, which is also in the [RC request](https://pagure.io/releng/issue/11094#comment-822428)

jlebon (approved):

PR #2031, opened 2022-10-24
tests: verify multiple nameservers config via kernel arguments work well

HuijingHei:


tests/kola/networking/nameserver
```diff
@@ -0,0 +1,27 @@
...
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# check nameserver in /etc/resolv.conf
+resolv=/etc/resolv.conf
+if ! (grep -q "nameserver 8.8.8.8" ${resolv} && \
+      grep -q "nameserver 1.1.1.1" ${resolv}); then
+    fatal "Error: can not find nameserver in ${resolv}"
```
jlebon: Can we `cat ${resolv}` before this line so we can tell what the actual file looked like?

tests/kola/networking/nameserver
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+# check nameserver in config file
+conf=/etc/NetworkManager/system-connections/default_connection.nmconnection
+if ! grep -q "dns=8.8.8.8;1.1.1.1;"; then
+    fatal "Error: can not find nameserver in ${conf}"
```
jlebon, line 36: Similarly here.

tests/kola/networking/nameserver
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+ok "multipule nameserver"
```
jlebon: Minor: multiple

jlebon (dismissed):
That's a nice test idea. Minor nits, but LGTM overall. I opened https://github.com/coreos/fedora-coreos-config/pull/2032 for the CI failure.

dustymabe:
This test isn't being tested in CI:

```
kola: Found non-executable file with shebang: nameserver
```

You need to make it executable in git with a `chmod +x file && git add file`.

tests/kola/networking/nameserver
```diff
@@ -0,0 +1,27 @@
...
+#   - This test should pass everywhere if it passes anywhere.
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# check nameserver in /etc/resolv.conf
+resolv=/etc/resolv.conf
+if ! (grep -q "nameserver 8.8.8.8" ${resolv} && \
+      grep -q "nameserver 1.1.1.1" ${resolv}); then
```
dustymabe: I don't think this test is going to work on FCOS where we have systemd-resolved. Maybe we can modify the test to work on both FCOS and RHCOS. on FCOS you should be able to run `resolvectl` to see what nameservers are configured.

dustymabe (changes requested):
Found a few things that need to be changed.

tests/kola/networking/nameserver
dustymabe: This isn't resolved. See the test failure in CI.

HuijingHei:
> Found a few things that need to be changed.

Done, thanks!

tests/kola/networking/nameserver
```diff
@@ -0,0 +1,39 @@
...
+        grep -q "nameserver 1.1.1.1" ${resolv}); then
+        fatal "Error: can not find nameserver in ${resolv}"
+    fi
+
+    # check nameserver in config file
+    conf=/etc/NetworkManager/system-connections/default_connection.nmconnection
+    cat ${conf}
+    if ! grep -q "dns=8.8.8.8;1.1.1.1;" ${conf}; then
+        fatal "Error: can not find nameserver in ${conf}"
+    fi
```
dustymabe: I think this part can be checked on both RHCOS and FCOS, right?

cgwalters (approved):

dustymabe (approved):

PR #2032, opened 2022-10-24
ci: bump memory request to 9Gi

jlebon:
Noticed in https://github.com/coreos/fedora-coreos-config/pull/2031 that the `kola testiso` was killed:

    FAIL: iso-offline-install (bios + metal) (1m35.505s)
        signal: killed
    Error: signal: killed

This is likely due to hitting up against the memory limit. It seems like 512Mi of overhead isn't enough. Let's bump it to 1Gi.


bgilbert (approved):

dustymabe (approved):
LGTM

PR #2034, opened 2022-10-25
denylist: extend snooze for ext.config.platforms.aws.nvme test

dustymabe:
It looks like our F37+ streams pass this test now [1] so let's also only deny the test on streams where it's known to fail.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1306#issuecomment-1289910315


bgilbert (approved):
LGTM

HuijingHei (approved):

PR #2038, opened 2022-10-28
[next-devel] platforms: add secondary serial console on OpenStack, VirtualBox, VMware

bgilbert:
Previously, `openstack console log <server>` was useful for collecting console logs from OpenStack VMs, and FCOS docs included instructions on connecting to the VirtualBox/VMware serial console to get a console log.  However, we've disabled the serial console on all three platforms.  Re-enable it as a secondary console so those techniques continue to work while avoiding interfering with the primary graphical console.

The OpenStack change is untested.

Fixes: [OCPBUGS-2926](https://issues.redhat.com/browse/OCPBUGS-2926)
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/567#issuecomment-1271536899

### DIFF
--- a/platforms.yaml
+++ b/platforms.yaml
@@ -143,3 +143,23 @@ x86_64:
     kernel_arguments:
       - console=tty0
       - console=ttyS0,115200n8
+  virtualbox:
+    # Graphical console primary, serial console available for logging
+    # https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-virtualbox/#_troubleshooting_first_boot_problems
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=ttyS0,115200n8
+      - console=tty0
+  vmware:
+    # Graphical console primary, serial console available for logging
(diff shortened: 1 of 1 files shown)


travier (dismissed):

cgwalters:
Hmm I'm a little confused honestly, it seems like we're getting to a point where we're enabling the console on pretty much every platform *except* metal?    If that's getting to be the case I wonder if we should change things to go back to more having a default console setup that can be overridden per-platform - in particular, we have the same kernel arguments across aws/azure/gcp for example.

Anyways, seems OK to me!

dustymabe (dismissed):
LGTM

Is there somewhere we can link to for these new platforms that tells us what the values should be for a serial console on these platforms (similar to the docs we link to for the original entries in this file)?

jlebon (approved):

bgilbert:
@cgwalters Some platforms want serial console only, some want graphical console only, and some want both with serial console either primary or secondary.  Since enabling serial console requires us to be thoughtful about how we configure it, I think the current default is useful to ensure we do that.

@dustymabe I've added a link for OpenStack.  VirtualBox and VMware both just emulate standard PCs, and how we use the serial ports is up to us, so the links to our docs are in some sense the authoritative ones.

dustymabe (approved):

PR #2039, opened 2022-10-28
denylist: extend snooze for tests affected by tmpfiles issue

dustymabe:
The systemd fix has landed but we've got it pinned right now for another issue so we still need to snooze this one. See
https://github.com/coreos/fedora-coreos-tracker/issues/1215#issuecomment-1295021109


bgilbert (approved):
LGTM

PR #2043, opened 2022-11-01
denylist: exclude ext.config.networking.nameserver on s390x

dustymabe:
This test was recently introduced and fails on s390x. See https://github.com/coreos/fedora-coreos-tracker/issues/1332


dustymabe:
I'll skip CI here since CI won't test this change.

jlebon (approved):

PR #2048, opened 2022-11-01
overlay/15fcos: retroactively fix BLS `grub_users` setting for CVE-2022-3675

bgilbert:
Starting with FCOS 36.20220906.1.0, 36.20220906.2.0, and 36.20220820.3.0, coreos-assembler inadvertently failed to configure ostree to set `grub_users=""` in non-default BLS configs, allowing old deployments to be booted without a GRUB password.  Add a service that fixes this setting on the first boot after upgrade, if the aleph version corresponds to an affected release.  This service can be removed after the next update barrier in all streams.

Add a CI test that fails if `if $` or `if "$` appears in a shell script, since `if [ $` may have been intended and the test will always return false if so.  This should help prevent similar future bugs in fedora-coreos-config.  The test will generate false positives if a command is being run indirectly via a shell variable, but we can work around that later if needed.

Add a kola test for the Butane GRUB users sugar, since we didn't have one before.

Requires https://github.com/coreos/coreos-assembler/pull/3155.  For https://github.com/coreos/fedora-coreos-tracker/issues/1333.


overlay.d/15fcos/usr/libexec/coreos-fix-grub-users
```diff
@@ -0,0 +1,57 @@
...
+
+STAMP=/var/lib/coreos/fix-grub-users.stamp
+
+aleph_ver=$(jq -r .build < /sysroot/.coreos-aleph-version.json)
+date=$(echo "$aleph_ver" | cut -f2 -d.)
+stream=$(echo "$aleph_ver" | cut -f3 -d.)
+
+if [ "$stream" = "3" ]; then
+    start=20220819
+    end=20221012
```
jlebon, line 16: Did you just pick a number between `20221001` (current `stable`) and `20221014` (current `testing`, which will be promoted to `stable` tomorrow)?

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-grub-users.service
```diff
@@ -0,0 +1,19 @@
+# Remove after the next update barrier in all streams.
+
+[Unit]
+Description=Fix ostree grub_users Setting
```
jlebon, line 4: Can we add a `Documentation` here with the tracker issue?

jlebon (dismissed):
One suggestion, but LGTM as is.
Thank you *so much* for working on this! :heart:

overlay.d/15fcos/usr/libexec/coreos-fix-grub-users
bgilbert: Yeah, I was trying to avoid off-by-one errors.  The corresponding test checks the versions we're interested in.

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-grub-users.service
bgilbert: Yup, done.

jlebon (approved):

PR #2050, opened 2022-11-02

coreosbot-releng:


lucab (dismissed):
This is currently causing a downgrade and dropping the punycode security fixes.

dustymabe:
> This is currently causing a downgrade and dropping the punycode security fixes.

maybe mirroring issue where the GH action was seeing one thing and the CI run saw something different? At least we have the check :)

dustymabe (dismissed):

dustymabe:
actually - this won't pass until the next run of `bump-lockfile` it makes sense now.

dustymabe (approved):
LGTM

PR #2056, opened 2022-11-02
tests/kola/selinux: add test that policy isn't recompiled

jlebon:
This is a test for https://github.com/openshift/os/issues/1036. It also exists in the rpm-ostree CI, but let's have it here too since other packages can break this.

### DIFF
--- a/tests/kola/selinux/default
+++ b/tests/kola/selinux/default
@@ -1,0 +1,15 @@
+#!/bin/bash
+# Verify that the SELinux policy isn't marked as modified.
+# See: https://github.com/openshift/os/issues/1036.
+## kola:
+##   exclusive: false
+##   tags: "platform-independent"
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+if ostree admin config-diff | grep 'selinux/targeted/policy'; then
+    fatal "SELinux policy is marked as modified"
+fi
+ok "SELinux policy not marked as modified"


jlebon:
Let's hold this until [the rpm-ostree patch](https://github.com/coreos/rpm-ostree/pull/4122) is in f37.

jlebon:
Actually, all we need is for the rpm-ostree patch to get into cosa, which it is now. So this is ready for review!

tests/kola/selinux/default
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# kola
+#   exclusive: false
+#   # if it works on QEMU, it should work everywhere
+#   platforms: qemu
```
dustymabe:
```suggestion
## kola:
##   exclusive: false
##   tags: "platform-independent"
```

jlebon: Done!

cgwalters: Heh, we may need some sort of test for the test syntax...

tests/kola/selinux/default
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# kola
+#   exclusive: false
+#   tags: "platform-independent"
```
dustymabe: I think we still need some updates here. IIUC we need two `##` at the beginning of the line and the first line needs to be `## kola:` (with a `:`). Am I missing something?

```suggestion
## kola:
##   exclusive: false
##   tags: "platform-independent"
```

dustymabe: may also be nice to get a  description of the problem and maybe a link to https://github.com/openshift/os/issues/1036?

I also wonder if we shouldn't have an FCOS issue tracker issue for this problem.

jlebon: OK, I followed the format used in the one other test that uses `platform-independent`.

dustymabe: You mean this one:

https://github.com/coreos/fedora-coreos-config/blob/010ad3d732241c8f2eb19133511c325d34503b52/tests/kola/content-origins/test.sh#L3-L8

tests/kola/selinux/default
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Verify that the SELinux policy isn't marked as modified.
+# See: https://github.com/openshift/os/issues/1036
+## kola
```
dustymabe:
```suggestion
## kola:
```

Is this not needed?

tests/kola/selinux/default
jlebon: Added link back to https://github.com/openshift/os/issues/1036. Not sure it's worth the FCOS issue tracker at this point. Were you thinking for release notes?

dustymabe (approved):
LGTM - thanks for writing this test!

tests/kola/selinux/default
dustymabe: Mostly just for a record that this affected `next` stream nodes. I'm not too worried about it, though. Fine without it.

PR #2065, opened 2022-11-04
tests/kola/butane/grub-users-fix: increase timeout again

bgilbert:
15 minutes still isn't reliably enough for ppc64le.

### DIFF
--- a/tests/kola/butane/grub-users-fix
+++ b/tests/kola/butane/grub-users-fix
@@ -6,8 +6,8 @@
 ##  architectures: "!s390x"
 ##  # Running on multiple platforms won't prove anything further
 ##  platforms: qemu
-##  # 10 minutes is not quite enough on ppc64le runners
-##  timeoutMin: 15
+##  # 15 minutes is not quite enough on ppc64le runners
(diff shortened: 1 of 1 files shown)


tests/kola/butane/grub-users-fix
```diff
@@ -6,8 +6,8 @@
 ##  architectures: "!s390x"
 ##  # Running on multiple platforms won't prove anything further
 ##  platforms: qemu
-##  # 10 minutes is not quite enough on ppc64le runners
-##  timeoutMin: 15
+##  # 10 minutes is not enough on ppc64le runners
```
dustymabe: should we update this to say 15?

dustymabe (dismissed):
LGTM

tests/kola/butane/grub-users-fix
bgilbert: 10 minutes is the default.

bgilbert: Oh, I see what you're saying.  Sure.

dustymabe (approved):
LGTM

PR #2073, opened 2022-11-09
networking/tests: add no-persist-ip and rd-net-timeout-carrier cases

HuijingHei:
- Verify `coreos.no_persist_ip` karg will disable initramfs network propagation
- Verify nm-initrd-generator supports `rd.net.timeout.carrier=`

### DIFF
--- a/tests/kola/networking/no-persist-ip
+++ b/tests/kola/networking/no-persist-ip
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# This test verifies that the coreos.no_persist_ip kernel argument will
+# prevent propagating kernel argument based networking configuration into
+# the real root. It does this by providing karg static networking config
+# for eth1 and then verifying that DHCP is used in the real root.
+
+## kola:
+##   # This test should pass everywhere if it passes anywhere.
+##   platforms: qemu
+##   # Add 1 NIC for this test
+##   additionalNics: 1
+##   # The functionality we're testing here and the configuration for the NIC.
+##   # We use net.ifnames=0 to disable consistent network naming here because on
+##   # different firmwares (BIOS vs UEFI) the NIC names are different.
+##   # See https://github.com/coreos/fedora-coreos-tracker/issues/1060
+##   appendKernelArgs: "ip=10.10.10.10::10.10.10.1:255.255.255.0:myhostname:eth1:none net.ifnames=0 coreos.no_persist_ip"
+##   # appendKernelArgs doesn't work on s390x, see https://github.com/coreos/coreos-assembler/issues/2776
+##   architectures: !s390x
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# Verify eth1 gets ip address via dhcp
+nic_name="eth1"
+nic_ip=$(get_ipv4_for_nic ${nic_name})
+if [ ${nic_ip} != "10.0.2.31" ]; then
+    fatal "Error: get ${nic_name} ip = ${nic_ip}, expected is 10.0.2.31"
+fi
+ok "get ${nic_name} ip is 10.0.2.31"
+
+if ! journalctl -b -u coreos-teardown-initramfs | grep -q "info: skipping propagating initramfs settings"; then
+    fatal "Error: can not get log: (info: skipping propagating initramfs settings)"
+fi
+ok "test coreos.no_persist_ip disable initramfs network propagation"
--- a/tests/kola/networking/rd-net-timeout-carrier/config.bu
+++ b/tests/kola/networking/rd-net-timeout-carrier/config.bu
@@ -1,0 +1,6 @@
+variant: fcos
+version: 1.4.0
+kernel_arguments:
+  should_exist:
+    - rd.net.timeout.carrier=15
+    - rd.neednet=1
--- a/tests/kola/networking/rd-net-timeout-carrier/test.sh
+++ b/tests/kola/networking/rd-net-timeout-carrier/test.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# This test verifies that NetworkManager supports configuring the
+# carrier timeout via the `rd.net.timeout.carrier=` karg. Without
+# recreating an environment that requires this setting to be set
+# (which would be hard to do), we'll test this by just making sure
+# that setting the kernel argument ensures the runtime configuration
+# file /run/NetworkManager/conf.d/15-carrier-timeout.conf gets created.
+# 
+# Checking that this file gets created is also tricky because we
+# `rm -rf /run/NetworkManager` in coreos-teardown-initramfs on first boot,
+# so we'll workaround that by setting the karg permanently and then
+# performing a reboot so we can check on the second boot if the file exists.
+#
+# See:
+# - https://bugzilla.redhat.com/show_bug.cgi?id=1917773
+# - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/e300138892ee0fc3824d38b527b60103a01758ab#
+
+## kola:
+##   # This test should pass everywhere if it passes anywhere.
+##   platforms: qemu-unpriv
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+  "")
+      ok "first boot"
+      /tmp/autopkgtest-reboot rebooted
+      ;;
+  
+  rebooted)
+      ok "second boot"
+      grep rd.net.timeout.carrier=15 /proc/cmdline
+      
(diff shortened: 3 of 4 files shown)


gursewak1997 (dismissed):

dustymabe:
I'll review this soon. Sorry for the delay.

tests/kola/networking/rd-net-timeout-carrier
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
dustymabe: I'm not sure how valuable this test is as written. We're mostly making sure that `nm-initrd-generator` supports writing out an appropriate config, which should be true for any future release of NM. 

It would be a better test to set this `rd.net.timeout.carrier=15` on boot and actually watch it take effect in an evironment where it is needed, but that environment is harder to set up in an automated fashion.

Maybe an in between would be to set this parameter on the kernel command line for the test and then verify the config gets written out. The issue with that is that we will [clean it up](https://github.com/coreos/fedora-coreos-config/blob/94dde18f1c3db963100653fa1a4dee14911b191e/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh#L234). A workaround for that is to set the kernel arguments permanently for the VM using [butane kernel args](https://github.com/coreos/fedora-coreos-config/blob/94dde18f1c3db963100653fa1a4dee14911b191e/tests/kola/kdump/crash/config.bu#L3) and then do a reboot (see [here](https://github.com/coreos/fedora-coreos-config/blob/94dde18f1c3db963100653fa1a4dee14911b191e/tests/kola/kdump/crash/test.sh) for an example of how) and then verify on the 2nd boot (we won't run the cleanup on that boot) that `/run/NetworkManager/conf.d/15-carrier-timeout.conf` exists and matches expected output.

tests/kola/networking/no-persist-ip
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# kola: { "platforms": "qemu", "additionalNics": 1, "appendKernelArgs": "ip=10.10.10.10::10.10.10.1:255.255.255.0:myhostname:eth1:none net.ifnames=0 coreos.no_persist_ip", "architectures": "!s390x"}
```
dustymabe: We now support yaml for this with inline comments. It just needs to be preceded with two `##`. See [here](https://github.com/coreos/fedora-coreos-config/blob/94dde18f1c3db963100653fa1a4dee14911b191e/tests/kola/platforms/aws/assert-xen#L4-L13) for an example.

tests/kola/networking/rd-net-timeout-carrier
HuijingHei: Thanks for the detail info, will change according to the comments.

tests/kola/networking/no-persist-ip
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
```
dustymabe, line 2: Let's add the description of the test up here. Something like:

```
# This test verifies that the coreos.no_persist_ip kernel argument will
# prevent propagating kernel argument based networking configuration into
# the real root. It does this by providing karg static networking config
# for eth1 and then verifying that DHCP is used in the real root.
```

tests/kola/networking/no-persist-ip
```diff
@@ -0,0 +1,37 @@
...
+## kola:
+##   # Setup configuration for a single NIC:
+##   # kargs provide static network config for eth1 and coreos.no_persist_ip
+##   # Expected result:
+##   # coreos.no_persist_ip karg will disable initramfs network propagation, 
+##   # verify that eth1 gets ip via dhcp
```
dustymabe: We can delete this if we add the description at the top.

tests/kola/networking/no-persist-ip
```diff
@@ -0,0 +1,37 @@
...
+##   platforms: qemu
+##   # This test should pass everywhere if it passes anywhere.
```
dustymabe: Move the descriptions of the settings here to above where the setting is defined:

```suggestion
##   # This test should pass everywhere if it passes anywhere.
##   platforms: qemu
```

Do this for `additionalNics` `appendKernelArgs` `architecture` below too.

tests/kola/networking/rd-net-timeout-carrier/config.bu
```diff
@@ -0,0 +1,5 @@
+variant: fcos
+version: 1.4.0
+kernel_arguments:
+  should_exist:
+    - rd.net.timeout.carrier=15
```
dustymabe, line 5: Hmm. I would have thought you would have needed `rd.neednet=1` too:

```suggestion
    - rd.net.timeout.carrier=15
    - rd.neednet=1
```

Did you not need that?

tests/kola/networking/rd-net-timeout-carrier/test.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# kola: {"platforms": "qemu-unpriv"}
```
dustymabe: optional: convert this to yaml too like we did the other file.

tests/kola/networking/rd-net-timeout-carrier/test.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Verify that nm-initrd-generator supports `rd.net.timeout.carrier=`,
+# this will create a new `15-carrier-timeout.conf` file in
+# /run/NetworkManager/conf.d with the parameter value as specified.
+# See:
+# - https://bugzilla.redhat.com/show_bug.cgi?id=1917773
+# - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/e300138892ee0fc3824d38b527b60103a01758ab#
```
dustymabe: Since this test is doing something non-obvious (i.e. adding permanent kernel arguments and forcing a reboot in order to test a feature of NetworkManager) let's add more context here:

```
# This test verifies that NetworkManager supports configuring the
# carrier timeout via the `rd.net.timeout.carrier=` karg. Without
# recreating an environment that requires this setting to be set
# (which would be hard to do), we'll test this by just making sure
# that setting the kernel argument ensures the runtime configuration
# file /run/NetworkManager/conf.d/15-carrier-timeout.conf gets created.
# 
# Checking that this file gets created is also tricky because we
# rm -rf /run/NetworkManager in coreos-teardown-initramfs on first boot
# so we'll workaround that by setting the karg permanently and then
# performing a reboot so we can check on the second boot if the file exists.
#
# See:
# - https://bugzilla.redhat.com/show_bug.cgi?id=1917773
# - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/e300138892ee0fc3824d38b527b60103a01758ab#
```

tests/kola/networking/rd-net-timeout-carrier/test.sh
```diff
@@ -0,0 +1,39 @@
...
+      expected_conf="[device-15-carrier-timeout]
+match-device=*
+carrier-wait-timeout=15000"
+
+      # remove empty and comment lines
+      generator_conf=$(cat /run/NetworkManager/conf.d/15-carrier-timeout.conf | grep -v "^[[:space:]]*$" | grep -v "^#")
+
+      if ! diff -u <(echo "$expected_conf") <(echo "$generator_conf"); then
+        fatal "Error: can not generate ${generator_conf} with karg rd.net.timeout.carrier"
+      fi
```
dustymabe: It might be simpler (less headache) over time if we just check if the file exists rather than verifying the contents of the file match exactly what we expect.

tests/kola/networking/rd-net-timeout-carrier/test.sh
```diff
@@ -0,0 +1,39 @@
...
+      ok "Test nm-initrd-generator supports rd.net.timeout.carrier"
```
dustymabe: optional wording enhancement:

```suggestion
      ok "NM generated rd.net.timeout.carrier configuration in the initramfs"
```

tests/kola/networking/rd-net-timeout-carrier/config.bu
dustymabe: Oh yes. `nm-initrd-generator` runs [regardless of if network is needed or not](https://github.com/dracutdevs/dracut/blob/fd9cd02cb21f25d32fec4cd9889457ea6361f45e/modules.d/35network-manager/nm-lib.sh#L7-L12), but the network is [only brought up IF it is needed](https://github.com/dracutdevs/dracut/blob/fd9cd02cb21f25d32fec4cd9889457ea6361f45e/modules.d/35network-manager/nm-lib.sh#L17).

Let's add `rd.neednet=1` here so we don't rely on that fact.

HuijingHei:
Updated, thanks @dustymabe for the review and suggestion!

dustymabe (approved):
LGTM - thanks for working on this.

PR #2077, opened 2022-11-10
tests/validate-symlinks: Add RHCOS logic

c4rt0:
Modified validate-symlinks kola test logic related to the https://github.com/openshift/os/issues/1030
following a discussion https://github.com/openshift/os/issues/1003

### DIFF
--- a/tests/kola/files/validate-symlinks
+++ b/tests/kola/files/validate-symlinks
@@ -1,5 +1,5 @@
 #!/bin/bash
-# kola: { "exclusive": false, "distros": "fcos", "platforms": "qemu" }
+# kola: { "exclusive": false, "platforms": "qemu" }
 #
 # Check if there are broken symlinks in /etc/ and /usr/.
 # See: https://github.com/coreos/fedora-coreos-config/issues/1782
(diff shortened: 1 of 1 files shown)


tests/kola/files/validate-symlinks
```diff
@@ -28,8 +28,30 @@ list_broken_symlinks_skip=(
     '/usr/share/rhel/secrets/rhsm'
 )
 
-find /usr/ /etc/ -type l -not -path "/usr/etc*"| while read -r file_name; do
-    real_path=$(realpath -m "${file_name}");
+is_rhcos="false"
+
+# Check if RHCOS
+is_rhcos() {
```
travier: You should use this function from the commonlib.sh instead of redefining it

tests/kola/files/validate-symlinks
```diff
@@ -28,8 +28,30 @@ list_broken_symlinks_skip=(
...
+    source /etc/os-release
+    [ "${ID}" == "rhcos" ]
+}
+
+# If RHCOS, update the array of ignored symlinks
+update_ignored_symlinks() {
```
travier: Let's not create a new function and add the code inline

travier (changes requested):
Looks like a good start. Some comments

travier:
Let's use something like that for the commit title and message:

```
tests/validate-symlinks: Add support for RHCOS

Add RHCOS specific list of broken symlinks to ignore. We will investigate them later.
Needs manifest changes from openshift/os PR.

See: openshift/os issue.
```

travier (approved):

PR #2083, opened 2022-11-11
manifest: Move to Fedora Linux 37 content

dustymabe:
Also seeded the architecture specific lockfiles from `next-devel`.

### DIFF
--- a/ci/buildroot/Dockerfile
+++ b/ci/buildroot/Dockerfile
@@ -5,6 +5,6 @@
 #
 # This image is used by CoreOS CI to build software like
 # Ignition, rpm-ostree, ostree, coreos-installer, etc...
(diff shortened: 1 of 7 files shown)


jmarrero (dismissed):
lgtm

dustymabe:
Looks like CI will fail on this until next tuesday when F37 is released because the yum repos aren't quite ready yet.

dustymabe:
ok CI should pass on this now. @jmarrero, can I get another round of review?

jmarrero (approved):

dustymabe:
```
echo 'Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.'
```

The one package that is being downgraded isn't a concern. I've sent a message to the package maintainer to see if we can get it pushed through in F37/Bodhi.

PR #2097, opened 2022-11-21
image-base.yaml: set `compressor: xz`

jlebon:
This will be picked up by `cosa compress` to know to compress with xz.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3215

dustymabe (approved):
LGTM

PR #2098, opened 2022-11-22
tests: verify using the target system's grub2-install to install bootloader

HuijingHei:
See https://github.com/coreos/coreos-assembler/pull/3190#issuecomment-1322147805

### DIFF
--- a/tests/kola/boot/grub2-install
+++ b/tests/kola/boot/grub2-install
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# Verify to install BIOS/PReP bootloader using grub2-install from 
+# the target system.
+# See:
+#  - https://github.com/coreos/coreos-assembler/pull/3190
+#  - https://github.com/coreos/coreos-assembler/issues/3148
+
+## kola:
+##   # This test should pass everywhere if it passes anywhere.
+##   platforms: qemu
+##   # The test is not available for aarch64 and s390x,
+##   # as aarch64 is UEFI only and s390x is not using grub2
+##   architectures: "!aarch64 !s390x"
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+boot_dev=/boot
+arch=$(arch)
+case ${arch} in
+  x86_64)
+    target=i386-pc
+    core="${boot_dev}/grub2/${target}/core.img"   
+    partition=$(findmnt -no SOURCE ${boot_dev})
+    device=$(lsblk --paths -no PKNAME ${partition})
+    # sync grub2-install parameter according to image build script
+    # https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh
+    grub_install_args="--modules mdraid1x"
+    ;;
+  
+  ppc64le)
+    target=powerpc-ieee1275
+    core="${boot_dev}/grub2/${target}/core.elf"
+    device=$(realpath /dev/disk/by-partlabel/PowerPC-PReP-boot)
+    # sync grub2-install parameter according to image build script
+    # https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh    
+    grub_install_args="--no-nvram"   
+    ;;
+
+  *)
+    fatal "grub2-install testing is not supported on ${arch}"
+    ;;
+esac
+
+[ -e ${core} ] || fatal "file ${core} doesn't exist"
+core_sum=$(sha256sum ${core} | awk '{print $1}')
+
+mount -o remount,rw ${boot_dev}
+grub2-install --target ${target} --boot-directory ${boot_dev} ${grub_install_args} ${device}
+
+new_core_sum=$(sha256sum ${core} | awk '{print $1}')
+
+# compare core.img checksum before and after run grub2-install
+if [ "${core_sum}" != "${new_core_sum}" ]; then
+  fatal "Error: not using grub2-install from target system"
+fi
+ok "using grub2-install from target system on ${arch}"


travier (commented):
Weak LGTM as I did not follow those changes.

travier:
Can you test this one on RHCOS too?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+# kola: { "exclusive": false }
```
cgwalters: This will run concurrently with other tests, but...

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
...
+  x86_64)
+    core=/boot/grub2/i386-pc/core.img
+    boot_dev=/boot
+    device=/dev/vda
+
+    [ -e ${core} ]
+    core_size=$(stat --printf="%s" ${core})
+
+    mount -o remount,rw ${boot_dev}
+    grub2-install --target i386-pc --boot-directory ${boot_dev} --modules mdraid1x ${device}
```
cgwalters: This is a global system-mutating action which seems like it would at least potentially conflict with any other nonexclusive tests which want to e.g. introspect `/boot` state.

I'd say the simplest fix is to just drop `exclusive: false`.

cgwalters (dismissed):
Looks sane to me, just one nit.

HuijingHei:
> Can you test this one on RHCOS too?

Build rhcos locally and run, result is passed.
```
$ kola run --qemu-image rhcos/builds/413.86.202211220853-0/x86_64/rhcos-413.86.202211220853-0-qemu.x86_64.qcow2 -E config ext.config.boot.grub2-install
=== RUN   ext.config.boot.grub2-install
--- PASS: ext.config.boot.grub2-install (55.41s)
PASS, output in _kola_temp/qemu-unpriv-2022-11-22-1332-104
```

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
jlebon, line 1: We can limit this to `platforms: qemu`, no?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
...
+
+# Make sure to install BIOS bootloader using the target system's grub2-install
+# See https://github.com/coreos/coreos-assembler/pull/3190
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+arch=$(arch)
+case ${arch} in
+  x86_64)
```
jlebon, line 22: Instead of this, we can do `architectures: x86_64`.

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
...
+    [ -e ${core} ]
+    core_size=$(stat --printf="%s" ${core})
+
+    mount -o remount,rw ${boot_dev}
+    grub2-install --target i386-pc --boot-directory ${boot_dev} --modules mdraid1x ${device}
+
+    generator_core_size=$(stat --printf="%s" ${core})
+
+    # compare core.img size before and after call grub2-install
+    if [ ${core_size} -ne ${generator_core_size} ]; then
```
jlebon: Should we be comparing checksums instead?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
...
+      fatal "Error: not using target system's grub2-install"
+    fi
```
jlebon: For all the other files in `/boot/grub2/i386-pc`, should we check that their checksums match the versions in `/usr/lib/grub/i386-pc`?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
...
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+arch=$(arch)
+case ${arch} in
+  x86_64)
+    core=/boot/grub2/i386-pc/core.img
+    boot_dev=/boot
+    device=/dev/vda
```
jlebon: Minor: you can use `findmnt -no SOURCE ${boot_dev}` to avoid hardcoding this.

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
...
+    generator_core_size=$(stat --printf="%s" ${core})
+
+    # compare core.img size before and after call grub2-install
+    if [ ${core_size} -ne ${generator_core_size} ]; then
+      fatal "Error: not using target system's grub2-install"
+    fi
+    ok "ok using target system's grub2-install on ${arch}"
+    ;;
+  *)
+    echo "skip grub2-install testing on ${arch}"
```
dustymabe: hmm. [The problem](https://github.com/coreos/coreos-assembler/issues/3148) that caused us to uncover the fact that we were using the grub from Fedora and thus lead to this correction of appropriate behavior was against `ppc64le`. Should we not test at least on that architecture too?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,33 @@
...
+
+arch=$(arch)
+case ${arch} in
+  x86_64)
+    core=/boot/grub2/i386-pc/core.img
+    boot_dev=/boot
+    device=/dev/vda
+
+    [ -e ${core} ]
+    core_size=$(stat --printf="%s" ${core})
```
dustymabe: Is size a good enough metric here? Since hashing tools are readily available to us then why not use something like `sha256sum` instead?

There may be a good reason to not use a checksum utility here, but including that rationale in a comment would be useful in general.

tests/kola/boot/grub2-install
HuijingHei: I think you are right, will add it, thanks

tests/kola/boot/grub2-install
HuijingHei: Just check some files, many are different (with size) in the following list. 
Maybe only need to check some files to verify, instead of running `grub2-install` and check checksum of `core.img`
```
[root@cosa-devsh ~]# ls /boot/grub2/i386-pc/ -l | head -n 10
total 2954
-rw-r--r--. 1 root root  13252 Nov  8 23:21 acpi.mod
-rw-r--r--. 1 root root   1428 Nov  8 23:21 adler32.mod
-rw-r--r--. 1 root root   7072 Nov  8 23:21 affs.mod
-rw-r--r--. 1 root root   9076 Nov  8 23:21 afs.mod
-rw-r--r--. 1 root root   1788 Nov  8 23:21 afsplitter.mod
-rw-r--r--. 1 root root  21532 Nov  8 23:21 ahci.mod
-rw-r--r--. 1 root root    492 Nov  8 23:21 all_video.mod
-rw-r--r--. 1 root root   1068 Nov  8 23:21 aout.mod
-rw-r--r--. 1 root root  14852 Nov  8 23:21 appended_signature_test.mod

[root@cosa-devsh ~]# ls /usr/lib/grub/i386-pc -l | head -n 10
total 3568
-rw-r--r--. 2 root root  13380 Jan  1  1970 acpi.mod
-rw-r--r--. 2 root root   1524 Jan  1  1970 adler32.mod
-rw-r--r--. 2 root root   7248 Jan  1  1970 affs.mod
-rw-r--r--. 2 root root   9108 Jan  1  1970 afs.mod
-rw-r--r--. 2 root root  21644 Jan  1  1970 ahci.mod
-rw-r--r--. 2 root root    492 Jan  1  1970 all_video.mod
-rw-r--r--. 2 root root   1148 Jan  1  1970 aout.mod
-rw-r--r--. 2 root root  14980 Jan  1  1970 appended_signature_test.mod
-rw-r--r--. 2 root root  45368 Jan  1  1970 appendedsig.mod
```

HuijingHei: There are 2 options to verify using the target system's grub2-install to install bootloader:
1) Check files in `/boot/grub2/i386-pc`, that their checksums match the versions in `/usr/lib/grub/i386-pc`
- The upside is nothing to change, just compare the files checksum.
- The downside is there are 284 files include core.img, maybe only pick some files.

2) Run `grub2-install` to create new `core.img` and compare checksum
- The upside is only need to check core.img.
- The downside is need to remount /boot with rw.

Which option is better?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,61 @@
...
+    
+    [ -e ${core} ]
+    core_sum=$(sha256sum ${core} | awk '{print $1}')
+
+    mount -o remount,rw ${boot_dev}
+    grub2-install --target ${target} --boot-directory ${boot_dev} --no-nvram ${device}    
+    ;;
+
+  *)
+    echo "skip grub2-install testing on ${arch}"
```
dustymabe: since we now have an `architectures:` line at the top of this file we should error here with `fatal` if we make it into this `*` case.

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,61 @@
...
+device=
+target=
+arch=$(arch)
+case ${arch} in
+  x86_64)
+    target=i386-pc
+    core="${boot_dev}/grub2/${target}/core.img"   
+    device=$(findmnt -no SOURCE ${boot_dev} | sed 's/[0-9]$//g')
+
+    [ -e ${core} ]
```
dustymabe: I feel like this probably does what you want (fail if the file doesn't exist) because we `set -e` at the top of this script but it's not very easy to read. Maybe something like this would be better:

```suggestion
    [ -e ${core} ] || fatal "file ${core} doesn't exist"
```

Same for `ppc64le` case below.

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,61 @@
...
+    grub2-install --target ${target} --boot-directory ${boot_dev} --no-nvram ${device}    
+    ;;
+
+  *)
+    echo "skip grub2-install testing on ${arch}"
+    exit 0
+    ;;
+esac
+
+generator_core_sum=$(sha256sum ${core} | awk '{print $1}')
```
dustymabe: where is the word `generator` coming from? Maybe `new_core_sum` would be a better variable name?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,61 @@
...
+  x86_64)
+    target=i386-pc
+    core="${boot_dev}/grub2/${target}/core.img"   
+    device=$(findmnt -no SOURCE ${boot_dev} | sed 's/[0-9]$//g')
+
+    [ -e ${core} ]
+    core_sum=$(sha256sum ${core} | awk '{print $1}')
+
+    mount -o remount,rw ${boot_dev}
+    grub2-install --target ${target} --boot-directory ${boot_dev} --modules mdraid1x ${device}
```
dustymabe: Where is the `--modules mdraid1x` coming from? i.e. is there corresponding code in COSA (or somewhere else) that we need to keep this in sync with?

HuijingHei: Yes, according to https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh#L481, also sync https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh#L494 for `ppc64le`

dustymabe: I think it may be worth a comment mentioning create_disk.sh (maybe not a link to exact line numbers because those will go stale, but a link to the file). before this `grub2-install` and the one below.

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,64 @@
...
+# See:
+#  - https://github.com/coreos/coreos-assembler/pull/3190
+#  - https://github.com/coreos/coreos-assembler/issues/3148
+
+## kola:
+##   # This test should pass everywhere if it passes anywhere.
+##   platforms: qemu
+##   # The test is not available for aarch64 and s390x,
+##   # as aarch64 is UEFI only and s390x is not using grub2
+##   architectures: "!aarch64 s390x"
```
jlebon: I think this needs to be

```suggestion
##   architectures: "!aarch64 !s390x"
```

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,64 @@
...
+boot_dev=/boot
+core=
+device=
+target=
+arch=$(arch)
+case ${arch} in
+  x86_64)
+    target=i386-pc
+    core="${boot_dev}/grub2/${target}/core.img"   
+    device=$(findmnt -no SOURCE ${boot_dev} | sed 's/[0-9]$//g')
```
jlebon: Oh I see, sorry for giving you an incomplete suggestion. I missed that we need the parent device here and not the partition. Cool with this, but e.g. if we ever loosen this test to also run on AWS, this would fail on NVMe instances I think. A more generic way to get at the parent device is:

```sh
partition=$(findmnt -no SOURCE ${boot_dev})
device=$(lsblk --paths -no PKNAME ${partition})
```

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,64 @@
...
+##   # as aarch64 is UEFI only and s390x is not using grub2
+##   architectures: "!aarch64 s390x"
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+boot_dev=/boot
+core=
+device=
+target=
```
jlebon: Minor: is there a reason to declare these upfront?

tests/kola/boot/grub2-install
```diff
@@ -0,0 +1,64 @@
...
+    core="${boot_dev}/grub2/${target}/core.elf"
+    device=$(findfs PARTLABEL=PowerPC-PReP-boot)
+    
+    [ -e ${core} ] || fatal "file ${core} doesn't exist"
+    core_sum=$(sha256sum ${core} | awk '{print $1}')
+
+    mount -o remount,rw ${boot_dev}
+    # sync grub2-install command according to image build script
+    # https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh
+    grub2-install --target ${target} --boot-directory ${boot_dev} --no-nvram ${device}
```
jlebon: Minor: It looks like this is almost exactly the same steps on both arches apart from the `--no-nvram` vs `--modules mdraid1x`. WDYT about defining a variable like `grub_install_args` instead and then factoring out this code to after the `case` structure?

jlebon (approved):
LGTM, thanks! Were you able to test this on ppc64le?

HuijingHei:
> LGTM, thanks! Were you able to test this on ppc64le?

Thanks! I just start ppc64le vm (based on https://github.com/coreos/coreos-assembler/pull/3201) and run the script, result is passed.

PR #2100, opened 2022-11-25
tests: fix skip no-persist-ip on s390x

HuijingHei:


travier:
Aparrently '!' has a meaning in YAML so we need quotes: https://stackoverflow.com/questions/9664113/what-does-a-single-exclamation-mark-do-in-yaml

travier (approved):

travier:
Can you check all other tests?

travier:
If you could convert all other tests to the new YAML format that would be great too

jlebon (approved):
Thanks!

jlebon:
This should fix existing known test failures, so let's get this in and do any other cleanups as follow-ups.

jlebon:
> If you could convert all other tests to the new YAML format that would be great too

```
$ git grep 'kola: {' tests/kola | wc -l
89
```

Hmm, that's more work than the comment implies. Let's track that in a GitHub issue and see if someone wants to tackle it?

travier:
> Hmm, that's more work than the comment implies. Let's track that in a GitHub issue and see if someone wants to tackle it?

Agree. Sorry, I was unclear. I was suggesting that a follow-up. I'll make an issue.

travier:
Made https://github.com/coreos/fedora-coreos-config/issues/2106

PR #2110, opened 2022-11-29
boot/grub2-install: fix `skip s390x` issue

HuijingHei:
Fix `fatal 'grub2-install testing is not supported on s390x'`


gursewak1997 (approved):
Definitely got confused initially but it does make sense why we don't want the `!` for both arches. The change works so LGTM.

HuijingHei:
Also confused, `!aarch64 !s390x` should work

HuijingHei:
Thanks @gursewak1997 for the review!
Anyway, we can merge this to let CI works smoothly, and do some tweak in the later patch.

PR #2113, opened 2022-12-01
[stable] image-base.yaml: set `compressor: xz`

jlebon:
This will be picked up by `cosa compress` to know to compress with xz.

(cherry picked from commit 7d3412cbfbd06dbaf923aaa781314de768a39dcb)


dustymabe (approved):
LGTM

PR #2114, opened 2022-12-01
denylist: extend snooze for ext.config.kdump.crash on ppc64le

dustymabe:
The 6.1 kernel hasn't made it out yet so we'll wait another month for this test.


jlebon (approved):

dustymabe:
Merging without CI since CI doesn't test on ppc64le.

PR #2115, opened 2022-12-01
Revert: "s390x: Remove SetGID dotlockfile from liblockfile"

dustymabe:
This reverts e79c4f1 and 5053bc7 (and a small part of 45434ed), which added an exclusion for dotlockfile. The liblockfile package, which provides dotlockfile, should no longer be getting used by s390utils so we can drop this.

See https://github.com/coreos/fedora-coreos-tracker/issues/1253#issuecomment-1313519008


jlebon (approved):
:tada:

PR #2116, opened 2022-12-07
kola/files/validate-symlinks: Update broken symlinks list

gursewak1997:
The `/usr/lib/firmware/brcm/brcmfmac43455-sdio.AW-CM256SM.txt.xz` is fixed and can be dropped from the broken symlinks list

### DIFF
--- a/tests/kola/files/validate-symlinks
+++ b/tests/kola/files/validate-symlinks
@@ -20,7 +20,6 @@ list_broken_symlinks_skip=(
     '/etc/swid/swidtags.d/fedoraproject.org'
     '/etc/xdg/systemd/user'
     '/usr/lib/.build-id/'
-    '/usr/lib/firmware/brcm/brcmfmac43455-sdio.AW-CM256SM.txt.xz'
     '/usr/lib/firmware/qcom/LENOVO/21BX.xz'
     '/usr/lib/modules/'
     '/usr/share/rhel/secrets/etc-pki-entitlement'


dustymabe (approved):
LGTM

jlebon:
Did you just notice this by inspection BTW? Do we have some mechanism to keep track of when broken symlinks are fixed?

gursewak1997:
> Did you just notice this by inspection BTW? Do we have some mechanism to keep track of when broken symlinks are fixed?

Yes, I found it when I was [verifying the symlinks](https://github.com/coreos/fedora-coreos-tracker/issues/1310#issuecomment-1334950234) to other firmware files. I realized the test for this file didn't fail. I am not too sure if there's a mechanism where we keep track of this though.

jlebon:
Ack gotcha. It'd be nice to have something. E.g. we could have the test error out if a symlink we expected was broken no longer was. It's harsh but would make sure that we eventually trim down the list. (Alternatively, the idea of having tests "warn" has come up before.) Also, ideally each item in that list would have an associated e.g. RHBZ tracking it.

jlebon:
Hmm, I'm confused by this test. There are a bunch of symlinks on that list that shouldn't belong:
- `/etc/mtab` is a valid symlink to `/proc/self/mounts`
- `/etc/xdg/systemd/user` is a valid symlink to `/etc/systemd/user`
- `/usr/lib/modules`, `/etc/ssl`, and `/usr/lib/.build-id` are not symlinks but valid directories

I can imagine the first one having ended up on that list because it looks broken on a non-booted rootfs. I'm not sure how the other ones got there. I think it'd be good to have someone refresh the list and dig a bit into the remaining ones.

gursewak1997:
@jlebon I can dig more into it and will update the list accordingly.

gursewak1997:
Update to symlinks testing as I tested locally:

- For `etc/mtab`- Error: `/etc/mtab` symlink to `/proc/5013/mounts` which does not exist
- For `/usr/lib/modules`- Error: `/usr/lib/modules/6.1.0-0.rc8.58.fc38.x86_64/build` symlink to `/usr/src/kernels/6.1.0-0.rc8.58.fc38.x86_64` which does not exist
- For `/usr/lib/.build`- Error: `/usr/lib/.build-id/00/7ab8afbe1a66812fa69a679a3e97b84afe1af8` symlink to `/usr/sbin/xfs_scrub` which does not exist

For `/etc/ssl` and `/etc/xdg/systemd/user`, I don't see any direct failure so far. Will wait for another ack (probably  @travier?) before I update the list itself in case it fails on some other platforms/cases.

Also, the `list_broken_symlinks_skip`  list does have valid directories (not symlinks) because when we run the test, we also check the file's existence BELOW that directory. Ref: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/files/validate-symlinks#L47

travier:
The test is still very basic and the list is mostly here as an ignore/skip list (see the `*` in https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/files/validate-symlinks#L47). The initial discussion was in https://github.com/coreos/fedora-coreos-config/issues/1782.

> /etc/mtab is a valid symlink to /proc/self/mounts

The test currently uses `realpath -m <file>` which fully resolves the destination thus you will always end up with an invalid destination as the process that resolved it (`realpath`) is gone. So `etc/mtab` is fine and it was initially easier to skip it than to fix the logic.

> /usr/lib/modules, /etc/ssl, and /usr/lib/.build-id are not symlinks but valid directories

Those are directories that may contain "broken symlinks" and are thus skipped altogether to simplify things too. It looks like `/etc/ssl` and `/etc/xdg/systemd/user` may be fine too and it may just be missing logic in the test or a previously broken version of the test and we did not update the list.

`/usr/lib/.build-id` is here to skip all the broken build-id links that come from packages and that were not broken by us. We should investigate those but it's very low priority.

Similarly, for `/usr/lib/modules` it was easier to just skip the whole directory.

---

It also looks like https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/files/validate-symlinks#L31 is incorrect: the `[[ ]]` are not needed here according to ShellCheck.

travier:
Filed: https://github.com/coreos/fedora-coreos-config/issues/2117

PR #2118, opened 2022-12-09
tests/kola: add test for systemd-timesyncd

dustymabe:


tests/kola/ntp/chrony/dhcp-propagation
```diff
@@ -78,13 +79,12 @@ main() {
     # Name: time-c-g.nist.gov  IP address: 129.6.15.30
     NTPSERVER=""
```
jlebon: Minor: I think we can drop this too now?

tests/kola/ntp/chrony/dhcp-propagation
```diff
@@ -78,13 +79,12 @@ main() {
...
-    while [[ $retries -gt 0 && ! $NTPSERVER =~ "time-c-g.nist.gov" ]]; do
-         NTPSERVER=$(chronyc sources)
-         if [[ $NTPSERVER != *"time-c-g.nist.gov"* ]]; then
-             echo "waiting for ntp server to appear"
-             sleep 1
-             retries=$((retries - 1))
-         fi
+    while [[ $retries -gt 0 ]]; do
+         NTPSERVER=$(chronyc -n sources)
+         [[ "$NTPSERVER" =~ "$NTPHOSTIP" ]] && break
```
jlebon: Feels like the commit message here is missing the rationale for why we've moving away from being name resolution-based. Did you observe a failure?

tests/kola/binfmt/qemu
```diff
@@ -22,7 +22,7 @@ set -xeuo pipefail
 
 case "$(arch)" in
     aarch64|s390x)
-        containerArch=$(podman run --arch=amd64 --rm registry.fedoraproject.org/fedora:36 arch)
+        containerArch=$(podman run --arch=amd64 --rm registry.fedoraproject.org/fedora:37 arch)
```
jlebon, line 25: Should we add a step for this in the rebase checklist?

tests/kola/ntp/chrony/dhcp-propagation
```diff
@@ -1,96 +1,40 @@
...
-#     that.
-# - timeoutMin: 15
-#   - Pulling and building the container can take a long time if a
-#     slow mirror gets chosen.
-# - minMemory: 1536
-#   - There's a bug in dnf that is causing OOM on low memory systems:
-#     https://bugzilla.redhat.com/show_bug.cgi?id=1907030
-#     https://pagure.io/releng/issue/10935#comment-808601
+## kola: 
+##   # Add the needs-internet tag. This test builds a container from remot
```
jlebon: remote

tests/kola/ntp/timesyncd/dhcp-propagation/config.bu
```diff
@@ -0,0 +1,17 @@
...
+    - name: systemd-timesyncd.service
+      enabled: true
+# There currently isn't any NetworkManager dispatcher that
+# propagates DHCP info to timesyncd in Fedora so we'll just
+# pull one that we can use from the internet.
+storage:
+  files:
+    - path: /etc/NetworkManager/dispatcher.d/30-timesyncd
+      contents:
+        source: "https://github.com/eworm-de/networkmanager-dispatcher-timesyncd/raw/master/30-timesyncd"
```
jlebon, line 16: I'd prefer just having it be inlined here to keep it self-contained. It's not very large. (With linkback of course.)

tests/kola/ntp/chrony/dhcp-propagation
dustymabe: yep

tests/kola/ntp/chrony/dhcp-propagation
dustymabe: basically it makes it easier for the code to support timesyncd in the following commits. I added a note to the commit message.

tests/kola/binfmt/qemu
dustymabe: probably

tests/kola/ntp/timesyncd/dhcp-propagation/config.bu
dustymabe: ehh. I'm hesitant to copy other people's code around. What's the concern here? That the repo disappears?

If that happens then we should probably stop testing this.

jlebon: That the file moves, disappears, or is modified and that we're unnecessarily adding a dependency in our CI we could easily avoid.

dustymabe: Unfortunately the code at https://github.com/eworm-de/networkmanager-dispatcher-timesyncd doesn't have a license. I guess we could just not add the test - or rework it in some way.

dustymabe: I'm thinking if it disappears or starts breaking we can just delete it.

jlebon (approved):
Ack, sounds good!

travier:
This is failing on RHCOS: https://github.com/openshift/os/pull/1088
Maybe this should be FCOS only?

travier:
More precisely, `ext.config.shared.ntp.timesyncd.dhcp-propagation` is failing.

dustymabe:
> This is failing on RHCOS: [openshift/os#1088](https://github.com/openshift/os/pull/1088) Maybe this should be FCOS only?

`ntp.timesyncd.dhcp-propagation` absolutely should be `fcos` only. An earlier version of this PR had it that way but then I lost it in a copy/paste.

PR to fix: 
https://github.com/coreos/fedora-coreos-config/pull/2127

PR #2119, opened 2022-12-12
ci/buildroot: Add +ShellCheck and file

travier:
Will be used to check Bash scripts in CI to catch common errors.

See: https://github.com/coreos/repo-templates/pull/39
See: https://github.com/openshift/os/pull/1001


travier:
Adding it directly to the buildroot will also make it easier to use for RHCOS.

travier:
https://github.com/openshift/os/pull/1001

jlebon (approved):

PR #2125, opened 2022-12-13
tests/kola: remove date conditional from files.console-config test

dustymabe:
First, the dates mentioned in [1] have passed so we no longer need this check. Second, checking the date inspecting the value from $VERSION was actually not quite correct because `stable` lags behind so the actual date in its $VERSION string is the date from the timestamp of the yumrepos (which was two weeks in the past). This means the `stable` threshold=20221209 isn't checking correctly (the current `stable` build we are trying to build today has a version string of `37.20221127.3.0`).

Just dropping the check solves both of these problems. Once merged we'll forward port this to the `stable` branch too.

[1] https://lists.fedoraproject.org/archives/list/coreos-status@lists.fedoraproject.org/message/GHLXX4MXNHUEAXQLK6BZN45IQYHRVQB4/

### DIFF
--- a/tests/kola/files/console-config
+++ b/tests/kola/files/console-config
@@ -37,33 +37,10 @@ check_platforms_json() {
 }
 
 # Check whether platforms.json exists
-if is_fcos; then
-    # Schedule based on:
-    # https://lists.fedoraproject.org/archives/list/coreos-status@lists.fedoraproject.org/message/GHLXX4MXNHUEAXQLK6BZN45IQYHRVQB4/
-    case "$(get_fcos_stream)" in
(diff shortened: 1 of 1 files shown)


jlebon (approved):

dustymabe:
wow ci failed because it couldn't pull from GitHub:

```
Command "git fetch --no-tags --force --progress -- https://github.com/coreos/fedora-coreos-config.git +refs/pull/2125/head:refs/remotes/origin/PR-2125 +refs/heads/testing-devel:refs/remotes/origin/testing-devel" returned status code 128:

```

PR #2126, opened 2022-12-13
[stable] tests/kola: remove date conditional from files.console-config test

dustymabe:
First, the dates mentioned in [1] have passed so we no longer need this check. Second, checking the date inspecting the value from $VERSION was actually not quite correct because `stable` lags behind so the actual date in its $VERSION string is the date from the timestamp of the yumrepos (which was two weeks in the past). This means the `stable` threshold=20221209 isn't checking correctly (the current `stable` build we are trying to build today has a version string of `37.20221127.3.0`).

Just dropping the check solves both of these problems. Once merged we'll forward port this to the `stable` branch too.

[1] https://lists.fedoraproject.org/archives/list/coreos-status@lists.fedoraproject.org/message/GHLXX4MXNHUEAXQLK6BZN45IQYHRVQB4/

(cherry picked from commit 720c9f8562e747162b1bef68e342930e1874c8d2)

### DIFF
--- a/tests/kola/files/console-config
+++ b/tests/kola/files/console-config
@@ -37,33 +37,10 @@ check_platforms_json() {
 }
 
 # Check whether platforms.json exists
-if is_fcos; then
-    # Schedule based on:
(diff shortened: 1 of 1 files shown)


jlebon (approved):

dustymabe:
Not sure why CI isn't running but the pipeline run will fail if this doesn't work so I'm going to merge.

PR #2127, opened 2022-12-13
tests/kola: limit timesyncd test to FCOS

dustymabe:
We only care about timesyncd in Fedora. It's not available elsewhere.


cgwalters (approved):

cgwalters:
Just going to force this, CI failure is unrelated

travier:
Thanks!

PR #2130, opened 2022-12-14
tests: Use qemu-unpriv as much as possible

travier:
Tests that are run locally in an unprivileged container will run under the `qemu-unpriv` platform by default. We should try to run as many tests as possible under this platform and only keep the test requiring privileged `qemu` using this platform.

### DIFF
--- a/tests/kola/binfmt/qemu
+++ b/tests/kola/binfmt/qemu
@@ -1,9 +1,9 @@
 #!/bin/bash
-# kola: {"platforms": "qemu", "exclusive": false, "tags": "needs-internet", "distros": "fcos", "architectures": "aarch64 s390x"}
+# kola: {"platforms": "qemu-unpriv", "exclusive": false, "tags": "needs-internet", "distros": "fcos", "architectures": "aarch64 s390x"}
 # Test the x86_64 emulator on aarch64 and s390x images for now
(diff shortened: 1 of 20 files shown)


dustymabe:
My understanding was that privileged qemu no longer existed in mantle and `qemu` was just an alias for `qemu-unpriv`. If that's the case what we should do here is remove all references to `qemu-unpriv` and update any callers to use `qemu`.

travier:
Hum, OK, so we should just switch everything to qemu and basically alias qemu-unpriv to qemu in kola?

dustymabe:
> Hum, OK, so we should just switch everything to qemu 

Yes

> and basically alias qemu-unpriv to qemu in kola?

or do what we probably should have done a while back and just have there be only one - no aliases.

cgwalters:
Yeah, changing everything to use `qemu` makes sense to me now.

cgwalters:
I think I'd say let's get https://github.com/coreos/fedora-coreos-config/pull/2131 in first, then this one can be a straightforward sed job on top of it.

travier:
Arg, I had not seen this one yet. Good point

travier:
Will be replaced by https://github.com/coreos/fedora-coreos-config/pull/2133

PR #2132, opened 2022-12-14
denylist: aws.nvme: extend snooze and re-apply to all streams

c4rt0:
It seems as if the fix that AWS had applied is no longer working.
See: https://github.com/coreos/fedora-coreos-tracker/issues/1306#issuecomment-1337660156


dustymabe (approved):
LGTM

PR #2134, opened 2022-12-15
[rawhide] overrides: fast-track selinux-policy-38.1-1.fc38

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -34,3 +34,15 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1341
       type: pin
+  selinux-policy:
+    evra: 38.1-1.fc38.noarch
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-cdeb268b2f
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1364
+      type: fast-track
+  selinux-policy-targeted:
+    evra: 38.1-1.fc38.noarch
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-cdeb268b2f
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1364
+      type: fast-track


prestist (approved):

prestist:
LGTM; This is to get us passed a failing tests in rawhide; see more [here](https://github.com/coreos/fedora-coreos-tracker/issues/1364)

manifest-lock.overrides.yaml
```diff
@@ -34,3 +34,15 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1341
       type: pin
+  selinux-policy:
+    evra: 38.1-1.fc38.noarch
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2022-cdeb268b2f
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1364
+      type: fast-track
```
dustymabe, line 42: Let's switch this to a pin. I think you can re-run the github action and select `pin` and provide the NVR of the package (i.e. `selinx-policy-38.1-1.fc38`) rather than the bodhi URL for this.

dustymabe: The significance of using a `pin` versus a fast-track is that we have a process that "graduates" fast-tracks by removing them from this file once an equal or newer RPM is in the repos.

We don't want that "graduate overrides" process to remove this pin, since there is no known fix yet.

dustymabe (changes requested):
switch to a pin

manifest-lock.overrides.yaml
prestist: Ah, ok, well that makes sense. 

I tried to rebuild it but I didn't get an option to change workflow inputs so I opted for a new workflow instance? 

I did ended up getting build not found, which is confusing https://github.com/coreos/fedora-coreos-config/actions/runs/3708111431/jobs/6285312616

I verified the build was here https://koji.fedoraproject.org/koji/buildinfo?buildID=2091670

dustymabe: weird. I ran it and got the same thing. Not sure what's going on but running the command worked locally so maybe it's something about the environment inside the GH action.

This isn't something you need to investigate. I just opened https://github.com/coreos/fedora-coreos-config/pull/2135 with the results of my locally run command.

PR #2135, opened 2022-12-15
[rawhide] overrides: pin selinux-policy-38.1-1.fc38.noarch

dustymabe:
The SELinux is causing a test failure in the ostree.hotfix test. See https://github.com/coreos/fedora-coreos-tracker/issues/1364


prestist (approved):
LGTM

PR #2141, opened 2023-01-02
Make handling of `boot=` and `root=` kargs idempotent

cgwalters:
dracut/boot-edit: Do nothing if we have a boot= karg

In this case, we shouldn't have anything to do.  This is needed
for `bootc install` which currently sets up these kargs in advance.
https://github.com/containers/bootc/pull/30

---

dracut/ignition-setup: Use boot= karg if it exists

This will be used with bootc which will default to injecting
a karg for `boot`.

---

ignition-uuid-root: no-op if we already have a root= karg

For bootc compatibility.

---

overlay: More no-ops if we have a boot= and/or root= karg

We can assume the disk image is already set up in this case.

---


jlebon:
Hmm, so this is to support a `bootc`-installed FCOS, right? I think it'd be good to discuss what the long-term goal is for `bootc` and FCOS.

Re. this PR specifically, the use of `/dev/disk/by-label/boot` is pervasive throughout this codebase. But anyway, we also now have code that ensures that only one filesystem labeled `boot` exists so it shouldn't be necessary, right? Or are you trying to support systems where that is no longer true?

cgwalters:
> Hmm, so this is to support a bootc-installed FCOS, right? I think it'd be good to discuss what the long-term goal is for bootc and FCOS.

Yep, I plan to get something scheduled for this soon this month!

> But anyway, we also now have code that ensures that only one filesystem labeled boot exists so it shouldn't be necessary, right?

This scenario arises when one boots one FCOS-derived system, and is using it to install to a different disk.

But this isn't critical to merge soon, so I just marked as draft.

cgwalters:
OK, lifting draft on this one.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-setup-user.sh
```diff
@@ -14,6 +14,60 @@ copy_file_if_exists() {
...
+            value="${arg#*=}"
+        fi
+    done
+    echo "${value}"
+}
+
+# Copied from
+# https://github.com/dracutdevs/dracut/blob/9491e599282d0d6bb12063eddbd192c0d2ce8acf/modules.d/99base/dracut-lib.sh#L586
+# rather than sourcing it.
+label_uuid_to_dev() {
```
cgwalters, line 31: Maybe this should move to rdcore or so?

cgwalters:
```
[2023-01-17T21:20:18.693Z] + cosa kola testiso -S --qemu-multipath --scenarios iso-offline-install --output-dir /srv/coreos/tmp/kolaTestIso-7qsbl/kola-testiso-multipath
[2023-01-17T21:20:18.693Z] kola -p qemu-unpriv --output-dir /srv/coreos/tmp/kolaTestIso-7qsbl/kola-testiso-multipath testiso -S --qemu-multipath --scenarios iso-offline-install
[2023-01-17T21:20:18.693Z] Testing scenarios: [iso-offline-install]
[2023-01-17T21:20:18.693Z] Ignoring verification of signature on metal image
[2023-01-17T21:21:26.668Z] FAIL: iso-offline-install ( + metal + multipath) (1m1.819s)
[2023-01-17T21:21:26.668Z]     entered emergency.target in initramfs
[2023-01-17T21:21:26.668Z] Error: entered emergency.target in initramfs
```

Hm, not seeing the failure there in the logs.

dustymabe:
> Hm, not seeing the failure there in the logs.

https://github.com/coreos/fedora-coreos-tracker/issues/1373

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-uuid-root.service
```diff
@@ -9,6 +9,7 @@ After=ignition-disks.service
 # If we've reprovisioned the rootfs, then there's no need to restamp
 ConditionPathExists=!/run/ignition-ostree-transposefs
 ConditionKernelCommandLine=!rootfs.roothash
+ConditionKernelCommandLine=!root
```
jlebon, line 12: I think this would break multipath, where currently you have to put the `root` karg at install time but we do still want the restamping. Is this actually required for bootc? The service will run but the script should no-op if the filesystem UUID was already randomized.

Similarly for `coreos-gpt-setup.service`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-setup-user.sh
```diff
@@ -14,6 +14,60 @@ copy_file_if_exists() {
...
+            echo "/dev/disk/by-uuid/$(echo "${_dev#UUID=}" | tr "[:upper:]" "[:lower:]")"
+            ;;
+        PARTUUID=*)
+            echo "/dev/disk/by-partuuid/$(echo "${_dev#PARTUUID=}" | tr "[:upper:]" "[:lower:]")"
+            ;;
+    esac
+}
+
+# This is copied from coreos-boot-mount-generator which we should likely run
+# in the initramfs too
```
jlebon, line 51: Yeah, I think that'd be a safer path. This seems to be trying to undo the assumption that `/dev/disk/by-label/boot` points to our boot disk at first boot, but only for one service, whereas it's an assumption embedded all over the place. The generator could put the found device in e.g. `/run/coreos/bootdev`?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-setup-user.sh
```diff
@@ -14,6 +14,60 @@ copy_file_if_exists() {
...
+elif [ -n "${bootkarg}" ]; then
+    # Adapted from https://github.com/dracutdevs/dracut/blob/9491e599282d0d6bb12063eddbd192c0d2ce8acf/modules.d/01fips/fips.sh#L17
+    case "$bootkarg" in
+        LABEL=* | UUID=* | PARTUUID=* | PARTLABEL=*)
+            bootdev="$(label_uuid_to_dev "$bootkarg")";;
+        /dev/*) bootdev=$bootkarg;;
+        *) echo "Unknown boot karg '${bootkarg}'; falling back to ${bootdev}";;
+    esac
+# This is used for the first boot only
+elif [ -f /run/coreos/bootfs_uuid ]; then
```
jlebon, line 67: This is written by `coreos-boot-edit` and isn't relevant here.

jlebon (commented):
Seems like the last two commits could be merged?

cgwalters:
For now, won't be targeting FCOS with `bootc install`.

PR #2151, opened 2023-01-04
[testing] denylist: drop snoozes before promotion to stable

dustymabe:
We don't want these tests to run when the next `stable` build happens so let's drop the snoozes here in the `testing` branch.


jlebon (approved):
We should probably figure out some way to not have to do this. One idea is to have kola use the timestamp in the build ID instead of today's date when comparing snooze dates.

dustymabe:
> We should probably figure out some way to not have to do this. One idea is to have kola use the timestamp in the build ID instead of today's date when comparing snooze dates.

Maybe we can address this like so: https://github.com/coreos/fedora-coreos-streams/pull/629

PR #2152, opened 2023-01-04
overrides: drop grub2 pin, bump to 2.06-72.fc37

jlebon:
Issue #1352 is now fixed in the latest grub2 in f37 as per https://github.com/coreos/fedora-coreos-tracker/issues/1352#issuecomment-1371225262.


dustymabe:
That version of grub2 is already in the Fedora `stable` bodhi repos so let's just delete the files to drop the override instead?

jlebon:
Yeah, I debated doing that. If we delete it, an immediate build would still respect the base lockfile containing the old grub2 and we wouldn't have the bump until bump-lockfile runs. Doing it this way means CI actually tests the new grub2 package.

dustymabe:
does the override graduating code handle the arch specific overrides files? If it doesn't then maybe just hand edit the generated lockfiles in this PR too and delete the arch specific overrides files. I just don't want to make more work for ourselves than we need to.

jlebon:
It doesn't currently. Updated to bump base lockfiles directly!

dustymabe:
From the commit message:

> Issue #1352 is now fixed in the latest grub2 in f37 as per

In the GitHub web UI this points us at https://github.com/coreos/fedora-coreos-config/pull/1352 which is not what you wanted. Can you either give it a full identifier or remove the reference (the later link suffices).

dustymabe (dismissed):
one optional suggestion.

LGTM

dustymabe (approved):
LGTM

PR #2153, opened 2023-01-04
[rawhide] overrides: drop grub2 pin

jlebon:
Issue #1352 is now fixed in the latest grub2 in f38 as per https://github.com/coreos/fedora-coreos-tracker/issues/1352#issuecomment-1371225262.


dustymabe:
Let's just delete the files to drop the override instead?

dustymabe:
From the commit message:

> Issue #1352 is now fixed in the latest grub2 in f37 as per

In the GitHub web UI this points us at https://github.com/coreos/fedora-coreos-config/pull/1352 which is not what you wanted. Can you either give it a full identifier or remove the reference (the later link suffices).

dustymabe (dismissed):
one optional suggestion.

LGTM

dustymabe:
There are two CI tests failing here that are unrelated to this change. See:

- https://github.com/coreos/fedora-coreos-tracker/issues/1364#issuecomment-1371727078
- https://github.com/coreos/fedora-coreos-tracker/issues/1369


Merging this..

PR #2156, opened 2023-01-05
manifests: install the x86_64 emulator on ppc64le instances

dustymabe:
This gets ppc64le in line with the other non-x86_64 arches.

### DIFF
--- a/tests/kola/binfmt/qemu
+++ b/tests/kola/binfmt/qemu
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Test the x86_64 emulator on aarch64 and s390x images for now
+# Test the x86_64 emulator on non-x86_64 instances.
 # https://github.com/coreos/fedora-coreos-tracker/issues/1237
 #
 ## kola:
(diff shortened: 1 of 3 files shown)


tests/kola/binfmt/qemu
```diff
@@ -12,15 +12,15 @@
 ##   tags: needs-internet
 ##   # Only run on fcos, as rhel does not support emulation
 ##   distros: fcos
-##   # We decided to ship x86_64 emulator to only these arches for now.
-##   architectures: aarch64 s390x
+##   # We ship the x86_64 on all non-x86_64 arches.
+##   architectures: "!x86_64"
```
jlebon: Minor/optional: I think we discussed this recently-ish somewhere, but a suggestion I had for this was to start writing it as `"! x86_64"` instead so it's less misleading. Other tests are still using that syntax so feel free to defer if we want to do that in one bulk change instead.

jlebon (dismissed):

jlebon:
I think also this is missing adding the package to the base ppc64le lockfile too.

tests/kola/binfmt/qemu
dustymabe: yeah, let's defer so we don't have to cycle CI again.

Maybe open an issue for the bulk change and we can get @c4rt0 to work on it?

dustymabe:
> I think also this is missing adding the package to the base ppc64le lockfile too.

that's not really necessary is it? I did a local build on a ppc64le machine and it worked fine without it. `bump-lockfile` will set things right when it runs next.

jlebon:
> > I think also this is missing adding the package to the base ppc64le lockfile too.
> 
> that's not really necessary is it? I did a local build on a ppc64le machine and it worked fine without it. `bump-lockfile` will set things right when it runs next.

Yup, the next `bump-lockfile` will fix it but until then any strict build (like the pipeline) will fail. It's more a best practice thing so that we're never in a "not buildable" state.

tests/kola/binfmt/qemu
jlebon: Sure, split into https://github.com/coreos/fedora-coreos-config/issues/2157.

dustymabe:
updated!

tests/kola/binfmt/qemu
dustymabe: ok I fixed this now that I needed to do another upload.

jlebon (approved):
LGTM!

PR #2158, opened 2023-01-06
manifests: explicitly name criu criu-libs

### LINKED ISSUE #1370: New Package Request: criu-libs
### What, if any, are the additional dependencies on the package? (i.e. does it pull in Python, Perl, etc)

 criu-libs is needed for podman to be able to subspend and resume a container.

### What is the size of the package and its dependencies?

dnf info criu shows
Size         : 1.5 M

### What problem are you trying to solve with this package? Or what functionality does the package provide?

Podman support for CRIU.
We moved this from a requires to a recommends, so it is no longer installed in coreos.

### Can the software provided by the package be run from a container?  Explain why or why not.

No, since it is needed for containers.

### Can the tool(s) provided by the package be helpful in debugging container runtime issues?

N/A

### Can the tool(s) provided by the package be helpful in debugging networking issues?

N/A

### Is it possible to layer the package onto the base OS as a day 2 operation?  Explain why or why not.

Yes, but not really worth it.

### In the case of packages providing services and binaries, can the packaging be adjusted to just deliver binaries?

Just binaries.

### Can the tool(s) provided by the package be used to do things we’d rather users not be able to do in FCOS?

No

### Does the software provided by the package have a history of CVEs?

Not that I know of.

dustymabe:
The `crun` upstream code switched to dynamically linking against libcriu dynamically [1] and the `crun` RPM switched the `criu` dep from a Requires to a Recommends. Let's just name them both here because we know we want them installed.

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1370

[1] https://github.com/containers/crun/commit/8c3fc125ba40d54675726fff55b4c03693a0d3dc
[2] https://src.fedoraproject.org/rpms/crun/c/c14e1218f5937622122158e11f5e777c1ee7065f


jlebon (dismissed):
Should this be in a shared manifest for RHCOS?

dustymabe:
> Should this be in a shared manifest for RHCOS?

I don't know, should it?

dustymabe:
OK added to a shared manifest (also where `crun` happens to be named).

jlebon (approved):

PR #2159, opened 2023-01-06
tests/kola/rpm-ostree: add test for kernel replacement

jmarrero:
closes https://github.com/coreos/rpm-ostree/issues/4205

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -1,0 +1,122 @@
+#!/bin/bash
+## kola:
+##   # This test only runs on FCOS due to a problem with skopeo copy on
+##   # RHCOS. See: https://github.com/containers/skopeo/issues/1846
+##   distros: fcos
+##   # We fetch packages from centos mirrors.
+##   tags: needs-internet
+#
+# Build container image with a new kernel and reboot into the new image.
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
+
+set -euo pipefail
+. $KOLA_EXT_DATA/commonlib.sh
+set -x
+cd $(mktemp -d)
+
+# TODO: It'd be much better to test this via a registry
+image_dir=/var/tmp/fcos
+image=oci:$image_dir
+image_pull=ostree-unverified-image:$image
+tmp_imagedir=/var/tmp/fcos-tmp
+arch=$(arch)
+fversion="37"
+kver="6.0.7-301.fc$fversion.${arch}"
+
+case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+  "")
+    # Take the existing ostree commit, and export it to a container image, then rebase to it.
+    checksum=$(rpm-ostree status --json | jq -r '.deployments[0].checksum')
+    v0=$(rpm-ostree status --json | jq -r '.deployments[0].version')
+    rm ${image_dir} -rf
+    # Since we're switching OS update stream, turn off zincati
+    systemctl mask --now zincati
+    ostree container encapsulate --repo=/ostree/repo ${checksum} "${image}"
+    # This one keeps --experimental, but we also test without it below
+    rpm-ostree rebase --experimental "$image_pull"
+    ostree container image list --repo=/ostree/repo | tee imglist.txt
+    # Test rebasing back to ostree https://github.com/coreos/rpm-ostree/issues/3677
+    rpm-ostree rebase "$checksum"
+    rpm-ostree rebase "$image_pull"
+    /tmp/autopkgtest-reboot 1
+    ;;
+  1)
+    # Setup
+    # copy the OCI dir to containers-storage for a local build
+    skopeo copy $image containers-storage:localhost/fcos
+    rm "${image_dir}" -rf
+    td=$(mktemp -d)
+    cd ${td}
+    version=$(rpm-ostree --version | grep Version)
+cat > dracut_call.sh << EOF
+#!/bin/bash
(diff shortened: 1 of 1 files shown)


jmarrero:
This test passes on Fedora CoreOS now. I'll submit a follow up PR once: https://github.com/containers/skopeo/issues/1846 is fixed/resolved.

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -0,0 +1,124 @@
...
+  1)
+    # Setup
+    # copy the OCI dir to containers-storage for a local build
+    skopeo copy $image containers-storage:localhost/fcos
+    rm "${image_dir}" -rf
+    td=$(mktemp -d)
+    cd ${td}
+    arch=$(arch)
+    version=$(rpm-ostree --version | grep Version)
+    kver="5.14.0-226.el9.${arch}"
```
cgwalters: One thing we quickly learned after https://github.com/openshift/os/pull/1087 is that the kernel packages will get GC'd.  So...we will need to do one of:

- query dynamically
- if we're on rhel, replace with the latest centos kernel always  (but this doesn't help the test on centos)
- copy a fixed kernel package to some sort of lookaside place (tempting actually to stick it in a container image)

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -0,0 +1,124 @@
...
+    derived=oci:$image_dir:derived
+    skopeo copy containers-storage:localhost/fcos-derived $derived
+    rpm-ostree --version
+    rpm-ostree rebase ostree-unverified-image:$derived
+    ostree container image list --repo=/ostree/repo
+    rm $image_dir -rf
+    /tmp/autopkgtest-reboot 2
+    ;;
+  2)
+    kver="5.14.0-226.el9.${arch}"
```
cgwalters: Let's move this assignment to the top of the file so it's not duplicated.

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -0,0 +1,124 @@
...
+    if uname -a | grep $kver; then
```
cgwalters: Should be able to do something more like:
```
un=$(uname -r)
if test "$un" != "$kver"; then
   echo "Expected kernel $kver but found $un"
   exit 1
fi
```
or so?

tests/kola/rpm-ostree/kernel-replace
jmarrero: I can stick them on a container for sure. The only question is do we have anywhere to push containers like that? I can use my repo on quay.io or ghcr but we probably don't want that.

cgwalters: We could create something like github.com/coreos/test-fixtures-container that has this and potentially other stuff for other tests?  It'd be pushed to `quay.io/coreos-assembler/test-fixtures` or so. Curious if anyone else on the team has opinons!

dustymabe: What I don't fully understand here is why we're doing a kernel-replace of a centos kernel on FCOS. Technically it should work, but is that something we really want to test here? I wouldn't recommend any of our users do that.

My preference here would be that we pin on a Fedora kernel that is in our Fedora CoreOS Pool koji tag/distrepo. Those packages don't get GC until we manually do it once a Fedora cycle as part of the [rebase checklist](https://github.com/coreos/fedora-coreos-tracker/blob/main/.github/ISSUE_TEMPLATE/rebase.md). So we can just pin on one of those (for example [`kernel-6.0.18-300.fc37.x86_64.rpm`](https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/latest/x86_64/Packages/k/kernel-6.0.18-300.fc37.x86_64.rpm)) and then we update it once a cycle when we GC.

cgwalters: Right, for Fedora we could also hardcode the "GA" kernel e.g. f37.  That's what we actually do in the rpm-ostree tests related to this.

But, we don't quite have that "GA" for centos stream AFAIK?

dustymabe: Right now this test has `distros: fcos`, but I assume we want to run it on the others.

We should probably have a Fedora kernel that we use to replace on Fedora and a CentOS/RHEL kernel that we use on RHCOS/SCOS. For the Fedora kernel we can adopt the policy I mentioned above. For CentOS/RHEL we could just add a step for the user when they update the Fedora kernel to pick a new CentOS/RHEL kernel as well (from the location we're pulling them from).

I really don't think us managing our own copy of kernel files for this test (i.e. in a test fixtures container) as useful. There probably needs to be regular updating of the test to use a newer kernel for the test and the GC would force that to happen.

cgwalters: > We should probably have a Fedora kernel that we use to replace on Fedora and a CentOS/RHEL kernel that we use on RHCOS/SCOS. For the Fedora kernel we can adopt the policy I mentioned above.

Yes, agree that's fine.

> For CentOS/RHEL we could just add a step for the user when they update the Fedora kernel to pick a new CentOS/RHEL kernel as well (from the location we're pulling them from).

That's the sticking point though, we don't have the same pool repo c9s side.  Arguably of course we should.  Right now looking at https://composes.stream.centos.org/development/ there's at least ~two months of history, so if we pick a new one now fixing the test every 2 months may not be so bad.

Or, we change this test to be dynamic and parse the composes.  May not be too hard, just ugly.

I think I'd say for now let's:

- Change this test to use the fedora kernel
- merge
- look at centos/rhel as a followup

?

jmarrero: I just downloaded packages from: https://koji.fedoraproject.org/koji/buildinfo?buildID=1849252 we can probably use any of the koji builds and don't worry about updating this test for fedora for a year? maybe more. Would be nice to have a stable package to test against and not having to update a test every few months. @dustymabe why a using the coreos pool better?

jlebon: Agree re. matching what rpm-ostree does: https://github.com/coreos/rpm-ostree/blob/c20e9834b35722679cf0fc02318966cfd7d71186/tests/vmcheck/test-override-kernel.sh#L36-L47.

> we can probably use any of the koji builds and don't worry about updating this test for fedora for a year? maybe more

I think that will work (at least until userspace adds requirements on newer kernels), but bigger picture I think for this test to stay realistic, we should be bumping it so it's tested against newer RPMs. Bumping it around rebase times seems reasonable to me.

dustymabe: Yeah using the GA kernel seems like a good option.

dustymabe: > I think I'd say for now let's:
> 
> * Change this test to use the fedora kernel
> * merge
> * look at centos/rhel as a followup

Works for me.

jmarrero: PR updated.

cgwalters (approved):

cgwalters:
```
[2023-01-14T17:24:46.234Z] --- PASS: ext.config.rpm-ostree.kernel-replace (434.87s)
```

PR #2165, opened 2023-01-10
Consistently Use Title Case for Service Descriptions

bgilbert:
Also do some minor wording cleanups.

Fixes https://issues.redhat.com/browse/COS-1840.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-unique-boot.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-unique-boot.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=Ensure filesystem labeled `boot` is unique
+Description=Ensure Unique `boot` Filesystem Label
 ConditionPathExists=/etc/initrd-release
 DefaultDependencies=no
 Before=ignition-diskful.target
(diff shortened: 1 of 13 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-unique-boot.service
```diff
@@ -1,5 +1,5 @@
 [Unit]
-Description=Ensure filesystem labeled `boot` is unique
+Description=Ensure Filesystem Labeled `boot` Is Unique
```
jlebon: Minor: while we're here, should we rename this

```suggestion
Description=Ensure Unique `boot` Filesystem Label
```

to make it less wordy?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-subsequent-sysroot.service
```diff
@@ -1,6 +1,6 @@
 # Note this unit is conditionally enabled by ignition-ostree-generator
 [Unit]
-Description=CoreOS: Mount (subsequent) /sysroot
+Description=CoreOS: Mount (Subsequent) /sysroot
```
jlebon: Wonder if these should be

> CoreOS: Mount /sysroot (First Boot)

and

> CoreOS: Mount /sysroot (Subsequent Boot)

instead.

jlebon (dismissed):

bgilbert:
Updated!

jlebon (approved):

PR #2170, opened 2023-01-12
overlay/35-coreos-live: Lazily unmount /proc/cmdline

cgwalters:
It's possible that some other process has a file descriptor open for `/proc/cmdline`; if so our unmount will fail.  Do the unmount lazily - it was designed for exactly this situation.

This was seen in an OCP installation on vSphere at least once.


bgilbert (approved):
Oh, interesting.  Good catch!

PR #2168, opened 2023-01-11
[next] tree: promote changes from testing-devel at b04dc5eb6ae6a2019a2029628af501b7cb99481f

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/636.

PR #2171, opened 2023-01-12
workflows: normalize the denylist in the promotion-diff workflow

dustymabe:
In https://github.com/coreos/fedora-coreos-releng-automation/pull/165 we started stripping out the snooze lines so we need to do the same here in the origin path to make it so we don't get unwanted warnings.


dustymabe:
I think this should fix the problems shown in https://github.com/coreos/fedora-coreos-config/pull/2168#issuecomment-1379512197 and https://github.com/coreos/fedora-coreos-config/pull/2167#issuecomment-1379509998

bgilbert (approved):
LGTM!

PR #2173, opened 2023-01-13
overrides: add kernel to TRIVIAL_FAST_TRACKS; fast-track kernel-6.1.5-200

dustymabe:


dustymabe:
CI failure hit https://github.com/coreos/fedora-coreos-tracker/issues/1382

Retrying..

ci/overrides.py
```diff
@@ -21,6 +21,9 @@
     # Packages that don't need a reason URL when fast-tracking
     'console-login-helper-messages',
     'ignition',
+    'kernel',
+    'kernel-headers',
+    'kernel-tools',
```
jlebon, line 26: I'm OK with the premise of early testing, but I don't like that it loosens it even for the case where we do actually want to do a kernel fast-track. I wonder if another approach is to just establish a convention that for these "early testing" runs, we put the Bodhi URL in as the reason URL as well?

dustymabe: If there is a reason other than "let's test the kernel early" then we'll put in a URL won't we?

For something like this we could link to some blanket comment about doing kernel fast-tracks, but choosing where to make that comment and find it in the future to link to is kind of annoying.

bgilbert: I'm hesitant.  The kernel is a complex beast and we don't own it, so I'd favor making sure we have clear documentation on why we've fast-tracked.  Otherwise we might accidentally omit a reason when there really is one.

Also, if we do this routinely, how will we make sure an unpushed kernel doesn't land in a release?

dustymabe: The point of this PR was mostly for testing (which CI has done now).

The reason I chose to do the TRIVIAL_FAST_TRACKS thing is because I'm lazy and wanted to just be able to use the GH action :)

I'll close this out since I got the information about CI pass/fail.

PR #2176, opened 2023-01-16
add azure SRIOV udev rules in initramfs; limit coreos-enable-network to Ignition boot

dustymabe:
I noticed recently that Azure is bringing up networking in the initramfs on every boot. This wasn't the intention of c1cf903 I don't believe. I believe the intention was just to make sure that it does come up in the initramfs on first boot (Ignition boot).

Noticed this when investigating
https://github.com/coreos/fedora-coreos-tracker/issues/1383#issuecomment-1384526420


bgilbert:
Both of the Azure-specific reasons cited in the code comment in that commit apply to subsequent boots as well.  On FCOS I believe we defer checkin until the real root, but the hostname part would still be relevant.

jlebon:
I think the canonical way to make a service only run on the Ignition boot is to have it be pulled in by `ignition-complete.target`. That way, it's not even in the transaction at all.

> Both of the Azure-specific reasons cited in the code comment in that commit apply to subsequent boots as well. On FCOS I believe we defer checkin until the real root, but the hostname part would still be relevant.

Hmm, in that case there's a gap there. The `afterburn-hostname.service` only runs on first boot AFAICT: https://github.com/coreos/afterburn/blob/1fa6a6e1091958eb7dc2cdedb0a13ae839bba141/dracut/30afterburn/module-setup.sh#L25

bgilbert:
> Hmm, in that case there's a gap there. The afterburn-hostname.service only runs on first boot AFAICT

Yeah, I was just discovering that.  That seems like a bug.  Hostnames can change if a machine is stopped and restarted.

OTOH, the Afterburn checkin portion turns out not to be relevant on RHCOS either.  The downstream RHCOS checkin unit only runs on first boot, and on subsequent boots we let the upstream unit in the rootfs handle it.

jlebon:
> I think the canonical way to make a service only run on the Ignition boot is to have it be pulled in by ignition-complete.target. That way, it's not even in the transaction at all.

I think `s/initrd.target/ignition-complete.target/` in the associated `module-setup.sh` should do it.

dustymabe:
ok there are now two commits in this PR:

```
commit ebc628a11aa8fa7bf4ce9e6bcbdcddfb89334753
Date:   Mon Jan 16 22:15:32 2023 -0500

    overlay.d: create new 25-azure-udev-rules overlay
    
    We previously had the 68-azure-sriov-nm-unmanaged.rules but we
    figured out we also want those rules in the initramfs. This
    commit makes a new overlay.d entry and groups together a new
    dracut module along with the udev rules in that overlay.d directory.
    
    Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1383

commit 932e6714b535f055b13b35b1f22acc0cb6785eb0
Date:   Mon Jan 16 15:22:57 2023 -0500

    overlay.d: only run coreos-enable-network on first boot
    
    I noticed recently that Azure is bringing up networking in the initramfs
    on every boot [1]. This behavior was originally introduced in c1cf903. After
    some discussion with @jlebon and @bgilbert we are going to convert the
    coreos-enable-network.service to only be pulled in and enalbed on the
    Ignition boot so that it doesn't run on subsequent boots.
    
    [1] https://github.com/coreos/fedora-coreos-tracker/issues/1383#issuecomment-1384526420

```

I still need to do some testing on it.

cgwalters:
One thing I [noticed in the PR discussion](https://github.com/Azure/WALinuxAgent/pull/1622#issuecomment-557137721):

> It would do harm though, if it was present outside Azure, which is why it can't be shipped with NetworkManager.

But we're effectively doing that.  I guess the worry might have been conflicts with anything else sharing that interface name?  Not sure how real that concern is but cc @lkundrak

jlebon:
If we only want to make it apply on Azure, we could add something like this to the rules:

```
IMPORT{cmdline}=="ignition.platform.id=azure"
```

dustymabe:
OK I tested this now. Added a new commit that handles the initramfs case better now.

As far as changing the udev rule as pointed out by @cgwalters and followed up on by @jlebon - maybe we can open a new issue for that? Apparently things seem to be working fine for people today. Maybe the rule is safe to be installed everywhere and it can be just put in NetworkManager's `85-nm-unmanaged.rules`.

cgwalters:
> If we only want to make it apply on Azure, we could add something like this to the rules:

Right, though that'd only work on FCOS derived systems though today.

But of course...say we had an effort to try to unify with the other Fedora Cloud images, nothing actually stops us from trying to add the Ignition platform IDs by default to those images...

cgwalters (approved):

jlebon (approved):
Right, this patch series isn't changing the fact that we're already today applying this everywhere. It just makes it also apply in the initramfs. So if it's the wrong behaviour, at least we're wrong consistently. :)

PR #2180, opened 2023-01-18
Adds mlx[45]_core to the udev rule for the Azure Accelerated Networking interfaces to exclude from the NetworkManager management

aleskandro:
coreos/fedora-coreos-config#2176 introduced an udev rule to avoid the management of the SRIOV network accelerator interfaces on the Azure VMs. The aarch64 Azure VMs' network accelerators use the `mlx[45]_core` modules. This PR makes the introduced rule to also catch the interfaces backed by the `mlx[45]_core` modules.

Refers #1383 OCPBUGS-5569

cc @Prashanth684

### DIFF
--- a/overlay.d/25-azure-udev-rules/usr/lib/udev/rules.d/68-azure-sriov-nm-unmanaged.rules
+++ b/overlay.d/25-azure-udev-rules/usr/lib/udev/rules.d/68-azure-sriov-nm-unmanaged.rules
@@ -2,3 +2,5 @@
 # This interface is transparently bonded to the synthetic interface,
 # so NetworkManager should just ignore any SRIOV interfaces.
 SUBSYSTEM=="net", DRIVERS=="hv_pci", ACTION=="add|change|move", ENV{NM_UNMANAGED}="1"
+IMPORT{cmdline}=="ignition.platform.id=azure", SUBSYSTEM=="net", DRIVERS=="mlx[45]_core", ACTION=="add|change|move", ENV{NM_UNMANAGED}="1"
+


dustymabe:
The `mlx[45]_core` modules are used on x86_64 too, but a parent device of the NIC is matched by the `hv_pci` driver. I'm not sure how the original udev rule for this was developed, but I hesitate to change it.

I'll have a new `aarch64` Azure build I can share with you soon and you can verify.

aleskandro:
Hi, yes.. I did something wrong when live applying your change yesterday. After re-applying it, it's working fine.

Thanks!

freedge:
note that existing rule is already matching.
I am wondering how you tested your change.

dustymabe:
> note that existing rule is already matching. I am wondering how you tested your change.

Let's see if the additional context I added in https://github.com/coreos/fedora-coreos-tracker/issues/1383#issuecomment-1397116857 helps explain things.

aleskandro:
We are facing installation failures for OCP4.12 on arm64, and one of the things we've seen is the mlx5 interfaces is that they have an IP address assigned (this also happens on amd64 nodes, just with less likelihood, for example, 4/6 nodes with no IP, others had it)

In the arm64 clusters I tested with, all the nodes had IPs for the VF interfaces.

I (1) applied a MachineConfig to add the new udev rule as a file in /etc/udev/rules.d/, (2) enabled the tracking of etc files with `rpm-ostree initramfs-etc --track=/etc/udev` and (3) rebooted the nodes.
After these steps, the VFs' IPs are gone. 

Is this verification reliable?

PS: sorry if I messed up the discussion maybe and this might stay better in the issue, but it would not have a clear context.

freedge:
sorry for my comment, I thought this pull request was merged while in fact it's just closed and not merged, and the changes from dustymabe completely fixes it already (tested on x64 so far)
It fixes an extra issue (on your 4/6 nodes with no IP, I believe they waited instead 90s for the DHCP to get an IP, delaying the boot).
What had me very puzzled was that the interface was already NM_UNMANAGED=1 with the current rules, so rules appeared already OK.

dustymabe:
> I (1) applied a MachineConfig to add the new udev rule as a file in /etc/udev/rules.d/, (2) enabled the tracking of etc files with `rpm-ostree initramfs-etc --track=/etc/udev` and (3) rebooted the nodes.
> After these steps, the VFs' IPs are gone.

That should work fine, but it would be nice if after the fix is delivered this could be reversed from the nodes where the workaround was applied.

PR #2181, opened 2023-01-18
ignition-ostree-firstboot-uuid: nuke libblkid cache after UUID restamp

### LINKED ISSUE #1373: Test `iso-offline-install` on multipath on ppc64le and aarch64 failing `coreos-ignition-unique-boot.service` check
```
[2023-01-07T14:56:33.035Z] kola -p qemu-unpriv --output-dir /home/jenkins/agent/workspace/build-arch/tmp/kolaTestIso-r5FDc/kola-testiso-multipath testiso -S --qemu-multipath --scenarios iso-offline-install
...
[2023-01-07T14:59:17.380Z] FAIL: iso-offline-install ( + metal + multipath) (2m41.811s)
[2023-01-07T14:59:17.380Z]     entered emergency.target in initramfs
[2023-01-07T14:59:17.380Z] Error: entered emergency.target in initramfs
```

coreos-installer runs successfully and reboots the machine, and then:

```
Error: System has 2 devices with a filesystem labeled 'boot': ["/dev/sdb3", "/dev/mapper/mpatha3"]
coreos-ignition-unique-boot.service: Main process exited, code=exited, status=1/FAILURE
coreos-ignition-unique-boot.service: Failed with result 'exit-code'.
Failed to start coreos-ignition-unique-boot.service - CoreOS Ensure Unique Boot Filesystem.
```

I suspect something is going wrong with `rdcore verify-unique-fs-label`'s multipath detection.

[iso-offline-install.zip](https://github.com/coreos/fedora-coreos-tracker/files/10373957/iso-offline-install.zip)

jlebon:
We're hitting an issue right now where
`coreos-ignition-unique-boot.service` (backed by `rdcore`) is failing on multipath with:

```
Error: System has 2 devices with a filesystem labeled 'boot': ["/dev/sdb3", "/dev/mapper/mpatha3"]
```

The unique label detection code in `rdcore` determines whether multiple lower-level devices actually refer to the same higher-level device (e.g. multipath or RAID1) by looking at the filesystem UUID. It uses blkid to query device UUIDs.

libblkid maintains a cache of devices to avoid reprobing all devices all the time. This cache normally gets updated (I *think* via udev, but I'm not sure) when changes occur. But something changed recently at least in the multipath case where the cache is only updated for the multipathed device, but not the underlying backing paths.

This then leads `rdcore` to think that they're separate devices. We probably should make `rdcore` smarter here in how it handles multipath devices, but still we don't want to have this stale cache around for the sake of other tools relying on it.

We started hitting this more frequently starting with kernel v6.0.17, but the issue triggers equally as easily on v6.0.16 when reproduced artificially. So I think we've just been lucky so far that this hasn't bit us (possibly we raced with another service that helped refresh the cache).

There's likely a bug here either in the kernel, or multipath or blkid. This is tracked by https://bugzilla.redhat.com/show_bug.cgi?id=2162151. Until then, nuke the blkid cache to force a reprobe on the next call.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1373

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
@@ -62,6 +62,10 @@ if [ "${TYPE}" == "${orig_type}" ] && [ "${UUID}" == "${orig_uuid}" ]; then
     *) echo "unexpected filesystem type ${TYPE}" 1>&2; exit 1 ;;
   esac
   udevadm settle || :
+  # Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2162151.
+  # We nuke the blkid cache containing stale UUIDs so that future blkid calls
+  # (or tools leveraging libblkid) will be forced to re-probe.
+  rm -rf /run/blkid
   echo "Regenerated UUID for ${target}"
 else
   echo "No changes required for ${target} TYPE=${TYPE} UUID=${UUID}"


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
```diff
@@ -62,6 +62,10 @@ if [ "${TYPE}" == "${orig_type}" ] && [ "${UUID}" == "${orig_uuid}" ]; then
     *) echo "unexpected filesystem type ${TYPE}" 1>&2; exit 1 ;;
   esac
   udevadm settle || :
+  # Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2162151.
+  # We nuke the blkid cache containing stale UUIDs so that future blkid calls
+  # (or tools leveraging libblkid) will be forced to re-probe.
+  rm -rf /run/blkid
```
dustymabe, line 68: the only concern I have here is this racing with something else accessing the data at the same time, but I don't think it's enough to block anything here as this clearly is solving a problem that we're hitting in the pipeline multiple times a day.

dustymabe (approved):
LGTM

dustymabe:
Thank you for working on this @jlebon

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-firstboot-uuid
jlebon: I don't think that's a concern. Checking the code quickly, it just does an `open` to open the cache file. If we delete it before, it will treat it as a cache miss and rebuild. If we delete it after, it'll still have an fd to the file.

karelzak:
Don't use cached blkid at all. If you really cannot use lsblk and you really need to read the device then use `blkid -p` to avoid the cache.

jlebon:
> Don't use cached blkid at all. If you really cannot use lsblk and you really need to read the device then use `blkid -p` to avoid the cache.

Ack, thanks for the info. I think there were instances in the past where `lsblk` wouldn't return information that `blkid` would (e.g. https://github.com/coreos/coreos-installer/pull/813), which is why we changed to use `blkid`. But now I wonder if what happened there was that the device became inaccessible somehow and `blkid` worked because it was returning cached data. We'll have to revisit that.

karelzak:
lsblk uses udev DB as a primary source; if udev is unavailable, it falls back to blkid (without cache), but if have a bad experience with lsblk then using blkid is fine, but always with `-p' ;-)

jlebon:
We should be able to revert this once https://github.com/coreos/coreos-installer/pull/1094 gets into FCOS.

jlebon:
> lsblk uses udev DB as a primary source; if udev is unavailable, it falls back to blkid (without cache), but if have a bad experience with lsblk then using blkid is fine, but always with `-p' ;-)

Is there a way to force `lsblk` to not use udev if we suspect that we may be running in a scenario where the udev db is not reliable? I couldn't find any relevant option in `lsblk(8)`. `udevadm settle` is not foolproof unfortunately and we've been bit by its inherent raciness in the past.

karelzak:
> Is there a way to force `lsblk` to not use udev if we suspect that we may be running in a scenario where the udev db is not reliable?

This is currently impossible, added to TODO: https://github.com/util-linux/util-linux/issues/2047

PR #2184, opened 2023-01-19
overlay: always use `-p` with `blkid`

jlebon:
By default, `blkid` will return cached data, which we don't want because it might be stale. Add `-p` to make sure we always bypass the cache. Some of these callsites could probably be changed to use `lsblk`, which uses the udev database, but it's safer to keep using `blkid`.

See also: https://github.com/coreos/fedora-coreos-config/pull/2181#issuecomment-1397386896


dustymabe:
should we not revert e41fd27 as part of this?

jlebon:
> should we not revert [e41fd27](https://github.com/coreos/fedora-coreos-config/commit/e41fd27e3e5dafe671be47c449358aaee984b143) as part of this?

Once this and https://github.com/coreos/coreos-installer/pull/1094 are in. I'll stage a revert PR.

dustymabe (approved):
LGTM

PR #2188, opened 2023-01-19
overlay/live-generator: Allow 50% of RAM for /etc and /var

### LINKED ISSUE #1344: Live environment only allocates 20% of RAM for ephemeral `/var`
### Describe the bug

I am provisioning my Fedora Core OS from RAM. However when running I see the filesystem is of 683M only (iso size I suppose).

The docs do not explain if I am supposed to resize the filesystem or what and when using the OS I have no space left on device.

### Reproduction steps

1. Create iPXE install script: https://gist.github.com/UnconventionalMindset/f4ca2d66b943f916b5fecfad698b47ea
2. Create your installation Butane and generate an Ignition from it: https://github.com/UnconventionalMindset/coreos-setup/blob/main/coreos.bu
3. Embed it to iPXE: https://gist.github.com/UnconventionalMindset/9c9daa66727e7f45a85bad30a7844caa
4. Run it in Proxmox
5. Pull an image like the one from LinuxServer.io of homeassistant.

### Expected behavior

Filesystem should be as big as the size of the RAM and the image should be pulled successfully.

### Actual behavior

Filesystem is as big as the size of the ISO and the image cannot be saved due to not enough space left.

image

### System details

- 36.20221001.3.0 stable
- Proxmox VE 7.2-11
- iPXE install script: https://gist.github.com/UnconventionalMindset/f4ca2d66b943f916b5fecfad698b47ea
- Script to embed to iPXE: https://gist.github.com/UnconventionalMindset/9c9daa66727e7f45a85bad30a7844caa

### Ignition config

Install butane: https://github.com/UnconventionalMindset/coreos-setup/blob/main/coreos.bu
Install ignition: https://github.com/UnconventionalMindset/coreos-setup/blob/main/coreos.ign

### Additional information

_No response_

cgwalters:
Today we default to sizing the ephemeral storage at the tmpfs default
of only 20%, but the goal here is to be able to run fully from RAM,
so there's no reason not to allow more.

(We should also likely enable zram here, but that's another topic)

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1344

---

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
@@ -168,9 +168,22 @@ EOF
     add_requires workaround-stalled-media-iso-mount.service basic.target
 fi
 
-# It turns out that `tmpfs` currently munches all SELinux labels
+# The systemd default for /run is 20% of RAM, but we want to use basically all of RAM,
+# so we make a distinct tmpfs that allows more.  Note the file we create there is
+# "thinly provisioned", so we're not actually allocating all of that.
+cat >>"${UNIT_DIR}/run-ephemeral_base.mount" <<EOF
+[Unit]
+DefaultDependencies=false
+[Mount]
+What=tmpfs
+Where=/run/ephemeral_base
+Type=tmpfs
+Options=size=50%%
+EOF
+
+# But it turns out that `tmpfs` currently munches all SELinux labels
 # we set before policy is loaded, so we make an XFS filesystem
-# loopback mounted that's sized the same as /run.
+# loopback mounted on top of the base tmpfs.
 # https://github.com/coreos/fedora-coreos-config/pull/499
 cat >"${UNIT_DIR}/sysroot-xfs-ephemeral-mkfs.service" <<'EOF'
 [Unit]
@@ -179,14 +192,15 @@ DefaultDependencies=false
 # can run really early.
 After=systemd-tmpfiles-setup-dev.service
 ConditionPathExists=/usr/lib/initrd-release
+RequiresMountsFor=/run/ephemeral_base
 # Something seems to be causing us to rerun?
-ConditionPathExists=!/run/ephemeral
+ConditionPathExists=!/run/ephemeral_base/loopfs
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/bin/sh -c 'set -euo pipefail; mem=$$(($$(stat -f -c "%%b * %%s / 1024" /run))) && /bin/truncate -s $${mem}k /run/ephemeral.xfsloop'
-ExecStart=/sbin/mkfs.xfs /run/ephemeral.xfsloop
+ExecStart=/bin/sh -c 'set -euo pipefail; mem=$$(($$(stat -f -c "%%b * %%s / 1024" /run/ephemeral_base))) && /bin/truncate -s $${mem}k /run/ephemeral_base/loopfs'
+ExecStart=/sbin/mkfs.xfs /run/ephemeral_base/loopfs
 ExecStart=/bin/mkdir /run/ephemeral
 EOF
 add_requires sysroot-xfs-ephemeral-mkfs.service initrd-root-fs.target
@@ -197,7 +211,7 @@ DefaultDependencies=false
 Requires=sysroot-xfs-ephemeral-mkfs.service
 After=sysroot-xfs-ephemeral-mkfs.service
 [Mount]
-What=/run/ephemeral.xfsloop
+What=/run/ephemeral_base/loopfs
 Where=/run/ephemeral
 Type=xfs
 Options=loop,discard


bgilbert:
Hmm, is it safe to set the limit that high?  tmpfs defaults to 50%, and I can imagine that we'd create some memory pressure if 90% of RAM is unreclaimable.

cgwalters:
In this scenario, anything that's not on the original disk image is unreclaimable.  What this enables is workloads that pull potentially large container images, but the launched processes don't actually take a lot of RAM.

I think this is actually a real use case with things like the OCP assisted installer that pull container images that inspect the host to gather data to send up to a provisioning server before running `coreos-installer`.  Those images may be large, but their workload is basically "send a HTTP request and long poll for a result" - there's not a lot of allocated RAM.

I'm not opposed to going to 50% if it'd help you feel more comfortable; most real-world use cases of the Live ISO are realistically at least 32GB of RAM, and I can't imagine people pulling 16GB of containers just to do an install.  It's even unlikely that a small 8GB device pulls 4GB of containers.

(But, it is more likely that an 8GB device pulls 1.6G of containers (default tmpfs 20% scenario) I think...and that's how we got here)

jlebon:
Rather than making it percentage based, how about making it a fixed amount of base memory we reserve for the host (e.g. 1G)? 90% on a 32G host is very different than 90% on a 2G host (the currently documented minimum requirement for PXE boot).

dustymabe:
OR make it have a minimum amount. i.e. 1G is reserved for the host at a minimum. So we could do a higher percentage, but bail out of that percentage if the amount left for the host would be less than some threshold. I think this is similar to how SwapOnZram calculates how much memory it should use.

cgwalters:
Hold on a second here...this loopback file is *thinly provisioned*.  We're not hard reserving 90% of RAM.

jlebon:
Previous discussions on this topic in https://github.com/coreos/fedora-coreos-config/pull/685#discussion_r504917352. There, it basically came down to "would you rather have the OOMkiller randomly kill processes (or in the worst case, have the kernel crash) or get an `ENOSPC` earlier?". I think the same applies here.

Bumping the default to 50% seems reasonable to me. But the user in the ticket is running workloads without a persistent `/var` and they might want even more. Maybe we should consider eventually making this configurable via a karg.

dustymabe:
`50%` WFM. It handles our small case (on `2Gi` you get `1Gi` for host and `1Gi` for FS) and also gives us substantial gain over our current allocation of `20%`. We can always adjust in the future if needed.

I'd prefer not to have a kernel arg for now if we can avoid it.

cgwalters:
OK, moved to 50%

jlebon (dismissed):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -179,14 +192,16 @@ DefaultDependencies=false
...
 # Something seems to be causing us to rerun?
-ConditionPathExists=!/run/ephemeral
+ConditionPathExists=!/run/ephemeral_base/loopfs
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/bin/sh -c 'set -euo pipefail; mem=$$(($$(stat -f -c "%%b * %%s / 1024" /run))) && /bin/truncate -s $${mem}k /run/ephemeral.xfsloop'
-ExecStart=/sbin/mkfs.xfs /run/ephemeral.xfsloop
+# Stat gives us units of 4k blocks, so multiply by 4
```
dustymabe: is it possible the size would ever be different than 4k? Should we still get that value from `%s` and do the math like we were doing before?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -179,14 +192,16 @@ DefaultDependencies=false
...
+ExecStart=/bin/sh -c 'set -euo pipefail; mem=$$(stat -f -c '%%f' /run/ephemeral_base) && /bin/truncate -s $$(($$mem * 4))k /run/ephemeral_base/loopfs'
```
dustymabe: we're using `%f` here (free blocks in the filesystem) as opposed to `%b` (total blocks in the filesystem) because something else is using some of that space? Sorry it wasn't clear to me.

dustymabe (dismissed):
LGTM. I asked a few questions for my own understanding.

dustymabe:
@cgwalters - ready to merge this?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
cgwalters: Nothing else is using that space, but...if something did somehow start to, we'd want to use free blocks right?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
cgwalters: Ah, good point.  Actually I think block size on tmpfs is equal to page size, which is indeed not 4k on e.g. ppc64le.

cgwalters: *This* type of bug is why we have code review; I think what happened here is I went to test out some of this shell script manually and ended up changing things while doing so.

You correctly asked "why did the logic change" when it shouldn't have.

cgwalters:
Pushed a fix for the bug you spotted, I'm retesting locally now.

cgwalters:
Yep, works.

dustymabe (approved):

PR #2189, opened 2023-01-20
overlay: rename azure overlay to conform with others

dustymabe:
The other overlays don't have a `-` between the number and the start of the name so let's drop the dash here.


dustymabe:
When I saw this it triggered my OCD :( - sorry for the churn.

cgwalters (approved):

PR #2197, opened 2023-01-23
overrides: fast-track qemu-7.0.0-13.fc37

coreosbot-releng:


jlebon (approved):
LGTM, but note this package isn't in x86_64. The compose will still work, but I think it'll prevent auto-graduation to trigger so it'll have to be cleared manually.

dustymabe:
> LGTM, but note this package isn't in x86_64.

Good point. I'll keep an eye on it. It also means that CI won't test this change so I'm merging without waiting on CI to run.

PR #2199, opened 2023-01-25

coreosbot-releng:


dustymabe:
oh hey. It worked @jlebon :)

dustymabe (approved):
LGTM

jlebon:
Ahh nice. Looks like `qemu-user-static-x86` does exist for x86_64 even though we don't ship it. That's why the script still worked.

PR #2200, opened 2023-01-26
[rhcos-4.9] 05core/coreos-liveiso-success: fix hang with systemd 239

ravanelli:
RHEL 8 has systemd 239, which has
https://github.com/systemd/systemd/issues/9374.  As a result, "journalctl -f ... | head -1" hangs because journalctl doesn't properly exit.  Rewrite the check to call loginctl once per second until it finds a session.  This approach is uglier, but the service only runs in the live ISO, on QEMU, when there's no Ignition config, and when kola's virtio port is open, so the minimal extra overhead shouldn't affect production systems.

Fixes the kola testiso iso-live-login scenario in RHCOS.

(cherry picked from commit a2aef3ee6194dcba0d9c507a3479a393d7787844)


dustymabe (dismissed):
LGTM

dustymabe:
may want to also backport 5be9e14 so we don't see CI failures for this older branch.

jlebon (approved):

PR #2201, opened 2023-01-26
coreos-livepxe-rootfs: use `curl --fail`

### LINKED ISSUE #1390: PXE boot fails with confusing "hash mismatch" error if the rootfs does not exist on the server
Hi all.

it appears that the pxe live initramfs for the 37.20230110.3.1 image that was just released over night has the wrong hash of the live rootfs embedded within it. This works perfectly for the prior stable release, and existing operational fedora coreos instances upgrade to this release perfectly but any new installations using pxe on this release fail.

image

jlebon:
Currently, if the server returns a 404 when fetching the rootfs, we
still try to hash the server's HTML response and compare it, which
leads to a confusing error message about mismatched hashes.

Do a single `HEAD` after establishing server connectivity, but before
the pipeline. Use `--fail` so that `curl` exits with a nonzero code and
clearly prints out the erroring HTTP code it got. Also use `--fail`
during the pipeline invocation for the same reason.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1390

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
@@ -59,6 +59,10 @@ elif [[ -n "${rootfs_url}" ]]; then
         echo "Couldn't fetch, verify, and unpack image specified by:" >&2
         echo "coreos.live.rootfs_url=${rootfs_url}" >&2
         echo "Check that the URL is correct and that the rootfs version matches the initramfs." >&2
+        source /etc/os-release
+        if [ -n "${OSTREE_VERSION:-}" ]; then
+            echo "The version of this initramfs is ${OSTREE_VERSION}." >&2
+        fi
         exit 1
     fi
 else
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
@@ -47,12 +47,24 @@ elif [[ -n "${rootfs_url}" ]]; then
         sleep 5
     done
 
+    # We shouldn't need a --retry from here on since we've just successfully
+    # HEADed the file, but let's add one just to be safe (e.g. if the
+    # connection just went online and flickers or something).
+    curl_common_args+=" --retry 5"
+
+    # Do a HEAD again but just once and with `--fail` so that if e.g. it's
+    # missing, we get a clearer error than if it were part of the pipeline
+    # below.
+    curl_common_args+=" --fail"
+    if ! curl --head $curl_common_args "${rootfs_url}" >/dev/null; then
+        echo "Couldn't query the server for the rootfs specified by:" >&2
+        echo "coreos.live.rootfs_url=${rootfs_url}" >&2
+        exit 1
+    fi
+
     # bsdtar can read cpio archives and we already depend on it for
     # coreos-liveiso-persist-osmet.service, so use it instead of cpio.
-    # We shouldn't need a --retry here since we've just successfully HEADed the
-    # file, but let's add one just to be safe (e.g. if the connection just went
-    # online and flickers or something).
-    if ! curl $curl_common_args --retry 5 "${rootfs_url}" | \
+    if ! curl $curl_common_args "${rootfs_url}" | \
             rdcore stream-hash /etc/coreos-live-want-rootfs | \
             bsdtar -xf - -C / ; then
         echo "Couldn't fetch, verify, and unpack image specified by:" >&2
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
@@ -37,7 +37,8 @@ elif [[ -n "${rootfs_url}" ]]; then
     # Doing this allows us to retry all errors (including transient
     # "no route to host" errors during startup). Note we can't use
     # curl's --retry-all-errors here because it's not in el8's curl yet.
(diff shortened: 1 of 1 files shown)


jlebon:
(Not tested yet; just throwing this up since I think it would've helped with https://github.com/coreos/fedora-coreos-tracker/issues/1390.)

dustymabe (approved):
Trivial LGTM

jlebon:
Tested working now!

dustymabe (dismissed):

jlebon:
OK since the original commit wasn't significant, I decided to convert this PR to submit a more important patch that gets to the root of https://github.com/coreos/fedora-coreos-tracker/issues/1390.

dustymabe (approved):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
```diff
@@ -46,19 +47,33 @@ elif [[ -n "${rootfs_url}" ]]; then
...
     done
 
-    # We don't need to verify TLS certificates because we're checking the
-    # image hash.
+    # We shouldn't need a --retry from here on since we've just successfully
+    # HEADed the file, but let's add one just to be safe (e.g. if the
+    # connection just went online and flickers or something).
+    curl_common_args+=" --retry 5"
+
+    # Do a HEAD again but just once and with `--fail` so that if e.g. it's
```
cgwalters, line 55: "just once" seems like an incorrect comment since it still has `--retry 5` no?

jlebon: "just once" is in contrast to the code just above where we also do `HEAD` but in a loop. The `--retry` here catches transient errors and has its own rationale above. If we get a 404, curl doesn't retry.

bgilbert (approved):
I'm okay with this, but it does feel a little redundant to HEAD the object twice.  I'd be tempted to skip the second HEAD and just pass `--fail` to the real fetch.  `rdcore` will throw a spurious error, sure, but our error message afterward is sufficiently general to cover this case, and we'll still have curl's error to explain what's going on.

jlebon:
> I'm okay with this, but it does feel a little redundant to HEAD the object twice. I'd be tempted to skip the second HEAD and just pass `--fail` to the real fetch. `rdcore` will throw a spurious error, sure, but our error message afterward is sufficiently general to cover this case, and we'll still have curl's error to explain what's going on.

Heh, that's actually what I originally had. The reason I switched it (on top of just making it easier to understand) is that all the spurious errors are enough to rotate out of view the real error from `curl` in [the 10 lines we output at the emergency shell](https://github.com/coreos/fedora-coreos-config/blob/1a305128e5979aba7f2ebfc04170a0deb679ec0a/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-shell-setup/emergency-shell.sh#L66-L67). So you get:

```
Jan 30 15:13:53 coreos-livepxe-rootfs[655]: bsdtar: Failed to set default locale
Jan 30 15:13:54 coreos-livepxe-rootfs[654]: Error: premature end of input data at offset 0
Jan 30 15:13:54 coreos-livepxe-rootfs[632]: Couldn't fetch, verify, and unpack image specified by:
Jan 30 15:13:54 coreos-livepxe-rootfs[632]: coreos.live.rootfs_url=http://10.0.2.2:26766/fedora-coreos-37.20230118.dev.0-live-rootfs.x86_64.img
Jan 30 15:13:54 coreos-livepxe-rootfs[632]: Check that the URL is correct and that the rootfs version matches the initramfs.
Jan 30 15:13:54 coreos-livepxe-rootfs[632]: The version of this initramfs is 37.20230118.dev.0.
Jan 30 15:13:54 systemd[1]: coreos-livepxe-rootfs.service: Main process exited, code=exited, status=1/FAILURE
Jan 30 15:13:54 systemd[1]: coreos-livepxe-rootfs.service: Failed with result 'exit-code'.
Jan 30 15:13:54 systemd[1]: Failed to start coreos-livepxe-rootfs.service - Acquire Live PXE rootfs Image.
Jan 30 15:13:54 systemd[1]: coreos-livepxe-rootfs.service: Triggering OnFailure= dependencies.
```

vs with the preliminary HEAD:

```
Jan 30 15:16:52 coreos-livepxe-rootfs[639]: Fetching rootfs image from http://10.0.2.2:14239/fedora-coreos-37.20230118.dev.0-live-rootfs.x86_64.img...
Jan 30 15:16:52 coreos-livepxe-rootfs[660]: curl: (22) The requested URL returned error: 404
Jan 30 15:16:52 coreos-livepxe-rootfs[639]: Couldn't query the server for the rootfs specified by:
Jan 30 15:16:52 coreos-livepxe-rootfs[639]: coreos.live.rootfs_url=http://10.0.2.2:14239/fedora-coreos-37.20230118.dev.0-live-rootfs.x86_64.img
Jan 30 15:16:52 systemd[1]: coreos-livepxe-rootfs.service: Main process exited, code=exited, status=1/FAILURE
Jan 30 15:16:52 systemd[1]: coreos-livepxe-rootfs.service: Failed with result 'exit-code'.
Jan 30 15:16:52 systemd[1]: Failed to start coreos-livepxe-rootfs.service - Acquire Live PXE rootfs Image.
Jan 30 15:16:52 systemd[1]: coreos-livepxe-rootfs.service: Triggering OnFailure= dependencies.
```

Obviously, you'd get the 404 message if you did a `journalctl -u coreos-livepxe-rootfs` at the console, but the clearer error seemed worth the extra code.

bgilbert:
Ahh, okay.  Might be worth a comment, but okay as is too. :+1:

jlebon:
> Ahh, okay. Might be worth a comment, but okay as is too. +1

:+1: I'll merge it to not waste CI but will add it in a separate PR.

jlebon:
> > Ahh, okay. Might be worth a comment, but okay as is too. +1
> 
> +1 I'll merge it to not waste CI but will add it in a separate PR.

https://github.com/coreos/fedora-coreos-config/pull/2223

PR #2202, opened 2023-01-27
kola/setuid: Update known files with SetUID bit set

### LINKED ISSUE #1391: kola: rawhide: `ext.config.files.setuid` fails due to additional unknown file with SetUID/GID bit set
### Describe the bug

There's an additional unknown file (`/usr/libexec/openssh/ssh-keysign`) with SetUID/GID bit set which leads to `ext.config.files.setuid` test failure. Not too sure if the addition of this file is intentional or not.

### Reproduction steps

1. Run `cosa kola run ext.config.files.setuid`

### Expected behavior

It passes with no additional unknown file(s) with SetUID/GID bit set

### Actual behavior

It fails with additional unknown file(s) with SetUID/GID bit set. That unknown file is `/usr/libexec/openssh/ssh-keysign`

### System details

[rawhide][x86_64] ⚡ 38.20230125.91.0

gursewak1997:
Update known files with SetUID bit set. Reason: https://github.com/coreos/fedora-coreos-tracker/issues/1391#issuecomment-1406299597

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1391

### DIFF
--- a/tests/kola/files/setgid
+++ b/tests/kola/files/setgid
@@ -6,6 +6,9 @@ set -xeuo pipefail
 . $KOLA_EXT_DATA/commonlib.sh
 
 # List of known files and directories with SetGID bit set
+# Drop '/usr/libexec/openssh/ssh-keysign' after
+# https://src.fedoraproject.org/rpms/openssh/c/b615362fd0b4da657d624571441cb74983de6e3f?branch=rawhide
+# is in all OS.
 list_setgid_files=(
     '/usr/bin/write'
     '/usr/libexec/openssh/ssh-keysign'
--- a/tests/kola/files/setuid
+++ b/tests/kola/files/setuid
@@ -20,6 +20,7 @@ list_setuid_files=(
     '/usr/bin/sudo'
     '/usr/bin/umount'
     '/usr/lib/polkit-1/polkit-agent-helper-1'
+    '/usr/libexec/openssh/ssh-keysign'
     '/usr/sbin/grub2-set-bootflag'
     '/usr/sbin/mount.nfs'
     '/usr/sbin/pam_timestamp_check'


dustymabe:
I think we also need to drop the file from the setgid test list.

dustymabe:
> I think we also need to drop the file from the setgid test list.

Hmm. But if we do that it poses a problem because the file is still setgid in F37 and also does this test run against RHCOS?

cc @travier

gursewak1997:
> I think we also need to drop the file from the setgid test list.

Hmm, I think I need to go further into the bugs specified in the issue by travier. The setgid test failed locally when I dropped the specified file from there.

dustymabe:
> The setgid test failed locally when I dropped from there.

Were you testing against rawhide?

gursewak1997:
> Were you testing against rawhide?

Yes

cgwalters:
This test is really a perfect example of something we should be upstreaming into Fedora and not gate on in our own tests.

dustymabe:
> This test is really a perfect example of something we should be upstreaming into Fedora and not gate on in our own tests.

I'm all for it.

dustymabe:
> > I think we also need to drop the file from the setgid test list.
> 
> Hmm, I think I need to go further into the bugs specified in the issue by travier. The setgid test failed locally when I dropped the specified file from there.

I just tested it and it PASSED for me when testing against `rawhide`:

```patch
diff --git a/tests/kola/files/setgid b/tests/kola/files/setgid
index 63df1b07..91073398 100755
--- a/tests/kola/files/setgid
+++ b/tests/kola/files/setgid
@@ -8,7 +8,6 @@ set -xeuo pipefail
 # List of known files and directories with SetGID bit set
 list_setgid_files=(
     '/usr/bin/write'
-    '/usr/libexec/openssh/ssh-keysign'
     '/usr/libexec/utempter/utempter'
 )
 
diff --git a/tests/kola/files/setuid b/tests/kola/files/setuid
index 51a0494a..d173ec76 100755
--- a/tests/kola/files/setuid
+++ b/tests/kola/files/setuid
@@ -20,6 +20,7 @@ list_setuid_files=(
     '/usr/bin/sudo'
     '/usr/bin/umount'
     '/usr/lib/polkit-1/polkit-agent-helper-1'
+    '/usr/libexec/openssh/ssh-keysign'
     '/usr/sbin/grub2-set-bootflag'
     '/usr/sbin/mount.nfs'
     '/usr/sbin/pam_timestamp_check'
```


though there is still the open question of what we should do about F37 and RHEL. I imagine maybe we leave the entry in the setgid test list but add a comment that it can be dropped from that list once https://src.fedoraproject.org/rpms/openssh/c/b615362fd0b4da657d624571441cb74983de6e3f?branch=rawhide is in all OS that we care about.

gursewak1997:
> I just tested it and it PASSED for me when testing against `rawhide`:
> 

Huh, kinda weird why it is consistently failing for me "against" rawhide. Adding a comment to setgid test for now.

travier:
It's fine indeed keeping the entry in setgid for now.

travier (dismissed):

travier:
> This test is really a perfect example of something we should be upstreaming into Fedora and not gate on in our own tests.

There are tests like that in Fedora somewhere. Unfortunately I'm afraid that nobody looks at them.

jlebon:
Related: https://github.com/coreos/fedora-coreos-tracker/issues/1394

dustymabe (dismissed):
LGTM

dustymabe:
CI failure should be fixed now that https://github.com/coreos/fedora-coreos-config/pull/2203 merged.

Could you rebase on that and also update your git commit message to have more information (like the description of this PR has)?

dustymabe:
actually - you need to change the target branch for this to merge into to be `testing-devel` since the tests themselves get synced to other branches from `testing-devel` by config-bot.

dustymabe (approved):
LGTM

PR #2204, opened 2023-01-29
tests/kola: test Ignition spec 3.4.0 LUKS `discard` and `openOptions` fields

bgilbert:
Rather than adding a dedicated test for relatively obscure options, change the rootfs test to enable them, check that it does so, and check that the existing LUKS `/var/log` test doesn't do so.  In each case, check both the first and second boots.

The rootfs is the important case, since its mount flow doesn't read `/etc/crypttab`.

### DIFF
--- a/tests/kola/root-reprovision/luks/config.ign
+++ b/tests/kola/root-reprovision/luks/config.ign
@@ -1,6 +1,6 @@
 {
   "ignition": {
-    "version": "3.2.0"
+    "version": "3.4.0"
   },
   "storage": {
     "luks": [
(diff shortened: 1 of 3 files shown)


tests/kola/root-reprovision/luks/config.ign
```diff
@@ -10,6 +10,7 @@
         "clevis": {
           "tpm2": true
         },
+        "discard": true,
```
jlebon, line 13: While we're here, should we also test out `openOptions`? E.g. add `--perf-no_read_workqueue` and verify it's persistent?

jlebon (dismissed):

tests/kola/root-reprovision/luks/config.ign
bgilbert: Done!

jlebon (dismissed):

jlebon (approved):

PR #2206, opened 2023-01-31
Removing ext.config.kdump.crash

Shilpi-Das1:
This PR removes ext.config.kdump.crash from kola deny list
Issue Link: https://github.com/coreos/coreos-assembler/issues/2725

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -5,17 +5,6 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: podman.workflow
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
-- pattern: ext.config.kdump.crash
-  tracker: https://github.com/coreos/coreos-assembler/issues/2725#issuecomment-1292616121
-  snooze: 2023-02-04
-  arches:
-    - ppc64le
-  streams:
-    - next-devel
-    - next
-    - testing-devel
-    - testing
-    - stable
 - pattern: coreos.boot-mirror.luks
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:


kola-denylist.yaml
```diff
@@ -5,17 +5,6 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: podman.workflow
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
-- pattern: ext.config.kdump.crash
-  tracker: https://github.com/coreos/coreos-assembler/issues/2725#issuecomment-1292616121
-  snooze: 2023-02-04
-  arches:
-    - ppc64le
-  streams:
```
ravanelli, line 13: We are not building ppc64le for some streams, as the fix probably wasn't promoted to other streams, let's  keep it as:
```suggestion
streams:
     next-devel
     next
     testing
     stable
  ```

Comment (dustymabe) [reply kola-denylist.yaml:13]: if @Shilpi-Das1 tested this against `testing-devel` (Fedora 37) and it worked then we can delete all the entries (i.e. we can merge this PR like it is now).

**Comment by dustymabe** [comment]
let's update the commit message to be more meaningful:

```
denylist: remove ext.config.kdump.crash denial

The underlying issue for kdump crash test failure on ppc64le
has been fixed in the upstream kernel [1]. We can remove
this test from the denylist now since the fix has made it to
all streams that we care about.

[1] https://github.com/coreos/coreos-assembler/issues/2725#issuecomment-1292616121
```

kola-denylist.yaml
dustymabe: Just so it is clear.. we are already running this test for `aarch64`, `x86_64`, and `s390x` everywhere, now we will for `ppc64le` too!

ravanelli: A good. I wasn't sure if it had landed in all streams. No changes needed in this case, just ignore my comments above

Shilpi-Das1:
@dustymabe @ravanelli Resolved comment Please review it again. Thanks

Shilpi-Das1:
> let's update the commit message to be more meaningful:
> 
> ```
> denylist: remove ext.config.kdump.crash denial
> 
> The underlying issue for kdump crash test failure on ppc64le
> has been fixed in the upstream kernel [1]. We can remove
> this test from the denylist now since the fix has made it to
> all streams that we care about.
> 
> [1] https://github.com/coreos/coreos-assembler/issues/2725#issuecomment-1292616121
> ```

Resolved this comment.

ravanelli (approved):

dustymabe (approved):

PR #2207, opened 2023-01-31
tests/kola/ntp: use different server for NTP for DHCP propagation tests

dustymabe:
The time-c-g.nist.gov can't resolve in our Fedora OpenShift because they require DNSSEC validation and that server started failing. I picked another server on the https://tf.nist.gov/tf-cgi/servers.cgi page that is passing validation and also added some commentary.


bgilbert (approved):
LGTM

PR #2210, opened 2023-02-01
denylist: drop ntp.timesyncd.dhcp-propagation from denylist

dustymabe:
This test was denylisted for rawhide because the fix for [1],
which was an updated container-selinux RPM, was held back [2]
because it needed a newer selinux-policy RPM, which was pinned
because of [3]. Now that selinux-policy and container-selinux
are no longer pinned, we can start running this test in rawhide.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1359
[2] https://github.com/coreos/fedora-coreos-tracker/issues/1364#issuecomment-1371727078
[3] https://github.com/coreos/fedora-coreos-tracker/issues/1364

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -40,8 +40,3 @@
 - pattern: ext.config.platforms.aws.nvme
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1306#issuecomment-1378864963
   snooze: 2023-02-10
-- pattern: ext.config.ntp.timesyncd.dhcp-propagation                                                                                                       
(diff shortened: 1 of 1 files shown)


jlebon:
Hmm, the commit message doesn't seem to match the commit title.

dustymabe:
> Hmm, the commit message doesn't seem to match the commit title.

The commit message is accurate, but it's not super clear because it's an indirect relationship.

I updated it to hopefully be more clear.

jlebon (approved):

PR #2215, opened 2023-02-02
[rawhide] overrides: drop grub2-2.06-78.fc38 pin

dustymabe:
The new grub2-2.06-80.fc38 (and newer) have the signing problem fixed.

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1403


ravanelli (approved):
lgtm

PR #2223, opened 2023-02-08
coreos-livepxe-rootfs: add comment about separate `--fail`

jlebon:
Follow-up to 84905567 ("coreos-livepxe-rootfs: use `curl --fail`") and review feedback in #2201.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
@@ -54,7 +54,10 @@ elif [[ -n "${rootfs_url}" ]]; then
 
     # Do a HEAD again but just once and with `--fail` so that if e.g. it's
     # missing, we get a clearer error than if it were part of the pipeline
-    # below.
+    # below. Otherwise, the `curl` error emitted there would get lost among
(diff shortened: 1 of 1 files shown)


jlebon:
(Killed CI for this. Not worth the resources.)

bgilbert:
Mmm, I find the new text hard to parse (and it has a typo).  Could we clarify/expand a bit?

jlebon:
Updated! Is that better?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
```diff
@@ -54,7 +54,10 @@ elif [[ -n "${rootfs_url}" ]]; then
 
     # Do a HEAD again but just once and with `--fail` so that if e.g. it's
     # missing, we get a clearer error than if it were part of the pipeline
-    # below.
+    # below (otherwise, the `curl` error emitted there would get lost among
```
bgilbert: Optional: parenthetical could be a full separate sentence instead.

bgilbert (dismissed):
:laughing: Definitely expanded upon, yes.  Thanks.

bgilbert (approved):

dustymabe (approved):
LGTM

PR #2224, opened 2023-02-10
denylist: extend snooze again for platforms.aws.nvme test

dustymabe:


dustymabe:
Will skip CI on this one because it won't test this change anyway.

jlebon (approved):

PR #2226, opened 2023-02-13
ignition-ostree-transposefs: add workaround when removing the zram device

### LINKED ISSUE #1149: [4.13-9.2/aarch64,ppc64le] luks.sss.* failure
```
14:12:30  --- FAIL: luks.sss.t1 (164.48s)
14:12:30          luks.go:171: Unable to create test machine: machine 62293c5f-e7d6-4e82-9157-4fa8e27499f9 entered emergency.target in initramfs
14:17:28  --- FAIL: luks.tang (155.82s)
14:17:28          luks.go:171: Unable to create test machine: machine f20837d6-42e7-4992-a5d5-b9a9f1f3ef48 entered emergency.target in initramfs
...
14:33:51  --- FAIL: luks.sss.t2 (151.35s)
14:33:51          luks.go:171: Unable to create test machine: machine fc34fc8a-fdb4-423e-a877-944067354a0f entered emergency.target in initramfs
```

This is a frequent flake, but only on ppc64le.  The failure here is:

```
[   91.620717] ignition-ostree-transposefs[2224]: /usr/libexec/ignition-ostree-transposefs: line 293: echo: write error: Device or resource busy
```

Code is https://github.com/coreos/fedora-coreos-config/blob/65d942edf9cc56597ccb3e9ad8d51389380b00df/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh#L293

Something seems to be holding the zram block device open...I'd need to do a custom build to determine whether a retry loop would help.

Edit(jl): logs from a run I reproduced: [logs.zip](https://github.com/openshift/os/files/10734004/logs.zip)

HuijingHei:
Add workaround for luks.tang and luks.sss.* tests failure on aarch64 and ppc64le.
Fix https://github.com/openshift/os/issues/1149

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -24,6 +24,11 @@ saved_prep=${saved_data}/prep
 zram_dev=${saved_data}/zram_dev
 partstate_root=/run/ignition-ostree-rootfs-partstate.sh
 
+is_rhcos9() {
+    source /etc/os-release
+    [ "${ID}" == "rhcos" ] && [ "${RHEL_VERSION%%.*}" -eq 9 ]
+}
+
 # Print jq query string for wiped filesystems with label $1
 query_fslabel() {
     echo ".storage?.filesystems? // [] | map(select(.label == \"$1\" and .wipeFilesystem == true))"
@@ -289,6 +294,12 @@ case "${1:-}" in
         if [ -d "${saved_data}" ]; then
             read dev < "${zram_dev}"
             umount "${saved_data}"
+            # After unmounting, make sure zram device state is stable before we remove it.
+            # See https://github.com/openshift/os/issues/1149
+            # Should remove when https://bugzilla.redhat.com/show_bug.cgi?id=2172058 is fixed.
+            if [ $(uname -m) != x86_64 ] && is_rhcos9; then
+                udevadm trigger --settle "/dev/zram${dev}"
+            fi
             rm -rf "${saved_data}" "${partstate_root}"
             echo "${dev}" > /sys/class/zram-control/hot_remove
         fi


HuijingHei:
The related tests will not be triggered until the patch is synced, and remove them in deny lists refer to https://github.com/openshift/os/pull/1150 and https://github.com/openshift/os/pull/1155

Do testing locally with the patch on ppc64le, result is PASSED except the `podman-system-generator` error, will confirm with podman container team. Not do any testing on aarch64 yet.
```
$ kola run luks.tang luks.sss.t1 luks.sss.t2
=== RUN   luks.sss.t1
=== RUN   luks.sss.t2
=== RUN   luks.tang

qemu-system-ppc64: warning: kernel_irqchip allowed but unavailable: IRQ_XIVE capability must be present for KVM
Falling back to kernel-irqchip=off
qemu-system-ppc64: warning: kernel_irqchip allowed but unavailable: IRQ_XIVE capability must be present for KVM
Falling back to kernel-irqchip=off
qemu-system-ppc64: warning: kernel_irqchip allowed but unavailable: IRQ_XIVE capability must be present for KVM
Falling back to kernel-irqchip=off
2023-02-13T05:46:05Z kola: Found systemd generator failure (/usr/lib/systemd/system-generators/podman-system-generator) on machine 460145ba-5191-493f-81ff-17032e2e037c console
2023-02-13T05:46:05Z kola: Found systemd generator failure (/usr/lib/systemd/system-generators/podman-system-generator) on machine 39dbeb79-3699-4042-ac03-fe063a0fa99d console
--- PASS: luks.sss.t1 (212.75s)

<same logs with luks.sss.t1>
--- PASS: luks.tang (188.99s)

<same logs with luks.sss.t1>
--- PASS: luks.sss.t2 (193.50s)
PASS, output in tmp/kola/qemu-unpriv-2023-02-13-0542-84342
```

HuijingHei:
`podman-system-generator` error is fixed by https://github.com/containers/podman/pull/17398, and `podman-4.4.1` will include the fixed patch.

ravanelli (dismissed):
LGTM

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -290,7 +290,15 @@ case "${1:-}" in
             read dev < "${zram_dev}"
             umount "${saved_data}"
             rm -rf "${saved_data}" "${partstate_root}"
-            echo "${dev}" > /sys/class/zram-control/hot_remove
+            # Workaround for https://github.com/openshift/os/issues/1149.
+            # Get retry time is 2 when do testing.
+            for x in {0..10}; do
+                if ! echo "${dev}" > /sys/class/zram-control/hot_remove 2>/dev/null; then
+                    sleep 0.01
```
dustymabe: this sleep seems aggressive. The value is so low it's basically like not having any sleep at all and busy looping.

Maybe we should bump the value up to at least a half second or a tenth of a second?

jlebon:
Your hypothesis in https://github.com/openshift/os/issues/1149#issuecomment-1423747513 that it may be the umount somehow needing time to finish sounds plausible. I wonder if udev is involved here and somehow shortly holding/probing the underlying XFS filesystem after it's unmounted. Can you see if instead of a retry loop, adding `udevadm settle || :` after the `umount` helps?

HuijingHei:
> Your hypothesis in [openshift/os#1149 (comment)](https://github.com/openshift/os/issues/1149#issuecomment-1423747513) that it may be the umount somehow needing time to finish sounds plausible. I wonder if udev is involved here and somehow shortly holding/probing the underlying XFS filesystem after it's unmounted. Can you see if instead of a retry loop, adding `udevadm settle || :` after the `umount` helps?

That sounds good to me.
Try adding `udevadm settle || :` after the `umount`, build image and run `luks.tang` and `luks.sss.t1` passed, but `luks.sss.t2` failed, the failed logs:

```
[   84.456275] ignition-ostree-transposefs[2208]: + umount /run/ignition-ostree-transposefs^M
[   84.457080] ignition-ostree-transposefs[2208]: + udevadm settle^M
[   84.465293] systemd[1]: sysroot-var.mount: Deactivated successfully.^M
[   84.465598] systemd[1]: run-ignition\x2dostree\x2dtransposefs.mount: Deactivated successfully.^M
[   84.468209] systemd[1]: ignition-ostree-mount-var.service: Deactivated successfully.^M
[   84.468968] systemd[1]: Stopped Mount OSTree /var.^M
[   84.474285] coreos-teardown-initramfs[2206]: info: taking down network device: enp0s3^M
[   84.475281] coreos-teardown-initramfs[2223]: RTNETLINK answers: Operation not supported^M
[   84.478204] coreos-teardown-initramfs[2206]: info: flushing all routing^M
[   84.480705] ignition-ostree-transposefs[2208]: + rm -rf /run/ignition-ostree-transposefs /run/ignition-ostree-rootfs-partstate.sh^M
[   84.486131] ignition-ostree-transposefs[2208]: + echo 0^M
[   84.486208] ignition-ostree-transposefs[2208]: /usr/libexec/ignition-ostree-transposefs: line 295: echo: write error: Device or resource busy
```

Also try with adding `udevadm trigger --settle "/dev/zram${dev}"` after `umount`, build image and run `luks.tang luks.sss.t1 luks.sss.t2` passed (skip the `podman-system-generator` error). Does this make sense?

```
$ kola run luks.tang luks.sss.t1 luks.sss.t2
=== RUN   luks.tang
=== RUN   luks.sss.t2
=== RUN   luks.sss.t1
--- PASS: luks.tang (206.43s)
--- PASS: luks.sss.t1 (204.48s)
--- PASS: luks.sss.t2 (191.37s)
PASS, output in tmp/kola/qemu-unpriv-2023-02-15-0357-63036
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -289,6 +289,8 @@ case "${1:-}" in
         if [ -d "${saved_data}" ]; then
             read dev < "${zram_dev}"
             umount "${saved_data}"
+            # Wait for umount events to finish
```
jlebon: Maybe something like:

```sh
# After unmounting, make sure zram device state is stable before we remove it.
```

Also, let's add a link back to https://github.com/openshift/os/issues/1149 here.

jlebon (dismissed):
The current workaround seems acceptable. I like it better than the retry loop. :) 

I think the commit message needs to be updated for the latest approach? Otherwise LGTM.

jlebon:
That said, there seems to be a bug here that we could bubble up to the kernel (or systemd/udev). So ideally, we'd file an RHBZ and then reference it in this code. But I'm OK not blocking on that.

I did try to reproduce this locally by doing a tight loop of "create zram device, format it, mount it, unmount it, remove the device" but couldn't reproduce it. So unfortunately filing a bug would probably require detailing how to reproduce this by running our tests, which is harder for other devs to act on.

HuijingHei:
I can not reproduce it locally on ppc64le either, run following scripts in loop 10 times.

```
#!/bin/bash
set -xeuo pipefail

target=/run/testzram/repo/
mem_available=$(grep MemAvailable /proc/meminfo | awk '{print $2}')

modprobe zram num_devices=0
read dev < /sys/class/zram-control/hot_add
# disksize is set arbitrarily large, as zram is capped by mem_limit
echo 10G > /sys/block/zram"${dev}"/disksize
echo $(( mem_available * 90 / 100 ))K > /sys/block/zram"${dev}"/mem_limit
mkfs.xfs -q /dev/zram"${dev}"
[ -d "${target}" ] || mkdir -p "${target}"
mount -o rw /dev/zram"${dev}" "${target}"

cp -aR /sysroot/ostree/repo/* "${target}"

if [ -d "${target}" ]; then
    dev=0
    umount "${target}"
    rm -rf "${target}"
    echo "${dev}" > /sys/class/zram-control/hot_remove
fi
```

HuijingHei:
> That said, there seems to be a bug here that we could bubble up to the kernel (or systemd/udev). So ideally, we'd file an > RHBZ and then reference it in this code. But I'm OK not blocking on that.

- Using `37.20230215.20.0` which has `kernel 6.1.11-200.fc37`, run luks.* test passed.
- Rebuild image with fedora kernel `6.1.11-200.fc37` (current `kernel-5.14.0-252.el9`), run luks.* test passed.
- Rebuild image with latest kernel `5.14.0-273.el9` on brew, run luks.sss.t* failed, luks.tang passed.

Seems that we have the issue even with the latest kernel, need some patch from upstream, but not easy to find which patch. Not sure how much we sync with upstream code, but can open a bug first.

jlebon:
Nice investigation @HuijingHei! I think we have enough information to file an RHBZ, but let me try to do a rough bisect to make it more useful. Meanwhile, one thing we can do is make this patch conditional on el9 for now. That'll make it safer to remove once/if it gets fixed. You can make a `is_rhcos9()` helper in this file similar to the `is_rhcos8()` helper in `commonlib.sh`.

HuijingHei:
> Meanwhile, one thing we can do is make this patch conditional on el9 for now. That'll make it safer to remove once/if it gets fixed. You can make a `is_rhcos9()` helper in this file similar to the `is_rhcos8()` helper in `commonlib.sh`.

Does it necessary to add condition for `aarch64,ppc64le`, as the issue is not found on `x86_64 or s390`?

HuijingHei:
Create bug https://bugzilla.redhat.com/show_bug.cgi?id=2172058 to track the issue, not sure about the component is correct, feel free to change

jlebon:
Cross-posting from https://bugzilla.redhat.com/show_bug.cgi?id=2172058#c2:

> Testing with more Fedora kernels, I've narrowed this down to v6.1. kernel-6.0.18-300.fc37 fails (the latest v6.0 packaged in Fedora), kernel-6.1.0-0.rc0.20221006git833477fce7a1.4.fc38 passes (the earliest v6.1 packaged in Fedora). Bisecting directly from the kernel source is feasible, but non-trivial. Going over `git log 833477fce7a1 ^v6.0.18`, I don't see anything obvious related to zram or mounts that could be implicated. (If this is something racy, it could also be that the race still theoretically exists but has just become much less likely in v6.1+.)

> Does it necessary to add condition for `aarch64,ppc64le`, as the issue is not found on `x86_64 or s390`?

That's a good idea. I'd also be OK with just `!= x86_64` to keep it simple. Not running it on x86_64 greatly reduces the scope of this workaround.

HuijingHei:
> That's a good idea. I'd also be OK with just `!= x86_64` to keep it simple. Not running it on x86_64 greatly reduces the scope of this workaround.

Nice detail instruction in [comment](https://bugzilla.redhat.com/show_bug.cgi?id=2172058#c4). Updating to exclude `x86_64` is done, thanks!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -289,6 +294,12 @@ case "${1:-}" in
         if [ -d "${saved_data}" ]; then
             read dev < "${zram_dev}"
             umount "${saved_data}"
+            # After unmounting, make sure zram device state is stable before we remove it.
+            # See https://github.com/openshift/os/issues/1149
+            # Should remove when https://bugzilla.redhat.com/show_bug.cgi?id=2172058 is fixed.
+            if [ $(uname -m) != x86_64 ] && is_rhcos9; then
```
jlebon, line 300: I just updated this to use `uname -m` instead, which I think is what we want here and is more commonly used.

jlebon (approved):
LGTM! Thanks a lot for digging into this with me! We didn't get to the bottom of it (yet), but having a bug report and scoping it to only where it happens makes it easier to maintain workarounds like this.

travier:
Nice digging and work here @jlebon @HuijingHei

cgwalters:
I'm not sure this works, we hit this failure still in a RHCOS build that has this patch.
`/usr/libexec/ignition-ostree-transposefs: line 304: echo: write error: Device or resource busy`

I have a suspicion that either the conditional logic we added isn't actually firing, or the udev thing didn't actually fix the problem. The thing is it's a race condition, so testing is inherently going to be a bit unreliable unless one is careful to run many times.

It's not clear to me that all the attempts here to isolate and refine the fix here are *really* worth it just a brute force retry that isn't OS conditional (i.e. the first patch).  I mean, yes - it's good for our team to avoid technical debt.  It's a good instinct.  But on the other hand, there's a GIANT TOWER OF STUFF queued up behind us trying to do the rhel9 switch, and like literally every day we miss that, the tower grows.

jlebon:
I'm fine with a retry loop hack constrained on el9 + !x86_64.

One problem that I see is that we're not really good at keeping track and paying down this type of tech debt. We have a bunch of links to RHBZs a bit everywhere, but it's incredibly hard to find the time to push those forward. So one response is to make the barrier to adding them in the first place higher. If we had a process by which to prioritize and undo this tech debt, that'd help a lot I think.

cgwalters:
In summary what I'm saying is that this PR adds a little bit of debt, but it's in service to us switching to RHEL9 which itself will let us then drop a *lot* of other debt.  I didn't happen to be the one reviewing the PR, but if I did that rationale would have led me to just click merge on the original patch here.

HuijingHei:
Thanks @jlebon and @cgwalters for the discussion, new workaround is https://github.com/coreos/fedora-coreos-config/pull/2272

PR #2229, opened 2023-02-13
denylist: drop multipath.day2

jlebon:
This is fixed now as per [1] and following.

[1]  https://github.com/coreos/fedora-coreos-tracker/issues/1206#issuecomment-1424249539


dustymabe (approved):
LGTM

dustymabe:
skipping CI because it won't test this change (doesn't test on s390x)

PR #2232, opened 2023-02-14
denylist: Snooze ext.config.toolbox until 2023-02-21 for rawhide

travier:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1416
See: https://github.com/containers/toolbox/issues/1233


dustymabe (approved):
LGTM

PR #2233, opened 2023-02-14
[rawhide] overrides: pin glib2-2.74.1-3.fc38 in rawhide

gursewak1997:
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1418

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,6 +9,11 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  glib2:
+    evr: 2.74.1-3.fc38
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1418
+      type: pin
   openssh:
     evr: 9.0p1-9.fc38
     metadata:


manifest-lock.overrides.yaml
```diff
@@ -9,6 +9,11 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  glib2:
+    evr: 2.74.1-3.fc38
+    metadata:
+      reason: https://github.com/coreos/rpm-ostree/issues/4306
```
dustymabe: We don't usually link to subproject issues because we need to track it at the FCOS level (i.e. even if a fix merges upstream we need to wait for a new release and RPM). Let's make a new FCOS issue tracker issue for this.

gursewak1997: Sure. I will let this CI run as I am expecting additional kola test failures. If so, I'll try to incorporate more pins into this PR if required.
Will update the PR with the new issue then: https://github.com/coreos/fedora-coreos-tracker/issues/1418

gursewak1997: Wait for this PR to merge: https://github.com/coreos/fedora-coreos-config/pull/2235
Then will rebase this one.

dustymabe (approved):
LGTM

PR #2234, opened 2023-02-15
denylist: Changing tracker issues for coreos.boot-mirror and coreos-boot-mirror.luks

Shilpi-Das1:
Changed tracker issues for  coreos.boot-mirror and coreos.boot-mirror.luks for ppc64le arch Updated the issues with latest kola runs
https://github.com/coreos/coreos-assembler/issues/3360 (coreos.boot-mirror.luks) https://github.com/coreos/coreos-assembler/issues/3361 (coreos.boot-mirror)
Signed-off-by: Shilpi Das <shdas@redhat.com>

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -6,11 +6,11 @@
 - pattern: podman.workflow
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: coreos.boot-mirror.luks
-  tracker: https://github.com/coreos/coreos-assembler/issues/2725
+  tracker: https://github.com/coreos/coreos-assembler/issues/3360
   arches:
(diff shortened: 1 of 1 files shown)


jlebon (approved):
I think the root cause is very likely the same between these two test failures so I'm not sure having two separate issues to track them is necessary. But since you already have them filed, we might as well roll with it. :)

Re. the commit message, note we usually use the imperative mood. So a suggestion for the commit title is:

> denylist: change tracker issue for coreos.boot-mirror patterns

dustymabe:
Thanks for doing this work @Shilpi-Das1 !

dustymabe (approved):

ravanelli (approved):

PR #2235, opened 2023-02-15
denylist: snooze `ext.config.extensions.module` for now

gursewak1997:
Issue: coreos/fedora-coreos-tracker#1420

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -41,3 +41,8 @@
   snooze: 2023-02-21
   streams:
     - rawhide
+- pattern: ext.config.extensions.module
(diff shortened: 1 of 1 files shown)


kola-denylist.yaml
```diff
@@ -41,3 +41,8 @@
   snooze: 2023-02-21
   streams:
     - rawhide
+- pattern: ext.config.extensions.module
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1420
+  snooze: 2023-02-22
```
jlebon: Let's snooze for longer? E.g. 3 weeks.

jlebon (approved):

PR #2240, opened 2023-02-15
[rawhide] overrides: pin selinux-policy-38.1-1.fc38 and container-selinux-2.193.0-1.fc38

marmijo:
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1412

Note: `container-selinux` needed to be pinned as well because the latest version required the latest version of `selinux-policy`.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,6 +9,11 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  container-selinux:
+    evra: 2:2.193.0-1.fc38.noarch
+    metadata:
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

dustymabe:
ok with this `pin` the `ext.config.ntp.timesyncd.dhcp-propagation` test fails.. which reminds me of https://github.com/coreos/fedora-coreos-config/pull/2210

Long story short it's going to be more work than it's worth to get this all passing tests (i.e. we'd have to also denylist a few other tests). We can just live with the Azure failures for now (they're non-fatal) and try to get this resolved soon with the selinux maintainers.

marmijo:
Sounds good to me, I'll go ahead and close this out and open a BZ ticket for the SELinux maintainers.

PR #2243, opened 2023-02-17
[rawhide] overrides: Remove pin for glib2-2.74.1-3.fc38 in rawhide

gursewak1997:
Related Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1422

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,11 +9,6 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
-  glib2:
-    evr: 2.74.1-3.fc38
-    metadata:
-      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1418
-      type: pin
   openssh:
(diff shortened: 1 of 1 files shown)


dustymabe:
@travier - it appears the toolbox test still isn't passing (even though the container does exist now) - can you take a look?

travier:
Highly likely that polkit has been rebuild / linked against the new glib which we reverted.

travier:
```
kola-runext-podman-tmpfs-context[2025]: /usr/bin/conmon: symbol lookup error: /usr/bin/conmon: undefined symbol: g_string_free_and_steal
```

travier:
Everything rebuilt against the new glib will start failing

travier:
I'm pushing the new rpm-ostree build so that we can update glib again

travier:
Hold on, this is rawhide, we should be able to unpin glib now that rpm-ostree is fixed there.

travier:
So this should unpin glib instead

aaradhak (approved):

PR #2244, opened 2023-02-17
F38 Change: Kola test for Shorter shutdown timer

Adam0Brien:
tests if a systems shutdown timer is 2 mins and 45 seconds

### DIFF
--- a/tests/kola/systemd/test.sh
+++ b/tests/kola/systemd/test.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##
+## Checks if default shutdown time is 2 mins an 45 seconds
+##
+
+# set -xeuo pipefail
+
+# . $KOLA_EXT_DATA/commonlib.sh
+source data/commonlib.sh
+
+
+timeout=$(systemctl show --property=DefaultTimeoutStartUSec | cut -d= -f2)
+
+timeout_seconds=$(echo $timeout | sed -E 's/([0-9]+)min ([0-9]+)s/\1*60+\2/' | bc)
+#echo "Timeout is set to $timeout_seconds seconds"
+
+#Verify that the default shutdown time is 2 minutes and 45 seconds (165 seconds)
+if [ $timeout_seconds -eq 165 ]; then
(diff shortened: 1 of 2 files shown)


tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##
+## Checks if default shutdown time is 2 mins an 45 seconds
+##
```
travier, line 6:
```suggestion
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##
+## Checks if default shutdown time is 2 mins an 45 seconds
```
travier, line 5:
```suggestion
# Checks if default shutdown time is 2 mins an 45 seconds
```

This needs to be a single `#`. We use double `#` for special kola comments, and single `#` for normal comments.

travier: See https://github.com/coreos/fedora-coreos-config#tests-layout

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,24 @@
...
+##
+
+# set -xeuo pipefail
+
+# . $KOLA_EXT_DATA/commonlib.sh
```
travier, line 10: You need to create a folder and a symbolic link for this to work like all other tests. Look for them in other tests folders.

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##
+## Checks if default shutdown time is 2 mins an 45 seconds
+##
+
+# set -xeuo pipefail
```
travier, line 8: We need this one. Your test should pass with this set.

travier (changes requested):

travier:
Moved to https://github.com/coreos/fedora-coreos-config/pull/2247

PR #2246, opened 2023-02-17
[rawhide] manifest: rawhide is now F39

aaradhak:


marmijo (approved):
LGTM

gursewak1997 (approved):

jlebon:
I think this technically requires an updated coreos-installer package with https://github.com/coreos/coreos-installer/pull/1113. Once it's building in the pipeline, we'll be signing artifacts with the f39 key, and then upgrade tests on the second f39 build will fail because it won't be able to verify the previous build's downloaded QEMU image. I'd be OK temporarily disabling upgrade tests if we know a coreos-installer release is coming soon but anyway we don't have an easy way to do that in the pipeline right now for just rawhide. A package rebuild shouldn't be hard though. Will ask.

Add this as a thing we'll want to update the checklist template for at the end of this process. :)

Edit: https://src.fedoraproject.org/rpms/rust-coreos-installer/pull-request/48

jlebon (approved):
Package built and will be in the next cosa rebuild.

PR #2247, opened 2023-02-20
F38 Changes: Kola test for shorter shutdown timer

Adam0Brien:
New PR as the last one was pushing to main

### DIFF
--- a/tests/kola/systemd/data/commonlib.sh
+++ b/tests/kola/systemd/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh
--- a/tests/kola/systemd/default-unit-timeouts
+++ b/tests/kola/systemd/default-unit-timeouts
@@ -1,0 +1,52 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+#
+# Checks if default shutdown time is 1m 30s
+#
+# See: https://github.com/coreos/fedora-coreos-tracker/issues/1404
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec)
+DefaultTimeoutStopUSec=$(systemctl show --value --property=DefaultTimeoutStopUSec)
+DefaultTimeoutAbortUSec=$(systemctl show --value --property=DefaultTimeoutAbortUSec)
+DefaultDeviceTimeoutUSec=$(systemctl show --value --property=DefaultDeviceTimeoutUSec)
+
+error="DefaultTimeoutStartUsec=$DefaultTimeoutStartUSec DefaultTimeoutStopUSec=$DefaultTimeoutStopUSec DefaultTimeoutAbortUSec=$DefaultTimeoutAbortUSec DefaultDeviceTimeoutUSec=$DefaultDeviceTimeoutUSec"
+timeout="1min 30s"
+
+systemd_version=$(rpm -q --queryformat '%{VERSION}' systemd)
+required_version="252.4"
+
+# Verify that the timeouts are 1 minute and 30 seconds (90 seconds)
+
+if [[ "$DefaultTimeoutStartUSec" == "$timeout" ]]; then
+    ok "The default start timeout is 1m 30 seconds."
+else
+    fatal $error
+fi
+
+if [[ $DefaultTimeoutStopUSec == "$timeout" ]]; then
+    ok "The default stop timeout is 1m 30 seconds."
+else
+    fatal $error
+fi
+
+if [[ $DefaultTimeoutAbortUSec == "$timeout" ]]; then
+    ok "The default abort timeout time is 1m 30 seconds."
+else
+    fatal $error
+fi
+
+if [[ "${systemd_version}" == "${required_version}" || "${systemd_version}" > "${required_version}" ]]; then
+    if [[ $DefaultDeviceTimeoutUSec == "$timeout" ]]; then
+        ok "The default device timeout is 1m 30 seconds."
+    else
+        fatal $error
+    fi
+else
+    ok "This variable is not available in the current systemd version"
+fi


tests/kola/data/commonlib.sh
```diff
@@ -1,82 +1 @@
-# This is a common library created for the ok & fatal function and symlinks
```
travier: This is the shared lib file. You should not remove this one.

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##
+
+#Checks if default shutdown time is 2 mins an 45 seconds
```
travier:
```suggestion
#
# Checks if default shutdown time is 2 mins an 45 seconds
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,22 @@
...
+. $KOLA_EXT_DATA/commonlib.sh
+
+timeout=$(systemctl show --property=DefaultTimeoutStartUSec | cut -d= -f2)
+
+timeout_seconds=$(echo $timeout | sed -E 's/([0-9]+)min ([0-9]+)s/\1*60+\2/' | bc)
+#echo "Timeout is set to $timeout_seconds seconds"
+
+#Verify that the default shutdown time is 2 minutes and 45 seconds (165 seconds)
+if [ $timeout_seconds -eq 165 ]; then
+  ok "PASS: The default shutdown time is 2 minutes and 45 seconds."
```
travier:
```suggestion
  ok "The default shutdown time is 2 minutes and 45 seconds."
```

You don't need to add PASS/FAIL, this is done by kola

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,22 @@
...
+else
+  fatal "FAIL: The default shutdown time is not 2 minutes and 45 seconds. This system is $timeout"
```
travier:
```suggestion
  fatal "The default shutdown time is not 2 minutes and 45 seconds. This system is $timeout"
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,22 @@
...
+#Checks if default shutdown time is 2 mins an 45 seconds
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+timeout=$(systemctl show --property=DefaultTimeoutStartUSec | cut -d= -f2)
+
+timeout_seconds=$(echo $timeout | sed -E 's/([0-9]+)min ([0-9]+)s/\1*60+\2/' | bc)
+#echo "Timeout is set to $timeout_seconds seconds"
```
travier:
```suggestion
#echo "Timeout is set to $timeout_seconds seconds"
```

Do not include commented code. Either we need it and it should run, or we don't and thus we should not include it.

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,22 @@
...
+
+#Verify that the default shutdown time is 2 minutes and 45 seconds (165 seconds)
```
travier:
```suggestion
# Verify that the default shutdown time is 2 minutes and 45 seconds (165 seconds)
```

travier (changes requested):

travier:
If I run that on an FCOS system right now I get:
```
$ systemctl show --property=DefaultTimeoutStartUSec
DefaultTimeoutStartUSec=1min 30s
```

travier:
You should also check all values, not just one.

travier:
Can you also add a link to the issue you're fixing to the commit message and update the first comment here?

Something like:

```
Fixes: <URL>
```

Adam0Brien:
> 

when you say check all values do you mean these three : 
DefaultTimeoutStartUSec=1min 30s
DefaultTimeoutStopUSec=1min 30s
DefaultTimeoutAbortUSec=1min 30s

travier:
All the values mentioned in the configuration that we are changing in https://src.fedoraproject.org/rpms/fedora-release/pull-request/249 from https://github.com/coreos/fedora-coreos-tracker/issues/1404

travier:
So that two times 3 values as we are changing user & system values.

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,30 @@
...
+
+DefaultDeviceTimeoutSec=$(systemctl show --property=DefaultDeviceTimeoutSec | cut -d= -f2)
+
+timeout_seconds_start=$(echo $DefaultTimeoutStartUSec | sed -E 's/([0-9]+)min ([0-9]+)s/\1*60+\2/' | bc)
+timeout_seconds_stop=$(echo $DefaultTimeoutStopUSec | sed -E 's/([0-9]+)min ([0-9]+)s/\1*60+\2/' | bc)
+timeout_seconds_abort=$(echo $DefaultTimeoutAbortUSec | sed -E 's/([0-9]+)min ([0-9]+)s/\1*60+\2/' | bc)
+timeout_seconds_default=$(echo $DefaultDeviceTimeoutSec | sed -E 's/([0-9]+)min ([0-9]+)s/\1*60+\2/' | bc)
+
+# Verify that the default shutdown time is 2 minutes and 45 seconds (165 seconds)
+if [[ $timeout_seconds_start && $timeout_seconds_stop && $timeout_seconds_abort && $DefaultDeviceTimeoutSec -eq 165 ]]; then
```
travier: I don't think that this checks what you want. This will only check the last value. The other ones are true because they are non zero.

Adam0Brien: Will i make a separate if statement for each value or is there a better solution?

travier: You need an if statement for each value

Adam0Brien: > You need an if statement for each value

changes pushed!

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+#
+# Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1404
+#
+# Checks if default shutdown time is 2 mins an 45 seconds
```
travier:
```suggestion
# Checks if default shutdown time is 1m 90s
# See: https://github.com/coreos/fedora-coreos-tracker/issues/1404
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,38 @@
...
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+DefaultTimeoutStartUSec=$(systemctl show --property=DefaultTimeoutStartUSec | cut -d= -f2)
```
travier:
```suggestion
DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec)
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,38 @@
...
+DefaultTimeoutAbortUSec=$(systemctl show --property=DefaultTimeoutAbortUSec | cut -d= -f2)
+DefaultDeviceTimeoutUSec=$(systemctl show --property=DefaultDeviceTimeoutUSec | cut -d= -f2)
+
+timeout_seconds_start=$(echo $DefaultTimeoutStartUSec | awk '{split($0,a,"min "); printf "%.0f%s\n", a[1]*60+a[2], ""}')
+timeout_seconds_stop=$(echo $DefaultTimeoutStopUSec | awk '{split($0,a,"min "); printf "%.0f%s\n", a[1]*60+a[2], ""}')
+timeout_seconds_abort=$(echo $DefaultTimeoutAbortUSec | awk '{split($0,a,"min "); printf "%.0f%s\n", a[1]*60+a[2], ""}')
+timeout_seconds_default=$(echo $DefaultDeviceTimeoutUSec | awk '{split($0,a,"min "); printf "%.0f%s\n", a[1]*60+a[2], ""}')
+
+# Verify that the DefaultTimeoutStartUSec time is 2 minutes and 45 seconds (165 seconds)
+if [[ $timeout_seconds_start -eq 165 ]]; then
```
travier: You should probably avoid doing complex computation and match strings directly instead.

It looks like the values computed above are not correct.

tests/kola/systemd/test.sh
travier: If I run the following command on a stable Fedora CoreOS, I get:

```
$ systemctl show --property=DefaultTimeoutStopUSec
DefaultTimeoutStopUSec=1min 30s
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,38 @@
...
+  ok "The default start timeout is 2 minutes and 45 seconds. "
+elif [[ $timeout_seconds_stop -eq 165 ]]; then
+  ok "The default stop timeout is 2 minutes and 45 seconds."
+elif [[ $timeout_seconds_abort -eq 165 ]]; then
+  ok "The default abort timeout time is 2 minutes and 45 seconds."
+elif [[ $timeout_seconds_default -eq 165 ]]; then
+  ok "The default device timeout is 2 minutes and 45 seconds."
+else
+  fatal "
```
travier: This likely needs to be on a single line.

travier (changes requested):

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Checks if default shutdown time is 1m 30s
+#
+# See: https://github.com/coreos/fedora-coreos-tracker/issues/1404
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec) #$timeout
```
travier:
```suggestion
DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec)
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,46 @@
...
+DefaultTimeoutStopUSec=$(systemctl show --value --property=DefaultTimeoutStopUSec)
+DefaultTimeoutAbortUSec=$(systemctl show --value --property=DefaultTimeoutAbortUSec)
+DefaultDeviceTimeoutUSec=$(systemctl show --value --property=DefaultDeviceTimeoutUSec) 
+
+error="DefaultTimeoutStartUsec=$DefaultTimeoutStartUSec\nDefaultTimeoutStopUSec=$DefaultTimeoutStopUSec\nDefaultTimeoutAbortUSec=$DefaultTimeoutAbortUSec\nDefaultDeviceTimeoutUSec=$DefaultDeviceTimeoutUSec"
+
+timeout="1min 30s"
+
+# Verify that the timeouts are 1 minute and 90 seconds (150 seconds)
```
travier:
```suggestion
# Verify that the timeouts are 1 minute and 30 seconds (90 seconds)
```
?

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [[ "$DefaultTimeoutStartUSec" == "$timeout" ]]; then
+  ok "The default start timeout is 1m 30 seconds."
+else
+  fatal -e $error
```
travier:
```suggestion
  fatal $error
```

This relies on the implementation of the fatal function. Let's just use a space instead.

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,46 @@
...
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec) #$timeout
+DefaultTimeoutStopUSec=$(systemctl show --value --property=DefaultTimeoutStopUSec)
+DefaultTimeoutAbortUSec=$(systemctl show --value --property=DefaultTimeoutAbortUSec)
+DefaultDeviceTimeoutUSec=$(systemctl show --value --property=DefaultDeviceTimeoutUSec) 
+
+error="DefaultTimeoutStartUsec=$DefaultTimeoutStartUSec\nDefaultTimeoutStopUSec=$DefaultTimeoutStopUSec\nDefaultTimeoutAbortUSec=$DefaultTimeoutAbortUSec\nDefaultDeviceTimeoutUSec=$DefaultDeviceTimeoutUSec"
```
travier:
```suggestion
error="DefaultTimeoutStartUsec=$DefaultTimeoutStartUSec DefaultTimeoutStopUSec=$DefaultTimeoutStopUSec DefaultTimeoutAbortUSec=$DefaultTimeoutAbortUSec DefaultDeviceTimeoutUSec=$DefaultDeviceTimeoutUSec"
```

travier (changes requested):

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Checks if default shutdown time is 1m 30s
+#
+# See: https://github.com/coreos/fedora-coreos-tracker/issues/1404
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec)
```
travier:
```suggestion
DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec)
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,46 @@
...
+# See: https://github.com/coreos/fedora-coreos-tracker/issues/1404
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+DefaultTimeoutStartUSec=$(systemctl show --value --property=DefaultTimeoutStartUSec) 
+DefaultTimeoutStopUSec=$(systemctl show --value --property=DefaultTimeoutStopUSec)
+DefaultTimeoutAbortUSec=$(systemctl show --value --property=DefaultTimeoutAbortUSec)
+DefaultDeviceTimeoutUSec=$(systemctl show --value --property=DefaultDeviceTimeoutUSec)
```
travier:
```suggestion
DefaultDeviceTimeoutUSec=$(systemctl show --value --property=DefaultDeviceTimeoutUSec)
```

tests/kola/systemd/test.sh
travier: You should configure your editor to highlight whitespace, especially end of line whitespace, or to trim it automatically.

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+if [[ $DefaultTimeoutStopUSec == "$timeout" ]]; then
+  ok "The default stop timeout is 1m 30 seconds."
+else
+  fatal $error
+fi
+
+if [[ $DefaultTimeoutAbortUSec == "$timeout" ]]; then
+  ok "The default abort timeout time 1m 30 seconds."
```
travier:
```suggestion
  ok "The default abort timeout time is 1m 30 seconds."
```

travier (changes requested):

travier:
Alright, this is now failing on:

```
[2023-02-23T10:57:20.211Z] Feb 23 10:57:18 qemu0 kola-runext-test.sh[1725]: + [[ '' == \1\m\i\n\ \3\0\s ]]
[2023-02-23T10:57:20.211Z] Feb 23 10:57:18 qemu0 kola-runext-test.sh[1725]: + fatal DefaultTimeoutStartUsec=1min 30s DefaultTimeoutStopUSec=1min 30s DefaultTimeoutAbortUSec=1min 30s DefaultDeviceTimeoutUSec=
[2023-02-23T10:57:20.211Z] Feb 23 10:57:18 qemu0 kola-runext-test.sh[1725]: + echo DefaultTimeoutStartUsec=1min 30s DefaultTimeoutStopUSec=1min 30s DefaultTimeoutAbortUSec=1min 30s DefaultDeviceTimeoutUSec=
[2023-02-23T10:57:20.211Z] Feb 23 10:57:18 qemu0 kola-runext-test.sh[1725]: DefaultTimeoutStartUsec=1min 30s DefaultTimeoutStopUSec=1min 30s DefaultTimeoutAbortUSec=1min 30s DefaultDeviceTimeoutUSec=
```

which is expected given that this is only in the systemd version in rawhide now.

What we need to do now is to check the systemd version using the rpm command (look for that logic in other tests here) and check this value only if systemd is recent enough.

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,52 @@
...
+  fatal $error
+fi
+
+if [[ "${systemd_version}" == "${required_version}" || "${systemd_version}" > "${required_version}" ]]; then
+  if [[ $DefaultDeviceTimeoutUSec == "$timeout" ]]; then
+    ok "The default device timeout is 1m 30 seconds."
+  else
+    fatal $error
+  fi
+  else
```
dustymabe: spacing:
```suggestion
else
```

travier:
```suggestion
else
```

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,52 @@
...
+    ok "This variable is not available in the current rpm version"
```
travier:
```suggestion
    ok "This variable is not available in the current systemd version"
```

travier (dismissed):
Two nits but LGTM otherwise

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,52 @@
...
+  fatal $error
+fi
+
+if [[ $DefaultTimeoutAbortUSec == "$timeout" ]]; then
+  ok "The default abort timeout time is 1m 30 seconds."
+else
+  fatal $error
+fi
+
+if [[ "${systemd_version}" == "${required_version}" || "${systemd_version}" > "${required_version}" ]]; then
```
dustymabe: I'm not sure if the comparison will work with the . in the middle. Does it work?

Adam0Brien: I've run the test on rawhide and fcos and they both seem to be passing at the moment. Here are the results i got :

Output 1:

```
`adamobrien@fedora:~/OS-Builds/fcos » cosa kola run ext.config.systemd
+cosa:16> podman run --rm -ti --security-opt 'label=disable' --privileged '--uidmap=1000:0:1' '--uidmap=0:1:1000' --uidmap 1001:1001:64536 -v /home/adamobrien/OS-Builds/fcos:/srv/ --device /dev/kvm --device /dev/fuse --tmpfs /tmp -v /var/tmp:/var/tmp quay.io/coreos-assembler/coreos-assembler:latest kola run ext.config.systemd
kola -p qemu-unpriv --output-dir tmp/kola run ext.config.systemd
⚠️  Skipping kola test pattern "fcos.internet":
  👉 https://github.com/coreos/coreos-assembler/pull/1478
⚠️  Skipping kola test pattern "podman.workflow":
  👉 https://github.com/coreos/coreos-assembler/pull/1478
🕒 Snoozing kola test pattern "ext.config.platforms.aws.nvme" until Mar 10 2023:
  👉 https://github.com/coreos/fedora-coreos-tracker/issues/1306#issuecomment-1426106534
=== RUN   non-exclusive-test-bucket-0/ext.config.systemd
--- PASS: non-exclusive-test-bucket-0 (27.15s)
    --- PASS: non-exclusive-test-bucket-0/ext.config.systemd (1.52s)
PASS, output in tmp/kola
+cosa:24> rc=0 
+cosa:24> set +x`
```

Output 2:

```
`adamobrien@fedora:~/OS-Builds/fcos-rawhide » cosa kola run ext.config.systemd          
+cosa:16> podman run --rm -ti --security-opt 'label=disable' --privileged '--uidmap=1000:0:1' '--uidmap=0:1:1000' --uidmap 1001:1001:64536 -v /home/adamobrien/OS-Builds/fcos-rawhide:/srv/ --device /dev/kvm --device /dev/fuse --tmpfs /tmp -v /var/tmp:/var/tmp quay.io/coreos-assembler/coreos-assembler:latest kola run ext.config.systemd
kola -p qemu-unpriv --output-dir tmp/kola run ext.config.systemd
⚠️  Skipping kola test pattern "fcos.internet":
  👉 https://github.com/coreos/coreos-assembler/pull/1478
⚠️  Skipping kola test pattern "podman.workflow":
  👉 https://github.com/coreos/coreos-assembler/pull/1478
🕒 Snoozing kola test pattern "ext.config.platforms.aws.nvme" until Mar 10 2023:
  👉 https://github.com/coreos/fedora-coreos-tracker/issues/1306#issuecomment-1426106534
=== RUN   non-exclusive-test-bucket-0/ext.config.systemd
--- PASS: non-exclusive-test-bucket-0 (27.33s)
    --- PASS: non-exclusive-test-bucket-0/ext.config.systemd (1.58s)
PASS, output in tmp/kola
+cosa:24> rc=0 
+cosa:24> set +x`
```

dustymabe: I think if you do some more testing you'll find that while it may work in this limited set of use cases it would be better if we made something a little more comprehensive here. I propose we add some new functions to the common library and then use them here. See what you think about https://github.com/coreos/fedora-coreos-config/pull/2280

tests/kola/systemd/test.sh
```diff
@@ -0,0 +1,52 @@
...
+error="DefaultTimeoutStartUsec=$DefaultTimeoutStartUSec DefaultTimeoutStopUSec=$DefaultTimeoutStopUSec DefaultTimeoutAbortUSec=$DefaultTimeoutAbortUSec DefaultDeviceTimeoutUSec=$DefaultDeviceTimeoutUSec"
+timeout="1min 30s"
+
+systemd_version=$(rpm -q --queryformat '%{VERSION}' systemd)
+required_version="252.4"
+
+# Verify that the timeouts are 1 minute and 30 seconds (90 seconds)
+
+if [[ "$DefaultTimeoutStartUSec" == "$timeout" ]]; then
+  ok "The default start timeout is 1m 30 seconds."
```
dustymabe: I think in our bash files we usually use 4 spaces for indentation. Can you update any current 2 spaced indentation to be 4?

dustymabe:
eventually we'll need to squash all the commits down into a single commit.

dustymabe:
mind again squashing down into a single commit?

Adam0Brien:
> mind again squashing down into a single commit?

will do!

dustymabe (approved):
LGTM - I'll address https://github.com/coreos/fedora-coreos-config/pull/2247#discussion_r1127349269 in https://github.com/coreos/fedora-coreos-config/pull/2280

cgwalters:
It's a problem for us to write tests like this that explicitly hardcode something about the operating system.  Today, when tests fail we *stop shipping operating system updates* by default.  In my opinion, this test should be at best informative.  (We don't have a good way to express that today)

However, I think there's a simpler fix: Instead of hardcoding expecting to find 1m 30s, let's hardcode that we expect a timeout that is *at least* that value.  Then we don't need to do anything with respect to the systemd or operating sytsem version.

cgwalters:
(The above comment is motivated by this test failing in the midst of us trying to ship RHEL CoreOS 9.2)

dustymabe:
> It's a problem for us to write tests like this that explicitly hardcode something about the operating system. Today, when tests fail we _stop shipping operating system updates_ by default. In my opinion, this test should be at best informative. (We don't have a good way to express that today)

The config bumps are gated by CI. When this config bump landed it should have failed in CI so IMO "no, this shouldn't have stopped us shipping operating system updates".

> 
> However, I think there's a simpler fix: Instead of hardcoding expecting to find 1m 30s, let's hardcode that we expect a timeout that is _at least_ that value. 

That's a bit hard here since we're dealing with strings.

> Then we don't need to do anything with respect to the systemd or operating system version.

The only reason we are checking systemd version is because `DefaultDeviceTimeoutUSec` doesn't exist in systemd prior to `252.4`, which isn't going to be solved by checking anything greater than or equal to `1m 30s`.

cgwalters:
> The config bumps are gated by CI. When this config bump landed it should have failed in CI so IMO "no, this shouldn't have stopped us shipping operating system updates".

Yes, but this is not true for RHCOS today.

> That's a bit hard here since we're dealing with strings.

Yeah...external tests are nice, bash is not.

> The only reason we are checking systemd version is because DefaultDeviceTimeoutUSec doesn't exist in systemd prior to 252.4, which isn't going to be solved by checking anything greater than or equal to 1m 30s.

I think the test can skip checking the property if it is empty.

dustymabe:
> > The config bumps are gated by CI. When this config bump landed it should have failed in CI so IMO "no, this shouldn't have stopped us shipping operating system updates".
> 
> Yes, but this is not true for RHCOS today.

If that's not true then this is the thing we should be complaining about.

> > The only reason we are checking systemd version is because DefaultDeviceTimeoutUSec doesn't exist in systemd prior to 252.4, which isn't going to be solved by checking anything greater than or equal to 1m 30s.
> 
> I think the test can skip checking the property if it is empty.

Looking at the test failure:

```
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2114]: ++ rpm -q --queryformat '%{VERSION}' systemd
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + systemd_version=239
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + required_version=252.4
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + [[ 1min 30s == \1\m\i\n\ \3\0\s ]]
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + ok 'The default start timeout is 1m 30 seconds.'
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + echo ok 'The default start timeout is 1m 30 seconds.'
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: ok The default start timeout is 1m 30 seconds.
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + [[ 1min 30s == \1\m\i\n\ \3\0\s ]]
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + ok 'The default stop timeout is 1m 30 seconds.'
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + echo ok 'The default stop timeout is 1m 30 seconds.'
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: ok The default stop timeout is 1m 30 seconds.
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + [[ '' == \1\m\i\n\ \3\0\s ]]
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + fatal DefaultTimeoutStartUsec=1min 30s DefaultTimeoutStopUSec=1min 30s DefaultTimeoutAbortUSec= DefaultDeviceTimeoutUSec=
Mar 16 13:06:30 qemu0 kola-runext-default-unit-timeouts[2108]: + echo DefaultTimeoutStartUsec=1min 30s DefaultTimeoutStopUSec=1min 30s DefaultTimeoutAbortUSec= DefaultDeviceTimeoutUSec=
```

It's apparent that `DefaultTimeoutAbortUSec` (in addition to `DefaultDeviceTimeoutUSec`) doesn't exist in the systemd in RHEL/RHCOS.

dustymabe:
> I think the test can skip checking the property if it is empty.

That seems like a good idea, until systemd upstream changes the property name and then we have a test that's not checking for the right value.

cgwalters:
> If that's not true then this is the thing we should be complaining about.

Complaining to...?  We own this whole thing and need to keep in mind both c9s and rhel9 too.

> until systemd upstream changes the property name and then we have a test that's not checking for the right value.

That seems very unlikely to me.  Changing the property name would break anyone who was trying to configure it.

travier:
fedora-coreos-config bumps are gated in Prow CI in openshift/os.

This very specific one got through because I disabled 8.6 CI too early in https://github.com/openshift/release/pull/35789 as we decided to keep the 8.6 branch alive a bit longer.

This is only impacting RHEL 8.6 builds so it should not block 9.2 builds.

travier:
We'll skip it in https://github.com/openshift/os/pull/1205

cgwalters:
> fedora-coreos-config bumps are gated in Prow CI in openshift/os.

Yeah, but crucially *not* on package changes.

> This is only impacting RHEL 8.6 builds so it should not block 9.2 builds.

But yes, I think I got build logs crossed here, so it didn't impact 9.2 work, so indeed not too important.

(But the larger point still stands I think)

PR #2252, opened 2023-02-20
denylist: snooze tests that use Butane 1.5.0-experimental

bgilbert:
We're ratcheting in a spec stabilization.  The Ignition repo has a workaround for ratcheting Ignition configs using the experimental spec, but we've started using Butane configs in tests, and that requires a double ratchet.  For now, just snooze the affected tests.

cc @prestist


dustymabe (approved):
LGTM

PR #2253, opened 2023-02-21
denylist: snooze files.file-directory-permissions test

dustymabe:
This is failing because the 98-default-mac-none.link  from systemd-udev has group write permissions. See
https://github.com/coreos/fedora-coreos-tracker/issues/1427


aaradhak:
LGTM

aaradhak (approved):

dustymabe:
skipping Ci on this since it won't test this change anyway.

PR #2254, opened 2023-02-21
[next-devel] Move to Fedora Linux 38

aaradhak:
Switching next-devel release version to F38


jlebon:
This step (and following steps) in the checklist need to be done when Fedora 38 Beta is released. So we can stop here for now. We'll be tracking f38 content via the `branched` stream until Beta.

dustymabe:
Pushed an update here.

jlebon (approved):
LGTM. Went over the package changes (https://github.com/coreos/fedora-coreos-tracker/issues/1435). Nothing alarming.

PR #2255, opened 2023-02-21
workflow: don't remove graduated overrides on next-devel when disabled

jlebon:
The `next-devel` stream uses the -next repos which recently became invalid since it still assumes f37. The graduation workflow has been failing since then and I've been receiving notifications each time it runs.

Since `next-devel` is unmaintained when it's disabled, it doesn't make sense to run this workflow there (e.g. there's currently a stale NetworkManager pin there).

Rework the job so that the matrix is built dynamically, and we only include `next-devel` in the matrix when it's enabled.

### DIFF
--- a/.github/workflows/remove-graduated-overrides.yml
+++ b/.github/workflows/remove-graduated-overrides.yml
@@ -9,19 +9,40 @@ permissions:
   contents: read
 
 jobs:
+  buildmatrix:
+    name: "Build job matrix"
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


jlebon:
(Killed CoreOS CI for this since it doesn't cover this.)

I didn't test this in a fork; was thinking of just triggering it manually and iterating with any fixups as needed. That said, the "dynamic matrix" and "query next-devel" bits come from existing GitHub Action workflows.

dustymabe (approved):
LGTM

jlebon:
Seems to be working well!

PR #2257, opened 2023-02-22
denylist: snooze another test that uses Butane 1.5.0-experimental

bgilbert:
This one only runs on AWS, so we missed it before.

Fixes #2252.


dustymabe (approved):
LGTM

dustymabe:
skipping CI here since this change won't be tested.

PR #2261, opened 2023-02-23
denylist: deny kdump.crash test on aarch64 for f38 streams

dustymabe:
It's failing with the latest kernel updates. See
https://github.com/coreos/fedora-coreos-tracker/issues/1430


jlebon (approved):

dustymabe:

PR #2262, opened 2023-02-23
overlay.d/35coreos-live: pull in coreos-enable-network.service on live boots

### LINKED ISSUE #1423: No network when PXE provisioning on bare metal since 37.20230122.3.0
### Describe the bug

We are using https://typhoon.psdn.io/ to provision Fredora CoreOS on bare metal. Since Fedora CoreOS **37.20230122.3.0** no network is available

- Using DNS of the matchbox server

```
coreos-livepxe-rootfs: curl: (6) Could not resolve host: xxx
coreos-livepxe-rootfs: Couldn't establish connectivity with the server specified by:
coreos-livepxe-rootfs: coreos.live.rootfs_url=http://xxx
```

- Using IP of the matchbox server

```
coreos-livepxe-rootfs: curl: (7) Couldn't connect to server
coreos-livepxe-rootfs: Couldn't establish connectivity with the server specified by:
coreos-livepxe-rootfs: coreos.live.rootfs_url=http://xxx
```

[coreos-enable-network.service](https://github.com/coreos/fedora-coreos-config/blob/stable/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.service)

It looks like it is related to #1383.

Is it intended behavior?

Setting `rd.neednet=1` explicitly  fixes it for us.

CC: @dghubble

### Reproduction steps

1. Provision Fedora CoreOS 37.20230122.3.0 on bare metal

### Expected behavior

Network is available.

### Actual behavior

Network is not available.

### System details

- Bare Metal
- Fedora CoreOS 37.20230122.3.0

### Butane or Ignition config

_No response_

### Additional information

_No response_

dustymabe:
After 937d329 coreos-enable-network.service is load bearing for coreos-livepxe-rootfs.service if ignition.firstboot isn't provided on the kernel command line. Let's pull it back in for live boots.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1423

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
@@ -30,6 +30,9 @@ fi
 add_requires sysroot.mount     initrd-root-fs.target
 add_requires sysroot-etc.mount initrd-root-fs.target
 add_requires sysroot-var.mount initrd-root-fs.target
+# make sure we enable network if required for coreos-livepxe-rootfs
+# https://github.com/coreos/fedora-coreos-tracker/issues/1423
+add_requires coreos-enable-network.service initrd-root-fs.target
 
 mkdir -p "${UNIT_DIR}/ostree-prepare-root.service.d"
 cat > "${UNIT_DIR}/ostree-prepare-root.service.d/10-live.conf" <<EOF


jlebon (approved):
LGTM. I was trying to see if there was a cleaner way to do this using systemd builtins. But I think the full boolean logic we need ("on one of these platforms and Ignition enabled or on a live boot and `coreos.live.rootfs_url` is specified") isn't easily achievable via unit conditionals.

Did you sanity-check this works as expected on a PXE boot without Ignition enabled?

dustymabe:
I tested with the following patch to `kola`:

```patch
diff --git a/mantle/platform/metal.go b/mantle/platform/metal.go
index 69d53fb6c..fed10695e 100644
--- a/mantle/platform/metal.go
+++ b/mantle/platform/metal.go
@@ -42,7 +42,7 @@ const (
 )
 
 // TODO derive this from docs, or perhaps include kargs in cosa metadata?
-var baseKargs = []string{"rd.neednet=1", "ip=dhcp", "ignition.firstboot", "ignition.platform.id=metal"}
+var baseKargs = []string{"ignition.platform.id=metal"}
 
 var (
        // TODO expose this as an API that can be used by cosa too
```

The pxe test failed but I was able to test with and without this change and see (by watching the console log) that without it we got stuck in the initramfs and with it we were able to complete the install.

PR #2265, opened 2023-02-28
tests: identify and mark non-exclusive tests 

Adam0Brien:
Added Exclusive false to kola tests to make it easier to distinguish what tests can run together in a single vm.
 
 https://github.com/coreos/fedora-coreos-config/issues/1229

### DIFF
--- a/tests/kola/boot/grub2-install
+++ b/tests/kola/boot/grub2-install
@@ -34,7 +34,7 @@ case ${arch} in
     device=$(realpath /dev/disk/by-partlabel/PowerPC-PReP-boot)
     # sync grub2-install parameter according to image build script
     # https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh
-    grub_install_args="--no-nvram"   
+    grub_install_args="--no-nvram"
     ;;
 
   *)
--- a/tests/kola/clhm/ignition-warnings/test.sh
+++ b/tests/kola/clhm/ignition-warnings/test.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 ## kola:
+##   exclusive false
 ##   tags: "platform-independent"
 ##   # We intentionally exclude the Install section from a systemd unit.
 ##   # This is valid but not ideal, so Butane warns about it.
--- a/tests/kola/clhm/ignition-warnings/test.sh
+++ b/tests/kola/clhm/ignition-warnings/test.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 ## kola:
-##   exclusive false
 ##   tags: "platform-independent"
 ##   # We intentionally exclude the Install section from a systemd unit.
 ##   # This is valid but not ideal, so Butane warns about it.
--- a/tests/kola/gshadow
+++ b/tests/kola/gshadow
@@ -20,3 +20,4 @@ echo 'grp1:*::somebody.a1,somebody.a2,somebody.a3,somebody.a4,somebody.a5,somebo
 echo 'grp2:*::root' >> /etc/gshadow
 systemd-sysusers
 echo "ok sysusers gshadow"
+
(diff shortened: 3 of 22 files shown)


tests/kola/ignition/remote/test.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 ## kola:
 ##   tags: needs-internet
+## exclusive false
```
bgilbert: Invalid syntax.

tests/kola/clhm/ignition-warnings/test.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 ## kola:
+##   exclusive false
```
bgilbert: Invalid syntax.

bgilbert (commented):
There are a number of tests here that reboot the machine or reconfigure it in some way.  We'd need to decide how aggressive we want to be in assuming that those tests won't interfere with each other, or with a test that's added later.

tests/kola/boot/grub2-install
```diff
@@ -34,7 +34,7 @@ case ${arch} in
     device=$(realpath /dev/disk/by-partlabel/PowerPC-PReP-boot)
     # sync grub2-install parameter according to image build script
     # https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh
-    grub_install_args="--no-nvram"   
+    grub_install_args="--no-nvram"
```
bgilbert, line 37: It's not a big deal here, but normally we'd put formatting changes in a separate commit.

dustymabe:
> There are a number of tests here that reboot the machine or reconfigure it in some way. We'd need to decide how aggressive we want to be in assuming that those tests won't interfere with each other, or with a test that's added later.

I think probably a safe rule of thumb here is that if a test has a butane config (or Ignition config) then we should probably assume it's exclusive and not mark it `exclusive: false`.

Maybe that means that there are no remaining tests to mark as exclusive? I'm not sure

tests/kola/gshadow
```diff
@@ -20,3 +20,4 @@ echo 'grp1:*::somebody.a1,somebody.a2,somebody.a3,somebody.a4,somebody.a5,somebo
 echo 'grp2:*::root' >> /etc/gshadow
 systemd-sysusers
 echo "ok sysusers gshadow"
+
```
dustymabe: this isn't a meaningful change.. delete?

bgilbert:
> I think probably a safe rule of thumb here is that if a test has a butane config (or Ignition config) then we should probably assume it's exclusive and not mark it `exclusive: false`.

Also, if there are reboots, the test may be changing the system, or at least is distinguishing between first and second boots.  Multiple tests can't do that, either in series or in parallel.

tests/kola/boot/grub2-install
Adam0Brien: sorry about that, i have vs code set up to auto trim white space..

tests/kola/gshadow
bgilbert: This originally added a newline and is now removing one.  On Unix systems, text files traditionally end with a single newline.

bgilbert:
All of the remaining tests specify test-specific kernel arguments, and some also configure additional network devices.  I think we should probably just close this.

dustymabe:
> I think we should probably just close this.

Yeah probably. Sorry Adam for the goose chase. At least we can probably close https://github.com/coreos/fedora-coreos-config/issues/1229 now.

tests/kola/gshadow
Adam0Brien: > > I think we should probably just close this.
> 
> Yeah probably. Sorry Adam for the goose chase. At least we can probably close #1229 now.

No problem at all, still learned a lot about kola tests :+1:

Adam0Brien:
> > I think we should probably just close this.
> 
> Yeah probably. Sorry Adam for the goose chase. At least we can probably close #1229 now.

No problem still learned alot about kola tests :+1:

PR #2267, opened 2023-02-28
tests/kola/selinux: Add kola SELinux test verifying permissive mode

c4rt0:
See: https://issues.redhat.com/browse/COS-2080


c4rt0:
hey @jlebon - I created this PR following dropping the `Go` test from [the coreos-assembler issue](https://github.com/coreos/coreos-assembler/issues/3369) discussed with @dustymabe

jlebon:
I think the idea with https://github.com/coreos/coreos-assembler/issues/3369 and the discussions in https://github.com/coreos/coreos-assembler/pull/3373 is that we would enhance the existing `ext.config.selinux.enforcing` test instead to cover the bits missing. The flow would roughly be something like this:

```sh
check that we're enforcing by default
setenforce 0
check that we're permissive
setenforce 1
check that we're back to enforcing
sed /etc/selinux/config to persist permissive mode
reboot
check that we're permissive
```

Rebooting in these tests is done a bit differently. You can grep for `autopkgtest-reboot` in this repo to see other tests that reboot and copy that structure.

PR #2268, opened 2023-03-01
denylist: extend snooze on Butane 1.5.0-experimental tests

bgilbert:
We'll need a bit more time to stabilize the Butane spec.


dustymabe (approved):
LGTM

PR #2269, opened 2023-03-01
networking-tools: Add nmstate

qinqon:
As [discussed](https://github.com/coreos/fedora-coreos-tracker/issues/1175) the nmstatectl tool is going to be added to FCOS, it will allow to declare the network configuration at boot time

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -32,3 +32,6 @@ enable rtas_errd.service
 enable clevis-luks-askpass.path
 # Provide status information about the Ignition run
 enable coreos-ignition-write-issues.service
+# Apply network configuration from /etc/nmstate/*.yml
+# We can drop this after merging https://src.fedoraproject.org/rpms/fedora-release/pull-request/255
+enable nmstate.service
--- a/tests/kola/networking/nmstate/data/nmstate-common.sh
+++ b/tests/kola/networking/nmstate/data/nmstate-common.sh
@@ -1,0 +1,23 @@
+. $KOLA_EXT_DATA/commonlib.sh
+
+main() {
+    nmstatectl show
+    journalctl -u nmstate
+    
+    local prefix="first boot"
+    if [ "${AUTOPKGTEST_REBOOT_MARK:-}" == rebooted ]; then
+        prefix="second boot"
+    fi
+    
+    if ! nmcli c show br-ex; then
+        fatal "${prefix}: bridge not configured"
+    fi
+
+    if ! ls /etc/nmstate/*applied; then
+        fatal "${prefix}: nmstate yamls files not marked as applied"    
+    fi
+
+    if [ "${AUTOPKGTEST_REBOOT_MARK:-}" == "" ]; then
+        /tmp/autopkgtest-reboot rebooted
+    fi
+}
--- a/tests/kola/networking/nmstate/policy/test.sh
+++ b/tests/kola/networking/nmstate/policy/test.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+## kola:
+##   tags: "platform-independent needs-internet"
+#
+# Configure a DHCP linux bridge using butane and nmstate service with policy
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/nmstate-common.sh
+
+main
--- a/tests/kola/networking/nmstate/state/config.bu
+++ b/tests/kola/networking/nmstate/state/config.bu
@@ -1,0 +1,17 @@
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/nmstate/br-ex.yml
+      contents:
+        inline: |
+          interfaces:
+           - name: br-ex
+             type: linux-bridge
+             state: up
+             ipv4:
+               enabled: false
+             ipv6:
+               enabled: false
+             bridge:
+               port: []
--- a/tests/kola/networking/nmstate/state/test.sh
+++ b/tests/kola/networking/nmstate/state/test.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+## kola:
+##   tags: "platform-independent needs-internet"
+#
+# Configure a DHCP linux bridge using butane and nmstate service with state
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/nmstate-common.sh
+
+main
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -34,3 +34,8 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1441
       type: pin
+  nmstate:
+    evr: 2.2.8-2.fc37
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-59543876eb
+      type: fast-track
--- a/manifests/networking-tools.yaml
+++ b/manifests/networking-tools.yaml
@@ -19,3 +19,6 @@ packages:
   - iptables nftables
   # Interactive network tools for admins
   - socat net-tools bind-utils
+  # Declarative network configuration
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1175
+  - nmstate
--- a/tests/kola/networking/nmstate/policy/config.bu
+++ b/tests/kola/networking/nmstate/policy/config.bu
@@ -1,0 +1,24 @@
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/nmstate/br-ex-policy.yml
+      contents:
+        inline: |
+          capture:
+            default-gw-route: routes.running.destination=="0.0.0.0/0"
+            default-gw-iface: interfaces.name==capture.default-gw-route.routes.running.0.next-hop-interface
+          desiredState:
+            interfaces:
+            - name: "{{ capture.default-gw-iface.interfaces.0.name }}"
+              type: ethernet
+              state: up
+            - name: br-ex
+              type: linux-bridge
+              copy-mac-from: "{{ capture.default-gw-iface.interfaces.0.name }}"
+              state: up
+              ipv4: "{{ capture.default-gw-iface.interfaces.0.ipv4 }}"
+              ipv6: "{{ capture.default-gw-iface.interfaces.0.ipv6 }}"
+              bridge:
+                port:
+                - name: "{{ capture.default-gw-iface.interfaces.0.name }}"
--- a/tests/kola/networking/nmstate/policy/data/commonlib.sh
+++ b/tests/kola/networking/nmstate/policy/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../../data/commonlib.sh
--- a/tests/kola/networking/nmstate/policy/data/nmstate-common.sh
+++ b/tests/kola/networking/nmstate/policy/data/nmstate-common.sh
@@ -1,0 +1,1 @@
+../../data/nmstate-common.sh
--- a/tests/kola/networking/nmstate/state/data/commonlib.sh
+++ b/tests/kola/networking/nmstate/state/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../../data/commonlib.sh
--- a/tests/kola/networking/nmstate/state/data/nmstate-common.sh
+++ b/tests/kola/networking/nmstate/state/data/nmstate-common.sh
@@ -1,0 +1,1 @@
+../../data/nmstate-common.sh


bgilbert:
I believe you'll need to manually add `nmstate` to the `manifest-lock` files.

It'd also be good to add a [kola test](https://coreos.github.io/coreos-assembler/kola/external-tests/) that configures the network by writing an nmstate config from a Butane config.

cathay4t:
@qinqon Please also include `nmstate-libs` in case someone want a C library instead of CLI.

bgilbert:
If `nmstate-libs` isn't needed by `nmstate`, we probably shouldn't include the library package.  We don't encourage users to install custom software in the host, and so we don't ship libraries that aren't needed by an OS component.

qinqon:
@cathay4t @bgilbert 
This would be needed if one of the os tools use nmstate programatically linking dynamically to it right ?

For static linking the nmstate-lib should be at the dev/build env but not af the target system.

qinqon:
Looks like `nmstatectl service` do not support nmpolicy 

```
Mar 02 13:59:48 qemu0 kola-runext-test.sh[1762]: [2023-03-02T13:59:48Z ERROR nmstatectl::service] Failed to apply state file /etc/nmstate/br-ex-policy.yml: serde_yaml::Error: Unsupported keys found: ["capture", "desiredState"]
```

nmstatectl apply is fine

bgilbert:
> This would be needed if one of the os tools use nmstate programatically linking dynamically to it right ?

In that case, presumably the RPM of the tool would have `Requires: nmstate-libs`, right?  So we shouldn't need to explicitly add it here.

tests/kola/networking/nmstate/config.bu
```diff
@@ -0,0 +1,32 @@
+variant: fcos                                                                   
+version: 1.4.0                                                                  
+storage:                                                                        
+  files:                                                                        
+    - path: /etc/nmstate/br-ex-policy.yaml                                      
+      overwrite: true
```
bgilbert: nit: This line can be dropped since the file won't already exist.

tests/kola/networking/nmstate/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+# Configure an ovs-bridge using nmstate and butane
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+
+if ! nmstatectl apply /etc/nmstate/br-ex-policy.yaml; then
```
bgilbert: Is this needed?  I thought the idea was that policy files in `/etc/nmstate` would be automatically applied by a systemd service provided by nmstate.

tests/kola/networking/nmstate/test.sh
```diff
@@ -0,0 +1,21 @@
...
+    fatal "cannot create bridge"
+fi
+
+nmstatectl show > /tmp/state.yaml
+
+if ! grep -q br-ex /tmp/state.yaml ; then
```
bgilbert: It's okay to check that nmstate thinks it applied the configuration, but it might also be good to look at the actual state of the system to confirm that the bridge was created.

(Minor cleanup: could this be a pipeline with `nmstatectl show` and skip the tempfile?)

tests/kola/networking/nmstate/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+# Configure an ovs-bridge using nmstate and butane
```
bgilbert: This test shouldn't have platform-specific quirks, right?  Let's add a metadata block with `tags: platform-independent` so tests aren't launching an extra VM instance on every cloud platform.

cathay4t:
> 

No. The `nmstate` does not depend on `nmstate-libs`.

bgilbert:
Great.  So it sounds as though we don't need `nmstate-libs`.

tests/kola/networking/nmstate/test.sh
qinqon: Checking with ip link

qinqon:
Have reconstructed the PR and now there are two tests one configuring a normal state and the other using a policy, so we exercise the two main modes to configure networking with nmstate

qinqon:
> Great. So it sounds as though we don't need `nmstate-libs`.

qinqon:
@bgilbert looks like nmstate.service is running before NetworkManger is ready, [this](https://github.com/nmstate/nmstate/blob/base/packaging/nmstate.service) is the systemd unit:

```systemd
[Unit]
Description=Apply nmstate on-disk state
Documentation=man:nmstate.service(8) https://www.nmstate.io
After=NetworkManager.service
Requires=NetworkManager.service

[Service]
Type=oneshot
ExecStart=/usr/bin/nmstatectl service

[Install]
WantedBy=multi-user.target
```

Logs
```
Mar  6 07:50:05.269335 systemd[1]: Starting nmstate.service - Apply nmstate on-disk state...
Mar  6 07:50:05.277998 systemd[1]: Finished systemd-homed-activate.service - Home Area Activation.  
Mar  6 07:50:05.278000 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=systemd-homed-activate comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success' 
Mar  6 07:50:05.280000 audit: BPF prog-id=67 op=LOAD                            
Mar  6 07:50:05.281000 audit: BPF prog-id=68 op=LOAD                            
Mar  6 07:50:05.281000 audit: BPF prog-id=69 op=LOAD                            
Mar  6 07:50:05.290390 systemd[1]: Starting systemd-hostnamed.service - Hostname Service...
Mar  6 07:50:05.294788 systemd[1]: Started chronyd.service - NTP client/server. 
Mar  6 07:50:05.298000 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=chronyd comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Mar  6 07:50:05.308385 nmstatectl[1263]: [2023-03-06T07:50:05Z WARN  nmstate::nm::nm_dbus::error] Unknown DBUS error MethodError("org.freedesktop.DBus.Error.UnknownMethod", Some("Object does not exist at path “/org/freedesktop/NetworkManager”"), Msg { type: Error, sender: ":1.3", reply-serial: 2, body: Signature: [
Mar  6 07:50:05.308385 nmstatectl[1263]:             s (115),                    
Mar  6 07:50:05.308385 nmstatectl[1263]:     ] })                                
Mar  6 07:50:05.310309 nmstatectl[1263]: [2023-03-06T07:50:05Z ERROR nmstatectl::service] Failed to apply state file /etc/nmstate/br-ex.yml: NmstateError: Bug: DbusConnectionError: org.freedesktop.DBus.Error.UnknownMethod: Object does not exist at path “/org/freedesktop/NetworkManager”
Mar  6 07:50:05.310867 systemd[1]: nmstate.service: Deactivated successfully.    

```

cathay4t:
Above service dependency issue will be fixed by https://github.com/nmstate/nmstate/pull/2265

bgilbert:
I [commented](https://github.com/nmstate/nmstate/pull/2265#issuecomment-1455779778) in the PR, but in short, I don't think that PR is the right fix.

cathay4t:
@bgilbert I have update the PR with new approach (retry on daemon not start failure) as NM team confirmed this is a bug of NM.

cathay4t:
> Great. So it sounds as though we don't need `nmstate-libs`.

We need `nmstate-libs` as it will be sued by openshift assisted installer team or other team prefer library over CLI tool(nmstate package).

bgilbert:
Non-OS software should run in a container or be installed via package layering.  If OpenShift has a particular need for `nmstate-libs`, it might be added as an RHCOS-specific package, but that isn't necessarily an argument for adding it to FCOS.  Do you have specific information about how the package would be used?

qinqon:
> > Great. So it sounds as though we don't need `nmstate-libs`.
> 
> We need `nmstate-libs` as it will be sued by openshift assisted installer team or other team prefer library over CLI tool(nmstate package).

@cathay4t do you know if they need that just at their buld environment ? At the end they are linking statically rust with golang and delivering it with the "openshit-installer" binary right ?

bgilbert:
Noting for the record that nmstate has opted to [work around](https://github.com/nmstate/nmstate/pull/2265) the race condition in https://github.com/coreos/fedora-coreos-config/pull/2269#issuecomment-1455727702 by [sleeping for two seconds](https://github.com/nmstate/nmstate/pull/2265/files#diff-ae6ab4ca7ea12f71746b70b459afebc7a9141af402bd09c239daa567c8f4bdd7R27-R30).

Edit: [NetworkManager issue](https://bugzilla.redhat.com/show_bug.cgi?id=2177188) to track fixing the race upstream.

cathay4t:
Nmstate has released new update for Fedora 37 with fixes(workaround) of above race problem https://bodhi.fedoraproject.org/updates/FEDORA-2023-59543876eb

qinqon:
> Nmstate has released new update for Fedora 37 with fixes(workaround) of above race problem https://bodhi.fedoraproject.org/updates/FEDORA-2023-59543876eb

@bgilbert Do you know how much take for CI to consume it ?

bgilbert:
You should be able to just update the lockfiles in this PR.

qinqon:
> You should be able to just update the lockfiles in this PR.

After talking with Gris looks like we will have to wait a week

```
warning: Couldn't find locked package 'nmstate-2.2.8-2.fc37.x86_64'

[2023-03-13T07:37:22.675Z]   Packages matching NEVRA: 0; Mismatched checksums: 0

[2023-03-13T07:37:22.675Z]   Packages matching name and arch (2):

[2023-03-13T07:37:22.675Z]   nmstate-2.1.4-1.fc37.x86_64 (fedora)

[2023-03-13T07:37:22.675Z]   nmstate-2.2.5-1.fc37.x86_64 (fedora-updates)
```

bgilbert:
Oh, sorry, I wrote without thinking.  If you roll back that last set of lockfile changes and add a fast-track to the override lockfile (see the Git history for examples) CI should pick up the new packages.

dustymabe:
I pushed up a fix that should handle the package not being found. Since it's not in bodhi stable it needed to be added as an override.

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
```diff
@@ -32,3 +32,5 @@ enable rtas_errd.service
 enable clevis-luks-askpass.path
 # Provide status information about the Ignition run
 enable coreos-ignition-write-issues.service
+# Apply network configuration from /etc/nmstate/*.yml
+enable nmstate.service
```
dustymabe: feels like this should be in Fedora's preset files. Any reason why it should be here instead of [there](https://src.fedoraproject.org/rpms/fedora-release/blob/rawhide/f/90-default.preset)?

qinqon: Don't think so, we just want the service to be anabled, @bgilbert do you think it's better to have it [here](https://src.fedoraproject.org/rpms/fedora-release/blob/rawhide/f/90-default.preset) ?

bgilbert: If nmstate should be enabled (if installed) on all editions, as I assume it should, then I agree it should go in the Fedora-wide preset.

qinqon: Where do we add a patch to do so ?

qinqon:
failure is unrelated

cathay4t:
I have added `nmstate-2.2.8-2.fc37` into [bodhi overrides][1].


[1]: https://bodhi.fedoraproject.org/overrides/nmstate-2.2.8-2.fc37

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
dustymabe: I would just open a PR to https://src.fedoraproject.org/rpms/fedora-release

See https://src.fedoraproject.org/rpms/fedora-release/pull-request/252 for an example.

dustymabe:
> I have added `nmstate-2.2.8-2.fc37` into [bodhi overrides](https://bodhi.fedoraproject.org/overrides/nmstate-2.2.8-2.fc37).

Thanks @cathay4t - when I said `an override` [earlier](https://github.com/coreos/fedora-coreos-config/pull/2269#issuecomment-1466250002) I was actually referring to our FCOS overrides (see the modified force push I did, which solves the problem). There's nothing you need to do here. 

Sorry for the confusion.

cathay4t:
> 

No worry. I have expired the bodhi override.

cathay4t:
> 



> > > Great. So it sounds as though we don't need `nmstate-libs`.
> > 
> > 
> > We need `nmstate-libs` as it will be sued by openshift assisted installer team or other team prefer library over CLI tool(nmstate package).
> 
> @cathay4t do you know if they need that just at their buld environment ? At the end they are linking statically rust with golang and delivering it with the "openshit-installer" binary right ?

I got no reply from assisted installer team yet. And [asked][1] in `openshift-installer`.
Let's ignore `nmstate-libs` for now and only add it when they require so.

[1]: https://github.com/openshift/installer/pull/6589

qinqon:
@cathay4t looks like the test are passing

qinqon:
Failed on policy
```
Mar 14 06:33:52.465057 nmstatectl[1260]: [2023-03-14T06:33:52Z INFO  nmstate::nm::show] Got unsupported interface type generic: lo, ignoring
Mar 14 06:33:52.466050 nmstatectl[1260]: [2023-03-14T06:33:52Z ERROR nmstatectl::service] Failed to apply state file /etc/nmstate/br-ex-policy.yml: NmstateError: PolicyError: route with 'destination=::/0' not found
Mar 14 06:33:52.466050 nmstatectl[1260]:     | routes.running.destination=="::/0"
Mar 14 06:33:52.466050 nmstatectl[1260]:     | .......^                         
Mar 14 06:33:52.466767 systemd[1]: nmstate.service: Deactivated successfully.   
```

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
qinqon: @dustymabe I think my user there (quiquell) do not have permission for fork or create a PR

qinqon:
@dustymabe @bgilbert can you re-run the tests to ensure that this is not a false possitive ?

dustymabe:
> @dustymabe @bgilbert can you re-run the tests to ensure that this is not a false possitive ?

The tests look like they passed..

qinqon:
> > @dustymabe @bgilbert can you re-run the tests to ensure that this is not a false possitive ?
> 
> The tests look like they passed..

The error we have before was stating that there where no ipv6 default gw, so I was wandering why now is working, maybe the network configuration is different depending at what CI node the tests are running ?

tests/kola/networking/nmstate/data/commonlib.sh
bgilbert, line 1: Did you intend this file to be `test.sh` in the `policy` and `state` directories?  At present, the file is being ignored and neither test is being run.

qinqon: Forgot to `git add` them :-/

tests/kola/networking/nmstate/state/config.bu
```diff
@@ -0,0 +1,25 @@
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/nmstate/br-ex.yml
+      contents:
+        inline: |
+          interfaces:
+           - name: ens4
```
bgilbert: Is this specific to qemu?  If so, we should drop `platform-independent` from this test and add `platforms: qemu` instead.

tests/kola/networking/nmstate/data/nmstate-common.sh
```diff
@@ -0,0 +1,28 @@
...
+main() {
+    case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+      "")
+            run-checks "first boot"
+            /tmp/autopkgtest-reboot rebooted
+            ;;
+
+      rebooted)
+            run-checks "second boot"
+            if ! ls /etc/nmstate/*applied; then
```
bgilbert: Is this only true on the second boot, or will it also be true after `nmstate.service` runs on the first boot?

tests/kola/networking/nmstate/policy/test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+## kola:
+##   tags: "platform-independent"
+#
+# Configure a DHCP linux bridge using butane and nmstate service with policy
+
+set -xeuo pipefail
+
+. data/nmstate-common.sh
```
bgilbert:
```bash
. $KOLA_EXT_DATA/nmstate-common.sh
```

tests/kola/networking/nmstate/data/nmstate-common.sh
qinqon: Is moved after apply, so you right, the only different thing on second boot is that the policy is that the state is not applied since it has being moved.

Will change the test

tests/kola/networking/nmstate/state/config.bu
qinqon: Have change the test to just create the bridge, so it's platform independant

bgilbert (dismissed):

bgilbert (dismissed):

bgilbert:
Merge conflicts need resolution.

bgilbert (dismissed):

dustymabe:
tests are failing.

also can we get someone to answer the [questions](https://src.fedoraproject.org/rpms/fedora-release/pull-request/255#comment-134311) and try to get the fedora-release PR merged?

qinqon:
> tests are failing.
> 
> also can we get someone to answer the [questions](https://src.fedoraproject.org/rpms/fedora-release/pull-request/255#comment-134311) and try to get the fedora-release PR merged?

looks like qemu is running as ipv6 single stack at CI

dustymabe:
> looks like qemu is running as ipv6 single stack at CI

I'd be surprised if that was the case.

tests/kola/networking/nmstate/state/test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+## kola:
+##   tags: "platform-independent"
```
dustymabe:
```suggestion
##   # This is testing networking so let's add the `needs-internet`
##   # tag to bring up the full stack. 
##   tags: "platform-independent needs-internet"
```

dustymabe:
> > looks like qemu is running as ipv6 single stack at CI
> 
> I'd be surprised if that was the case.

Maybe try adding https://github.com/coreos/fedora-coreos-config/pull/2269/files#r1143551266 to both your tests.

tests/kola/networking/nmstate/state/test.sh
qinqon: thanks

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
```diff
@@ -32,3 +32,6 @@ enable rtas_errd.service
 enable clevis-luks-askpass.path
 # Provide status information about the Ignition run
 enable coreos-ignition-write-issues.service
+# Apply network configuration from /etc/nmstate/*.yml
+# We can drop this after merging https://src.fedoraproject.org/rpms/fedora-release/pull-request/255
+enable nmstate.service
```
cgwalters, line 37: One thing related to this...I think we want it so that nmstate.service is enabled by default iff `NetworkManager.service` is.  I think we should change
https://github.com/nmstate/nmstate/blob/1728ae21acb8764f49e5db321a64cb5a5414b504/packaging/nmstate.service#L12
to be `WantedBy=NetworkManager.service`.

Specifically, if someone turns off NM today, we shouldn't have a `nmstate.service` pointlessly spinning.

bgilbert (approved):

bgilbert:
@qinqon I'm confused by https://github.com/nmstate/nmstate/pull/2283.  Does that mean `nmstate.service` is failing in a default install today?  CI in this PR should have failed if that's the case.

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
dustymabe: If that's true then do we need https://src.fedoraproject.org/rpms/fedora-release/pull-request/255 ?

bgilbert: Yes, because the `WantedBy` would go in the `Install` section, so wouldn't have any effect until the service is enabled.

bgilbert:
I've confirmed that `nmstate.service` is not failing if `/etc/nmstate` doesn't exist, so I'm not sure what nmstate/nmstate#2283 is doing.

cathay4t:
> 

I tried with old nmstatectl. The `systemd start nmstate.service` will fail on missing /etc/nmstate folder with error:

Mar 22 14:27:19 c9s nmstatectl[4929]: std::io::Error: No such file or directory (os error 2)

Anyway, PR https://github.com/nmstate/nmstate/pull/2283 should fixed in next fedora update.

overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
bgilbert: Fixed in https://github.com/nmstate/nmstate/pull/2285.

cathay4t:
> 

The detail rpm based install will have /etc/nmstate with README file in it, hence no failure on default install.

travier:
Shouldn't this PR close https://github.com/coreos/fedora-coreos-tracker/issues/1175?

bgilbert:
No, the remaining steps are in https://github.com/coreos/fedora-coreos-tracker/issues/1175#issuecomment-1196286556.

PR #2270, opened 2023-03-01
ci: update for new kolaTestIso()

jlebon:
The `extraArgs4k` parameter no longer does anything. Just remove it so it's clear that we're accepting the default set of tests from `kola testiso`.

This will run more tests than before. We'll likely cut down tests we consider redundant soon, but that should be transparent to this code.


bgilbert (approved):

bgilbert:
LGTM

PR #2271, opened 2023-03-02
Move chrony config from generator to systemd service

gursewak1997:
With the latest changes from systemd, we now run certain steps before `NetworkManager.service` to access and write to `/etc/sysconfig/network`. We completely got rid of the usage of generators.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1402


dustymabe:
Unfortunately this is getting complicated :( 

> Long-term, we are looking to eliminate that file entirely and just use systemd unit service.

Can you expand on the strategy there?

I'm thinking `1.` maybe we should push harder to get everything out of a generator now OR `2.` take a more tactical approach to updating `/etc/sysconfig/network` and leave most of the logic in the generator for now (until we can get rid of it completely.

I actually have ideas on how we can achieve either `1.` or `2.`. Maybe we can discuss it tomorrow.

overlay.d/20platform-chrony/usr/lib/systemd/system/coreos-platform-chrony-config.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Call coreos-chrony script that inspects
+# the platform and reconfigures the chrony
```
dustymabe: This looks weird because the statement is broken up across two lines one of them being a comment

Let's just simplify it and get it down to one line: 

```suggestion
Description=Configure chrony based on the platform
```

overlay.d/20platform-chrony/usr/lib/systemd/system/chronyd.service.d/coreos-platform-chrony.conf
```diff
@@ -0,0 +1,4 @@
+# Override the env file to /run/sysconfig-coreos-chrony
+# Supports writing new OPTIONS cariable to this file
```
dustymabe: How about we say where the `/run/sysconfig-coreos-chrony` file comes from. 

```suggestion
# Override chrony configuration with the output of
# coreos-platform-chrony-config.service.
```

dustymabe (dismissed):
A few small comment fixups.. otherwise LGTM

How did this perform in testing?

dustymabe:
looking at the CI failure we probably need to delete the `coreos-platform-chrony-generator-permissions` test now that we no longer have a generator.

dustymabe (dismissed):
LGTM - tagging in @jlebon for a second pair of eyes.

overlay.d/20platform-chrony/usr/lib/systemd/system/coreos-platform-chrony-config.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Configure chrony based on the platform
```
jlebon: Minor: we usually prefix our units with "CoreOS". Also, the convention for systemd unit descriptions is Title Case.

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
```diff
@@ -3,37 +3,37 @@ set -euo pipefail
...
 
-# Exit early if chrony configuration as been changed from the image default
+# Exit early if chrony configuration has been changed from the image default
 if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
     echo "$self: /etc/chrony.conf is modified; not changing the default"
     exit 0
 fi
 
+confpath=/run/coreos-platform-chrony.conf
+altenvfilepath=/run/sysconfig-coreos-chrony
```
jlebon: While we're here, I wonder if we should put these under `/run/coreos` instead? (And drop the `coreos` in the basename.) Now that it's two files, it starts to look a bit messy putting it in the top-level.

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
```diff
@@ -82,14 +82,11 @@ esac
...
-mkdir -p $(dirname "${unitconfpath}")
-cat >"${unitconfpath}" << EOF
-[Service]
-ExecStart=
-ExecStart=/usr/sbin/chronyd -f ${confpath} \$OPTIONS
-EOF
+# Read in the existing $OPTIONS variable setting from /etc/sysconfig/chrony
+# and write out a new $OPTIONS variable (with specified new configuration path)
+# to /run/sysconfig-coreos-chrony
+source ${altenvfilepath}
```
jlebon: Shouldn't this be `/etc/sysconfig/chronyd`?

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
```diff
@@ -82,14 +82,11 @@ esac
...
-UNIT_DIR="${1:-/tmp}"
 
-unitconfpath="${UNIT_DIR}/chronyd.service.d/coreos-platform-chrony.conf"
-mkdir -p $(dirname "${unitconfpath}")
-cat >"${unitconfpath}" << EOF
-[Service]
-ExecStart=
-ExecStart=/usr/sbin/chronyd -f ${confpath} \$OPTIONS
-EOF
+# Read in the existing $OPTIONS variable setting from /etc/sysconfig/chrony
```
jlebon:
```suggestion
# Read in the existing $OPTIONS variable setting from /etc/sysconfig/chronyd
```

jlebon:
Just noticed this PR is targeting the `rawhide` branch. It needs to target `testing-devel`. Were there concerns with doing this change in f37?

dustymabe:
> Just noticed this PR is targeting the `rawhide` branch. It needs to target `testing-devel`. Were there concerns with doing this change in f37?

good catch. Yes, this should target `testing-devel`.

overlay.d/20platform-chrony/usr/lib/systemd/system/coreos-platform-chrony-config.service
gursewak1997: Can you elaborate on the description part? Also, don't we already have "coreos" as the prefix for this service?

dustymabe: I think he's saying that we usually prefix our units with `CoreOS` in the description so that in the logs it's clear to us that it's one of our systemd units we defined.

Based on his suggestion something like:

```
CoreOS Configure Chrony Based On The Platform
```

Would work.

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
dustymabe: this still needs to be updated

gursewak1997:
With the recent changes, the `basic` test passes but the `ext.config.ntp.chrony.coreos-platform-chrony-generator` fails. The CI here isn't testing that as this test only runs on `aws azure gce`. Working on the fix as the logs aren't helping much.

dustymabe:
> With the recent changes, the `basic` test passes but the `ext.config.ntp.chrony.coreos-platform-chrony-generator` fails. The CI here isn't testing that as this test only runs on `aws azure gce`. Working on the fix as the logs aren't helping much.

For one, this still needs to be addressed: https://github.com/coreos/fedora-coreos-config/pull/2271/files#r1124818300

overlay.d/20platform-chrony/usr/lib/systemd/system/chronyd.service.d/platform-chrony.conf
```diff
@@ -0,0 +1,4 @@
+# Override chrony configuration with the output of
+# coreos-platform-chrony-config.service.
+[Service]
+EnvironmentFile=-/etc/sysconfig/chronyd
```
dustymabe: this should go back to the value it was before.

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
```diff
@@ -3,37 +3,36 @@ set -euo pipefail
...
-
-# Exit early if chrony configuration as been changed from the image default
+# Exit early if chrony configuration has been changed from the image default
 if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
     echo "$self: /etc/chrony.conf is modified; not changing the default"
     exit 0
 fi
 
+confpath=/run/coreos/platform-chrony.conf
+altenvfilepath=/etc/sysconfig/chronyd
```
dustymabe: This should go back to the value it was before

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
```diff
@@ -82,14 +81,11 @@ esac
...
-mkdir -p $(dirname "${unitconfpath}")
-cat >"${unitconfpath}" << EOF
-[Service]
-ExecStart=
-ExecStart=/usr/sbin/chronyd -f ${confpath} \$OPTIONS
-EOF
+# Read in the existing $OPTIONS variable setting from /etc/sysconfig/chronyd
+# and write out a new $OPTIONS variable (with specified new configuration path)
+# to /etc/sysconfig/chronyd
+source ${altenvfilepath}
```
dustymabe: This should say `source /etc/sysconfig/chronyd`

dustymabe (approved):
LGTM if the tests pass!

PR #2272, opened 2023-03-02
transposefs: add new workaround using try in loop

HuijingHei:
Seems the previous workaround https://github.com/coreos/fedora-coreos-config/pull/2226 can not completely resolve the race issue, try in loop with a small sleep for el9 + !x86_64.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -294,14 +294,28 @@ case "${1:-}" in
         if [ -d "${saved_data}" ]; then
             read dev < "${zram_dev}"
             umount "${saved_data}"
+            rm -rf "${saved_data}" "${partstate_root}"
             # After unmounting, make sure zram device state is stable before we remove it.
             # See https://github.com/openshift/os/issues/1149
             # Should remove when https://bugzilla.redhat.com/show_bug.cgi?id=2172058 is fixed.
+            # Seems the previous workaround https://github.com/coreos/fedora-coreos-config/pull/2226
+            # can not completely resolve the race issue, try in loop with a small sleep for el9 + !x86_64.
             if [ $(uname -m) != x86_64 ] && is_rhcos9; then
-                udevadm trigger --settle "/dev/zram${dev}"
+                for x in {0..10}; do
+                    if ! echo "${dev}" > /sys/class/zram-control/hot_remove 2>/dev/null; then
+                        sleep 0.1
+                    else
+                        dev=
+                        break
+                    fi
+                done
+                # try it one last time and let it possibly fail
+                if [ -n "${dev}" ]; then
+                    echo "${dev}" > /sys/class/zram-control/hot_remove
+                fi
+            else
+                echo "${dev}" > /sys/class/zram-control/hot_remove
             fi
-            rm -rf "${saved_data}" "${partstate_root}"
-            echo "${dev}" > /sys/class/zram-control/hot_remove
         fi
         ;;
     *)


HuijingHei:
I have not tested with this patch yet.
/hold

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -294,14 +294,23 @@ case "${1:-}" in
...
             # After unmounting, make sure zram device state is stable before we remove it.
             # See https://github.com/openshift/os/issues/1149
             # Should remove when https://bugzilla.redhat.com/show_bug.cgi?id=2172058 is fixed.
+            # Seems the previous workaround https://github.com/coreos/fedora-coreos-config/pull/2226
+            # can not completely resolve the race issue, try in loop with a small sleep for el9 + !x86_64.
             if [ $(uname -m) != x86_64 ] && is_rhcos9; then
-                udevadm trigger --settle "/dev/zram${dev}"
+                for x in {0..10}; do
+                    if ! echo "${dev}" > /sys/class/zram-control/hot_remove 2>/dev/null; then
+                        sleep 0.01
```
cgwalters: Let's do `0.1` per Dusty's comment; reprovisioning the OS is so slow that anything that takes hundredths of a second is in the noise.  I think we could easily wait an additional 10 seconds, but my suggestion takes us just to waiting a max of 1 instead of 1/10.

cgwalters:
One suggestion when testing this is to use `--parallel 5 --multiply 5 ext.luks.*`.

travier:
For reference, the `jira` label only works on issues, not PRs unfortunately.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -294,14 +294,23 @@ case "${1:-}" in
...
+            rm -rf "${saved_data}" "${partstate_root}"
             # After unmounting, make sure zram device state is stable before we remove it.
             # See https://github.com/openshift/os/issues/1149
             # Should remove when https://bugzilla.redhat.com/show_bug.cgi?id=2172058 is fixed.
+            # Seems the previous workaround https://github.com/coreos/fedora-coreos-config/pull/2226
+            # can not completely resolve the race issue, try in loop with a small sleep for el9 + !x86_64.
             if [ $(uname -m) != x86_64 ] && is_rhcos9; then
-                udevadm trigger --settle "/dev/zram${dev}"
+                for x in {0..10}; do
+                    if ! echo "${dev}" > /sys/class/zram-control/hot_remove 2>/dev/null; then
```
jlebon, line 305: Note that right now if this fails all 10 times, we keep going instead of erroring out. One idea is in the `else` block, we do `dev=` before we break and then outside the loop, we check if `dev` is still non-empty and if so, do a final `echo` and let it succeed or fail.

HuijingHei: Sound reasonable to me.
I have not tested it yet, sorry that just reserve a machine with ppc64le platform, will do more testing tomorrow.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -294,14 +294,28 @@ case "${1:-}" in
...
+                        sleep 0.1
+                    else
+                        dev=1
```
jlebon: Ahh sorry, I should've been more clear. What I meant here is to empty out the `dev` variable. 1 could be a valid value for `dev`. So this would be

```suggestion
                        dev=
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -294,14 +294,28 @@ case "${1:-}" in
...
+                        break
+                    fi
+                done
+                # if dev != 1, means action failed after full loop
+                if [ ${dev} -ne 1 ]; then
```
jlebon: And then here, e.g.

```suggestion
                # try it one last time and let it possibly fail
                if [ -n "${dev}" ]; then
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: If you agree with the latest comments below, I can update this PR and test it today if you'd like. :)

jlebon:
With the suggested changes above applied, I ran `kola run luks.sss.t1 --multiply 30` on ppc64le twice: once as is, and once with `--parallel 5`. None hit the race condition. Almost all passed (a few timed out after 10 minutes, but well past the point we care about here).

Assuming you apply the changes above, you can consider this tested.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
HuijingHei: Yes, agree, go ahead. Thanks!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
HuijingHei: I saw your comments about `dev=`, just thought `dev=1` could look better. I am OK with the change `dev=`. Thanks!

jlebon (approved):
Thanks @HuijingHei!

PR #2280, opened 2023-03-07
tests/kola: add version comparison functions to commonlib.sh

dustymabe:
Leveraging these functions will make it easier to do comparision checks inside of our tests.

Specifically for RPM version comparisons something like /usr/bin/rpmdev-vercmp would be better but that would require having rpmdevtools installed in FCOS, which we don't have.

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -86,3 +86,31 @@ is_service_active() {
     # return actual result
     systemctl is-active "${service}"
 }
+
+## Some functions to support version comparison
+# Returns true iff $1 is equal to $2
+vereq() {
+    [ "$1" == "$2" ]
+}
+
+# Returns true iff $1 is less than $2
+verlt() {
+    vereq $1 $2 && return 1
+    local lowest="$(echo -e "$1\n$2" | sort -V | head -n 1)"
+    [ "$1" == "$lowest" ]
+}
+
+# Returns true iff $1 is less than or equal to $2
+verlte() {
+    vereq $1 $2 || verlt $1 $2
+}
+
+# Returns true iff $1 is greater than to $2
+vergt() {
+    ! verlte $1 $2
+}
+
+# Returns true iff $1 is greater than or equal to $2
+vergte() {
+    vereq $1 $2 || vergt $1 $2
+}
--- a/tests/kola/systemd/default-unit-timeouts
+++ b/tests/kola/systemd/default-unit-timeouts
@@ -1,6 +1,7 @@
 #!/bin/bash
 ## kola:
 ##   exclusive: false
+##   tags: "platform-independent"
 #
 # Checks if default shutdown time is 1m 30s
 #
--- a/tests/kola/systemd/default-unit-timeouts
+++ b/tests/kola/systemd/default-unit-timeouts
@@ -41,7 +41,7 @@ else
     fatal $error
 fi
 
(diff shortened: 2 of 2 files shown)


dustymabe:
tested this with:

```
$ cat test-sort.sh 
. tests/kola/data/commonlib.sh
compare() {
    A=$1; B=$2
    echo "Comparing $A and $B"
    vereq $A $B && echo "$A is equal to $B"
    verlt $A $B && echo "$A is < $B"
    verlte $A $B && echo "$A is <= $B"
    vergt $A $B && echo "$A is > $B"
    vergte $A $B && echo "$A is => $B"
}

compare "0" "1"
compare "1" "1"
compare "1" "2"
compare "1.2.3.4" "1.2.3.5"
compare "100.1" "2"
compare "253.1" "2002.1"
compare "253.1" "253.1"


$ bash test-sort.sh 
Comparing 0 and 1
0 is < 1
0 is <= 1
Comparing 1 and 1
1 is equal to 1
1 is <= 1
1 is => 1
Comparing 1 and 2
1 is < 2
1 is <= 2
Comparing 1.2.3.4 and 1.2.3.5
1.2.3.4 is < 1.2.3.5
1.2.3.4 is <= 1.2.3.5
Comparing 100.1 and 2
100.1 is > 2
100.1 is => 2
Comparing 253.1 and 2002.1
253.1 is < 2002.1
253.1 is <= 2002.1
Comparing 253.1 and 253.1
253.1 is equal to 253.1
253.1 is <= 253.1
253.1 is => 253.1
```

tests/kola/data/commonlib.sh
```diff
@@ -86,3 +86,26 @@ is_service_active() {
...
 }
+
+# Some functions to support version comparison
+vereq() {
+    [ "$1" == "$2" ]
+}
+
+verlt() {
+    vereq $1 $2 && return 1
+    sorted="$(echo -e "$1\n$2" | sort -V)"
```
jlebon: `local`

tests/kola/data/commonlib.sh
```diff
@@ -86,3 +86,26 @@ is_service_active() {
...
+    [  "$1" == $(echo "$sorted" | head -n 1) ]
```
jlebon: Minor: would read better as

```suggestion
    local lowest="$(echo -e "$1\n$2" | sort -V | head -n 1)"
    [  "$1" == $lowest ]
```

tests/kola/data/commonlib.sh
```diff
@@ -86,3 +86,26 @@ is_service_active() {
     # return actual result
     systemctl is-active "${service}"
 }
+
+# Some functions to support version comparison
+vereq() {
+    [ "$1" == "$2" ]
+}
+
+verlt() {
```
jlebon, line 97: A comment for this to make it clear that it succeeds if the first arg is less than the second arg could be nice.

jlebon (dismissed):
Overall LGTM

dustymabe:
@jlebon - pushed an update that should address those comments.

Adam0Brien (dismissed):
LGTM

dustymabe:
Looks like my new comments like:

```
# Returns true if $1 is less than or equal to $2
```

tripped up https://github.com/coreos/fedora-coreos-config/pull/2048/commits/41e87994811e7eb69d708279fd1521dc2ecae4ea. I guess that code doesn't consider if the string it's searching for is part of a comment in bash or not.

dustymabe:
because I'm lazy I just replaced `s/if/iff` in the comments.

jlebon (approved):

PR #2281, opened 2023-03-07
platform-chrony: Make /run/coreos dir so we can create platform-chrony.conf

gursewak1997:
We are hitting `/usr/libexec/coreos-platform-chrony-config: line 61: /run/coreos/platform-chrony.conf: No such file or directory` error for a few kola tests. To fix that, we will create the required directory(`/run/coreos/`) beforehand.

### DIFF
--- a/overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
+++ b/overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
@@ -19,6 +19,7 @@ if ! cmp {/usr,}/etc/chrony.conf >/dev/null; then
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
```diff
@@ -48,6 +48,7 @@ PEERNTP=no
 EOF
 fi
 
+mkdir -p /run/coreos
```
dustymabe: Maybe let's put it up on line 21/22 so it's clear why it's needed (i.e. those following two vars are going to use that directory).

dustymabe (approved):
LGTM

PR #2282, opened 2023-03-07
denylist: bump snooze for files.file-directory-permissions

dustymabe:
The systemd package in F38 (next-devel/branched) is frozen right now so we won't get the fix for this until after beta is released. Let's snooze again for now.


gursewak1997 (approved):
Lgtm

PR #2283, opened 2023-03-08
tests: add "ext.config.secex.ensure" test

nikita-dubrovskii:
This is the 1st test for qemu-secex images.
To run `--qemu-secex` and `--tag secex` options should be provided:
```
cosa kola run --qemu-secex --tag secex
```

Requires: https://github.com/coreos/coreos-assembler/pull/3384

### DIFF
--- a/tests/kola/secex/ensure/test.sh
+++ b/tests/kola/secex/ensure/test.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+## kola:
+##   # This test verifies the s390x Secure Execution QEMU image works. It also
+##   # implicitly tests Ignition config decryption. We don't run it by default
+##   # because it requires running with `--qemu-secex`.
+##   architectures: s390x
+##   platforms: qemu
+##   requiredTag: secex
+##   timeoutMin: 3
+
+set -xeuo pipefail
+
+check_luks() {
+    local mnt dev type
+    mnt=${1}
+    dev=$(findmnt -nvr ${mnt} -o SOURCE)
+    type=$(lsblk -o TYPE "${dev}" --noheadings)
+    [[ ${type} == crypt ]]
+}
+
+# 1 means system runs with Secure Execution
+grep -q 1 /sys/firmware/uv/prot_virt_guest
+
+# Check firstboot kargs have dm-verity hashes
+grep -q rootfs.roothash /proc/cmdline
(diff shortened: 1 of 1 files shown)


tests/kola/secex/ensure/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -xeuo pipefail
+# kola: { "architectures": "s390x", "platforms": "qemu-unpriv", "tags": "secex", "requiredTag": "secex", "minMemory": 4096, "timeoutMin": 5 }
```
jlebon: - The kola line is usually placed right after the first line.
- New tests use YAML now instead of JSON (see other tests)
- We don't need `tags` if we have `requiredTag`

tests/kola/secex/ensure/test.sh
```diff
@@ -0,0 +1,10 @@
...
+
+grep -q rootfs.roothash /proc/cmdline
+grep -q bootfs.roothash /proc/cmdline
+grep -q 1 /sys/firmware/uv/prot_virt_guest
+test -h /dev/disk/by-label/se
+test -h /dev/disk/by-label/crypt_bootfs
+test -h /dev/disk/by-label/crypt_rootfs
```
jlebon: Instead of/in addition to this, we could check directly that the block device that `/` and `/boot` are on are LUKS. See e.g. `tests/kola/root-reprovision/luks/test.sh`.

tests/kola/secex/ensure/test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+## kola:
+##  architectures: s390x
+##  platforms: qemu
+##  requiredTag: secex
+##  timeoutMin: 3
```
jlebon: For consistency with other tests, maybe let's add a comment before these knobs with some context. Something like:

```sh
## # This test verifies the s390x Secure Execution QEMU image works. It also
## # implicitly tests Ignition config decryption. We don't run it by default
## # because it requires running with `--qemu-secex`.
```

Also, suuuper minor: also for consistency with other tests, let's use an indentation of 2 spaces?

jlebon (dismissed):
One comment about comments, otherwise LGTM!

jlebon (approved):

PR #2285, opened 2023-03-08
tests/kola: add test for custom SSH host key with mode 640

bgilbert:
Exclude the embedded SSH private key from Red Hat secret scanning.

For https://github.com/coreos/fedora-coreos-tracker/issues/1394.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,5 @@
+[allowlist]
+description = "Exclude test keys from Red Hat secret scanning"
+files = [
+    ".*custom-host-key-permissions.*",
+]
--- a/tests/kola/ssh/custom-host-key-permissions/test.sh
+++ b/tests/kola/ssh/custom-host-key-permissions/test.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+## kola:
+##   tags: platform-independent
+#
+# Check that sshd still works with a custom host key with mode 640 and
+# group ssh_keys
+# https://github.com/coreos/fedora-coreos-tracker/issues/1394
+# https://src.fedoraproject.org/rpms/openssh/pull-request/37
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# recent Fedora sshd binaries will fail to start if all configured host keys
+# have mode > 600 and their modes haven't automatically been fixed
+# that part is implicitly tested by kola, which needs sshd to connect
+
+# make sure our key was actually used
+# grep -q causes sshd -T to fail on SIGPIPE
+if ! sshd -T | grep "^hostkey /etc/ssh-host-key$" >/dev/null; then
+    sshd -T
+    fatal "configured host key not used by sshd"
+fi
+ok "configured host key used by sshd"
+
+# sshd starts successfully if any keys are mode 600, which would invalidate
+# the test except that our HostKey directive should replace the default
+# keys.  verify this.
+if [[ $(sshd -T | grep "^hostkey " | wc -l) != 1 ]]; then
+    sshd -T | grep "^hostkey "
+    fatal "sshd uses multiple host keys"
+fi
+ok "sshd uses only the configured host key"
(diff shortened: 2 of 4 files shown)


tests/kola/ssh/custom-host-key-permissions/test.sh
```diff
@@ -0,0 +1,22 @@
...
+# Checks that sshd still works with a custom host key with mode 640
+# https://github.com/coreos/fedora-coreos-tracker/issues/1394
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# recent sshd releases will fail to start if a host key has mode > 600 and
+# its mode hasn't automatically been fixed
+# that part is implicitly tested by kola, which needs sshd to connect
```
dustymabe, line 16: I imagine we could do more here and check that the ssh-host-key-migration service ran and succeeded (on f38+), but like you said here it's not strictly necessary because we can't run this test if we can't SSH in.

dustymabe (dismissed):
LGTM

tests/kola/ssh/custom-host-key-permissions/test.sh
bgilbert: Yeah, I originally had code to do that, but didn't want to tie this test to F38+ or to the exact workaround we're currently using.

bgilbert:
Updated to (hopefully) exclude the private key from Red Hat's secret scanning.

.gitleaks.toml
```diff
@@ -0,0 +1,5 @@
+[allowlist]
```
jlebon, line 1: TIL! Did you get an email from this? :laughing:

tests/kola/ssh/custom-host-key-permissions/test.sh
jlebon: From when I was testing this, sshd only fails to start if none of the keys have valid permissions (OK yup, just retested this). Since we don't disable the regular sshd keygen services, the default keys will always be valid. So I think it does make sense to verify that the custom key we added was migrated correctly on f38+? Alternatively, we can mask `sshd-keygen@.service`.

.gitleaks.toml
bgilbert: Yup.  coreos-installer has a [similar config](https://github.com/coreos/coreos-installer/blob/f7a7bcafec58cf66deb04cbba660ef2655dd16cf/.gitleaks.toml).

tests/kola/ssh/custom-host-key-permissions/test.sh
bgilbert: Ah, good point.  It actually doesn't matter here, though, since the `HostKey` directive replaces the default list of host keys.  I've added a check to ensure we're only using the single host key.

bgilbert:
Okay, I think I got the GitLeaks syntax correct now.

tests/kola/ssh/custom-host-key-permissions/test.sh
```diff
@@ -0,0 +1,31 @@
...
+#
+# Checks that sshd still works with a custom host key with mode 640
+# https://github.com/coreos/fedora-coreos-tracker/issues/1394
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# recent sshd releases will fail to start if a host key has mode > 600 and
+# its mode hasn't automatically been fixed
```
jlebon: Minor/optional: this comment could use a tweak.

jlebon (dismissed):
LGTM! Thanks for writing this.

bgilbert:
It turns out that older OpenSSH packages don't allow mode 640 with arbitrary group ownership; the group has to be `ssh_keys`.  We [ship that group](https://github.com/coreos/fedora-coreos-config/blob/423235a7b1b8ced48511ca5ae6161cbe04f82126/manifests/group#L26) on F37, F38, and RHCOS, so it's safe to use here.

jlebon (approved):

PR #2289, opened 2023-03-10
[next] tree: promote changes from next-devel at b8ebc03c9b8b01bf08bf3cbf794d4d4a95cfa408

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/663.


dustymabe:
I triggered this change so won't leave an official review, but will say LGTM

manifest.yaml
```diff
@@ -2,6 +2,6 @@ variables:
   stream: next
   prod: true
 
-releasever: 37
+releasever: 38
```
dustymabe, line 5: This change is expected (and done in a separate commit) so we can ignore this "Unexpected change" warning.

jlebon (approved):

PR #2290, opened 2023-03-10
platforms.yaml: Add serial console information for KubeVirt platform.

dustymabe:


platforms.yaml
```diff
@@ -115,6 +115,18 @@ x86_64:
...
+    # No official docs on this for kubevirt. Requested some:
+    # https://github.com/kubevirt/kubevirt/issues/9400
+    # The web UI displays the VNC console first, so go with
+    # graphical console primary, serial console secondary.
+    grub_commands:
+      - serial --speed=19200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=ttyS0,19200n81
```
dustymabe: the [blog](https://kubevirt.io/2020/Customizing-images-for-containerized-vms.html#building-standard-centos-8-image) has `console=ttyS0,19200n81` but I almost wonder if we should just go with `115200n8` since there isn't any official guidance here.

bgilbert: If 115200 works (as it should, since this is just qemu), we should absolutely use it.  19200 will slow down the console substantially.

bgilbert: `81` isn't a thing.  The karg syntax provides no way to specify stop bits.

dustymabe: ok i'll update and test it.

bgilbert:
Commit message nit: the formal capitalization is KubeVirt.

platforms.yaml
dustymabe: yeah I was just taking the entry from the [blog post](https://kubevirt.io/2020/Customizing-images-for-containerized-vms.html#building-standard-centos-8-image). Fixed.

bgilbert (approved):

platforms.yaml
dustymabe: tested, working!

PR #2303, opened 2023-03-16
kola: Add ignition.journald-log test

gursewak1997:
Converting `coreos.ignition.journald-log` test to an external one.
Ref: https://github.com/coreos/coreos-assembler/pull/3391#pullrequestreview-1339641509

### DIFF
--- a/tests/kola/ignition/data/commonlib.sh
+++ b/tests/kola/ignition/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh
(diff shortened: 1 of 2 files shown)


dustymabe:
The code LGTM, but I think we should move this up one level. All of the other ignition tests are in their own subdirectory because they either provide their own butane config or a ignition config. This one doesn't have a specific ignition/butane config (it's exclusive=false) so it doesn't need to be in it's own subdir.

```
cd tests/kola/ignition/
git rm journald-log/data/commonlib.sh
git mv journald-log/test.sh ./
rmdir journald-log
git mv test.sh journald-log
mkdir data
cd data; ln -s ../../data/commonlib.sh ./
git add commonlib.sh
git commit --amend
```

dustymabe (approved):
LGTM

PR #2304, opened 2023-03-16
[next-devel] overrides: pin on kernel-6.1.18-200.fc37 from F37

dustymabe:
The new 6.2 kernel can cause aarch64 systems originally installed on F36- to not boot. For now while we figure out the best path forward we'll ship the newest 6.1 kernel we can find, which just happens to be built against F37.

See https://github.com/coreos/fedora-coreos-tracker/issues/1441

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -19,3 +19,18 @@ packages:
     metadata:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-8c9cfc425e
       type: fast-track
+  kernel:
+    evr: 6.1.18-200.fc37
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1441
+      type: pin
+  kernel-core:
+    evr: 6.1.18-200.fc37
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1441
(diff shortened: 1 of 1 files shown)


dustymabe:
I don't know if this is a good idea or not, but CI did at least pass (but failed on the RPM downgrade detection).

jmflinuxtx:
The potential problems with running an F37 kernel on F38 userspace include the inability to load a module built on the system.  This one is pretty easy as the error is clear, and it will not load. Additionally, kernel-headers are always forward compatible, but not always backwards compatible. The full rebuild for F38 was done with 6.2 kernel headers.  It is possible that there are userspace packages which will fail in various ways if they are using something introduced in 6.2 which did not exist in 6.1. We have done no audit to see how likely that the changes in uapi headers  would trigger an issue, but for reference:
```
git diff v6.1..v6.2 include/uapi/ | wc -l
4444
```
diffstat shows  89 files changed, 1921 insertions(+), 620 deletions(-)

There is a potential for userspace program crashes which are not obvious at all, and not reproducible outside of trying to run a 6.1.x kernel on an F38 userspace.

dustymabe:
Thanks for that perspective @jmflinuxtx - I'll leave this open for now, but will probably close it tomorrow after we discuss our options amongst the team.

PR #2308, opened 2023-03-20
overlay/15fcos: upgrade bootloader on aarch64 at boot

bgilbert:
On aarch64, kernel 6.2 won't boot with older versions of GRUB.  In preparation for switching to the new kernel, add a systemd service that uses bootupd to update the bootloader on aarch64 systems.

Revert this after the next barrier release.

For https://github.com/coreos/fedora-coreos-tracker/issues/1441.  Not heavily tested.


dustymabe (approved):
LGTM

dustymabe:
I'm OK to go ahead and get this in and a build running while we continue to test.

PR #2310, opened 2023-03-21
tests/kola: add extended upgrade test

dustymabe:
This test will attempt to test an upgrade from a given starting point (assumed by the caller passing in a specific `cosa kola run --build=x.y.z`) all the way to the latest build that is staged to be released. The test is basic in that it essentially tests 1) updates work 2) boot works.

An example invocation for this test would look like:

```
cosa buildfetch --stream=next --build=34.20210904.1.0 --artifact=qemu
cosa decompress --build=34.20210904.1.0
cosa kola run --build=34.20210904.1.0 --tag extended-upgrade
```

You can then monitor the console.txt or journal.txt to watch the upgrades happen via `tail -f tmp/kola/ext.config.upgrade.extened/*/console.txt`.

Another useful thing to do is to just monitor the major events:

```
$ tail -f  tmp/kola/ext.config.upgrade.extended/*/journal.txt | grep --color -i 'ok reached version'
Mar 20 02:25:12.584096 kola-runext-test.sh[2085]: ok Reached version: 34.20210904.1.0
Mar 20 02:26:37.107738 kola-runext-test.sh[3538]: ok Reached version: 35.20210924.1.0
Mar 20 02:28:08.509482 kola-runext-test.sh[1668]: ok Reached version: 36.20220418.1.0
Mar 20 02:28:41.242555 kola-runext-test.sh[1649]: ok Reached version: 36.20220906.1.0
Mar 20 02:29:33.635979 kola-runext-test.sh[1635]: ok Reached version: 37.20221111.1.0
Mar 20 02:30:16.657854 kola-runext-test.sh[1469]: ok Reached version: 37.20230303.1.0
```

### DIFF
--- a/tests/kola/upgrade/extended/config.bu
+++ b/tests/kola/upgrade/extended/config.bu
@@ -1,0 +1,24 @@
+variant: fcos
+# Must use version 1.0.0 here to use Ignition spec 3.0.0 for
+# our oldest supported starting points (i.e. 31.20200108.3.0)
+version: 1.0.0
+storage:
+  files:
+    - path: /etc/systemd/journald.conf.d/forward-to-console.conf
+      mode: 0644
+      contents:
+        inline: |
+          # Send journal messages to the console. This is so we can see the
+          # progress even after autopkgtest-reboot-prepare has been called.
+          [Journal]
+          ForwardToConsole=yes
+    - path: /etc/zincati/config.d/99-config.toml
+      mode: 0644
+      contents:
+        inline: |
+          # Don't wait for any rollout window, update ASAP
+          [identity]
+          rollout_wariness = 0.0000
+          # Increase the frequency at which we check for updates
+          [agent.timing]
+          steady_interval_secs = 20
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -1,0 +1,158 @@
+#!/bin/bash
+## kola:
+##   # - needs-internet: to pull updates
+##   tags: "needs-internet"
+##   # Extend the timeout since a lot of updates/reboots can happen.
+##   timeoutMin: 45
+##   # Only run this test when specifically requested.
+##   requiredTag: extended-upgrade
+
+set -eux -o pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# This test will attempt to test an upgrade from a given starting
+# point (assumed by the caller passing in a specific
+# `cosa kola run --build=x.y.z`) all the way to the latest build
+# that is staged to be released. The test is basic in that it
+# essentially tests 1) updates work 2) boot works.
+#
+# An example invocation for this test would look like:
+
+# ```
+# cosa buildfetch --stream=next --build=34.20210904.1.0 --artifact=qemu
+# cosa decompress --build=34.20210904.1.0
+# cosa kola run --build=34.20210904.1.0 --tag extended-upgrade
+# ```
+#
+# You can monitor the progress from the console and journal:
+#   - everything:
+#       - tail -f tmp/kola/ext.config.upgrade.extended/*/console.txt
+#   - major events:
+#       - tail -f tmp/kola/ext.config.upgrade.extended/*/journal.txt | grep --color -i 'ok reached version'
+#
+# For convenience, here is a list of the earliest releases on each
+# stream/architecture:
+#
+# stable
+#   - x86_64  31.20200108.3.0
+#   - aarch64 34.20210821.3.0
+#   - s390x   36.20220618.3.1
+# testing
+#   - x86_64  30.20190716.1
+#   - aarch64 34.20210904.2.0
+#   - s390x   36.20220618.2.0
+# next
+#   - x86_64  32.20200416.1.0
+#   - aarch64 34.20210904.1.0
+#   - s390x   36.20220618.1.1
+
+. /etc/os-release # for $VERSION_ID
+
+# delete the disabling of updates that was done by the test framework
+if [ -f /etc/zincati/config.d/90-disable-auto-updates.toml ]; then
+    rm -f /etc/zincati/config.d/90-disable-auto-updates.toml
+    systemctl restart zincati
+fi
+
+version=$(rpm-ostree status  --json | jq -r '.deployments[0].version')
+stream=$(rpm-ostree status  --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]')
+
+test -f /srv/releases.json || \
+    curl -L "https://builds.coreos.fedoraproject.org/prod/streams/${stream}/releases.json" > /srv/releases.json
+test -f /srv/builds.json || \
+    curl -L "https://builds.coreos.fedoraproject.org/prod/streams/${stream}/builds/builds.json" > /srv/builds.json
+
+last_release=$(jq -r .releases[-1].version /srv/releases.json)
+target_version=$(jq -r .builds[0].id /srv/builds.json)
+
+grab-gpg-keys() {
+    # For older FCOS we had an issue where when we tried to pull the
+    # commits from the repo it would fail if we were on N-2 because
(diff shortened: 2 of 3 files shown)


tests/kola/upgrade/extended/config.bu
```diff
@@ -0,0 +1,14 @@
...
+# our oldest supported starting points (i.e. 31.20200108.3.0)
+version: 1.0.0
+storage:
+  files:
+    - path: /etc/systemd/journald.conf.d/forward-to-console.conf
+      mode: 0644
+      contents:
+        inline: |
+          # Send journal messages to the console. This is so we can see the
+          # progres even after autopkgtest-reboot-prepare has been called.
```
jlebon:
```suggestion
          # progress even after autopkgtest-reboot-prepare has been called.
```

tests/kola/upgrade/extended/test.sh
```diff
@@ -0,0 +1,157 @@
+#!/bin/bash
+## kola:
+##   # - needs-internet: to pull updates
+##   tags: "needs-internet"
+##   # Extend the timeout since a lot of updates/reboots can happen.
+##   timeoutMin: 45
+##   # Only run this test when specifically requested.
+##   requiredTag: extended-upgrade
+
+###   platforms: qemu
```
jlebon: Leftover?

tests/kola/upgrade/extended/test.sh
```diff
@@ -0,0 +1,157 @@
...
+#
+# An example invocation for this test would look like:
+
+# ```
+# cosa buildfetch --stream=next --build=34.20210904.1.0 --artifact=qemu
+# cosa decompress --build=34.20210904.1.0
+# cosa kola run --build=34.20210904.1.0 --tag extended-upgrade
+# ```
+#
+# You can montior the progress from the console and journal:
```
jlebon:
```suggestion
# You can monitor the progress from the console and journal:
```

tests/kola/upgrade/extended/test.sh
```diff
@@ -0,0 +1,157 @@
...
+        ;;
+    *) fatal "unexpected stream: $stream";;
+esac
+
+# If we have made it all the way to the last release then
+# we have one more test. We'll now rebase to the latest
+# in the compose OSTree repo.
+if vereq $version $last_release; then
+    systemctl stop zincati
+    rpm-ostree rebase fedora-compose:
```
jlebon: Shouldn't we check here that the version we staged matches `$target_version`? Or alternatively, we can make this `rpm-ostree rebase fedora-compose: $target_version`. That way rpm-ostree itself errors out if the version somehow isn't in the compose repo.

tests/kola/upgrade/extended/test.sh
```diff
@@ -0,0 +1,157 @@
...
+    /tmp/autopkgtest-reboot reboot # execute the reboot
+    sleep infinity
+fi
+
+# Watch the Zincati logs to see if it got a lead on a new update.
+# Timeout after 30s if no update. Unset pipefail since the
```
jlebon:
```suggestion
# Timeout after 60s if no update. Unset pipefail since the
```

tests/kola/upgrade/extended/test.sh
```diff
@@ -0,0 +1,157 @@
...
+# journalctl -f will give a bad exit code when grep exits early.
+set +o pipefail
+timeout 60s journalctl -b 0 -f --no-tail -u zincati.service \
+    | grep --max-count=1 'proceeding to stage it' && \
+        /tmp/autopkgtest-reboot-prepare reboot && sleep infinity
+
+# If we make it here then there was no update; let's error.
+fatal "Updating the system stalled out on version: $version"
```
jlebon: This looks cryptic. Why not

```bash
if ! timeout ... | grep ...; then
  fatal "Updating the system stalled out on version: $version"
fi

/tmp/autopkgtest-reboot-prepare reboot
sleep infinity
```

?

jlebon (commented):
This looks sane overall (thanks for writing it!). But I think ideally we'd rework this so that it works with the locally built artifacts. I've been thinking about merging the kola upgrade test into an external test, which is relevant here.

One thing we could do is add a new `startingImage` YAML key, which can take e.g. `previous` (for the current upgrade test) and `first` (for this one). Then kola would implement the logic to resolve the version and fetch the image to boot (it already has most of the code for this). Then the external test could use `injectContainer: true` to have access to the OCI archive.

tests/kola/upgrade/extended/test.sh
dustymabe: yep

tests/kola/upgrade/extended/test.sh
dustymabe: will change it so that it's not inaccurate.

tests/kola/upgrade/extended/test.sh
dustymabe: I didn't think what I had was too bad, but I reworked it a bit now

tests/kola/upgrade/extended/test.sh
dustymabe: applied

dustymabe:
> This looks sane overall (thanks for writing it!).

No problem.

> But I think ideally we'd rework this so that it works with the locally built artifacts. I've been thinking about merging the kola upgrade test into an external test, which is relevant here.

I support something better than this, but probably can't invest much time in it for now.

> 
> One thing we could do is add a new `startingImage` YAML key, which can take e.g. `previous` (for the current upgrade test) and `first` (for this one). Then kola would implement the logic to resolve the version and fetch the image to boot (it already has most of the code for this). Then the external test could use `injectContainer: true` to have access to the OCI archive.

👍

tests/kola/upgrade/extended/test.sh
```diff
@@ -0,0 +1,161 @@
...
+    # For older FCOS we had an issue where when we tried to pull the
+    # commits from the repo it would fail if we were on N-2 because
+    # the newer commits would be signed with a key the old OS didn't
+    # know anything about. We applied a workaround in newer releases,
+    # so this workaround should be limited to zincati older than v0.0.24
+    # https://github.com/coreos/fedora-coreos-tracker/issues/749
+    max_version=${target_version:0:2} # i.e. 36, 37, 38, etc..
+    max_version=38 # hardcode for now because `next` went from
+                   # 37.20230303.1.0 -> 38.20230310.1.0 -> 37.20230303.1.1.
+                   # Can drop once a new 38 commit is at the tip of `next`.
```
dustymabe: Since we're about to run the `next` build we can just hold this PR for 4-12 hours and I can drop this hack/workaround (I think).

jlebon (dismissed):

tests/kola/upgrade/extended/test.sh
jlebon: This is still there, did you mean to drop it?

tests/kola/upgrade/extended/test.sh
```diff
@@ -0,0 +1,160 @@
...
+if vereq $version $last_release; then
+    systemctl stop zincati
+    rpm-ostree rebase fedora-compose: $target_version
+    /tmp/autopkgtest-reboot reboot # execute the reboot
+    sleep infinity
+fi
+
+# Watch the Zincati logs to see if it got a lead on a new update.
+# Timeout after 400s if no update. Zincati retries after 5 minutes
+# if it fails (DNS, network blip, Server side error) on first checkin.
```
jlebon: I think https://github.com/coreos/coreos-assembler/blob/3715f622e5a2e253bf318e1a2ad54c2726807262/mantle/kola/tests/upgrade/basic.go#L101-L105 would help here. (See https://github.com/coreos/zincati/issues/203 for more info.)

jlebon (dismissed):

tests/kola/upgrade/extended/test.sh
dustymabe: updated!

jlebon (approved):

PR #2311, opened 2023-03-22

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,16 +9,6 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
-  coreos-installer:
-    evr: 0.17.0-1.fc37
-    metadata:
-      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-d6f3402ea1
-      type: fast-track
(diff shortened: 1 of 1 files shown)


bgilbert (dismissed):

bgilbert:
CI failing with:

```
error: Packages not found: nmstate
```

Do we have to do something special to tag nmstate into the pool?

jlebon:
> CI failing with:
> 
> ```
> error: Packages not found: nmstate
> ```
> 
> Do we have to do something special to tag nmstate into the pool?

Nothing special. The issue is that nmstate isn't in the base lockfiles yet so dropping it from the overrides lockfile here means it's now completely unlocked and `cosa fetch --strict` doesn't like that. Ideally https://github.com/coreos/fedora-coreos-config/pull/2269 should've added nmstate to the base lockfiles. This will be fixed by the next bump-lockfile run.

dustymabe (approved):
LGTM

bgilbert:
> Ideally #2269 should've added nmstate to the base lockfiles.

Oof.  It originally did, and I thought it still did at merge time, but apparently that was dropped from the PR and I didn't notice.  Apologies for the noise.

PR #2316, opened 2023-03-22
denylist: bump snooze for ext.config.kdump.crash on aarch64

dustymabe:
This is still a problem. See
https://github.com/coreos/fedora-coreos-tracker/issues/1430#issuecomment-1479842542


dustymabe:
will skip CI on this one since it won't test this change.

jlebon (approved):

PR #2315, opened 2023-03-22
[testing] tree: promote changes from testing-devel at a71bd956e3352cc1c4df564ae4a1ac2fa937aace

coreosbot-releng:
Triggered by @prestist in https://github.com/coreos/fedora-coreos-streams/issues/662.

PR #2317, opened 2023-03-22
Revert "overlay/15fcos: upgrade bootloader on aarch64 at boot"

dustymabe:
This reverts commit 8ce6fd6827b06e3c462ddc8c18d2fe1bdc7f5e86.

The testing-devel promotion with this has already happened in https://github.com/coreos/fedora-coreos-config/pull/2315 and we have already shipped a `next` with the unit, so we can drop it now (before we execute the `next-devel`->`next` promotion) to prevent it from shipping in more than one release per stream.

We can do this because we have barriers.

Full context in https://github.com/coreos/fedora-coreos-tracker/issues/1441


dustymabe:
skipping CI on this because I need it in `next-devel` soon so I can run `bump-lockfile` (which will do a full CI run on all arches) to get a lockfile bump before we promote `next-devel`->`next` this cycle.

jlebon (approved):

PR #2318, opened 2023-03-22
[next-devel] overrides: fast-track dracut-059-2.fc38

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,22 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  dracut:
+    evr: 059-2.fc38
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-e8ca690ff3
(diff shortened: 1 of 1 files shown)


dustymabe:
Will rebase this once the bump-lockfile run finishes (should make CI pass).

jlebon:
Hmm, this is fast-tracking it right before re-releasing f38. I'd personally prefer we skip it for this release and land it right after the promotion.

dustymabe:
The only reason I think it'd be nice to get it in here is because this is probably what we'll be executing tests for test day on.

I'm cool with leaving it until next time too.

Either way this PR should be able to be merged, we'll just wait until after promotion if we don't want it to go into the release we're going to build today.

jlebon:
> I'm cool with leaving it until next time too.

:+1: I'm not strongly opposed either. Just that this dracut release looks like it has a lot of changes, and given our many integration points with it, I'd prefer to have it bake more. The Bodhi update is still fresh and hasn't really seen any testing yet it looks like.

> Either way this PR should be able to be merged, we'll just wait until after promotion if we don't want it to go into the release we're going to build today.

Yeah, that's what I mean but it likely wasn't clear enough. :) It'll also allow us to provide negative feedback on the Bodhi update if needed.

dustymabe:
rebased

jlebon (approved):

PR #2320, opened 2023-03-22
40ignition-ostree: add autosave-xfs transposefs unit

### LINKED ISSUE #1183: coreos autoinstall creates huge number of xfs allocation groups
**Describe the bug**
coreos resizing the filesystem during the install process results in a pathological amount of allocation groups. On a large disk (10TB) the agcount is >20k and the system is not bootable, with the initial mount timing out.

**Reproduction steps**
1. autoinstall (--dest-ignition/--dest-device) coreos on a VM with a 10TB /dev/vda
2. the installer runs succesfully, taking a while to do some resizing
3. boot and reboot, mounting xfs times out, not bootable

**Expected behavior**
System boots normally after install.

**Actual behavior**
https://access.redhat.com/solutions/5587281 describes this exact issue. 
**System details**
 - Bare Metal/QEMU/AWS/GCP/etc.
  proxmox
 - Fedora CoreOS version 
  next

**Ignition config**
nothing filesystem related

**Additional information**
image

I don't think you need a 10tb disk to see the problem. The same image provisioned on a 200gb disk has an agcount of 430, which still seems too high.

jlebon:
Add a new transposefs unit: `autosave-xfs`. This unit runs after
`ignition-disks` and `ignition-ostree-growfs,` but before the `restore`
transposefs unit.

If the XFS root was grown, it checks if the allocation group count
(agcount) is within a reasonable amount (128 is chosen here). If
it isn't, it saves the rootfs and reformats the filesystem. The
`restore` unit will then restore it as usual. In the case of in-place
reprovisioning like LUKS (i.e. where the partition table isn't modified
by the Ignition config), the rootfs is still saved only once.

Ideally, instead of adding a new transposefs unit, we would make it
part of the initial `save` unit. But at that point, there's no way to
tell whether we should autosave without gazing even more deeply into the
Ignition config. We also don't want to unconditionally save the rootfs
when we may not need it.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1183


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -124,6 +127,62 @@ mount_and_save_filesystem_by_label() {
...
+        echo "Filesystem is not XFS (found $fstype); skipping"
+        return 1
+    fi
+    local agcount
+    eval $(xfs_info "${root_part}" | grep -o 'agcount=[0-9]*')
+    # Semi-arbitrarily chosen: this is roughly ~64G currently (based on initial
+    # ag sizing at build time) which seems like a good rootfs size at which to
+    # discriminate between "throwaway/short-lived systems" and "long-running
+    # workload systems". It's not like XFS performance is way worse at 128.
+    if [ "$agcount" -lt 128 ]; then
```
jlebon, line 147: @sandeen, ended up going with 128 as a threshold for this.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
```diff
@@ -28,6 +28,8 @@ install() {
         systemd-sysusers \
         systemd-tmpfiles \
         sort \
+        xfs_info \
+        xfs_spaceman \
```
cgwalters, line 32: :astronaut:

cgwalters (dismissed):
Gave this a read, looks good.

jlebon:
Rebased for conflicts now! With FCOS releases freshly out, I think we can get this in.

tests/kola/disks/growfs
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+
```
dustymabe, line 4: Would be nice to get a description up here for some context.

tests/kola/root-reprovision/autosave-xfs/test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+## kola:
+##   # This test reprovisions the rootfs automatically.
+##   tags: "platform-independent reprovision"
+##   # Trigger automatic XFS reprovisioning
+##   minDisk: 64
```
dustymabe, line 6: I suspect because of thin provisioning we won't hit an issue with using a large disk here, but is there any case where you could see this causing us to run out of space in our pipeline?

tests/kola/root-reprovision/luks/test.sh
```diff
@@ -14,57 +14,13 @@ set -xeuo pipefail
...
-    fatal "missing allow_discards in root DM table: ${table}"
-fi
-if ! grep -q no_read_workqueue <<< "${table}"; then
-    fatal "missing no_read_workqueue in root DM table: ${table}"
+# check that we didn't run automatic XFS reprovisioning
+if [ -z "${AUTOPKGTEST_REBOOT_MARK:-}" ]; then
+    if [ -f /run/ignition-ostree-autosaved-xfs.stamp ]; then
+        fatal "unexpected autosaved XFS"
+    fi
+    ok no autosaved XFS on large disk
```
dustymabe: Optional:

```suggestion
    ok "no autosaved XFS on large disk"
```

tests/kola/root-reprovision/luks/data/luks-test.sh
dustymabe, line 1: Feels like this might be better as a shared library (with a nice comment at the top) instead of just a script that gets run. They are basically the same thing, but at least for me it's easier to understand. For example we have [ntplib.sh](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/ntp/data/ntplib.sh) that's called by both [`ntp/timesyncd`](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/ntp/timesyncd/dhcp-propagation/test.sh) and [`ntp/chrony`](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/ntp/chrony/dhcp-propagation) tests.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -5,17 +5,40 @@ set -euo pipefail
...
+    # We've seen this on Fedora kernels with debug enabled (common in `rawhide`).
+    # See https://github.com/coreos/fedora-coreos-tracker/issues/1092
+    actual_dev=$(realpath "/dev/disk/by-label/$label" || :)
+    if [ "$actual_dev" != "$expected_dev" ]; then
+        echo "Expected /dev/disk/by-label/$label to point to $expected_dev, but points to $actual_dev; triggering udev"
+        udevadm trigger --settle "$expected_dev"
+    fi
+}
+ignition_cfg=/run/ignition.json
+expected_dev=$(jq -r '.storage?.filesystems? // [] | map(select(.label == "root")) | .[0].device // ""' "${ignition_cfg}")
```
dustymabe, line 29: a lot of `jq` here :) - The `man` page is my friend.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -96,8 +96,10 @@ mount_and_restore_filesystem_by_label() {
     local mountpoint=$1; shift
     local saved_fs=$1; shift
     local new_dev
-    new_dev=$(jq -r "$(query_fslabel "${label}") | .[0].device" "${ignition_cfg}")
-    udev_trigger_on_label_mismatch "${label}" "${new_dev}"
+    new_dev=$(jq -r "$(query_fslabel "${label}") | .[0].device // \"\"" "${ignition_cfg}")
+    if [ -n "${new_dev}" ]; then
+        udev_trigger_on_label_mismatch "${label}" "${new_dev}"
+    fi
```
dustymabe: Might be worth adding the context from the commit message here.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -219,6 +240,22 @@ case "${1:-}" in
             mkdir "${saved_prep}"
         fi
         ;;
+    autosave-xfs)
+        if should_autosave_rootfs; then
```
dustymabe: I'm trying to determine if we should try to handle errors here. For example, because this script is `set -euo pipefail` we could get a return code from the call that isn't `0` or `1`. Should we behave any different?

dustymabe (commented):
Mostly LGTM - some suggestions

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jlebon: To be clear, this is also copied from `transposefs.sh`. I kept it stuck to the `udev_trigger_on_label_mismatch` function to try to make that clear, but added another comment instead.

tests/kola/root-reprovision/autosave-xfs/test.sh
jlebon: Not that I can think of. But I guess we'll find out soon enough and will adapt if somehow this turns out to be an issue. :)

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: Good call, will try to tweak this a bit.

tests/kola/root-reprovision/luks/data/luks-test.sh
jlebon: I initially did want to structure it that way, but it felt funny to have the library function also do a reboot (and I didn't want to duplicate the reboot stuff in the caller scripts), so it just felt more natural to leave it as a straight up sourced file. We have a similar pattern in ostree at least (see e.g. https://github.com/ostreedev/ostree/blob/main/tests/basic-test.sh).

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
dustymabe: Sorry. This was just a comment. Wasn't necessarily asking for a change.

tests/kola/root-reprovision/luks/data/luks-test.sh
dustymabe: 👍

dustymabe: Maybe just add a comment line at the top of the file for some context.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
dustymabe: Sounds good. I'm not sure if we'd do anything different if we get a different error, but it is at least worth thinking about.

jlebon:
Updated for comments!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -202,6 +244,23 @@ case "${1:-}" in
             mkdir "${saved_prep}"
         fi
         ;;
+    autosave-xfs)
+        should_autosave=$(should_autosave_rootfs)
```
dustymabe, line 248: so (by design) now the script will exit here if an unexpected error occurs (because of `set -euo pipefail`) and that is how we'll handle the cases we weren't handling before?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -202,6 +244,23 @@ case "${1:-}" in
...
+        if [ "${should_autosave}" = "1" ]; then
```
dustymabe, line 249:
```suggestion
        if [ "${should_autosave}" == "1" ]; then
```

It always throws me off that a single `=` works.

You don't need to change this, I'm just calling it out.

dustymabe (approved):
LGTM

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jlebon: Yup, exactly. It's no longer part of an if-statement, so the errexit should be respected.

cgwalters:
Downstream bug https://issues.redhat.com/browse/OCPBUGS-16157

cgwalters:
Further fallout from this in https://issues.redhat.com/browse/OCPBUGS-16724

cgwalters:
Followup PR in https://github.com/coreos/fedora-coreos-config/pull/2565

PR #2322, opened 2023-03-23
Remove the fedora-rawhide-nodebug-kernel repo from fedora-rawhide.repo

dustymabe:
This was originally added in ff9bf0e. There is no longer any need to pull from the separate repo because the policy has changed [1] and now forced debug won't be on for any kernels built and submitted to rawhide.

[1] https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/MDJUE324EVG7CVG5UCTZCXPCOMOWMX5M/


jlebon (approved):
:tada:

PR #2327, opened 2023-03-25
overrides: fast-track xfsprogs-6.1.0-2.fc37

coreosbot-releng:


bgilbert:
Update for F37 to re-enable creation of small filesystems.  This is F37-only, so there's no fast-track for `next`.

dustymabe (approved):
LGTM

PR #2331, opened 2023-03-28
[next-devel] overrides: fast-track openssh-9.0p1-14

travier:
See: https://bodhi.fedoraproject.org/updates/FEDORA-2023-bb12d0efe6
See: https://github.com/coreos/fedora-coreos-tracker/issues/1394
See: https://bugzilla.redhat.com/show_bug.cgi?id=2172956

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,9 +9,45 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  fedora-release-common:
+    evra: 38-0.32.noarch
+    metadata:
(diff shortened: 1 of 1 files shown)


travier:
Note that we may want to not merge this one and wait for the update to land normally.

This is mostly for testing it in FCOS to make sure it works.

dustymabe:
> Note that we may want to not merge this one and wait for the update to land normally.

Yeah. This one wasn't super urgent for us because the previous update covered us already (because of how we handle presets), but I'm OK with a fast-track here too.

dustymabe (approved):
LGTM

PR #2336, opened 2023-03-29
[rhcos-4.13] overlay.d: create new 30gcp-udev-rules overlay

ravanelli:
- This PR injects/updates the udev rules from google-guest-config package into our overlay.
 - It works as an workaround for the BZ #2176212. The longer term solution is to add these packages as part of the Fedora/RHEL itself, so we won't need to maintain it, as we can entirely drop it.

For more info:
https://github.com/GoogleCloudPlatform/guest-configs/tree/master/src/lib/udev


dustymabe:
A few things:

- Let's rename the commit to not mention the bug but rather say what we are doing. We can mention the bug in the commit message.
- Do you mind adding a first commit that breaks out the existing udev rule into it's own module (i.e. look at `25azure-udev-rules`). This helps us in the future evaluate our "layers" and work to eliminate them over time.
- In the commit message of the second commit can you mention what commit hash you used when you synced over these 3 files from https://github.com/GoogleCloudPlatform/guest-configs ? I think it will be useful for historical purposes. We could/should also link to https://bugzilla.redhat.com/show_bug.cgi?id=2182865 as an opportunity for us to drop this when the new subpackage exists that we can pull in.

jlebon:
>     * In the commit message of the second commit can you mention what commit hash you used when you synced over these 3 files from [GoogleCloudPlatform/guest-configs](https://github.com/GoogleCloudPlatform/guest-configs) ?

I would go further and include that information directly as comments at the top of those files.

dustymabe:
> I would go further and include that information directly as comments at the top of those files.

I wouldn't - since the files are currently diff-able with what is upstream, but I would include it in the README for the module. i.e. look at what we did for azure: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/README.md#25-azure-udev-rules

jlebon:
> > I would go further and include that information directly as comments at the top of those files.
> 
> I wouldn't - since the files are currently diff-able with what is upstream, but I would include it in the README for the module. i.e. look at what we did for azure: [`testing-devel`/overlay.d/README.md#25-azure-udev-rules](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/README.md?rgh-link-date=2023-03-30T14%3A29%3A35Z#25-azure-udev-rules)

Personally, when copy/pasting stuff like this directly from other places, I want to be really clear that happened and so document it at the top of the file. It doesn't have to be the full context; just a "Copied from $permalink" is enough. Having that provenance be inline is helpful not only to us but in this case also to users perusing those files on the host too.

dustymabe:
> Personally, when copy/pasting stuff like this directly from other places, I want to be really clear that happened and so document it at the top of the file. It doesn't have to be the full context; just a "Copied from $permalink" is enough. Having that provenance be inline is helpful not only to us but in this case also to users perusing those files on the host too.

I think that's fair. Maybe we can do both?

dustymabe:
Nice commit messages @ravanelli!

I think we need to still update the [README](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/README.md) and also add a line at the top of the copied files mentioning where exactly they were copied from (permalink to github file would be nice).

ravanelli:
> Nice commit messages @ravanelli!
> 
> I think we need to still update the [README](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/README.md) and also add a line at the top of the copied files mentioning where exactly they were copied from (permalink to github file would be nice).

Done =)

ravanelli:
The CI still failing, seems we need a backport to rhcos-413 from: https://github.com/coreos/fedora-coreos-config/commit/b0ffe7c29d74d2d520a816ee4e194be13ff9fcce

overlay.d/30gcp-udev-rules/usr/lib/udev/google_nvme_id
```diff
@@ -0,0 +1,308 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Used to generate symlinks for NVMe devices (both local SSD and
+# persistent disk) using the disk names reported by the metadata server.
+
+# ATTENTION: It is a copy from https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/lib/udev/google_nvme_id
+# This file is synced with the source using the last update: Latest commit 3841945 on Feb 17 2023
```
dustymabe: For these let's move this directly under the `#!/bin/bash` above and also just use the permalink which is self explanatory:
```suggestion
# ATTENTION: This is a copy from https://github.com/GoogleCloudPlatform/guest-configs/blob/18fbc050b135461879e631a5ec2dd2cd7259d8e2/src/lib/udev/google_nvme_id
```

overlay.d/README.md
```diff
@@ -66,3 +66,12 @@ bits to include the rules in the initramfs too.
 [1] https://github.com/coreos/fedora-coreos-tracker/issues/1383
 [2] https://github.com/Azure/WALinuxAgent/pull/1622
 [3] https://src.fedoraproject.org/rpms/WALinuxAgent/pull-request/4
+
+
+30gcp-udev-rules
+-------------------
+
+Add udev rules and scripts needed from google-guest-configs [1] for disk
+configuration in GCP, such as local SSD controllers (nvme and scsi).
```
dustymabe, line 75: let's also mention here the open BZ to get this into the RPM and that we can drop this and just include the RPM when the BZ is fixed.

jlebon (approved):
LGTM!

dustymabe (approved):
LGTM - The only other thing that might be nice is adding a test for this (only runs on GCP) that validates some of the rules are acting appropriately.

jlebon:
Wait, I just noticed this was targeting `rhcos-4.13`. I think we want this in FCOS proper too though. Can you open the same PR against `testing-devel` too?

dustymabe:
oh yeah. I thought this was targetting the `testing-devel` branch.

ravanelli:
> LGTM - The only other thing that might be nice is adding a test for this (only runs on GCP) that validates some of the rules are acting appropriately.

I created a new issue to work on it:
https://github.com/coreos/fedora-coreos-tracker/issues/1457

ravanelli:
> Wait, I just noticed this was targeting `rhcos-4.13`. I think we want this in FCOS proper too though. Can you open the same PR against `testing-devel` too?

A my bad, probably opened it first against 413 since the issue was in this branch.

travier:
New PR in https://github.com/coreos/fedora-coreos-config/pull/2350

PR #2344, opened 2023-04-03
tests: add "ext.config.secex.reboot" test

nikita-dubrovskii:
This test requires valid hostkey to succeed:
```
export COREOS_ASSEMBLER_CONTAINER_RUNTIME_ARGS="-v $(realpath /path/to/HKD-*.crt):/secex-hostkey"
cosa kola run --qemu-secex --tag secex --qemu-secex-hostkey /secex-hostkey
```

### DIFF
--- a/tests/kola/secex/reboot/test.sh
+++ b/tests/kola/secex/reboot/test.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+## kola:
+##   # This test verifies the qemu-secex image reboots with SE enabled. It also
+##   # implicitly tests Ignition config decryption. We don't run it by default
+##   # because it requires running with `--qemu-secex --qemu-secex-hostkey HKD-<serial>.crt`.
+##   architectures: s390x
+##   platforms: qemu
+##   requiredTag: secex
+##   timeoutMin: 3
+
+set -xeuo pipefail
+
+grep -q 1 /sys/firmware/uv/prot_virt_guest
+
+case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+"")
+    rpm-ostree kargs --append secex_test
+    /tmp/autopkgtest-reboot rebooted
+    ;;
+rebooted)
+    grep -q rd.luks.name=$(cryptsetup luksUUID /dev/disk/by-label/crypt_rootfs)=root /proc/cmdline
+    grep -q secex_test /proc/cmdline
+    ;;
+*)
+    echo "unexpected mark: ${AUTOPKGTEST_REBOOT_MARK}";
+    exit 1
+    ;;
+esac


jlebon:
Instead of creating a new test, WDYT about enhancing the existing `secex/ensure` test to now also reboot?

nikita-dubrovskii:
I did this intentionally: it may happen that builder's hostkey (due to upgrade or hw-repair) would change and therefore we would end up with only 1 broken secex-test. on my dev-builder machine hostkey from time to time gets changed

jlebon:
> I did this intentionally: it may happen that builder's hostkey (due to upgrade or hw-repair) would change and therefore we would end up with only 1 broken secex-test. on my dev-builder machine hostkey from time to time gets changed

I wasn't aware this could happen. Doesn't this break any Secure Execution workloads though? Or is the old hostkey still around for decrypting the sdboot blob?

nikita-dubrovskii:
> I wasn't aware this could happen. Doesn't this break any Secure Execution workloads though? Or is the old hostkey still around for decrypting the sdboot blob?

image signed with universal key would start, but reboot won't survive.

jlebon (approved):
Were you able to verify this works well on our s390x builder with the host key Jan added?

jschintag:
I tested it on the builder. My only comment is if we should use `/data.secex/hostkeys/HKD-*.crt` as the default value?
```
[coreos-assembler]$ cosa kola run --qemu-secex --tag secex --qemu-secex-hostkey /data.secex/hostkeys/HKD-*.crt --build 414.92.202304042047-0
kola -p qemu-unpriv --build 414.92.202304042047-0 run --qemu-secex --tag secex --qemu-secex-hostkey /data.secex/hostkeys/HKD-*.crt --output-dir tmp/kola
⚠️  Skipping kola test pattern "skip-console-warnings":
  👉 https://bugzilla.redhat.com/show_bug.cgi?id=2164765
⚠️  Skipping kola test pattern "ext.config.rpm-ostree.replace-rt-kernel":
  👉 https://github.com/openshift/os/issues/1099
⚠️  Skipping kola test pattern "rhcos.network.*":
  👉 https://github.com/coreos/coreos-assembler/issues/3376
🕒 Snoozing kola test pattern "ext.config.shared.networking.nmstate.*" until Apr 17 2023:
  👉 https://github.com/openshift/os/issues/1228
=== RUN   ext.config.shared.secex.reboot
=== RUN   ext.config.shared.secex.ensure
--- PASS: ext.config.shared.secex.reboot (145.58s)
--- PASS: ext.config.shared.secex.ensure (127.58s)
PASS, output in tmp/kola
```
and
```
[coreos-assembler]$ cat tmp/kola/ext.config.shared.secex.reboot/7a828d27-f0fb-4923-a974-5ec684de9b1e/ext.config.shared.secex.reboot.txt
Apr 05 08:38:11 qemu0 kola-runext-test.sh[3158]: + grep -q 1 /sys/firmware/uv/prot_virt_guest
Apr 05 08:38:11 qemu0 kola-runext-test.sh[3158]: + case "${AUTOPKGTEST_REBOOT_MARK:-}" in
Apr 05 08:38:11 qemu0 kola-runext-test.sh[3158]: + rpm-ostree kargs --append secex_test
Apr 05 08:38:13 qemu0 kola-runext-test.sh[3160]: Staging deployment...done
Apr 05 08:38:13 qemu0 kola-runext-test.sh[3160]: Freed: 2.0 MB (pkgcache branches: 0)
Apr 05 08:38:13 qemu0 kola-runext-test.sh[3160]: Changes queued for next boot. Run "systemctl reboot" to start a reboot
Apr 05 08:38:13 qemu0 kola-runext-test.sh[3158]: + /tmp/autopkgtest-reboot rebooted
Apr 05 08:38:13 qemu0 kola-runext-test.sh[3189]: + /var/home/core/kolet reboot-request rebooted
Apr 05 08:38:13 qemu0 kola-runext-test.sh[3189]: + reboot
-- Boot d16a74c99838499d9fc6f0100d57887b --
Apr 05 08:38:31 qemu0 kola-runext-test.sh[1478]: + grep -q 1 /sys/firmware/uv/prot_virt_guest
Apr 05 08:38:31 qemu0 kola-runext-test.sh[1478]: + case "${AUTOPKGTEST_REBOOT_MARK:-}" in
Apr 05 08:38:31 qemu0 kola-runext-test.sh[1480]: ++ cryptsetup luksUUID /dev/disk/by-label/crypt_rootfs
Apr 05 08:38:31 qemu0 kola-runext-test.sh[1478]: + grep -q rd.luks.name=b4c8843d-ff35-4931-b931-2b83016ee5b2=root /proc/cmdline
Apr 05 08:38:31 qemu0 kola-runext-test.sh[1478]: + grep -q secex_test /proc/cmdline
```

nikita-dubrovskii:
> I tested it on the builder. My only comment is if we should use `/data.secex/hostkeys/HKD-*.crt` as the default value?

Thx! Do you mean hardcore this path withnin `coreos-assembler` ? hmm, i would keep explicit

jschintag:
> Thx! Do you mean hardcore this path withnin `coreos-assembler` ? hmm, i would keep explicit

Yes, I would propose:
1. If `--qemu-secex-hostkey` is set, use it and error out if no file is found
2. If not set, check if `/data.secex/hostkeys/HKD-*.crt` exists, if not use fallback dummy key.

This would be in line with how the `genprotimgvm` is found. (Except fallback, as there is no fallback for vm)

jlebon:
> Yes, I would propose:
> 
>   1. If `--qemu-secex-hostkey` is set, use it and error out if no file is found
>   2. If not set, check if `/data.secex/hostkeys/HKD-*.crt` exists, if not use fallback dummy key.
> 
> This would be in line with how the `genprotimgvm` is found. (Except fallback, as there is no fallback for vm)

Not opposed if @nikita-dubrovskii wants to whip that up, but also fine to just hardcode it in the pipeline as part of [the proposed patch](https://github.com/coreos/fedora-coreos-pipeline/pull/849) for now since we're already special-casing s390x there anyway. I agree matching the existing semantics established by the `genprotimgvm` bits would be nice. (But long-term, it'd be nice to figure out a cleaner way to wire all this stuff through.)

nikita-dubrovskii:
> Yes, I would propose:
> 
>     1. If `--qemu-secex-hostkey` is set, use it and error out if no file is found

If specified key doesn't exist , than test fails

>     2. If not set, check if `/data.secex/hostkeys/HKD-*.crt` exists, if not use fallback dummy key.
> This would be in line with how the `genprotimgvm` is found. (Except fallback, as there is no fallback for vm)

I don't want to hide the key. Now the SecEx is fresh, but after couple of months (or when new people join) it'll be difficult to figure out what's going on here. So key should be set explicitly

PR #2350, opened 2023-04-05
overlay.d: create new 30gcp-udev-rules overlay

ravanelli:
This is a cherry-pick from https://github.com/coreos/fedora-coreos-config/pull/2336, where I merged first in 413 by mistake


travier:
I've just realized that this PR and the parent PR does not change the list of overlays in the manifest and is thus not taken into account.

travier:
https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/fedora-coreos-base.yaml#L18

ravanelli:
> https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/fedora-coreos-base.yaml#L18

Good catch! Fixing it

dustymabe (approved):
LGTM

travier:
Could we have at least a manual test that this works? A basic automated test would be great.

ravanelli:
> Could we have at least a manual test that this works? A basic automated test would be great.

The manually check done was to validate if the udev files exist and if the symlinks created  were different from before for the nvme disk using a GCP confidencial /non confidencial VM

ravanelli:
@travier  Since people is waiting for this PR I think make sense for us go head and merge it without the test, since we already validate it manually. I will also cherry-pick the missing part of the manifest to 4.13

marmijo (approved):
LGTM

dustymabe:
> @travier Since people is waiting for this PR I think make sense for us go head and merge it without the test, since we already validate it manually. I will also cherry-pick the missing part of the manifest to 4.13

Agree. Let's get this in. @ravanelli can you work on https://github.com/coreos/fedora-coreos-tracker/issues/1457 soon?

dustymabe (approved):
LGTM

PR #2354, opened 2023-04-07
denylist: snooze autosave-xfs and luks.autosave-xfs for aarch64 and ppc64le

marmijo:
These tests are failing and blocking FCOS pipeline multi-arch builds. 
See: https://github.com/coreos/fedora-coreos-tracker/issues/1458

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -34,3 +34,15 @@
   snooze: 2023-04-12
   arches:
     - aarch64
+- pattern: ext.config.root-reprovision.autosave-xfs
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1458
+  snooze: 2023-04-21
+  arches:
+    - aarch64
+    - ppc64le
+- pattern: ext.config.root-reprovision.luks.autosave-xfs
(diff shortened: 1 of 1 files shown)


kola-denylist.yaml
```diff
@@ -34,3 +34,13 @@
...
+- pattern: ext.config.root-reprovision.autosave-xfs
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1458
+  arches:
+    - aarch64
+    - ppc64le
+- pattern: ext.config.root-reprovision.luks.autosave-xfs
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1458
+  arches:
+    - aarch64
+    - ppc64le
```
bgilbert: Missing trailing newline.

ashcrow (dismissed):
:+1:

bgilbert (commented):
Should we snooze these for investigation rather than denying them?

marmijo:
Snoozing is probably a better route. These are new tests, so we can snooze for a bit to investigate. I'll make that change.

kola-denylist.yaml
marmijo: Updated

bgilbert (approved):

PR #2355, opened 2023-04-10
coreos-teardown-initramfs: use NetworkManager to teardown networking

### LINKED ISSUE #1459: Use NetworkManager to tear down networking on Ignition boot
https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh#L190

In thinking more about the whole service that effectively does `for i in $(ls /sys/class/net) do; ip link set $i down; done`. I don't see why this is needed, wouldn't in effect the same thing be accomplished if simply the networking service were stopped?

dustymabe:
Now that NetworkManager is running as a systemd service in the initramfs (in both Fedora and RHEL9+) we can use NetworkManager itself to take down networking in the initramfs. Let's do that to simplify our life a bit.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1459

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -168,55 +168,15 @@ propagate_initramfs_hostname() {
     fi
 }
 
-down_interface() {
-    echo "info: taking down network device: $1"
-    # On recommendation from the NM team let's try to delete the device
-    # first and if that doesn't work then set it to down and flush any
-    # associated addresses. Deleting virtual devices (bonds, teams, bridges,
-    # ip-tunnels, etc) will clean up any associated kernel resources. A real
-    # device can't be deleted so that will fail and we'll fallback to setting
-    # it down and flushing addresses.
-    if ! ip link delete $1; then
-        ip link set $1 down
-        ip addr flush dev $1
-    fi
-}
-
-# Iterate through the interfaces in the machine and take them down.
-# Note that in the future we would like to possibly use `nmcli` networking off`
-# for this. See the following two comments for details:
-# https://github.com/coreos/fedora-coreos-tracker/issues/394#issuecomment-599721763
-# https://github.com/coreos/fedora-coreos-tracker/issues/394#issuecomment-599746049
-down_interfaces() {
-    if ! [ -z "$(ls /sys/class/net)" ]; then
-        for f in /sys/class/net/*; do
-            interface=$(basename "$f")
-            # The `bonding_masters` entry is not a true interface and thus
-            # cannot be taken down. Also skip local loopback
-            case "$interface" in
-                "lo" | "bonding_masters")
-                    continue
-                    ;;
-            esac
-            # When we start taking down devices some other devices can
-            # start to disappear (for example vlan on top of interface).
-            # If the device we're about to take down has disappeared
-            # since the start of this loop then skip taking it down.
-            if [ ! -e "$f" ]; then
-                echo "info: skipping teardown of ${interface}; no longer exists."
-                continue
-            fi
-            down_interface $interface
-        done
-    fi
-}
-
 main() {
     # Load libraries from dracut
     load_dracut_libs
 
-    # Take down all interfaces set up in the initramfs
-    down_interfaces
+    # Take down all networking set up in the initramfs
+    if systemctl is-active --quiet nm-initrd.service; then
+        echo "info: taking down initramfs networking"
+        nmcli networking off
+    fi
 
     # Clean up all routing
     echo "info: flushing all routing"
@@ -234,10 +194,13 @@ main() {
         propagate_initramfs_networking
     fi
 
-    # Now that the configuration has been propagated (or not)
-    # clean it up so that no information from outside of the
-    # real root is passed on to NetworkManager in the real root
-    rm -rf /run/NetworkManager/
+    # Configuration has been propagated, but we can't clean up
+    # /run/NetworkManager because NM is still running. Let's drop
+    # down a tmpfiles.d snippet so that it's cleaned up first thing
+    # in the real root. Doing it this way prevents us having to write
+    # another unit to do the cleanup after this service has finished.
+    echo "R! /run/NetworkManager - - - - -" > \
+        /run/tmpfiles.d/15-teardown-initramfs-networkmanager.conf
 
     rm -f /run/udev/rules.d/80-coreos-boot-disk.rules
     rm -f /dev/disk/by-id/coreos-boot-disk
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.service
@@ -18,13 +18,13 @@ DefaultDependencies=false
 Before=ignition-mount.service
 Before=ignition-complete.target
 
-# Since we are tearing down networking we need to make sure
-# NetworkManager has been stopped, otherwise it'll be trying
-# to react to our delete/down operations. Since the ordering
-# for ExecStop is the opposite of ExecStart we need to use
-# `Before=nm-initrd.service`.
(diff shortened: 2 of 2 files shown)


dustymabe:
Marking this PR as draft because I think we want to let the F38 rebase propagate out to all channels before we get this in.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -168,55 +168,15 @@ propagate_initramfs_hostname() {
...
-    # Take down all interfaces set up in the initramfs
-    down_interfaces
+    # Take down all networking set up in the initramfs
+    if systemctl is-active --quiet nm-initrd.service; then
+        echo "info: taking down initramfs networking"
+        nmcli networking off
+    fi
 
     # Clean up all routing
     echo "info: flushing all routing"
```
dustymabe, line 182: @jtoppins - with the changes in this PR do we still need these `ip route` commands to flush the routing table? Maybe `nmcli networking off` does that for us?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -168,55 +168,15 @@ propagate_initramfs_hostname() {
...
-}
-
 main() {
     # Load libraries from dracut
     load_dracut_libs
 
-    # Take down all interfaces set up in the initramfs
-    down_interfaces
+    # Take down all networking set up in the initramfs
+    if systemctl is-active --quiet nm-initrd.service; then
```
cgwalters, line 176: This is fine, but it'd be better instead to change this unit to be `WantedBy=nm-initrd.service`.

cgwalters (dismissed):
LGTM as is; one optional.

(Also I find this service confusingly named; the name "coreos-stop-initramfs-networking.service" or so would be a lot more obvious to me)

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
jtoppins: I agree, I don't see why this if statement is needed if the dependency references for the unit were correct.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
jtoppins: I was expecting the switchroot unit to be able to declare the networking service be brought down. That way there is no one off custom networking disable call or similar in a script somewhere and systemd's state is correctly reflected across the switchroot boundary. Meaning systemd knows it needs to restart the networkmanager service once it is in the real root. By doing it this way all the knowledge of how to properly take down the service is isolated to the service's unit file.

dustymabe: > I was expecting the switchroot unit to be able to declare the networking service be brought down.

It does. Generally (some exceptions I'm sure) it takes down all processes running before switching root.

> That way there is no one off custom networking disable call or similar in a script somewhere and systemd's state is correctly reflected across the switchroot boundary. Meaning systemd knows it needs to restart the networkmanager service once it is in the real root. By doing it this way all the knowledge of how to properly take down the service is isolated to the service's unit file.

The problem here is that taking down the NetworkManager daemon doesn't actually deconfigure networking. It just stops the daemon (try it), which is not what we want. The `nmcli networking off` is needed to actually take networking down.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
dustymabe: The problem here is that this unit started as just a "take down and propagate networking" script, but over time it grew other responsibilities and we can't just run it if `nm-initrd.service` was up unless we properly split out those other pieces.

dustymabe:
> (Also I find this service confusingly named; the name "coreos-stop-initramfs-networking.service" or so would be a lot more obvious to me)

We intentionally renamed it when it grew more than just networking responsibilities. See https://github.com/coreos/ignition-dracut/pull/186

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
jtoppins: It might be time to split things.
It has been a while since I played with systemd so I had to reference its docs. When reading the bootup[1] doc, does coreos have something like `coreos-first-boot.target` that only exists in the install media initrd?

[1] https://www.freedesktop.org/software/systemd/man/bootup.html#Bootup%20in%20the%20initrd

dustymabe: we do have a mechanism to have it only run in the first boot based on `ignition-complete.target`.  See https://github.com/coreos/fedora-coreos-config/blob/eef40155245d7b66918955e30c20233bba8eb0db/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh#L58-L62

I think splitting up the responsibility part is straightforward, just have to decide if it's something worth the effort and that we want to do. Will see what the rest of the team thinks on this when they get back from Holiday.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
jtoppins: Not great but ok. Will nmcli work if the NetworkManager.service is stopped? Meaning is this ordered correctly after `network.target`? You didn't have to worry about ordering before when doing a brute force down all interfaces with iproute2, now I think you do.

dustymabe: > Not great but ok. Will nmcli work if the NetworkManager.service is stopped? Meaning is this ordered correctly after `network.target`?

The systemd unit ordering [here](https://github.com/coreos/fedora-coreos-config/blob/7c8e8b004940dc4833abe7ede4fccd3e23776394/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.service#L21-L27) (part of this PR) guarantees we run while `nm-initrd.service` and `dbus-broker.service` are still running



> You didn't have to worry about ordering before when doing a brute force down all interfaces with iproute2, now I think you do.

Actually we did need to worry about ordering, we just didn't know it until the screen share I was doing last week with you. We started caring about ordering in https://github.com/coreos/fedora-coreos-config/commit/a44c50e22f01906476d1e62144a79aa45a8fe6e9

cgwalters:
What may actually be most sustainable is to add an option to NM to teardown networking when it shuts down; then we can just inject that config into the initramfs.

jtoppins:
> What may actually be most sustainable is to add an option to NM to teardown networking when it shuts down; then we can just inject that config into the initramfs.

I agree, however, remember you only want to do this in a first boot install media. I didn't mention it in my other replies but I was thinking coreos might want to ship a customized version of `NetworkManager.service` until you can convince upstream NM that it would be a good idea to do this.

dustymabe:
Now that this week's releases have all been built I'm pulling this out of draft (queue up for next round of releases).

dustymabe:
Now that the F38 rebase is cleared out can I get another round of review on this?

cgwalters (approved):
Looks good to me, though I still think a patch to NM to add a config option to teardown networking itself before shutdown would be cleanest.

dustymabe:
> Looks good to me, 

Thanks!

> though I still think a patch to NM to add a config option to teardown networking itself before shutdown would be cleanest.

I'll circle back with the NM team on this to see if they think it would be useful. I need to sync with them on a different topic anyway.

PR #2356, opened 2023-04-10
tests/kola: upgrade.extended: get info about booted deployment

dustymabe:
I've seen a case where zincati/rpm-ostree stage an update before the kola script even runs so the information we pick up from `rpm-ostree status  --json | jq -r '.deployments[0]` is actually from the staged (not booted) deployment, which leads to us thinking we're one step further than we need to be in the process and then hitting an error when the `rpm-ostree rebase` fails because there is an rpm-ostree operation already in progress.

Let's pick up the information from the booted deployment to get rid of this race condition.


marmijo (approved):
LGTM

PR #2358, opened 2023-04-10
[rhcos-4.13] coreos-teardown-initramfs: teardown after NM has exited

dustymabe:
We haven't seen this before, but recently there was a case where in RHCOS based on RHEL 9 we saw the coreos-teardown-initramfs happening before the nm-initrd.service had been stopped during switch root. We need to guarantee that NetworkManager is stopped before we start the teardown.

Seen in https://issues.redhat.com/browse/OCPBUGS-11052

(cherry picked from commit a44c50e22f01906476d1e62144a79aa45a8fe6e9)


cgwalters (approved):
Nice job tracking this down!

dustymabe:
> Nice job tracking this down!

Thanks @cgwalters

PR #2361, opened 2023-04-11
[rawhide] overrides: pin util-linux-2.38.1-4.fc38

marmijo:
The new util_linux-2.39-0.4.fc39 package causes kola reprovision tests to fail in rawhide. Pin on util-linux-2.38.1-4.fc38 so these tests pass.
see: https://github.com/coreos/fedora-coreos-tracker/issues/1462


dustymabe (approved):
LGTM

PR #2362, opened 2023-04-11
40ignition-ostree: give filesystem type when mounting zram-based XFS

jlebon:
We should do this anyway to be more explicit, but it also happens to satisfy util-linux v2.39 in rawhide. (But likely we've exposed a bug in that new code.)

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1462

PR #2363, opened 2023-04-12
Revert "[rawhide] overrides: pin util-linux-2.38.1-4.fc38"

dustymabe:
This reverts commit dcf8f253be19c7b10f85b07c08c4a1ecfe930367.

This is no longer needed now that we specify the filesystem when mounting (see https://github.com/coreos/fedora-coreos-config/pull/2362).

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -19,31 +19,6 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1479
       type: pin
-  libblkid:
-    evr: 2.38.1-4.fc38
-    metadata:
-      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1462
-      type: pin
-  libfdisk:
-    evr: 2.38.1-4.fc38
-    metadata:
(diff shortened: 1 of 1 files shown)


bgilbert (dismissed):

dustymabe:
cc @jlebon - some of the testiso tests still failed.

jlebon:
Started digging into this. Opened https://github.com/coreos/coreos-assembler/pull/3434 to help. Only 4Kn-related tests seem to be failing. It's like none of the filesystems on the disk can be found. So possibly some regression in util-linux v2.39 + 4Kn disks.

jlebon:
Both the original issue here (related to zram) and the issue in CI here (related to 4Kn disks) will be fixed in the next util-linux rawhide build (see https://github.com/util-linux/util-linux/issues/2161#issuecomment-1513460425).

jlebon (approved):
Just rebased this and fixed the conflicts. This should pass now that we have a new enough `util-linux` build in rawhide.

dustymabe:
> Just rebased this and fixed the conflicts. This should pass now that we have a new enough `util-linux` build in rawhide.

Thanks. I had planned to do that today too when the other pin landed.

PR #2364, opened 2023-04-12
overlay.d: Add 30gcp-udev-rules dracut module

ravanelli:
- Add dracut module for 30gcp-udev-rules;
 - Udpate overlay documentation about it.

### DIFF
--- a/overlay.d/30gcp-udev-rules/usr/lib/dracut/modules.d/30gcp-udev-rules/module-setup.sh
+++ b/overlay.d/30gcp-udev-rules/usr/lib/dracut/modules.d/30gcp-udev-rules/module-setup.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/bash
+# Install 64-gce-disk-removal.rules, 65-gce-disk-naming.rules and
+# google_nvme_id into the initramfs
+
+# called by dracut
+install() {
+    inst_simple /usr/lib/udev/google_nvme_id
+    inst_multiple \
+        /usr/lib/udev/rules.d/64-gce-disk-removal.rules \
+        /usr/lib/udev/rules.d/65-gce-disk-naming.rules
+}
--- a/overlay.d/README.md
+++ b/overlay.d/README.md
@@ -74,9 +74,13 @@ bits to include the rules in the initramfs too.
 Add udev rules and scripts needed from google-guest-configs [1] for disk
 configuration in GCP, such as local SSD controllers (nvme and scsi).
 
-There is an opened BZ [2] requesting a subpackage of google-compute-engine-guest-configs
+The udev rules are also needed in the initramfs [2] and are delivered here via a dracut
+module.
+
+There is an opened BZ [3] requesting a subpackage of google-compute-engine-guest-configs
 containing only what we need. Once we get it, we can include this rpm in the
 OS (Fedora/RHEL) and drop this module entirely.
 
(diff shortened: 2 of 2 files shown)


overlay.d/30gcp-udev-rules/usr/lib/dracut/modules.d/30gcp-udev-rules/google_nvme_id
```diff
@@ -0,0 +1,308 @@
+#!/bin/bash
```
jlebon: Hmm, isn't the file already in the overlay? You should be able to `inst_simple` it directly from where it is in `/usr/lib/udev`.

overlay.d/30gcp-udev-rules/usr/lib/dracut/modules.d/30gcp-udev-rules/module-setup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# called by dracut
+check() {
+    return 0
+}
+
+# called by dracut
+depends() {
+    return 0
+}
```
jlebon: Minor: this shouldn't be necessary

overlay.d/30gcp-udev-rules/usr/lib/dracut/modules.d/30gcp-udev-rules/module-setup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# called by dracut
+install() {
+
```
jlebon: Super minor: extraneous space

cgwalters (dismissed):

overlay.d/30gcp-udev-rules/usr/lib/dracut/modules.d/30gcp-udev-rules/module-setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/bash
+# Install 64-gce-disk-removal.rules, 65-gce-disk-naming.rules and
+# google_nvme_id into the initramfs
+
+# called by dracut
+install() {
+    inst_simple google_nvme_id /usr/lib/udev/google_nvme_id
```
jlebon: Huh, I would've expected

```suggestion
    inst_simple /usr/lib/udev/google_nvme_id
```

here.

What you have clearly works (I don't see a dracut error in the compose) but I'm confused how. I would think this gets interpreted as asking dracut to install the executable `google_nvme_id` into `/usr/lib/udev/google_nvme_id` but `google_nvme_id` isn't in the `PATH` so it shouldn't find it. Can you sanity-check that the file indeed ends up in there? Alternatively, maybe cleaner to just use the simpler invocation? (But if you've already verified this, then feel free to ship as is).

jlebon (dismissed):

overlay.d/30gcp-udev-rules/usr/lib/dracut/modules.d/30gcp-udev-rules/module-setup.sh
ravanelli: Yeah it works in the end:
```
$ ls -ls /usr/lib/udev/google_nvme_id
12 -rw-r--r--. 2 root root 9186 Jan  1  1970 /usr/lib/udev/google_nvme_id
```
But let me do it this way to be simpler

jlebon (approved):

PR #2366, opened 2023-04-12
manifests: Merge EL9 manifests and overlays

travier:
Now that we only have RHCOS based on RHEL 9 and SCOS based on C9S, we can merge the manifests again.


travier:
Will be paired and tested with an RHCOS PR.

travier:
RHCOS PR: https://github.com/openshift/os/pull/1254

cgwalters (dismissed):

jlebon (dismissed):

overlay.d/05core/usr/lib/dracut/dracut.conf.d/coreos-zstd.conf
dustymabe, line 1: can drop this comment from the file now:

```
#                                                               
# We can't use this in RHCOS 8 because the kernel doesn't enable
# CONFIG_RD_ZSTD.                                               
```

travier:
Should be ready now

cgwalters (approved):

PR #2376, opened 2023-04-14
Move to Fedora Linux 38

dustymabe:
The Fedora 38 GA is to be released Tuesday April 18th.

### DIFF
--- a/ci/buildroot/Dockerfile
+++ b/ci/buildroot/Dockerfile
@@ -5,6 +5,6 @@
 #
 # This image is used by CoreOS CI to build software like
 # Ignition, rpm-ostree, ostree, coreos-installer, etc...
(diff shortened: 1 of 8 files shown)


dustymabe:
looking at the CI failure I guess we can't switch it over officially until the prod repos are in place for 38 (probably late Monday or early Tuesday).

jlebon (dismissed):

dustymabe:
Ok this is ready for final review now.

jlebon (approved):
Verified diff against `next-devel`. Looks good.

dustymabe:
CI fails due to downgrades being detected, but that's OK since we are switching to F38. The next `bump-lockfile` should update those packages so they would no longer be downgrades.

Merging.

PR #2382, opened 2023-04-17
tests/kola: fixes for upgrade test for early `next` releases

dustymabe:
```
commit 5c8aec846f44be8b77330946ecf9712fe96c689d
Date:   Mon Apr 17 13:26:11 2023 -0400

    tests/kola: fix version comparison for upgrade test
    
    I had inappropriately used 31.20200505.1.0 when it should have been
    32.20200505.1.0. Also fixed a typo in the comment URL above.

commit ea1f0faa9d56b519da1402de0a4c5434460ce19e
Date:   Mon Apr 17 13:24:25 2023 -0400

    tests/kola: workaround early next builds zincati config
    
    Early `next` releases before [1] had auto-updates disabled too. Let's
    drop that config if it exists.
    
    [1] https://github.com/coreos/fedora-coreos-config/commit/99eab318998441760cca224544fc713651f7a16d

```

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -59,6 +59,14 @@ if [ -f /etc/zincati/config.d/90-disable-auto-updates.toml ]; then
     need_zincati_restart='true'
 fi
 
+# Early `next` releases before [1] had auto-updates disabled too. Let's
+# drop that config if it exists.
+# [1] https://github.com/coreos/fedora-coreos-config/commit/99eab318998441760cca224544fc713651f7a16d
+if [ -f /etc/zincati/config.d/90-disable-on-non-production-stream.toml ]; then
+    rm -f /etc/zincati/config.d/90-disable-on-non-production-stream.toml
+    need_zincati_restart='true'
+fi
+
 get_booted_deployment_json() {
     rpm-ostree status  --json | jq -r '.deployments[] | select(.booted == true)'
 }
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -117,6 +117,16 @@ EOF
     need_zincati_restart='true'
 }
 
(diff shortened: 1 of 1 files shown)


tests/kola/upgrade/extended/test.sh
```diff
@@ -54,9 +54,20 @@ set -eux -o pipefail
...
+# Early `next` releases before [1] had auto-updates disabled too. Let's
+# drop that config if it exists.
+# [1] https://github.com/coreos/fedora-coreos-config/commit/99eab318998441760cca224544fc713651f7a16d
+if [ -f /etc/zincati/config.d/90-disable-on-non-production-stream.toml ]; then
+    rm -f /etc/zincati/config.d/90-disable-on-non-production-stream.toml
+    need_zincati_restart='true'
 fi
 
+# Do the restart if needed
+[ "${need_zincati_restart}" == "true" ] && systemctl restart zincati
```
jlebon: Doesn't this trigger `errexit` if `need_zincati_restart` is not `true`?

dustymabe: I don't think so because it's inside the `[ ]`.

dustymabe: testing seems to work fine.

dustymabe:
pushed up another commit to handle another problem related to https://github.com/coreos/fedora-coreos-tracker/issues/481

tests/kola/upgrade/extended/test.sh
jlebon: OK cool, this does work. I think it's actually because it's part of a `&&` list? `bash(1)` says about `set -e`:

> The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test following the if or elif reserved words, part of any command executed in a && or || list except the command following the final && or ||, any command in a pipeline but the last, or if the command's return value is being inverted with !.

jlebon (approved):

PR #2383, opened 2023-04-18
40ignition-ostree/ignition-ostree-growfs: Fix Secure Execution firstboot error

### LINKED ISSUE #1264: s390x/Secure Execution: Firstboot fails during `ignition-ostree-growfs.service`
The Secure Execution is unable to boot, failing at `ignition-ostree-growfs.service`.
```
Displaying logs from failed units: ignition-ostree-growfs.service
Apr 18 08:42:09 systemd[1]: Starting Ignition OSTree: Grow Root Filesystem...
Apr 18 08:42:09 ignition-ostree-growfs[2448]: mount: /sysroot: special device /dev/disk/by-label/root does not exist.
Apr 18 08:42:09 systemd[1]: ignition-ostree-growfs.service: Main process exited, code=exited, status=32/n/a
Apr 18 08:42:09 systemd[1]: ignition-ostree-growfs.service: Failed with result 'exit-code'.
Apr 18 08:42:09 systemd[1]: Failed to start Ignition OSTree: Grow Root Filesystem.

Generating "/run/initramfs/rdsosreport.txt"
```
This just started with `414.92.202304172216-0`, the build before is working. Looking at the diff, there are no relevant package diffs. The only notable difference between the builds is the f-c-c bump.
Need to investigate further.

jschintag:
Do not try to mount disk when running in Secure Execution mode.

Fixes: https://github.com/openshift/os/issues/1264

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
@@ -5,6 +5,12 @@ set -euo pipefail
 # partition, unless it determines that either the rootfs was moved or the
 # partition was already resized (e.g. via Ignition).
 
+# In the IBM Secure Execution case we use Ignition to grow and reencrypt rootfs
+# see overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+if [[ -f /run/coreos/secure-execution ]]; then
+    exit 0
+fi
+
 # This is copied from ignition-ostree-transposefs.sh.
 # Sometimes, for some reason the by-label symlinks aren't updated. Detect these
 # cases, and explicitly `udevadm trigger`.
@@ -41,12 +47,6 @@ saved_partstate=/run/ignition-ostree-rootfs-partstate.sh
 path=/sysroot
 src=/dev/disk/by-label/root
 mount "${src}" "${path}"
-
-# In the IBM Secure Execution case we use Ignition to grow and reencrypt rootfs
-# see overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
-if [[ -f /run/coreos/secure-execution ]]; then
-    exit 0
-fi
 
 if [ ! -f "${saved_partstate}" ]; then
     partition=$(realpath /dev/disk/by-label/root)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
```diff
@@ -35,19 +35,19 @@ fi
 # partition *before* ignition-disks.
 saved_partstate=/run/ignition-ostree-rootfs-partstate.sh
 
+# In the IBM Secure Execution case we use Ignition to grow and reencrypt rootfs
+# see overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+if [[ -f /run/coreos/secure-execution ]]; then
+    exit 0
+fi
+
```
jlebon: Let's move this even higher up. We don't need the `udev_trigger_on_label_mismatch` bits even. So maybe at line 7?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
```diff
@@ -3,6 +3,7 @@ Description=Ignition OSTree: Grow Root Filesystem
 DefaultDependencies=false
 ConditionKernelCommandLine=ostree
 ConditionPathExists=!/run/ostree-live
+Require=/dev/disk/by-label/root
```
jlebon: I'd like to think more on this one. Can you drop this hunk for now? Is that required for the error rate to go down to 0%, or is the secex hunk below sufficient?

jschintag: The secex part should be sufficient for that. I really only left this one here because the service does require the device to be present so it didn't hurt. But the require only really dropped the error rate for Secure Execution from 100% to 50% anyway.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.sh
jschintag: Works with line 7 as well.

jlebon (approved):
I just pushed a fixup to remove a no-longer-true part of the commit message and drop a stray hunk.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
jlebon: The reason why it wasn't clear to me why we needed this is that this service is already ordered after `ignition-ostree-uuid-root.service`, which itself is already ordered `After` the root device. So it should be redundant. But it's possible there's funky udev behaviour going on.

But also, note that the key name is `Requires`, and the device name usually needs to be encoded a-la-`systemd escape --path`. So I don't think this patch as is actually could've changed anything at the systemd level.

jschintag:
Thank you @jlebon

PR #2388, opened 2023-04-19
manifests/fedora-coreos-base: add systemd override for ssh host key migration

dustymabe:
In this case we'll override the ConditionPathExists from ssh-host-keys-migration.service [1] to force it to run every boot. We want to do this to handle the case where someone could do an upgrade->rollback->upgrade and end up locked out of their system [2].

[1] https://src.fedoraproject.org/rpms/openssh/blob/6f7c765ed4cf0e8eef664fb93b26f4ea2a14d955/f/ssh-host-keys-migration.service
[2] https://github.com/coreos/fedora-coreos-tracker/issues/1473

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -108,6 +108,22 @@ postprocess:
     set -xeuo pipefail
     rm -vrf /usr/lib/modules/*aarch64/dtb/qcom/
 
+  # Force the ssh-host-keys-migration to happen on every boot
+  # to handle cases where someone did a upgrade->rollback->upgrade
+  # See https://github.com/coreos/fedora-coreos-tracker/issues/1473
+  # We should remove this after the next barrier release.
+  - |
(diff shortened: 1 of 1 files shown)


dustymabe:
ok local tests of this seem promising

jlebon:
Is there a reason we're not doing this via an overlay instead?

dustymabe:
> Is there a reason we're not doing this via an overlay instead?

I started by just adding it to another overlay but I feel like some of this stuff gets lost in those larger overlays. I felt like it had more visibility here.

I think if I were going to move it to an overlay I'd create a new one (i.e. better visibility) and that felt a bit heavyweight for something we are planning to remove soon.

travier:
Would have preferred an overlay (harder to make mistakes) but ok

jlebon (approved):
Same, but not a strong opinion. It doesn't matter too much since it's short-lived. LGTM!

PR #2395, opened 2023-04-21
[rawhide] overrides: pin linux-firmware-20230310-148.fc39

marmijo:
The new linux-firmware-20230404-149.fc39 causes `files.validate-symlinks` kola tests to fail in rawhide. 
Pin on linux-firmware-20230310-148.fc39 so the tests pass. 

See: https://github.com/coreos/fedora-coreos-tracker/issues/1479


dustymabe (approved):
LGTM

PR #2396, opened 2023-04-23
overlay.d/20platform-chrony: accept `ec2` and `gce` platform IDs

bgilbert:
RHCOS 4.1 and 4.2 bootimages are still in use, and they produce systems that claim to be `ec2`/`gce` instead of `aws`/`gcp`.  Detect those and handle appropriately.

### DIFF
--- a/overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
+++ b/overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
@@ -83,11 +83,11 @@ case "${platform}" in
          echo 'refclock PHC /dev/ptp_hyperv poll 3 dpoll -2 offset 0'
          echo 'leapsectz right/UTC'
         ) >> "${confpath}" ;;
-    aws)
+    aws | ec2)
         (echo '# See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html'
          echo 'server 169.254.169.123 prefer iburst minpoll 4 maxpoll 4'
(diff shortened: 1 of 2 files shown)


dustymabe:
Was this a regression introduced by say https://github.com/coreos/fedora-coreos-config/pull/2271 or did we just never handle this case with the platform chrony code?

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
```diff
@@ -83,11 +83,11 @@ case "${platform}" in
          echo 'refclock PHC /dev/ptp_hyperv poll 3 dpoll -2 offset 0'
          echo 'leapsectz right/UTC'
         ) >> "${confpath}" ;;
-    aws)
+    aws|ec2)
```
dustymabe: optional: since `azure | azurestack` has a space above it would be nice if these were consistent. I don't have a preference of with space or without space but maybe just consistent.

dustymabe (dismissed):
LGTM

bgilbert:
> Was this a regression introduced by say https://github.com/coreos/fedora-coreos-config/pull/2271 or did we just never handle this case with the platform chrony code?

We never handled it.

overlay.d/20platform-chrony/usr/libexec/coreos-platform-chrony-config
bgilbert: Updated!

dustymabe (approved):
LGTM

PR #2398, opened 2023-04-24
[rawhide] kola-denylist: snooze new `ext.config.networking.nmstate.*` tests

marmijo:
The new nmstate tests are failing in rawhide. Let's snooze until we find a fix for them or disable them in FCOS.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1480

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -18,3 +18,6 @@
   snooze: 2023-05-03
   arches:
     - aarch64
+- pattern: ext.config.networking.nmstate.*
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1480
+  snooze: 2023-05-08


dustymabe (approved):
LGTM

kola-denylist.yaml
```diff
@@ -18,3 +18,6 @@
   snooze: 2023-05-03
   arches:
     - aarch64
+- pattern: ext.config.shared.networking.nmstate.*
```
dustymabe: I don't think we need the `shared` here: 

```suggestion
- pattern: ext.config.networking.nmstate.*
```

dustymabe (changes requested):

kola-denylist.yaml
marmijo: Updated!

kola-denylist.yaml
```diff
@@ -18,3 +18,6 @@
   snooze: 2023-05-03
   arches:
     - aarch64
+- pattern: ext.config.networking.nmstate.*
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1480
+  snooze: 2023-05-08
```
dustymabe, line 23: Might be good to limit the snooze to `rawhide` so we'll know if the regression is introduced in one of our prod streams.

```suggestion
  snooze: 2023-05-08
  streams:
    - rawhide
```

dustymabe (approved):

gursewak1997:
We need to open another PR which would target `testing-devel`. Looks like these changes are gone since we merged it to `rawhide`.
Edit: Let's just wait on the fix https://github.com/coreos/fedora-coreos-tracker/issues/1480#issuecomment-1521236302

PR #2402, opened 2023-04-26
kola/validate-symlinks: skip `/usr/lib/firmware` symlinks

marmijo:
A recent rawhide package of linux-firmware caused the validate-symlinks kola tests to fail. Let's modify the broken-symlinks list to skip all of `/usr/lib/firmware` so we can unblock FCOS rawhide builds.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1479

### DIFF
--- a/tests/kola/files/validate-symlinks
+++ b/tests/kola/files/validate-symlinks
@@ -11,14 +11,16 @@ set -xeuo pipefail
 
 . $KOLA_EXT_DATA/commonlib.sh
 
-# List of known broken symlinks that require further investigation:
+# List of known broken symlinks that require further investigation.
+# The sym links in /usr/lib/firmware are often broken and are
+# typically not an issue, so let's skip this location altogether.
 list_broken_symlinks_skip=(
     '/etc/mtab'
     '/etc/ssl/'
     '/etc/swid/swidtags.d/fedoraproject.org'
     '/etc/xdg/systemd/user'
     '/usr/lib/.build-id/'
-    '/usr/lib/firmware/qcom/LENOVO/21BX.xz'
+    '/usr/lib/firmware'
     '/usr/lib/modules/'
     '/usr/share/rhel/secrets/etc-pki-entitlement'
     '/usr/share/rhel/secrets/redhat.repo'


tests/kola/files/validate-symlinks
```diff
@@ -18,7 +18,7 @@ list_broken_symlinks_skip=(
     '/etc/swid/swidtags.d/fedoraproject.org'
     '/etc/xdg/systemd/user'
     '/usr/lib/.build-id/'
-    '/usr/lib/firmware/qcom/LENOVO/21BX.xz'
+    '/usr/lib/firmware/'
```
gursewak1997: I just hope we keep `'/usr/lib/firmware/qcom/LENOVO/21BX.xz'` after we revert this change( hoping this gets fixed soon)

gursewak1997 (dismissed):
lgtm

tests/kola/files/validate-symlinks
marmijo: I agree. I wasnt sure about this one. I could just add `/usr/lib/firmware` as it's own line so there is no loss of info here.

gursewak1997: Sure, let's do that

marmijo:
@gursewak1997 I went ahead and added it back to the list so we don't lose any information. cc @travier

gursewak1997 (dismissed):

travier:
LGTM as is. I think we can drop all firmware symlink tracking so dropping the Lenovo one would be fine.

travier (dismissed):

dustymabe:
I think what we've decided here is that the files under `/usr/lib/firmware/` are regularly broken and it's not an issue and we'll ignore those broken symlinks in the future.

If that's the case let's maybe add a comment indicating that and also maybe make the commit message to state it's a regular occurrence and we've decided we can ignore them.

tests/kola/files/validate-symlinks
gursewak1997: Looks like you can go back to your initial commit and add a comment:)

marmijo:
I've updated the commit and message to reflect that we're opting to skip validating sym links in the `/usr/lib/firmware` location altogether.

gursewak1997 (approved):

travier (approved):

PR #2403, opened 2023-04-26
[rawhide] overrides: unpin linux-firmware-20230310-148.fc39

marmijo:
Now that we are skipping sym link validation in `/usr/lib/firmware` we can drop this pin.
See: https://github.com/coreos/fedora-coreos-config/pull/2402


dustymabe (approved):
LGTM

PR #2410, opened 2023-05-04
denylist: bump snooze for ext.config.kdump.crash on aarch64

dustymabe:
Still not fixed yet, but is being worked on.


bgilbert (approved):

dustymabe:
skipping CI as it doesn't currently test `aarch64`

PR #2411, opened 2023-05-04
overlay.d/05core: add ifname= karg udev rule propagation

### LINKED ISSUE #553: forwarding NIC renaming udev rules from the initramfs
There is a dracut arg called `ifname=` that will allow someone to dynamically create a udev rule to change the name of a NIC to whatever they would like it to be. For example I could tell the system to name NIC with MAC `12:34:56:78:9a:bc` to `infra` with `ifname=infra:12:34:56:78:9a:bc`.

This gets applied via [ifname-genrules.sh](https://github.com/dracutdevs/dracut/blob/5c3d0a96473ac339fa2d1b25213b8f301c1cfd0d/modules.d/40network/ifname-genrules.sh) from the `40network` dracut module which creates a rule in the initramfs like:

```
# cat /etc/udev/rules.d/80-ifname.rules
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="12:34:56:78:9a:bc", ATTR{type}=="1", NAME="infra"
```

Since we persist other parts of the networking config from the initramfs (static IPs, hostname, etc), should we persist these udev rules as well? Or should we require a user to either write out the udev rule (or systemd link file) themselves OR add the `ifname=` karg so that it gets dynamically generated every time?

Just for posterity here is how to create a udev rule or systemd link file:

```
storage:
  files:
    - path: /etc/udev/rules.d/80-ifname.rules
      mode: 0644
      contents:
        inline: |
          SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="12:34:56:78:9a:bc", ATTR{type}=="1", NAME="infra"'
```

```
storage:
  files:
    - path: /etc/systemd/network/25-infra.link
      mode: 0644
      contents:
        inline: |
          [Match]
          MACAddress=12:34:56:78:9a:bc
          [Link]
          Name=infra
```

dustymabe:
This adds to coreos-teardown-initramfs.{service,sh} the ability to propagate udev rules that were generated by ifname-genrules.sh [1] to the real root of the machine to allow for an ephemeral ifname= karg on the first boot to have persistent configuration effect (similar to our other networking kernel arguments).

This also adds tests for both the firstboot and everyboot case for the ifname= karg.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/553

[1] https://github.com/dracutdevs/dracut/blob/5c3d0a96473ac339fa2d1b25213b8f301c1cfd0d/modules.d/40network/ifname-genrules.sh

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
@@ -168,6 +168,20 @@ propagate_initramfs_hostname() {
     fi
 }
 
+# Propagate the ifname= karg udev rules. The policy here is:
+#
+#     - IF ifname karg udev rule file exists
+#     - THEN copy it to the real root
+#
+propagate_ifname_udev_rules() {
+    local udev_file='/etc/udev/rules.d/80-ifname.rules'
+    if [ -e "${udev_file}" ]; then
+        echo "info: propagating ifname karg udev rules to the real root"
+        cp -v "${udev_file}" "/sysroot/${udev_file}"
+        coreos-relabel "${udev_file}"
+    fi
+}
+
 main() {
     # Load libraries from dracut
     load_dracut_libs
@@ -192,6 +206,7 @@ main() {
     else
         propagate_initramfs_hostname
         propagate_initramfs_networking
+        propagate_ifname_udev_rules
     fi
 
     # Configuration has been propagated, but we can't clean up
--- a/tests/kola/networking/ifname-karg/data/ifname-karg-lib.sh
+++ b/tests/kola/networking/ifname-karg/data/ifname-karg-lib.sh
@@ -1,0 +1,36 @@
+# This is a library created for our ifname-karg tests
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# check IP for given NIC name
+check_ip() {
+    # User provides the NIC name.
+    local nic_name=$1
+    # The expected IP is the first one in the range given out by QEMU
+    # user mode networking: https://www.qemu.org/docs/master/system/devices/net.html#using-the-user-mode-network-stack
+    local expected_ip="10.0.2.15"
+    # Verify the given nic name has the expected IP.
+    local nic_ip=$(get_ipv4_for_nic ${nic_name})
+    if [ "${nic_ip}" != "${expected_ip}" ]; then
+        fatal "Error: get ${nic_name} ip = ${nic_ip}, expected is ${expected_ip}"
+    fi
+    ok "get ${nic_name} ip is ${expected_ip}"
+}
+
+# simple file existence check
+check_file_exists() {
+    local file=$1
+    if [ ! -f $file ]; then
+        fatal "expected file ${file} doesn't exist on disk"
+    fi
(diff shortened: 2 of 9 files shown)


tests/kola/networking/ifname-karg/everyboot-systemd-link-file/config.bu
```diff
@@ -0,0 +1,8 @@
+variant: fcos
+version: 1.4.0
+kernel_arguments:
+  should_exist:
+    # Persistently set the ifname kernel argument to set the given MAC address to the
+    # NIC named `kolatest`. The MAC address is from:
+    # https://github.com/coreos/coreos-assembler/blob/5ef5305cda3736ac6a355d8d8ac25009a605f828/mantle/platform/metal.go#L519
```
jlebon: Minor: hmm, that's in the testiso code, which is unrelated here. That MAC address is the default one QEMU assigns. So maybe just:

> The MAC address is the default one QEMU assigns.

?

tests/kola/networking/ifname-karg/data/ifname-karg-lib.sh
```diff
@@ -0,0 +1,36 @@
...
+. $KOLA_EXT_DATA/commonlib.sh
+
+# check IP for given NIC name
+check_ip() {
+    # User provides the NIC name.
+    nic_name=$1
+    # The expected IP is the first one in the range given out by QEMU
+    # user mode networking: https://www.qemu.org/docs/master/system/devices/net.html#using-the-user-mode-network-stack
+    expected_ip="10.0.2.15"
+    # Verify the given nic name has the expected IP. F
```
jlebon: Minor: extraneous F

tests/kola/networking/ifname-karg/data/ifname-karg-lib.sh
```diff
@@ -0,0 +1,36 @@
...
+    nic_ip=$(get_ipv4_for_nic ${nic_name})
```
jlebon: Minor: local (and in the other functions)

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -168,6 +168,20 @@ propagate_initramfs_hostname() {
     fi
 }
 
+# Propagate the ifname= karg udev rules. The policy here is:
+#
+#     - IF ifname karg udev rule file exists
+#     - THEN copy it to the real root
+#
+propagate_ifname_udev_rules() {
+    udev_file='/etc/udev/rules.d/80-ifname.rules'
```
jlebon: Minor: local :)

Though I know in this file at least, we don't really do that, so feel free to leave it off and maybe we can address it all together in a separate commit.

jlebon (commented):
Looks sane to me! Just some minor comments.

tests/kola/networking/ifname-karg/everyboot-systemd-link-file/config.bu
dustymabe: Yeah I copied this comment from somewhere else. It was inaccurate there too. Will update.

dustymabe:
Addressed comments in the latest upload.

jlebon (approved):

PR #2414, opened 2023-05-10
overrides: pin rust-afterburn-5.4.0-2.fc38

coreosbot-releng:
Requested by @marmijo via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,6 +9,16 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  afterburn:
+    evr: 5.4.0-2.fc38
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1492
+      type: pin
+  afterburn-dracut:
+    evr: 5.4.0-2.fc38
+    metadata:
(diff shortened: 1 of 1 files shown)


bgilbert (approved):

jlebon (approved):
I think we need a similar one for rawhide too?

marmijo:
yes, see: https://github.com/coreos/fedora-coreos-config/pull/2415

marmijo:
CI failed with 
`[2023-05-10T17:53:39.427Z] Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.`
This is probably expected. How do we ignore it?

jlebon:
You can merge it in manually after verifying that the diff is indeed what we expected. (It is, but I'll let you do the honours. :) )

bgilbert:
That only works for repo administrators, though.  Ideally we'd break out the downgrade test into a separate non-required CI run to avoid that issue.

bgilbert:
Filed as https://github.com/coreos/fedora-coreos-config/issues/2416.

PR #2420, opened 2023-05-13
manifests: add passt

PhrozenByte:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1436

### DIFF
--- a/manifest-lock.aarch64.json
+++ b/manifest-lock.aarch64.json
@@ -990,6 +990,12 @@
     "pam-libs": {
       "evra": "1.5.2-16.fc38.aarch64"
     },
+    "passt": {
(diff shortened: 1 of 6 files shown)


PhrozenByte:
No idea whether the lockfiles are correct this way (no idea whether I've fiddled with them the way it's expected), but the workflow still fails. However, the failure doesn't seem to be related to `passt`, but some other package (`shim`)?

dustymabe:
I added a basic test for `--net=pasta` functionality.

jlebon (approved):
Didn't realize `^` was a valid character in RPM versioning. Will be interesting to see if anything breaks because of this.

sbrivio-rh:
Thank you so much @PhrozenByte and @dustymabe! I didn't see this until now.

PR #2422, opened 2023-05-15
tests/kola: cancel any pending rpm-ostree transaction in upgrade test

dustymabe:
We've seen it where zincati gets restarted but a previous rpm-ostree deploy (that had been scheduled by Zincati) was still running. Let's cancel anything that's pending before restarting Zincati.

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -186,7 +186,10 @@ if vereq $version $last_release; then
 fi
 
 # Restart Zincati if configuration was changed
-[ "${need_zincati_restart}" == "true" ] && systemctl restart zincati
+if [ "${need_zincati_restart}" == "true" ]; then
+    rpm-ostree cancel # in case anything was already in progress
+    systemctl restart zincati
+fi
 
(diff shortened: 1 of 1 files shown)


tests/kola/upgrade/extended/test.sh
```diff
@@ -186,7 +186,10 @@ if vereq $version $last_release; then
 fi
 
 # Restart Zincati if configuration was changed
-[ "${need_zincati_restart}" == "true" ] && systemctl restart zincati
+if [ "${need_zincati_restart}" == "true" ]; then
+    rpm-ostree cancel # in case anything was already in progress
```
jlebon, line 190: Minor: a larger hammer is `systemctl stop rpm-ostreed`. `rpm-ostree cancel` should normally work too, but stopping the service is more bulletproof.

jlebon (approved):

tests/kola/upgrade/extended/test.sh
dustymabe: yeah I wasn't sure if I'd need to restart it afterwards. This seemed to work in my tests so I just went with it.

PR #2427, opened 2023-05-18
kola/denylist: Denylist `ext.config.files.validate-symlinks` due to a broken symlink

gursewak1997:
The symlink `/usr/lib/debug/usr/lib64/bfd-plugins/libdep.so-2.40-7.fc39.x86_64.debug` is broken. So denylisting this test till we get a fix for that.
Tracker issue: https://github.com/coreos/fedora-coreos-tracker/issues/1496
BZ issue: https://bugzilla.redhat.com/show_bug.cgi?id=2208360

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -29,3 +29,8 @@
     - ppc64le
   streams:
     - rawhide
+- pattern: ext.config.files.validate-symlinks
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1496
+  snooze: 2023-06-01
+  streams:
+    - rawhide


dustymabe:
Is this broken symlink expected? I feel like we should investigate this further rather than just skip all `/usr/lib/debug/` broken symlinks.

dustymabe:
i.e. does the maintainer of the package that owns this file know it's a broken symlink? Is it intended to be broken?

We can open a tracker issue, denylist/snooze this test on rawhide for now and try to start a conversation with the package maintainer.

cgwalters:
This broken symlink stuff is a perfect example of something that should be a Fedora test that the people building the packages see and not something we own and gate on.

dustymabe:
> This broken symlink stuff is a perfect example of something that should be a Fedora test that the people building the packages see and not something we own and gate on.

Agreed. Having a chat with QA today.

dustymabe (approved):
LGTM

PR #2433, opened 2023-05-24
manifests/fedora-coreos: drop removal of sysusers.d/dbus.conf

dustymabe:
Originally introduced as part of d3a0cf9, the bug [1] was fixed by a PR to the RPM [2] a while back so we shouldn't need to carry this any longer.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2105177
[2] https://src.fedoraproject.org/rpms/dbus/pull-request/16#

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -131,11 +131,6 @@ postprocess:
     done
 
 
-remove-from-packages:
-  # Drop some buggy sysusers fragments which do not match static IDs allocation:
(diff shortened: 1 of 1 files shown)


dustymabe:
anybody know how to properly test this doesn't have any negative effects?

jlebon (approved):
We already have tests which verify the allocated IDs keep matching the expected ones (`ext.config.files.fcos_*`), so we should be covered.

dustymabe:
cool thanks!

PR #2435, opened 2023-05-26
selectively enable ostree autopruning

dustymabe:
This adds autopruning for ppc64le so we can start to officially release it.

It also adds back in the qualcomm dtbs and autopruning for aarch64 on F39+, which allows us to passively pick up the change when the `next-devel` branch is re-enabled at F39 beta time. If we re-enable the `next-devel` branch earlier then we can change this to pick it up quicker too.

Finally, there is a commit in there to drop some code that's no longer necessary.

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -29,6 +29,18 @@ conditional-include:
     # https://github.com/coreos/fedora-coreos-tracker/issues/1247#issuecomment-1355314761
     # https://github.com/coreos/fedora-coreos-tracker/issues/1495#issuecomment-1561765705
     include: ostree-autoprune.yaml
+  - if:
+      - basearch == "aarch64"
+      - releasever >= 39
+    # In F39+ we will stop removing the qcom dtb files and thus we'll leverage OSTree autopruning
+    # so that we don't run into https://github.com/coreos/fedora-coreos-tracker/issues/1464
+    # again. OSTree autopruning is new so we're selectively enabling it before making it the default.
+    include: ostree-autoprune.yaml
+  - if:
+      - basearch == "aarch64"
+      - releasever == 38
+    # Remove qcom dtb on F38 files since autopruning isn't in place yet
+    include: aarch64-drop-qcom-dtb-files.yaml
 
 ostree-layers:
   - overlay/15fcos
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -23,6 +23,12 @@ conditional-include:
   - if: basearch != "s390x"
     # And remove some cruft from grub2
     include: grub2-removals.yaml
+  - if: basearch == "ppc64le"
+    # Need OSTree autopruning on ppc64le (because kernels aren't compressed)
+    # until we increase the size of /boot.
+    # https://github.com/coreos/fedora-coreos-tracker/issues/1247#issuecomment-1355314761
+    # https://github.com/coreos/fedora-coreos-tracker/issues/1495#issuecomment-1561765705
+    include: ostree-autoprune.yaml
 
 ostree-layers:
   - overlay/15fcos
--- a/manifests/ostree-autoprune.yaml
+++ b/manifests/ostree-autoprune.yaml
@@ -1,0 +1,11 @@
+# Enable OSTree autopruning to help with /boot size constraints
+# https://github.com/coreos/fedora-coreos-tracker/issues/1495
+postprocess:
+  - |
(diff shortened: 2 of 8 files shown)


manifests/ostree-autoprune.yaml
```diff
@@ -0,0 +1,12 @@
...
+postprocess:
+  # Enable OSTree autopruning to help with /boot size constraints
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1495
+  - |
+    #!/usr/bin/env bash
+    mkdir -p /usr/lib/systemd/system/ostree-finalize-staged.service.d
+    cat <<'EOF' > /usr/lib/systemd/system/ostree-finalize-staged.service.d/ostree-autoprune.conf
+    [Service]
+    # https://github.com/coreos/fedora-coreos-tracker/issues/1495"
+    Environment=OSTREE_SYSROOT_OPTS=early-prune
```
dustymabe, line 10: Judging from the commit message in https://github.com/ostreedev/ostree/commit/c561e6179e965d11bba7e1d80ca35b7ad6fc7bc5 I thought I'd need to set `$OSTREE_ENABLE_AUTO_EARLY_PRUNE` here, but I couldn't get that to work and judging from the tests in that commit it needed to be `OSTREE_SYSROOT_OPTS=early-prune`. Am I missing something?

manifests/ostree-autoprune.yaml
```diff
@@ -0,0 +1,12 @@
+
+postprocess:
+  # Enable OSTree autopruning to help with /boot size constraints
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1495
+  - |
+    #!/usr/bin/env bash
+    mkdir -p /usr/lib/systemd/system/ostree-finalize-staged.service.d
+    cat <<'EOF' > /usr/lib/systemd/system/ostree-finalize-staged.service.d/ostree-autoprune.conf
+    [Service]
+    # https://github.com/coreos/fedora-coreos-tracker/issues/1495"
```
jlebon: Stray quote character here

manifests/ostree-autoprune.yaml
```diff
@@ -0,0 +1,12 @@
+
```
jlebon: Super minor: extraneous newline (optionally since we only have the postprocess here, maybe we can move the comments on line 3-4 here to act as a toplevel docstring).

(And similarly for the other submanifest we're adding in the next commit.)

manifests/ostree-autoprune.yaml
jlebon: Sigh, yes I forgot to update the commit message after addressing https://github.com/ostreedev/ostree/pull/2847#discussion_r1178339033.

manifests/fedora-coreos.yaml
```diff
@@ -29,6 +29,17 @@ conditional-include:
     # https://github.com/coreos/fedora-coreos-tracker/issues/1247#issuecomment-1355314761
     # https://github.com/coreos/fedora-coreos-tracker/issues/1495#issuecomment-1561765705
     include: ostree-autoprune.yaml
+  - if:
+      - basearch == "aarch64"
+      - releasever >= 39
+    # Need OSTree autopruning on aarch64 until we stop removing the qcom dtb files
+    # as was done in https://github.com/coreos/fedora-coreos-tracker/issues/1464
```
jlebon: That comment looks incorrect since we are not removing the blobs in this path.

manifests/ostree-autoprune.yaml
dustymabe: Haha, no worries. It took me at least 45m of beating my head against the wall to figure that out. Maybe at least update the [description](https://github.com/ostreedev/ostree/pull/2847#issue-1667154492) in the PR for anyone happening upon it?

manifests/ostree-autoprune.yaml
dustymabe: done

manifests/fedora-coreos.yaml
dustymabe: I think you are right. I rewrote the comment. Hopefully it is more accurate now.

jlebon (dismissed):

dustymabe:
ok I pushed up a commit to fast track a new ostree RPM with a autoprune corner case fixed. This should be ready to merge now (if it passes CI).

cgwalters (approved):

gursewak1997 (approved):
Reviewed the latest commit and went over the previous ones briefly. LGTM.

PR #2436, opened 2023-05-26
[rawhide] overrides: pin util-linux-2.39-3.fc39

coreosbot-releng:


marmijo (approved):
LGTM

PR #2437, opened 2023-05-26
tests/kola: upstream network-online login test from RHCOS

jlebon:
This test was originally added in RHCOS[[1]], but it's equally valid to run in FCOS. That would have caught the original issue when it was still present in Fedora.

[1]: https://github.com/openshift/os/pull/1294

### DIFF
--- a/tests/kola/systemd/network-online/config.bu
+++ b/tests/kola/systemd/network-online/config.bu
@@ -1,0 +1,19 @@
+variant: fcos
+version: 1.4.0
+systemd:
+  units:
+    - name: block-network-online.service
+      enabled: true
+      contents: |
+        [Unit]
+        After=NetworkManager-wait-online.service
+        Before=network-online.target
+        Wants=network-online.target
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=sleep infinity
+
+        [Install]
+        WantedBy=multi-user.target
--- a/tests/kola/systemd/network-online/test.sh
+++ b/tests/kola/systemd/network-online/test.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+## kola:
+##   description: Verify that network-online.target doesn't block login
+##   tags: platform-independent
+##   # this really shouldn't take long; if it does, it's that we're hitting the
+##   # very issue we're testing for
+##   timeoutMin: 3
+
+# If the user provides a systemd unit which pulls in `network-online.target`,
+# we want to make sure that logins don't block on `network-online.target` being
+# reached. `block-network-online.service` verifies this by pulling in `network-
+# online.target` and preventing it from being reached by running `Before=` it
+# and sleeping forever.
+
+# We hit this in RHCOS with iscsi.service causing network-online.target to block
+# remote-fs-pre.target and hence systemd-user-sessions.service from running:
+#
+# https://github.com/openshift/os/pull/1279
+# https://issues.redhat.com/browse/OCPBUGS-11124
+
+set -euo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# The fact that we're here means that `systemd-user-sessions.service` was
+# reached and logins work since kola was able to SSH to start us. But let's do
+# some sanity-checks to verify that the test was valid.
+
+# verify that block-network-online.service is still activating since it's stuck sleeping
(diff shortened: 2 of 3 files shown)


tests/kola/systemd/network-online/config.bu
```diff
@@ -0,0 +1,32 @@
+variant: fcos
+version: 1.4.0
+systemd:
+  units:
+    - name: ovs-configuration.service
+      enabled: true
+      contents: |
+        [Unit]
+        Requires=openvswitch.service
```
dustymabe: it's a bit weird here to use `ovs-configuration.service` and `openvswitch.service`. I think this is because in RHCOS those services exist?

Maybe we could add some context here about why they are useful in RHCOS and why using them in FCOS still has the desired effect even if they don't exist.

tests/kola/systemd/network-online/test.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+## kola:
+##   description: Verify that network-online.target doesn't block login
+##   tags: platform-independent
+##   # this really shouldn't take long; if it does, it's that we're hitting the
+##   # very issue we're testing for
+##   timeoutMin: 3
+
+# https://github.com/openshift/os/pull/1279
+# https://issues.redhat.com/browse/OCPBUGS-11124
```
dustymabe, line 19: I think it could be nice here to summarize what the units in the butane are doing in more detail. This is a bit of the age old argument of commenting the code versus letting people read the code (since the code never lies, only comments do :)), but my opinion is that it would be useful to people who are trying to glance and not deeply understand.

tests/kola/systemd/network-online/config.bu
```diff
@@ -0,0 +1,32 @@
...
+        After=NetworkManager-wait-online.service openvswitch.service
+        Before=network-online.target kubelet.service
+
```
dustymabe, line 12:
```suggestion
```

nit/optional: I would delete the blank lines in these units. I'm not a fan of blank lines in the yaml(butane) files.

tests/kola/systemd/network-online/test.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -euo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+# The fact that we're here means that logins must be working since kola was able
```
dustymabe:
```suggestion
# The fact that we're here means that `systemd-user-sessions.service` was able to
# start and logins must be working since kola was able
```

I think mentioning the unit that we were able to get to is useful here (for clarity). It's `systemd-user-sessions.service` right?

tests/kola/systemd/network-online/test.sh
```diff
@@ -0,0 +1,32 @@
...
+# to SSH to start us. But let's do some sanity-checks to verify that the test
+# was valid.
+
+# verify that ovs-configuration is still activating
```
dustymabe: optional: more explicit:

```suggestion
# verify that ovs-configuration is still activating because of the infinity sleep
```

tests/kola/systemd/network-online/test.sh
```diff
@@ -0,0 +1,32 @@
...
+if [[ $(systemctl show ovs-configuration.service -p ActiveState) != "ActiveState=activating" ]]; then
+  systemctl status ovs-configuration.service
+  fatal "ovs-configuration.service isn't activating"
+fi
+
+# verify that network-online is not yet active
```
dustymabe: optional: more explicit:

```suggestion
# verify that network-online.target is not yet active because it's waiting on ovs-configuration.service
```

dustymabe (commented):
A few comments. Sorry for requesting so much extra commentary. My brain is tired right now and I know in the future when I look at this I'm going to have all of these same questions.

tests/kola/systemd/network-online/config.bu
jlebon: I'd rather keep them since they help readability and is how you'd normally write systemd units.

jlebon:
I simplified the Butane config more and added more comments in the test.

dustymabe:
> I simplified the Butane config more and added more comments in the test.

The comments are beautiful and the test really is very clear now! Thank you!

dustymabe (approved):

jlebon:
This requires https://github.com/coreos/coreos-assembler/pull/3487.

PR #2440, opened 2023-05-28
Pin util-linux to 2.38.1-4.fc38

djds:
https://github.com/coreos/rpm-ostree/issues/4437

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -24,3 +24,8 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1476
       type: pin
+  util-linux:
+    evr: 2.38.1-4.fc38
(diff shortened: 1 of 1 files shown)


dustymabe:
Hi. I'm not sure this is necessary. We [already pinned](https://github.com/coreos/fedora-coreos-config/pull/2436) it for `rawhide`, which is where this problem surfaced. It hasn't hit the F38 packages yet. In fact there haven't been any new updates for F38 [for some time](https://bodhi.fedoraproject.org/updates/?search=&packages=util-linux&releases=F38). Did you see an error?

djds:
> Did you see an error?

I did, but it was a custom build overlaying `testing-devel`:
```yaml
include: fedora-coreos-config/manifests/fedora-coreos-base.yaml
```

Closing this as no longer relevant.

PR #2445, opened 2023-05-31

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,34 +9,6 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
-  afterburn:
-    evr: 5.4.2-1.fc38
-    metadata:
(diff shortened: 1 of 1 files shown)


bgilbert (approved):

dustymabe:
this won't succeed CI until the next bump-lockfile run completes successfully because the fedora-release rpm that this is removing from the overrides lockfile never made it into the json lockfiles. I'm investigating the bump-lockfile failure and hopefully will get it unblocked.

dustymabe:
> this won't succeed CI until the next bump-lockfile run completes successfully because the fedora-release rpm that this is removing from the overrides lockfile never made it into the json lockfiles. I'm investigating the bump-lockfile failure and hopefully will get it unblocked.

https://github.com/coreos/fedora-coreos-config/pull/2452

dustymabe:
ok this should pass tests now.

PR #2453, opened 2023-06-06
[rawhide] overrides: pin systemd-253.4-1.fc39

coreosbot-releng:


dustymabe:
CI will fail on this because of other issues in rawhide right now, but we can merge it assuming the ntp tests pass.

dustymabe:
ok this resolves the ntp tests. merging

dustymabe (approved):

PR #2457, opened 2023-06-07
overrides: pin passt-0^20230509.g96f8d55-1.fc38

coreosbot-releng:
Requested by @gursewak1997 via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


dustymabe (approved):
LGTM

PR #2458, opened 2023-06-08
overrides: pin systemd-253.4-1.fc38

coreosbot-releng:


aaradhak (approved):
LGTM

PR #2464, opened 2023-06-13
.cci.jenkinsfile: change path for download-overrides.py

c4rt0:
Once below PR's, requesting the move of download-overrides.py
has been merged, the file path will change to modified. 

See: https://github.com/coreos/coreos-assembler/pull/3510
See: https://github.com/coreos/fedora-coreos-releng-automation/pull/176

Issue:

See: https://github.com/coreos/fedora-coreos-tracker/issues/1185

PR #2466, opened 2023-06-14
[testing] .cci.jenkinsfile: change path for download-overrides.py

dustymabe:
This is a forward port of https://github.com/coreos/fedora-coreos-config/pull/2464 so we can go ahead and merge  https://github.com/coreos/fedora-coreos-releng-automation/pull/176

We'll merge it into the `testing` branch now and it will get promoted to `stable` next round so CI for `stable` will pass.


c4rt0 (approved):
LGTM

PR #2468, opened 2023-06-16
tests: check that we should not find `/dev/disk/by-id/google*` device links on non GCP

HuijingHei:
See https://issues.redhat.com/browse/OCPBUGS-13754 
https://github.com/coreos/fedora-coreos-tracker/issues/1475#issuecomment-1551857322

### DIFF
--- a/tests/kola/disks/no-google-device-links
+++ b/tests/kola/disks/no-google-device-links
@@ -1,0 +1,17 @@
+#!/bin/bash
+## kola:
+##   platforms: !gcp
+##   exclusive: false
+##   description: Verify no /dev/disk/by-id/*google* device links are found on non GCP.
+
+# See https://issues.redhat.com/browse/OCPBUGS-13754
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+links=$(find /dev/disk/by-id/ -iname "*google*")
+if [[ -n "${links:-}" ]]; then
+    fatal "Error: should not find /dev/disk/by-id/*google* device links on non GCP"
+fi
+ok "No /dev/disk/by-id/*google* device links are found on non GCP"


tests/kola/var-mount/scsi-id/test.sh
```diff
@@ -39,3 +40,9 @@ if ! grep -q "63-scsi-sg3_symlink.rules" $tempfile; then
 fi
 
 rm -f $tempfile
+
+# https://issues.redhat.com/browse/OCPBUGS-13754
+links=$(find /dev/disk/by-id/ -name google*)
```
travier:
```suggestion
links=$(find /dev/disk/by-id/ -iname "*google*")
```

tests/kola/var-mount/scsi-id/test.sh
```diff
@@ -39,3 +40,9 @@ if ! grep -q "63-scsi-sg3_symlink.rules" $tempfile; then
...
+if [ -n "${links:-}" ]; then
```
travier:
```suggestion
if [[ -n "${links:-}" ]]; then
```

travier (commented):
Some suggestions / nits.

dustymabe:
I feel like maybe we should create a new test for this since it feels weird to just add it into another (seemingly unrelated) test.

IIUC this doesn't require a specific disk setup so we could make it a non-exclusive test (that way we won't end up spawning an extra VM but just re-use the non-exclusive-tests VM.

HuijingHei:
> I feel like maybe we should create a new test for this since it feels weird to just add it into another (seemingly unrelated) test.
> 
> IIUC this doesn't require a specific disk setup so we could make it a non-exclusive test (that way we won't end up spawning an extra VM but just re-use the non-exclusive-tests VM.

SGTM. Thanks @dustymabe and @travier for the reivew.

tests/kola/disks/device-link
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+## kola:
+##   platforms: qemu
```
dustymabe: add ?

```
##   exclusive: false
```

dustymabe:
maybe name the test `tests/kola/disks/no-google-device-links`

tests/kola/disks/no-google-device-links
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+## kola:
+##   platforms: qemu
```
travier: I'd say that we should run this on all platforms but GCP?

HuijingHei: Yes, you are correct, maybe I should change to `!gcp` ?

dustymabe (approved):
LGTM

PR #2469, opened 2023-06-16
drop default fedora-repos-modular

juhp:
part of approved https://fedoraproject.org/wiki/Changes/No_default_fedora-repos-modular for F39

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -55,9 +55,6 @@ lockfile-repos:
 packages:
   - fedora-release-coreos
   - fedora-repos-ostree
-  # fedora-repos-modular was converted into its own subpackage in f33
-  # Continue to include it in case users want to use it.
-  - fedora-repos-modular
   # the archive repo for more reliable package layering
   # https://github.com/coreos/fedora-coreos-tracker/issues/400
   - fedora-repos-archive


dustymabe:
Thanks for the PR. We'd want to implement this only for F39 (don't worry I can take care of that).

I think maybe I should also bring this up in the weekly meeting because it could cause trouble if people have existing layered packages and this repo goes away.

juhp:
(Of course fedora-repos-modular can still be layered I think)

dustymabe:
> part of approved https://fedoraproject.org/wiki/Changes/No_default_fedora-repos-modular for F39

I'll note here that as of today this isn't yet an approved change.

dustymabe:
Issue where we are discussing this: https://github.com/coreos/fedora-coreos-tracker/issues/1513

juhp:
Well just noting the Change _was_ formally [approved by Fesco](https://pagure.io/fesco/issue/3007#comment-861463) on 13 June (I should have noted that explicitly perhaps): it took a while before the Change page status was updated...
Anyway CoreOs is not obliged to follow this Change immediately: it is your call of course.
Though I would assume most CoreOS modular users would be consuming it more in containers?

dustymabe:
Yeah. We want to follow the change request, just want to be careful about leaving users behind. Discussion in https://github.com/coreos/fedora-coreos-tracker/issues/1513

juhp:
Though see also https://fedoraproject.org/wiki/Changes/RetireModularity
Okay you already noted in your tracker.

dustymabe:
ok I posted a more comprehensive version of this in https://github.com/coreos/fedora-coreos-config/pull/2508 which should be able to merge now because the retire modularity change proposal was approved by FESCO. I'll close this in favor of https://github.com/coreos/fedora-coreos-config/pull/2508

PR #2473, opened 2023-06-21
s390x: create 'by-partlabel' symlinks for labeled filesystems on DASD-CDL partitions

nikita-dubrovskii:
CDL formatted DASD disks support only up to 3 partitions and don't use GPT, so there is no `/dev/disk/by-partlabel/*` symlinks but only `/dev/disk/by-label/*`.

### DIFF
--- a/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-dasd-partlabel.rules
+++ b/overlay.d/05core/usr/lib/udev/rules.d/90-coreos-dasd-partlabel.rules
@@ -1,0 +1,15 @@
+# CoreOS-specific symlinks for DASD disks
+# DASDs don't use GPT, but in some cases it'd be good to have '/dev/disk/by-partlabel/*' symlinks
+
+ACTION!="add|change", GOTO="dasd_partlabel_end"
+SUBSYSTEM!="block", GOTO="dasd_partlabel_end"
+KERNEL!="dasd*", GOTO="dasd_partlabel_end"
+
+# Act only on partitions with labeled filesystems
+ENV{DEVTYPE}!="partition", GOTO="dasd_partlabel_end"
+ENV{ID_FS_USAGE}!="filesystem", GOTO="dasd_partlabel_end"
+ENV{ID_FS_LABEL_ENC}!="?*", GOTO="dasd_partlabel_end"
+
+SYMLINK+="disk/by-partlabel/$env{ID_FS_LABEL_ENC}"
+
+LABEL="dasd_partlabel_end"


overlay.d/05core/usr/lib/udev/rules.d/90-coreos-dasd-partlabel.rules
```diff
@@ -0,0 +1,15 @@
+# CoreOS-specific symlinks for DASD disks
```
cgwalters, line 1: Looks sane to me but...this isn't actually CoreOS specific, is it?  I think we could upstream this to systemd, right?

cgwalters (approved):

nikita-dubrovskii:
https://github.com/coreos/butane/issues/453

overlay.d/05core/usr/lib/udev/rules.d/90-coreos-dasd-partlabel.rules
nikita-dubrovskii: TBH i don't know.  There is an issue when installing CoreOS on zVM with config generated by Butane + `boot-device` syntax sugar, where `by-partlabel` is used by default for LUKS encryption.

nikita-dubrovskii: /cc @bgilbert

bgilbert:
We shouldn't do this.  Those partitions don't actually have partition labels, but this PR is incorrectly claiming that they do, and repurposing a completely different label (the filesystem label) just to have a name to use for the symlink.  This seems likely to increase confusion without solving an actual problem.  The proper fix for https://github.com/coreos/butane/issues/453 is to add Butane layouts for s390x, not to try to force the x86_64 template to work on s390x.

I'll go ahead and close.

cgwalters:
OK, I do generally agree, however I would say that this is pretty widespread issue for us, beyond just that LUKS example.  If I understand correctly, every use of `/dev/disk/by-partlabel` we recommend in docs is broken on s390x.  That's...a lot!  It's used in e.g. https://docs.fedoraproject.org/en-US/fedora-coreos/storage/ too.

bgilbert:
Yes, it's always been like that.  If we want to switch our docs to `/dev/disk/by-label`, we can likely do that in some places, though probably not when reprovisioning storage where the underlying filesystem is going to be wiped (since that would create symlink races).  We might need to have s390x-specific instructions in some places.

nikita-dubrovskii:
> We shouldn't do this. Those partitions don't actually have partition labels, but this PR is incorrectly claiming that they do, and repurposing a completely different label (the filesystem label) just to have a name to use for the symlink. This seems likely to increase confusion without solving an actual problem. 

I had similar concerns about this udev-rule, on the other hand DASDs don't support GPT and usage of filesystems `labels` as `partlabels` shouldn't add any issues.

> The proper fix for [coreos/butane#453](https://github.com/coreos/butane/issues/453) is to add Butane layouts for s390x, not to try to force the x86_64 template to work on s390x.

Agree. At least i refreshed my memory regarding udev

bgilbert:
> usage of filesystems `labels` as `partlabels` shouldn't add any issues.

Well, filesystem and partition labels have different lifecycles.  `mkfs.ext4 /dev/disk/by-partlabel/foo` has no effect on the symlink, while `mkfs.ext4 /dev/disk/by-label/foo` causes the latter symlink to be deleted.

PR #2475, opened 2023-06-21
denylist: snooze rpmostree and ostree tests

marmijo:
Snooze the following tests for 3 weeks while we wait for a fix in the kernel to the mlx5 driver.
- `rpmostree.install-uninstall`
- `rpmostree.upgrade-rollback`
- `ostree.hotfix`

These three tests started failing in rawhide on Azure. See: https://github.com/coreos/fedora-coreos-tracker/issues/1515

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -25,3 +25,24 @@
     - ppc64le
   streams:
     - rawhide
+- pattern: rpmostree.install-uninstall
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1515
+  snooze: 2023-07-12
+  platforms:
+    - azure
+  streams:
+    - rawhide
+- pattern: rpmostree.upgrade-rollback
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1515
+  snooze: 2023-07-12
+  platforms:
+    - azure
+  streams:
+    - rawhide
+- pattern: ostree.hotfix
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1515
+  snooze: 2023-07-12
+  platforms:
+    - azure
+  streams:
+    - rawhide


cgwalters:
I'd be surprised if it was really just those tests; I think any test that pulls from the network is going to be subject to this.

I think instead we need to pin the kernel, no?

kola-denylist.yaml
```diff
@@ -25,3 +25,24 @@
     - ppc64le
   streams:
     - rawhide
+- pattern: rpmostree.install-uninstall
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1515
+  snooze: 2023-07-12
+  arches:
+    - x86_64
```
dustymabe: Since we're only seeing this on azure WDYT of something like:

```suggestion
  platforms:
    - azure
```

dustymabe:
> I'd be surprised if it was really just those tests; I think any test that pulls from the network is going to be subject to this.
> 
> I think instead we need to pin the kernel, no?

If this were `testing-devel` (hence closer to being promoted to production streams) I'd advocate for pinning the kernel. Since this is `rawhide` (no risk of users hitting this problem) I'm more in favor of snoozing/denylistting these tests so we can continue to test kernel updates in rawhide.

kola-denylist.yaml
marmijo: Thanks for the suggestion. I updated the commit

marmijo:
> I'm more in favor of snoozing/denylistting these tests so we can continue to test kernel updates in rawhide.

I think I agree. This isnt hitting any test in any other arch or cloud test runs. I think we should continue to test kernel updates in rawhide until we start seeing this in other areas.

gursewak1997 (approved):

dustymabe (approved):
LGTM

PR #2485, opened 2023-06-26
denylist: remove rpmostree and ostree tests

marmijo:
https://github.com/coreos/fedora-coreos-pipeline/pull/881 will snooze the entire kola-azure cloud test on rawhide for a month. These tests wont run during that time, they can be removed.

This requires https://github.com/coreos/fedora-coreos-pipeline/pull/881 to merge first, otherwise these tests will still run on rawhide.

see: https://github.com/coreos/fedora-coreos-tracker/issues/1515


dustymabe (approved):
LGTM

Issue #2157, opened 2023-01-06
Add space after `!` in kola YAML fields that want to negate semantics


jlebon:
In recent conversations, one suggestion I had was to write e.g.

```yaml
architectures: "!x86_64 aarch64"
```

as

```yaml
architectures: "! x86_64 aarch64"
```

instead, so that it's clearer that the `!` is distributive and not per element. Consider having kola tests in this repo follow this convention. (And then we could consider enforcing it in kola too.)

PR #2488, opened 2023-06-28
tests/kola: fix exclusion for gcp platform on no-google-device-links test

dustymabe:
A "!" at the beginning of an element has a special meaning in YAML [1] so we need to put it in quotes. While we are here also insert a space to comply with the future work that will happen as part of [2].

[1] https://stackoverflow.com/questions/9664113/what-does-a-single-exclamation-mark-do-in-yaml
[2] https://github.com/coreos/fedora-coreos-config/issues/2157

### DIFF
--- a/tests/kola/disks/no-google-device-links
+++ b/tests/kola/disks/no-google-device-links
@@ -1,6 +1,6 @@
 #!/bin/bash
 ## kola:
-##   platforms: !gcp
+##   platforms: "! gcp"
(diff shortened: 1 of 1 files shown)


bgilbert (approved):
LGTM

HuijingHei:
Thanks for the fix.
/lgtm
/approve

HuijingHei (approved):

PR #2494, opened 2023-07-05
denylist: Snooze kdump.crash in ppc64le for rawhide

gursewak1997:
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1523


marmijo (approved):
LGTM

PR #2500, opened 2023-07-10
overrides: fast-track ignition-2.16.1-1.fc38

coreosbot-releng:


dustymabe (approved):
LGTM

PR #2504, opened 2023-07-13
[rawhide] overrides: fast-track ignition-2.16.2-1.fc39

coreosbot-releng:
Requested by @gursewak1997 via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,9 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  ignition:
+    evr: 2.16.2-1.fc39
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-19f5506eed
+      type: fast-track


marmijo (approved):
LGTM

bgilbert:
Hmm, why was this needed?  Rawhide updates usually go to stable immediately.

dustymabe:
> Hmm, why was this needed? Rawhide updates usually go to stable immediately.

It goes to "stable" state in bodhi immediately but it's not available in any yum repos until the next nightly pungi compose of rawhide runs. If the pungi compose fails for a day (or multiple days) then the `n-1` version is still what gets used. Since the `n-1` version causes failures on `s390x` and `ppc64le` and the first compose today [failed](https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20230713.n.0/STATUS) which caused this morning's run of rawhide to fail.. This PR was a reaction to those morning failures.

dustymabe:
TL;DR - 9/10 times fast-tracking in rawhide isn't necessary or useful, but in this case it was useful.

PR #2508, opened 2023-07-14
Drop modularity support/tests

dustymabe:
In Fedora 39 modularity is going away [1]. Let's drop the repos for F39+ and also drop the test that we have. We do still want to test installing crio works so let's add crio to the ext.config.extensions.package test.

[1] https://fedoraproject.org/wiki/Changes/RetireModularity


manifests/fedora-repos-modular.yaml
```diff
@@ -0,0 +1,4 @@
+# Modularity is going away in F39+ so we'll only include this
```
marmijo: small nit
```suggestion
# Modularity is going away in F39+ so we'll only include
```

manifests/fedora-coreos.yaml
```diff
@@ -41,6 +41,11 @@ conditional-include:
       - releasever == 38
     # Remove qcom dtb on F38 files since autopruning isn't in place yet
     include: aarch64-drop-qcom-dtb-files.yaml
+  - if:
+      - releasever == 38
+    # Modularity is going away in F39+ so we'll only include this
```
marmijo: small nit
```suggestion
    # Modularity is going away in F39+ so we'll only include
```

marmijo (dismissed):
LGTM with two small nits

PR #2511, opened 2023-07-17
[rawhide] overrides: pin systemd-253.5-6.fc39

coreosbot-releng:


dustymabe (approved):
LGTM

aaradhak:
Reference to the tracker issue : https://github.com/coreos/fedora-coreos-tracker/issues/1527

PR #2512, opened 2023-07-18
kola: mark denylisted tests with 'warn: true'

nikita-dubrovskii:
Requires: https://github.com/coreos/coreos-assembler/pull/3539


cgwalters:
I have a general question about the structure of things that seeing this PR raises for me https://github.com/coreos/coreos-assembler/issues/3344#issuecomment-1641193186

PR #2514, opened 2023-07-18
initramfs: Ensure we `Requires=ostree-prepare-root.service`

cgwalters:
Today if that unit doesn't start we can continue on and continue on to `ignition-ostree-check-rootfs-size.service` which will fail.


cgwalters:
OK that's weird, it fails the same way in CI here with the old systemd...but that worked locally.

cgwalters:
Ohh wait, I see.  Because `sysroot.mount` isn't statically generated, we can't actually order based on it.

cgwalters:
Just going to roll this into https://github.com/coreos/fedora-coreos-config/pull/2515

PR #2515, opened 2023-07-18
initramfs: Remove manual sysroot mounting

### LINKED ISSUE #1527: rawhide: systemd-254 drops ostree-prepare-root from transaction
### Describe the bug

The kola basic tests are failing on the latest rawhide x86_64 builds since build [39.20230716.91.0](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/build/1545/)

- basic
```
[2023-07-17T12:49:24.039Z] --- FAIL: basic (8.35s)
[2023-07-17T12:49:24.040Z]         harness.go:1741: mach.Start() failed: machine c78e2b87-ddfa-4237-b769-7d16ddf5fc56 entered emergency.target in initramfs
```

- fcos.upgrade.basic 
```
[2023-07-17T12:57:57.919Z]     --- FAIL: fcos.upgrade.basic/upgrade-from-current (120.97s)
[2023-07-17T12:57:57.919Z]             basic.go:318: failed waiting for machine reboot: timed out after 2m0s waiting for machine to reboot
```

Observed on checking the console logs that these tests began failing after the systemd version upgrade from `systemd 253.5-6.fc39 -> systemd 254~rc2-1.fc39`

console.log output related to systemd:
```
[    5.165851] systemd[1]: Failed to start ignition-ostree-check-rootfs-size.service - Ignition OSTree: Check Root Filesystem Size.
[[0;32m  OK  [0m] Stopped [0;1;39mdracut-cmdline.service[0m - dracut cmdline hook.
[    5.167786] systemd[1]: Dependency failed for ignition-complete.target - Ignition Complete.
[[0;32m  OK  [0m] Stopped [0;1;39mdracut-cmdline-ask.service[0m… for additional cmdline parameters.
[    5.170887] systemd[1]: Dependency failed for initrd.target - Initrd Default Target.
[    5.171559] systemd[1]: initrd.target: Job initrd.target/start failed with result 'dependency'.
[[0;32m  OK  [0m] Started [0;1;39memergency.service[0m - Emergency Shell.
[[0;32m  OK  [0m] Reached target [0;1;39memergency.target[0m - Emergency Mode.
         Starting [0;1;39mignition-virtio-dump-jour…ce[0m - Dump Journal to Virtio Port...
[    5.179091] systemd[1]: initrd.target: Triggering OnFailure= dependencies.
[    5.180358] systemd-journald[259]: Received client request to sync journal.
[    5.180937] systemd[1]: ignition-complete.target: Job ignition-complete.target/start failed with result 'dependency'.
[    5.181807] systemd[1]: systemd-ask-password-console.path: Deactivated successfully. 
```
 
This is the kola basic output from one of the latest rawhide builds:
[kola-basic-7146e04d.zip](https://github.com/coreos/fedora-coreos-tracker/files/12072038/kola-basic-7146e04d.zip)

### Reproduction steps

1. build and run latest FCOS rawhide [image](https://builds.coreos.fedoraproject.org/browser?stream=rawhide&arch=x86_64)
2. Run kola test - basic
3. kola basic test fails

### Expected behavior

kola basic test to pass with upgraded systemd-254~rc2-1.fc39

### Actual behavior

kola basic test fails with systemd-254~rc2-1.fc39
 ```
 harness.go:1741: mach.Start() failed: machine 7bc8c78e-6a7c-401e-9e30-bbbf318f1706 entered emergency.target in initramfs
  FAIL, output in /home/jenkins/agent/workspace/build/tmp/kola-Off5m/kola-basic/nvme-true
```

### System details

Fedora CoreOS latest x86_64 Rawhide Version - https://builds.coreos.fedoraproject.org/browser?stream=rawhide&arch=x86_64

### Butane or Ignition config

_No response_

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by cgwalters**
Our dependency graph for firstboot is today kind of broken
because we only implicitly generate `sysroot.mount` as a side
effect of executing `ignition-ostree-mount-sysroot.sh`
(which is part of `ignition-ostree-mount-{firstboot,subsequent}-sysroot.service`)s.

As of relatively recently, we switched to auto-synthesizing the `root=`
karg after ignition (in `coroes-boot-edit.service`).  This means
that the systemd-fstab-generator will always win on subsequent boots,
so the systemd unit `-subsequent-sysroot.service` is dead code.

On firstboot, let's generate `sysroot.mount` out of the diskful
generator, which will ensure it always exists and fix our dependency
graph.

This should improve compatibility with newer systemd, and just
make everything saner too.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1527

### CODE REVIEW

File: overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh [hunk 1]
```diff
@@ -1,18 +0,0 @@
-#!/bin/bash
-set -euo pipefail
-
-# Note that on *new machines* this script is now only ever used on firstboot. On
-# subsequent boots, systemd-fstab-generator mounts /sysroot from the
-# root=UUID=... and rootflags=... kargs.
-
-# We may do a migration window at some point where older machines have these
-# kargs injected so that we can simplify the model further.
```
bgilbert, line 9: To be clear, are you planning to implement this migration so we don't break older machines?

cgwalters: Well spotted.  This is a good thing to discuss and we should explicitly test this.

However, I believe we don't need a migration window, we can just do this:

```
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
index acb7b3ce..1c400b24 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -41,10 +41,6 @@ add_requires() {
 }
 
-if ! $(cmdline_bool 'ignition.firstboot' 0); then
-    exit 0
-fi
-
 if ! is-live-image; then
     mkdir -p ${UNIT_DIR}
     # We want to generate sysroot.mount on ostree systems, but we don't want
@@ -60,6 +56,13 @@ Where=/sysroot
 EOF
         add_requires sysroot.mount initrd-root-fs.target
     fi
+fi
+
+if ! $(cmdline_bool 'ignition.firstboot' 0); then
+    exit 0
+fi
+
+if ! is-live-image; then
     # coreos-ignition-setup-user.service should depend on the boot device node
     # only on diskful boots
     mkdir -p "${UNIT_DIR}/coreos-ignition-setup-user.service.d"
```

Basically, the generator always generates `sysroot.mount` on every boot that doesn't have a `root=` karg, which includes old installs and our firstboot.

(In practice, I think it'd make sense to switch to injecting `root=/dev/disk/by-label/root` in our default kargs, but that's a potential followup)

cgwalters: (Repushed with that change)

bgilbert: That feels awkward to me.  We've generally been careful to distinguish Ignition code that runs on the first boot from code that runs on subsequent boots.

cgwalters:
I've verified that this fixes our compatiblity with the new `systemd-254~rc2-4.fc39.x86_64`.  I've added a Closes on https://github.com/coreos/fedora-coreos-tracker/issues/1527

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh
cgwalters: I think this new code is *much less* awkward and easier to understand than the previous two systemd units and the shell script to run `mount`.

But yes, any alternative suggestions?  Are you saying you think this should be a separate generator?  `coreos-sysroot-generator`?

bgilbert: Yeah, if this is the route we want to go (I haven't thought about it carefully) we should probably have a distinct generator outside of the `ignition` namespaces.

cgwalters: Is the vague intention that the things in 35coreos-ignition could be used by (ignition-but-not-ostree) systems?

Is the concern then the check for `ostree=`?  Because that's actually not necessary at all...nothing relates to ostree here.  I just kept it out of an excessive conservatism.  But it's highly likely today in practice that if our initramfs code somehow ends up in e.g. a (dnf, cloud-init/anaconda) system it's likely going to break something...

This is just about gluing together/integrating with the "fixed partition model" encouraged a bit by ignition/coreos-installer world, and we already have units in this dracut module like `coreos-boot-edit.service` which are similar logically.

bgilbert: When we moved this code out of Ignition into fedora-coreos-config, I think we implicitly assumed that it would be specific to CoreOS.  It'd be nice to minimize ostree dependencies, sure, but that's not my point here.

The concern is one of clarity: everything in this directory is currently only relevant to the first boot, and it's confusing to add a bit of code that runs on every boot.  So let's add the code somewhere that signals more clearly that it runs every time.

cgwalters:
Oh right fun, we have the `coreos-rootflags` handling which needs to fetch the rootfs flags from Ignition; it also has the special handling for `prjquota` for xfs.  I think that was part of the reason we had this as a dynamic script before.  Well...messy.  Perhaps the least hacky thing to do is have a service in the same place as `ignition-ostree-mount-firstboot-sysroot.service` that adds a drop-in for `sysroot.mount` with the `Options=` (if non-empty) and invokes `systemctl daemon-reload` too.

dustymabe:
> Oh right fun, we have the `coreos-rootflags` handling which needs to fetch the rootfs flags from Ignition; it also has the special handling for `prjquota` for xfs. I think that was part of the reason we had this as a dynamic script before. Well...messy. Perhaps the least hacky thing to do is have a service in the same place as `ignition-ostree-mount-firstboot-sysroot.service` that adds a drop-in for `sysroot.mount` with the `Options=` (if non-empty) and invokes `systemctl daemon-reload` too.

I picked this up and moved it forward in https://github.com/coreos/fedora-coreos-config/pull/2543. I really didn't want to `systemctl daemon-reload` in the middle of bootup so I found a way to keep the dynamic behavior using `EnvironmentFile` instead.

PR #2516, opened 2023-07-19
tests/manual: add script to launch a Hyper-V runner in AWS

bgilbert:
Most of the script is EC2 boilerplate to get a publicly routable IP address.  Unfortunately our test accounts have a pruner that deletes the default VPC, so we can't just use that.  For simplicity, create all our own infrastructure, and don't clean it up because the pruner will get it.

Most of the rest of the script is boilerplate for connecting to a Windows instance.  The only things that are Hyper-V specific are running on a bare-metal instance and installing Hyper-V.

### DIFF
--- a/tests/manual/create-hyperv-environment.sh
+++ b/tests/manual/create-hyperv-environment.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+# This script launches an AWS Windows instance suitable for running Hyper-V.
+# Ideally kola would support running automated Hyper-V tests by launching
+# such an instance and driving it over SSH.
+
+set -euo pipefail
+
+REGION="us-east-2"
+AZ="us-east-2a"
+AMI_DESC="Windows_Server-2022-English-Full-Base"
+# least expensive Windows metal option
+INSTANCE_TYPE="m5zn.metal"
+VPC="coreos-hyperv"
+SG="coreos-hyperv"
+INSTANCE="coreos-hyperv"
+KEY_PREFIX="coreos-hyperv"
+DISK_GB=100
+DISP_W=1280
+DISP_H=950
+
+if [ $# != 1 ]; then
+    echo "Usage: $0 <workdir>"
+    exit 1
+fi
+
+dir="$1"
+mkdir -p "$dir"
+
+export AWS_DEFAULT_REGION="$REGION"
+export AWS_DEFAULT_OUTPUT="json"
+
+for req in aws jq python3 xfreerdp ; do
+    if ! which "$req" &>/dev/null; then
+        echo "No $req command. Can't continue." >&2
+        exit 1
+    fi
+done
+
+set -x
+
+# find AMI
+ami=$(aws ssm get-parameters --names "/aws/service/ami-windows-latest/$AMI_DESC" | jq -r ".Parameters[].Value")
+
+# get or create VPC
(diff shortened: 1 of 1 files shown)


tests/manual/create-hyperv-environment.sh
```diff
@@ -0,0 +1,123 @@
...
+
+if [ $# != 1 ]; then
+    echo "Usage: $0 <workdir>"
+    exit 1
+fi
+
+dir="$1"
+mkdir -p "$dir"
+
+aws="aws --region=$REGION"
```
dustymabe: why not just set `export AWS_DEFAULT_REGION=$REGION` here?

tests/manual/create-hyperv-environment.sh
```diff
@@ -0,0 +1,123 @@
...
+
+set -x
+
+# find AMI
+ami=$($aws ssm get-parameters --names "/aws/service/ami-windows-latest/$AMI_DESC" --region "$REGION" | jq -r ".Parameters[].Value")
```
dustymabe: redundant?

```suggestion
ami=$($aws ssm get-parameters --names "/aws/service/ami-windows-latest/$AMI_DESC" | jq -r ".Parameters[].Value")
```

dustymabe: also - I had to `export AWS_DEFAULT_OUTPUT=json` to get this script to work looks like we assume `aws` is going to output json - maybe we should set it in the script?

tests/manual/create-hyperv-environment.sh
```diff
@@ -0,0 +1,123 @@
...
+igw=$($aws ec2 describe-internet-gateways --filters "Name=attachment.vpc-id,Values=$vpc" | jq -r ".InternetGateways[].InternetGatewayId")
+if [ -z "$igw" ]; then
+    igw=$($aws ec2 create-internet-gateway | jq -r .InternetGateway.InternetGatewayId)
+    $aws ec2 attach-internet-gateway --internet-gateway-id "$igw" --vpc-id "$vpc"
+fi
+
+# get or create subnet
+subnet=$($aws ec2 describe-subnets --filters "Name=vpc-id,Values=$vpc" "Name=availability-zone,Values=$AZ" | jq -r ".Subnets[].SubnetId")
+if [ -z "$subnet" ]; then
+    octet=$(python -c "print(ord(\"$AZ\"[-1]) - ord('a'))")
```
dustymabe: on my system I only have `python3` - maybe let's check our deps up front (`python`, `aws`, `jq`, etc) like is done over in https://github.com/coreos/fedora-coreos-config/blob/2aa59bf58e6a1c8e6c4b3a5f19067de3c52e9ad0/tests/manual/coreos-network-testing.sh#L240-L264

dustymabe (commented):
some initial comments. I'm still executing the process and going to try to run through the documenation proposed in https://github.com/coreos/fedora-coreos-docs/pull/545

dustymabe:
Once I got an instance up I was able to test https://github.com/coreos/fedora-coreos-docs/pull/545 and everything seemed to work OK.

bgilbert:
Thanks, updated!

dustymabe (approved):
LGTM - I'll merge once I run one more test with it.

PR #2517, opened 2023-07-20
kola: mark denylisted tests with 'warn: true'

nikita-dubrovskii:
This patch adds behavior described here: https://github.com/coreos/coreos-assembler/issues/3344

Requires: https://github.com/coreos/coreos-assembler/pull/3539

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -2,23 +2,29 @@
 # coreos-assembler to automatically skip some tests. For more information,
 # see: https://github.com/coreos/coreos-assembler/pull/866.
 - pattern: fcos.internet
+  warn: true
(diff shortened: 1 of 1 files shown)


dustymabe:
We don't want to make any modifications to the current denylist. If you're going off of what was described in https://github.com/coreos/coreos-assembler/issues/3344#issuecomment-1642556614 and https://github.com/coreos/coreos-assembler/issues/3344#issuecomment-1642566943 this isn't needed. i.e. when no `warn:` value exists we won't run the test (the current behavior). IMO we should close this PR.

After https://github.com/coreos/coreos-assembler/pull/3539 merges (or any other PR that fulfills https://github.com/coreos/coreos-assembler/issues/3344) we can then start to use it here in the config repo but really I would only use it for specific tests that we want the functionality for.

PR #2519, opened 2023-07-21
denylist: extend snooze for ext.config.ntp.* tests

aaradhak:
ext.config.ntp.chrony.dhcp-propagation &
ext.config.ntp.timesyncd.dhcp-propagation gets times out in rawhide build. It is not fixed in the upstream.
see: coreos/fedora-coreos-tracker#1508


mike-nguyen (approved):
LGTM

PR #2524, opened 2023-07-24
denylist: snooze root-reprovision on testing-devel `ppc64le`

marmijo:
These tests are now failing on testing-devel. Remove the rawhide field to snooze these tests on testing-devel as well.

see: https://github.com/coreos/fedora-coreos-tracker/issues/1489


c4rt0 (approved):
/lgtm

PR #2525, opened 2023-07-26
tests/validate-symlinks: Skip '/etc/pki/tls/'

travier:
```
Error: /etc/pki/tls/fips_local.cnf symlink to /etc/crypto-policies/back-ends/openssl_fips.config which does not exist
```

See: https://github.com/openshift/os/pull/1328

### DIFF
--- a/tests/kola/files/validate-symlinks
+++ b/tests/kola/files/validate-symlinks
@@ -32,8 +32,9 @@ list_broken_symlinks_skip=(
 if is_scos || is_rhcos; then
     rhcos_list=(
         '/etc/grub2-efi.cfg'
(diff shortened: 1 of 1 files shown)


travier:
```
$ rpm -qf /etc/pki/tls/fips_local.cnf
openssl-libs-3.0.7-24.el9.x86_64
```

Comes from https://gitlab.com/redhat/centos-stream/rpms/openssl/-/commit/d30c497ed1125a1e780be4cf1c713041353bc88a

travier:
Manually verified passing on SCOS.

marmijo (approved):
LGTM

PR #2526, opened 2023-07-26
ignition-ostree: make sure we don't mount /sysroot before transposefs

jmarrero:
https://issues.redhat.com/browse/OCPBUGS-16157

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-autosave-xfs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-autosave-xfs.service
@@ -5,6 +5,11 @@ After=ignition-disks.service
 # Avoid racing with UUID regeneration
 After=ignition-ostree-uuid-root.service
 After=ignition-ostree-growfs.service
+Before=initrd-root-fs.target
+# https://issues.redhat.com/browse/OCPBUGS-16157
+# On multipath systems mounting the /sysroot before
+# the ignition-ostree services causes the transpose to fai.
+Before=sysroot.mount
 Before=ignition-ostree-transposefs-restore.service
 OnFailure=emergency.target
 OnFailureJobMode=isolate
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-restore.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-restore.service
@@ -6,6 +6,11 @@ After=ignition-disks.service
 After=ignition-ostree-uuid-root.service
 After=ignition-ostree-growfs.service
 Before=ignition-ostree-mount-firstboot-sysroot.service
+# https://issues.redhat.com/browse/OCPBUGS-16157
+# On multipath systems mounting the /sysroot before
+# the ignition-ostree services causes the transpose to fai.
+Before=initrd-root-fs.target
+Before=sysroot.mount
 OnFailure=emergency.target
 OnFailureJobMode=isolate
 
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-save.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-save.service
@@ -7,6 +7,11 @@ ConditionKernelCommandLine=ostree
 ConditionPathIsDirectory=/run/ignition-ostree-transposefs
 # Any services looking at mounts need to order after this
 # because it causes device re-probing.
+# https://issues.redhat.com/browse/OCPBUGS-16157
+# On multipath systems mounting the /sysroot before
+# the ignition-ostree services causes the transpose to fai.
+Before=initrd-root-fs.target
+Before=sysroot.mount
 After=coreos-gpt-setup.service
 OnFailure=emergency.target
 OnFailureJobMode=isolate


dustymabe:
Without more details in the commit message or comments here I would classify this as a workaround. i.e. we know that when multipath is enabled the autosave-xfs stuff will fail, but we don't know why. 

Does the problem that autosave-xfs solve still happen on multipath? Should it work? Is there an underlying bug we need to investigate and figure out? Do we want it to work in the future? Should we add a test for it when it does work?

jmarrero:
Oh yeah this is a workaround, I am trying a possible real fix too. Just opened this in case we needed to provide them an urgent workaround. 

With this workaround:
Does the problem that autosave-xfs solve still happen on multipath? Yes 
Should it work? Yes
Is there an underlying bug we need to investigate and figure out? Yes (hopefully I can fix it before having to push the workaround)
Do we want it to work in the future? Yes
Should we add a test for it when it does work? Yes

dustymabe:
OK. Thanks for the context!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -139,13 +138,15 @@ should_autosave_rootfs() {
         return
     fi
     local agcount
-    eval $(xfs_info "${root_part}" | grep -o 'agcount=[0-9]*')
+    xfsinfo=$(xfs_info "${root_part}")
+    agcount=$(echo "$xfsinfo" | grep -o 'agcount=[0-9]*' | tr -dc '0-9')
```
dustymabe: might be easier to read if we use `cut` here:

```suggestion
    agcount=$(echo "$xfsinfo" | grep -o 'agcount=[0-9]*' | cut -d '=' -f 2)
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -205,6 +206,13 @@ check_and_set_secex_config() {
 # info from config, so just check cached one and reset to secex.ign now
 check_and_set_secex_config
```
dustymabe: add comment here for the below block of code?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -205,6 +206,13 @@ check_and_set_secex_config() {
...
 
+root_part=$(karg root)
+if test -z "${root_part}"; then
+    root_part=/dev/disk/by-label/root
```
dustymabe: maybe leave this set above (in the section where all the default values are set) and then only override it here if `karg root` is something.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
cgwalters: The use of `eval` here that was present before is IMO a giant red flag...I can't say I understand why he chose to do it, but explicitly setting the variable is way better!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -139,13 +138,15 @@ should_autosave_rootfs() {
...
     fi
     local agcount
-    eval $(xfs_info "${root_part}" | grep -o 'agcount=[0-9]*')
+    xfsinfo=$(xfs_info "${root_part}")
+    agcount=$(echo "$xfsinfo" | grep -o 'agcount=[0-9]*' | tr -dc '0-9')
     # Semi-arbitrarily chosen: this is roughly ~64G currently (based on initial
     # ag sizing at build time) which seems like a good rootfs size at which to
     # discriminate between "throwaway/short-lived systems" and "long-running
     # workload systems". It's not like XFS performance is way worse at 128.
+    echo "Filesystem agcount is $agcount"
```
jmarrero: Ohh this explains all the time I have spent trying to debug why sending this message to std_err would make it start failing again...

```
Jul 28 19:35:27 localhost ignition-ostree-transposefs[1896]: ++ echo 1
Jul 28 19:35:27 localhost ignition-ostree-transposefs[1890]: + should_autosave='Filesystem agcount is 159
Jul 28 19:35:27 localhost ignition-ostree-transposefs[1890]: 1'
```

should_autosave is returning garbage... so the script does not continue which means we are basically skipping it. Oh man... no fix yet...

dustymabe: right - because you added a `echo "Filesystem agcount is $agcount"` on line 147?

jmarrero:
Now this, ends up in the same place as the reported issue.
```
Displaying logs from failed units: ignition-ostree-transposefs-autosave-xfs.service
Jul 28 20:47:07 systemd[1]: ignition-ostree-transposefs-autosave-xfs.service: Main process exited, code=exited, status=32/n/a
Jul 28 20:47:07 ignition-ostree-transposefs[1986]: ++ realpath /dev/disk/by-label/dm-mpath-root
Jul 28 20:47:07 systemd[1]: ignition-ostree-transposefs-autosave-xfs.service: Failed with result 'exit-code'.
Jul 28 20:47:07 ignition-ostree-transposefs[1866]: + echo 'Mounting /dev/disk/by-label/dm-mpath-root ro (/dev/dm-4) to /var/tmp/mnt'
Jul 28 20:47:07 ignition-ostree-transposefs[1866]: Mounting /dev/disk/by-label/dm-mpath-root ro (/dev/dm-4) to /var/tmp/mnt
Jul 28 20:47:07 ignition-ostree-transposefs[1866]: + mkdir -p /var/tmp/mnt
Jul 28 20:47:07 ignition-ostree-transposefs[1866]: + mount -o ro /dev/disk/by-label/dm-mpath-root /var/tmp/mnt
Jul 28 20:47:07 systemd[1]: Failed to start Ignition OSTree: Autosave XFS Rootfs Partition.
Jul 28 20:47:07 ignition-ostree-transposefs[1988]: mount: /var/tmp/mnt: /dev/mapper/mpatha4 already mounted on /sysroot.
Jul 28 20:47:07 systemd[1]: ignition-ostree-transposefs-autosave-xfs.service: Triggering OnFailure= dependencies.
```
I continue debugging.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
jmarrero: done

jmarrero:
This boots correctly and does not seem to cause any errors, multipath devices are still being used when I check `df -h`

jmarrero:
Tested these 4 variants and all boot OK.

`cosa run -c --qemu-size=120G --qemu-memory=6096`

`cosa run -c --qemu-size=20G --qemu-memory=6096`

`cosa run -c --qemu-size=120G --qemu-memory=6096 --qemu-multipath=true --kargs='rd.multipath=default root=/dev/disk/by-label/dm-mpath-root rw'`

`cosa run -c --qemu-size=20G --qemu-memory=6096 --qemu-multipath=true --kargs='rd.multipath=default root=/dev/disk/by-label/dm-mpath-root rw'`

cgwalters:
> [2023-08-01T17:20:28.219Z] --- PASS: multipath.day1 (55.06s)

etc :tada: 

What we could do as a followup is a new kola test which expands to a big size with mpath but this LGTM

cgwalters (approved):

jmarrero:
Adding test issue so I don't forget: https://github.com/coreos/fedora-coreos-config/issues/2533. I will get to this ASAP.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-autosave-xfs.service
```diff
@@ -5,6 +5,11 @@ After=ignition-disks.service
 # Avoid racing with UUID regeneration
 After=ignition-ostree-uuid-root.service
 After=ignition-ostree-growfs.service
+Before=initrd-root-fs.target
```
cgwalters, line 8: OK now that I look I think I'm starting to understand this better.  Before this PR, we only had `Before=ignition-ostree-transposefs-restore.service` on this service, which is itself just `Before=ignition-ostree-mount-firstboot-sysroot.service`.

Yet *that* service is just a no-op if we have a `root` kernel argument.

So there were no constraints and we just end up racing.

So the multipath bits exposed this core race condition because it is done via a `root=` karg.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-restore.service
```diff
@@ -6,6 +6,11 @@ After=ignition-disks.service
 After=ignition-ostree-uuid-root.service
 After=ignition-ostree-growfs.service
 Before=ignition-ostree-mount-firstboot-sysroot.service
+# https://issues.redhat.com/browse/OCPBUGS-16157
+# On multipath systems mounting the /sysroot before
+# the ignition-ostree services causes the transpose to fai.
+Before=initrd-root-fs.target
+Before=sysroot.mount
```
cgwalters, line 13: This is 100% fine, but I'd note that normally `sysroot.mount` is itself `Before=initrd-root-fs.target`, so we could just be `Before=sysroot.mount` too.

Or maybe the simple way to say it is: anywhere we have `Before=ignition-ostree-mount-firstboot-sysroot.service` we *also* need `Before=sysroot.mount`.

PR #2528, opened 2023-07-28
tests/kola: ignore newlines in default network behavior change test

dustymabe:
Recently an updated glib2 caused failures here because of a whitespace change [1]. Let's ignore empty lines when comparing configurations.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1530


marmijo (approved):
LGTM

aaradhak (approved):

PR #2529, opened 2023-07-28
[testing-devel] overrides: pin kernel-6.3.12-200.fc38

marmijo:
We decided to pin the kernel the kernel in `testing-devel` instead of snoozing the repvosion tests while we track down a solution to https://github.com/coreos/fedora-coreos-tracker/issues/1489.

Pin kernel `kernel-6.3.12-200.fc38` in testing-devel and remove the snooze for the reprovision tests on `testing-devel`, which reverts commit 17ec963ed7b581e8ae461e48a8b37d7ec64c5fa2.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1489#issuecomment-1654472521

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -30,6 +30,8 @@
   snooze: 2023-08-04
   arches:
     - ppc64le
+  streams:
+    - rawhide
 - pattern: ext.config.ntp.chrony.dhcp-propagation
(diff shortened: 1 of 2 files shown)


dustymabe (approved):
LGTM

dustymabe:
CI is complaining about the kernel being downgraded but we can ignore that because the downgrade is intentional.

PR #2530, opened 2023-07-28
[rawhide] overrides: drop glib2-2.76.4-1.fc39 pin

dustymabe:
The change here was determined to be intentional and the test was updated to ignore newlines in the comparison of NM configs in https://github.com/coreos/fedora-coreos-config/pull/2528.

See https://github.com/coreos/fedora-coreos-tracker/issues/1530 for more context.


aaradhak (approved):
LGTM

PR #2534, opened 2023-08-01
[rhcos-4.14] ignition-ostree: make sure we don't mount /sysroot before transposefs

jmarrero:
When enabling multipath the ignition-ostree-mount-firstboot-sysroot service loses to the systemd's generator, which causes /sysroot to be mounted before we finish transposing the fs. This change makes sure we wait to mount the /sysroot until we are done. This was reported via: https://issues.redhat.com/browse/OCPBUGS-16157

Cherry-picked from #2526 & #2535


cgwalters:
Makes sense...I didn't think we'd branched yet for 4.14 though?  But the branch exists so we must have?

cgwalters (dismissed):

jmarrero:
converting to draft to pull in #2535 once it's merged too.

dustymabe (approved):
LGTM

PR #2535, opened 2023-08-02
ignition-ostree: remove not needed Before= checks

jmarrero:
Follow up per:
https://github.com/coreos/fedora-coreos-config/pull/2526#discussion_r1281205015

Tested and this works as expected. Thanks @cgwalters

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-save.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-save.service
@@ -7,11 +7,6 @@ ConditionKernelCommandLine=ostree
 ConditionPathIsDirectory=/run/ignition-ostree-transposefs
 # Any services looking at mounts need to order after this
 # because it causes device re-probing.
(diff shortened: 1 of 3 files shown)


cgwalters (approved):

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs-save.service
```diff
@@ -7,11 +7,6 @@ ConditionKernelCommandLine=ostree
 ConditionPathIsDirectory=/run/ignition-ostree-transposefs
 # Any services looking at mounts need to order after this
 # because it causes device re-probing.
-# https://issues.redhat.com/browse/OCPBUGS-16157
-# On multipath systems mounting the /sysroot before
-# the ignition-ostree services causes the transpose to fai.
-Before=initrd-root-fs.target
-Before=sysroot.mount
```
cgwalters, line 14: Yep and these ones were unnecessary (but also harmless) because this service is already `Before=ignition-disks.service` which is itself already

```
Before=initrd-root-device.target
Before=sysroot.mount
```

PR #2537, opened 2023-08-03
overrides: pin kernel-6.3.12-200 on `ppc64le` only

marmijo:
In the community meeting this week, we decided to pin `kernel-6.3.12-200` on `ppc64le` only for the next release cycle while we wait on a fix to the upstream issue: https://github.com/coreos/fedora-coreos-tracker/issues/1489.

Unpin the kernel from all of testing-devel and then add a `manifest-lock.overrides.ppc64le.yaml` to pin `kernel-6.3.12-200` on `ppc64le` only. This file is consumed when building on `ppc64le` and ignored on all other arches.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1489#issuecomment-1662811923


dustymabe (approved):
LGTM - thanks for working on this!

PR #2539, opened 2023-08-07
overlay/sysusers: remove `10-groups-basic.conf` which comes from `setup` package

HuijingHei:
As `setup` shipped its own sysusers.d entries,
see https://src.fedoraproject.org/rpms/setup/c/ff8169d809a5de20864deb6bb41c4b4104a3fb61?branch=f38 and https://src.fedoraproject.org/rpms/setup/c/9174b59bde57a991c284c10dac749d061f93aaad?branch=f38, we should remove the duplicated entries in `10-groups-basic.conf`.

Xerf to https://github.com/coreos/fedora-coreos-config/pull/1836


travier:
I've compared `/usr/lib/sysusers.d/20-setup-users.conf`, `/usr/lib/sysusers.d/20-setup-groups.conf` and https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/15fcos/usr/lib/sysusers.d/10-setup-basic.conf on a live system and the content mostly matches excepted for the games group case which links to https://github.com/systemd/systemd/issues/24217 which should be fixed now.

travier (approved):

HuijingHei:
Thanks @travier for the checking, and `nobody` mismatched issue is tracked by https://github.com/coreos/fedora-coreos-tracker/issues/1201

cgwalters:
Oh wow, I didn't realize we had *another* copy of that data!  Nice work to delete it!

cgwalters:
OK now here's the tricky thing though...this content also gets pulled into openshift/os pretty quickly in the development stream, so anything that lands here generally will need to work on RHEL 9.2 as well.

So the basic question is did https://github.com/systemd/systemd/pull/24249 make it into RHEL 9.2 yet?

cgwalters (approved):

cgwalters:
(If it didn't, we can still land this PR here, but re-add the file in openshift/os)

HuijingHei:
> this content also gets pulled into openshift/os pretty quickly in the development stream

The overlay sysuser config files are under dir `15fcos`, that will not included in `openshift/os` xref to https://github.com/openshift/os/tree/master/overlay.d, so this is safe for RHCOS.

> did [systemd/systemd#24249](https://github.com/systemd/systemd/pull/24249) make it into RHEL 9.2 yet?

Not yet, and sysusers configuration fragment for `setup` is also not included in RHEL 9.2 ([setup-2.13.7-9.el9](https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2311885)), maybe both will be added in future.

travier:
> Thanks @travier for the checking, and `nobody` mismatched issue is tracked by [coreos/fedora-coreos-tracker#1201](https://github.com/coreos/fedora-coreos-tracker/issues/1201)

The "legacy" nobody group is enforced by: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/15fcos/usr/lib/sysusers.d/00-coreos-nobody.conf

PR #2543, opened 2023-08-09
initramfs: Remove manual sysroot mounting

dustymabe:
This is a continuation of https://github.com/coreos/fedora-coreos-config/pull/2515. This PR now handles the coreos-rootflags dynamic behavior that was missing from https://github.com/coreos/fedora-coreos-config/pull/2515.

Please review commits individually.


dustymabe:
Did a new push here to fix CI failure. Anyone want to review this so we can unblock the systemd update in `rawhide`?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-rootflags.service
```diff
@@ -0,0 +1,27 @@
...
+# we must run before sysroot.mount
+Before=sysroot.mount
+# In case the disk was manipulated by Ignition and we need to pick
+# up mount options from Ignition config.
+After=ignition-disks.service
+# The script checks if the rootfs was reprovisioned so let's wait
+# until after that has completed if it was requested.
+After=ignition-ostree-transposefs-restore.service
+# Requires the root device to exist
+Requires=dev-disk-by\x2dlabel-root.device
```
cgwalters: Hmm, do we need this?  Remember that `Requires` are always pulled in even if the conditions don't apply.

I think eventually we want to get away from having labels at all after firstboot - today we switch to mounting via UUID, but we do keep the label (I think for backcompat reasons).

But ISTM we should eventually try to drop the labels.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -90,4 +90,5 @@ install() {
     inst_multiple jq blkid
     inst_script "$moddir/coreos-rootflags.sh" \
         "/usr/sbin/coreos-rootflags"
+    install_ignition_unit coreos-rootflags.service initrd-root-fs.target
```
cgwalters, line 93: I think it'd be better to have the generator enable this unit alongside generating `sysroot.mount`, instead of having it enabled statically with a lot of "don't run this" conditions.

Then we could keep the `Requires=` above and drop all the conditions.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-rootflags.service
```diff
@@ -0,0 +1,27 @@
...
+After=dev-disk-by\x2dlabel-root.device
+
+[Service]
+Type=oneshot
+StandardError=journal
+StandardOutput=append:/run/coreos-rootflags
+ExecStartPre=echo -n 'OPTIONS='
+ExecStart=/usr/sbin/coreos-rootflags
```
cgwalters, line 25: Perhaps cleaner to add `coreos-rootflags -o /run/coreos-rootflags`?

cgwalters (commented):
The denial changes and the rhel8 thing could land independently of this right?

Overall this seems fine, though we didn't end up addressing bgilbert's concern of trying to keep firstboot and subsequent boot code separated, right?  It seems valid, but I'm not sure of the cost/benefit of fixing that.

cgwalters (approved):
Technically I think this is OK to land as is with followups.

dustymabe:
> Technically I think this is OK to land as is with followups.

Sorry I had set it to automerge, so it merged as soon as the Green Check mark was waived.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-rootflags.service
dustymabe: I didn't think of adding a new option. Could have done it that way.

The one thing I was trying to do was make sure the output of the script didn't change because `rdcore` uses it too.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-rootflags.service
dustymabe: Yeah I dropped the `Requires` in a later commit in this PR, but wanted to keep it in a separate commit to illustrate further why the requires needed to be dropped.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
dustymabe: That's a good idea. I think I'll do a followup for it.

dustymabe:
> The denial changes and the rhel8 thing could land independently of this right?

The RHEL8 thing was completely independent.

The denial changes are there because this work unblocks us so that we can remove the systemd pin in `rawhide`. The denylist changes are needed for that to succeed.

> 
> Overall this seems fine, though we didn't end up addressing bgilbert's concern of trying to keep firstboot and subsequent boot code separated, right? It seems valid, but I'm not sure of the cost/benefit of fixing that.

Indeed. I tried to add a large comment block in the diskful generator to help alleviate this concern, but the fundamental undesirable structure is still there.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
dustymabe: followup in https://github.com/coreos/fedora-coreos-config/pull/2547

PR #2544, opened 2023-08-10
overlay.d/05core: support booting ISO completely from RAM

dustymabe:
It has been presented to us a use case where the user wants to actually "install" directly to the same device that was used to boot the live ISO image (i.e. the ISO was written to a block device and then booted from). The way the ISO boots today it mounts the ISO at /run/media/iso and then mounts the CoreOS rootfs from a file under that mountpoint. In this scenario /run/media/iso will be busy and can't be unmounted because the root filesystem of the live running system depends on it.

We can adjust the strategy slightly to get a system that runs completely   from memory (similar to live PXE systems) by just copying the file out of /run/media/iso and into memory before doing the rootfs mount, which will allow /run/media/iso to be unmounted before the installation commences.

This patchset adds a new coreos.liveiso.fromram option to allow requesting this new behavior. We could consider making this new behavior the default, but we'd need to consider the new RAM requirements.  For example, before this change on a 2G FCOS system the usage after booting looks like:

```
[core@localhost ~]$ free -m
               total        used        free      shared  buff/cache   available
Mem:            1953         163        1286         125         503        1519
Swap:              0           0           0
```

With this change (and the coreos.liveiso.fromram kernel arguemnt
specified) the usage looks like:

```
[core@localhost ~]$ free -m
               total        used        free      shared  buff/cache   available
Mem:            1953         173         688         808        1091         824
Swap:              0           0           0
```

The usage would scale with the size of the rootfs. i.e. RHCOS is larger so the memory requirements would be larger too.


dustymabe:
tests for this over in https://github.com/coreos/coreos-assembler/pull/3555

cgwalters:
In theory too what we could do is make the whole thing dynamic using something like devicemapper - basically we only need to move ourself to RAM when we go to do an install.  So we could e.g. set up [devicemapper raid](https://docs.kernel.org/admin-guide/device-mapper/dm-raid.html) in raid1 backed just by the ISO originally.

Then when we want to run coreos-installer, we create the ramfs and add it to the raid, then wait for the raid to sync, then unplug the disk.

Or maybe [dm switch](https://docs.kernel.org/admin-guide/device-mapper/switch.html) ?

But...dunno, maybe a kernel argument is just simpler.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-liveiso-persist-osmet.service
```diff
@@ -1,12 +1,12 @@
 [Unit]
 Description=Persist Osmet Files (ISO)
-DefaultDependencies=false
 ConditionPathExists=/run/ostree-live
 ConditionKernelCommandLine=coreos.liveiso
 RequiresMountsFor=/run/media/iso
-# on el8, the ISO is mounted by our own systemd unit
-After=run-media-iso-mount.service
 Before=initrd-switch-root.target
+# DefaultDependencies=true so this unit gets stopped on switchroot to
```
cgwalters, line 7: Setting this is redundant...I'm not objecting to keeping this comment, but it feels like it's just the `DefaultDependencies=false` case that requires documenting.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -7,6 +7,7 @@
 exec 1>/dev/kmsg; exec 2>&1
 
 command -v getarg >/dev/null || . /usr/lib/dracut-lib.sh
+command -v getargbool >/dev/null || . /usr/lib/dracut-lib.sh
```
cgwalters: I can't imagine that this is actually necessary since both come from the same place.  Did you copy it from somewhere?  I'd argue to delete the redundant import in the other place.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -90,32 +86,8 @@ else
 After=${isosrc_escaped}
 Requires=${isosrc_escaped}
 EOF
-    # Temporary hack for rhel8, where our mount unit can race with udev's
```
cgwalters, line 93: Hooray to deleting hacks!

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -127,13 +99,42 @@ Wants=systemd-udev-settle.service
 After=sysinit.target
 After=initrd-root-device.target
 Before=initrd-root-fs.target
+# This will make /run/media/iso get unmounted when no longer needed,
+# which will aid in the coreos.liveiso.fromram case.
+StopWhenUnneeded=true
```
cgwalters, line 103: Interesting.  Hmm, I wonder if ostree should be doing something similar.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -127,13 +99,42 @@ Wants=systemd-udev-settle.service
...
+        sysrootrequiresmountsfor=""
+        sysrootfsimg=/rootfs.img
+        # Add service to copy the rootfs.img from the ISO to memory
+        cat >"${UNIT_DIR}/coreos-liveiso-run-media-iso-cp-rootfsimg.service" <<EOF
+# Automatically generated by live-generator
+
+[Unit]
+RequiresMountsFor=/run/media/iso
+Before=sysroot.mount
+# DefaultDependencies=true so this unit gets stopped on switchroot to
```
cgwalters, line 124: Same comment as above

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -127,13 +99,42 @@ Wants=systemd-udev-settle.service
...
+# allow for /run/media/iso to get unmounted.
+DefaultDependencies=true
+
+[Service]
+Type=oneshot
+ExecStart=cp -v /run/media/iso/images/pxeboot/rootfs.img /rootfs.img
```
cgwalters, line 130: Hum, so `/rootfs.img` ends up in the initramfs tmpfs?  Maybe to be a bit more explicit, something like `/tmp/rootfs.img`?

I don't think `/tmp` gets preserved across switchroot, only `/run`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-liveiso-persist-osmet.service
```diff
@@ -1,12 +1,12 @@
 [Unit]
 Description=Persist Osmet Files (ISO)
-DefaultDependencies=false
 ConditionPathExists=/run/ostree-live
 ConditionKernelCommandLine=coreos.liveiso
```
cgwalters, line 4: (Unrelated to this PR, but we could also have the generator enable this service)

cgwalters (commented):
An interesting thing here is this becomes an API we need to maintain ~forever...but so far I don't think our initramfs glue has really had any public APIs?  And hence there's no explicit place to document this in this repo, we'd need to add it to fedora-coreos-docs.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
dustymabe: you are right. I mostly put this in place to be explicit about the functions we were using. I can delete this if it's too much.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-liveiso-persist-osmet.service
dustymabe: Yeah. I struggled here. We default in most of these early boot units to `DefaultDependencies=false`. I figured it would be easy for someone to come in and add that back if I didn't state explicitly why we need it to be true.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
dustymabe: 🎉

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
dustymabe: yep. I'll delete if we delete the other one. LMK what you think in https://github.com/coreos/fedora-coreos-config/pull/2544#discussion_r1293815377

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
dustymabe: I think what I was trying to mimic here is what is done for `coreos.live.rootfs_url`. In that case the rootfs is streamed from the remote [via `rdcore`](https://github.com/coreos/fedora-coreos-config/blob/377823b0efd146b78153024abfdd85cc9b07eb61/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh#L71-L72) to `/root.squashfs` and then [mounted from there](https://github.com/coreos/fedora-coreos-config/blob/377823b0efd146b78153024abfdd85cc9b07eb61/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator#L76).

Since that was the model already used I assumed it was safest just to cp `rootfs.img` into `/`.

dustymabe:
> An interesting thing here is this becomes an API we need to maintain ~forever...but so far I don't think our initramfs glue has really had any public APIs? And hence there's no explicit place to document this in this repo, we'd need to add it to fedora-coreos-docs.

I imagine we'd document it wherever `coreos.live.rootfs_url` is documented.

dustymabe:
> In theory too what we could do is make the whole thing dynamic using something like devicemapper - basically we only need to move ourself to RAM when we go to do an install. So we could e.g. set up [devicemapper raid](https://docs.kernel.org/admin-guide/device-mapper/dm-raid.html) in raid1 backed just by the ISO originally.
> 
> Then when we want to run coreos-installer, we create the ramfs and add it to the raid, then wait for the raid to sync, then unplug the disk.
> 
> Or maybe [dm switch](https://docs.kernel.org/admin-guide/device-mapper/switch.html) ?
> 
> But...dunno, maybe a kernel argument is just simpler.

Yeah. In this kernel arg scenario `coreos-installer` itself doesn't need to know or care, which I think makes things more simple.

When it comes to modifying `coreos-installer`, I think we should still do the `kexec` option in https://github.com/coreos/coreos-installer/pull/791. However, this PR would still have the benefit of not having to do a second "install" boot to get to a fully "fromram" system.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-liveiso-persist-osmet.service
cgwalters: I don't think we have anyone who wakes up, grabs their coffee and decides to just start changing systemd unit ordering for fun :smile: 

Anyways, no big deal, I'm fine as is too.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
cgwalters: Seems deletion worthy yes

cgwalters (dismissed):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
dustymabe: deleted!

dustymabe:
ok did a new push with the one requested deletion. I also rebased on top of latest `testing-devel` to get a full test run on top of the changes from https://github.com/coreos/fedora-coreos-config/pull/2543

PR #2547, opened 2023-08-14
overlay.d/05core: enable coreos-rootflags.service when needed

dustymabe:
Enabling coreos-rootflags.service in the generator makes it so that we can only enable it on boots where it is needed and we won't see superfluous messsages about it "not starting due to a condition" in the journal.


cgwalters (dismissed):

dustymabe:
Ok I fixed a bug. I think this should be good now.

cgwalters (approved):

PR #2550, opened 2023-08-15
#1534 Change: Allow Removal of tzdata 

### LINKED ISSUE #1534: F39 Change: Allow Removal of tzdata
For the [proposed Fedora 39 changes](https://github.com/coreos/fedora-coreos-tracker/issues/1491) we decided the following change needed action:

-  [Allow Removal of tzdata](https://fedoraproject.org/wiki/Changes/AllowRemovalOfTzdata)
     - Allow the removal of tzdata especially on containers in order to minimize size.
     - Tracking bug:

yasminvalim:
This PR is adding the tzdata package in Fedora CoreOS.  :package:

<br class="Apple-interchange-newline">To know more about this change, please check this [documentation ](https://fedoraproject.org/wiki/Changes/AllowRemovalOfTzdata).

### DIFF
--- a/manifests/system-configuration.yaml
+++ b/manifests/system-configuration.yaml
@@ -39,6 +39,8 @@ packages:
   - stalld
   # Ignition aware SSH key management
   - ssh-key-dir
+  # Allow for configuring different timezones
+  - tzdata
 
 postprocess:
   # Mask systemd-repart. Ignition is responsible for partition setup on first


dustymabe:
In this case maybe let's not include a link to the issue in the comment:

```suggestion
  # Allow for configuring different timezones
  - tzdata
```

But we do want to include a link to the issue in the git message.

Also, let's put this over in `system-configuration.yaml` rather than here since it's not specific to Fedora CoreOS. i.e. we'll want this same thing in RHCOS too when the tzdata dep drops out there.

dustymabe:
Can we get the commit message updated? Something like:

```
manifests: name tzdata package in packages list

The tzdata package is undergoing a change [1] to allow it
to not be required by some other packages (like glibc). We
don't want the tzdata package to drop out of Fedora CoreOS
so let's name the package here in our manifest.

[1] https://fedoraproject.org/wiki/Changes/AllowRemovalOfTzdata

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1534
```

yasminvalim:
Done! Thanks for your review. :blush:

dustymabe (approved):
LGTM

PR #2551, opened 2023-08-16
denylist: add denials for tests that layer packages on rawhide

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/1549


HuijingHei (approved):
LGTM

PR #2552, opened 2023-08-16
denylist: bump snooze for `ext.config.kdump.crash` on aarch64

gursewak1997:
This test is still failing.
See: coreos/fedora-coreos-tracker#1430


dustymabe (approved):
LGTM

PR #2556, opened 2023-08-17
tmpfiles: Move root bash files copy to tmpfiles.d

### LINKED ISSUE #136: /etc/skel files in /root should be copied by tmpfiles

HuijingHei:
Fixes https://github.com/coreos/fedora-coreos-config/issues/136

### DIFF
--- a/overlay.d/05core/usr/lib/tmpfiles.d/root-bash.conf
+++ b/overlay.d/05core/usr/lib/tmpfiles.d/root-bash.conf
@@ -1,0 +1,5 @@
+# This really should be C! instead of L, but https://github.com/systemd/systemd/issues/12467 prevents that.
+# TODO: switch to C! when https://github.com/systemd/systemd/pull/12750 makes it into fcos
+L	/var/roothome/.bashrc	-	-	-	-	/usr/etc/skel/.bashrc
+L	/var/roothome/.bash_profile	-	-	-	-	/usr/etc/skel/.bash_profile
+L	/var/roothome/.bash_logout	-	-	-	-	/usr/etc/skel/.bash_logout
--- a/overlay.d/05core/usr/lib/tmpfiles.d/root-bash.conf
+++ b/overlay.d/05core/usr/lib/tmpfiles.d/root-bash.conf
@@ -1,5 +1,4 @@
-# This really should be C! instead of L, but https://github.com/systemd/systemd/issues/12467 prevents that.
-# TODO: switch to C! when https://github.com/systemd/systemd/pull/12750 makes it into fcos
-L	/var/roothome/.bashrc	-	-	-	-	/usr/etc/skel/.bashrc
-L	/var/roothome/.bash_profile	-	-	-	-	/usr/etc/skel/.bash_profile
-L	/var/roothome/.bash_logout	-	-	-	-	/usr/etc/skel/.bash_logout
+# Use C! according to https://github.com/coreos/fedora-coreos-config/pull/137/files#diff-4c7caeddac3f43acdf59b41ffa4e3533bb846d2e3bfa31f885a65e9598a2c4a1R1-R2
+C!	/var/roothome/.bashrc	-	-	-	-	/usr/etc/skel/.bashrc
+C!	/var/roothome/.bash_profile	-	-	-	-	/usr/etc/skel/.bash_profile
+C!	/var/roothome/.bash_logout	-	-	-	-	/usr/etc/skel/.bash_logout
--- a/tests/kola/files/root-bash
+++ b/tests/kola/files/root-bash
@@ -1,0 +1,17 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   description: Verify /var/roothome/.bash* files exist.
+
+# See https://bugzilla.redhat.com/show_bug.cgi?id=1193590
+set -xeuo pipefail
+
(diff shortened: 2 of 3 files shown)


HuijingHei:
Also do testing on rhcos, which also works.
```
[root@cosa-devsh ~]# ls -al /var/roothome/
total 16
drwx------.  3 root root   74 Aug 22 05:54 .
drwxr-xr-x. 24 root root 4096 Aug 22 05:54 ..
-rw-r--r--.  1 root root   18 Aug 22 05:54 .bash_logout
-rw-r--r--.  1 root root  141 Aug 22 05:54 .bash_profile
-rw-r--r--.  1 root root  492 Aug 22 05:54 .bashrc

[root@cosa-devsh ~]# rm -rf .bash*

[root@cosa-devsh ~]# rpm-ostree usroverlay 

[root@cosa-devsh ~]# cat /usr/lib/tmpfiles.d/root-bash.conf
# Find `C` in https://github.com/systemd/systemd/blob/main/man/tmpfiles.d.xml
C	/var/roothome/.bashrc	-	-	-	-	/usr/etc/skel/.bashrc
C	/var/roothome/.bash_profile	-	-	-	-	/usr/etc/skel/.bash_profile
C	/var/roothome/.bash_logout	-	-	-	-	/usr/etc/skel/.bash_logout

[root@cosa-devsh ~]# systemd-tmpfiles --create --remove --exclude-prefix=/dev
/usr/lib/tmpfiles.d/tmp.conf:12: Duplicate line for path "/var/tmp", ignoring.
/usr/lib/tmpfiles.d/var.conf:14: Duplicate line for path "/var/log", ignoring.
/usr/lib/tmpfiles.d/var.conf:19: Duplicate line for path "/var/cache", ignoring.
/usr/lib/tmpfiles.d/var.conf:21: Duplicate line for path "/var/lib", ignoring.
/usr/lib/tmpfiles.d/var.conf:23: Duplicate line for path "/var/spool", ignoring.
"/home" already exists and is not a directory.
"/srv" already exists and is not a directory.
"/root" already exists and is not a directory.
[root@cosa-devsh ~]# ls -alh
total 16K
drwx------.  3 root root   74 Aug 22 05:56 .
drwxr-xr-x. 24 root root 4.0K Aug 22 05:54 ..
-rw-r--r--.  1 root root   18 Jan  1  1970 .bash_logout
-rw-r--r--.  1 root root  141 Jan  1  1970 .bash_profile
-rw-r--r--.  1 root root  492 Jan  1  1970 .bashrc
```

tests/kola/files/root-bash
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   description: Verify /var/roothome/.bash* files are existed.
```
travier:
```suggestion
##   description: Verify /var/roothome/.bash* files exist.
```

tests/kola/files/root-bash
```diff
@@ -0,0 +1,17 @@
...
+
+# See https://bugzilla.redhat.com/show_bug.cgi?id=1193590
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+for bashfile in .bashrc .bash_profile .bash_logout
+do 
+    if ! test -f "/var/roothome/${bashfile}"; then
+        fatal "Error: not find /var/roothome/${bashfile}"
```
travier:
```suggestion
        fatal "Error: could not find /var/roothome/${bashfile}"
```

overlay.d/05core/usr/lib/tmpfiles.d/root-bash.conf
```diff
@@ -0,0 +1,4 @@
+# Find `C` in https://github.com/systemd/systemd/blob/main/man/tmpfiles.d.xml
```
travier: https://github.com/coreos/fedora-coreos-config/pull/137/files#diff-4c7caeddac3f43acdf59b41ffa4e3533bb846d2e3bfa31f885a65e9598a2c4a1R1-R2 mentions that is should be `C!` and I think that's what we want if I understand correctly. From https://www.freedesktop.org/software/systemd/man/systemd-tmpfiles-setup.service.html#--boot:

```
Also execute lines with an exclamation mark. Lines that are not safe to be executed on a running system
may be marked in this way. systemd-tmpfiles is executed in early boot with --boot specified and will execute
those lines. When invoked again later, it should be called without --boot.
```

travier (changes requested):
Some comments.

Let's also add a "Fixes: https://github.com/coreos/fedora-coreos-config/issues/136" ?

overlay.d/05core/usr/lib/tmpfiles.d/root-bash.conf
HuijingHei: Thanks @travier, saw the comment that you mentioned, but I can not find it in definitions, let me try using `C!`.

travier: I found it in https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html#Type%20Modifiers

cgwalters (approved):

travier (approved):

PR #2560, opened 2023-08-21
[stable] tree: promote changes from testing at 07b1de960009ca66245f71a498c380e53581346b

coreosbot-releng:
Triggered by @marmijo in https://github.com/coreos/fedora-coreos-streams/issues/747.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -15,19 +15,12 @@
     - ppc64le
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1430
-  # snooze: 2023-08-02 (disabled on promotion)
+  # snooze: 2023-08-16 (disabled on promotion)
   arches:
     - aarch64
-- pattern: ext.config.kdump.crash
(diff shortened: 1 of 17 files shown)


dustymabe:
CI failed with:

```
FAIL: iso-offline-install-fromram.4k.uefi
```

Can we add a 2nd commit to this PR that denylists that test? That new feature is only in latest `testing-devel` so it won't pass this round.

marmijo:
> Can we add a 2nd commit to this PR that denylists that test? 

Sure, I'll add that now

dustymabe:
actually there are a few more - all added in https://github.com/coreos/coreos-assembler/commit/0955101533acae9b72f155136f5c7a9d1fdbbc00 

```
iso-offline-install-fromram.4k.uefi
iso-offline-install-fromram.ppcfw
iso-offline-install-fromram.uefi
```

marmijo:
I added the three tests to the denylist in a seperate commit. `Check promotion diffs` is now failing because the changes were unexpected. We should be good to ignore this.

dustymabe (approved):
LGTM

PR #2561, opened 2023-08-23
denylist: mark snoozed tests as warn: true

dustymabe:
This is a new feature that was added recently in
https://github.com/coreos/coreos-assembler/pull/3539.

Now when a snooze expires we can configure it to warn us and continue to build rather than marking the entire run as a failure.


nikita-dubrovskii (approved):

marmijo (commented):
LGTM

PR #2563, opened 2023-08-23
test: Collection of cleanups & Action for whitespace check

travier:
test: Disable 'source' ShellCheck warnings

---

test: Remove unneeded whitespace at EOL

---

overlay.d/usr/libexec/coreos-check-modularity: Remove whitespace at EOL

---

ci: Add action to check for whitespace at EOL

While usually harmless, this can create issue in some edge cases and
creates noise when looking at the code in editors that highlight
unneeded whitespace.

### DIFF
--- a/tests/kola/binfmt/qemu
+++ b/tests/kola/binfmt/qemu
@@ -15,7 +15,8 @@
 
 set -xeuo pipefail
 
-. $KOLA_EXT_DATA/commonlib.sh
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
 
 case "$(arch)" in
     aarch64|ppc64le|s390x)
--- a/.github/workflows/find-whitespace.yml
+++ b/.github/workflows/find-whitespace.yml
@@ -1,0 +1,22 @@
+name: Find whitespace
+
+on:
+  pull_request:
+    branches: [testing-devel]
+
+permissions:
+  contents: read
+
+jobs:
+  find-whitespace:
+    name: Find whitespace
+    runs-on: ubuntu-latest
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+      # https://github.com/actions/checkout/issues/760
+      - name: Mark git checkout as safe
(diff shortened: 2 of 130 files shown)


travier:
Maybe we should make a GitHub Action that checks for whitespace at EOL.

tests/kola/binfmt/qemu
```diff
@@ -15,7 +15,8 @@
 
 set -xeuo pipefail
 
-. $KOLA_EXT_DATA/commonlib.sh
+# shellcheck disable=SC1091
```
dustymabe, line 18: Reading over:

- https://www.shellcheck.net/wiki/SC1090
- https://www.shellcheck.net/wiki/SC1091

maybe we should try to tell shellcheck where the file is?

Would this work?

```
# shellcheck source=data/commonlib.sh
```

dustymabe (dismissed):
one optional comment - LGTM

tests/kola/binfmt/qemu
travier: I though about this too, but I'm not sure how the path and the ENV variable are handled. Will give this a try.

travier: and right now it does not work with the default shellcheck / syntastic (vim plugin) config as it does not enable `-x` / external sources for shellcheck (we do in the script here in the CI).

travier: OK, I've done some tests and it does not take the path relative to the file but relative to where you run the shellcheck command.

travier: So we could do `# shellcheck source=tests/kola/data/commonlib.sh`

travier: But that would not work for "local checks", only CI checks.

travier: So unfortunately I can not find a good solution for all cases. We already verify the lib file in CI so I'd say this is not strictly needed.

travier: How to enable external sources for the syntastic vim plugin: https://github.com/vim-syntastic/syntastic/issues/1957#issuecomment-347658033

dustymabe (approved):
LGTM

PR #2564, opened 2023-08-24
Collection of ShellCheck fixes + Make tests/logrotate-service platform independent

travier:
tests/logrotate-service: Make platform independent

This should pass on all platforms if it passes on one.

---

tests/kola/data/commonlib.sh: ShellCheck fixes

---

overlay.d/*: ShellCheck fix for /proc/cmdline reading

Fix ShellCheck warning:
```
In ./overlay.d/05core/usr/lib/coreos/generator-lib.sh line 11:
    local cmdline=( $(</proc/cmdline) )
                    ^---------------^ SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).
```

### DIFF
--- a/overlay.d/05core/usr/lib/coreos/generator-lib.sh
+++ b/overlay.d/05core/usr/lib/coreos/generator-lib.sh
@@ -8,7 +8,7 @@ UNIT_DIR="${1:-/tmp}"
 
 have_karg() {
     local arg="$1"
-    local cmdline=( $(</proc/cmdline) )
+    IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
     local i
     for i in "${cmdline[@]}"; do
         if [[ "$i" =~ "$arg=" ]]; then
@@ -20,7 +20,7 @@ have_karg() {
 
 karg() {
     local name="$1" value="${2:-}"
-    local cmdline=( $(</proc/cmdline) )
+    IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
     for arg in "${cmdline[@]}"; do
         if [[ "${arg%%=*}" == "${name}" ]]; then
             value="${arg#*=}"
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -1,3 +1,4 @@
+# shellcheck shell=bash
 # This is a common library created for the ok & fatal function and symlinks
 # added to the data/ in each directory
 
@@ -12,12 +13,13 @@ fatal() {
 
 get_ipv4_for_nic() {
     local nic_name=$1
-    local ip=$(ip -j addr show ${nic_name} | jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local')
+    local ip
+    ip="$(ip -j addr show "${nic_name}" | jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local')"
     if [ -z "$ip" ]; then
         echo "Error: can not get ip for ${nic_name}"
         exit 1
     fi
-    echo $ip
+    echo "$ip"
 }
 
 get_fcos_stream() {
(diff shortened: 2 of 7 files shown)


travier:
Split from previous PR to make it easier to review

overlay.d/05core/usr/lib/coreos/generator-lib.sh
```diff
@@ -8,7 +8,7 @@ UNIT_DIR="${1:-/tmp}"
 
 have_karg() {
     local arg="$1"
-    local cmdline=( $(</proc/cmdline) )
+    IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
```
cgwalters, line 11: This looks like Perl-level modem line noise to me, but I'll believe it's the right thing to do...

cgwalters (approved):

overlay.d/05core/usr/lib/coreos/generator-lib.sh
travier: Agree it's not great. I had made similar "fixes" in https://github.com/coreos/coreos-installer/pull/999 & https://github.com/coreos/coreos-installer/pull/1194 in the past so this is a copy / paste.

travier:
Notes for future reference:
- https://www.shellcheck.net/wiki/SC2207 The ShellCheck page with the recommendation this change is based on
- Similar changes (in case we decide to change one of those, we should fix all of them):
  - https://github.com/coreos/coreos-installer/pull/999
  - https://github.com/coreos/coreos-installer/pull/1194
- Other "simpler" potential solutions:
  - https://stackoverflow.com/questions/11393817/read-lines-from-a-file-into-a-bash-array
  - https://stackoverflow.com/questions/30988586/creating-an-array-from-a-text-file-in-bash
  - https://unix.stackexchange.com/questions/80362/what-does-mean
  - https://unix.stackexchange.com/questions/76402/command-line-instead-of

tests/kola/data/commonlib.sh
```diff
@@ -12,12 +13,13 @@ fatal() {
 
 get_ipv4_for_nic() {
     local nic_name=$1
-    local ip=$(ip -j addr show ${nic_name} | jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local')
+    local ip
+    ip="$(ip -j addr show "${nic_name}" | jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local')"
```
dustymabe, line 17: Are we splitting this because the line is too long? I'd almost rather line break after the | if that was the case:

```
local ip="$(ip -j addr show "${nic_name}" | \
            jq -r '.[0].addr_info | map(select(.family == "inet")) | .[0].local')"
```

tests/kola/data/commonlib.sh
```diff
@@ -65,7 +67,7 @@ is_scos() {
     [ "${ID}" == "scos" ] && [ "${VARIANT_ID}" == "coreos" ]
 }
 
-cmdline=( $(</proc/cmdline) )
+IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
```
dustymabe, line 70: Should this be a part of the other commit?

dustymabe: oh I guess this commit is specific to the library. nvm

dustymabe (approved):

tests/kola/data/commonlib.sh
travier: It's for this rule: https://www.shellcheck.net/wiki/SC2155

tests/kola/data/commonlib.sh
travier: Indeed. I'll do that if we need to do another change to avoid wasting CI on a commit re-split.

PR #2565, opened 2023-08-24
transposefs: Only autosave-xfs for much larger filesystems

cgwalters:
The change in https://github.com/coreos/fedora-coreos-config/pull/2320
has been very problematic for OpenShift because our default node
configuration is *always* over the threshold, and that causes
significant latency on instance provisioning.

Experimentally bumping to 400 allocation groups, which is about 700GiB.
This is comfortably about the default OpenShift node root disk sizes,
and returns us to the prior status quo.

While we're here, rework the logging a bit so that we *always*
log the `agcount` for debugging purposes.

---

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -139,17 +139,24 @@ should_autosave_rootfs() {
         return
     fi
     local agcount
+    # This runs xfs_info on the unmounted filesystem, because mounting an
+    # XFS filesystem that has grown an excessive number of allocation groups
+    # can be very slow.
     eval $(xfs_info "${root_part}" | grep -o 'agcount=[0-9]*')
-    # Semi-arbitrarily chosen: this is roughly ~64G currently (based on initial
-    # ag sizing at build time) which seems like a good rootfs size at which to
-    # discriminate between "throwaway/short-lived systems" and "long-running
-    # workload systems". It's not like XFS performance is way worse at 128.
-    if [ "$agcount" -lt 128 ]; then
-        echo "Filesystem agcount is $agcount; skipping" >&2
+    # This is roughly ~700GiB currently (based on initial ag sizing at build time)
+    # which ensures we grow only on "large" root filesystems.
+    # Specifically for e.g. OpenShift, this ensures we don't reprovision on default
+    # worker node root filesystems.
+    local threshold
+    threshold=400
+    if [ "$agcount" -lt "${threshold}" ]; then
+        echo "autosave-xfs: ${root_part} agcount=$agcount is lower than threshold=${threshold}" >&2
         echo 0
         return
+    else
+        echo "autosave-xfs: ${root_part} agcount=$agcount meets threshold=${threshold}" >&2
+        echo 1
     fi
-    echo 1
 }
 
 ensure_zram_dev() {
--- a/tests/kola/root-reprovision/autosave-xfs/test.sh
+++ b/tests/kola/root-reprovision/autosave-xfs/test.sh
@@ -2,8 +2,8 @@
 ## kola:
 ##   # This test reprovisions the rootfs automatically.
 ##   tags: "platform-independent reprovision"
-##   # Trigger automatic XFS reprovisioning
-##   minDisk: 100
+##   # Trigger automatic XFS reprovisioning (heuristic)
+##   minDisk: 1000
 ##   # Root reprovisioning requires at least 4GiB of memory.
 ##   minMemory: 4096
 ##   # This test includes a lot of disk I/O and needs a higher
@@ -11,6 +11,7 @@
 ##   timeoutMin: 15
 ##   description: Verify the root reprovision with XFS
 ##     on large disk triggers autosaved.
+##     This test is meant to cover ignition-ostree-transposefs-autosave-xfs.service
 
 set -xeuo pipefail
 
@@ -20,10 +21,13 @@ set -xeuo pipefail
 if [ ! -f /run/ignition-ostree-autosaved-xfs.stamp ]; then
     fatal "expected autosaved XFS"
 fi
+# Verify we printed something about the agcount
+journalctl -u ignition-ostree-transposefs-autosave-xfs.service --grep=agcount
 ok "autosaved XFS on large disk"
 
 eval $(xfs_info / | grep -o 'agcount=[0-9]*')
-if [ "$agcount" -gt 4 ]; then
-    fatal "expected agcount of at most 4, got ${agcount}"
+expected=4
+if [ "$agcount" -gt "$expected" ]; then
+    fatal "expected agcount of at most ${expected}, got ${agcount}"
 fi
 ok "low agcount on large disk"
--- a/tests/kola/root-reprovision/luks/autosave-xfs/test.sh
+++ b/tests/kola/root-reprovision/luks/autosave-xfs/test.sh
@@ -10,9 +10,10 @@
 ##   # timeout value than the default.
 ##   timeoutMin: 15
 ##   # Trigger automatic XFS reprovisioning
-##   minDisk: 100
+##   minDisk: 1000
 ##   description: Verify the root reprovision with XFS and TPM
 ##     on large disk triggers autosaved.
+##     This test is meant to cover ignition-ostree-transposefs-autosave-xfs.service
 
 set -xeuo pipefail
 
@@ -27,8 +28,9 @@ if [ -z "${AUTOPKGTEST_REBOOT_MARK:-}" ]; then
     ok "autosaved XFS on large disk"
 
     eval $(xfs_info / | grep -o 'agcount=[0-9]*')
-    if [ "$agcount" -gt 4 ]; then
-        fatal "expected agcount of at most 4, got ${agcount}"
+    expected=4
+    if [ "$agcount" -gt "${expected}" ]; then
+        fatal "expected agcount of at most ${expected}, got ${agcount}"
     fi
     ok "low agcount on large disk"
 fi


cgwalters:
Draft since this could use some more testing

cgwalters:
Downstream: https://issues.redhat.com/browse/OCPBUGS-16724

cgwalters:
Here's a script I wrote to analyze xfs AG counts as the system is grown:

```
#!/bin/bash
set -euo pipefail
initial=$1
shift
grown=$1
shift

runv() {
    echo "$ $@"
    "$@"
} 

tmpd=$(mktemp -d -p /var/tmp)
target="${tmpd}/backing-file"
runv truncate -s "$initial" "$target"
runv mkfs.xfs "${target}"
runv truncate -s "$grown" "$target"
mntpath="${tmpd}/mnt"
mkdir "${mntpath}"
runv mount -o loop -t xfs "${target}" "${mntpath}"
runv xfs_growfs "${mntpath}" 1>/dev/null
runv xfs_info "${mntpath}"
umount "${mntpath}"
rm ${tmpd} -rf
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -139,13 +139,20 @@ should_autosave_rootfs() {
...
-    # workload systems". It's not like XFS performance is way worse at 128.
-    if [ "$agcount" -lt 128 ]; then
-        echo "Filesystem agcount is $agcount; skipping" >&2
+    # workload systems".  Specifically for e.g. OpenShift, it's below the default node
+    # sizes.
+    local threshold
+    threshold=400
+    echo "${root_part} agcount=$agcount" >&2
+    if [ "$agcount" -lt "${threshold}" ]; then
+        echo "agcount=$agcount is lower than threshold=${threshold}"
```
dustymabe: The other echo in this function prints to stderr:

```suggestion
        echo "agcount=$agcount is lower than threshold=${threshold}" >&2
```

Also it might be nice to give one more piece of info:

```suggestion
        echo "agcount=$agcount is lower than the autosave-xfs threshold=${threshold}" >&2
```

tests/kola/root-reprovision/autosave-xfs/test.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 ## kola:
 ##   # This test reprovisions the rootfs automatically.
+##   # Tests: ignition-ostree-transposefs-autosave-xfs.service
```
dustymabe: Maybe this could go in the `description:` field below.

tests/kola/root-reprovision/autosave-xfs/test.sh
```diff
@@ -1,9 +1,10 @@
...
 ##   tags: "platform-independent reprovision"
-##   # Trigger automatic XFS reprovisioning
-##   minDisk: 100
+##   # Trigger automatic XFS reprovisioning (heuristic)
+##   minDisk: 1000
```
dustymabe, line 6: Should this be 1000 or something closer to the 700GiB threshold so that we can properly test that something right above the threshold gets reprovisioned?

Making it lower might make CI go faster, but I'm not sure.

Probably need to update the luks autosave-xfs test too.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -139,13 +139,20 @@ should_autosave_rootfs() {
         return
     fi
     local agcount
+    # We want to run xfs_info on the unmounted filesystem, because actually just
+    # even mounting an XFS filesystem that has grown an excessive number of allocation groups
+    # can be very slow.
```
dustymabe: This comment confuses me. You didn't change the following line of code so you were trying to add context? Is the filesystem unmounted or are you saying we want to operate on an unmounted filesystem in the future?

Maybe this would help:

```suggestion
    # We run xfs_info on the unmounted filesystem, because actually just
    # even mounting an XFS filesystem that has grown an excessive number of allocation groups
    # can be very slow.
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -139,13 +139,20 @@ should_autosave_rootfs() {
...
-    # Semi-arbitrarily chosen: this is roughly ~64G currently (based on initial
-    # ag sizing at build time) which seems like a good rootfs size at which to
+    # This is roughly ~700GiB currently (based on initial ag sizing at build time) which
+    # seems like a good rootfs size at which to
     # discriminate between "throwaway/short-lived systems" and "long-running
-    # workload systems". It's not like XFS performance is way worse at 128.
-    if [ "$agcount" -lt 128 ]; then
-        echo "Filesystem agcount is $agcount; skipping" >&2
+    # workload systems".  Specifically for e.g. OpenShift, it's below the default node
+    # sizes.
```
dustymabe: I think this comment is wrong now. The `discriminate between "throwaway/short-lived systems" and "long-running workload systems"` comment really doesn't apply for a 700GiB threshold IMO.

Maybe we can just say that ~400 agcount (which is ~700GiB) seemed to be a good cutoff for not losing too much performance on grown filesystems and not having to require everyone to autosave-xfs on first boot.

dustymabe (commented):
Thanks for working on this. A few comments

dustymabe:
I added the whitespace fixes over in https://github.com/coreos/fedora-coreos-config/pull/2566 so you can ignore that CI failure here.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
cgwalters: Personally, I always use `journalctl -o with-unit` which clearly shows the origin of things; it isn't reliable or scalable to try to prefix all the output inside a unit. 

 But sure, done anyways.

cgwalters: > The other echo in this function prints to stderr:

I think these are normal conditions, so we should print to stdout.  Fixed the other cases.

dustymabe: Yes, these are "normal" conditions, but we have to consider the way the function is used here. stdout gets swallowed by the caller and then the string is used in conditional logic.

cgwalters: Oh wow you're right yes...writing correct shell script is just a mess.  Fixed

tests/kola/root-reprovision/autosave-xfs/test.sh
cgwalters: I don't quite see the value in having it be close; we don't care about the exact value so much as its rough range.

> Probably need to update the luks autosave-xfs test too.

Done.

dustymabe:
> I added the whitespace fixes over in #2566 so you can ignore that CI failure here.

I broke the whitespace fixes out into https://github.com/coreos/fedora-coreos-config/pull/2570

tests/kola/root-reprovision/autosave-xfs/test.sh
dustymabe: That's fair. We don't need to change it.

The only value I see in changing it is that the behavior is more well defined for us. i.e. if someone comes to us and they say they have a 800GiB disk (probably not common) then we can know if it should or should not have had autosave-xfs run.

tests/kola/root-reprovision/autosave-xfs/test.sh
```diff
@@ -20,10 +21,13 @@ set -xeuo pipefail
 if [ ! -f /run/ignition-ostree-autosaved-xfs.stamp ]; then
     fatal "expected autosaved XFS"
 fi
+# Verify we printed something about the agcount
+journalctl -u ignition-ostree-transposefs-autosave-xfs.service --grep=agcount
```
dustymabe, line 25: Do we also want this in the luks.autosave-xfs test?

dustymabe (dismissed):
LGTM

tests/kola/root-reprovision/autosave-xfs/test.sh
cgwalters: Would probably more make sense to refactor these to share a helper script, but that doesn't seem worth the churn yet.

travier (dismissed):

cgwalters:
Blah, missed a case of output redirection.  Fixed now and cleaned up.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -139,17 +139,24 @@ should_autosave_rootfs() {
...
+    # This is roughly ~700GiB currently (based on initial ag sizing at build time)
+    # which ensures we grow only on "large" root filesystems.
+    # Specifically for e.g. OpenShift, this ensures we don't reprovision on default
+    # worker node root filesystems.
+    local threshold
+    threshold=400
+    if [ "$agcount" -lt "${threshold}" ]; then
+        echo "autosave-xfs: ${root_part} agcount=$agcount is lower than threshold=${threshold}" >&2
         echo 0
         return
```
dustymabe, line 155: totally optional: this `return` can be dropped now.

dustymabe (approved):
LGTM

travier (approved):

PR #2566, opened 2023-08-25
overlay.d: add 30lvmdevices overlay

dustymabe:
Populate an lvmdevices(8) file to limit LVM from autoactivating all
devices it sees in a system. By default systems will get a "blank"
configuration file with a comment in it explaining what it is used
for. There is also a one-time "populate" service that will run and
add any devices it sees into the devices file. This will serve to
import existing devices on upgrading systems or new systems with
pre-existing LVM devices attached. See the tracker issue [1] for more
information.

[1] coreos/fedora-coreos-tracker#1517

### DIFF
--- a/overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
+++ b/overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
@@ -1,0 +1,25 @@
+[Unit]
+Description=CoreOS Populate LVM Devices File
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1517
+# Only run this import once.
+ConditionPathExists=!/var/lib/coreos-populate-lvmdevices.stamp
+# Don't add default dependencies so we can run early enough to populate
+# the devices file before any LVM devices are used.
+DefaultDependencies=false
+# Since our ConditionPathExists file lives under /var/lib/ let's make
+# sure any filesystems/mounts that exist and are needed to access that
+# directory are set up.
+RequiresMountsFor=/var/lib
+# On OpenShift/Kubernetes we want to ensure we run before kubernetes
+# comes up where storage drivers may be initiated and present more LVM
+# devices to the system that we don't want to be considered.
+Before=kubelet.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/coreos-populate-lvmdevices
+ExecStartPost=touch /var/lib/coreos-populate-lvmdevices.stamp
+
+[Install]
+WantedBy=default.target
--- a/tests/kola/disks/lvmdevices
+++ b/tests/kola/disks/lvmdevices
@@ -1,0 +1,65 @@
+#!/bin/bash
+## kola:
+##   description: Verify LVM devices file handling works as expected.
+##                This test is meant to cover coreos-populate-lvmdevices.service.
+##   # additionalDisks is only supported on qemu.
+##   platforms: qemu
+##   # A few extra disks to set up LVM on.
+##   additionalDisks: ["1G", "2G"]
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+LVMDEVICESFILENAME="system.devices"
+LVMDEVICESFILE="/etc/lvm/devices/${LVMDEVICESFILENAME}"
+
+# Check that coreos-populate-lvmdevices did run
+if [ ! -e /var/lib/coreos-populate-lvmdevices.stamp ]; then
+  fatal "coreos-populate-lvmdevices didn't run"
+fi
+
+case "${AUTOPKGTEST_REBOOT_MARK:-}" in
+  "")
+      # Verify the lvmdevices file by default matches what was shipped.
+      if ! diff -u "/usr/${LVMDEVICESFILE}" "${LVMDEVICESFILE}"; then
+          fatal "Detected modified $LVMDEVICESFILE file."
+      fi
+
+      # Set up LVM on the two disks and set up a vg/lv/fs/mount on one
+      # of them. Specify a blank devicesfile so the *create commands
+      # won't touch our devices file.
+      pvcreate --devicesfile="" /dev/vda /dev/vdb
+      vgcreate --devicesfile="" vg1 /dev/vda
+      lvcreate --devicesfile="" vg1 --name=lv1 -l 100%FREE
+      mkfs.ext4 /dev/vg1/lv1
+      echo "/dev/vg1/lv1 /srv/ ext4 defaults 0 2" >> /etc/fstab
+
+      # Remove the stamp file to force the "migration" to happen on
+      # next boot.
+      rm -f /var/lib/coreos-populate-lvmdevices.stamp
+
+      # reboot to simulate running migration for first time on a
+      # system with pre-existing LVM devices
+      /tmp/autopkgtest-reboot rebooted
+      ;;
+
+  rebooted)
+      # Check that the devices are in the devices file.
+      grep -q 'DEVNAME=/dev/vda' "$LVMDEVICESFILE" || fatal "Missing vda in devices file"
+      grep -q 'DEVNAME=/dev/vdb' "$LVMDEVICESFILE" || fatal "Missing vdb in devices file"
+
+      # Check that we can see the PVs
+      if [ "$(pvs --noheadings | wc -l)" != '2' ]; then
+        fatal "Incorrect number of LVM PVs detected"
+      fi
+
+      # Check that /srv/ is a mountpoint
+      if ! mountpoint /srv; then
+        fatal "/srv/ is not mounted, but it should be"
+      fi
+
+      ok "LVM Devices file populated correctly"
+      ;;
+  *) fatal "unexpected mark: ${AUTOPKGTEST_REBOOT_MARK}";;
+esac
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -16,6 +16,7 @@ ostree-layers:
   - overlay/09misc
   - overlay/20platform-chrony
   - overlay/25azure-udev-rules
+  - overlay/30lvmdevices
 
 # Be minimal
 recommends: false
--- a/overlay.d/30lvmdevices/etc/lvm/devices/system.devices
+++ b/overlay.d/30lvmdevices/etc/lvm/devices/system.devices
@@ -1,0 +1,8 @@
+# LVM uses devices listed in this file.
+#
+# This is an empty lvmdevices(8) file placed here by the CoreOS overlays.
+# The existence of the file prevents LVM from auto-activating any LVM devices
+# that aren't explicitly allowlisted by being added to this file. Any newly
+# added PV/VG devices that get created via pvcreate/vgreate will have an entry
+# added here automatically by the tools. For more information on this see
+# https://github.com/coreos/fedora-coreos-tracker/issues/1517
--- a/overlay.d/30lvmdevices/statoverride
+++ b/overlay.d/30lvmdevices/statoverride
@@ -1,0 +1,2 @@
+# Config file for overriding permission bits on overlay files/dirs
+# Format: =<file mode in decimal> <absolute path to a file or directory>
--- a/overlay.d/30lvmdevices/usr/lib/systemd/system-preset/45-coreos-populate-lvmdevices.preset
+++ b/overlay.d/30lvmdevices/usr/lib/systemd/system-preset/45-coreos-populate-lvmdevices.preset
@@ -1,0 +1,1 @@
+enable coreos-populate-lvmdevices.service
--- a/overlay.d/30lvmdevices/usr/libexec/coreos-populate-lvmdevices
+++ b/overlay.d/30lvmdevices/usr/libexec/coreos-populate-lvmdevices
@@ -1,0 +1,46 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script will detect any LVM devices and add them to the lvmdevices
+# file, which will force this host to only consider those devices in
+# the future. This script should be run once to do the population and
+# should not need to be run again. See
+# https://github.com/coreos/fedora-coreos-tracker/issues/1517
+
+LVMDEVICESFILENAME="system.devices"
+LVMDEVICESFILE="/etc/lvm/devices/${LVMDEVICESFILENAME}"
+
+# If the devices file doesn't exist that is a bit odd because we
+# shipped it in the same update this migration script runs but let's
+# just bail out. Someone could have deleted the lvmdevices file and
+# then later accidentally run the migration script again.
+if [ ! -f $LVMDEVICESFILE ]; then
+    echo "$LVMDEVICESFILE does not exist. Exiting."
+    exit 0
+fi
+
+# If the file exists and the file is different than what was shipped
+# then we exit early. In this case the system likely already had an
+# lvmdevices file defined already.
+if ! diff -u "/usr/${LVMDEVICESFILE}" "${LVMDEVICESFILE}"; then
+    echo "Detected modified $LVMDEVICESFILE file. Exiting."
+    exit 0
+fi
+
+# Detect all existing PVs using `pvs` with a blank devicesfile
+# setting, which will un-limit the search.
+PVS=$(pvs --devicesfile="" --noheadings -o pv_name)
+
+if [ -z "$PVS" ]; then
+    echo "No LVM devices detected. Exiting."
+    exit 0
+fi
+
+echo "Populating lvmdevices file with detected devices."
+for pv in $(pvs --devicesfile="" --noheadings -o pv_name); do
+    echo "Adding ${pv} to lvmdevices file $LVMDEVICESFILE"
+    lvmdevices --journal output --adddev "$pv" --devicesfile "$LVMDEVICESFILENAME"
+done
+
+echo "Activating lvmdevices after devices file population"
+pvscan --cache --activate ay
--- a/overlay.d/README.md
+++ b/overlay.d/README.md
@@ -67,7 +67,7 @@ such as `azure`, `aws`, `gcp`. The chrony config for these NTP servers
 should override other chrony configuration (e.g. DHCP-provided)
 configuration.
 
-25-azure-udev-rules
+25azure-udev-rules
 -------------------
 
 Add udev rules for SRIOV networking on Azure. The udev rules are also
--- a/overlay.d/README.md
+++ b/overlay.d/README.md
@@ -79,3 +79,17 @@ bits to include the rules in the initramfs too.
 [1] https://github.com/coreos/fedora-coreos-tracker/issues/1383
 [2] https://github.com/Azure/WALinuxAgent/pull/1622
 [3] https://src.fedoraproject.org/rpms/WALinuxAgent/pull-request/4
+
+30lvmdevices
+-------------------
+
+Populate an lvmdevices(8) file to limit LVM from autoactivating all
+devices it sees in a system. By default systems will get a "blank"
+configuration file with a comment in it explaining what it is used
+for. There is also a one-time "populate" service that will run and
+add any devices it sees into the devices file. This will serve to
+import existing devices on upgrading systems or new systems with
+pre-existing LVM devices attached. See the tracker issue [1] for more
+information.
+
+[1] https://github.com/coreos/fedora-coreos-tracker/issues/1517


tests/kola/disks/lvmdevices
```diff
@@ -0,0 +1,64 @@
...
+  "")
+      # Verify the lvmdevices file by default matches what was shipped.
+      if ! diff -u "/usr/${LVMDEVICESFILE}" "${LVMDEVICESFILE}"; then
+          fatal "Detected modified $LVMDEVICESFILE file."
+      fi
+
+      # Set up LVM on the two disks and set up a vg/lv/fs/mount on one
+      # of them. Specify a blank devicesfile so the *create commands
+      # won't touch our devices file.
+      pvcreate --devicesfile="" /dev/vda /dev/vdb
```
travier, line 32: If I run:
```
cosa run --qemu-image fedora-coreos-38.20230806.1.0-qemu.x86_64.qcow2 --add-disk 1G --add-disk 2G
```
I get:
```
[core@cosa-devsh ~]$ lsblk 
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
vda    252:0    0    1G  0 disk 
vdb    252:16   0    2G  0 disk 
vdc    252:32   0   10G  0 disk 
├─vdc1 252:33   0    1M  0 part 
├─vdc2 252:34   0  127M  0 part 
├─vdc3 252:35   0  384M  0 part /boot
└─vdc4 252:36   0  9.5G  0 part /var
                                /sysroot/ostree/deploy/fedora-coreos/var
                                /usr
                                /etc
                                /
                                /sysroot
```
so shouldn't this be vdb & vdc ?

tests/kola/disks/lvmdevices
```diff
@@ -0,0 +1,64 @@
...
+      vgcreate --devicesfile="" vg1 /dev/vda
```
travier, line 33: vdb?

tests/kola/disks/lvmdevices
```diff
@@ -0,0 +1,64 @@
...
+      rm -f /var/lib/coreos-populate-lvmdevices.stamp
+
+      # reboot to simulate running migration for first time on a
+      # system with pre-existing LVM devices
+      /tmp/autopkgtest-reboot rebooted
+      ;;
+
+  rebooted)
+      # Check that the devices are in the devices file.
+      grep -q 'DEVNAME=/dev/vda' "$LVMDEVICESFILE" || fatal "Missing vda in devices file"
```
travier, line 49: vdb vdc?

travier (commented):
Some comments but looks great otherwise

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
```diff
@@ -0,0 +1,26 @@
...
+After=var.mount var-lib.mount
+# On OpenShift/Kubernetes we want to ensure we run before kubernetes
+# comes up where storage drivers may be initiated and present more LVM
+# devices to the system that we don't want to be considered.
+Before=kubelet.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+StandardOutput=journal+console
```
cgwalters: Why output to the console?

tests/kola/disks/lvmdevices
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+## kola:
+##   description: Verify LVM devices file handling works as expected
```
cgwalters: In another PR I explicitly included the name of the systemd unit being tested which is *really* useful for grepping.

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
```diff
@@ -0,0 +1,26 @@
...
+Description=CoreOS Populate LVM Devices File
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1517
+# Only run this import once.
+ConditionPathExists=!/var/lib/coreos-populate-lvmdevices.stamp
+# Don't add default dependencies so we can run early enough to populate
+# the devices file before any LVM devices are used.
+DefaultDependencies=false
+# Since our ConditionPathExists file lives under /var/lib/ let's make
+# sure they are set up if those filesystems/mountpoints exist separately.
+After=var.mount var-lib.mount
```
cgwalters: I think this is just `RequiresMountsFor=/var/lib`

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
```diff
@@ -0,0 +1,26 @@
...
+# On OpenShift/Kubernetes we want to ensure we run before kubernetes
+# comes up where storage drivers may be initiated and present more LVM
+# devices to the system that we don't want to be considered.
+Before=kubelet.service
```
cgwalters, line 16: A bit of a layering violation here, which ultimately at least for OCP we'll fix with something in the shape of https://github.com/openshift/machine-config-operator/pull/3865

But obviously a good idea to have here now.

tests/kola/disks/lvmdevices
dustymabe: `vdc` is our FCOS disk. `vda` and `vdb` are the additional disks that get added that I use for LVM. I'm not 100% sure why the "additional disks" show up before the boot disk but that was the behavior so I just took it as was.

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
dustymabe: Yeah. It's probably not needed since we have DefaultDependencies=false, but I was trying to make sure we had the case covered. I'm happy to remove it if you think it's safe to do so.

tests/kola/disks/lvmdevices
dustymabe: The comment above covers it:

```

      # Set up LVM on the two disks and set up a vg/lv/fs/mount on one
      # of them.
```

It's just an arbitrary test, but I am intending to only set up a mount on just one of the LVM devices.

tests/kola/disks/lvmdevices
dustymabe: see https://github.com/coreos/fedora-coreos-config/pull/2566/files#r1307779743

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
dustymabe: In a lot of systems there may be no extra filesystem for `/var` or `/var/lib`. `RequiresMountsFor=` will add a `Requires=` too, which we don't want IIUC.

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
dustymabe: When I was developing the corresponding test added in this PR I had things misconfigured and the system didn't come up fully. Having this extra output on the console was useful.

Since this unit only runs once and there is opportunity for peoples systems to get stuck during boot (i.e. if they are mounting a FS on top of LVM and the populate doesn't happen properly) I think having the output go to the console will be more useful than not.

tests/kola/disks/lvmdevices
dustymabe: added!

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
cgwalters: This operates correctly if there isn't a distinct filesystem, see some of the examples in

```
$ grep -r RequiresMounts /usr/lib/systemd/system
/usr/lib/systemd/system/rpmdb-migrate.service:RequiresMountsFor=/var
/usr/lib/systemd/system/rpmdb-rebuild.service:RequiresMountsFor=/usr
/usr/lib/systemd/system/rpm-ostreed.service:RequiresMountsFor=/boot
/usr/lib/systemd/system/ostree-boot-complete.service:RequiresMountsFor=/boot
/usr/lib/systemd/system/ostree-finalize-staged-hold.service:RequiresMountsFor=/sysroot /boot
/usr/lib/systemd/system/ostree-finalize-staged.service:RequiresMountsFor=/sysroot /boot
/usr/lib/systemd/system/basic.target:RequiresMountsFor=/var /var/tmp
/usr/lib/systemd/system/podman-clean-transient.service:RequiresMountsFor=%t/containers
/usr/lib/systemd/system/podman-kube@.service:RequiresMountsFor=%t/containers
/usr/lib/systemd/system/rpcbind.service:RequiresMountsFor=/run/rpcbind
/usr/lib/systemd/system/systemd-journal-flush.service:RequiresMountsFor=/var/log/journal
/usr/lib/systemd/system/systemd-machined.service:RequiresMountsFor=/var/lib/machines
/usr/lib/systemd/system/systemd-nspawn@.service:RequiresMountsFor=/var/lib/machines/%i
/usr/lib/systemd/system/systemd-portabled.service:RequiresMountsFor=/var/lib/portables
/usr/lib/systemd/system/systemd-random-seed.service:RequiresMountsFor=/var/lib/systemd/random-seed
/usr/lib/systemd/system/systemd-rfkill.socket:RequiresMountsFor=/var/lib/systemd/rfkill
/usr/lib/systemd/system/systemd-update-utmp-runlevel.service:RequiresMountsFor=/var/log/wtmp
/usr/lib/systemd/system/systemd-update-utmp.service:RequiresMountsFor=/var/log/wtmp
```

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
cgwalters: I guess there's a sort of ultimate question here for whether the default systemd unit status output is what we want.

But individual units overriding to write to the console in an ad-hoc fashion seems suboptimal.

I mean, if the system didn't come up fully, we already have documented how to enable `systemd.journald.forward_to_console=1` etc right?

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
dustymabe: I was going mostly off of the description in the man page which says:

```
RequiresMountsFor=
    Takes a space-separated list of absolute paths. Automatically adds dependencies
    of type Requires= and After= for all mount units required to access the
    specified path.

```

From that description `RequiresMountsFor=/var/lib` would translate directly into `Requires=var-lib.mount` and `After=var-lib.mount` which would force `coreos-populate-lvmdevices.service` to get [dropped from the transaction in systemd 254+](https://github.com/coreos/fedora-coreos-tracker/issues/1527#issuecomment-1656786452). 

And indeed `Requires=var-lib.mount` and `After=var-lib.mount` yields me with:

```
[core@qemu0 ~]$ systemctl status coreos-populate-lvmdevices.service 
○ coreos-populate-lvmdevices.service - CoreOS Populate LVM Devices File
     Loaded: loaded (/usr/lib/systemd/system/coreos-populate-lvmdevices.service; enabled; preset: enabled)
             └─10-timeout-abort.conf
     Active: inactive (dead)
       Docs: https://github.com/coreos/fedora-coreos-tracker/issues/1517

Aug 28 20:18:29 localhost systemd[1]: coreos-populate-lvmdevices.service: Cannot add dependency job to transaction, deleting job coreos-populate-lvmdevices.service/start again: Unit var-lib.mount not found.
Aug 28 20:18:29 localhost systemd[1]: coreos-populate-lvmdevices.service: Cannot add dependency job, ignoring: Unit var-lib.mount not found.
Aug 28 20:19:18 qemu0 systemd[1]: coreos-populate-lvmdevices.service: Next restart interval calculated as: 100ms
```


but `RequiresMountsFor=/var/lib` apparently has different behavior because it works fine (passes the test).

I guess maybe the difference is `RequiresMountsFor=` will just look at what mounts exist and only mark them required if they exist and are required to access that path?

dustymabe: Another question is will `RequiresMountsFor=/var/lib` also add a requirement for `var.mount`? I would assume it would have done both `var.mount` and `var-lib.mount` but then why does systemd have `RequiresMountsFor=/var /var/tmp` separate in https://github.com/systemd/systemd/blob/337e8504f71074686a00dbec4ed9542a0a0c2224/units/basic.target#L21 ?

cgwalters: > I guess maybe the difference is RequiresMountsFor= will just look at what mounts exist and only mark them required if they exist and are required to access that path?

From a quick glance at the source code (see `unit_add_mount_dependencies`) it looks like that.

dustymabe: ok added `RequiresMountsFor=/var /var/lib` in the latest upload.

dustymabe: > > I guess maybe the difference is RequiresMountsFor= will just look at what mounts exist and only mark them required if they exist and are required to access that path?
> 
> From a quick glance at the source code (see `unit_add_mount_dependencies`) it looks like that.

Yeah. I was looking at that function but it wasn't 100% clear to me. I'm guessing `manager_load_unit_prepare` will load a unit for `var-lib.mount` even if no configuration exists for one and it'll get deleted from the transaction later. At least that's what the debug level systemd logging looked like was happening to me when I tested it.

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
cgwalters: This one is quite small, it's not a super big deal.  In OpenShift though [someone did this for ovs-configuration.service](https://github.com/openshift/machine-config-operator/blob/2ea9a64d51497060e4bef9e87fc9a55baf85b1f0/templates/common/_base/units/ovs-configuration.service.yaml#L18) which completely takes over the console output.

dustymabe: Just getting back to this.. I'm not strongly tied to this configuration. i.e. if you want I'll remove the `StandardOutput=` and `StandardError=` lines. I just found them useful during my debugging because I ended up with a hung system and for some reason the journal from the kola test didn't get captured. Only the console.

dustymabe:
Once we get the code review close I'll spin another RHCOS/OpenShift test for this. Let me know.

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
cgwalters: There's *almost* nothing in the set today writing stdout to the console, and not much to stderr:

```
$ grep -r 'Standard.*=.*console' /usr/lib/systemd/system
/usr/lib/systemd/system/coreos-installer-reboot.service:StandardOutput=kmsg+console
/usr/lib/systemd/system/coreos-installer-reboot.service:StandardError=kmsg+console
/usr/lib/systemd/system/coreos-installer.service:StandardOutput=kmsg+console
/usr/lib/systemd/system/coreos-installer.service:StandardError=kmsg+console
/usr/lib/systemd/system/ostree-prepare-root.service:StandardError=journal+console
/usr/lib/systemd/system/ostree-remount.service:StandardError=journal+console
/usr/lib/systemd/system/selinux-autorelabel.service:StandardOutput=journal+console
/usr/lib/systemd/system/selinux-check-proper-disable.service:StandardOutput=journal+console
$
```

I think there should be a pretty high bar in writing non-errors to the console by default.
(Even `StandardError` is sparse too, but that's a distinct thing)

dustymabe: ok I dropped the `journal+console` configuration from this PR

teigland:
Hi, I happened to pull up this issue again today while researching the design of the lvm feature (bug 2224641) that I hope would handle this in the future (for this and another image-based installations). 

The basic idea is that on first boot, lvm would create a new system.devices file for the root VG if no system.devices exists.  The method I see above is including all visible PVs.  Is it important for you to include all PVs vs just devices for the root VG?

Also, I'm thinking about whether the automatic creation of system.devices can happen by default, or whether it needs to be enabled explicitly.   If it needs to be enabled, then you'd need to do something like edit a file in the image to turn it on before booting the image for the first time.  How difficult would that editing be?

Looking at the script for populating system.devices, it looks like a fine approach, except it's unfortunate that it has to do activation, which  is a fairly significant divergence from normal system behavior (although it makes sense that first boot has some latitude for doing things differently.)  I can see why an activation command would be needed in the script, and don't have any good suggestion.  I'd note that 'vgchange -aay' is the standard/preferred command for autoactivation, as opposed to 'pvscan --cache -aay' which is no longer used in RHEL9 (although it should still work.)

dustymabe:
> Hi, I happened to pull up this issue again today while researching the design of the lvm feature (bug 2224641) that I hope would handle this in the future (for this and another image-based installations).
> 
> The basic idea is that on first boot, lvm would create a new system.devices file for the root VG if no system.devices exists. The method I see above is including all visible PVs. Is it important for you to include all PVs vs just devices for the root VG?

With CoreOS systems it's almost impossible for the root filesystem to be on a VG. Ignition doesn't support LVM and we don't offer any way to provision a system to use LVM for the root filesystem so that's not a use case I'm considering here.

The goal of this work is to identify and add all existing LVM devices on the system to a devices file. Think if someone had added a device to a VG called vgdata and then an LV on top and ran mkfs and mounted that under `/var/data/`. That's the use case I'm trying to capture here. i.e. on upgrade their LV continues to work even though we are now delivering a devices file by default.

The service runs before kubelet is started so hopefully any Block storage delivered by storage drivers that may have LVM on subordinate devices for guests wouldn't be up yet. 

> 
> Also, I'm thinking about whether the automatic creation of system.devices can happen by default, or whether it needs to be enabled explicitly. If it needs to be enabled, then you'd need to do something like edit a file in the image to turn it on before booting the image for the first time. How difficult would that editing be?

Right now the way this PR works is that there is an empty LVM devices file delivered with the system by default.

> 
> Looking at the script for populating system.devices, it looks like a fine approach, except it's unfortunate that it has to do activation, which is a fairly significant divergence from normal system behavior (although it makes sense that first boot has some latitude for doing things differently.) I can see why an activation command would be needed in the script, and don't have any good suggestion. I'd note that 'vgchange -aay' is the standard/preferred command for autoactivation, as opposed to 'pvscan --cache -aay' which is no longer used in RHEL9 (although it should still work.)

For `vgchange` versus `pvscan`, don't I need some sort of "scan" to happen to search for devices that had previously been ignored? Would `vgchange` do that?

I wish I didn't need any scan at all. When I ran the test (see the test added in this PR) on Fedora I didn't need the `pvcscan`, the sytem just picked up the change to the devices file and it worked without the scan. On RHEL I needed the scan. Maybe it's a race condition of sorts. Either way the `pvscan` was the big hammer approach.

dustymabe:
@teigland - I'd be happy to get together to discuss this further! There's a lot of nuance here and boucing ideas off each other would be quite useful.

tests/kola/disks/lvmdevices
travier: Oh yeah, my bad, I did not even read my output correctly here.

teigland:
> With CoreOS systems it's almost impossible for the root filesystem to be on a VG. Ignition doesn't support LVM and we don't offer any way to provision a system to use LVM for the root filesystem so that's not a use case I'm considering here.
> 
> The goal of this work is to identify and add all existing LVM devices on the system to a devices file. Think if someone had added a device to a VG called vgdata and then an LV on top and ran mkfs and mounted that under `/var/data/`. That's the use case I'm trying to capture here. i.e. on upgrade their LV continues to work even though we are now delivering a devices file by default.

Interesting, it's quite a different problem than I've been working on solving.  The problem here sounds like an OS upgrade, and I'd been focusing on fresh image-based installations that use lvm for the system root.  Upgrades are obviously complicated by existing system customizations -- I've never seriously considered trying to enable the devices file in a RHEL8 to 9 upgrade.   8 to 9 upgrades is the main reason that the lvmdevices feature is disabled when no system.devices exists.

cgwalters:
> With CoreOS systems it's almost impossible for the root filesystem to be on a VG. Ignition doesn't support LVM and we don't offer any way to provision a system to use LVM for the root filesystem so that's not a use case I'm considering here.

(Note that `bootc install-to-filesystem` though is explicitly intended to be used for things like installing on top of LVM, and I definitely plan to continue to "productize" bootc in the CoreOS context, so Ignition-LVM support is not necessary to use CoreOS with LVM in the future)

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
```diff
@@ -0,0 +1,25 @@
...
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1517
+# Only run this import once.
+ConditionPathExists=!/var/lib/coreos-populate-lvmdevices.stamp
+# Don't add default dependencies so we can run early enough to populate
+# the devices file before any LVM devices are used.
+DefaultDependencies=false
+# Since our ConditionPathExists file lives under /var/lib/ let's make
+# sure any filesystems/mounts that exist and are needed to access that
+# directory are set up.
+RequiresMountsFor=/var /var/lib
```
cgwalters: nit: I think the separate `/var` is redundant and the systemd example doing this is also redundant

cgwalters (dismissed):

overlay.d/30lvmdevices/usr/lib/systemd/system/coreos-populate-lvmdevices.service
dustymabe: Yeah, I didn't know, but figured following the example in their own repo was safe.

dustymabe: Note I did start a [discussion upstream](https://github.com/systemd/systemd/issues/29003) about the behavior of `RequiresMountsFor`. Ultimately it will probably lead to some clarification in the docs.

dustymabe: I updated it to `RequiresMountsFor=/var/lib`.

dustymabe:
> Once we get the code review close I'll spin another RHCOS/OpenShift test for this. Let me know.

OK I'm going to give this another round of testing on RHCOS/OpenShift now and then merge if everything looks good.

PR #2568, opened 2023-08-28
buildroot: Add clippy

cgwalters:
For the same reason we have `rustfmt`; I plan to rework the ostree-rs-ext build flow to use this there.

### DIFF
--- a/ci/buildroot/buildroot-reqs.txt
+++ b/ci/buildroot/buildroot-reqs.txt
@@ -25,8 +25,9 @@ util-linux
 which
 xz
 
-# For rust projects like rpm-ostree
+# For rust projects
 rustfmt
+clippy
 
 # For unit tests at least.
 ostree


jmarrero:
`[+] Found files with whitespace at the end of line` :eyes:

aaradhak:
Looks like the whitespace failures are handled in this - https://github.com/coreos/fedora-coreos-config/pull/2566

cgwalters:
Two mistakes:

- Adding a new CI check without fixing the existing problems it finds (and also, trailing whitespace is *not* some kind of critical blocker)
- Adding unrelated fixes *only* into an existing PR instead of splitting them out into something that can be merged separately now

jmarrero (approved):

travier:
> * Adding a new CI check without fixing the existing problems it finds (and also, trailing whitespace is _not_ some kind of critical blocker)

I've added the whitespace check in https://github.com/coreos/fedora-coreos-config/pull/2563 and I fixed the whitespace errors that were in tree at the time. Maybe we merged something with whitespace errors right before we merge the new GitHub Action, which is exactly the point of the Action.

As I wrote in the commit:
> While usually harmless, this can create issue in some edge cases and
creates noise when looking at the code in editors that highlight
unneeded whitespace.

PR #2562, opened 2023-08-23
Enable fwupd refresh timer on Fedora 39

travier:
Enable fwupd refresh timer on Fedora 39

See: https://fedoraproject.org/wiki/Changes/EnableFwupdRefreshByDefault
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1512

---

overlay/README: Update for 17fedora-modularity overlay

PR #2569, opened 2023-08-28
manifest: Add fwupd for all arches

ravanelli:
- Add fwupd package for all architecture, eliminating the exclusion for 's390x';
 - Now that we enabled/start 'fwupd' by default in #2562. We should have this package to all arhces.

### DIFF
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -13,16 +13,14 @@ packages:
  - linux-firmware
  # rpm-ostree
  - rpm-ostree nss-altfiles
+ # firmware updates
+ - fwupd
 
 # bootloader
 packages-aarch64:
   - grub2-efi-aa64 efibootmgr shim
-  # firmware updates
-  - fwupd
 packages-ppc64le:
   - grub2 ostree-grub2
-  # firmware updates
-  - fwupd
 packages-s390x:
   # On Fedora, this is provided by s390utils-core. on RHEL, this is for now
   # provided by s390utils-base, but soon will be -core too.
@@ -32,8 +30,6 @@ packages-s390x:
 packages-x86_64:
   - grub2 grub2-efi-x64 efibootmgr shim
   - microcode_ctl
-  # firmware updates
-  - fwupd
 
 exclude-packages:
   # Exclude kernel-debug-core to make sure that it doesn't somehow get


ravanelli:
The `fwupd` package was removed before in https://github.com/coreos/fedora-coreos-config/commit/4cae1d0bb381927e776cc041d5c8cdffaa81d30c.

Checking it for RHOS 4.14. I can't see the same errors:

```
[root@cosa-devsh core]# journalctl -u systemd-modules-load.service
Aug 28 18:08:58 localhost systemd-modules-load[347]: Inserted module 'fuse'
Aug 28 18:08:58 localhost systemd-modules-load[347]: Inserted module 'pkey'
Aug 28 18:08:58 localhost systemd[1]: Finished Load Kernel Modules.
Aug 28 18:09:04 localhost systemd[1]: systemd-modules-load.service: Deactivated successfully.
Aug 28 18:09:04 localhost systemd[1]: Stopped Load Kernel Modules.
```

cc: @jschintag

manifests/bootable-rpm-ostree.yaml
```diff
@@ -13,16 +13,14 @@ packages:
  - linux-firmware
  # rpm-ostree
  - rpm-ostree nss-altfiles
-
+ - fwupd
 # bootloader
 packages-aarch64:
   - grub2-efi-aa64 efibootmgr shim
   # firmware updates
```
dustymabe:
```suggestion
```

manifests/bootable-rpm-ostree.yaml
```diff
@@ -13,16 +13,14 @@ packages:
  - linux-firmware
  # rpm-ostree
  - rpm-ostree nss-altfiles
-
+ - fwupd
```
dustymabe, line 17: Add the comment and also the newline back:

```suggestion
 # firmware updates
 - fwupd
 
```

manifests/bootable-rpm-ostree.yaml
```diff
@@ -13,16 +13,14 @@ packages:
...
 # bootloader
 packages-aarch64:
   - grub2-efi-aa64 efibootmgr shim
   # firmware updates
-  - fwupd
 packages-ppc64le:
   - grub2 ostree-grub2
   # firmware updates
```
dustymabe:
```suggestion
```

manifests/bootable-rpm-ostree.yaml
```diff
@@ -33,7 +31,6 @@ packages-x86_64:
   - grub2 grub2-efi-x64 efibootmgr shim
   - microcode_ctl
   # firmware updates
```
dustymabe:
```suggestion
```

dustymabe:
I broke the whitespace fixes out into https://github.com/coreos/fedora-coreos-config/pull/2570

cgwalters (commented):
Looks sane to me!

jschintag:
I think i remember that s390x was excluded because there is no fwupd package for it.
Is there now one?
I have to admit i am unsure about s390x firmware, but i do not think that is updateable through the OS, but rather only from the Hardware Management Console or maybe even with physical access to the machine...

ravanelli:
> I think i remember that s390x was excluded because there is no fwupd package for it. Is there now one? I have to admit i am unsure about s390x firmware, but i do not think that is updateable through the OS, but rather only from the Hardware Management Console or maybe even with physical access to the machine...

Yes, there is one now. Let's say how it goes now, and if we find some issue again we should open an issue against fwupd to get that address this time.

dustymabe (approved):
LGTM

travier:
Note that fwupd is not only used to update the system firmwares but also the firmwares of devices that are plugged into a system. Not sure how much use that would be on s390x however.

travier (approved):

PR #2571, opened 2023-08-30
tests/kola: don't run fwupd-refresh-timer test on !fcos

dustymabe:


travier (approved):
Thanks. I was careful with the manifest inclusion but of course forgot it for the test.

PR #2572, opened 2023-08-30
denylist: extend snooze for `ext.config.networking.*` tests

marmijo:
These tests are still failing. 
Extend the snooze while we wait for a fix on https://github.com/coreos/fedora-coreos-tracker/issues/1542.


dustymabe (approved):
LGTM

PR #2575, opened 2023-09-01
denylist: deny kdump.crash test on rawhide

aaradhak:
Adding `ext.config.kdump.crash` to the denylist as it is failing right now with selinux enabled. 
see https://github.com/coreos/fedora-coreos-tracker/issues/1560

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -89,3 +89,9 @@
   warn: true
   platforms:
     - azure
+- pattern: ext.config.kdump.crash
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1560
+  snooze: 2023-09-30
+  warn: true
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
There are two `ext.config.kdump.crash` entries now. I think that's OK but we'll find out.

dustymabe:
> There are two `ext.config.kdump.crash` entries now. I think that's OK but we'll find out.

yeah, it doesn't work :) - looks like only the first one is getting picked up.

dustymabe:
> yeah, it doesn't work :) - looks like only the first one is getting picked up.

actually no, it does work - but we needed to specify `stream` and not `platform` here. I added a fix to a second commit in https://github.com/coreos/fedora-coreos-config/pull/2576

PR #2576, opened 2023-09-02
denylist: extend and adjust kdump.crash aarch64 denial

dustymabe:
The new kexec-tools v2.0.27 is in rawhide now so the problem should be fixed there. Let's extend the snooze some time to see if the new version makes it into the other streams.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -15,10 +15,17 @@
     - ppc64le
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1430
-  snooze: 2023-08-31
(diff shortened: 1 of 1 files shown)


marmijo (approved):
LGTM

kola-denylist.yaml
```diff
@@ -15,10 +15,17 @@
     - ppc64le
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1430
-  snooze: 2023-08-31
+  snooze: 2023-09-13
   warn: true
   arches:
     - aarch64
+  platforms:
```
travier, line 22: Shouldn't that be streams as well here?

PR #2585, opened 2023-09-07
kola secex tests: Increase test timeout to 5min.

jschintag:
We are experiencing test timeouts in the RHCOS Pipeline. Local testing shows that 3min. seems to be around the time needed for the test. Thus increasing the timeout from 3min. to 5min. should prevent future trouble. The secex.ensure test does not timeout, but is on the border with 160s. Increase timeout for it as well to prevent potential problems later.


jlebon (approved):

marmijo (approved):
LGTM

PR #2584, opened 2023-09-07
[rhcos-4.14] kola secex tests: Increase test timeout to 5min.

jschintag:
We are experiencing test timeouts in the RHCOS Pipeline. Local testing shows that 3min. seems to be around the time needed for the test. Thus increasing the timeout from 3min. to 5min. should prevent future trouble. The secex.ensure test does not timeout, but is on the border with 160s. Increase timeout for it as well to prevent potential problems later.

Signed-off-by: Jan Schintag <jan.schintag@de.ibm.com>
(cherry picked from commit 81d97de70f3cdb3f9230c816ff7d33b3841dd964)

PR #2586, opened 2023-09-07
[rhcos-4.13] kola secex tests: Increase test timeout to 5min.

jschintag:
We are experiencing test timeouts in the RHCOS Pipeline. Local testing shows that 3min. seems to be around the time needed for the test. Thus increasing the timeout from 3min. to 5min. should prevent future trouble. The secex.ensure test does not timeout, but is on the border with 160s. Increase timeout for it as well to prevent potential problems later.

(cherry picked from commit 81d97de70f3cdb3f9230c816ff7d33b3841dd964)

### DIFF
--- a/tests/kola/secex/ensure/test.sh
+++ b/tests/kola/secex/ensure/test.sh
@@ -6,7 +6,11 @@
 ##   architectures: s390x
 ##   platforms: qemu
 ##   requiredTag: secex
-##   timeoutMin: 3
+##   timeoutMin: 5
+##   description: Verify the s390x Secure Execution QEMU image works. It also
+##     implicitly tests Ignition config decryption.
+
+# We don't run it by default because it requires running with `--qemu-secex`.
 
 set -xeuo pipefail
 
--- a/tests/kola/secex/reboot/test.sh
+++ b/tests/kola/secex/reboot/test.sh
@@ -6,7 +6,12 @@
 ##   architectures: s390x
 ##   platforms: qemu
 ##   requiredTag: secex
-##   timeoutMin: 3
+##   timeoutMin: 5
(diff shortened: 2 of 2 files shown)


cgwalters (approved):
Trivial LGTM.

Thoughts related to the topic: I think what would also help at some point is a pipeline-global timeout multiplier.  Also timeouts in general are a crude mechanism; we can have distinct timeouts for "system booting" (e.g. 10m) versus "test execution" etc.

marmijo (approved):
LGTM

jlebon:
This looks out of order. Let's get #2585 in first and then update the cherry-pick references here and in #2584 to that one. (Because we're not using merge commits in this repo, the hash isn't actually known until it gets in.)

jschintag:
> This looks out of order. Let's get #2585 in first and then update the cherry-pick references here and in #2584 to that one. (Because we're not using merge commits in this repo, the hash isn't actually known until it gets in.)

Sounds fine to me, and yes i did 4.14 first and then remembered i needed to do the other 2...

jlebon:
> Also timeouts in general are a crude mechanism; we can have distinct timeouts for "system booting" (e.g. 10m) versus "test execution" etc.

We have some of that in kola already (see e.g. https://github.com/coreos/coreos-assembler/pull/2568). There's more granularity possible of course, but it's diminishing returns.

jschintag:
When reading it, it says timeout starts when ignition starts, which might be part of the problem here.
Essentially on firstboot ignition is encrypting the whole disk, which simply takes time. Guess that is eating into the timeout here, since it is part of ignition provisioning the node.

jlebon (approved):

PR #2591, opened 2023-09-12
manifests/fedora-coreos: ban initscripts-service

jlebon:
We want to make sure we never unknowingly start shipping this package. Relevant with the recent addition of audit which used to pull it in:

https://github.com/coreos/fedora-coreos-tracker/issues/1362

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -186,6 +186,8 @@ exclude-packages:
   # Let's make sure initscripts doesn't get pulled back in
   # https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-611566254
   - initscripts
+  # nor /usr/sbin/service
+  - initscripts-service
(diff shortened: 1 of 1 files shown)


travier (approved):
Thanks!

dustymabe:
IIUC this will cause the build to fail if that package gets pulled in.

We might want to figure out why the package is getting pulled in before we merge this because currently in `rawhide`/`branched` it is.

dustymabe:
> IIUC this will cause the build to fail if that package gets pulled in.
> 
> We might want to figure out why the package is getting pulled in before we merge this because currently in `rawhide`/`branched` it is.

Ahh. I mispoke. It's only getting pulled in on `branched`, not `rawhide`. That's because the new package update for audit isn't in the stable repos for F39 yet. I'll open a PR to fast-track it.

dustymabe:
- https://github.com/coreos/fedora-coreos-config/pull/2594

dustymabe (approved):

PR #2597, opened 2023-09-13
denylist: drop coreos.boot-mirror tests

jlebon:
These should now be fixed by
https://github.com/coreos/coreos-assembler/pull/3611.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3611

dustymabe (approved):

dustymabe:
This can be merged now I think.

dustymabe:
Actually - we might need to revert https://github.com/coreos/coreos-assembler/pull/3611 - hold off.

dustymabe:
The fix is in now: https://github.com/coreos/coreos-assembler/pull/3620

PR #2602, opened 2023-09-14
denylist: extend snooze for `ext.config.networking.*` tests

marmijo:
Extend the snooze on these tests again while we wait for a fix for https://github.com/coreos/fedora-coreos-tracker/issues/1542


dustymabe (approved):
LGTM

PR #2603, opened 2023-09-15
[next-devel] overrides: fast-track audit-3.1.2-4.fc39

coreosbot-releng:
Requested by @travier via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -10,15 +10,15 @@
 
 packages:
   audit:
-    evr: 3.1.2-3.fc39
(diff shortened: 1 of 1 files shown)


travier:
For https://github.com/coreos/fedora-coreos-tracker/issues/1362. Should fix the initscripts-service dep.

dustymabe (approved):

PR #2609, opened 2023-09-15
manifests: Move kernel into separate manifest

cgwalters:
This is prep for rebasing on [sagano](https://gitlab.com/cgwalters-playground/sagano)
where I hit on the issue that to build images with `kernel-rt`,
we need to clearly separate the `kernel` package from userspace
stuff (`systemd`, `rpm-ostree`) etc.

Note that we'll need to make the same change in RHCOS - until
we rebase both on sagano.

---

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -3,6 +3,7 @@
 # core functionality.
 
 include:
+  - kernel.yaml
   - system-configuration.yaml
   - ignition-and-ostree.yaml
   - file-transfer.yaml
--- a/manifests/bootable-rpm-ostree.yaml
+++ b/manifests/bootable-rpm-ostree.yaml
@@ -1,12 +1,11 @@
-# This minimal base starts just from: kernel + systemd + rpm-ostree + bootloader.
+# This minimal base is the userspace: systemd + rpm-ostree + bootloader.
 # The intent of this is to inherit from this if you are doing something highly
 # custom that e.g. might not involve Ignition or podman, but you do want
 # rpm-ostree.
 # We expect most people though using coreos-assembler to inherit from
 # fedora-coreos-base.yaml.
 packages:
- # Kernel + systemd.
- - kernel systemd
+ - systemd
  # linux-firmware now a recommends so let's explicitly include it
  # https://gitlab.com/cki-project/kernel-ark/-/commit/32271d0cd9bd52d386eb35497c4876a8f041f70b
  # https://src.fedoraproject.org/rpms/kernel/c/f55c3e9ed8605ff28cb9a922efbab1055947e213?branch=rawhide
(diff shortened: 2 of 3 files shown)


jlebon:
Hmm, odd that this is required. The `kernel-rt-core` package should Provide `kernel` so the `kernel` requirement here should be satisfied.

What was the depsolve error that you got?

cgwalters:
The problem is easy to replicate with this patch to the sagano configs:

```
diff --git a/tier-0/kernel-rt.yaml b/tier-0/kernel-rt.yaml
index cdcff10..e9dad97 100644
--- a/tier-0/kernel-rt.yaml
+++ b/tier-0/kernel-rt.yaml
@@ -4,7 +4,7 @@ repos:
 
 # Enable the "realtime" AKA soft-realtime AKA latency-optimized kernel.
 packages:
- - kernel-rt-core kernel-rt-modules kernel-rt-modules-extra kernel-rt-kvm
+ - kernel kernel-rt-core kernel-rt-modules kernel-rt-modules-extra kernel-rt-kvm
 
 exclude-packages:
   - kernel-rt-debug-core
```

```
$ rpm-ostree compose image --cachedir=/builds/compose.cache --initialize-mode if-not-exists --format=oci centos-tier-0-rt-stream9.yaml out.oci
...
error: Finalizing rootfs: During kernel processing: Multiple kernels (vmlinuz) found in: usr/lib/modules: 5.14.0-366.el9.x86_64+rt 5.14.0-366.el9.x86_64
```

(Clearly we should error out in the depsolve phase, not finalization here if multiple things provide kernels)

cgwalters:
There are of course multiple other ways I can think of to address this. 

The first would be something like
```
package_slots:
  kernel:
    - kernel
```

Under our inheritance rules this would allow something else to later do:

```
package_slots:
  kernel:
    - kernel-rt
```

and override instead of union.


Even then though, it seems like just a Good Idea to me to separate the userspace configs from the kernel as is being done here.

manifests/fedora-coreos-base.yaml
```diff
@@ -96,6 +96,8 @@ postprocess:
 # available in RHCOS or not desired in RHCOS). All other packages should go
 # into one of the sub-manifests listed at the top.
 packages:
+  # We use the default kernel package, but note c9s may differ
+  - kernel
```
jlebon: The kernel looks out of place in this package list. I think it'd be less odd and more structured (and even more of a Sagano prep) if we instead added the `tier-0/kernel.yaml` file into this repo for now and included it from this manifest.

.cci.jenkinsfile
```diff
@@ -4,6 +4,9 @@
 cosaPod(cpus: 4, memory: "9Gi") {
     checkoutToDir(scm, 'config')
 
+    // Until https://github.com/coreos/coreos-ci-lib/pull/153 lands
+    shwrap("(cd config && git submodule update --init --recursive)")
```
jlebon: Should be able to drop this now.

jlebon (dismissed):
Some comments but looks sane overall.

cgwalters:
OK, addressed comments and queued for merge.

jlebon (approved):
LGTM!

PR #2613, opened 2023-09-16
denylist: add next stream to denial for networking selinux issue

dustymabe:
Now that we have moved the next stream over to F39 we'll need these denials to apply there too.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -23,6 +23,7 @@
     - rawhide
     - branched
     - next-devel
+    - next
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Hmm, I think we talked about this in the past, but might be interesting to add releasever-based denials since most of the case, that's the real determinant when using `streams:`.

dustymabe:
> Hmm, I think we talked about this in the past, but might be interesting to add releasever-based denials since most of the case, that's the real determinant when using `streams:`.

100% - would love to have this feature implemented.

PR #2622, opened 2023-09-20
tests/kola/docker: Add a basic docker run test

travier:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1578

### DIFF
--- a/tests/kola/docker/basic
+++ b/tests/kola/docker/basic
@@ -1,0 +1,16 @@
+#!/bin/bash
+## kola:
+##   # Must not run this test with another podman test
+##   exclusive: true
+##   # This test pulls a container image from remote sources.
+##   tags: "platform-independent needs-internet"
+##   description: Verify that running a basic container with docker works.
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+docker run --rm registry.fedoraproject.org/fedora:38 true
+
+ok "basic docker run successfully"
--- a/tests/kola/docker/data/commonlib.sh
+++ b/tests/kola/docker/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh


travier:
This will fail on next.

tests/kola/docker/basic
```diff
@@ -0,0 +1,14 @@
...
+##   # This test pulls a container image from remote sources.
+##   tags: "platform-independent needs-internet"
+##   description: Verify that running a basic container with docker works.
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+docker run --rm -ti quay.io/fedora/fedora:latest true
```
dustymabe: all other uses of a fedora container image in this git repo uses the registry.fedoraproject.org/fedora location. I'm fine with migrating them all, but I think it would be good to be consistent.

dustymabe (dismissed):
One optional comment, otherwise LGTM.

Also, for some reason CI is failing even on testing-devel.

tests/kola/docker/basic
travier: We should indeed migrate them all. Will make a PR.

travier: https://discussion.fedoraproject.org/t/registry-fedoraproject-org-to-quay-io-migration/88231

tests/kola/docker/basic
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+## kola:
+##   # This test pulls a container image from remote sources.
+##   tags: "platform-independent needs-internet"
+##   description: Verify that running a basic container with docker works.
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
```
dustymabe, line 12: I think you need to add the symlink so this file can be found when the test is run.

tests/kola/docker/basic
travier: Moved back to registry.fedoraproject.org/fedora:38. I'll make another PR

travier:
Updated.

dustymabe (dismissed):
LGTM

travier:
Fix interactive/tty not working in a systemd unit

dustymabe (approved):
LGTM

dustymabe:
```
[2023-09-21T15:15:38.461Z] --- PASS: ext.config.docker.basic (40.21s)
```

Merging so we can get this in to next-devel and next sooner.

PR #2623, opened 2023-09-21
ci & tests: Switch to quay.io/fedora/fedora

travier:
Fedora is planning on deprecating the registry and moving to quay.io for
all containers.

Existing containers are mirrors to quay.io so we should be able to
switch now.

See: https://discussion.fedoraproject.org/t/registry-fedoraproject-org-to-quay-io-migration/88231
See: https://pagure.io/fedora-infrastructure/issue/10386

### DIFF
--- a/.github/workflows/add-override.yml
+++ b/.github/workflows/add-override.yml
@@ -37,7 +37,7 @@ jobs:
   add-override:
     name: Add package override
     runs-on: ubuntu-latest
-    container: registry.fedoraproject.org/fedora:latest
+    container: quay.io/fedora/fedora:latest
     steps:
       - name: Install dependencies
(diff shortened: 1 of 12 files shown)


travier:
Based on https://github.com/coreos/fedora-coreos-config/pull/2622 to simplify rebases. We don't have to do it now.

jlebon (dismissed):

jlebon:
There's also https://github.com/coreos/fedora-coreos-config/blob/745417d2a51ad93ceea02510550dd3336c3f525a/.github/workflows/add-override.yml#L40 we might as well change while we're here.

jlebon (dismissed):

dustymabe:
Let's rebase this and get it merged.

travier:
Rebased

dustymabe (approved):

PR #2628, opened 2023-09-22
upgrade/extended: Fix last_release_index extraction method

gursewak1997:
Initially, I was getting the last_release_index from the update json file itself under `key org.fedoraproject.coreos.releases.age_index` whereas we are using the actual index value of the versions as edge values in edges array

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -79,7 +79,7 @@ stream=$(get_booted_deployment_json | jq -r '.["base-commit-meta"]["fedora-coreo
 test -f /srv/updateinfo.json || \
     curl -L "https://updates.coreos.fedoraproject.org/v1/graph?basearch=$(arch)&stream=${stream}" > /srv/updateinfo.json
 last_release=$(jq -r .nodes[-1].version /srv/updateinfo.json)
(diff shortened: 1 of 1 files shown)


dustymabe:
Hmm. I would have thought `org.fedoraproject.coreos.releases.age_index` would have been the right thing to use.

dustymabe:
ok I think I see. with the x86_64 data returned by the server the `age_index` corresponds to the nodes[] index perfectly (because every release we've ever done has had an x86_64 release as part of it).

For example:

```
$ curl 'https://updates.coreos.fedoraproject.org/v1/graph?basearch=x86_64&stream=next' 2>/dev/null | gron | grep 'age_index' | grep '125'
json.nodes[125].metadata["org.fedoraproject.coreos.releases.age_index"] = "125";
```

But for the other architectures there wasn't an entry in the nodes graph for every release ever done of FCOS so the age_index is still 125 but the node index isn't:

```
$ curl 'https://updates.coreos.fedoraproject.org/v1/graph?basearch=aarch64&stream=next' 2>/dev/null | gron | grep 'age_index' | grep '125'
json.nodes[76].metadata["org.fedoraproject.coreos.releases.age_index"] = "125";
```

The edges are edges based on the node index.

dustymabe (approved):
LGTM

gursewak1997:
Yes, the reason the most recent `aarch64` upgrade builds failed but the `x86_64` passed.

PR #2612, opened 2023-09-16
[next] tree: promote changes from next-devel at ad8729d2564e9f1e18cd2da2b6f069fc65893f5f

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/762.

PR #2632, opened 2023-09-25
[next-devel] Revert "overrides: use grub2-2.06-98.fc39"

marmijo:
Now that there is a newer NVR version of grub2 (grub2-2.06-100.fc39) in the bodhi stable state than what is in Fedora 38 we can unpin the grub2 override that was necessary to prevent a downgrade in https://github.com/coreos/fedora-coreos-config/pull/2612#issuecomment-1722342103

This reverts commit 7ccc4f5c18804fa5e445f0f82a4758566a7deada.


dustymabe:
> Now that the `next` stream has had a build based on Fedora 39, we can remove this pin.

I'm fine with leaving the commit message as is but I think the "reason" is a bit different than this.

i.e. something like this may be more appropriate:

Now that there is a newer NVR version of grub2 (grub2-2.06-100.fc39) in the bodhi stable state than what is in Fedora 38 we can unpin the grub2 override that was necessary to prevent a downgrade in https://github.com/coreos/fedora-coreos-config/pull/2612#issuecomment-1722342103

dustymabe (approved):
LGTM

marmijo:
> i.e. something like this may be more appropriate:

Updated, along with the PR description. Thanks for also updating the title :smile:

dustymabe (approved):
LGTM

PR #2633, opened 2023-09-25
denylist: drop ext.config.kdump.crash test

aaradhak:
ext.config.kdump.crash test can be dropped from the denylist for as the new selinux-policy-38.28-1 pkg has landed in F39 and rawhide.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1560

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -21,14 +21,6 @@
   warn: true
   platforms:
     - azure
-- pattern: ext.config.kdump.crash
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1560
-  snooze: 2023-09-30
-  warn: true
-  streams:
-    - rawhide
-    - next-devel
-    - next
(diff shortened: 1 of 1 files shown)


kola-denylist.yaml
```diff
@@ -26,8 +26,6 @@
   snooze: 2023-09-30
   warn: true
   streams:
-    - rawhide
-    - next-devel
```
dustymabe, line 30: `next-devel` will get promoted to `next` so we should just be able to remove this entire entry.

dustymabe:
> selinux-policy-38.28-1.fc40 pkg has landed in rawhide.

we can just say:

> selinux-policy-38.28-1 pkg has landed in F39 and rawhide.

kola-denylist.yaml
aaradhak: I assume its good to remove the entire array now then.

dustymabe (approved):
LGTM

PR #2635, opened 2023-09-27
denylist: extend snooze for `coreos.ignition.ssh.key`

gursewak1997:
This test is still failing so let's bump the snooze while we work to find a fix for coreos/fedora-coreos-tracker#1553


marmijo (approved):
LGTM

PR #2638, opened 2023-09-28
denylist: bump snooze for `ext.config.kdump.crash` on aarch64

gursewak1997:
This test is still failing.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -7,7 +7,7 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1430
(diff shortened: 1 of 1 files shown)


kola-denylist.yaml
```diff
@@ -7,7 +7,7 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1430
-  snooze: 2023-09-28
+  snooze: 2023-10-13
```
dustymabe:
```suggestion
  snooze: 2023-10-24
```

Let's just move this until the F39 final target release date. Then all streams will have moved to F39 and this will no longer be an issue.

dustymabe (approved):

PR #2650, opened 2023-10-04
manifests: add crun-wasm.yaml

jlebon:
This will be sourced by `next-devel` and `next` as part of https://github.com/coreos/fedora-coreos-tracker/issues/1375.


dustymabe (approved):
LGTM

PR #2652, opened 2023-10-04
[testing-devel] overrides: fast-track glibc-2.37-10.fc38

marmijo:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1591


dustymabe (approved):
LGTM

PR #2653, opened 2023-10-04
[next-devel] overrides: fast-track glibc-2.38-6.fc39

marmijo:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1591


dustymabe (approved):
LGTM

PR #2658, opened 2023-10-05
Revert "coreos-teardown-initramfs: use NetworkManager to teardown networking"

jlebon:
This reverts commit b8607dabb5e1a5a0795fe11b5089889038c150f6.

We're seeing around 75% failure rate with this patch when provisioning RHCOS on AWS with m4 instance types. The error message is:

    Error: failed to set networking: Timeout was reached

I confirmed that D-Bus and NM are definitely still running when we issue the `nmcli` call. Still, I think there's a subtle race somewhere (likely combined with the specific network adapter of m4 instance types), that we need to plug. We need more visibility into why NM isn't responding. Anyway, for now let's revert this to not regress.

Fixes: https://issues.redhat.com/browse/OCPBUGS-17154
Probably fixes: https://issues.redhat.com/browse/OCPBUGS-19870

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.service
@@ -18,13 +18,13 @@ DefaultDependencies=false
 Before=ignition-mount.service
 Before=ignition-complete.target
 
-# Since we are tearing down networking using NetworkManager itself
-# we need to make sure NetworkManager is running. Similarly
-# dbus-broker since nmcli talks to the daemon via DBUS. Since the
-# ordering for ExecStop is the opposite of ExecStart we need to use
-# `After=` here.
-After=nm-initrd.service
-After=dbus-broker.service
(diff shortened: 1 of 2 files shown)


jlebon:
We could also make this conditional on RHCOS vs FCOS, but my hope here is that we can get the right fix in soon and just revert this revert.

dustymabe:
Are we actually seeing the failure for FCOS or just for RHCOS?

jlebon:
I haven't seen it fail on FCOS, but given its flakiness, I should try harder.

jlebon:
100% success rate for latest FCOS stable (n=10)

dustymabe:
seems to indicate to me a bug that's fixed somewhere in the software between RHEL and Fedora.

jlebon:
> seems to indicate to me a bug that's fixed somewhere in the software between RHEL and Fedora.

Agreed. I'm working on finding it. But in the meantime, this puts us in a not broken state so we can ship 4.14 without this caveat.

dustymabe (approved):
LGTM

jlebon:
FYI: opened https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/1400 for this

PR #2659, opened 2023-10-05
tests/kola/docker/basic: Only run on FCOS

travier:
Fixes: aa437320 tests/kola/docker: Add a basic docker run test


dustymabe (approved):
LGTM

PR #2662, opened 2023-10-09
[next] tree: promote changes from next-devel at c0a0942643a347627b65252e2ce188eb368cc422

coreosbot-releng:
Triggered by @c4rt0 in https://github.com/coreos/fedora-coreos-streams/issues/783.


c4rt0 (approved):
LGTM

PR #2664, opened 2023-10-09
[stable] update glibc for CVE-2023-4911

c4rt0:
This stable release is an out-of-band release for a specific reason (the glibc CVE).
We won't do a normal promotion, as we only want one package to change.
What we need to do is to revert the previous commit and fast-track the glibc update.

See: https://github.com/coreos/fedora-coreos-streams/issues/778


manifest-lock.overrides.yaml
```diff
@@ -9,6 +9,17 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+<<<<<<< HEAD
```
dustymabe:
```suggestion
```

manifest-lock.overrides.yaml
```diff
@@ -9,6 +9,17 @@
...
+  coreos-installer:
+    evr: 0.18.0-1.fc38
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-c34621867f
+      type: fast-track
+  coreos-installer-bootinfra:
+    evr: 0.18.0-1.fc38
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-c34621867f
+=======
```
dustymabe:
```suggestion
```

manifest-lock.overrides.yaml
```diff
@@ -32,9 +43,28 @@ packages:
     metadata:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-2b8c11ee75
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1591
+>>>>>>> e6e5fb76 (overrides: fast-track glibc-2.37-10.fc38)
```
dustymabe:
```suggestion
```

dustymabe (changes requested):

manifest-lock.overrides.yaml
c4rt0: > @dustymabe - all requested above changes are now implemented.

Ah, sorry - not yet.

dustymabe:
LGTM!

PR #2666, opened 2023-10-10
ci/buildroot: fixes to make `fcos-buildroot` image buildable on all arches

jlebon:
Part of adding multi-arch support to CoreOS CI.
See individual commits.


ci/buildroot/install-buildroot.sh
```diff
@@ -19,7 +19,14 @@ echo "${deps}" | xargs dnf -y install
...
 brs=$(grep -v '^#' "${dn}"/buildroot-buildreqs.txt)
-echo "${brs}" | xargs dnf -y builddep
+(cd "${tmpd}" && mkdir rpmbuild
+ echo "${brs}" | xargs dnf download --source
+ # rebuild the SRPM for this arch; see
+ # https://bugzilla.redhat.com/show_bug.cgi?id=1402784#c6
+ find . -name '*.src.rpm' -print0 | xargs -0n 1 rpmbuild -rs --nodeps \
+    -D "%_topdir $PWD/rpmbuild" -D "%_tmppath %{_topdir}/tmp"
+ dnf builddep -y rpmbuild/SRPMS/*.src.rpm)
+rm -rf "${tmpd:?}"/*
```
dustymabe, line 29: this shouldn't be needed because of line 7 (i.e. it will get cleaned up when this script exists anyway?

ci/buildroot/install-buildroot.sh
```diff
@@ -19,7 +19,14 @@ echo "${deps}" | xargs dnf -y install
...
 echo "Installing build dependencies of primary packages"
 brs=$(grep -v '^#' "${dn}"/buildroot-buildreqs.txt)
-echo "${brs}" | xargs dnf -y builddep
+(cd "${tmpd}" && mkdir rpmbuild
+ echo "${brs}" | xargs dnf download --source
+ # rebuild the SRPM for this arch; see
+ # https://bugzilla.redhat.com/show_bug.cgi?id=1402784#c6
+ find . -name '*.src.rpm' -print0 | xargs -0n 1 rpmbuild -rs --nodeps \
+    -D "%_topdir $PWD/rpmbuild" -D "%_tmppath %{_topdir}/tmp"
+ dnf builddep -y rpmbuild/SRPMS/*.src.rpm)
```
dustymabe, line 28: this is ugly, but I guess necessary. I guess the only way to fix this in the future would be to have source repos be arch specific too and build an SRPM for each arch?

dustymabe (approved):

ci/buildroot/install-buildroot.sh
jlebon: The directory gets reused below. It follows the same pattern of emptying out before the next step as those too.

dustymabe: ahh - ok

ci/buildroot/install-buildroot.sh
jlebon: Right, yeah. But I guess that comes with more overhead too so maybe it's not worth it. Another approach is to make `dnf builddep` smarter to handle some of this underneath (maybe not create a temporary SRPM, but unpacking the SRPM and pointing at the spec file seems more realistic).

PR #2667, opened 2023-10-10
upgrade/extended: Add rollout_wariness=0 to curl updates

gursewak1997:
Adding this option so we pick the changes to the edges to point to the latest release during the rollout window.


dustymabe (approved):
LGTM

PR #2668, opened 2023-10-12
[next-devel] overrides: fast-track curl-8.2.1-3.fc39

coreosbot-releng:
Requested by @travier via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


dustymabe (approved):
LGTM

PR #2670, opened 2023-10-12
tests/ignition: check file ownership for system users

HuijingHei:
Based on https://github.com/coreos/fedora-coreos-config/pull/774

### DIFF
--- a/tests/kola/ignition/sysusers/config.bu
+++ b/tests/kola/ignition/sysusers/config.bu
@@ -1,12 +1,11 @@
----
 variant: fcos
-version: 1.0.0
+version: 1.4.0
 storage:
   files:
-    - path: /etc/zincati/config.d/00-dummy-placeholder.toml
+    - path: /etc/dnsmasq/config.d/00-dummy-placeholder.toml
       mode: 0644
       user:
-        name: "zincati"
+        name: "dnsmasq"
       contents:
         inline: |
           # Dummy placeholder
--- a/tests/kola/ignition/sysusers/config.fcc
+++ b/tests/kola/ignition/sysusers/config.fcc
@@ -1,0 +1,12 @@
+---
+variant: fcos
+version: 1.0.0
+storage:
+  files:
+    - path: /etc/zincati/config.d/00-dummy-placeholder.toml
+      mode: 0644
+      user:
+        name: "zincati"
+      contents:
+        inline: |
+          # Dummy placeholder
--- a/tests/kola/ignition/sysusers/data/commonlib.sh
+++ b/tests/kola/ignition/sysusers/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../../data/commonlib.sh
--- a/tests/kola/ignition/sysusers/test.sh
+++ b/tests/kola/ignition/sysusers/test.sh
@@ -1,20 +1,17 @@
 #!/usr/bin/env bash
-set -euo pipefail
+## kola:
+##   platforms: qemu
+##   description: Verify file ownership can reference system users.
 
-ok() {
-    echo "ok" "$@"
-}
+set -xeuo pipefail
 
-fatal() {
-    echo "$@" >&2
-    exit 1
-}
+. "$KOLA_EXT_DATA/commonlib.sh"
 
-TARGET="/etc/zincati/config.d/00-dummy-placeholder.toml"
+TARGET="/etc/dnsmasq/config.d/00-dummy-placeholder.toml"
 OWNER=$(stat -c '%U' "${TARGET}")
 
 # make sure the placeholder file is owned by the proper system user.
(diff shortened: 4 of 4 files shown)


HuijingHei:
Check the test is passed without `ignition-ostree-sysusers.service` in https://github.com/coreos/fedora-coreos-config/pull/774, still not quite sure where the fix is.
```
$ kola run -E fedora-coreos-config --qemu-image rhcos-414.92.202310100209-0-qemu.x86_64.qcow2 ext.fedora-coreos-config.ignition.sysusers
=== RUN   ext.fedora-coreos-config.ignition.sysusers
--- PASS: ext.fedora-coreos-config.ignition.sysusers (41.09s)
PASS, output in _kola_temp/qemu-2023-10-12-1412-79

$ kola run -E fedora-coreos-config --qemu-image fedora-coreos-38.20231009.20.0-qemu.x86_64.qcow2 ext.fedora-coreos-config.ignition.sysusers
=== RUN   ext.fedora-coreos-config.ignition.sysusers
--- PASS: ext.fedora-coreos-config.ignition.sysusers (40.49s)
PASS, output in _kola_temp/qemu-2023-10-12-1420-138
```

jmarrero (dismissed):
lgtm

travier:
Some context for https://github.com/coreos/fedora-coreos-config/pull/774:

The use case where we need to run sysusers in the initramfs (https://github.com/coreos/fedora-coreos-config/pull/774) is when you want to declare a system user in a sysusers config file (written by Ignition to the real root) and also include a file in the Ignition config that is owned by that new system user. As the users will not be created in the real root when ignition-files runs, it would fail.

So we need something like https://github.com/coreos/fedora-coreos-config/pull/774 to pre-create the users so that Ignition can set the ownership of the files it writes to the real root.

---

This test looks fine to me but I'm not sure how much we already have testing that in kola.

HuijingHei:
> Some context for #774:
> 
> The use case where we need to run sysusers in the initramfs (#774) is when you want to declare a system user in a sysusers config file (written by Ignition to the real root) and also include a file in the Ignition config that is owned by that new system user. As the users will not be created in the real root when ignition-files runs, it would fail.

Thanks @travier for the info, I tried the following Ignition config, it works in `fedora-coreos-38.20231009.20.0`, is this what you meant? Can I understand the issue is fixed without #774 ?

```
$ cat config.bu 
variant: fcos
version: 1.5.0
passwd:
  users:
    - name: newuser
storage:
  files:
    - path: /etc/newuser/config.d/00-dummy-placeholder.toml
      mode: 0644
      user:
        name: "newuser"
      contents:
        inline: |
          # Dummy placeholder

[coreos-assembler]$ cosa run --qemu-image fedora-coreos-38.20231009.20.0-qemu.x86_64.qcow2 -m 4096 -B config.bu

[core@cosa-devsh ~]$ ls -l /etc/newuser/config.d/00-dummy-placeholder.toml
-rw-r--r--. 1 newuser root 20 Oct 13 13:28 /etc/newuser/config.d/00-dummy-placeholder.toml

```

HuijingHei:
Thanks @jmarrero @travier for the review, update the test script, might need re-review, thanks!

travier:
Here is an example config that should fail:

```
variant: fcos
version: 1.5.0
storage:
  files:
    - path: /etc/foo
      user:
        name: "foo"
      contents:
        inline: |
          # Dummy placeholder
    - path: /etc/sysusers.d/foo.conf
      contents:
        inline: |
          u foo - "Foo user" - -
```

See: https://www.freedesktop.org/software/systemd/man/sysusers.d.html

HuijingHei:
> Here is an example config that should fail:
> 
> ```
> variant: fcos
> version: 1.5.0
> storage:
>   files:
>     - path: /etc/foo
>       user:
>         name: "foo"
>       contents:
>         inline: |
>           # Dummy placeholder
>     - path: /etc/sysusers.d/foo.conf
>       contents:
>         inline: |
>           u foo - "Foo user" - -
> ```
> 
> See: https://www.freedesktop.org/software/systemd/man/sysusers.d.html


This should fail with `ignition-ostree-sysusers.service` refer to https://github.com/coreos/fedora-coreos-config/pull/774#issuecomment-745427947

jlebon:
So the issue here is that the `sysusers.d` dropin gets written by the files stage, so it's not yet present when the new `ignition-ostree-sysusers.service` unit in #2679 runs.

I commented in https://github.com/coreos/fedora-coreos-tracker/issues/155#issuecomment-1775749284 to try to keep design discussions there.

HuijingHei:
Thanks @jlebon @travier for the reply.

Does this script make sense?
Can I understand that we support that the entries in Ignition configuration can reference the existed system users (see [doc](https://docs.fedoraproject.org/en-US/fedora-coreos/managing-files/))?

PR #2679, opened 2023-10-19
core/dracut/ignition-ostree: add `ignition-ostree-sysusers.service`

HuijingHei:
Based on https://github.com/coreos/fedora-coreos-config/pull/774.

Copy text here:
This introduces a new `ignition-ostree-sysusers.service`, which takes
care of poulating users and groups on the target sysroot before the
`Ignition` files stage.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Needed to work around the initrd `rootfs` / filesystem not being a valid
+# mount to pivot out of. For reference, see:
+#  - https://github.com/torvalds/linux/blob/26bc672134241a080a83b2ab9aa8abede8d30e1c/fs/namespace.c#L3605
+#  - https://gist.github.com/jlebon/fb6e7c6dcc3ce17d3e2a86f5938ec033
+set -euo pipefail
+
+TMP_CHROOT_DIR=""
+
+main() {
+    setup_chroot_tmpdir
+    run_chrooted_bwrap "$@"
+}
+
+setup_chroot_tmpdir() {
+    TMP_CHROOT_DIR=$(mktemp --directory --tmpdir=/mnt '.coreos-sysroot-bwrap.tmp.XXXXXXXXXX')
+    mount --bind / "${TMP_CHROOT_DIR}"
+    mount --make-private "${TMP_CHROOT_DIR}"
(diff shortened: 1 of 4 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-sysusers.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Populate OSTree sysusers
```
jlebon: For consistency, let's make it e.g.

```suggestion
Description=Ignition OSTree: Create System Users
```

?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-sysroot-bwrap
```diff
@@ -0,0 +1,53 @@
...
+}
+
+cleanup() {
+  if test -z "${TMP_CHROOT_DIR}"; then
+    return
+  fi
+
+  umount --lazy --recursive "${TMP_CHROOT_DIR}"
+  umount --recursive "${TMP_CHROOT_DIR}"
+  rmdir "${TMP_CHROOT_DIR}"
```
jlebon: Super minor: `rm --dir "${TMP_CHROOT_DIR}"` avoids having to add `rmdir` to the initramfs.

jlebon (commented):
LGTM overall, though could we add an external test that tries to create a file owned by a user in a `sysusers.d` dropin? (E.g. `zincati`)

HuijingHei:
> could we add an external test that tries to create a file owned by a user in a `sysusers.d` dropin? (E.g. `zincati`)

Thanks @jlebon for the review!

Do you mean like following (use `dnsmasq` instead, as rhcos also includes this user)?
The interesting thing is the config works well without new `ignition-ostree-sysusers.service`, see https://github.com/coreos/fedora-coreos-config/pull/2670
```
$ cat config.bu
variant: fcos
version: 1.4.0
storage:
  files:
    - path: /etc/dnsmasq/config.d/00-dummy-placeholder.toml
      mode: 0644
      user:
        name: "dnsmasq"
      contents:
        inline: |
          # Dummy placeholder

$ cosa run --qemu-image fedora-coreos-38.20231009.20.0-qemu.x86_64.qcow2 -m 4096 -B config.bu

[core@cosa-devsh ~]$ ls -al /etc/dnsmasq/config.d/00-dummy-placeholder.toml
-rw-r--r--. 1 dnsmasq root 20 Oct 22 13:19 /etc/dnsmasq/config.d/00-dummy-placeholder.toml
```

HuijingHei:
@jlebon I am not sure how I can add special test for this patch, as all systemd sysusers are generated during build process, any pointers? Thanks!

PR #2673, opened 2023-10-15
[next-devel] overrides: fast-track downgraded packages

marmijo:
F39 is now in final freeze. This means some packages in F38 will sort as newer than packages in F39. We'll prevent downgrades by fast-tracking any packages that would violoate this "no downgrade" rule.

Today they are: aardvark-dns, fwupd, google-compute-engine-guest-configs, grub2, gvisor-tap-vsock, gcc, libnet, libsolv, netavark, passt, rpm-ostree, selinux-policy, vim, zchunk.

PR #2680, opened 2023-10-19

marmijo:
F39 is now in final freeze. This means some packages in F38 will sort as newer than packages in F39. We'll prevent downgrades by fast-tracking any packages that would violoate this "no downgrade" rule.

I missed `aardvark-dns` in last week's PR: https://github.com/coreos/fedora-coreos-config/pull/2673, so I'm adding it now. `kernel` is also on the downgraded list this week.


jlebon (approved):

dustymabe (approved):
LGTM

PR #2684, opened 2023-10-20
buildroot: Enable testing-devel

cgwalters:
Shorten the cycle for using new ostree in CI.

### DIFF
--- a/ci/buildroot/install-buildroot.sh
+++ b/ci/buildroot/install-buildroot.sh
@@ -3,6 +3,10 @@ set -euo pipefail
 
 # This is invoked by Dockerfile
 
+dnf -y install dnf-plugins-core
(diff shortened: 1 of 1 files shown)


cgwalters:
Motivated by https://github.com/coreos/rpm-ostree/pull/4640

dustymabe:
any potential downsides to doing this?

ci/buildroot/install-buildroot.sh
```diff
@@ -3,6 +3,9 @@ set -euo pipefail
 
 # This is invoked by Dockerfile
```
dustymabe, line 5: Optional: Might be useful to add a comment with rationale here for why this is useful

dustymabe (dismissed):
Code changes LGTM

cgwalters:
(This isn't tested by this CI, so merging)

PR #2685, opened 2023-10-23
tests/kernel-replace: update to f38 kernel; drop cruft

dustymabe:
Here we update to the kernel from F38 GA and also drop the rpm-ostree version check since all streams are past that. We also drop the kernel-modules-core check since that's no longer needed either.

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -44,7 +44,7 @@ image=oci:$image_dir
 image_pull=ostree-unverified-image:$image
 tmp_imagedir=/var/tmp/fcos-tmp
 arch=$(arch)
-kver="6.0.7-301.fc37.${arch}"
+kver="6.2.9-300.fc38.${arch}"
 
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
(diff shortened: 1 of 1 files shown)


dustymabe:
@jmarrero we might be able to also revisit https://github.com/coreos/fedora-coreos-config/blob/64f2dc320787eb27ea1d28f2932c2463ee925473/tests/kola/rpm-ostree/kernel-replace#L8-L10

cgwalters (approved):

jmarrero:
> @jmarrero we might be able to also revisit
> 
> https://github.com/coreos/fedora-coreos-config/blob/64f2dc320787eb27ea1d28f2932c2463ee925473/tests/kola/rpm-ostree/kernel-replace#L8-L10

cool, I will give it a try and see if looks good on RHCOS too and remove the restriction if it does.

jmarrero:
@dustymabe started to look at it and it seems like we already have this covered via the rt-kernel test on RHCOS which is failing currently. I am taking: https://github.com/openshift/os/issues/1383 which first part test the `override replace`.

dustymabe:
@jmarrero thanks! maybe we should update the comment in this test to mention this is covered in RHCOS by the kernel-rt test

PR #2686, opened 2023-10-23
[next] tree: promote changes from next-devel at c1e318a50c7ba241873d7a956bc5d1d50f52538b

coreosbot-releng:
Triggered by @prestist in https://github.com/coreos/fedora-coreos-streams/issues/794.


marmijo (approved):
LGTM

PR #2687, opened 2023-10-23
[rawhide] overrides: pin kexec-tools-2.0.27-1.fc40 on ppc64le only

marmijo:
Recent rawhide ppc64le builds have been failing with a newer version of kexec-tools. 
Let's pin on kexec-tools-2.0.27-1.fc40 until we can find a fix for: https://github.com/coreos/fedora-coreos-tracker/issues/1598


dustymabe:
it's worth noting that we typically pin a package for all architectures, but in this case, for rawhide, we don't ever ship rawhide to end users. Pinning for a specific architecture in that case seems OK. The only thing I'm not sure about is if the "autolock" stuff will kick in and reject the delta.

dustymabe (approved):
LGTM

PR #2689, opened 2023-10-24
denylist: extend snoozes for `kdump.crash`

gursewak1997:
For rawhide ppc64le failure, we don't have a fixed kernel yet. For the aarch64 selinux failure, it's been fixed in kexec-tools 2.0.27 which is now in F39. Will remove this once the prod streams have moved to F39

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -7,7 +7,7 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1430
-  snooze: 2023-10-24
+  snooze: 2023-11-15
   warn: true
   arches:
     - aarch64
@@ -29,7 +29,7 @@
     - rawhide
 - pattern: ext.config.kdump.crash
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1588
-  snooze: 2023-10-24
+  snooze: 2023-11-15
   warn: true
   arches:
     - ppc64le


dustymabe:
> For rawhide, as ppc64le builds fail with latest kexec-tools-2.0.27-3.fc40 we cannot test the kdump.crash against it.

Right, the builds fail with `2.0.27-3.fc40` but we [pinned](https://github.com/coreos/fedora-coreos-config/pull/2687) to `2.0.27-1.fc40` for ppc64le so that we could get a build. 

> Also we are still seeing kdump.crash failure for aarch64

https://github.com/coreos/fedora-coreos-tracker/issues/1430 should be fixed by kexec-tools 2.0.27, which we now have and https://github.com/coreos/fedora-coreos-tracker/issues/1560 should be fixed by a recent selinux RPM update.

Can you look into the failures to confirm they are the same?

dustymabe:
> > Also we are still seeing kdump.crash failure for aarch64
> 
> [coreos/fedora-coreos-tracker#1430](https://github.com/coreos/fedora-coreos-tracker/issues/1430) should be fixed by kexec-tools 2.0.27, which we now have and [coreos/fedora-coreos-tracker#1560](https://github.com/coreos/fedora-coreos-tracker/issues/1560) should be fixed by a recent selinux RPM update.

ahh yes. the denial is just for `stable`, `testing`, `testing-devel` where we don't have kexec-tools 2.0.27 yet. So extending that denial makes sense.

gursewak1997:
> Right, the builds fail with `2.0.27-3.fc40` but we [pinned](https://github.com/coreos/fedora-coreos-config/pull/2687) to `2.0.27-1.fc40` for ppc64le so that we could get a build.

True, apparently the problem now is that due to https://github.com/coreos/fedora-coreos-tracker/issues/1598, we cannot get past `cosa build` step and test the `kdump.crash`, hence why I extended the ppc64le denylist entry.
Since, we have the version pinned to the previous version, the test will fail.

dustymabe:
> True, apparently the problem now is that due to [coreos/fedora-coreos-tracker#1598](https://github.com/coreos/fedora-coreos-tracker/issues/1598), we cannot get past `cosa build` step and test the `kdump.crash`, hence why I extended the ppc64le denylist entry. Since, we have the version pinned to the previous version, the test will fail.]

The build failure shouldn't matter much here. We are pinned on the previous version, but IIUC the `2.0.27-3.fc40` kexec-tools doesn't actually have any fixes. We need a fixed kernel for that and it's [not really fixed upstream yet](https://bugzilla.redhat.com/show_bug.cgi?id=2241399#c11).

dustymabe:
I think we just need to update the commit message here and then we can merge this.

dustymabe (approved):
LGTM

PR #2690, opened 2023-10-26
manifests: include wasm packages in fedora-coreos-base manifest

dustymabe:
It's a bit confusing to folks that this didn't make it into the testing and stable streams [1]. Let's just include them everywhere like we agreed to originally.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1375#issuecomment-1750114083


benoitf:
it would be awesome

travier (approved):
Let's do this for now even though I don't really like it. Hopefully we can revisit this later.

PR #2691, opened 2023-10-26
[next-devel] remove wasm includes in manifest.yaml

dustymabe:
If https://github.com/coreos/fedora-coreos-config/pull/2690 merges we won't need this stream specific configuration.


dustymabe:
Marking as ready for review

travier (approved):

PR #2698, opened 2023-10-30
Add `sysusers` option in treefile

HuijingHei:
This is based on https://github.com/coreos/rpm-ostree/pull/4680


HuijingHei:
Do not merge this, it is just to test https://github.com/coreos/rpm-ostree/pull/4680

dustymabe:
@HuijingHei @travier is this still useful?

travier:
Probably not as is but we still need work in this area to finalize the migration to sysusers managed users.

PR #2702, opened 2023-10-30
05core: add support for root on iSCSI

jlebon:
See individual commit messages. The high-level overview here is that we just shuffle some services around and neuter others as needed so that we get out of the way of the existing iSCSI code in dracut when the right kargs are present (notably, `rd.iscsi.firmware` or `netroot`).


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
```diff
@@ -229,13 +229,22 @@ main() {
...
     load_dracut_libs
 
-    # Take down all interfaces set up in the initramfs
-    down_interfaces
+    # If we're using iSCSI, then we can't tear down networking since we'll lose
+    # root. This means in that case that the network config written to the real
+    # root won't be applied "from scratch". But anyway, since networking must
+    # stay on, it's simply not supported to configure the real root in a way
+    # that would require tearing down the connection on the interface involved.
+    if dracut_func getargbool 0 rd.iscsi.firmware || dracut_func getarg netroot; then
```
cgwalters, line 237: No objections to this as is, but wouldn't it be a bit cleaner to do in the systemd unit via e.g.
```
ConditionKernelCommandLine=!netroot
ConditionKernelCommandLine=!rd.iscsi.firmware
```
?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -110,6 +110,11 @@ if test -n "$(cmdline_arg rd.iscsi.firmware)" || test -n "$(cmdline_arg netroot)
     # activate ignition-fetch.service
     mkdir -p /run/ignition
     touch /run/ignition/neednet
+
+    # With iSCSI, configuring networking via kargs is a hard requirement
+    # since it can't be picked up from the bootfs (i.e. injected via
+    # `--copy-network`). So neuter `coreos-copy-firstboot-network.service`.
+    ln -sf /dev/null "${EARLY_DIR}/ignition-complete.target.requires/coreos-copy-firstboot-network.service"
```
cgwalters, line 117: I guess this works but it feels a bit less clean than changing the generator to enable the unit only if we're not on iscsi instead of disabling it.

cgwalters (commented):
I have trouble adding this new path into my head for the boot flow.  But I think the code makes sense.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -84,17 +85,37 @@ if ! $(cmdline_bool 'ignition.firstboot' 0); then
...
-
-Requires=coreos-gpt-setup.service
-After=coreos-gpt-setup.service
 EOF
+cp "${UNIT_DIR}/coreos-ignition-setup-user.service.d/diskful.conf" \
+    "${UNIT_DIR}/coreos-copy-firstboot-network.service.d/diskful.conf"
+
+# If booting from iSCSI, then we need networking first before we
+# can see the bootfs. This has some implications.
+if test -n "$(cmdline_arg rd.iscsi.firmware)" || test -n "$(cmdline_arg netroot)"; then
```
cgwalters, line 102: We have these kargs being tested in a few places...it's OK as is but maybe cleaner long term to have one generator check them and then create e.g. `/run/coreos-iscsi-root` if they're detected, and then all the systemd units can use `ConditionPathExists=!/run/coreos-iscsi-root` (or without `!`) as appropriate.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-teardown-initramfs.sh
jlebon: There's some logic in this script we still want to execute.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
jlebon: Yeah, I was tempted to do this... The thing is that `ignition-fetch-offline.service` doesn't live in this repo and I didn't want to "smear" the iSCSI implementation details across multiple repos. Maintenance-wise, it feels easier to follow if we have a single place where we can say "OK, if we're in the iSCSI case, these are all the things we do differently".

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
jlebon: Confusingly, that unit also needs to run in the live case, so the logic would be like: if we're in the live case OR we're not in the iSCSI case, then enable this unit. Except we currently exit quite early in this script in the live case, so it'd be a little more awkward to follow.

Another way to look at this is that we default to always running this unit since we almost always want it to run, and iSCSI is the exception. This relates to wanting to keep it centralized (see other comment).

cgwalters (dismissed):

jlebon:
Added another commit now to fix a bug with s390x Secure Execution Ignition decryption!

jlebon:
Can I get a restamp on this one? I've tested this pretty heavily now and it's looking good. We've also now effectively cleared f39 GA (we already have our `testing` build staged).

cgwalters (approved):

PR #2703, opened 2023-10-31
denylist: extend snooze for failing kola tests

marmijo:
ext.config.docker.basic and coreos.ignition.ssh.key are still failing, so let's extend the snooze on them while we wait for a fix.


aaradhak (approved):

dustymabe (approved):
LGTM

PR #2704, opened 2023-11-01
tests: create partition scheme test

lukewarmtemp:
Create a test to verify partition numbers and sizes. This is to ensure that the partition scheme is valid during the transition to using OSBuild in the FCOS pipeline.

Jira Issue: https://issues.redhat.com/browse/COS-2527

### DIFF
--- a/tests/kola/disks/partition-scheme
+++ b/tests/kola/disks/partition-scheme
@@ -1,0 +1,61 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   description: Verify the partition scheme is what we expect.
+##
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+root_part=$(findmnt -n -o SOURCE /sysroot)
+disk_name=$(lsblk --json -o PKNAME --path "$root_part" | jq --raw-output '.blockdevices[].pkname')
+diskData=$(sfdisk --json "$disk_name" | jq '.partitiontable' )
+partitionData=$(echo $diskData | jq '.partitions[]')
+totalPartitions=$(echo $diskData | jq '.partitions | length')
+sector_size=$(echo $diskData | jq .sectorsize)
+
+if [[ $totalPartitions -ne 4 ]]; then
+    fatal "Expected 4 partitions, got $totalPartitions"
+fi
+
+# check sizes of each parition
+ONE_MiB=$(( 1024 * 1024 ))
+
+# An associative array that maps the partition label to the size (in MiB)
+# of the partition. For the root partition we set it to "" to skip the check
+# there because the growfs service runs on first boot. Checking start
+# should be enough there.
+declare -A expected=(
+    ["BIOS-BOOT"]="1"
+    ["EFI-SYSTEM"]="127"
+    ["boot"]="384"
+    ["root"]=""
+)
+
+# There is a 1MiB gap at the beginning of the disks
+expected_start=$(( 1 * $ONE_MiB / $sector_size ))
+
+# Iterate over the partitions and check their start and size
+for key in "${!expected[@]}"; do
+    size_MiB="${expected[${key}]}"
+    start=$(echo "$partitionData" | jq "select ( .name == \"$key\") | .start")
+    sectors=$(echo "$partitionData" | jq "select ( .name == \"$key\") | .size")
+    size=$(( $sectors * $sector_size ))
+    if [[ "$start" -ne "$expected_start" ]]; then
+        fatal "Expected $key partition start sector of $expected_start, got $start"
+    fi
+    if [ ! -z "$size_MiB" ]; then
+        expected_size=$(($size_MiB * $ONE_MiB))
+        if [[ "$expected_size" -ne "$size" ]]; then
+            fatal "Expected $key partition of size $expected_size, got $size"
+        fi
+    fi
+    # The expected start of the next partition will be the start of this partition
+    # plus the size of this partition.
+    expected_start=$(($expected_start + $size / $sector_size))
+done
+
+ok partition scheme
+exit 0


tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,41 @@
...
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+# check for 4 partitions
+totalPartitions=$(grep -c 'vda[0-9]' /proc/partitions)
+if [[ $totalPartitions -ne 4 ]]; then
+    fatal "Expected 4 partitions, got $totalPartitions"
+fi
```
dustymabe: Since `/dev/vd*` won't be the disk naming used on every platform I think there is a better heuristic we can use here.

Why don't we just find the partition related to our root filesystem and then the backing device behind that and then we can pass that device into the `partitionData=$(sudo lsblk` below to limit the output to only things related to that device?

```
[core@cosa-devsh ~]$ findmnt -nvr -o SOURCE /sysroot
/dev/vda4
[core@cosa-devsh ~[core@cosa-devsh ~]$ lsblk -o PKNAME --pairs  /dev/vda4
PKNAME="vda"]$ lsblk -o PKNAME --pairs --paths /dev/vda4
PKNAME="/dev/vda"
```

then we can check the number of children to verify we only got 4 partitions.

```
[core@cosa-devsh ~]$ sudo lsblk --json --bytes -o NAME,FSTYPE,SIZE,MOUNTPOINT,PARTLABEL,PARTTYPENAME /dev/vda | jq '.blockdevices[].children | length'
4
```

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,41 @@
...
+
+# get partition sizes
+partitionData=$(sudo lsblk --json -ba -o NAME,FSTYPE,SIZE,MOUNTPOINT,PARTLABEL,PARTTYPENAME |  jq .blockdevices[].children[] )
+biosBootSize=$( echo "$partitionData" | jq 'select ( .partlabel == "BIOS-BOOT" or .parttypename == "BIOS boot") | .size ' )
+efiSystemSize=$( echo "$partitionData" | jq 'select ( .partlabel == "EFI-SYSTEM" or .parttypename == "EFI System") | .size ' )
+bootSize=$( echo "$partitionData" | jq 'select ( .partlabel == "boot" or .mountpoint == "/boot") | .size ' )
+rootSize=$( echo "$partitionData" | jq 'select ( .partlabel == "root" or .mountpoint == "/sysroot") | .size ' )
+
+# check sizes of each parition
+if [[ $biosBootSize -ne 1048576 ]]; then
```
dustymabe: it might be easier to read this code in the future if we break it down to something easier to understand:

```
ONE_MiB=$(( 1024 * 1024 ))
expectedBIOSBootSize=$((1 * $ONE_MiB))
expectedEFISystemSize=$((127 * $ONE_MiB))
expectedBootSize=$((384 * $ONE_MiB))
etc..

File: tests/kola/disks/partition-scheme [hunk 3]
```diff
@@ -0,0 +1,41 @@
...
+if [[ $efiSystemSize -ne 133169152 ]]; then
+    fatal "Expected EFI-SYSTEM partition of size 133169152, got $efiSystemSize"
+fi
+
+if [[ $bootSize -ne 402653184 ]]; then
+    fatal "Expected boot partition of size 402653184, got $bootSize"
+fi
+
+if [[ $rootSize -ne 10199481856 ]]; then
+    fatal "Expected root partition of size 10199481856, got $rootSize"
```
dustymabe: I'm not sure across platforms we can verify the root partition size of a booted system (i.e. growpart/growfs runs on first boot). Maybe what we should do here is just verify the starting sector is what we expect.

dustymabe:
we might need to add some special consideration here for 4k disks, though hmm, do these tests even run for 4k disks or is that just testiso tests?

lukewarmtemp:
> we might need to add some special consideration here for 4k disks, though hmm, do these tests even run for 4k disks or is that just testiso tests?

Oh it didn't even cross my mind to test the different platforms. I just ran it on the `qcow2` built by OSBuild and COSA, but I can try running it for different platforms and see what the outputs are.

dustymabe:
> > we might need to add some special consideration here for 4k disks, though hmm, do these tests even run for 4k disks or is that just testiso tests?
> 
> Oh it didn't even cross my mind to test the different platforms. I just ran it on the `qcow2` built by OSBuild and COSA, but I can try running it for different platforms and see what the outputs are.

Like I said 4k disks probably won't be able to be tested with this because I think the only place we can use those is for bare metal and we only test the bare metal images using our `testiso` stuff.

But... having this test run everywhere (like AWS, GCP, etc) is useful, so making it a test that isn't limited to `qemu` will have value I think.

tests/kola/disks/partition-scheme
lukewarmtemp: Not sure if there's a better way to find the starting sector, but my current approach is to grep the journal logs for when the `growfs` service occurs:
```
$ journalctl -b | grep 'growfs' | grep 'start'
Nov 15 14:21:51 localhost ignition-ostree-growfs[990]: CHANGED: partition=4 start=1050624 old: size=4194304 end=5244927 new: size=19920863 end=20971486
```
I'm assuming that the starting sector is `start=1050624`, but it doesn't really divide cleanly into `ONE_MiB=$(( 1024 * 1024 ))`.

dustymabe: should just add `START` to your `lsblk` call: 


```
sudo lsblk --json -ba -o NAME,FSTYPE,START,SIZE,MOUNTPOINT,PARTLABEL,PARTTYPENAME
```

and then you can pull it from that.

lukewarmtemp:
> But... having this test run everywhere (like AWS, GCP, etc) is useful, so making it a test that isn't limited to qemu will have value I think.

Test runs on AWS successfully.

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,51 @@
...
+
+# check for 4 partitions
+root_part=$(findmnt -n -o SOURCE /sysroot)
+disk_name=$(lsblk --json -o PKNAME --path "$root_part" | jq --raw-output '.blockdevices[].pkname')
+totalPartitions=$(sudo lsblk --json --bytes -o NAME,FSTYPE,START,SIZE,MOUNTPOINT,PARTLABEL,PARTTYPENAME "$disk_name" | jq '.blockdevices[].children | length')
+if [[ $totalPartitions -ne 4 ]]; then
+    fatal "Expected 4 partitions, got $totalPartitions"
+fi
+
+# get partition sizes
```
dustymabe: Can we check both start and size of each partition (except root)? 

```suggestion
# get partition data and then verify the starting sector and size of all we can
```

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,51 @@
...
+bootSize=$( echo "$partitionData" | jq 'select ( .partlabel == "boot" or .mountpoint == "/boot") | .size ' )
+# get start size of root partition since the growfs service runs on first boot
+rootSize=$( echo "$partitionData" | jq 'select ( .partlabel == "root" or .mountpoint == "/sysroot") | .start ' )
+
+# check sizes of each parition
+ONE_MiB=$(( 1024 * 1024 ))
+expectedBIOSBootSize=$((1 * $ONE_MiB))
+expectedEFISystemSize=$((127 * $ONE_MiB))
+expectedBootSize=$((384 * $ONE_MiB))
+expectedRootSize=$(($ONE_MiB + 2048 ))
```
dustymabe:
```suggestion
expectedRootStart=$(($ONE_MiB + 2048 ))
```

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   description: Verify the partition scheme is correct.
```
dustymabe: I think the word "correct" here can be improved upon:

```suggestion
##   description: Verify the partition scheme is what we expect.
```

lukewarmtemp:
A point of investigation: the `qcow2` created by OSBuild does not have the `partlabel` field
```
lsblk --json --bytes -o NAME,FSTYPE,START,SIZE,MOUNTPOINT,PARTLABEL,PARTTYPENAME,LOG-SEC "$disk_name" | jq '.blockdevices[]'
{
  "name": "vda",
  "fstype": null,
  "start": null,
  "size": 10737418240,
  "mountpoint": null,
  "partlabel": null,
  "parttypename": null,
  "log-sec": 512,
  "children": [
    {
      "name": "vda1",
      "fstype": null,
      "start": 2048,
      "size": 1048576,
      "mountpoint": null,
      "partlabel": null,
      "parttypename": "BIOS boot",
      "log-sec": 512
    },
    {
      "name": "vda2",
      "fstype": "vfat",
      "start": 4096,
      "size": 133169152,
      "mountpoint": null,
      "partlabel": null,
      "parttypename": "EFI System",
      "log-sec": 512
    },
    {
      "name": "vda3",
      "fstype": "ext4",
      "start": 264192,
      "size": 402653184,
      "mountpoint": "/boot",
      "partlabel": null,
      "parttypename": "Linux filesystem",
      "log-sec": 512
    },
    {
      "name": "vda4",
      "fstype": "xfs",
      "start": 1050624,
      "size": 10199481856,
      "mountpoint": "/sysroot",
      "partlabel": null,
      "parttypename": "Linux filesystem",
      "log-sec": 512
    }
  ]
}
```

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,61 @@
...
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+root_part=$(findmnt -n -o SOURCE /sysroot)
+disk_name=$(lsblk --json -o PKNAME --path "$root_part" | jq --raw-output '.blockdevices[].pkname')
+diskData=$(lsblk --json --bytes -o NAME,FSTYPE,START,SIZE,MOUNTPOINT,PARTLABEL,PARTTYPENAME,LOG-SEC "$disk_name" | jq '.blockdevices[]' )
+partitionData=$(echo $diskData | jq '.children[]')
+totalPartitions=$(echo $diskData | jq '.children | length')
+sector_size=$(echo $diskData | jq '.["log-sec"]')
```
dustymabe: ok given our email thread with kzak let's just set this to 512 with something like:

```suggestion
# lsblk always reports `START` in sectors that are 512 bytes even if
# the disk is a 4k native disk with 4096 byte sectors. Let's hard code
# the value here and do our math based on that. See
# https://github.com/karelzak/util-linux/commit/61c40f484563881420e1606a27663aee930cbe53
sector_size=512
```

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,61 @@
...
+    ["boot"]="384"
+    ["root"]=""
+)
+
+# There is a 1MiB gap at the beginning of the disks
+expected_start=$(( 1 * $ONE_MiB / $sector_size ))
+
+# Iterate over the partitions and check their 
+for key in ${!expected[@]}; do
+    echo "${key}, ${expected[${key}]}"
```
dustymabe: Probably can drop this

```suggestion
```

dustymabe:
> A point of investigation: the `qcow2` created by OSBuild does not have the `partlabel` field

yep. definitely something that may be causing us problems. We'll need to remedy that I think.

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,68 @@
...
+)
+
+# There is a 1MiB gap at the beginning of the disks
+expected_start=$(( 1 * $ONE_MiB / $sector_size ))
+
+# Iterate over the partitions and check their 
+for key in ${!expected[@]}; do
+    size_in_sectors="${expected[${key}]}"
+    start=$(echo "$partitionData" | jq "select ( .name == \"$key\") | .start")
+    size=$(echo "$partitionData" | jq "select ( .name == \"$key\") | .size")
```
dustymabe: rather than change up all the other pieces I think we could just:


```suggestion
    sectors=$(echo "$partitionData" | jq "select ( .name == \"$key\") | .size")
    size=$(( $sectors * $sector_size ))
```

which I think would make things like `SECTOR_UNIT=$(( $ONE_MiB / $sector_size ))` not needed?

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,68 @@
...
+partitionData=$(echo $diskData | jq '.partitions[]')
+totalPartitions=$(echo $diskData | jq '.partitions | length')
+
+# sfdisk --json "$disk_name" | jq '.blockdevices[]'
+
+# lsblk always reports `START` in sectors that are 512 bytes even if
+# the disk is a 4k native disk with 4096 byte sectors. Let's hard code
+# the value here and do our math based on that. See
+# https://github.com/karelzak/util-linux/commit/61c40f484563881420e1606a27663aee930cbe53
+sector_size=512
```
dustymabe: one of the reasons to move to `sfdisk` is that the sector size reported is accurate (unlike `lsblk`) - let's pick this value up from the json?

dustymabe: See the `sectorsize` in the top level:

```
$ sudo sfdisk --json /dev/mmcblk1 | jq .
{
  "partitiontable": {
    "label": "gpt",
    "id": "1C8052E8-AABC-4586-88A5-95F2909B78DB",
    "device": "/dev/mmcblk1",
    "unit": "sectors",
    "firstlba": 34,
    "lastlba": 123596766,
    "sectorsize": 512,
    "partitions": [
      {
        "node": "/dev/mmcblk1p1",
        "start": 2048,
        "size": 2048,
        "type": "8DA63339-0007-60C0-C436-083AC8230908",
        "uuid": "71ACCFA0-8B5A-4C3D-9D0A-9BEDBF75AFC5",
        "name": "reserved"
      },
      {
        "node": "/dev/mmcblk1p2",
        "start": 4096,
        "size": 260096,
        "type": "C12A7328-F81F-11D2-BA4B-00A0C93EC93B",
        "uuid": "8F1053AC-D3A3-407C-9FB4-A09982575E95",
        "name": "EFI-SYSTEM"
      },
      {
        "node": "/dev/mmcblk1p3",
        "start": 264192,
        "size": 786432,
        "type": "0FC63DAF-8483-4772-8E79-3D69D8477DE4",
        "uuid": "A7D14714-1354-4A34-ABCB-DDECC1C2C5BB",
        "name": "boot"
      },
      {
        "node": "/dev/mmcblk1p4",
        "start": 1050624,
        "size": 122546143,
        "type": "0FC63DAF-8483-4772-8E79-3D69D8477DE4",
        "uuid": "FF042EB1-48D1-4239-B605-D4D3965FAAEF",
        "name": "root"
      }
    ]
  }
}
```

dustymabe:
I think this LGTM - can we fix the CI failures? Once we do that I'll give it a spin on a 4k disk just to verify it works (I know it won't ever get run against a 4k disk, but in theory it should just work).

lukewarmtemp:
>can we fix the CI failures

Not really sure why it's failing... I used `truncate -s -1` to get rid of the whitespace at the end of the file.

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,61 @@
...
+    ["BIOS-BOOT"]="1"
+    ["EFI-SYSTEM"]="127"
+    ["boot"]="384"
+    ["root"]=""
+)
+
+# There is a 1MiB gap at the beginning of the disks
+expected_start=$(( 1 * $ONE_MiB / $sector_size ))
+
+# Iterate over the partitions and check their
```
dustymabe: Here is the white space

```suggestion
# Iterate over the partitions and check their
```

though we probably want to finish the comment (seems like an incomplete thought).

tests/kola/disks/partition-scheme
```diff
@@ -0,0 +1,61 @@
...
+    ["EFI-SYSTEM"]="127"
+    ["boot"]="384"
+    ["root"]=""
+)
+
+# There is a 1MiB gap at the beginning of the disks
+expected_start=$(( 1 * $ONE_MiB / $sector_size ))
+
+# Iterate over the partitions and check their start and size
+for key in ${!expected[@]}; do
```
dustymabe: I think shellcheck is wanting you to do:

```suggestion
for key in "${!expected[@]}"; do
```

dustymabe:
let's squash this into one commit?

dustymabe (approved):
LGTM

PR #2707, opened 2023-11-02
[next] tree: promote changes from next-devel at 0554612314b50355b19ecbb95b65e912ce2c10e9

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/796.


dustymabe:
Not sure if we'll get an updated SELinux RPM or not, but just in case we don't I'm starting CI for this early.

jlebon (approved):

PR #2714, opened 2023-11-06
overrides: drop arch specific overrides

dustymabe:
These are all stable or obsolete now so we can drop the arch specific overrides.


dustymabe:
Will skip CI for this once approved and jump directly to the bump-lockfile job.

marmijo (approved):
LGTM

PR #2710, opened 2023-11-03
[testing] tree: promote changes from next at 1a3b193

dustymabe:

PR #2716, opened 2023-11-06
[stable] tree: promote changes from testing at b374b765d90e3a2f8ae2ac373c934424fc152d3a

coreosbot-releng:
Triggered by @marmijo in https://github.com/coreos/fedora-coreos-streams/issues/806.

### DIFF
--- a/ci/buildroot/install-buildroot.sh
+++ b/ci/buildroot/install-buildroot.sh
@@ -3,6 +3,10 @@ set -euo pipefail
 
 # This is invoked by Dockerfile
 
+dnf -y install dnf-plugins-core
+# We want to avoid a 7 day cycle for e.g. new ostree etc.
+dnf config-manager --set-enabled updates-testing
+
 dn=$(dirname "$0")
 tmpd=$(mktemp -d) && trap 'rm -rf ${tmpd}' EXIT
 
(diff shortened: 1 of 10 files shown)


dustymabe:
Hmm. so this is actually not right. We pre-emptively promoted `testing` from `next` in https://github.com/coreos/fedora-coreos-config/pull/2710 last Friday when the content set for GA was finalized and that messes up the flow here.

What we need to do is promote from b374b76 which was the last `testing` release. I can fix this up locally and push here.

dustymabe:
> I can fix this up locally and push here.

Done

marmijo:
check promotion diffs is failing because the original promotion in this PR was from a different commit. 
```
env:
    COMMIT_TITLE: [stable] tree: promote changes from testing at 7e2d1e47aa9dde
```
I tried to rerun it but it didn't pick up the new commit.

marmijo:
It doesnt seem like the check promotion diffs job is picking up the new commit from the PR title. I've manually checked the diff between https://github.com/coreos/fedora-coreos-config/commit/b374b765d90e3a2f8ae2ac373c934424fc152d3a and https://github.com/coreosbot-releng/fedora-coreos-config/commit/2c8b88894a23f33e86dd06e0561ce5699b3c3313:
``` diff
index fbb095f3..c8b774aa 100644
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,5 +1,5 @@
 variables:
-  stream: stable
+  stream: testing
   prod: true
 
 releasever: 38

```

marmijo (approved):

PR #2693, opened 2023-10-26
Move to Fedora 39 #1490

yasminvalim:
This PR, detailed in Issue [#1490](https://github.com/coreos/fedora-coreos-tracker/issues/1490) , follows the necessary steps to prepares for the upcoming Fedora CoreOS 39 GA release on October 30th by updating fedora-coreos-config.

PR #2719, opened 2023-11-07
Move to Fedora 39 #1490

yasminvalim:
This PR, [detailed in Issue]( https://github.com/coreos/fedora-coreos-tracker/issues/1490) , follows the necessary steps to prepares for the upcoming Fedora CoreOS 39 GA release by updating fedora-coreos-config.

FYI: Previously, there was another branch, but I set it as a draft and started working on this new one after running into some issues with the rebase. If you'd like to review the discussion on t[hat pull request,](https://github.com/coreos/fedora-coreos-config/pull/2693) please feel free to do so.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,8 +9,8 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
-  moby-engine:
-    evr: 20.10.23-1.fc38
+  containerd:
+    evr: 1.6.19-2.fc39
     metadata:
-      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1476
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1578
       type: pin
--- a/ci/buildroot/Dockerfile
+++ b/ci/buildroot/Dockerfile
@@ -5,6 +5,6 @@
 #
 # This image is used by CoreOS CI to build software like
 # Ignition, rpm-ostree, ostree, coreos-installer, etc...
-FROM quay.io/fedora/fedora:38
+FROM quay.io/fedora/fedora:39
 COPY . /src
 RUN ./src/install-buildroot.sh && yum clean all && rm /src -rf
--- a/image.yaml
+++ b/image.yaml
@@ -2,9 +2,3 @@
 # similarly to manifest.yaml. Unlike image-base.yaml, which is shared by all
 # streams.
 include: image-base.yaml
-
-# see https://github.com/coreos/coreos-assembler/pull/3607
-# Defaults for AWS
-aws-imdsv2-only: false
-aws-volume-type: "gp2"
-aws-x86-boot-mode: "legacy-bios"
--- a/manifest-lock.aarch64.json
+++ b/manifest-lock.aarch64.json
@@ -982,10 +982,10 @@
       "evra": "1.81-3.fc38.aarch64"
     },
     "ostree": {
-      "evra": "2023.7-2.fc38.aarch64"
+      "evra": "2023.6-1.fc38.aarch64"
     },
     "ostree-libs": {
-      "evra": "2023.7-2.fc38.aarch64"
+      "evra": "2023.6-1.fc38.aarch64"
     },
     "p11-kit": {
       "evra": "0.25.2-1.fc38.aarch64"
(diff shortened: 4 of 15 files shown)


jlebon:
This is looking close! I think we also want the `image.yaml` from `next-devel`.

Also, it looks like there's some minor end of file newline differences in the lockfiles (this is minor but it shows up in a `git diff` between the two, and it'd be nice to have that diff be minimized to the strict necessary). Something like this should work:

```
git restore --source=origin/next-devel image.yaml manifest-lock*
git add image.yaml manifest-lock*
git commit --amend --no-edit
```

manifest-lock.overrides.yaml
```diff
@@ -9,8 +9,20 @@
...
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-24872e50a0
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1490#issuecomment-1720020468
+      type: fast-track
+  selinux-policy-targeted:
+    evra: 39.1-1.fc39.noarch
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-24872e50a0
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1490#issuecomment-1720020468
+      type: fast-track
```
dustymabe: note that `next-devel` just dropped this in https://github.com/coreos/fedora-coreos-config/pull/2718

yasminvalim: Thanks for that. Fixed it.

yasminvalim:
> This is looking close! I think we also want the `image.yaml` from `next-devel`.
> 
> Also, it looks like there's some minor end of file newline differences in the lockfiles (this is minor but it shows up in a `git diff` between the two, and it'd be nice to have that diff be minimized to the strict necessary). Something like this should work:
> 
> ```
> git restore --source=origin/next-devel image.yaml manifest-lock*
> git add image.yaml manifest-lock*
> git commit --amend --no-edit
> ```

Thanks for reviewing! I added the `image.yaml` now. 
This newline differences are some default configuration in VScode, I am fixing it right now.

yasminvalim:
> This is looking close! I think we also want the `image.yaml` from `next-devel`.
> 
> Also, it looks like there's some minor end of file newline differences in the lockfiles (this is minor but it shows up in a `git diff` between the two, and it'd be nice to have that diff be minimized to the strict necessary). Something like this should work:
> 
> ```
> git restore --source=origin/next-devel image.yaml manifest-lock*
> git add image.yaml manifest-lock*
> git commit --amend --no-edit
> ```

I have two questions: 

**1 -**  In the `manifest.yaml`  file in next-devel, there is this part at the end:
```

conditional-include:
  # we only want these in next for now; see
  # https://github.com/coreos/fedora-coreos-tracker/issues/1375
  - if: basearch == "x86_64"
    include: manifests/crun-wasm.yaml
  - if: basearch == "aarch64"
    include: manifests/crun-wasm.yaml
```
Should it be included too?

**2 -** The other question is about the red circles in my PR who suggest that there is a empty line in some files. I already fix it in my vscode settings and committed  applying the changes, but is still appearing here. Do you have any idea how to solve it?

Thanks :smile_cat:

jlebon:
> I have two questions:
> 
> **1 -** In the `manifest.yaml` file in next-devel, there is this part at the end:
> 
> ```
> 
> conditional-include:
>   # we only want these in next for now; see
>   # https://github.com/coreos/fedora-coreos-tracker/issues/1375
>   - if: basearch == "x86_64"
>     include: manifests/crun-wasm.yaml
>   - if: basearch == "aarch64"
>     include: manifests/crun-wasm.yaml
> ```

For now, no. We may remove that bit soon separately (see https://github.com/coreos/fedora-coreos-config/pull/2690).

> Should it be included too?
> 
> **2 -** The other question is about the red circles in my PR who suggest that there is a empty line in some files. I already fix it in my vscode settings and committed applying the changes, but is still appearing here. Do you have any idea how to solve it?

Hmm, I'm not sure. It may still be VS Code playing tricks. The surest way to squash those is to do it from the command-line instead using the commands I provided in https://github.com/coreos/fedora-coreos-config/pull/2719#issuecomment-1799291638.

yasminvalim:
@jlebon  I was able to fix it, finally. I think everything is okay now. 
Thanks a lot! :smile_cat:

jlebon:
OK, this is looking great! But the `manifest.yaml` has unwanted changes. I might've misunderstood your question 1 in the previous comment. Apart from the `releasever` change, `manifest.yaml` needs to stay the same.

jlebon (dismissed):
LGTM, thanks for doing this!

dustymabe (approved):

PR #2721, opened 2023-11-08
[testing] overrides: pin rust-zincati-0.0.25-6.fc39

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -183,7 +183,7 @@ packages:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1490#issuecomment-1720020468
       type: fast-track
   zincati:
-    evr: 0.0.26-2.fc39
+    evr: 0.0.25-6.fc39
     metadata:
-      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-586e8dc477
-      type: fast-track
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1608#issuecomment-1802126860
+      type: pin
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -185,5 +185,5 @@ packages:
   zincati:
     evr: 0.0.25-6.fc39
     metadata:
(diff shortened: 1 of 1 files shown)


manifest-lock.overrides.yaml
```diff
@@ -183,7 +183,7 @@ packages:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1490#issuecomment-1720020468
       type: fast-track
   zincati:
-    evr: 0.0.26-2.fc39
+    evr: 0.0.25-6.fc39
     metadata:
-      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2023-586e8dc477
-      type: fast-track
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1608#issuecomment-1802126860
```
jlebon: https://github.com/coreos/fedora-coreos-tracker/issues/1608#issuecomment-1802057755 seems like a better comment for this?

jlebon (dismissed):

manifest-lock.overrides.yaml
dustymabe: copy paste error. I didn't intend for it to be a comment at all.

dustymabe:
```
[2023-11-09T02:23:56.369Z] ostree diff commit from: 694d31bf9cdf1d365f7870e73f5c4d17b71e37562204beddcb76f587ecb65ca3
[2023-11-09T02:23:56.369Z] ostree diff commit to:   66f015edda80ccf46669eb38a338086f3ecc4246978cc9ec27c1018fccad6aa8
[2023-11-09T02:23:56.369Z] Downgraded:
[2023-11-09T02:23:56.369Z]   zincati 0.0.26-2.fc39 -> 0.0.25-6.fc39
[2023-11-09T02:23:56.369Z] + grep -q Downgraded tmp/diff.txt
[2023-11-09T02:23:56.369Z] + echo 'Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.'
[2023-11-09T02:23:56.369Z] Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.
```

👍 - this is what we want!

PR #2722, opened 2023-11-08
overrides: pin rust-zincati-0.0.25-6.fc39

coreosbot-releng:


jlebon (approved):

jlebon:
CI failing on downgrade check, which is expected.

PR #2723, opened 2023-11-08
manifest: Enable ostree bootloader-naming-2 for f40

cgwalters:
- This removes the autoprune stuff which is already always on in ostree now
- Add the new "experiment" of bootloader-naming-2; xref https://github.com/ostreedev/ostree/pull/3007


manifests/fedora-coreos.yaml
```diff
@@ -23,19 +23,9 @@ conditional-include:
...
-    # https://github.com/coreos/fedora-coreos-tracker/issues/1495#issuecomment-1561765705
-    include: ostree-autoprune.yaml
   - if:
-      - basearch == "aarch64"
-      - releasever >= 39
-    # In F39+ we will stop removing the qcom dtb files and thus we'll leverage OSTree autopruning
-    # so that we don't run into https://github.com/coreos/fedora-coreos-tracker/issues/1464
-    # again. OSTree autopruning is new so we're selectively enabling it before making it the default.
-    include: ostree-autoprune.yaml
+      - releasever >= 40
```
jlebon, line 27: Minor: this can be squashed to a single line. It only needs to be a list if there are multiple conditions.

jlebon (approved):
LGTM as is.

dustymabe:
> * This removes the autoprune stuff which is already always on in ostree now

oh nice. I missed that [this](https://github.com/ostreedev/ostree/pull/3012) merged and made it in [2023.7](https://github.com/ostreedev/ostree/releases/tag/v2023.7). I did check and it is in `testing-devel` and `next-devel` so we should be good.

dustymabe:
Looks good:

```
[core@cosa-devsh ~]$ ls /boot/loader/entries/
ostree-1.conf  ostree-2.conf
State: idle
Deployments:
● fedora-compose:fedora/x86_64/coreos/rawhide
                  Version: 40.20231108.91.0 (2023-11-08T12:47:04Z)
                   Commit: a8f7ddc642f751e71ba514d3b2a94e8944360352fda9349fa5d12a3e72a572cf

  fedora:fedora/x86_64/coreos/rawhide
                  Version: 40.20231109.dev.0 (2023-11-09T19:45:24Z)
                   Commit: 8d5f9ca25e88b8c9062dfd39c4abf076641975b8744de56a3de4ab23bad6629f
             GPGSignature: (unsigned)

```

Though, I don't know if the desire is for the bls entry filename in the delivered image (qcow2) to have this form. If it is that isn't what happens IIUC.

dustymabe:
if you rebase on top of latest `testing-devel` I think it will fix the CI failure.

dustymabe (approved):
LGTM

cgwalters:
Yeah, it's different work to make it affect the disk images too, would need an image.yaml knob I guess.

PR #2725, opened 2023-11-09
tests: verify systemd-sysuser.service with `ConditionNeedsUpdate=` will be triggered with new deployment

HuijingHei:
See https://github.com/ostreedev/ostree/issues/3069#issuecomment-1798115799

### DIFF
--- a/tests/kola/systemd/condition-needs-update
+++ b/tests/kola/systemd/condition-needs-update
@@ -1,0 +1,64 @@
+#!/bin/bash
+## kola:
+##   tags: "platform-independent"
+##   description: Verify systemd-sysuser.service with `ConditionNeedsUpdate=`
+##     will be triggered with new deployment at boot time.
+
+# See https://github.com/ostreedev/ostree/issues/3069#issuecomment-1798115799
+
+# Systemd units using ConditionNeedsUpdate= run if the mtime of .updated in
+# the specified directory is newer than /usr. Since /usr has an mtime of
+# 0, there's no way to have an older .updated file refer to
+# https://ostreedev.github.io/ostree/repo/#content-objects. Systemd units
+# typically specify ConditionNeedsUpdate=/etc or ConditionNeedsUpdate=/var
+# to support stateless systems like ostree.
+
+# Remove the file .updated from the new deployment's /etc and the OS's /var
+# regardless of where they came from to ensure that these systemd units
+# run when booting new deployments, see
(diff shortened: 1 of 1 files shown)


tests/kola/systemd/condition-needs-update
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+## kola:
+##   description: Verify systemd-sysuser.service with `ConditionNeedsUpdate=`
+##     will be triggered with new deployment at boot time.
```
jlebon, line 5: Can we mark this `platform-independent`?

tests/kola/systemd/condition-needs-update
```diff
@@ -0,0 +1,59 @@
...
+    if getent passwd ${username}; then
+      fatal "should not get user ${username} after second boot"
+    fi
+    # create a new deployment
+    ref=$(rpm-ostree status --json | jq -r '.deployments[0].origin')
+    ostree admin deploy --karg-append=somedummykarg=1 $ref
+    /tmp/autopkgtest-reboot third-boot
+    ;;
+
+  third-boot)
```
jlebon, line 51: Should we be checking in this block that `somedummykarg` is present in `/proc/cmdline`?

tests/kola/systemd/condition-needs-update
```diff
@@ -0,0 +1,59 @@
...
+
+  second-boot)
+    # reboot to check user footest not created
+    ok "second boot"
+    if getent passwd ${username}; then
+      fatal "should not get user ${username} after second boot"
+    fi
+    # create a new deployment
+    ref=$(rpm-ostree status --json | jq -r '.deployments[0].origin')
+    ostree admin deploy --karg-append=somedummykarg=1 $ref
```
jlebon: Minor: this could be `rpm-ostree kargs --append=...`.

jlebon (dismissed):
LGTM overall, thanks!

tests/kola/systemd/condition-needs-update
HuijingHei: Sounds more reasonable, thanks!

tests/kola/systemd/condition-needs-update
HuijingHei: Updated, thanks!

jlebon (approved):

HuijingHei:
Thanks @jlebon for the review!

travier:
Looks great. Thanks!

PR #2729, opened 2023-11-09
tests/kernel-replace: Adapt to already being in a container

cgwalters:
This is so far the only test here which breaks when we're first booted into a container.  rpm-ostree's own tests are *so* not ready for this, but we can get to that later...


cgwalters:
Blah, shellcheck...ok, should be fixed now

HuijingHei (approved):
LGTM

PR #2731, opened 2023-11-13
kola/butane/grub-users: Update the test

gursewak1997:
Adapt the `ext.config.butane.grub-users` test as per the changes from https://github.com/coreos/fedora-coreos-config/pull/2723. Now, we git grep from `ostree-x.conf` instead of `ostress-x-fedora-coreos.conf`. Keeping the test to grep against `ostree-x*.conf` so we can still run the test with old file names.

### DIFF
--- a/tests/kola/butane/grub-users/test.sh
+++ b/tests/kola/butane/grub-users/test.sh
@@ -36,18 +36,18 @@ rebooted)
         fatal "Rebooted into old deployment"
     fi
     # cross-check karg with BLS configs
-    if grep -q test-added-karg /boot/loader.0/entries/ostree-1-fedora-coreos.conf; then
+    if grep -q test-added-karg /boot/loader.0/entries/ostree-1*.conf; then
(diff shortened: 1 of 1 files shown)


tests/kola/butane/grub-users/test.sh
```diff
@@ -36,18 +36,18 @@ rebooted)
         fatal "Rebooted into old deployment"
     fi
     # cross-check karg with BLS configs
-    if grep -q test-added-karg /boot/loader.0/entries/ostree-1-fedora-coreos.conf; then
+    if grep -q test-added-karg /boot/loader.0/entries/ostree-1.conf; then
```
jlebon: We need the test to work with the old filename too. Maybe simplest is just e.g.

```suggestion
    if grep -q test-added-karg /boot/loader.0/entries/ostree-1*.conf; then
```

marmijo (approved):
LGTM

PR #2733, opened 2023-11-17
denylist: skip PXE tests on ppc64le

jlebon:
These tests are broken with the latest GRUB BOOTP binaries in cosa, which is now on f39.


dustymabe (approved):
LGTM

PR #2735, opened 2023-11-20
[stable] tree: promote changes from testing at cad432926095b9d1a1723bd5188f96930b839dd6

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/809.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -2,9 +2,3 @@
 # similarly to manifest.yaml. Unlike image-base.yaml, which is shared by all
 # streams.
 include: image-base.yaml
-
-# see https://github.com/coreos/coreos-assembler/pull/3607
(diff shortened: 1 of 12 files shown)


jlebon (approved):
```
$ git diff upstream/testing | cat
index d3c24364..9a46b85a 100644
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,5 +1,5 @@
 variables:
-  stream: testing
+  stream: stable
   prod: true

 releasever: 39
```

dustymabe:
The "unexpected change" is `releasever: 39`, which we are doing this round because we are promoting to Fedora 39 🎉

PR #2740, opened 2023-11-22
overrides: fast-track ignition-2.17.0-1.fc39

coreosbot-releng:


jlebon (approved):

dustymabe:
Since `next-devel` and `testing-devel` aren't in sync right now we'd need to do this for `next-devel` too.

dustymabe:
Oh, I see it now in https://github.com/coreos/fedora-coreos-config/pull/2741 - it just didn't merge yet because CI failed.

PR #2742, opened 2023-11-23
Revert "denylist: skip PXE tests on ppc64le"

jlebon:
This reverts commit 077d4188c2dabb49d3d0e7dfe22b3944502ad323.

Those tests should work now with the workaround added in https://github.com/coreos/coreos-assembler/pull/3661.


dustymabe (approved):
LGTM

PR #2743, opened 2023-11-23
[rawhide] overrides: pin `systemd` due to failing networking kola tests

gursewak1997:
Recent rawhide builds have failed with a newer version of systemd. Let's pin on `systemd-254.5-2.fc40` till we get a fix for that.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1620


jlebon:
I think we need to adapt our tests instead. I opened https://github.com/coreos/fedora-coreos-config/pull/2744 for that.

PR #2745, opened 2023-11-28
denylist: skip `coreos.unique.boot.failure` on aarch64

marmijo:
This test has been failing on aarch64 builds. Let's denylist the test while we work to find a fix for:
https://github.com/coreos/coreos-assembler/issues/3669


dustymabe (approved):
LGTM

PR #2746, opened 2023-11-30
s390x: Fix ext.config.disks.partition-scheme

jschintag:
s390x only has 2 partitions. Adjust the test accordingly.

### DIFF
--- a/tests/kola/disks/partition-scheme
+++ b/tests/kola/disks/partition-scheme
@@ -16,23 +16,38 @@ partitionData=$(echo $diskData | jq '.partitions[]')
 totalPartitions=$(echo $diskData | jq '.partitions | length')
 sector_size=$(echo $diskData | jq .sectorsize)
 
-if [[ $totalPartitions -ne 4 ]]; then
-    fatal "Expected 4 partitions, got $totalPartitions"
+arch=$(uname -m)
+
+if [[ "$arch" == "s390x" ]]; then
+    if [[ $totalPartitions -ne 2 ]]; then
+        fatal "Expected 2 partitions, got $totalPartitions"
+    fi
+    # An associative array that maps the partition label to the size (in MiB)
+    # of the partition. For the root partition we set it to "" to skip the check
+    # there because the growfs service runs on first boot. Checking start
+    # should be enough there.
+    declare -A expected=(
+        ["boot"]="384"
+        ["root"]=""
+    )
+else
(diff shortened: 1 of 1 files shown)


jlebon:
Won't we break on the other arches next since they don't have `BIOS-BOOT`? Might as well fix those too while we're here and add the missing partitions (aarch64 has a reserved partition, and ppc64le has the prep boot partition and a reserved partition).

dustymabe:
@jschintag if you want to make this easier for now just set this test to only run on x86_64 and will fix the others later. Sorry for breaking things.

jschintag:
While i don't know about the partition layout on the other arches, i would like to point out that all the other arches passed the test in the pipeline. So i think adjusting it to also support s390x is the best way, instead of just disabling the test.
Tests for aarch64 and ppc64le:
https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/build-arch/3094/
https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/build-arch/3095/

dustymabe:
It turns out that it actually is failing on the other arches, the failure just isn't being detected:

```
Dec 01 15:57:46 qemu0 kola-runext-partition-scheme[2113]: ++ jq 'select ( .name == "BIOS-BOOT") | .size'
Dec 01 15:57:47 qemu0 kola-runext-partition-scheme[2091]: + sectors=
Dec 01 15:57:47 qemu0 kola-runext-partition-scheme[2091]: /usr/local/bin/kola-runext-partition-scheme: line 45: * 512 : syntax error: operand expected (error token is "* 512 ")
Dec 01 15:57:47 qemu0 kola-runext-partition-scheme[2091]: + ok partition scheme
Dec 01 15:57:47 qemu0 kola-runext-partition-scheme[2091]: + echo ok partition scheme
Dec 01 15:57:47 qemu0 kola-runext-partition-scheme[2091]: ok partition scheme
Dec 01 15:57:47 qemu0 kola-runext-partition-scheme[2091]: + exit 0
```

dustymabe:
What's weird is the encounter of the syntax error doesn't appear to stop the entire script from running, just the for loop so it drops down into the `ok()` and then exits 0.

dustymabe:
I'll just merge this and try to fix it up.

dustymabe (approved):

PR #2747, opened 2023-11-30
denylist: extend expired snoozes

marmijo:
The following tests are still failing as of today:
- `coreos.ignition.ssh.key`
- `ext.config.docker.basic`
- `ext.config.kdump.crash`

Extend the snooze on these tests for another two weeks while we wait for a fix on them.


dustymabe (approved):
LGTM

PR #2752, opened 2023-12-04
[testing] overrides: pin podman-4.7.2-1.fc39

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -14,6 +14,16 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1578
       type: pin
+  podman:
+    evr: 5:4.7.2-1.fc39
+    metadata:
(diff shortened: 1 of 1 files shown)


jlebon (approved):
I think we need one for `next` too.

jlebon:
Sorry, I got confused here.

We're cutting a release this week, so we'll repromote from testing-devel as part of that (https://github.com/coreos/fedora-coreos-streams/issues/825).

So this should be opened against testing-devel and next-devel instead.

dustymabe:
> So this should be opened against testing-devel and next-devel instead.

yes. I think what we need to do is try to determine if the reported issues (reported against silverblue) are an issue for FCOS too and then open a tracker issue and pin. I think the issues may have something to do with upgrading systems, which we don't have great CI coverage for. 

Hence our CI passes fine, but that doesn't mean the new podman is without regressions.

dustymabe:
- https://github.com/coreos/fedora-coreos-config/pull/2753
- https://github.com/coreos/fedora-coreos-config/pull/2754

PR #2754, opened 2023-12-04
[next-devel] overrides: pin podman-4.7.2-1.fc39

coreosbot-releng:


jlebon (approved):

dustymabe:
CI failed on a downgrade detection, which is expected.

PR #2757, opened 2023-12-05
overrides: drop zincati pin; add podman fast-track

dustymabe:


cgwalters (approved):

jlebon (approved):
Debating whether we should ship it in next first. OK either way.

PR #2760, opened 2023-12-06
manifests: move `amd-ucode-firmware` to FCOS-specific manifest

jlebon:
This change hasn't hit RHCOS/C9S yet so we can't put it in a shared manifest.

Add a test so we know when RHCOS gets it.

Fixes 898bf23a ("manifests: Explicitly include amd-ucode-firmware").


mike-nguyen (approved):

mike-nguyen:
LGTM

travier:
Thanks for fixing my double mistake! (x86_86 only and FCOS only) 🙂

PR #2761, opened 2023-12-07

coreosbot-releng:


dustymabe:
CI here is warning of a downgrade, which is fine since the next bump-lockfile will fix it. I'm just going to go ahead and merge this.

PR #2762, opened 2023-12-07
[rhcos-4.15] manifests: move `amd-ucode-firmware` to FCOS-specific manifest

mike-nguyen:
This change hasn't hit RHCOS/C9S yet so we can't put it in a shared manifest.

Add a test so we know when RHCOS gets it.

Fixes 898bf23a ("manifests: Explicitly include amd-ucode-firmware").

(cherry picked from commit 76362abe815a6ec6c559745962064e7437030acc)


jlebon:
We'll want https://github.com/coreos/fedora-coreos-config/pull/2763 too.

mike-nguyen:
Added #2763

jlebon (approved):

PR #2763, opened 2023-12-07
tests/kola/files/amd-ucode-firmware: limit to x86_64

jlebon:
This test is only relevant on x86_64.

Fixes 76362abe ("manifests: move `amd-ucode-firmware` to FCOS-specific manifest").


dustymabe (approved):
LGTM

PR #2766, opened 2023-12-13
[rawhide] overrides: remove kexec-tools-2.0.27-1.fc40 pin on ppc64le

marmijo:
The fix for https://github.com/coreos/fedora-coreos-tracker/issues/1598 landed upstream. Let's remove the pin to pick up the latest version of kexec-tools.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1598#issuecomment-1850504070


dustymabe (approved):
LGTM

PR #2769, opened 2023-12-15
overlay.d: add 40grub overlay

dustymabe:
bootupd upstream was modified to store default static grub configs and allow for OS integrators to add their own snippets by placing files in /usr/lib/bootupd/grub2-static/configs.d/ [1] When we move to building with osbuild we will start taking advantage of this functionality and calling bootupd with --with-static-configs. Let's go ahead and add our snippets here so they can be leveraged when we start calling bootupd with --with-static-configs.

The snippets added here come from the parts of the config currently defined in coreos-assembler that weren't lifted into bootupd [2][3][4].

[1] https://github.com/coreos/bootupd/pull/543
[2] https://github.com/coreos/coreos-assembler/blob/4636b1a5c6dc00b1d6a58b1bfbb199431444336b/src/grub.cfg#L58-L60
[3] https://github.com/coreos/coreos-assembler/blob/4636b1a5c6dc00b1d6a58b1bfbb199431444336b/src/grub.cfg#L71-L85
[4] https://github.com/coreos/coreos-assembler/blob/4636b1a5c6dc00b1d6a58b1bfbb199431444336b/src/grub.cfg#L87-L91

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -18,6 +18,7 @@ ostree-layers:
   - overlay/20platform-chrony
   - overlay/25azure-udev-rules
   - overlay/30lvmdevices
+  - overlay/40grub
 
 # Be minimal
 recommends: false
--- a/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/30_console.cfg
+++ b/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/30_console.cfg
@@ -1,0 +1,4 @@
+
+# Any non-default console settings will be inserted here.
+# CONSOLE-SETTINGS-START
+# CONSOLE-SETTINGS-END
--- a/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/40_coreos-ignition.cfg
+++ b/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/40_coreos-ignition.cfg
@@ -1,0 +1,18 @@
+# Remove soon when Ignition is providing this:
+# https://github.com/coreos/fedora-coreos-config/pull/2769#discussion_r1428152480
+#
+# Determine if this is a first boot and set the ${ignition_firstboot} variable
+# which is used in the kernel command line.
+set ignition_firstboot=""
+if [ -f "/ignition.firstboot" ]; then
+    # Default networking parameters to be used with ignition.
+    set ignition_network_kcmdline=''
+
+    # Source in the `ignition.firstboot` file which could override the
+    # above $ignition_network_kcmdline with static networking config.
+    # This override feature is also by coreos-installer to persist static
+    # networking config provided during install to the first boot of the machine.
+    source "/ignition.firstboot"
+
+    set ignition_firstboot="ignition.firstboot ${ignition_network_kcmdline}"
+fi
--- a/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/70_coreos-user.cfg
+++ b/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/70_coreos-user.cfg
@@ -1,0 +1,6 @@
+
+# Import user defined configuration
+# tracker: https://github.com/coreos/fedora-coreos-tracker/issues/805
+if [ -f $prefix/user.cfg ]; then
+  source $prefix/user.cfg
+fi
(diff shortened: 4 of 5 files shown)


overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/50_coreos.cfg
```diff
@@ -0,0 +1,26 @@
+
+# Any non-default console settings will be inserted here.
+# CONSOLE-SETTINGS-START
+# CONSOLE-SETTINGS-END
+
+# Determine if this is a first boot and set the ${ignition_firstboot} variable
+# which is used in the kernel command line.
+set ignition_firstboot=""
```
cgwalters: See https://github.com/coreos/ignition/commit/f878c9ddbdb8bade9822b8b578f6641c054a9bc9

(Remember because IoT/Edge exist and currently use osbuild this drives us to ship stuff like this in packages)

I think what hasn't happened yet is to ship that content in an `ignition-grub` subpackage.

cgwalters: Tangentially related to this...we are currently leaving the functionality of `coreos-installer --console` on the floor in other paths (Anaconda, `bootc install`).

Maaybe we should try to move this to `bootupd`, like `bootupd sync-bootloader-kargs` that uses the kernel cmdline as source of truth and syncs it to `/boot/grub2/console.cfg` or something.

dustymabe: > See [coreos/ignition@f878c9d](https://github.com/coreos/ignition/commit/f878c9ddbdb8bade9822b8b578f6641c054a9bc9)
> 
> (Remember because IoT/Edge exist and currently use osbuild this drives us to ship stuff like this in packages)
> 
> I think what hasn't happened yet is to ship that content in an `ignition-grub` subpackage.

oh nice - I didn't know that. Yeah, we'd need an ignition-grub subpackage included, but maybe since the path is `/usr/lib/bootupd/grub2-static/configs.d/` maybe `ignition-bootupd-grub-config` subpackage?

We can easily switch out this to use that when it has shipped.

dustymabe: > Tangentially related to this...we are currently leaving the functionality of `coreos-installer --console` on the floor in other paths (Anaconda, `bootc install`).
> 
> Maaybe we should try to move this to `bootupd`, like `bootupd sync-bootloader-kargs` that uses the kernel cmdline as source of truth and syncs it to `/boot/grub2/console.cfg` or something.

I'm not really sure what the best path forward is here but maybe we can at least get the file right so I can update [coreos-installer](https://github.com/coreos/coreos-installer/pull/1360#pullrequestreview-1784456023) once and not twice?

Would `30_console.cfg` work?

cgwalters: > Would 30_console.cfg work?

Sure.

> I'm not really sure what the best path forward is here but maybe we can at least get the file right so I can update https://github.com/coreos/coreos-installer/pull/1360#pullrequestreview-1784456023 once and not twice?

If we go the bootupd path, the file wouldn't be written by coreos-installer.

dustymabe: > > Would 30_console.cfg work?
> 
> Sure.

ok cool - updated here. I split our configs out into 3 different ones now. I'll work on testing this.

> 
> > I'm not really sure what the best path forward is here but maybe we can at least get the file right so I can update [coreos/coreos-installer#1360 (review)](https://github.com/coreos/coreos-installer/pull/1360#pullrequestreview-1784456023) once and not twice?
> 
> If we go the bootupd path, the file wouldn't be written by coreos-installer.

I think that would make bootupd a requirement of coreos-installer, which I think we'd have to think through the implications of.

cgwalters: Yeah, would probably need to vendor as a crate I guess.

dustymabe:
local tests on this are looking good.

keep in mind these files currently won't be used in FCOS/RHCOS until we either change cause to call bootupd with `--with-static-config` or we switch to using `osbuild` and set with static configs to true.

This is mostly enablement, but doesn't make any concrete changes for now.

cgwalters (approved):

PR #2770, opened 2023-12-15
[Rawhide] overrides: pin `selinux-policy-40.5-1.fc40` and `container-selinux-2.224.0-1.fc40`

gursewak1997:
Pin selinux-policy-40.5-1.fc40 and container-selinux-2.224.0-1.fc40
Ref Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1630

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,19 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  container-selinux:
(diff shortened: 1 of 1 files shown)


dustymabe:
maybe we can shorten the commit title to:

```
overrides: pin selinux-policy and container-selinux
```

and then add a body that has a sentence about the problem and links to the tracker issue.

?

gursewak1997:
Need to override the CI.
`Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.`

dustymabe:
can you add your investigation of the issue into the tracker issue (with any relevant log messages) and then maybe open a BZ against selinux policy so we can get it fixed?

PR #2774, opened 2023-12-19
[testing] overrides: pin kernel-6.6.3-200.fc39

coreosbot-releng:


dustymabe:
turns out this won't solve any problems: https://github.com/coreos/fedora-coreos-tracker/issues/1637#issuecomment-1863123978

closing

PR #2780, opened 2024-01-02
 Add bootupd package for ppc64le

ravanelli:
- Bootupd is now available for ppc64le as part of our ongoing efforts to
integrate osbuild. Let's include it

- Let's also enable the existing test to check it.

### DIFF
--- a/manifest-lock.ppc64le.json
+++ b/manifest-lock.ppc64le.json
@@ -174,6 +174,12 @@
         "sourcerpm": "bind"
       }
     },
+    "bootupd": {
+      "evra": "0.2.16-2.fc39.ppc64le",
+      "metadata": {
+        "sourcerpm": "rust-bootupd"
+      }
+    },
     "brcmfmac-firmware": {
       "evra": "20231211-1.fc39.noarch",
       "metadata": {
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -205,8 +205,9 @@ packages-s390x:
 
 # See https://github.com/coreos/bootupd
 arch-include:
-  x86_64: bootupd.yaml
   aarch64: bootupd.yaml
+  ppc64le: bootupd.yaml
+  x86_64: bootupd.yaml
 
 remove-from-packages:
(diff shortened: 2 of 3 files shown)


cgwalters (approved):

dustymabe:
> * Bootupd is now available for ppc64le as part of our ongoing efforts to
>       integrate osbuild. Let's include it

but notably our current [disk building code](https://github.com/coreos/coreos-assembler/blob/5a81fdf106ce9c4da08d688f3b5baf3c9cf11505/src/create_disk.sh#L408) that calls `bootupctl` won't be called for `ppc64le`. This should change when we switch to OSBuild.

Given that ^^ will the test even pass?

ravanelli:
> > * Bootupd is now available for ppc64le as part of our ongoing efforts to
> >   integrate osbuild. Let's include it
> 
> but notably our current [disk building code](https://github.com/coreos/coreos-assembler/blob/5a81fdf106ce9c4da08d688f3b5baf3c9cf11505/src/create_disk.sh#L408) that calls `bootupctl` won't be called for `ppc64le`. This should change when we switch to OSBuild.
> 
> Given that ^^ will the test even pass?

Do you mean the kola bootupd test I'm changing in this PR?
If so, it is enabled [here](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset#L26C3-L27C22).

dustymabe:
> Do you mean the kola bootupd test I'm changing in this PR?

yes

> If so, it is enabled [here](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset#L26C3-L27C22).

what I mean is that if the disk image that was built didn't use bootupctl when it was created then will the test pass?

ravanelli:
> > Do you mean the kola bootupd test I'm changing in this PR?
> 
> yes
> 
> > If so, it is enabled [here](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset#L26C3-L27C22).
> 
> what I mean is that if the disk image that was built didn't use bootupctl when it was created then will the test pass?

I just tested it, and yes, it will pass.

dustymabe (approved):

PR #2783, opened 2024-01-03
[rawhide] Enable composefs

cgwalters:
The goal is to start shaking out integration issues.

### DIFF
--- a/overlay.d/08ostree-prepare-root/usr/lib/ostree/prepare-root.conf
+++ b/overlay.d/08ostree-prepare-root/usr/lib/ostree/prepare-root.conf
@@ -1,0 +1,3 @@
+[composefs]
+enabled = true
+
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -9,6 +9,9 @@ repos:
 
 include: manifests/fedora-coreos.yaml
 
+ostree-layers:
+  - overlay/08ostree-prepare-root
+
 # Ship this in rawhide because we want to enable it in the future, also
 # to shake out any integration issues.
 packages:


dustymabe:
We'll be switching to OSBuild for image builds soon and I'm not sure of its capability with respect to composefs. There is https://github.com/osbuild/osbuild/pull/1343 but it seems experimental.

Any changes we make at this point that rely on currently unused (in production streams) functionality implemented in create_disk.sh should be weighed heavily with the capabilities of OSBuild.

cgwalters:
It's only experimental because the ostree state is still experimental, which is experimental still because we need to flesh things out and fix more bugs, which will happen as we try to enable it in distribution branches like rawhide...

So keeping it out here because of that is just deadlocking ourselves :smile: 

You're absolutely right that when we switch to osbuild in cosa and start ignoring the `image.yaml` field here it's possible this change would be silently lost.  But that's not fatal, we can fix it.

jlebon:
Higher level, even after we switch to OSBuild, we should still retain the ability to ratchet features in via `image.yaml`. So in this case for example, we'd... add a new var under [`mpp-vars`](https://github.com/coreos/coreos-assembler/blob/5a81fdf106ce9c4da08d688f3b5baf3c9cf11505/src/coreos.osbuild.mpp.yaml#L2) I guess? But we need to teach cosa to dynamically change those (like the existing `XXX` there).

All `composefs: true` does is set the `ex-integrity.composefs` knob in the repo config. It looks like the `org.osbuild.experimental.ostree.config` stage added in https://github.com/osbuild/osbuild/pull/1343 does exactly that. So we could just call out to that stage in the mpp and pass through the new var.

jlebon (dismissed):
Cool getting this in for now.

cgwalters:
There's a few test failures to fix, will work on it

dustymabe:
> You're absolutely right that when we switch to osbuild in cosa and start ignoring the `image.yaml` field here it's possible this change would be silently lost. But that's not fatal, we can fix it.

+1 - as long as it is understood that this might silently drop I'm ok with making this change.

cgwalters:
This is now reworked on top of https://github.com/ostreedev/ostree/pull/3165/commits/cae4ceb6c5f8d3e1db52364a3ca6cd9ccb4e0a69 which means we can configure it in the commit/container, not at disk image generation time.

overlay.d/08ostree-prepare-root/usr/lib/ostree/prepare-root.conf
dustymabe, line 1: this file will get overwritten by config-bot. It will need to go in the `testing-devel` branch. A different approach may be to put this all in `testing-devel` but then do a `conditional-include` in one of the other manifests that only includes it based on the stream name.

overlay.d/08ostree-prepare-root/usr/lib/ostree/prepare-root.conf
```diff
@@ -0,0 +1,3 @@
+[composefs]
+enabled = true
+
```
dustymabe, line 3: Let's also update the `overlay.d/README.md` file with some context/rationale.

overlay.d/08ostree-prepare-root/usr/lib/ostree/prepare-root.conf
cgwalters: Oh.  OK, I tried this in https://github.com/coreos/fedora-coreos-config/pull/2856

cgwalters:
moving to https://github.com/coreos/fedora-coreos-config/pull/2856

PR #2785, opened 2024-01-03
platforms: [qemu] Also use hvc0 for x86_64/aarch64

cgwalters:
Pairs with https://github.com/coreos/coreos-assembler/pull/3689

Basically hvc0 is faster and more modern, but requires us to opt-in specifically.  We already had a few usages of hvc0 from cloud providers that default to this, no reason for us not to do it for our own qemu provider.

(One thing TODO here: I believe s390x qemu could do this too;
 as of right now we don't have any references to s390x here)

### DIFF
--- a/platforms.yaml
+++ b/platforms.yaml
@@ -52,6 +52,9 @@ aarch64:
     kernel_arguments:
       - console=ttyAMA0,115200
   qemu:
+    kernel_arguments:
+      - console=hvc0
+      - console=ttyAMA0,115200
     # The kernel successfully autodetects a serial console, but we still
     # want GRUB to use one
     grub_commands:
@@ -153,6 +156,7 @@ x86_64:
       - terminal_input serial console
       - terminal_output serial console
     kernel_arguments:
+      - console=hvc0
       - console=tty0
       - console=ttyS0,115200n8
   virtualbox:


cgwalters:
Draft since this needs some more testing...

cgwalters:
Ahh yes, Fedora (and c9s) currently uses
`CONFIG_VIRTIO_CONSOLE=m`
which I think is why I'm seeing the behavior that we only end up with console logging only after the initramfs has started if the hypervisor is only setting up a virtio console (as is done in https://github.com/coreos/coreos-assembler/pull/3689 ).

In contrast...a quick check of opensuse Leap (`kernel-kvmsmall-5.14.21-150500.55.39.1.x86_64`) shows
```
$ grep VIRTIO_CONSOLE /boot/config-5.14.21-150500.55.39-kvmsmall 
CONFIG_VIRTIO_CONSOLE=y
```

Though Debian (`linux-image-amd64 6.1.69-1`) has it as a module too.

I'm pretty strongly of the opinion though that we should change this to be built-in; we already have `CONFIG_VIRTIO=y` (and some other virtio modules) built in.

baude:
I agree it should be built-in.  My old friend `hvc` :)

dustymabe:
so we just need to submit a PR to the kernel configs? or is it more complicated than that?

platforms.yaml
```diff
@@ -52,6 +52,9 @@ aarch64:
     kernel_arguments:
       - console=ttyAMA0,115200
   qemu:
+    kernel_arguments:
+      - console=hvc0
+      - console=ttyAMA0,115200
```
dustymabe, line 57: I'm not sure if this will change the behavior on aarch64 qemu today. Right now it autodetects. So will this change it so that VGA console isn't the primary console?

cgwalters: I believe so, yes.  In our use cases we basically never use a graphical console (right?)

That said, honestly since podman is targeting applehv now as of recently it matters much less what we do with qemu.  So I'm just going to close this for now.

It probably *does* make sense for us to switch to hvc for qemu, but doesn't seem worth the churn for now.

PR #2788, opened 2024-01-05
manifests: update qcom dtb hack on aarch64

dustymabe:
We need this hack again to work around a new corner case in the /boot ENOSPC wars.

See https://github.com/coreos/fedora-coreos-tracker/issues/1637


jmarrero (approved):
/lgtm

PR #2789, opened 2024-01-05
[rhcos-4.14] .cci.jenkinsfile: disable testing on older RHCOS branch

dustymabe:
The tests won't pass because they run with latest COSA and not the branch specific COSA. Let's just disable them for now.

(cherry picked from commit 5be9e141a7fb90a9b41797f00064269955744a33)


jmarrero (approved):
/lgtm

PR #2791, opened 2024-01-05
overrides: fast-track ostree-2023.8-3.fc38

dustymabe:
Contains the fix for the corner case issue described in https://github.com/coreos/fedora-coreos-tracker/issues/1637


prestist (approved):
LGTM

PR #2794, opened 2024-01-05
[stable] The test for iscsi boot fails on somes arches

gursewak1997:
let's disable while we build images for all the arches and create a repo centralizing images that kola needs for the tests.


dustymabe (approved):
LGTM

PR #2796, opened 2024-01-08
denylist: snooze `rpm-ostree.kernel-replace`

gursewak1997:
Recent rawhide runs fails in `ext.config.rpm-ostree.kernel-replace` test because the `ostree container encapsulate command` just seems to hang forever. Snoozing till we fix it

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -26,3 +26,8 @@
 - pattern: iso-offline-install-iscsi.bios
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1638
   snooze: 2024-01-20
+- pattern: ext.config.rpm-ostree.kernel-replace
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1642
+  snooze: 2024-01-22
+  streams:
+    - rawhide


kola-denylist.yaml
```diff
@@ -26,3 +26,6 @@
 - pattern: iso-offline-install-iscsi.bios
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1638
   snooze: 2024-01-20
+- pattern: ext.config.rpm-ostree.kernel-replace
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1642
+  snooze: 2024-01-22
```
dustymabe, line 31: let's limit this snooze to the `rawhide` stream since that's the only place it is known to fail for now and we'd want to know if the problem creeps into any other stream.

dustymabe (changes requested):
https://github.com/coreos/fedora-coreos-config/pull/2796#discussion_r1445050229

dustymabe:
the update here LGTM but let's hold merging this until the currently running bump-lockfile job finishes (any git pushes will force that job to run again ☹️)

dustymabe (approved):

PR #2797, opened 2024-01-12
40ignition-ostree: avoid jq try-catch

jlebon:
There seems to be a bug in jq 1.7 where `try .foo catch "" | ascii_downcase` now fails.

It seems like jq implements short-circuiting boolean operators, so we can rework this to instead just use a simpler `and` operator.

BZ issue: https://bugzilla.redhat.com/show_bug.cgi?id=2258285

Reported-by: Gursewak Mangat <gursmangat@gmail.com>

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
@@ -36,7 +36,7 @@ query_fslabel() {
 
 # Print jq query string for partitions with type GUID $1
 query_parttype() {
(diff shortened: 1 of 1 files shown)


jlebon:
@gursewak1997 If you open an issue against jq, I can crosslink it here.

dustymabe (approved):

gursewak1997 (approved):

dustymabe:
disabled auto-merge since there is a bump-lockfile job running right now. If this merges it will cause all of that work to be thrown away. Will merge after that finishes.

gursewak1997:
Just a quick update: This was rather a issue in how  in `jq-1.6` try catches the error of ascii_downcase but it should not.
For more details refer to: https://github.com/jqlang/jq/issues/3016
Also closed the https://bugzilla.redhat.com/show_bug.cgi?id=2258285

PR #2804, opened 2024-01-15
denylist: update outdated entries

marmijo:
The issues with `ext.config.docker.basic`  and `coreos.unique.boot.failure` were both resolved and these tests are passing again. Remove the denylist entries for these two tests.

Bump the snooze for `coreos.ignition.ssh.key` while we continue to investigate the issue.

See: 
 - https://github.com/coreos/fedora-coreos-tracker/issues/1578
 -  https://github.com/coreos/coreos-assembler/issues/3669
 - https://github.com/coreos/fedora-coreos-tracker/issues/1553


dustymabe (approved):
LGTM

PR #2806, opened 2024-01-16
denylist: snooze `ext.config.root-reprovision.linear`

gursewak1997:
Most recent rawhide build fails `ext.config.root-reprovision.linear' kola test with the latest kernel release. Snoozing till it gets fixed.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1648


marmijo (approved):
LGTM

dustymabe (approved):

PR #2809, opened 2024-01-18
tests/bootupd: skip test on ppc64le and s390x if bootupd not installed

jlebon:
This test runs on both the 9.2 and 9.4 variants of RHCOS, but we only have bootupd on ppc64le and s390x on the latter. On those arches, make the test conditional on whether bootupd is installed.


dustymabe (approved):
LGTM

PR #2810, opened 2024-01-18
tests/grub2-install: update to match bootupd grub2-install invocation

dustymabe:
The bootupd invocation of grub2-install [1] specifies both mdraid1x and part_gpt modules. Let's update it here so the tests pass. 

[1] https://github.com/coreos/bootupd/blob/e1be3005d827fc3bacf82a2c2a98cd9c0706c65f/src/bios.rs#L66-L69

### DIFF
--- a/tests/kola/boot/grub2-install
+++ b/tests/kola/boot/grub2-install
@@ -18,23 +18,36 @@ set -xeuo pipefail
 
 boot_dev=/boot
 arch=$(arch)
+grub_install_args=""
+grub_modules=()
 case ${arch} in
   x86_64)
     target=i386-pc
     core="${boot_dev}/grub2/${target}/core.img"
     partition=$(findmnt -no SOURCE ${boot_dev})
     device=$(lsblk --paths -no PKNAME ${partition})
-    # sync grub2-install parameter according to image build script
-    # https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh
-    grub_install_args="--modules mdraid1x"
+    #
+    # Handle modules for both the create_disk.sh case that only
+    # includes "mdraid1x" and bootupd case (used when building using
+    # OSBuild) that adds "part_gpt". The aleph stage in osbuild is
+    # the only one that creates a file at /sysroot/.aleph-version.json
+    #
+    # sync grub2-install parameters with the bootupd invocation from:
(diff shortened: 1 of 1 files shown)


dustymabe:
@ravanelli - can you test this to make sure the grub2-install test still passes when run against an osbuild built qemu image?

tests/kola/boot/grub2-install
```diff
@@ -18,15 +18,17 @@ set -xeuo pipefail
...
   x86_64)
     target=i386-pc
     core="${boot_dev}/grub2/${target}/core.img"
     partition=$(findmnt -no SOURCE ${boot_dev})
     device=$(lsblk --paths -no PKNAME ${partition})
-    # sync grub2-install parameter according to image build script
-    # https://github.com/coreos/coreos-assembler/blob/main/src/create_disk.sh
-    grub_install_args="--modules mdraid1x"
+    # sync grub2-install parameters with the bootupd invocation from:
+    # https://github.com/coreos/bootupd/blob/main/src/bios.rs
```
jlebon, line 36: Shouldn't we still do at least this comment change for the ppc64le case too?

dustymabe: Yes. I'll update it. It was an oversight that I didn't in the first upload.

jlebon (dismissed):

jlebon:
One way to ratchet this in is in the test you check if it's osbuild-made (checking the aleph), and if so, add the module to the invocation. Then we don't really have to fix `create_disk.sh` (is that module actually fixing anything?), it'll just be fixed when we move over to osbuild.

Semi-OK with forcing it in, but also we've been breaking RHCOS intermittently a lot lately with all the churn.

ravanelli:
> @ravanelli - can you test this to make sure the grub2-install test still passes when run against an osbuild built qemu image?

ppc64le passed, it doesn't run for aarch64 and s390x.
However, it was already passing for ppc64le. Do you want me to give it a try for x86?

dustymabe:
@jlebon updated the strategy per your suggestion..

dustymabe:
> ppc64le passed, it doesn't run for aarch64 and s390x.
> However, it was already passing for ppc64le. Do you want me to give it a try for x86?

Nope. Thanks! I think that's all I needed.

jlebon (approved):
LGTM!

PR #2811, opened 2024-01-18
kola/lvmdevices: use serial to reference disks

### LINKED ISSUE #2807: s390x: ext.config.disks.lvmdevices fails ocassionally
```
22:33:11  Jan 16 21:33:04 qemu0 kola-runext-lvmdevices[2151]: + pvcreate --devicesfile= /dev/vda /dev/vdb
22:33:11  Jan 16 21:33:04 qemu0 kola-runext-lvmdevices[2153]:   Cannot use /dev/vdb: device is partitioned
22:33:11  Jan 16 21:33:06 qemu0 kola-runext-lvmdevices[2153]:   Physical volume "/dev/vda" successfully created.
```
Probable cause is that on s390x `/dev/vda` and such are not always guaranteed to be the same disk. This most likely resulted in the above example, where `/dev/vdb` is likely the root partition.

jlebon:
Using `/dev/vda`/`/dev/vdb` to refer to attached disks in inherently unreliable since the order may change across e.g. OS or QEMU versions.

Slap on a serial name instead and refer to it that way. Interestingly, it seems like the LVM tooling also picks up on this and bases its entry in the `lvmdevices` file on the device serial.

Closes: #2807

### DIFF
--- a/tests/kola/disks/lvmdevices
+++ b/tests/kola/disks/lvmdevices
@@ -5,7 +5,7 @@
 ##   # additionalDisks is only supported on qemu.
 ##   platforms: qemu
 ##   # A few extra disks to set up LVM on.
-##   additionalDisks: ["1G", "2G"]
+##   additionalDisks: ["1G:serial=disk1", "2G:serial=disk2"]
 
 set -xeuo pipefail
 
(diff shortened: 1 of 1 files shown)


tests/kola/disks/lvmdevices
```diff
@@ -46,8 +46,9 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
 
   rebooted)
       # Check that the devices are in the devices file.
-      grep -q 'DEVNAME=/dev/vda' "$LVMDEVICESFILE" || fatal "Missing vda in devices file"
-      grep -q 'DEVNAME=/dev/vdb' "$LVMDEVICESFILE" || fatal "Missing vdb in devices file"
+      cat "$LVMDEVICESFILE"
```
dustymabe: looks useful for dev/test.. intentionally left in the final PR?

dustymabe (dismissed):
LGTM

dustymabe (approved):

marmijo (approved):

PR #2814, opened 2024-01-20
denylist: extend snooze for iso-offline-install-iscsi.bios

dustymabe:


jlebon (dismissed):
I would just denylist it (unsnoozed) for now. (Did that for RHCOS in https://github.com/openshift/os/pull/1426).

dustymabe:
> I would just denylist it (unsnoozed) for now. (Did that for RHCOS in [openshift/os#1426](https://github.com/openshift/os/pull/1426)).

OK I updated the commit to just drop the snooze.

jlebon (approved):

marmijo (approved):

PR #2817, opened 2024-01-23
denylist: extend expired snoozes

marmijo:
`ext.config.rpm-ostree.kernel-replace` and `ext.config.root-reprovision.linear` are still failing.
Let's extend the snooze while we continue to investigate.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -15,12 +15,13 @@
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1638
 - pattern: ext.config.rpm-ostree.kernel-replace
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1642
-  snooze: 2024-01-22
+  snooze: 2024-02-05
+  warn: true
   streams:
     - rawhide
 - pattern: ext.config.root-reprovision.linear
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1648
-  snooze: 2024-01-22
+  snooze: 2024-02-05
   warn: true  
   streams:
     - rawhide


kola-denylist.yaml
```diff
@@ -15,12 +15,12 @@
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1638
 - pattern: ext.config.rpm-ostree.kernel-replace
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1642
-  snooze: 2024-01-22
+  snooze: 2024-02-05
   streams:
     - rawhide
```
dustymabe: did we ask someone from the corepackages team to look at this?

kola-denylist.yaml
```diff
@@ -15,12 +15,12 @@
...
 - pattern: ext.config.root-reprovision.linear
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1648
-  snooze: 2024-01-22
+  snooze: 2024-02-05
```
dustymabe, line 24: I feel like someone needs to start looking at this before the next two weeks have expired. Initially we suspected it may be part of initial RC kernel problems but now we are out of rc0 and on to rc1 and it still fails so we probably need to be proactive before the kernel gets released and then comes to our stable streams.

kola-denylist.yaml
dustymabe: note: adding `warn: true` here will at least make this non-fatal if he snooze expires again.

kola-denylist.yaml
marmijo: Actually, it looks like I jumped the gun a bit here. The rawhide build from this morning was a rerun of yesterday's build and only the multi-arch jobs were run. I extended the snooze here thinking the test had failed, but did not verify.
I started a new rawhide build to see if this issue has been resolved. For now, I'll remove the entry from this PR until we see the result.

kola-denylist.yaml
marmijo: I agree we should raise awareness of this issue. No, I haven't asked anyone to look into it yet.
I'll add the `warn: true`.
Because the rawhide job this morning was a rebuild of yesterday, let's hold off to see the result of today's build

kola-denylist.yaml
marmijo: rawhide is still pulling in rc0. The rc1 kernel update had some negative karma and is still in testing.
https://bodhi.fedoraproject.org/updates/FEDORA-2024-2cd93c3116
`ext.config.root-reprovision.linear` failed during the rawhide build with the rc0 kernel so I'll extend the snooze and hopefully the test will pass when we get to rc1

dustymabe: > so I'll extend the snooze and hopefully the test will pass when we get to rc1

+1

dustymabe (approved):

PR #2818, opened 2024-01-24
Add Podman v5 breaking changes checks.

gursewak1997:
Podman 5 will come with breaking changes affecting upgradability. CGroups v1 environments will be required to switch to CGroups v2 and CNI plugin environemnts will need to switch to netavark. In the above script we added the CLHM helpers to notifiy people of the incoming changes.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1629

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -33,6 +33,9 @@ conditional-include:
     # Remove this once we have Podman v5 in f39.
     # xref: https://github.com/coreos/fedora-coreos-tracker/issues/1629
     include: cni-plugins.yaml
+  - if: releasever == 39
+    # Checks for breaking changes that came with Podman v5.
+    include: podman-v5.yaml
 
 ostree-layers:
   - overlay/15fcos
--- a/overlay.d/14podman-v5/usr/lib/systemd/system/coreos-cni-networking-check.service
+++ b/overlay.d/14podman-v5/usr/lib/systemd/system/coreos-cni-networking-check.service
@@ -1,0 +1,11 @@
+# This service checks if the system nodes are still using
+# CNI networking. If so, they will be warned to move
+# their nodes to netavark respectively.
+[Unit]
+Description=Check If Podman Is Still Using CNI Networking
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-cni-networking-check
+RemainAfterExit=yes
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/14podman-v5/usr/libexec/coreos-cni-networking-check
+++ b/overlay.d/14podman-v5/usr/libexec/coreos-cni-networking-check
@@ -1,0 +1,37 @@
+#!/usr/bin/bash
+set -euo pipefail
+# Podman is dropping support for CNI networking.
+# Podman 5 changes: https://fedoraproject.org/wiki/Changes/Podman5
+# This script checks if the system nodes are still using CNI
+# networking. If so, they will warned to move their nodes to Netavark.
+
+# Change the output color to yellow
+warn=$(echo -e '\033[0;33m')
+# No color
+nc=$(echo -e '\033[0m')
+
+# Podman supports two network backends Netavark and CNI.
+# Netavark is the default network backend and was added in
+# Podman version 4.0. CNI is deprecated and is removed
+# in Podman version 5.0, in preference of Netavark.
+podmanBackend=$(podman info --format "{{.Host.NetworkBackend}}")
+
+if [[ $podmanBackend != "netavark" ]]; then
+    motd_path=/run/motd.d/35_cni_warning.motd
+
+    cat << EOF > "${motd_path}"
+${warn}
+##############################################################################
+WARNING: Podman is using CNI networking. CNI is deprecated and will be
+removed in the upcoming Podman v5.0, in preference of Netavark. To switch
+from CNI networking to Netavark, you must run 'podman system reset --force'.
+This will delete all of your images, containers, and custom networks.
+Depending on your setup it may be preferable to reprovision the whole machine
+from the latest images.
+
+To disable this warning, use:
+sudo systemctl disable coreos-cni-networking-check.service
+##############################################################################
+${nc}
+EOF
+fi
--- a/overlay.d/15fcos/usr/libexec/coreos-check-cgroups-version
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-cgroups-version
@@ -11,15 +11,16 @@ motd_path=/run/motd.d/30_cgroupsv1_warning.motd
 
 cat << EOF > "${motd_path}"
 ${warn}
-############################################################################
+##########################################################################
 WARNING: This system is using cgroups v1. For increased reliability
 it is strongly recommended to migrate this system and your workloads
-to use cgroups v2. For instructions on how to adjust kernel arguments
+to use cgroups v2. A future version of Podman will also drop support
+for cgroups v1. For instructions on how to adjust kernel arguments
 to use cgroups v2, see:
 https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/
 
 To disable this warning, use:
-sudo systemctl disable coreos-check-cgroups.service
-############################################################################
+sudo systemctl disable coreos-check-cgroups-version.service
+##########################################################################
 ${nc}
 EOF
--- a/manifests/podman-v5.yaml
+++ b/manifests/podman-v5.yaml
@@ -1,0 +1,5 @@
+# https://github.com/coreos/fedora-coreos-tracker/issues/1629
+# Delete this file once we are on Podman v5 everywhere.
+# i.e. We have moved past F39.
+ostree-layers:
+  - overlay/14podman-v5
--- a/overlay.d/14podman-v5/usr/lib/systemd/system-preset/14-fcos.preset
+++ b/overlay.d/14podman-v5/usr/lib/systemd/system-preset/14-fcos.preset
@@ -1,0 +1,1 @@
+enable coreos-cni-networking-check.service
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -1,5 +1,5 @@
 enable coreos-check-ssh-keys.service
 # Check if cgroupsv1 is still being used
-enable coreos-check-cgroups.service
+enable coreos-check-cgroups-version.service
 # https://fedoraproject.org/wiki/Changes/EnableFwupdRefreshByDefault
 enable fwupd-refresh.timer
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-cgroups-version.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-cgroups-version.service
@@ -5,7 +5,7 @@ Description=Check if cgroupsv1 Is Still Being Used
 ConditionControlGroupController=v1
 [Service]
 Type=oneshot
-ExecStart=/usr/libexec/coreos-check-cgroups
+ExecStart=/usr/libexec/coreos-check-cgroups-version
 RemainAfterExit=yes
 [Install]
 WantedBy=multi-user.target
--- a/overlay.d/README.md
+++ b/overlay.d/README.md
@@ -24,6 +24,14 @@ https://bugzilla.redhat.com/show_bug.cgi?id=1700056
 
 Warning about `/etc/sysconfig`.
 
+14podman-v5
+------
+
+Warn about breaking changes [1] with Podman v5 and link the
+users to how they can adapt their existing nodes.
+
+[1] https://github.com/coreos/fedora-coreos-tracker/issues/1629
+
 15fcos
 ------
 


jlebon:
Thanks, I think this is a good start. Now, before going further and refining this we need to do two things:
1. Try to reproduce the situation that triggers this. E.g. start an old enough FCOS then update all the way to e.g. rawhide with podman v5 (it doesn't look like it's in rawhide yet, but you can use the packages from https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/ for now) and confirm the issues.
2. Find and test the steps to take to remediate the issues. E.g. for cgroups, confirm that just dropping `systemd.unified_cgroup_hierarchy=0` and rebooting works (i.e. existing containers still work fine). For the CNI networking case, I think there's no choice but to nuke all the containers, but that needs investigation.

Once we have that info, let's write up a nice comment in e.g. https://github.com/coreos/fedora-coreos-tracker/issues/1629 that we can then link to from these issue dropins.

jlebon:
Also, note we'll need a systemd service to actually run the script.

gursewak1997:
I see that we have a script [coreos-check-cgroups](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/15fcos/usr/libexec/coreos-check-cgroups) that already checks and warns the users who use cgroupv1.
Do we keep and add networking check to it or delete that.

overlay.d/14container/usr/libexec/coreos-podman-changes-check
```diff
@@ -0,0 +1,43 @@
...
+cgroups v1. Move your nodes to cgroups v2 if not already. For instructions
+on how to adjust kernel arguments to use cgroups v2, see:
+https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/${nc}"
+fi
+
+# Podman supports two network backends Netavark and CNI.
+# Netavark is the default network backend and was added in
+# Podman version 4.0. CNI is deprecated and is removed
+# in Podman version 5.0, in preference of Netavark.
+podmanBackend=$(podman info --format "{{.Host.NetworkBackend}}")
```
dustymabe: I think we need to somehow not run this check if the users are using docker and not podman. Just running the `podman` or `docker` commands on systems that are using the other runtime can cause issues.

@jlebon - do you agree?

dustymabe:
> I see that we have a script [coreos-check-cgroups](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/15fcos/usr/libexec/coreos-check-cgroups) that already checks and warns the users who use cgroupv1.
> Do we keep and add networking check to it or delete that.

I think we should keep the checks separate (i.e. a new check for netavark) BUT we should somehow rename the cgroups check so that if they previously ignored the warning by running `systemctl disable coreos-check-cgroups.service` then the warning will re-appear.

We should also update the text of the cgroup check to reflect the current state.

overlay.d/14container/usr/libexec/coreos-podman-changes-check
jlebon: Hmm, I'm not sure but I would think `podman info` at least would be safe regardless. The larger point of only warning if podman is used makes sense, though I'm not sure there's a reliable way to know which runtime is in use. E.g. just checking for running containers and not finding any at that moment in time doesn't necessarily mean it's not in use.

If `podman info` is safe (/cc @mheon), then implicitly it's checking for the "in use" bit as well since it'll only return something other than the netavark default if the machine has state.

overlay.d/14container/usr/libexec/coreos-podman-changes-check
```diff
@@ -0,0 +1,43 @@
...
+
+if [[ $podmanBackend != "netavark" ]]; then
+ echo -e "${warn}
```
jlebon: Instead of just echoing, we need to generate an motd.

overlay.d/14container/usr/libexec/coreos-podman-changes-check
```diff
@@ -0,0 +1,43 @@
...
+###########################################################################
+WARNING: This system is using CNI networking. CNI is deprecated and will be
+removed in the upcoming Podman v5.0, in preference of Netavark. To switch
+from CNI networking to netavark, you must run 'podman system reset --force'
```
jlebon: Minor: Inconsistent capitalization. Also let's maybe put the command on its own line.

overlay.d/14container/usr/libexec/coreos-podman-changes-check
```diff
@@ -0,0 +1,43 @@
...
+command. This will delete all of your images,containers, and custom networks.${nc}"
```
jlebon:
```suggestion
command. This will delete all of your images, containers, and custom networks.${nc}"
```

overlay.d/14container/usr/libexec/coreos-podman-changes-check
dustymabe: I really think running any `podman` command does make changes to the system. @mheon would know better than I.

One thing we could try to do is look under certain known directories to see if files have been created there and assume podman is in use if some files exist.

gursewak1997:
> We should also update the text of the cgroup check to reflect the current state.

Should we also update the warning to show how to disable this service if needed for eg. add `To disable this warning, use: sudo systemctl disable xni-checks.service`
Technically they can always disable it but it might not be a good idea to explicitly have in the warning message since we are fully dropping the support for both of them and the transition to cgroupsv2 and netavark is definitely needed.

dustymabe:
> > We should also update the text of the cgroup check to reflect the current state.
> 
> Should we also update the warning to show how to disable this service if needed for eg. add `To disable this warning, use: sudo systemctl disable xni-checks.service` Technically they can always disable it but it might not be a good idea to explicitly have in the warning message since we are fully dropping the support for both of them and the transition to cgroupsv2 and netavark is definitely needed.

I think it depends on if `docker` still supports cgroupsv1. If it does then I'd say we should give them an option to disable the message from popping up. If not, then maybe not?

overlay.d/14container/usr/libexec/coreos-podman-changes-check
```diff
@@ -0,0 +1,43 @@
...
+# Ref: https://kubernetes.io/docs/concepts/architecture/cgroups/#check-cgroup-version
+# Update an existing system from cgroupsv1 to cgroupsv2 and immediately reboot
+# Run $ sudo rpm-ostree kargs --delete=systemd.unified_cgroup_hierarchy --reboot
+
+cgroupVersion=$(stat -fc %T /sys/fs/cgroup/)
+
+if [[ $cgroupVersion == "tmpfs" ]]; then
+ echo -e "${warn}
+##########################################################################
+WARNING: This system is using cgroups v1. Podman is dropping support for
```
mheon: Update here: We have formally deprecated CGv1, but are not actually removing until v6.0 (I have no idea when that will come out, but at least a year from now). There are too many stragglers still on v1 (including WSL2), for us to completely remove at this point. So Podman will still work on v1 systems, but will pop up an annoying warning  when you do so unless silenced by environment variable.

overlay.d/14container/usr/libexec/coreos-podman-changes-check
mheon: The big issue is usually conflicts in firewall rules, and that only happens if a container is started. `podman info` on a Docker system should not be a problem.

overlay.d/14container/usr/lib/systemd/system/coreos-cni-networking-check.service
```diff
@@ -0,0 +1,11 @@
+# This service checks if the system nodes are still using
+# CNI networking. If so, they will be warned to move
+# their nodes to netavark respectively.
+[Unit]
+Description=Check if nodes are still using CNI networking
```
jlebon: Minor: this should be Title Case

overlay.d/14container/usr/libexec/coreos-cni-networking-check
```diff
@@ -0,0 +1,30 @@
...
+nc=$(echo -e '\033[0m')
+
+# Podman supports two network backends Netavark and CNI.
+# Netavark is the default network backend and was added in
+# Podman version 4.0. CNI is deprecated and is removed
+# in Podman version 5.0, in preference of Netavark.
+podmanBackend=$(podman info --format "{{.Host.NetworkBackend}}")
+
+if [[ $podmanBackend != "netavark" ]]; then
+ motd_path=/run/motd.d/35_cni_warning.motd
```
jlebon: Minor: weird indentation level here. Let's do 2 or 4?

overlay.d/14container/usr/libexec/coreos-cni-networking-check
```diff
@@ -0,0 +1,30 @@
...
+
+ cat << EOF > "${motd_path}"
+${warn}
+###########################################################################
+WARNING: This system is using CNI networking. CNI is deprecated and will be
+removed in the upcoming Podman v5.0, in preference of netavark. To switch
+from CNI networking to netavark, you must run 'podman system reset --force'
+command. This will delete all of your images, containers, and custom networks.
```
jlebon: Missing trailing `###` divider.

overlay.d/14container/usr/libexec/coreos-cni-networking-check
```diff
@@ -0,0 +1,30 @@
...
+
+if [[ $podmanBackend != "netavark" ]]; then
+ motd_path=/run/motd.d/35_cni_warning.motd
+
+ cat << EOF > "${motd_path}"
+${warn}
+###########################################################################
+WARNING: This system is using CNI networking. CNI is deprecated and will be
+removed in the upcoming Podman v5.0, in preference of netavark. To switch
+from CNI networking to netavark, you must run 'podman system reset --force'
```
jlebon: Minor: Netavark (I think? Based on upstream docs, it seems like it's capitalized when talking about the project.)

overlay.d/14container/usr/libexec/coreos-cni-networking-check
```diff
@@ -0,0 +1,30 @@
...
+# in Podman version 5.0, in preference of Netavark.
+podmanBackend=$(podman info --format "{{.Host.NetworkBackend}}")
+
+if [[ $podmanBackend != "netavark" ]]; then
+ motd_path=/run/motd.d/35_cni_warning.motd
+
+ cat << EOF > "${motd_path}"
+${warn}
+###########################################################################
+WARNING: This system is using CNI networking. CNI is deprecated and will be
```
jlebon:
```suggestion
WARNING: Podman is using CNI networking. CNI is deprecated and will be
```

overlay.d/14container/usr/libexec/coreos-cni-networking-check
jlebon: Re. whether to show instructions to disable the warning, I think we probably always should provide them. Users should always be free to disregard warnings if they know why it doesn't apply to them.

overlay.d/15fcos/usr/libexec/coreos-check-cgroups-version
```diff
@@ -11,15 +11,14 @@ motd_path=/run/motd.d/30_cgroupsv1_warning.motd
...
 cat << EOF > "${motd_path}"
 ${warn}
-############################################################################
-WARNING: This system is using cgroups v1. For increased reliability
-it is strongly recommended to migrate this system and your workloads
-to use cgroups v2. For instructions on how to adjust kernel arguments
-to use cgroups v2, see:
+##########################################################################
+WARNING: This system is using cgroups v1. Podman has dropped support for
+cgroups v1. Move your nodes to cgroups v2 if not already. For instructions
```
jlebon: This needs to be adjusted given https://github.com/coreos/fedora-coreos-config/pull/2818#discussion_r1488540086. We could delay warning it until when Podman v6.0 arrives, but meh probably good to give an early heads up.

I think also we shouldn't lose the point about it benefiting the whole system, not just Podman.

OK, how about something like this:

> This system is using cgroups v1. For increased reliability
it is strongly recommended to migrate this system and your workloads
to use cgroups v2. A future version of Podman will also drop support
for cgroups v1. For instructions on how to adjust kernel arguments
to use cgroups v2, see:

manifests/fedora-coreos.yaml
```diff
@@ -36,6 +36,7 @@ conditional-include:
 
 ostree-layers:
   - overlay/15fcos
+  - overlay/14container
```
jlebon: I would call this overlay something clearer, e.g. `14podman-v5`. And add a README to document it. Also, let's make it conditional on Fedora 39? It doesn't hurt on Fedora 40 (we just will never warn since podman can't support anything other than Netavark), but it makes it clearer when we can drop it.

gursewak1997: just confirming, we don't really need it to be conditional right? Aiming to keep it and update the README.

jlebon: We don't _need_ it to be conditional, but we also don't _need_ it to be in Fedora 40. What I mean is making it conditional will make it more likely to get cleaned up when we scan all the overlays we can drop when we rebase on top of f40.

overlay.d/14container/usr/libexec/coreos-cni-networking-check
dustymabe: will `podman system reset --force` delete your existing containers? If so, maybe we should have a FAQ entry in our docs we should point people to instead for more context so they can make the right decisions.

dustymabe: ok I see line 27 below now. :(

overlay.d/14container/usr/libexec/coreos-cni-networking-check
dustymabe: Maybe we should mention that "depending on your setup it may be preferable to reprovision the whole machine from the latest images". because reprovisioning will set things up the right way too.

dustymabe (dismissed):
I don't see anything alarming. I added some comments.

I'll trust @jlebon to review from here on out.

overlay.d/14podman-v5/usr/libexec/coreos-cni-networking-check
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/bash
```
jlebon, line 1: Missing `set -euo pipefail`

overlay.d/14podman-v5/usr/libexec/coreos-cni-networking-check
```diff
@@ -0,0 +1,36 @@
...
+if [[ $podmanBackend != "netavark" ]]; then
+    motd_path=/run/motd.d/35_cni_warning.motd
+
+    cat << EOF > "${motd_path}"
+${warn}
+##############################################################################
+WARNING: Podman is using CNI networking. CNI is deprecated and will be
+removed in the upcoming Podman v5.0, in preference of Netavark. To switch
+from CNI networking to Netavark, you must run 'podman system reset --force'
+command. This will delete all of your images, containers, and custom networks.
```
jlebon: Hmm, "command." looks weird here. Either e.g. "you must run the command '...'." or "you must run '...'.".

overlay.d/15fcos/usr/libexec/coreos-check-cgroups-version
```diff
@@ -11,15 +11,16 @@ motd_path=/run/motd.d/30_cgroupsv1_warning.motd
...
+to use cgroups v2. A future version of Podman will also drop support
+for cgroups v1. For instructions on how to adjust kernel arguments
 to use cgroups v2, see:
 https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/
 
 To disable this warning, use:
-sudo systemctl disable coreos-check-cgroups.service
-############################################################################
+sudo systemctl disable coreos-check-cgroups-version.service
+###########################################################################
```
jlebon: Super minor: this is longer than the first divider.

overlay.d/14podman-v5/usr/lib/systemd/system/coreos-cni-networking-check.service
```diff
@@ -0,0 +1,11 @@
+# This service checks if the system nodes are still using
+# CNI networking. If so, they will be warned to move
+# their nodes to netavark respectively.
+[Unit]
+Description=Check if Nodes Are Still Using CNI Networking
```
jlebon:
```suggestion
Description=Check If Podman Is Still Using CNI Networking
```

jlebon (dismissed):
Some superficial things, but LGTM overall. Thanks!

jlebon (approved):

PR #2819, opened 2024-01-25
ci/overrides.py: add podman to trivial fast tracks

dustymabe:
I'd say it's routine that we want to pick up the newest version of podman before it is stable in Fedora (especially if it has passed tests).

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -22,6 +22,7 @@ TRIVIAL_FAST_TRACKS = [
     'console-login-helper-messages',
     'ignition',
     'ostree',
+    'podman',
     'rpm-ostree',
     'rust-afterburn',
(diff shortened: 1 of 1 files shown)


jlebon:
Hmm, not sure about this. The reason why a lot of the packages on that list are there is because they're very specific to FCOS and so don't really get any valuable testing in Bodhi anyway. That's not the case for podman.

Not strongly against either, but given that podman is an important component, I'd rather default to letting it bake in Bodhi (and fast-track when we really know we want it).

dustymabe:
Yeah. I was thinking we usually would want to fast track podman to decrease the time it takes for a podman release to hit `testing` and thus `stable`.

I'm OK with not doing this though.

PR #2824, opened 2024-01-29
[rawhide] overrides: drop `kernel-6.7.0-68.fc40 pin`

gursewak1997:
The issue was fixed with the latest kernel build.
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1647


jlebon (approved):

dustymabe:
Noting that apparently the new kernel isn't in the rawhide repos yet :(

Maybe I'll just fast-track it.

marmijo:
I agree to fast-track it. We havent had a rawhide kernel update in 2 weeks since https://bodhi.fedoraproject.org/updates/FEDORA-2024-4b6043736b

PR #2825, opened 2024-01-30
[rawhide] overrides: fast-track kernel-6.8.0-0.rc2.20240130git861c0981648f.20.fc40

coreosbot-releng:


marmijo (approved):
This bodhi update is still in the testing state with negative karma. 
+1 from me to fast-track this kernel for now so we can pick up the latest updates in rawhide.

PR #2826, opened 2024-01-30
tests/kernel-replace: tag with "reprovision"

dustymabe:
This is a hack to get this test to run not in parallel with other tests. The performance on our ppc64le builder has been abysmal and this is one lever we can pull. We have bumped the timeout for the test in the past. It's already at 20m. I say we just start running this test serially.

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -9,7 +9,11 @@
 ##   # RHCOS. See: https://github.com/containers/skopeo/issues/1846
 ##   distros: fcos
 ##   # Needs internet access as we fetch files from koji
-##   tags: "needs-internet platform-independent"
+##   # We add the "reprovision" tag here even though we aren't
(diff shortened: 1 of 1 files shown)


marmijo (approved):
LGTM

jlebon:
It's always been a bit awkward how we mentioned the `reprovision` tag in the `kola()` step. We should probably add a new `serial` tag instead to all those tests and use that in `kola()`.

dustymabe:
> It's always been a bit awkward how we mentioned the `reprovision` tag in the `kola()` step. We should probably add a new `serial` tag instead to all those tests and use that in `kola()`.

I'd be happy with that. Though I think it would be nicer if there was a way to have it happen internally in `kola` instead of having to do `--parallel=4 --tag=!serial` and then `--tag=serial`. 

i.e. tests with the `serial` tag would run serially even when `--parallel=4` was specified

jlebon:
We could certainly make it a special tag. The caveat is that on many developer laptops, you do have enough resources to run them in parallel.

PR #2829, opened 2024-01-31
[rawhide] overrides: pin kernel-6.8.0-0.rc2.19.fc40 on aarch64

gursewak1997:
Since we are still seeing the booting [error](https://github.com/coreos/fedora-coreos-tracker/issues/1647) on aarch64 using the latest kernel, let's pin it to `6.8.0-0.rc2.19.fc40`


dustymabe (approved):
LGTM

dustymabe:
CI won't test this because it's aarch64 specific.

PR #2838, opened 2024-02-06
40ignition-ostree: add ignition-ostree-mount-state-overlays.service

jlebon:
The new experimental `opt-usrlocal-overlays` treefile knob[1] allows users to overlay packages with `/opt`/`/usr/local` content (and eventually, rebase to container images with content in those places).

However, booting a base compose with this knob will break Ignition configs that currently write data in e.g. `/opt` since that now points to `/usr/lib/opt`, which is of course read-only.

We need to assemble the state overlays from the initramfs so that those configs keep working seamlessly.

This is a no-op if state overlays are off (status quo), but it'll make it easier for people and CI to test the feature.

This is a tiny part of the required bits if we want to eventually turn this on in FCOS/RHCOS. The other major part is migrating existing systems.

[1] https://github.com/coreos/rpm-ostree/pull/4728

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+set -euo pipefail
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if [ $# -ne 1 ] || { [[ $1 != mount ]] && [[ $1 != umount ]]; }; then
+    fatal "Usage: $0 <mount|umount>"
+fi
+
+# if state overlays are not enabled, there's nothing to do
+if ! ls /sysroot/usr/lib/systemd/system/local-fs.target.requires/ostree-state-overlay@*.service 2>/dev/null; then
+    exit 0
+fi
+
+do_mount() {
+    # be nice to persistent /var; if the top-level state overlay dir exists,
+    # then assume it's properly labeled
+    relabel=1
+    state_overlays_dir=/sysroot/var/ostree/state-overlays
+    if [ -d ${state_overlays_dir} ]; then
+        relabel=0
+    fi
+    for overlay in /usr/lib/opt /usr/local; do
+        escaped=$(systemd-escape --path "${overlay}")
+        overlay_dirs=${state_overlays_dir}/${escaped}
+        mkdir -p "${overlay_dirs}"/{upper,work}
+        # ideally we'd use `ostree admin state-overlay`, but that'd require
+        # pulling in bwrap and chroot which isn't yet in the FCOS initrd
+        mount -t overlay overlay /sysroot/${overlay} -o "lowerdir=/sysroot/${overlay},upperdir=${overlay_dirs}/upper,workdir=${overlay_dirs}/work"
+    done
+    if [ $relabel = 1 ]; then
+        coreos-relabel /var/ostree
+        # the above relabel will have relabeled the upperdir too; relabel that
+        # from the perspective of the mount point so it's not var_t
+        for overlay in /usr/lib/opt /usr/local; do
+            coreos-relabel ${overlay}
+        done
+    fi
+}
+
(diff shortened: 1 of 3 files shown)


overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# if state overlays are disabled, there's nothing to do
+if ! jq -e '.["opt-usrlocal-overlays"]' /sysroot/usr/share/rpm-ostree/treefile.json; then
+    exit 0
+fi
+
+do_mount() {
+    for overlay in /usr/lib/opt /usr/local; do
+        escaped=$(systemd-escape --path "${overlay}")
+        overlay_dirs=/sysroot/var/ostree/state-overlays/${escaped}
```
dustymabe: Since we only operate on one overlay dir at a time maybe:

```suggestion
        overlay_dir=/sysroot/var/ostree/state-overlays/${escaped}
```

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
```diff
@@ -0,0 +1,45 @@
...
+        # be nice to persistent /var; if the dirs already exist, assume they're
+        # properly labeled
+        need_relabeling=0
+        if [ ! -d "${overlay_dirs}" ]; then
+            mkdir -p "${overlay_dirs}"/{upper,work}
+            coreos-relabel /var/ostree
```
dustymabe: So this gets executed multiple times.. any reason that needs to happen or can we just move `coreos-relabel /var/ostree` to the `need_relabeling` if down on line 34?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+
+if [ $# -ne 1 ] || { [[ $1 != mount ]] && [[ $1 != umount ]]; }; then
+    fatal "Usage: $0 <mount|umount>"
+fi
+
+# if state overlays are disabled, there's nothing to do
+if ! jq -e '.["opt-usrlocal-overlays"]' /sysroot/usr/share/rpm-ostree/treefile.json; then
+    exit 0
+fi
```
dustymabe: feels a bit icky to dig into the treefile to determine if this is supported but I guess it's the only way and we probably have precedent for doing something like this in the past?

dustymabe (commented):
mostly LGTM - a few comments

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
jlebon: Changed the approach for this!

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
jlebon: A better source of truth for this is actually whether the systemd services are enabled. I changed it to checked that (which actually also matches some rpm-ostree code).

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-state-overlays.sh
jlebon: It's plural because that dir contains `work` and `upper`. This matches the wording in e.g. https://github.com/ostreedev/ostree/blob/da892140657bfb67d475f80395af97af96b1b6af/src/ostree/ot-admin-builtin-state-overlay.c#L48.

dustymabe (approved):
LGTM

PR #2841, opened 2024-02-06
overlay/05core: use generator vs static drop-in for rescue/emergency services

bsherman:
This uses a systemd-generator to replace the static drop-ins which provide SYSTEMD_SULOGIN_FORCE=1 for rescue.service and emergency.service.

This improves security by not grantng a root shell upon a fsck failure, for example. This respects password protected bootloaders and should be compatible with hardening efforts such as:
https://github.com/coreos/fedora-coreos-tracker/issues/805

This was recently merged into the Universal Blue images: https://github.com/ublue-os/main/pull/488


cgwalters (approved):
This looks sane to me.

PR #2844, opened 2024-02-07
manifests: add new passwd manifest, include only on F39

dustymabe:
The passwd RPM is no longer available in F40+ [1]. The passwd
binary is provided by shadow-utils, which we already name. Let's
configure it so that passwd gets named in the package lists only
on F39.

[1] https://src.fedoraproject.org/rpms/shadow-utils/c/91360f25a8c8b810d59bec2803a2477a2647c775?branch=rawhide

Also did a bunch of cleanup in the manifests. See individual commit messages.


AdamWill:
+1 on the passwd change, I have already done this in comps and workstation-ostree-config and it fixed the openQA failure to build ostree installer images. Can't speak to the rest.

jlebon (approved):
Nice!

PR #2846, opened 2024-02-08
manifests/bootupd: Remove 'mkdir -p /run' workaround

travier:
This was added a while back as a fast-track for a fix in rpm-ostree and we don't need it anymore.

Reverts: https://github.com/coreos/fedora-coreos-config/commit/cb300804dbb0ffea4e04b84cfb1109b4585c29c5


jbtrystram (approved):
lgtm

PR #2847, opened 2024-02-08
05core: skip sysroot.mount for remote kdump

jbtrystram:
If kdump is setup to upload logs to a remote host, the XFS module may not be loaded, which cause sysroot.mount to fail.

If the target for logs is SSH, kdump will boot with a `kdump_remote_ip` karg. Skip generating sysroot.mount in that case so kdump can boot without having to load XFS.

See https://issues.redhat.com/browse/OCPBUGS-27935

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -66,7 +66,11 @@ mkdir -p ${UNIT_DIR}
 # We want to generate sysroot.mount on ostree systems, but we don't want
 # to run if there's already a root= karg, where the systemd-fstab-generator
 # should win.
-if test -n "$(cmdline_arg ostree)" && test -z "$(cmdline_arg root)"; then
+# We also don't want to generate sysroot.mount if we are booting a kdump kernel
+# that aims to upload logs to a remote target, as the XFS kernel module won't be loaded
+# https://issues.redhat.com/browse/OCPBUGS-27935
+# FIXME: this kdump case should be removed when we are done with https://github.com/coreos/fedora-coreos-tracker/issues/1675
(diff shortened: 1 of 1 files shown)


dustymabe:
would an alternative approach to this be to just make sure the XFS module gets loaded in the initramfs?

jbtrystram:
It's what I suggested on https://issues.redhat.com/browse/OCPBUGS-27935
But if I understand things correctly, we don't control what is shipped in kdump initramfs, as it's generated each boot, and it depends on what is in the kdump config, which may be overloaded by the users

jlebon (dismissed):
Yeah, this is not ideal but at least it's contained.

I think the real fix here is probably a barrier release to finally migrate older systems to add a `root` karg, and then we'd change the logic here to only emit `sysroot.mount` on first boot (i.e. move it below the `ignition.firstboot` check).

dustymabe:
another option would be to convince kdump upstream to name the module in their default config. but if we want to do the `root` karg dance/migration anyway then maybe we just focus on that?

jbtrystram:
I am not so sure about I should do here , after reading comments from @dustymabe and @jlebon :)

jlebon:
IMO I think we should get this in to unbreak anyone using kdump+SSH. Doing the migration is good too, but must be done carefully.

How about filing a tracker issue for the migration and then linking that as a comment in the code mentioning that we'll be able to clean this up once it's implemented?

jbtrystram:
> IMO I think we should get this in to unbreak anyone using kdump+SSH. Doing the migration is good too, but must be done carefully.
> 
> How about filing a tracker issue for the migration and then linking that as a comment in the code mentioning that we'll be able to clean this up once it's implemented?

Ok, I updated the commit message and also add a link in the code

jlebon (approved):

travier:
We can not rely on loading the right filesystem module as kdump maybe be used in a "diskless" / upload to remote mode which will not touch the disks at all, for example if those are LUKS encrypted.

PR #2852, opened 2024-02-13
manifest: drop `containernetworking-plugins` and `podman-plugins`

gursewak1997:
Since podman is dropping support for CNI networking entirely we can drop `containernetworking-plugins` and  `podman-plugins`. Podman stopped shipping `podman-plugins` rpm with the Podmanv5. Subsequently, we wouldn't need `containernetworking-plugins` as it is just libraries for writing CNI plugins.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1629#issuecomment-1939353789

### DIFF
--- a/manifests/cni-plugins.yaml
+++ b/manifests/cni-plugins.yaml
@@ -1,0 +1,5 @@
+# Issue: https://github.com/coreos/fedora-coreos-config/pull/2818
+# Delete this file once we are on Podman v5 everywhere.
+# i.e. We have moved past F39.
+packages:
+  - containernetworking-plugins podman-plugins
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -104,11 +104,10 @@ packages:
   # Containers
   - systemd-container catatonit
   - fuse-overlayfs slirp4netns
-  # support for old style CNI networks and name resolution for
-  # podman containers with CNI networks
+  # Some host applications(i.e. NetworkManager) use dnsmasq
+  # as the binary for some various utility operations.
   # https://github.com/coreos/fedora-coreos-tracker/issues/519
-  # https://github.com/coreos/fedora-coreos-tracker/issues/1128#issuecomment-1071338097
-  - containernetworking-plugins podman-plugins dnsmasq
+  - dnsmasq
   # For podman v4 netavark gets pulled in but it only recommends
   # aardvark-dns (which provides name resolution based on container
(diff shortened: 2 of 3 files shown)


jlebon:
This needs to be made conditional on Fedora 40. Or in other words, you would move the `containernetworking-plugins` and `podman-plugins` bits to a separate manifest, and add a conditional include for f39 for that manifest. Grep for `conditional-include` to see examples of this.

Also /cc @mheon to sanity-check: there's no point in having these packages in f40, right?

manifests/fedora-coreos-base.yaml
```diff
@@ -104,11 +104,10 @@ packages:
   # Containers
   - systemd-container catatonit
   - fuse-overlayfs slirp4netns
-  # support for old style CNI networks and name resolution for
-  # podman containers with CNI networks
+  # Some host applications((podman,NetworkManager, etc) ) use
```
dustymabe: The only known use now is NM

```suggestion
  # Some host applications(i.e. NetworkManager) use
```

mheon:
Podman-plugins we definitely have no further use for. Containernetworking-plugins is still used by CRI-O but I'm not sure how that ships in Fedora

jlebon:
> Podman-plugins we definitely have no further use for. Containernetworking-plugins is still used by CRI-O but I'm not sure how that ships in Fedora

Ack, thanks.

Looks like the CRI-O in Fedora `Suggests` it. Since CRI-O currently needs to be layered separately, I think we could also have any users doing that also layer this if they need it. We should mention it as part of our f40 communications.

manifests/cni-plugins.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
```
dustymabe, line 4: I would add a comment up at the top here to let us know when we can delete this file. See for example the comment at the top of the removed bash-color-prompt.yaml from 24142b2

dustymabe (dismissed):
one suggestion. otherwise LGTM

manifests/cni-plugins.yaml
```diff
@@ -0,0 +1,4 @@
+# Issue: https://github.com/coreos/fedora-coreos-config/pull/2818
+# Delete this file once Podman v5 releases to f39
```
dustymabe: podman v5 will never go to F39. 

```suggestion
# Delete this file once we are on Podman v5 everywhere. 
# i.e. We have moved past F39.
```

dustymabe (approved):
LGTM

PR #2856, opened 2024-02-15
Enable composefs for 41+

cgwalters:
We really want to aim to eventually enable this by default, let's test things out in rawhide.

A thing that this is known to break is the "chattr -i" hack for new toplevel dirs (xref https://github.com/coreos/rpm-ostree/issues/337 )

Basically if you want that, you either need to make a derived image, or enable transient root.

PR #2857, opened 2024-02-15
Enable composefs for 41+

cgwalters:
This is testing https://github.com/coreos/fedora-coreos-config/pull/2856 applied to rawhide


jbtrystram:
Done in https://github.com/coreos/fedora-coreos-config/pull/3009

PR #2858, opened 2024-02-16
manifests: enable cliwrap on Fedora 40+

jlebon:
As part of the bootable containers effort, we want to put emphasis on a consistent experience when deriving images. A big part of that is being able to type `dnf install -y ...` in one's `Containerfile`.

Let's turn on cliwrap for this. This was previously agreed[[1]], but we ended never shipping it since there was a less invasive way to test it.[[2]] Here, we bundle it as part of the Fedora 40 rebase. This will naturally allow it to bake in `branched` and `next` for a while before making it to `stable`.

Eventually, once dnf5 is ready, it could take on this role (including package layering on the client-side), at which point we will no longer need to wrap it.

[1]: https://github.com/coreos/fedora-coreos-config/pull/830#issuecomment-780840462
[2]: https://github.com/coreos/fedora-coreos-config/pull/830#issuecomment-789806308

### DIFF
--- a/manifests/cliwrap.yaml
+++ b/manifests/cliwrap.yaml
@@ -1,0 +1,2 @@
+# https://github.com/coreos/fedora-coreos-tracker/issues/730
+cliwrap: true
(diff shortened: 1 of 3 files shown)


jlebon:
Also done in RHCOS in https://github.com/openshift/os/pull/1442.

jlebon:
Rebased! As per https://github.com/coreos/fedora-coreos-tracker/issues/730#issuecomment-1957763169, I think this is ready for review now.

dustymabe (approved):
LGTM

PR #2859, opened 2024-02-16
Move versionary script here

jlebon:
When we emptied this script out of
https://github.com/coreos/fedora-coreos-releng-automation/ in [[1]], I'm not sure why I decided to move it to cosa instead of this repo, where it clearly belongs much better.

[1]: https://github.com/coreos/coreos-assembler/pull/3113

### DIFF
--- a/versionary
+++ b/versionary
@@ -1,0 +1,173 @@
+#!/usr/bin/python3 -u
+
+# This file originally lived in
+# https://github.com/coreos/fedora-coreos-releng-automation. See that repo for
+# archeological git research.
+
+'''
+    Implements the Fedora CoreOS versioning scheme as per:
+        https://github.com/coreos/fedora-coreos-tracker/issues/81
+        https://github.com/coreos/fedora-coreos-tracker/issues/211
+'''
+
+import argparse
+import json
+import os
+import re
+import subprocess
+import sys
+import time
+import yaml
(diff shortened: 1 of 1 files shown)


dustymabe:
> When we emptied this script out of https://github.com/coreos/fedora-coreos-releng-automation/ in [[1](https://github.com/coreos/coreos-assembler/pull/3113)], I'm not sure why I decided to move it to cosa instead of this repo, where it clearly belongs much better.

I kind of understand why we moved it to COSA. This repo doesn't really have much code in it other than what is under the `ci/` dir and used for CI.

dustymabe:
> This repo doesn't really have much code in it other than what is under the `ci/` dir and used for CI.

and I'm not sure if that's the most ideal path forward, but clearly it makes whatever you are trying to do here easier so I'm fine with it.

dustymabe (approved):

jlebon:
Yeah. The way I think of it is that the versioning scheme is a core part of the OS definition, and so belongs better here. Another approach that would capture this more would be to fold this into the rpm-ostree manifest itself, e.g. via some `automatic-version-script: path/to/script` knob, but that doesn't seem worth the complexity. Not opposed to moving it to `ci/` (or maybe better a `scripts/` dir?).

PR #2860, opened 2024-02-19
[rawhide] overrides: fast-track `coreos-installer` and `bootupd`, drop `util-linux`

marmijo:
A new util-linux package was released with a fix for the issue, let's remove the pin to pick up the latest rpm.
See: https://github.com/coreos/fedora-coreos-tracker/issues/1666


dustymabe (dismissed):

dustymabe:
Updated this with a few fast-tracks.

gursewak1997 (approved):

dustymabe:
The only failed kola test was a `warn: true` test. Merging.

PR #2861, opened 2024-02-19
overlay.d/40grub: delete 30_console.cfg file

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/1671 where we discovered that having this be a static config delivered by bootupd was a bad idea.

Instead we'll have something else create a file in /boot/grub2/console.cfg (most likely https://github.com/osbuild/osbuild/pull/1589) and we'll have the bootupd static grub configs source it (https://github.com/coreos/bootupd/pull/619).


dustymabe:
ok this is looking good in testing. lifting draft

jlebon (approved):

PR #2863, opened 2024-02-20
[testing] Move versionary script here

jlebon:
When we emptied this script out of
https://github.com/coreos/fedora-coreos-releng-automation/ in [[1]], I'm not sure why I decided to move it to cosa instead of this repo, where it clearly belongs much better.

[1]: https://github.com/coreos/coreos-assembler/pull/3113

### DIFF
--- a/versionary
+++ b/versionary
@@ -1,0 +1,173 @@
+#!/usr/bin/python3 -u
+
+# This file originally lived in
+# https://github.com/coreos/fedora-coreos-releng-automation. See that repo for
+# archeological git research.
+
+'''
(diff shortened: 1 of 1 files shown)


jlebon:
Fast-track of https://github.com/coreos/fedora-coreos-config/pull/2859 so it gets to stable faster.

jlebon:
This code is inert until https://github.com/coreos/fedora-coreos-pipeline/pull/961 is actually merged, which we'll do only after the next set of releases.

dustymabe (approved):

PR #2864, opened 2024-02-20
denylist: snooze tests on branched stream

marmijo:
F40 has branched from rawhide, so we'll be tracking it in the branched stream once enabled: https://github.com/coreos/fedora-coreos-pipeline/pull/962.
Let's also snooze `ext.config.var-mount.scsi-id` on branched since it's been failing on F40.

EDIT: removed `ext.config.files.console-config`

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -21,3 +21,4 @@
   warn: true
   streams:
     - rawhide
+    - branched


kola-denylist.yaml
```diff
@@ -15,9 +15,11 @@
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1654
   streams:
     - rawhide
+    - branched
```
dustymabe: actually not needed as we've only enabled OSBuild on `rawhide`, but it's fine to leave it like this too.

dustymabe (dismissed):

kola-denylist.yaml
marmijo: I can remove it if it's unnecessary :+1:

jlebon (approved):

PR #2868, opened 2024-02-21
tests/var-mount/scsi-id: simplify bootloader entry finding

jlebon:
We don't have to be super strict here in how we find the bootloader
entry. There should only be one, so simplify the logic using a glob
instead.

Motivated by the fact that this will break otherwise as part of
openshift/os#1445 where the `ID` will be
`centos`, but the stateroot will still be `scos`.


dustymabe (approved):

dustymabe:
I assume this has nothing to do with https://github.com/coreos/fedora-coreos-tracker/issues/1670

PR #2869, opened 2024-02-23
denylist: extend snooze for var-mount.scsi-id test

dustymabe:
This is still outstanding. Let's snooze it so we don't keep getting warnings from builds.


aaradhak (approved):
LGTM

PR #2876, opened 2024-02-26
 denylist: drop `ext.config.files.console-config` snooze

gursewak1997:
In reference to https://github.com/coreos/coreos-assembler/pull/3742 this test should be passing now. Updating the denylist accordingly.


dustymabe (approved):
LGTM

PR #2877, opened 2024-02-26
manifests: Split container engines and deps into sub manifests

travier:
manifests: Split container engines into sub-manifests

- Split podman, moby-engine and related packages into distinct
  sub-manifests
- Include both container engines by default
- Ensure that they are excluded if specified as such

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -11,6 +11,9 @@ rojig:
 add-commit-metadata:
   fedora-coreos.stream: ${stream}
 
+# Include all container engines by default
+variables:
+  container_engines: true
 
 include: fedora-coreos-base.yaml
 conditional-include:
@@ -27,17 +30,25 @@ conditional-include:
     # passwd RPM was obsoleted by shadow-utils in F40+, but we need to keep
     # including it in F39. Remove this when we are on F40+ on all streams.
     include: passwd.yaml
-  - if: releasever == 39
+  - if:
+      - releasever == 39
+      - container_engines == true
     # No longer need CNI plugins in f40 since Podman dropped
     # support for CNI networking entirely in Podmanv5
     # Remove this once we have Podman v5 in f39.
     # xref: https://github.com/coreos/fedora-coreos-tracker/issues/1629
     include: cni-plugins.yaml
-  - if: releasever == 39
+  - if:
+      - releasever == 39
+      - container_engines == true
     # Checks for breaking changes that came with Podman v5.
     include: podman-v5.yaml
   - if: releasever >= 40
     include: cliwrap.yaml
+  - if: container_engines == true
+    include:
+      - podman.yaml
+      - moby-engine.yaml
 
 ostree-layers:
   - overlay/15fcos
@@ -55,10 +66,6 @@ packages:
   # the archive repo for more reliable package layering
   # https://github.com/coreos/fedora-coreos-tracker/issues/400
   - fedora-repos-archive
-  # CL ships this.
-  - moby-engine
-  # Already pulled in by moby-engine, but let's be explicit. Typhoon uses it.
-  - containerd
   # Updates
   - zincati
   # Include and set the default editor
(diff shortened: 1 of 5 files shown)


manifest.yaml
```diff
@@ -1,6 +1,7 @@
 variables:
   stream: testing-devel
   prod: false
+  podman-machine-os: false
```
dustymabe: I imagine we'll use a new stream name for the new stream. My proposal is that we make our conditional logic based on that and not a new var here.

manifests/fedora-coreos.yaml
```diff
@@ -38,6 +38,10 @@ conditional-include:
     include: podman-v5.yaml
   - if: releasever >= 40
     include: cliwrap.yaml
+  - if: podman-machine-os == false
+    include:
+      - podman.yaml
+      - moby-engine.yaml
```
dustymabe, line 51: note we'll probably also need to add an extra condition to the cni-plugins.yaml statement above to make it not included (because it would pull in podman anyway).

I guess that is unless we can assume the podman-machine-os stuff would only happen for F40+.

manifests/podman.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - podman
+  - crun
```
dustymabe, line 5: any openshift/os changes we need to make because we're breaking this out into a separate manifest?

manifest.yaml
dustymabe: This has the added benefit of not requiring us to change all of our `manifest.yaml` files on every branch.

travier: There is likely going to be multiple streams, one per version of podman

manifests/fedora-coreos.yaml
travier: podman v5 is F40 only so that's a good assumption. I can tweak that as well.

manifests/podman.yaml
travier: Likely, will look at this next.

manifest.yaml
travier: But if we define the streams exhaustively that could work. Or we can probably set a "default" value somewhere that is overridden in more top level manifest?

dustymabe: so add multiple conditions?

manifests/fedora-coreos.yaml
travier: done

manifest.yaml
travier: I've added a variable set with a default value that can be overridden by a top level manifest.

travier:
Things to clarify:
- Should we include/exclude runc/crun?
- Should we include/exclude skopeo (pulls-in container-commons)?

manifests/podman.yaml
```diff
@@ -0,0 +1,19 @@
...
+packages-aarch64:
+  - crun-wasm wasmedge-rt
+
+postprocess:
+  # Enable SELinux booleans used by OpenShift
+  # https://github.com/coreos/fedora-coreos-tracker/issues/284
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    setsebool -P -N container_use_cephfs on  # RHBZ#1692369
```
jlebon, line 34: Hmm, this seems more general than a podman thing. Let's keep that bit in `fedora-coreos-base.yaml`?

jlebon (commented):
> Things to clarify:
> 
>     * Should we include/exclude runc/crun?

What you have right now looks reasonable to me. I would expect anyone using `container_engines: false` would want to make that choice themselves in their derived layers. (But in practice, likely they'd just let the container engine of choice pull in its dep.)

>     * Should we include/exclude skopeo (pulls-in container-commons)?

This is used by rpm-ostree for the container update flow.

Actually, it's a runtime dep so a minimal compose would break anyway if it were excluded.

manifests/podman.yaml
travier: This requires the containers selinux policy module otherwise it fails.

travier:
This is working for me with https://github.com/travier/podman-machine-config

dustymabe:
Converted back to draft. Let's hold this while we discuss with the podman folks more.

dustymabe:
Should we close this?

travier:
I think it's still worth doing, even if it's not urgent anymore.

PR #2879, opened 2024-02-27
workflows/openshift-os allow custom branches

jbtrystram:
Allow specifying the source branch to open a PR from as well as the target branch to compare against.
This is useful for backporting work

### DIFF
--- a/.github/workflows/openshift-os.yml
+++ b/.github/workflows/openshift-os.yml
@@ -6,11 +6,21 @@ on:
     - cron: '0 0 * * *'
   workflow_dispatch:
     inputs:
-      branch:
+      branch-name:
         # Allow overriding branch for syncs that need manual fixups
-        description: PR branch
+        description: PR branch name
         required: true
         default: fcc-sync
+      branch:
+        # Allow specifying the source branch for backports
+        description: Source branch to use for PR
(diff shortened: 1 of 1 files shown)


yasminvalim (approved):
LGTM!

dustymabe:
do we often sync from a branch other than `testing-devel` ?

dustymabe:
I guess we do sync over to the release branches but we've been just doing that manually up until this point?

jlebon:
A more streamlined/less error-prone tweak on this is to have it take a single input instead [of type `choice`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs) where you can choose an OCP release and it knows which f-c-c branch to open against which openshift/os branch.

jbtrystram:
@jlebon yes, i thought of that, but we'd have to update this workflow file for each ocp release, that sounds like something we will forgot to do :)

jlebon:
> @jlebon yes, i thought of that, but we'd have to update this workflow file for each ocp release, that sounds like something we will forgot to do :)

The default would just be e.g. `dev` or something to track `testing-devel` -> `master`. For branched releases, we have a checklist already of things to do that we could add to.

PR #2880, opened 2024-02-27
workflows/openshift-os allow custom branches

jbtrystram:
Allow specifying the source branch to open a PR from
as well as the target branch to compare against.
This is useful for backporting work

Also update the action as there were some deprecation warnings

### DIFF
--- a/.github/workflows/openshift-os.yml
+++ b/.github/workflows/openshift-os.yml
@@ -6,22 +6,20 @@ on:
     - cron: '0 0 * * *'
   workflow_dispatch:
     inputs:
-      branch-name:
-        # Allow overriding branch for syncs that need manual fixups
-        description: PR branch name
-        required: true
-        default: fcc-sync
       branch:
         # Allow specifying the source branch for backports
         description: Source branch to use for PR
         required: true
         default: testing-devel
-      target-branch:
-        # Allow specifying the target branch for backports
-        description: Target branch to open PR against
-        required: true
-        default: master
-
+        type: choice
+        options:
+          - testing-devel
+          - rhcos-4.15
+          - rhcos-4.14
+          - rhcos-4.13
+          - rhcos-4.12
+          - rhcos-4.11
+          - rhcos-4.10
 permissions:
   # none at all
   contents: none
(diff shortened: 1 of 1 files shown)


.github/workflows/openshift-os.yml
```diff
@@ -77,10 +79,10 @@ jobs:
           EOF
 
       - name: Open pull request
-        uses: peter-evans/create-pull-request@v4.2.3
+        uses: peter-evans/create-pull-request@v6.0.0
         with:
           token: ${{ secrets.COREOSBOT_RELENG_TOKEN }}
-          push-to-fork: coreosbot-releng/os
+          push-to-fork: openshift/os
```
jlebon: This looks incorrect.

.github/workflows/openshift-os.yml
```diff
@@ -67,7 +69,7 @@ jobs:
               echo "No non-trivial changes; exiting"
               exit 0
           fi
-          git checkout $SOURCE_BRANCH
```
jlebon, line 70: Without this, there's nothing actually updating the submodule, right?

jlebon (commented):
My 2c is to rework this to use a `choice` input instead. 3 inputs is a lot to get right (esp. the `branch-name` one; forgetting to update that one would stomp on an open PR opened against a completely different branch -- though maybe the `create-pull-request` action would error out on that).

.github/workflows/openshift-os.yml
jbtrystram: yes, that was what I got confused about yesterday !

jbtrystram:
Ok, here is a working version of the workflow :) 

Successful run (on my fork to avoid noise on openshift/os ) : https://github.com/jbtrystram/fedora-coreos-config/actions/runs/8081693445/job/22080845871
Created PR : (here backporting 4.15 release) -> https://github.com/jbtrystram/openshift-os/pull/1

.github/workflows/openshift-os.yml
```diff
@@ -6,22 +6,23 @@ on:
...
-        description: Target branch to open PR against
-        required: true
-        default: master
-
+        type: choice
+        options:
+          - testing-devel
+          - branched
+          - rawhide
+          - next-devel
```
jlebon: We can drop these. We never normally need to sync those.

.github/workflows/openshift-os.yml
```diff
@@ -6,22 +6,23 @@ on:
...
+          - rhcos-4.10
+          - rhcos-4.11
+          - rhcos-4.12
+          - rhcos-4.13
+          - rhcos-4.14
+          - rhcos-4.15
```
jlebon: Is the order here the same order the choices are laid out when you trigger the action? If so, should we order it the other way instead? (I.e. 4.15 first, then down to 4.10.)

.github/workflows/openshift-os.yml
```diff
@@ -30,35 +31,54 @@ jobs:
...
+                echo "BRANCH_NAME=fcc-sync" >> $GITHUB_ENV
+              ;;
+              rhcos-*)
+              # split the string around the -
+                array=(${SOURCE_BRANCH//-/ })
+                OCP_VERSION=${array[1]}
+                echo "TARGET_BRANCH=release-${OCP_VERSION}" >> $GITHUB_ENV
+                echo "BRANCH_NAME=fcc-sync-${SOURCE_BRANCH}" >> $GITHUB_ENV
+                ;;
+              *)
```
jlebon, line 48: We should just error out in this path since it shouldn't be reachable.

jlebon (commented):
This looks great!

.github/workflows/openshift-os.yml
jbtrystram: yes. Fixed.

.github/workflows/openshift-os.yml
jbtrystram: I am not sure if the default values are set when running on the `schedule` event. the syntax does not make it look like so, so I'd keep that default case there IMO

jbtrystram: yep, further testing confirm this: the `on_schedule` event do not initialize the variables, so we need this block :)

jbtrystram: I pushed an update enforcing `testing-devel`, `master` and `fcc-sync` in that default case ( to keep consistency the previous action behaviour for the daily sync)

jlebon (approved):

PR #2881, opened 2024-02-27
[branched] overrides: pin authselect-1.5.0-3.fc40

coreosbot-releng:
Requested by @jmarrero via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,14 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  authselect:
+    evr: 1.5.0-3.fc40
+    metadata:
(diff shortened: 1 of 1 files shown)


manifest-lock.overrides.yaml
```diff
@@ -8,4 +8,14 @@
...
+packages:
+  authselect:
+    evr: 1.5.0-3.fc40
+    metadata:
+      reason: https://bugzilla.redhat.com/show_bug.cgi?id=2266344
+      type: pin
+  authselect-libs:
+    evr: 1.5.0-3.fc40
+    metadata:
+      reason: https://bugzilla.redhat.com/show_bug.cgi?id=2266344
```
dustymabe: if you don't mind, let's open an FCOS tracker issue for this and use that link here.

jmarrero:
done via #2884

PR #2883, opened 2024-02-27
[rawhide] overrides: pin authselect-1.5.0-3.fc40

jmarrero:
Same as #2881 but for rawhide. However, no f41 package available of previous version.


dustymabe (approved):
LGTM

PR #2886, opened 2024-02-28
[branched] overrides: fast-track rust-coreos-installer-0.21.0-1.fc40

coreosbot-releng:


aaradhak (approved):
LGTM

PR #2887, opened 2024-02-29
Reapply "manifests: enable cliwrap on Fedora 40+"

jlebon:
This reverts commit 3789b068627af980a3914fbab42153adc2df5815.

The reported issue that motivated the revert[[1]] is due to an rpm-ostree bug[[2]].

We can trivially work around this bug until the rpm-ostree fix lands in FCOS, so let's do that and re-enable cliwrap.

[1]: https://github.com/coreos/fedora-coreos-tracker/issues/1679
[2]: https://github.com/coreos/rpm-ostree/pull/4848

### DIFF
--- a/manifests/cliwrap.yaml
+++ b/manifests/cliwrap.yaml
@@ -1,2 +1,5 @@
 # https://github.com/coreos/fedora-coreos-tracker/issues/730
-cliwrap: true
+cliwrap-binaries:
+  - dnf
+  # until https://github.com/coreos/rpm-ostree/issues/4726
+  - kernel-install
--- a/manifests/cliwrap.yaml
+++ b/manifests/cliwrap.yaml
@@ -1,0 +1,2 @@
+# https://github.com/coreos/fedora-coreos-tracker/issues/730
+cliwrap: true
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -36,6 +36,8 @@ conditional-include:
   - if: releasever == 39
     # Checks for breaking changes that came with Podman v5.
     include: podman-v5.yaml
+  - if: releasever >= 40
+    include: cliwrap.yaml
 
 ostree-layers:
   - overlay/15fcos
--- a/tests/kola/extensions/package
+++ b/tests/kola/extensions/package
@@ -52,3 +52,9 @@ if [[ -n "${failed}" ]]; then
   fatal "could not install: ${failed}"
 fi
 ok "successfully installed os rpm package extensions"
+
+# also try the wrapped dnf
+if jq -e .cliwrap /usr/share/rpm-ostree/treefile.json; then
+  dnf install -y 'ltrace'
+  ok "dnf cliwrap"
+fi
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -87,7 +87,8 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
     version=$(rpm-ostree --version | grep Version)
 cat > Dockerfile << EOF
 FROM localhost/fcos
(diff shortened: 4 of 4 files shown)


dustymabe:
Doing this means we are knowingly shipping a bug to anything on F40+? Normally this might be fine, but we just entered beta freeze which means we potentially ship this in `next`. I’m not super comfortable with that.

I think it would be better to get the fix back ported to rpm-ostree in Fedora and then fast track it to all branches that are on F40+?

dustymabe:
An option here would be to just not workaround the issue in the test and just denylist that test on branched and rawhide. This would protect against us shipping a bug in our prod streams but would set a timer on getting rpm-ostree fixed.

jlebon:
Yes, this needs to be fixed in RHCOS too. I wasn't planning on immediately backporting everywhere in case more bugs show up. But I did plan to eventually make sure it's fixed before f40 GA, whether by backporting or just because we've cut a new release. (Notice I didn't mark this PR as closing https://github.com/coreos/fedora-coreos-tracker/issues/1679 :) ). Not too concerned about leaving this in Beta for now.

Guess we could denylist, yeah... Though kernel replacement in the layering flow is a important feature. I don't want us to lose coverage of that in f40+ in the meantime.

dustymabe (dismissed):
LGTM

cgwalters:
FWIW (as the person who pushed for cliwrap and wrote a lot of the code) I now consider it a technological dead end and also a giant social mistake (effectively agreeing with initial feedback!)

The big technology problem here is that upgrading a package in a container build (whether via rpm or dnf or rpm-ostree if it ever supported `rpm-ostree upgrade <package>` is that the librpm stack will happily blow away the cliwrap entrypoint wrapper.

jlebon:
> FWIW (as the person who pushed for cliwrap and wrote a lot of the code) I now consider it a technological dead end and also a giant social mistake (effectively agreeing with initial feedback!)
> 
> The big technology problem here is that upgrading a package in a container build (whether via rpm or dnf or rpm-ostree if it ever supported `rpm-ostree upgrade <package>` is that the librpm stack will happily blow away the cliwrap entrypoint wrapper.

Yikes, yes that's not great.

I'm still also not a big fan of the approach overall. The main thing I'm interested in really is trying to make `dnf install` work in layered builds. Once we have a Python-free dnf that works well in image mode, I'd rather we ship that in FCOS/RHCOS and back out cliwrap. Hmm, so I think actually this argues for a `cliwrap-dnf: true` to limit the wrapping to just dnf? The upgrade issue shouldn't be an issue there.

jlebon:
Hmm but actually, the confusing thing here is that we definitely do at least need the `kernel-install` wrapper for kernel replacements to work, right? I suppose the plan there is to have the real `kernel-install` be adapted to Just Work but that hasn't happened yet?

I don't see a kernel replacement example in https://gitlab.com/bootc-org/examples and https://github.com/coreos/layering-examples/blob/main/replace-kernel/Containerfile still includes an explicit `RUN rpm-ostree cliwrap install-to-root /`. Should we add `cliwrap-binaries: [kernel-install]` in the meantime in the bootc base image composes?

cgwalters:
>  I suppose the plan there is to have the real kernel-install be adapted to Just Work but that hasn't happened yet?

yeah cc https://github.com/coreos/rpm-ostree/issues/4726

> Should we add cliwrap-binaries: [kernel-install] in the meantime in the bootc base image composes?

Yeah, would probably make sense except do note it comes from systemd and is hence pretty likely to get replaced when upgrading systemd

jlebon:
OK, I changed the proposal here to only wrap `dnf` and `kernel-install`!

> Yeah, would probably make sense except do note it comes from systemd and is hence pretty likely to get replaced when upgrading systemd

Done in https://github.com/CentOS/centos-bootc/pull/377.

jlebon:
Retracting this. Will open a tracker issue with another proposal.

PR #2890, opened 2024-03-08
OCPBUGS-29762: Reverts and further refines  previous kdump fix 

jbtrystram:
See https://github.com/coreos/fedora-coreos-config/pull/2847 
This one was merged with not enough testing : it turns out the kernel arg is only visible to dracut, which turns it into a drop-in file, in `/etc/cmdline.d/60kdumpip.conf`

Adding this as a condition to the sysroot.mount generator makes systemd skip the mount unit in kdump initramfs (only when kdump is configured to export logs to a remote IP)

Ref https://issues.redhat.com/browse/OCPBUGS-29762

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -6,7 +6,17 @@
 # With Ignition v 2.5.0, ignition-dracut was merged into Ignition and the CoreOS
 # specific bits were deposited here.
 
-set -e
+set +euo pipefail
+. /usr/lib/dracut-lib.sh
+set -euo pipefail
+
+# dracut isn't nice with `set -e`
+dracut_func() {
+    set +euo pipefail
+    "$@"; local rc=$?
+    set -euo pipefail
+    return $rc
+}
 
 # Generators don't have logging right now
 # https://github.com/systemd/systemd/issues/15638
@@ -17,7 +27,8 @@ EARLY_DIR="${2:-/tmp}"
 
 IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
 cmdline_arg() {
-    local name="$1" value="$2"
+    local name value
+    name="$1"
     for arg in "${cmdline[@]}"; do
         if [[ "${arg%%=*}" == "${name}" ]]; then
             value="${arg#*=}"
@@ -27,7 +38,8 @@ cmdline_arg() {
 }
 
 cmdline_bool() {
-    local value=$(cmdline_arg "$@")
+    local value
+    value=$(cmdline_arg "$@")
     case "$value" in
         ""|0|no|off) return 1;;
         *) return 0;;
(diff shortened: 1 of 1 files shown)


travier:
Hum, that feels very specific. What if kdump starts writing to another file named differently?

travier:
I'm not sure that I understand why this did not work. Isn't the kernel command line fixed once booted?

jbtrystram:
> I'm not sure that I understand why this did not work. Isn't the kernel command line fixed once booted?

I am not too sure but it looks like some kernel args are only visible by dracut.
When running `kdumpctl rebuild` you can see in the logs : 
`dracut-cmdline[270]: Using kernel command line parameters:  rd.neednet kdump_remote_ip=192.168.122.1`

However, once in the initramfs shell of kdump, I was not able to see the paramater in `/proc/cmdline`. In turns, the systemd generator was not skipping sysroot.mount.

> Hum, that feels very specific. What if kdump starts writing to another file named differently?

We will have to update that. I don't have any other ideas to work around it, but I am not an expert about our set of initramfs stuff

jlebon:
You can source the dracut library and use the dracut helper instead to query the karg, which is `/etc/cmdline.d`-aware. See e.g. https://github.com/coreos/fedora-coreos-config/blob/c6662caac7149597f0e7bfa8d3efb3ae1ee83f48/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.sh

jbtrystram:
@jlebon good tip ! 
I updated the PR using that

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -6,8 +6,20 @@
 # With Ignition v 2.5.0, ignition-dracut was merged into Ignition and the CoreOS
 # specific bits were deposited here.
 
+set +e
+. /usr/lib/dracut-lib.sh
+
 set -e
 
+# dracut isn't nice with `set -e`
```
travier, line 13: This comment is likely for the lines above

travier (commented):
Minor nit but otherwise LGTM. Let's have someone else test the fix?

travier:
Or we could also make a test for it 😉

jbtrystram:
@travier this one would be quite a complicated test setup (it requires an extra container) so i would rather test it manually for now and maybe create an issue to add a test later on. I can create a tracking issue for that, meanwhile we can ship the fix

jbtrystram:
this appears to have a side effect on the iscsi feature, i am looking into it

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -6,6 +6,17 @@
 # With Ignition v 2.5.0, ignition-dracut was merged into Ignition and the CoreOS
 # specific bits were deposited here.
 
+set +e
+. /usr/lib/dracut-lib.sh
```
jlebon: This probably needs to be

```suggestion
set +euo pipefail
. /usr/lib/dracut-lib.sh
set -euo pipefail
```

since IIRC the dracut lib doesn't run with `set -u` so any undefined variable will break us.

jlebon (commented):
Looks sane to me, though for the commit title, maybe let's be more clear. E.g. something like

> 05core: check for `kdump_remote_ip` as a dracut karg

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -6,7 +6,17 @@
 # With Ignition v 2.5.0, ignition-dracut was merged into Ignition and the CoreOS
 # specific bits were deposited here.
 
-set -e
+set +euxo pipefail
+. /usr/lib/dracut-lib.sh
+set -euxo pipefail
```
jlebon: Do we actually want `-x` in production though? I'm guessing that was more for debugging?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -85,7 +100,7 @@ EOF
     add_requires coreos-rootflags.service initrd-root-fs.target
 fi
 
-if ! $(cmdline_bool 'ignition.firstboot' 0); then
+if ! eval "$(cmdline_bool 'ignition.firstboot' 0)"; then
```
jlebon: Yeah, the previous code there looks weird. But confused about the new code too. I think this is just

```suggestion
if ! cmdline_bool 'ignition.firstboot'; then
```

right?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
jbtrystram: yes, it slipped through.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
jbtrystram: That was a shellcheck suggestion :)

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -17,7 +27,8 @@ EARLY_DIR="${2:-/tmp}"
 
 IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
 cmdline_arg() {
-    local name="$1" value="$2"
+    local name value
+    name="$1"
```
jlebon, line 31: I will note this is dropping the "pass in default as second argument" bit, but I understand why you did it (because of `set -u` which wasn't set before). One way to fix this is:

```
local name=$1; shift
local value=${1:-}
```

Though OK with this too (no caller of that function currently uses that feature).

jlebon (approved):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
jbtrystram: I removed it because I saw it was not used, I understand what you mean though.
let's fix it the day we actually need the feature, it shouldn't break existing code then

PR #2891, opened 2024-03-08
overrides: fast-track ignition-2.18.0-1.fc39

coreosbot-releng:
Requested by @yasminvalim via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


jlebon (approved):
Can you also open a fast-track against `branched`? (With the f40 update.)

PR #2892, opened 2024-03-11
tests: use `/dev/disk/by-id/coreos-boot-disk` instead of `/dev/vda`

### LINKED ISSUE #1683: s390x: sgdisk sometimes flakes during disk reprovision tests
When running kola tests on fcos rawhide stream on s390x, sometimes tests are failing with the error:
```
[   24.182309] ignition[950]: disks: createPartitions: op(2): [started]  partitioning "/run/ignition/dev_aliases/dev/vda"

[   24.182339] ignition[950]: disks: createPartitions: op(2): op(3): [started]  reading partition table of "/run/ignition/dev_aliases/dev/vda"

[   24.182761] ignition[950]: disks: createPartitions: op(2): op(3): [finished] reading partition table of "/run/ignition/dev_aliases/dev/vda"

[   24.182807] ignition[950]: disks: createPartitions: op(2): running sgdisk with options: [--pretend --new=4:0:+1048576 --new=5:0:+0 /run/ignition/dev_aliases/dev/vda]

[   24.185900] ignition[950]: disks: createPartitions: op(2): [failed]   partitioning "/run/ignition/dev_aliases/dev/vda": Failed to pretend to create partitions. Err: exit status 2. Stderr:
```
Last seen in https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/build-arch/3589, during `ext.config.var-mount.simple`

[ext.config.var-mount.simple.zip](https://github.com/coreos/fedora-coreos-tracker/files/14556007/ext.config.var-mount.simple.zip)

jlebon:
Hardcoding `/dev/vda` will break if there are more than one block device may be attached (like on s390x where we attach another disk to pass in Ignition). This is why the `/dev/disk/by-id/coreos-boot-disk` symlink was added. Use it.

While we're here, switch from `platform: qemu` to
`tags: platform-independent` since they should theoretically work on any platform now (though they are equivalent today).

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1683

### DIFF
--- a/tests/kola/root-reprovision/swap-before-root/config.bu
+++ b/tests/kola/root-reprovision/swap-before-root/config.bu
@@ -3,7 +3,7 @@ version: 1.4.0
 
 storage:
   disks:
-    - device: /dev/vda
+    - device: /dev/disk/by-id/coreos-boot-disk
       partitions:
         - number: 4
           label: swap
--- a/tests/kola/root-reprovision/swap-before-root/test.sh
+++ b/tests/kola/root-reprovision/swap-before-root/test.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 ## kola:
-##   # This test's config manually references /dev/vda and is thus QEMU only
-##   platforms: qemu
 ##   # Root reprovisioning requires at least 4GiB of memory.
 ##   minMemory: 4096
 ##   # This test includes a lot of disk I/O and needs a higher
@@ -11,7 +9,7 @@
 ##   # legal but usually not intended, so Butane warns about it.
 ##   allowConfigWarnings: true
 ##   # This test reprovisions the rootfs.
-##   tags: reprovision
+##   tags: reprovision platform-independent
 ##   description: Verify the root reprovision and swap enabled are supported.
 
 set -xeuo pipefail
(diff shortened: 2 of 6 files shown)


jschintag:
I haven't managed to really test it, since i choose the wrong machine, which is no really suited to running kola tests...
It is still running since this morning, so about 7h and i can't tell if it is failing or timeouts...
I just started a build + kola on another machine, but i guess i can only tell you the result tomorrow, since i'll be heading out shortly.

jschintag (approved):
I have tested it and it looks good. Given that https://github.com/coreos/fedora-coreos-tracker/issues/1683 is flaky, i'm not sure if it solves it.
Especially since the tests only use 1 disk as far as i can see. But consequently using disks by label instead of instable letters is a good thing anyway, so i'm down for checking if this helps.

LGTM

jlebon:
> Especially since the tests only use 1 disk as far as i can see.

I think the issue actually is that on s390x and ppc64le we do pass in another disk for the Ignition config, and that's where the flakiness is coming from depending on which one gets `/dev/vda`.

jschintag:
> > Especially since the tests only use 1 disk as far as i can see.
> 
> I think the issue actually is that on s390x and ppc64le we do pass in another disk for the Ignition config, and that's where the flakiness is coming from depending on which one gets `/dev/vda`.

Right, the ignition is a file mounted as a disk, there was something...

PR #2896, opened 2024-03-12
tests/bootupd: check for `/boot/bootupd-state.json`

jlebon:
`bootupctl status` will still be happy even if that file doesn't exist, so let's check for it explicitly to make sure that `bootupctl backend install` ran correctly at disk image generation time.


travier (approved):

marmijo (approved):
LGTM

PR #2899, opened 2024-03-14
tests/bootupd: don't check for `/boot/bootupd-state.json` on s390x

jlebon:
On x86_64 and aarch64, we always expect the file to exist. On ppc64le, we only expect it to exist if it was built using osbuild. On s390x, we never expect it to exist.

Fixes 10d5f64d ("tests/bootupd: check for `/boot/bootupd-state.json`").


dustymabe (approved):
LGTM

PR #2900, opened 2024-03-14
manifest: Unpin ostree

cgwalters:
It seems like something that was intended to unpin `fast-track` overrides isn't.


dustymabe:
It's a tooling gap. Our graduated overrides checker doesn't work for arch specific overrides.

Normally this gets cleaned up because weekly we have a "health sync" meeting where we review existing overrides and denylist entries. Not sure why that didn't get caught this time.

manifest-lock.overrides.yaml
dustymabe, line 1: I think we should keep this file. It already just had an empty package list. I'll add it back and merge this.

dustymabe (approved):

PR #2902, opened 2024-03-15
[rawhide] overrides: drop authselect pin; bump aarch64 kernel

dustymabe:


jlebon (dismissed):

dustymabe:
ok added a kernel pin for all here because of https://github.com/coreos/fedora-coreos-tracker/issues/1693

jlebon (approved):

PR #2903, opened 2024-03-15
manifests: add makedumpfile RPM on F41+

dustymabe:
This was made into a subpackage, but is required for kdump to work. See https://github.com/coreos/fedora-coreos-tracker/issues/1692

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -36,6 +36,9 @@ conditional-include:
   - if: releasever == 39
     # Checks for breaking changes that came with Podman v5.
     include: podman-v5.yaml
+  - if: releasever >= 41
+    # Include makedumpfile subpackage from kexec-tools (new in F41+)
+    include: makedumpfile.yaml
 
 ostree-layers:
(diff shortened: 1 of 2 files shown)


manifests/fedora-coreos.yaml
```diff
@@ -36,6 +36,9 @@ conditional-include:
   - if: releasever == 39
     # Checks for breaking changes that came with Podman v5.
     include: podman-v5.yaml
+  - if: releasever >= 40
```
jlebon:
```suggestion
  - if: releasever >= 41
```

dustymabe: fixed

jlebon (approved):
Hmm, weird. It feels like ideally we split out the `kdump` bits into its own subpackage and then have that hard require `makedumpfile`?

dustymabe:
> Hmm, weird. It feels like ideally we split out the `kdump` bits into its own subpackage and then have that hard require `makedumpfile`?

Yeah. I tagged in Carl in our matrix channel to see if there was any more context. Let's see what he says.

PR #2904, opened 2024-03-18
[rawhide] overrides: pin to shim-15.6-2

dustymabe:
The new shim-15.8-2 seems to not work for secure boot: https://github.com/coreos/fedora-coreos-tracker/issues/1694


aaradhak (approved):
lgtm

dustymabe:
The test failure here was `ext.config.var-mount.scsi-id` which is marked as `warn: true`. Merging!

PR #2906, opened 2024-03-18
[rawhide] overrides: fast-track passt; unpin kernel

dustymabe:
The new passt release allows us to unpin the kernel and still have all podman tests pass. See
https://github.com/coreos/fedora-coreos-tracker/issues/1693#issuecomment-2004260760


aaradhak (approved):
LGTM

dustymabe:
The CI failure here is a warn: true test so we can ignore the failure. All else passs.

PR #2901, opened 2024-03-15
Updating next-devel manifest for f40

aaradhak:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1674

PR #2907, opened 2024-03-18
denylist: snooze ext.config.var-mount.scsi-id on next & next-devel

aaradhak:
Let's also snooze `ext.config.var-mount.scsi-id` on `next` & `next-devel` since it's been failing on F40.
Ref: https://github.com/coreos/fedora-coreos-config/pull/2901


dustymabe (approved):
LGTM

PR #2911, opened 2024-03-20
manifests: remove ostree-bls.yaml

dustymabe:
`bootloader-naming-2` was turned on by default in ostree v2024.5 [1] so keeping it here isn't necessary any longer.

[1] https://github.com/ostreedev/ostree/releases/tag/v2024.5


dustymabe:
so one thing we have right now is [ostree-2024.5-1.fc39](https://bodhi.fedoraproject.org/updates/FEDORA-2024-ac4e52fdda) in COSA but not in any of our streams yet (we could change this with a fast-track), so it begs the question:

if the disk images get deployed with say:

```
[bound] -bash-5.2$ ls /boot/loader/entries/
ostree-1.conf
```

but then on upgrade they get:

```
Bootloader updated; bootconfig swap: yes; bootversion: boot.0.1, deployment count change: 1
[bound] -bash-5.2$ ls /boot/loader/entries/
ostree-1-fedora-coreos.conf  ostree-2-fedora-coreos.conf
```

is that OK?

i.e. are these file names enough of an internal detail that it doesn't matter if they flip flop a few times?

cgwalters:
Yes, it's OK for them to go back.  We'll eventually reconcile which is the important part.

jlebon (approved):

PR #2915, opened 2024-03-22
Add `dnf5` on Fedora 41+

jlebon:
As per https://github.com/coreos/fedora-coreos-tracker/issues/1687, we will for now add `dnf5` to the rawhide stream to test it out and re-evaluate at branching.

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -36,6 +36,13 @@ conditional-include:
   - if: releasever >= 41
     # Include makedumpfile subpackage from kexec-tools (new in F41+)
     include: makedumpfile.yaml
+  # On <41, we want to keep making sure dnf doesn't slip in somehow
+  # On 41+, we do want it
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1687
+  - if: releasever < 41
+    include: exclude-dnf.yaml
+  - if: releasever >= 41
+    include: include-dnf.yaml
 
 ostree-layers:
   - overlay/15fcos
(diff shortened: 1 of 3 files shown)


jlebon:
It'd be nice to add a test for this where we do a layered build with `dnf5` and then reboot into it, but currently there's no way to limit a test to a specific stream. Though I guess we could denylist it on all the other streams for now?

jlebon:
Though I did manually test this (see https://github.com/coreos/fedora-coreos-tracker/issues/1687#issuecomment-2015511773).

manifests/fedora-coreos.yaml
```diff
@@ -36,6 +36,13 @@ conditional-include:
   - if: releasever >= 41
     # Include makedumpfile subpackage from kexec-tools (new in F41+)
     include: makedumpfile.yaml
+  # On <41, we want to keep making sure dnf doesn't slip in somehow
+  # On 41+, we do want it
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1687
+  - if: releasever < 41
+    include: exclude-dnf.yaml
+  - if: releasever >= 41
+    include: include-dnf.yaml
```
dustymabe, line 45: I'm guessing there is no `else` directive?

jlebon: Heh, I actually had started looking at that, and then got distracted by https://github.com/coreos/rpm-ostree/issues/3729 instead which would also be really be useful. And then I stopped and just got this out. :)

dustymabe: yep :) - every time I mess with this stuff I long for https://github.com/coreos/rpm-ostree/issues/3729

dustymabe (approved):
LGTM

PR #2919, opened 2024-03-23
[next] tree: promote changes from next-devel at af418566f1a8125a6a651a2ae2a8d5713974b666

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/869.


manifest.yaml
```diff
@@ -2,6 +2,6 @@ variables:
   stream: next
   prod: true
 
-releasever: 39
+releasever: 40
```
dustymabe, line 5: we can ignore the warning here - this change was intended

jlebon (approved):

PR #2921, opened 2024-03-25
[next-devel] overrides: drop graduated arch specific overrides

dustymabe:
The shim update has been tagged into all Fedora releases and the others are now in stable repos, but the graduated overrides action doesn't know about arch specific overrides so I'll manually clean them up here.


jlebon (approved):

manifest-lock.overrides.aarch64.yaml
jbtrystram, line 1: why delete the whole file rather than just the values, keeping the comments at the top ?  I feel like the next person needed to pin a package would have benefited from the template file

dustymabe: It's just what we have been doing. For arch specific overrides files we delete them (less clutter in the directory) and then pull the template from git history when we re-instate.

Ironically for the main overrides yaml file when it is empty we don't delete it :).

I think the current status quo makes it easier to see what overrides there are. If we did it the other way we'd always have to check 5 files to see that. Currently often we only have to check 1, sometimes more, but almost always less than 5.

I'm happy with changing the approach, though. I appreciate you speaking up!

jbtrystram: It was more a question than disagreement :)
Keeping the main override file achieves the same goal so it works

PR #2922, opened 2024-03-26
[rawhide] denylist: add ext.config.dump.crash

jbtrystram:
~~Kdump currently fails on kernel-next. This have been fixed in kexec-tools-2.0.28-6.fc41~~

~~https://bodhi.fedoraproject.org/updates/FEDORA-2024-0582ec516b~~
~~It looks like this package should be in rawhide already, not sure if this PR is necessary really ?~~

Kdump currently fails on kernel-next, for ppc64le
This is a known upstream bug
See coreos/fedora-coreos-tracker#1698
Also see upstream bug : https://bugzilla.redhat.com/show_bug.cgi?id=2269991

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -20,3 +20,9 @@
     - branched
     - next
     - next-devel
+- pattern: ext.config.kdump.crash
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1698 
+  snooze: 2024-04-07
+  warn: true
+  streams:
+    - rawhide


aaradhak (dismissed):
LGTM

dustymabe:
let me look

dustymabe:
`kexec-tools-2.0.28-6.fc41.x86_64` is in rawhide now (obviously x86_64).

If I look at the [most recent ppc64le run](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/build-arch/detail/build-arch/3748/pipeline/187) (which failed) I do see `kexec-tools-2.0.28-6.fc41.ppc64le` was built into the OSTree.

Now that I look https://bugzilla.redhat.com/show_bug.cgi?id=2269991 if you look at comment 4 you'll see that this is a different bug than the commenter in comment 2 thought it was. This bug isn't fixed in any version in rawhide yet.

Can we update the commit message here?

kola-denylist.yaml
```diff
@@ -20,3 +20,9 @@
     - branched
     - next
     - next-devel
+- pattern: ext.config.kdump.crash
+  tracker: https://bugzilla.redhat.com/show_bug.cgi?id=2269991
```
dustymabe: this should point to https://github.com/coreos/fedora-coreos-tracker/issues/1698

jbtrystram: Thanks for digging a bit more than me Dusty ! 
I updated the commit message

dustymabe:
CI won't really test this change.. Merging

PR #2925, opened 2024-03-26
[branched] overrides: drop all overrides

dustymabe:
The branched stream is disabled now so let's wipe it of any overrides that will surely be stale once it is reinstated.


aaradhak (approved):
LGTM

PR #2927, opened 2024-03-28
tests: check default dropins file permissions

HuijingHei:
Add check in existing case, xerf to https://github.com/coreos/ignition/issues/1833

### DIFF
--- a/tests/kola/ignition/delete-config/test.sh
+++ b/tests/kola/ignition/delete-config/test.sh
@@ -29,6 +29,10 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
 "")
     # Ignition boot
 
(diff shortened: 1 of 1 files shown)


jbtrystram (commented):
I would create another test case ideally. Because having "ignition-delet-config" fail to a permission error might be confusing. 
I don't know if that's possible to have the test in the ignition test suite/repo ? 
But no hard opinions, i am fine merging that as well

dustymabe:
> I would create another test case ideally.

IIUC it could easily be a non-exclusive test too. so it can standalone, but also not cause an extra VM to be spun up.

jbtrystram (approved):

HuijingHei:
Thanks @jbtrystram @dustymabe for the review, both comments make sense to me, thanks!

PR #2928, opened 2024-03-28
denylist: only warn if `coreos.boot-mirror` tests fail on ppc64le

jlebon:
These tests have been flaky on ppc64le for a while. Let's just mark them as warn-only until we can actually look more into what's going on there.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -20,3 +20,6 @@
     - branched
     - next
     - next-devel
+- pattern: coreos.boot-mirror*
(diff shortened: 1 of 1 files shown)


aaradhak (approved):
LGTM

dustymabe:
I think my only concern here is marking it as warn for all streams. Should we conditionalize it? Have we noticed it failing on one stream more than others or is it failing on all of them?

jlebon:
I've seen it fail on every Fedora release.

PR #2930, opened 2024-03-28
[next-devel]: fast-track downgraded packages

marmijo:
F40 is now in final freeze. This means some packages in F39 will sort as newer than packages in F40. We'll prevent downgrades by fast-tracking any packages that would violoate this "no downgrade" rule.

Today they are: `gnutls`

See: https://github.com/coreos/fedora-coreos-tracker/issues/1674#issuecomment-2010200582


aaradhak (approved):
LGTM

dustymabe (approved):
LGTM

PR #2931, opened 2024-03-29
[rawhide] denylist: add ext.config.kdump.crash

aaradhak:
kdump currently fails on kernel-next for ppc64le
This is a known upstream bug
Ref: coreos/fedora-coreos-tracker#1698
Also see upstream bug : https://bugzilla.redhat.com/show_bug.cgi?id=2269991

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -23,3 +23,9 @@
 - pattern: coreos.boot-mirror*
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1659
   warn: true
+- pattern: ext.config.kdump.crash
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1698
+  snooze: 2024-04-07
+  warn: true
+  streams:
+    - rawhide


marmijo (approved):
LGTM

jbtrystram (commented):
Just so I learn, what's the difference with https://github.com/coreos/fedora-coreos-config/commit/10a96490a3b1b0178e950cab98e41dcab2f24325
I think it's mostly the same patch, but i did it on the `rawhide` branch. Did i made a mistake?

dustymabe:
> Just so I learn, what's the difference with [10a9649](https://github.com/coreos/fedora-coreos-config/commit/10a96490a3b1b0178e950cab98e41dcab2f24325) I think it's mostly the same patch, but i did it on the `rawhide` branch. Did i made a mistake?

It's subtle. Basically we have a config-bot that syncs almost all files in the branches of this repo from `testing-devel` into the other branches. The exceptions are a few files like the lockfiles, `manifest.yaml`, and `image.yaml` (IIRC). Hence we've chosen to share the denylist across all development/mechanical branches to minimize maintenance. Your PR to rawhide eventually got overwritten by config-bot, but this PR will persist.

jbtrystram:
@dustymabe thanks for the explanation, I wasn't aware of that. So I should have opened the PR against `testing-devel` in the first place. Good to know !

PR #2933, opened 2024-04-01
[next] tree: promote changes from next-devel at 781a9bba27b52a79979d7cb6f42df381a6cf32ed

coreosbot-releng:
Triggered by @marmijo in https://github.com/coreos/fedora-coreos-streams/issues/872.


kola-denylist.yaml
```diff
@@ -20,3 +20,12 @@
     - branched
     - next
     - next-devel
+- pattern: coreos.boot-mirror*
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1659
+  # warn: true (disabled on promotion)
```
dustymabe, line 25: this is purely commentary:

The intent of commenting these out was to make the "snooze" lines not have any effect (i.e. disallowing them to expire) on production streams, but I think in the case where we encounter a warn with no snooze we should leave it in the future.

Now, warn with no snooze isn't really something we've commonly done, but in this case we did do it on purpose.

Might PR this if I get a chance or someone else feel free.

dustymabe (approved):
LGTM

marmijo (approved):

PR #2936, opened 2024-04-01
[rawhide] overrides: bump aarch64 kernel pin

marmijo:
Let's bump to the latest rc kernel in aarch64 while we wait for a fix for https://github.com/coreos/fedora-coreos-tracker/issues/1647


dustymabe (approved):

dustymabe:
CI won't test this since it is a change for aarch64.

PR #2937, opened 2024-04-03
Manifests: move zram to system-configuration

jbtrystram:
Kubernetes 1.28 now supports swap on zram as beta, so we need to ship this package in RHCOS.
Moving zram to a separate manifest that could be optionnaly enabled for k8s use ?

See https://github.com/coreos/fedora-coreos-docs/issues/530

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -150,9 +150,6 @@ packages:
   - console-login-helper-messages-motdgen
   # i18n
   - kbd
-  # zram-generator (but not zram-generator-defaults) for F33 change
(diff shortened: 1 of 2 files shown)


dustymabe:
If we need it in both RHCOS and FCOS then maybe just move it into `manifests/system-configuration.yaml`. It will automatically flow into RHCOS then.

In the commit message you can mention why (i.e. new kube support for swap).

travier:
It's also useful for non-k8s deployments so I'd would also vote for not adding a k8s specific manifest.

travier (approved):

dustymabe (approved):
LGTM

PR #2938, opened 2024-04-03
manifests: Add back more wireless card firmwares

travier:
Temporarily add back more wireless card firmwares until we announce their removal and add a CLHM module to warn users about it.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1575


dustymabe (approved):

jlebon (approved):

jlebon:
You'll have to add them to the lockfiles as well.

dustymabe:
> You'll have to add them to the lockfiles as well.

they are already there in F39 since they are a requires in F39, but recommends in F40

jlebon:
> > You'll have to add them to the lockfiles as well.
> 
> they are already there in F39 since they are a requires in F39, but recommends in F40

Ahh right. So `next-devel` will fail to build until the next bump-lockfile run. Let's schedule one once this merges.

travier:
Do I need to make a PR for next as well?

dustymabe:
> Do I need to make a PR for next as well?

no. the bump-lockfile job will take care of it.

PR #2940, opened 2024-04-03
Misc: add a small tool to graph manifest includes

jbtrystram:
This is a simple script that display a tree of manifest includes. I wrote it as a quick tool to get an idea of how manifest were structured.
Requires the `anytree` python lib.
Simply pass the manifest you want to start from as an argument.
Sample output on RHCOS:
```
(L)manifest.yaml
└── (L)manifest-rhel-coreos-9.yaml
    └── manifest-rhel-9.2.yaml
        ├── common.yaml
        │   ├── fedora-coreos-config/manifests/system-configuration.yaml
        │   ├── fedora-coreos-config/manifests/ignition-and-ostree.yaml
        │   │   └── fedora-coreos-config/manifests/bootable-rpm-ostree.yaml
        │   ├── fedora-coreos-config/manifests/file-transfer.yaml
        │   ├── fedora-coreos-config/manifests/networking-tools.yaml
        │   ├── fedora-coreos-config/manifests/user-experience.yaml
        │   ├── fedora-coreos-config/manifests/shared-workarounds.yaml
        │   └── packages-rhcos.yaml
        └── packages-openshift.yaml
```

Agreed I would have spent less time doing it by hand, but it's not every day you get to write some recursive code ! 
I thought I'd share if that can be helpful to someone else. 
It's probably not the best place for it, let me know  !

### DIFF
--- a/manifest_graph
+++ b/manifest_graph
@@ -1,0 +1,60 @@
+#!/usr/bin/python3 -u
+
+'''
+    Create a tree chart with all the manifest includes
+    symlinks are displayed with a (L) prefix
+'''
+
+import sys
+import yaml
+import os
+from anytree import Node, RenderTree
+
+def collect_included_files(filename, parent):
+    node = None
+
+    if os.path.islink(filename):
+      node = Node("(L)"+filename, parent=parent)
+      dest = os.readlink(filename)
+      collect_included_files(dest, parent=node)
+      return node
+    else:
+      node = Node(filename, parent=parent)
+
+    with open(filename, 'r') as file:
+        try:
+            data = yaml.safe_load(file)
+            if data is None:
+                return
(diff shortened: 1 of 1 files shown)


dustymabe:
This actually looks pretty cool.

Maybe it would best fit in the COSA repo? or maybe rpm-ostree somewhere. Not sure.

mike-nguyen:
This would have been useful a few weeks ago when I was trying to track down where a package was being added!

I agree with Dusty that maybe rpm-ostree or cosa would be good places so we can use it in RHCOS or FCOS.

c4rt0:
I'm not sure what is the best place for it, but I agree with Dusty, this looks cool.

jbtrystram:
I'll move this to cosa

PR #2941, opened 2024-04-04
[rawhide] overrides: pin google-compute-engine-guest-configs-udev

dustymabe:
The package was retired but if we pin it we can still pull it from the coreos-pool repo for FCOS in rawhide for now while we try to find a new maintainer.

Follow https://github.com/coreos/fedora-coreos-tracker/issues/1702


jbtrystram (approved):
Lgtm

c4rt0 (approved):
LGTM

PR #2943, opened 2024-04-05
ext-tests/ext Fix var-mount/scsi-id failure on rawhide

jbtrystram:
Update the scsci-id test to set a WWN for the disk, and use reliable udev symlinks to adjust for a change in sg3_utils [1]

Note that the `wwn` value set is converted to base 16 by QEMU, so the symlink in the ignition config must reflects it.

This requires https://github.com/coreos/coreos-assembler/pull/3772 See https://github.com/coreos/fedora-coreos-tracker/issues/1670

[1] https://listman.redhat.com/archives/dm-devel/2023-March/053645.html

### DIFF
--- a/tests/kola/var-mount/scsi-id/config.bu
+++ b/tests/kola/var-mount/scsi-id/config.bu
@@ -2,7 +2,8 @@ variant: fcos
 version: 1.4.0
 storage:
   disks:
-    - device: /dev/disk/by-id/scsi-0NVME_VirtualMultipath_disk1
+    # the symlink is from wwn=123456789 converted to hexadecimal (base16)
+    - device: /dev/disk/by-id/scsi-300000000075bcd15
       wipe_table: true
       partitions:
         - number: 1
--- a/tests/kola/var-mount/scsi-id/test.sh
+++ b/tests/kola/var-mount/scsi-id/test.sh
@@ -2,7 +2,7 @@
 ## kola:
 ##   # additionalDisks is only supported on QEMU
 ##   platforms: qemu
-##   additionalDisks: ["5G:mpath"]
+##   additionalDisks: ["5G:channel=scsi,wwn=123456789"]
 ##   description: Verify udev rules /dev/disk/by-id/scsi-* symlinks exist
 ##     in initramfs.
 


tests/kola/var-mount/scsi-id/config.bu
```diff
@@ -2,7 +2,7 @@ variant: fcos
 version: 1.4.0
 storage:
   disks:
-    - device: /dev/disk/by-id/scsi-0NVME_VirtualMultipath_disk1
+    - device: /dev/disk/by-id/scsi-300000000075bcd15
```
dustymabe, line 6: Since this is yaml we can add a comment about the base16 conversion:

```suggestion
    # the symlink is wwn=123456789 converted to hexadecimal (base16)
    - device: /dev/disk/by-id/scsi-300000000075bcd15
```


Ok - another comment:

```
[dustymabe@media ~]$ python3 -c 'print(0x75bcd15)'
123456789
```

Where does the `3` on the front of `300000000075bcd15` come from?

jbtrystram: @dustymabe I believe the `3` comes from [here ](https://github.com/doug-gilbert/sg3_utils/blob/2355dc4b451989291df695148cd8d8d03b3d987e/scripts/58-scsi-sg3_symlink.rules#L34)

Another way would be to set some `kargs` for a custom symlink, as [explained here](https://github.com/doug-gilbert/sg3_utils/pull/42) but that sounds even more complicated

Also, see this [comment](https://github.com/coreos/coreos-assembler/pull/3772#discussion_r1553646402) for more details about the base16.

jbtrystram:
@dustymabe you were so quick to review I could not post my follow-up comment :) 

Let's hold this for now : https://github.com/coreos/coreos-assembler/pull/3772#issuecomment-2040052157

EDIT: now it's fixed :)

dustymabe (approved):
LGTM

c4rt0 (commented):
Even though this LGTM, I have no idea how the hardcoded `wwn` value works in `additionalDisks: ["5G:channel=scsi,wwn=123456789"]` - is it a templating thing?

c4rt0 (approved):

dustymabe:
> Even though this LGTM, I have no idea how the hardcoded `wwn` value works in `additionalDisks: ["5G:channel=scsi,wwn=123456789"]` - is it a templating thing?

That value is one we add so that we can ID the related block device inside the VM. My understanding is that in this case the driver takes the decimal value and converts it to hexadecimal underneath, so the symlink that shows up in the VM is hexadecimal.

jbtrystram:
@c4rt0 looks at https://github.com/coreos/coreos-assembler/pull/3772 for how the `wwn` option is passed to qemu. 
As dusty explains, the value nd up being converted to hex, i don't know why. 
I only found very few documentation around that wwn option.

PR #2945, opened 2024-04-06
overrides: drop kernel pin for aarch64

dustymabe:
Looks like there was a fix for the issue we were seeing sometime early in kernel 6.9 development: https://github.com/coreos/fedora-coreos-tracker/issues/1647#issuecomment-2040944967

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1647


jbtrystram (approved):
Lgtm

PR #2948, opened 2024-04-09
tests: Verify access permissions for shadow & gshadow

travier:
Merge with the existing passwd & group check and also check backup files.

See: https://github.com/coreos/rpm-ostree/pull/4911
See: https://github.com/coreos/rpm-ostree/security/advisories/GHSA-2m76-cwhg-7wv6
See: https://github.com/coreos/fedora-coreos-tracker/issues/1705

---

Waiting on an updated rpm-ostree version to land in COSA.

PR #2951, opened 2024-04-09
[stable] overrides: fast-track rpm-ostree-2024.4-6.fc39

coreosbot-releng:


dustymabe:
Added the test from #2948 here too.

dustymabe (approved):
LGTM

PR #2954, opened 2024-04-09
image-base: drop bootfs_metadata_csum_seed entry

dustymabe:
This is no longer a supported feature in COSA. See https://github.com/coreos/coreos-assembler/commit/383349c7791c483d344647f01b116e160cab0f5f


jmarrero (approved):
lgtm

PR #2956, opened 2024-04-10
Manifests: disable default ZRAM config

jbtrystram:
In https://github.com/coreos/fedora-coreos-config/pull/2937 we moved zram-generator to the system-configuration.yaml manifest so it's shared with RHCOS.
However in the centOS stream 9 zram-generator-0.3.2 [1] ships a default config file, resulting in a swap device being created by default.
This ends up breaking kubelet, until kubernetes 1.30 is released

[1] https://gitlab.com/redhat/centos-stream/rpms/rust-zram-generator/-/blob/c9s/zram-generator.conf?ref_type=heads

### DIFF
--- a/manifests/system-configuration.yaml
+++ b/manifests/system-configuration.yaml
@@ -54,3 +54,8 @@ postprocess:
   - |
     #!/usr/bin/env bash
     systemctl mask systemd-repart.service
+  # zram-generator-0.3.2 (shipped in centOS 9) provides a default
+  # zram-generator config, we want to disable it
+  - |
+    #!/usr/bin/env bash
+    rm -f /usr/lib/systemd/zram-generator.conf


travier:
This should probably go into openshift/os, or maybe we should update the package in RHEL as well

dustymabe:
In centos stream 10 I think the default config is in a subpackage like in Fedora (which makes sense): https://gitlab.com/redhat/centos-stream/rpms/rust-zram-generator/-/blob/c10s/rust-zram-generator.spec?ref_type=heads#L69

I think we should move this removal into a manifest in `openshift/os` and make it conditional on el9 (c9s or rhel9).

travier:
You can use the `remove-from-package` directive from rpm-ostree: https://coreos.github.io/rpm-ostree/treefile/

jbtrystram:
Thanks @jlebon @dustymabe 
I was not sure if the `postprocess` step needs to be in the same file as the package, i don't know if there is some ordering or not. Based on your answers I guess no. I'll push a separate manifest in rhcos

travier:
The manifest is first "resolved" (i.e. all includes imported, duplicated fields merged, etc.) and then processed by rpm-ostree so the "order" *generally* does not matter.

jbtrystram:
Moved to https://github.com/openshift/os/pull/1479

PR #2961, opened 2024-04-11
tests/manual: add in coreos-builds-bisect script

dustymabe:
This script will allow us to perform bisection on our list of builds for our CoreOS streams. It will enable us to find issues faster. It was inspired by rpm-ostree-bisect [1], which does somethign similar, but operates on the OSTree repo from inside the OS.

Example output looks something like:

```
+ echo 'Could not access /etc/shadow -> SUCCESS'
Could not access /etc/shadow -> SUCCESS
+ exit 0
38.20230819.3.0 passed
BISECT TEST RESULTS:
Last known good build: 38.20230819.3.0
First known bad build: 38.20230902.3.0
```

[1] https://github.com/ostreedev/ostree-releng-scripts/blob/master/rpm-ostree-bisecthis

### DIFF
--- a/tests/manual/coreos-builds-bisect.py
+++ b/tests/manual/coreos-builds-bisect.py
@@ -1,0 +1,303 @@
+#!/usr/bin/python3
+#
+# Copyright 2024 Dusty Mabe <dusty@dustymabe.com>
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU Lesser General Public
+#  License as published by the Free Software Foundation; either
+#  version 2 of the License, or (at your option) any later version.
+#
+#  This library is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+#  Lesser General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


jlebon (approved):
I didn't review the script in depth, but it looks sane to me!

Though I wonder if it should be in e.g. `tests/tools/` or `tests/scripts/` instead? All the other files in `tests/manual/` currently are actual scripts that test specific features and so are accurately described by the directory name. This seems like a more generic tool?

dustymabe:
> Though I wonder if it should be in e.g. `tests/tools/` or `tests/scripts/` instead? All the other files in `tests/manual/` currently are actual scripts that test specific features and so are accurately described by the directory name. This seems like a more generic tool?

I don't have a strong opinion there. Mostly the things in here are here because we didn't have a better place to put them and you generally know what you are looking for if you get down this far..

With as few of these that there are I'm not sure making other directories will add value (i.e. compared to the more searching) until we get into maybe 10s of files?

TL;DR Want it in a `scripts/` dir? I'll put it there if you think it's worth it.

jlebon:
> TL;DR Want it in a `scripts/` dir? I'll put it there if you think it's worth it.

Actually yeah, `scripts/` is even more appropriate. But no strong opinion either if you want to keep it there.

dustymabe:
I'll keep it where it is for now just to save a CI cycle. Thanks for the review!

PR #2963, opened 2024-04-12
Add missing wifi firmware checks

jbtrystram:
Some wifi-firmwares will be split into sub-packages in fedora 40 We will keep them in until fedora 41, but display a warning message in the console if NetworkManager-wifi is layered without the most popular the wifi firmwares.

See https://github.com/coreos/fedora-coreos-tracker/issues/1575

---
This is an initial draft to get some discussion. 
it's probably worth moving to a separate manifest and having it be a conditional include on `releasever < 41`


overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,45 @@
...
+# and provide remediation steps
+# See https://github.com/coreos/fedora-coreos-tracker/issues/1575
+
+layered_packages="$(rpm-ostree status --json -b | jq -r '.deployments[0].packages[]')"
+
+if echo "$layered_packages" | grep -q "NetworkManager-wifi"; then
+  if echo "$layered_packages" | grep -q "atheros-firmware"  && \
+     echo "$layered_packages" | grep -q "brcmfmac-firmware" && \
+     echo "$layered_packages" | grep -q "mt7xxx-firmware"   && \
+     echo "$layered_packages" | grep -q "realtek-firmware"
```
dustymabe: We are missing the two new ones here. Suggestion for improvement:

```suggestion

firmwares=(
    atheros-firmware
    brcmfmac-firmware
    mt7xxx-firmware
    nxpwireless-firmware
    realtek-firmware
    tiwilink-firmware
)
regex=$(IFS='|'; echo "${firmwares[*]}")

if grep -q "NetworkManager-wifi" <<< "$layered_packages"; then
    if grep -qP $regex <<< "$layered_packages"; then
```

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,45 @@
...
+# No color
+nc=$(echo -e '\033[0m')
+
+motd_path=/run/motd.d/30_wireless_firmwares_warning.motd
+
+cat << EOF > "${motd_path}"
+${warn}
+##########################################################################
+WARNING: The NetworkManager-wifi is installed on this system.
+However, some wifi-firmwares are not installed.
```
dustymabe:
```suggestion
WARNING: The NetworkManager-wifi RPM is a requested layered package
on this system, but no wifi drivers are requested. The wifi drivers
will be dropped from FCOS at a future date.
```

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,45 @@
...
+
+To layer the missing firmwares use the following :
+sudo rpm-ostree install atheros-firmware brcmfmac-firmware \
+mt7xxx-firmware realtek-firmware
```
dustymabe: Rather than tell them what to do here I think it would be better to link to either a comment in the GitHub issue OR (even better if you are up for it) a FAQ entry with more context. This would allow us to update it anytime with more information or better steps.

For example, the easy thing to do is to layer all wireless drivers, but it's not the most optimal. Installing a specific driver is more optimal, but the hard part is knowing which one. We should mention this as an option.

dustymabe:
one other thing we should do is include other wifi drivers (i.e. ones not explicitly included in the f39 fcos base layer today) in this check. i.e. if people have `NetworkManager-wifi` layered and `iwl7260-firmware` (one of the intel wireless ones) then we shouldn't warn them they are most likely covered.

travier:
Unfortunately I don't  know of a good way to get a full list of the wifi firmware RPMs in Fedora. I think it's mostly up to the maintainer of the linux-firmware package in Fedora which I think is Peter Ronbinson.

travier:
CC @nullr0ute

dustymabe:
> Unfortunately I don't know of a good way to get a full list of the wifi firmware RPMs in Fedora.

That's OK we can just do best effort here.

nullr0ute:
> Unfortunately I don't know of a good way to get a full list of the wifi firmware RPMs in Fedora. I think it's mostly up to the maintainer of the linux-firmware package in Fedora which I think is Peter Ronbinson.

I'm not sure what you mean with this statement, not sure if you would like something from me? Can you clarify pls?

nullr0ute:
> Some wifi-firmwares will be split into sub-packages in fedora 40.

From memory most of the WiFi firmware were split out in F-39, for stable releases we retain a hard Requires dep to pull them in, for the not stable releases that becomes a Recommends to enable those that don't want them to remove them (they are quite large). They are added to comps so if you're pulling in HW support group they should come along automatically.

dustymabe:
> They are added to comps so if you're pulling in HW support group they should come along automatically.

good point...

we can probably just pull the wifi related drivers from the [list in comps](https://pagure.io/fedora-comps/blob/main/f/comps-f41.xml.in#_2700) and that should be good enough.

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
jbtrystram: Thanks for the suggestions ! My shell scripting skills aren't great :)

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-wireless-firmwares.service
```diff
@@ -0,0 +1,10 @@
+# This service is used for printing a message if
+# some wireless firmwares are missing
+[Unit]
+Description=Check if all the wireless firmwares are installed
+[Service]
```
travier: Hum, don't we need something else to order this properly during boot?

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,60 @@
...
+zd1211-firmware
+)
+regex=$(IFS='|'; echo "${firmwares[*]}")
+
+layered_packages="$(rpm-ostree status --json -b | jq -r '.deployments[0].packages[]')"
+
+if grep -q "NetworkManager-wifi" <<< "$layered_packages"; then
+    if grep -qP $regex <<< "$layered_packages"; then
+    return 0
+  fi
```
dustymabe: fix spacing

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-wireless-firmwares.service
jlebon: All those units writing MOTDs should probably be ordered `Before=systemd-user-sessions.service`.

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,60 @@
...
+nxpwireless-firmware
+realtek-firmware
+tiwilink-firmware
+atmel-firmware
+bcm283x-firmware
+zd1211-firmware
+)
+regex=$(IFS='|'; echo "${firmwares[*]}")
+
+layered_packages="$(rpm-ostree status --json -b | jq -r '.deployments[0].packages[]')"
```
jlebon: Probably want to use `requested-packages` here instead. That one includes inactive packages, whereas `packages` only includes packages that were actually layered.

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-wireless-firmwares.service
jbtrystram: I don't think so. Is there an rpm-ostree service we could order after just to be safe ?

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,60 @@
...
+nc=$(echo -e '\033[0m')
+
+motd_path=/run/motd.d/30_wireless_firmwares_warning.motd
+
+cat << EOF > "${motd_path}"
+${warn}
+##########################################################################
+WARNING: The NetworkManager-wifi  is a requested layered package on this
+system, but no wifi drivers are requested. The wifi drivers will be
+dropped from FCOS at a future date.
```
jlebon:
```suggestion
WARNING: The NetworkManager-wifi is a requested layered package on this
system, but no Wi-Fi drivers are requested. The Wi-Fi drivers will no
longer be included by default in the future.
```

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,60 @@
...
+
+More context and remediation steps are available in the following FAQ entry:
+ https://docs.fedoraproject.org/en-US/fedora-coreos/faq/#_how_do_i_add_the_missing_wifi_firmwares
```
jlebon:
```suggestion
https://docs.fedoraproject.org/en-US/fedora-coreos/faq/#_how_do_i_add_the_missing_wifi_firmwares
```

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-wireless-firmwares.service
```diff
@@ -0,0 +1,10 @@
+# This service is used for printing a message if
+# some wireless firmwares are missing
+[Unit]
+Description=Check if all the wireless firmwares are installed
```
jlebon: Shouldn't it be "any" instead? Also Title Case.
How about just

```suggestion
Description=Check For Wireless Firmware Packages
```

?

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/bash
```
jlebon: Missing `set -euo pipefail`.

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,62 @@
...
+warn=$(echo -e '\033[0;33m')
+# No color
+nc=$(echo -e '\033[0m')
+
+motd_path=/run/motd.d/30_wireless_firmwares_warning.motd
+
+cat << EOF > "${motd_path}"
+${warn}
+##########################################################################
+WARNING: The NetworkManager-wifi is a requested layered package on this
```
dustymabe:
```suggestion
WARNING: NetworkManager-wifi is a requested layered package on this
```

dustymabe (dismissed):

dustymabe:
I guess the docs/faq needs to land first so we can make sure the link we are using is right.

dustymabe:
in CI:

```
Apr 16 15:25:54.013651 rpm-ostreed.service[1509]: client(id:cli dbus:1.7 unit:coreos-check-wireless-firmwares.service uid:0) added; new total=1                                              
Apr 16 15:25:54.016500 init.scope[1]: sshd-keygen@rsa.service: Deactivated successfully.                                                                                                     
Apr 16 15:25:54.018616 init.scope[1]: Finished sshd-keygen@rsa.service - OpenSSH rsa Server Key Generation.                                                                                  
Apr 16 15:25:54.033734 init.scope[1]: Reached target sshd-keygen.target.                                                                                                                     
Apr 16 15:25:54.038889 rpm-ostreed.service[1509]: client(id:cli dbus:1.7 unit:coreos-check-wireless-firmwares.service uid:0) vanished; remaining=0                                           
Apr 16 15:25:54.039581 rpm-ostreed.service[1509]: In idle state; will auto-exit in 62 seconds                                                                                                
Apr 16 15:25:54.052455 init.scope[1]: Starting console-login-helper-messages-gensnippet-ssh-keys.service - Generate SSH keys snippet for display via console-login-helper-messages...        
Apr 16 15:25:54.067790 coreos-check-wireless-firmwares.service[1403]: /usr/libexec/coreos-check-wireless-firmwares: line 38: return: can only `return' from a function or sourced script     
Apr 16 15:25:54.068766 init.scope[1]: coreos-check-wireless-firmwares.service: Main process exited, code=exited, status=2/INVALIDARGUMENT                                                    
Apr 16 15:25:54.068965 init.scope[1]: coreos-check-wireless-firmwares.service: Failed with result 'exit-code'.                                                                               
Apr 16 15:25:54.072502 init.scope[1]: Failed to start coreos-check-wireless-firmwares.service - Check For Wireless Firmware Packages.                                                        
Apr 16 15:25:54.193699 init.scope[1]: var-lib-containers-storage-overlay-volatile\x2dcheck3087635057-merged.mount: Deactivated successfully.                                                 
Apr 16 15:25:54.207598 init.scope[1]: var-lib-containers-storage-overlay-compat2047962956-lower\x2dmapped.mount: Deactivated successfully.
```

dustymabe:
you'll need to use `exit` instead of `return`

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
dustymabe: I think this change was missed?

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,62 @@
...
+libertas-firmware
+mt7xxx-firmware
+nxpwireless-firmware
+realtek-firmware
+tiwilink-firmware
+atmel-firmware
+bcm283x-firmware
+zd1211-firmware
+)
+regex=$(IFS='|'; echo "${firmwares[*]}")
```
c4rt0: This is a little confusing, maybe a comment with an explanation of what this part of code does would be a nice to see.

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,62 @@
...
+nc=$(echo -e '\033[0m')
+
+motd_path=/run/motd.d/30_wireless_firmwares_warning.motd
+
+cat << EOF > "${motd_path}"
+${warn}
+##########################################################################
+WARNING: The NetworkManager-wifi is a requested layered package on this
+system, but no Wi-Fi drivers are requested. The Wi-Fi drivers will no
+longer be included by default in the future.
```
yasminvalim: Do we have a possible date/version for when this will no longer be available? 
Should we warn the users about this deadline?

jbtrystram: good point. I think we agreed on removing the packages on F41.

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
jbtrystram: I think an explanation in the code would be a little bit clumsy, so i'll do it here : 
IFS is `Internal Field Separator` in bash. The default value is  "<space><tab><newline>".
This is how the shell separates values. 
e.g. in
```
fruits="apple pear orange"
for i in $fruits
do
   echo "'$i' is a fruit"
done
```
This will loop over "apple", "pear" then "orange".
writing: 
```
fruits="apple|pear|orange"
IFS='|'; for i in $fruits
do
   echo "'$i' is a fruit"
done
```
Would have the same result. 

So what `IFS='|'; echo "${firmwares[*]}` does is echoing all the values in `firmwares` but instead of the default <space> in between, the output will be `a|b|c|d`, which is convenient because it means OR

c4rt0: I was considering potential code review in the future, where it could take time for someone to understand this bit of code. It certainly took me a while - thank you for the explanation though!

dustymabe: rather than document in a comment what `IFS` does we could explain what the outcome of the following statement would be, which I think would achieve the goal:

```suggestion
# Get firmware names into `a|b|c|d` regex string
regex=$(IFS='|'; echo "${firmwares[*]}")
```

Then if the reader wants to understand how IFS works they could go look it up, but they at least understand the goal of the current code.

dustymabe:
sorry I pushed up something here by mistake.. fixing

dustymabe:
opening

dustymabe:
I guess I no longer have permissions to push to that branch in his fork :(

JB can you do a force push (or give me access to your fork and I can fix it).

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
jbtrystram: Oh yeah of course, i think i misunderstood Adam comment

jbtrystram:
@dustymabe  I re-enabled write access to maintainer on this branch, I  am not sure why it got turned off

dustymabe:
@jbtrystram i'm still not able to push to it. and I can't re-open this PR because currently the branch in the fork doesn't share any history with this repo

jbtrystram:
@dustymabe you pushed a branch from coreos/fedora-coreos-docs here. 

Sadly I realized that after I f***d my local repo trying to fix it. So i have no copies of the previous commit containing the MOTD script .
I should have let the coffee infuse more :sweat_smile:

jbtrystram:
Reflog to the rescue :) 

https://github.com/coreos/fedora-coreos-config/pull/2978

dustymabe:
> @dustymabe you pushed a branch from coreos/fedora-coreos-docs here.
> 

yeah it was a mistake on my part :(

> Sadly I realized that after I f***d my local repo trying to fix it. So i have no copies of the previous commit containing the MOTD script . I should have let the coffee infuse more 😅

I see below you found it, but yeah I had a local copy I was trying to push to fix but couldn't :(

PR #2964, opened 2024-04-12
Denylist: re-enable ext/var-mount/scsi-id

jbtrystram:
An update in rawhide broke /dev/disk-by symlinks that the scsi-id relied on. [1] fixed the test but i forgot to re-enable it

[1] https://github.com/coreos/fedora-coreos-config/pull/2943


c4rt0 (approved):
/LGTM

PR #2965, opened 2024-04-15
denylist: extend snooze for failing kola tests, remove kdump.crash

c4rt0:
The `coreos.ignition.ssh.key` is now expired. Extending the snooze to stop our pipeline from failing.

Also removing here the `ext.config.kdump.crash` since it should be fixed by the https://github.com/coreos/fedora-coreos-config/pull/2966

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -7,7 +7,7 @@
   tracker: https://github.com/coreos/coreos-assembler/pull/1478
 - pattern: coreos.ignition.ssh.key
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1553
-  snooze: 2024-04-15
+  snooze: 2024-04-29
   warn: true
   platforms:
     - azure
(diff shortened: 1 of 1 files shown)


jbtrystram:
weird, kdump.crash should be fixed : https://github.com/coreos/fedora-coreos-tracker/issues/1698#issuecomment-2041720310

c4rt0:
Right, I saw the warning on one of the many builds I looked at today. Let me update the denylist, thanks @jbtrystram.

jbtrystram:
@c4rt0 looking more into it, it's a different issue, sorry :)

c4rt0:
No worries, back to the beginning...

dustymabe:
> @c4rt0 looking more into it, it's a different issue, sorry :)

right, but a different issue that should be solved with https://github.com/coreos/fedora-coreos-config/pull/2966 so we can remove the denylist entry here I think??

This PR as is LGTM. If we do remove the kdump entry here then let's make it a separate second commit.

dustymabe (dismissed):

c4rt0:
Updated as per suggestion - thank you for the reviews.

dustymabe (approved):
LGTM

PR #2969, opened 2024-04-16
[next] tree: promote changes from next-devel at 485db993665e8920ab18afecff48a9395e43dab7

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/886.


c4rt0 (approved):
/lgtm

I paid the most attention to the `coreos-builds-bisect.py` here which is quite interesting.

dustymabe (approved):

PR #2971, opened 2024-04-17
overlay.d/05core: fix variable usage in coreos-diskful-generator

dustymabe:
Saw this scroll by in a log:

```
[    1.792490] /usr/lib/systemd/system-generators/coreos-diskful-generator: line 37: value: unbound variable
```

We need to set a value for value in order to prevent the unbound variable error.


jbtrystram (approved):
lgtm

PR #2974, opened 2024-04-19
tests/kola: handle flakes in quadlet test

dustymabe:
This test was flaking on ppc64le a lot. After debugging we were able to see that the `podman inspect systemd-test` was running before the container was actually started, but the call didn't fail because `systemd-test` matched the name of the volume and the network resource. So we need to do a few things here:

- make sure we don't try to check things before the volume and network has been created or the container has been started.
- limit the inspect call for the container to only containers by doing `podman container inspect` instead of generic `podman inspect`.

### DIFF
--- a/tests/kola/containers/quadlet/test.sh
+++ b/tests/kola/containers/quadlet/test.sh
@@ -9,23 +9,36 @@ set -xeuo pipefail
 # shellcheck disable=SC1091
 . "$KOLA_EXT_DATA/commonlib.sh"
 
-if [[ "$(podman volume inspect systemd-test | jq -r '.[0].Labels."org.test.Key"')" != "quadlet-test-volume" ]]; then
+# Test volume
+if ! is_service_active test-volume.service; then
+  fatal "test-volume.service failed to start"
+fi
+volume_info=$(podman volume inspect systemd-test)
+if [[ "$(jq -r '.[0].Labels."org.test.Key"' <<< "$volume_info")" != "quadlet-test-volume" ]]; then
     fatal "Volume not correctly created"
 fi
 
-if [[ "$(podman network inspect systemd-test | jq -r '.[0].labels."org.test.Key"')" != "quadlet-test-network" ]]; then
+# Test network
+if ! is_service_active test-network.service; then
+  fatal "test-network.service failed to start"
+fi
+network_info=$(podman network inspect systemd-test)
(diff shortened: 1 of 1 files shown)


tests/kola/containers/quadlet/test.sh
```diff
@@ -9,24 +9,41 @@ set -xeuo pipefail
...
+    fi
+    if [[ "$(jq -r '.[0].NetworkSettings.Networks[].NetworkID' <<< "$container_info")" != "systemd-test" ]]; then
+        fatal "Container not using the correct network"
+    fi
+    if [[ "$(jq -r '.[0].HostConfig.Binds[0]' <<< "$container_info")" != "systemd-test:/data:rw,rprivate,nosuid,nodev,rbind" ]]; then
+        fatal "Container not using the correct volume"
+    fi
+}
+
+quadlet_test
```
jbtrystram: what's the benefit of wrapping everything in a function ? I see only one call so it could be flattened

tests/kola/containers/quadlet/test.sh
```diff
@@ -9,24 +9,41 @@ set -xeuo pipefail
...
+    if [[ "$(jq -r '.[0].labels."org.test.Key"' <<< "$network_info")" != "quadlet-test-network" ]]; then
+        fatal "Network not correctly created"
+    fi
+
+    # Test container
+    if ! is_service_active test.service; then
+      fatal "test-network.service failed to start"
+    fi
+    container_info=$(podman container inspect systemd-test)
+    echo "${container_info} | jq ."
```
jbtrystram: is this a left-over debug statement ?

jbtrystram (changes requested):
A couple of nits, otherwise LGTM

tests/kola/containers/quadlet/test.sh
dustymabe: I had originally thought we were going to need a retry with a sleep, in which case it would have been easier to have this code in a function that we could then call from a while loop. after investigating further that wasn't needed, but I decided to leave the function anyway.

I removed it now.

tests/kola/containers/quadlet/test.sh
dustymabe: yep - removed!

jbtrystram (approved):
LGTM

PR #2978, opened 2024-04-22
Add missing wifi firmware checks

jbtrystram:
Some wifi-firmwares will be split into sub-packages in fedora 40
We will keep them in until fedora 41, but display a warning message
in the console if NetworkManager-wifi is layered without the most
popular the wifi firmwares.

See coreos/fedora-coreos-tracker#1575

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
+++ b/overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
@@ -1,0 +1,63 @@
+#!/usr/bin/bash
+# This script checks if
+# and will prints a message to the serial console
+# to warn the user about missing wifi firmware messages
+# and provide remediation steps
+# See https://github.com/coreos/fedora-coreos-tracker/issues/1575
+
+set -euo pipefail
+
+# List of wifi-firmwares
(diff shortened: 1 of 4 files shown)


c4rt0 (changes requested):

overlay.d/15fcos/usr/libexec/coreos-check-wireless-firmwares
```diff
@@ -0,0 +1,62 @@
...
+iwlwifi-mvm-firmware
+libertas-firmware
+mt7xxx-firmware
+nxpwireless-firmware
+realtek-firmware
+tiwilink-firmware
+atmel-firmware
+bcm283x-firmware
+zd1211-firmware
+)
```
c4rt0, line 28: This is way too much noise for a simple comment, but I will mention it again
```
# Get firmware names into `a|b|c|d` regex string
```

c4rt0: Actually, [this](https://github.com/coreos/fedora-coreos-config/pull/2963#discussion_r1572483100) clarifies it enough. Let's skip the comment (it might be unnecessary)

jlebon (dismissed):

jlebon (approved):

gursewak1997 (approved):

PR #2980, opened 2024-04-23
manifests: Don't include Wi-Fi firmwares starting with Fedora 41

### LINKED ISSUE #1575: dropping some firmware packages that provide Wi-Fi firmware
The linux-firmware package [dropped](https://src.fedoraproject.org/rpms/linux-firmware/c/be92a95e169422f1108bce0aee70ed96fc26fb0b?branch=rawhide) the following firmware subpackages as requirements in F39+:

```
    atheros-firmware-20230804-153.fc38.noarch
    brcmfmac-firmware-20230804-153.fc38.noarch
    mt7xxx-firmware-20230804-153.fc38.noarch
    realtek-firmware-20230804-153.fc38.noarch 
```

We should determine if this is a change we want to make.

travier:
manifests: Don't include Wi-Fi firmwares starting with Fedora 41

Move Wi-Fi firmwares to their own manifest so that we can automatically
drop then when we rebase to Fedora 41.

We can safely do that now that we have included a warning for existing
users.

See: https://github.com/coreos/fedora-coreos-config/pull/2978
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1575

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -157,8 +157,6 @@ packages:
   - iptables-legacy
   # GPU Firmware files (not broken out into subpackage of linux-firmware in RHEL yet)
   - amd-gpu-firmware intel-gpu-firmware nvidia-gpu-firmware
-  # Wifi/BT firmware files https://github.com/coreos/fedora-coreos-tracker/issues/1575
-  - atheros-firmware brcmfmac-firmware mt7xxx-firmware realtek-firmware nxpwireless-firmware tiwilink-firmware
 
 
 # - irqbalance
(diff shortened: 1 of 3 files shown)


jlebon (approved):
Code itself LGTM though shouldn't the commit title be e.g.

> manifests: don't ship Wi-Fi firmwares on f41+

?

jbtrystram (approved):
Thanks Timothée ! I indeed forgot that bit on the other PR !

travier:
> > manifests: don't ship Wi-Fi firmwares on f41+

Indeed, agree. Do you think it's worth pushing an update?

travier:
I'll do it. We are not in a hurry

jlebon (approved):

PR #2981, opened 2024-04-23
Move to Fedora Linux 40

aaradhak:
The Fedora 40 GA is to be released on Tuesday April 23rd.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1674

PR #2983, opened 2024-04-24
Updating container image to fedora-toolbox:40

aaradhak:
PR : https://github.com/coreos/fedora-coreos-config/pull/2981

The CI in PR-2891 fails as the 'mount' command could not be executed. Looks like F40 got rid of 'util-linux-core' package which provides the 'mount' command among other things.

In order to make the CI test PASS in PR-2981, the container image needs to be updated to quay.io/fedora/fedora-toolbox:40.


dustymabe:
i would just add this as a commit in https://github.com/coreos/fedora-coreos-config/pull/2981

aaradhak:
Added this change as a commit in https://github.com/coreos/fedora-coreos-config/pull/2981

PR #2984, opened 2024-04-24
DNM: experiment with unit requirements for multipath over iscsi work

dustymabe:
We are experimenting with iscsi and also multipath+iscsi.

In the multipath+iscsi case we end up with dependency issues in our initramfs because of `coreos-multipath-wait.target`:

- It blocks `cryptsetup-pre.target` and thus `basic.target`:

```
basic.target -> sysinit.target -> cryptsetup.target -> cryptsetup-pre.target -> coreos-multipath-wait.target -> /boot disk
```

- It blocks `dracut-initqueue.service` and that is what sets up the iscsi disks:

```
dracut-initqueue.service -> coreos-multipath-wait.target
```

I was testing this with a disk that was installed using:

```
sudo coreos-installer install /path/to/disk \
    --append-karg rd.multipath=default      \
    --append-karg root=/dev/disk/by-label/dm-mpath-root \
    --append-karg rw --append-karg rd.iscsi.firmware=1  \
    --console ttyS0,115200n8 -i config.ign --insecure   \
    --image-file fedora-coreos-40.20240424.dev.1-metal.x86_64.raw.xz
```

an a PXE config that looked like:

```
[dustymabe@media images]$ cat boot.ipxe 
#!ipxe
# set some random uuid as the initiator iqn, otherwise
# this will fail pointing to https://ipxe.org/1c0d6502
set initiator-iqn iqn.68cc69b9-1b54-4ff1-9d61-eedb570da8fd
sanboot iscsi:192.168.122.137::::iqn.2024-04.com.coreos:0 \
        iscsi:192.168.122.137::::iqn.2024-04.com.coreos:1

```

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-trigger.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-trigger.service
@@ -11,7 +11,7 @@ Description=CoreOS Trigger Multipath
 DefaultDependencies=false
 Requires=coreos-multipath-wait.target
 After=coreos-multipath-wait.target
-Before=cryptsetup-pre.target
+#Before=cryptsetup-pre.target
 
 [Service]
 Type=oneshot
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
@@ -1,7 +1,7 @@
 [Unit]
 Description=CoreOS Wait For Multipathed Boot
 DefaultDependencies=false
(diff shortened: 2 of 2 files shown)


dustymabe:
After talking with @jlebon I was able to investigate a bit more:

> basic.target -> sysinit.target -> cryptsetup.target -> cryptsetup-pre.target -> coreos-multipath-wait.target -> /boot disk

This chain isn't important or rather it's not what is blocking things. It's that `coreos-enable-network` was blocking on `basic.target`, which was holding up networking and dracut-initqueue.

> dracut-initqueue.service -> coreos-multipath-wait.target

This one was important. We were able to remove it.

After dropping the `After=basic.target` from `coreos-enable-network.service` and `Before=dracut-initqueue.service` from `coreos-multipath-wait.target` things seem to work.

dustymabe:
took the output of this investigation and opened https://github.com/coreos/fedora-coreos-config/pull/2987

PR #2985, opened 2024-04-24
[next-devel] disable candidate-compose repo

aaradhak:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1674


dustymabe:
as part of this let's clear out all overrides too in a separate commit just to reset this branch to steady state

aaradhak:
> as part of this let's clear out all overrides too in a separate commit just to reset this branch to steady state

I have dropped the overrides now.

dustymabe (approved):

PR #2987, opened 2024-04-26
change unit ordering to allow iscsi+multipath to work

dustymabe:
We want multipath on top of iscsi to work. We need to drop some unit
dependencies in order to get this to work. See 
https://github.com/coreos/fedora-coreos-config/pull/2984 for more
context.

This was tested with an iscsi disk that was installed with:

```
sudo coreos-installer install /path/to/disk \
    --append-karg rd.multipath=default      \
    --append-karg root=/dev/disk/by-label/dm-mpath-root \
    --append-karg rw --append-karg rd.iscsi.firmware=1  \
    --console ttyS0,115200n8 -i config.ign --insecure   \
    --image-file fedora-coreos-40.20240424.dev.1-metal.x86_64.raw.xz
```

and then shared over iPXE with:

```
set initiator-iqn iqn.68cc69b9-1b54-4ff1-9d61-eedb570da8fd
sanboot iscsi:192.168.122.137::::iqn.2024-04.com.coreos:0 \
        iscsi:192.168.122.137::::iqn.2024-04.com.coreos:1
```

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
@@ -1,7 +1,6 @@
 [Unit]
 Description=CoreOS Wait For Multipathed Boot
 DefaultDependencies=false
-Before=dracut-initqueue.service
 After=dracut-cmdline.service
 Requires=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
 After=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.service
@@ -2,7 +2,6 @@
 Description=CoreOS Enable Network
 ConditionPathExists=/etc/initrd-release
 DefaultDependencies=false
-After=basic.target
 
 # Triggering conditions for cases where we need network:
 #  * when Ignition signals that it is required for provisioning.


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.service
```diff
@@ -2,7 +2,6 @@
 Description=CoreOS Enable Network
 ConditionPathExists=/etc/initrd-release
 DefaultDependencies=false
-After=basic.target
```
jlebon, line 5: This definitely looked suspicious how it was before since `basic.target` is a default dependency and you'd use `DefaultDependencies=false` to not inherit that.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
```diff
@@ -1,7 +1,6 @@
 [Unit]
 Description=CoreOS Wait For Multipathed Boot
 DefaultDependencies=false
-Before=dracut-initqueue.service
```
jlebon, line 4: There's one case where I think this might be load-bearing: FIPS. The FIPS module needs to mount the bootfs to be able to get the HMAC of the kernel and verify it. So in theory, it could try to be doing I/O to the bootfs before multipathd takes over. And of course, dracut being dracut, there's no way to order our unit before just this FIPS code without also blocking the iSCSI code.

I think one way to work around this is to have `rdcore` inject the `boot` karg as `boot=/dev/disk/by-label/dm-mpath-boot` in the multipath case instead of the UUID. That'll force the FIPS code to wait until the multipathed-version actually shows up.

Anyway, we don't have to tackle this right now or block on this. The larger issue is that we don't have any CI for the 'non-optimized paths' case, so we have no way of verifying that scenario. Ideally, we'd add that, and then we could verify that the problem above is real, and verify the fix for it.

jlebon (approved):
Thanks for digging into this!

PR #2988, opened 2024-04-29
ci & tests: Use registry.fedoraproject.org for Fedora container images

travier:
The plan is to keep using the registry.fedoraproject.org name but redirect it to quay.io.

Reverts: https://github.com/coreos/fedora-coreos-config/pull/2623

See: https://pagure.io/fedora-infrastructure/issue/11543

### DIFF
--- a/tests/kola/containers/quadlet/config.bu
+++ b/tests/kola/containers/quadlet/config.bu
@@ -9,7 +9,7 @@ storage:
           Description=A minimal container
 
           [Container]
-          Image=quay.io/fedora/fedora-minimal:39
+          Image=registry.fedoraproject.org/fedora-minimal:39
           Exec=sleep 60
           Volume=test.volume:/data
           Network=test.network
--- a/tests/kola/containers/quadlet/config.bu
+++ b/tests/kola/containers/quadlet/config.bu
@@ -9,7 +9,7 @@ storage:
           Description=A minimal container
 
           [Container]
-          Image=registry.fedoraproject.org/fedora-minimal:39
+          Image=registry.fedoraproject.org/fedora-minimal:40
           Exec=sleep 60
           Volume=test.volume:/data
           Network=test.network
--- a/tests/kola/containers/quadlet/test.sh
+++ b/tests/kola/containers/quadlet/test.sh
@@ -32,7 +32,7 @@ if ! is_service_active test.service; then
   fatal "test-network.service failed to start"
 fi
 container_info=$(podman container inspect systemd-test)
-if [[ "$(jq -r '.[0].ImageName' <<< "$container_info")" != "quay.io/fedora/fedora-minimal:39" ]]; then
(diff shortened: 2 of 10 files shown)


tests/kola/containers/quadlet/config.bu
```diff
@@ -9,7 +9,7 @@ storage:
           Description=A minimal container
 
           [Container]
-          Image=quay.io/fedora/fedora-minimal:39
+          Image=registry.fedoraproject.org/fedora-minimal:39
```
c4rt0: Image=registry.fedoraproject.org/fedora-minimal:40

tests/kola/containers/quadlet/test.sh
```diff
@@ -32,7 +32,7 @@ if ! is_service_active test.service; then
   fatal "test-network.service failed to start"
 fi
 container_info=$(podman container inspect systemd-test)
-if [[ "$(jq -r '.[0].ImageName' <<< "$container_info")" != "quay.io/fedora/fedora-minimal:39" ]]; then
+if [[ "$(jq -r '.[0].ImageName' <<< "$container_info")" != "registry.fedoraproject.org/fedora-minimal:39" ]]; then
```
c4rt0: if [[ "$(jq -r '.[0].ImageName' <<< "$container_info")" != "registry.fedoraproject.org/fedora-minimal:40" ]]; then

c4rt0 (changes requested):

c4rt0:
Sorry - the above isn't that obvious. I'm referring to the `fedora-minimal:39` which still feels outdated. I understand this is only addressing the `registry.fedoraproject.org` change - I am wondering if we could do both here?

travier:
Thanks! I'll add the 39->40 change to this PR.

c4rt0:
> I'll add the 39->40 change to this PR.

Ah, sorry - I pushed the change before seeing your comment.

c4rt0 (approved):
/lgtm

travier:
> > I'll add the 39->40 change to this PR.
> 
> Ah, sorry - I pushed the change before seeing your comment.

Looks like I pushed it right after :)

c4rt0 (approved):

jlebon:
The only concern I have with this "redirect" plan is that it now makes our image pulls dependent on both registry.fedoraproject.org and Quay.io working, instead of just the latter. For that reason alone, I'd vote to just keep using Quay.io. ISTM like the main goal of the redirect approach is to not break existing docs/code, which is not a concern here since we've already done the work to move them over.

Commented in https://pagure.io/fedora-infrastructure/issue/11543#comment-907073.

dustymabe:
> For that reason alone, I'd vote to just keep using Quay.io.

It's also weird because we push directly to quay for our FCOS images. Inconsistency isn't great.

travier:
Let's do https://github.com/coreos/fedora-coreos-config/pull/2990 first and then we'll see for this one.

dustymabe:
Maybe we close this?

travier:
We can probably close this for now indeed.

PR #2989, opened 2024-04-29
denylist: extend snooze for `coreos.ignition.ssh.key`

marmijo:
This test is still failing. Extend the snooze to unblock the pipeline.


c4rt0 (approved):
/lgtm

dustymabe (approved):
LGTM - though maybe we can prioritize trying to root cause this in the next few weeks.

PR #2997, opened 2024-05-10
manifests: dnf5 is now default in rawhide; stop symlinking dnf

jlebon:
The postprocess bomb here went off this morning, which means we can simplify this by a lot!

### DIFF
--- a/manifests/include-dnf.yaml
+++ b/manifests/include-dnf.yaml
@@ -1,13 +1,2 @@
 packages:
   - dnf5
-
-# until dnf5 becomes the default, manually symlink dnf to it
-postprocess:
-  - |
-    #!/usr/bin/bash
-    set -euo pipefail
-    if command -v dnf; then
-      echo 'dnf5 is now the default, remove this postprocess script!' >&2
-      exit 1
-    fi
-    ln -s dnf5 /usr/bin/dnf


dustymabe (dismissed):
LGTM

manifests/include-dnf.yaml
```diff
@@ -1,13 +1,2 @@
 packages:
-  - dnf5
-
-# until dnf5 becomes the default, manually symlink dnf to it
-postprocess:
-  - |
-    #!/usr/bin/bash
-    set -euo pipefail
-    if command -v dnf; then
```
dustymabe, line 9: the way this test is constructed it looks like all it was doing was checking to see if the `dnf5` package provides `/usr/bin/dnf` and then warning us.

It looks like CI is still pulling in a `dnf` that still requires python. Maybe the old `dnf` package is still there, but `dnf5` now provides `/usr/bin/dnf` too?

jlebon: Yeah, looks like it.

```
[root@1710845a3e2e /]# dnf repoquery dnf dnf5
Updating and loading repositories:
Repositories loaded.
dnf-0:4.20.0-1.fc41.noarch
dnf5-0:5.2.1.0-1.fc41.i686
dnf5-0:5.2.1.0-1.fc41.x86_64
```

Updated to still request `dnf5`.

dustymabe: I guess we'll know now when we need to make a change here because presumably the dnf5 package will no longer exist? of maybe `dnf5` will be a virtual provides from the `dnf` package once it gets replaced?

dustymabe (approved):
LGTM

manifests/include-dnf.yaml
jlebon: I think the plan is for `dnf5` to obsolete `dnf` so what we have here should keep working, but yeah ideally eventually we can just say `dnf`. I'm not sure if the obsolete bit would be enough for that though.

PR #3000, opened 2024-05-16
overlay/coreos-check-ssh-keys: Remove invalid service key

jbtrystram:
The `Before` key is undefined for the `Service` Section. This was silentely ignored by systemd until systemd-v256-rc1 where it started emmitting a log message :
`Unknown key name 'Before' in section 'Service', ignoring.`

This causes the pipeline to fail as kola consider this an error.


dustymabe (approved):
LGTM

PR #3001, opened 2024-05-16
buildroot: update URL to rpm-ostree spec

jlebon:
This was changed in https://github.com/coreos/rpm-ostree/pull/4964.


jlebon:
Need to wait until that PR merges.

cgwalters (approved):

PR #3002, opened 2024-05-16
tests/kdump: bump timeout for kdump.service

jlebon:
Saw a CI run timeout because `kdump.service` took 65s to build its initramfs but we only wait 60s. Let's just bump this to have a larger margin of error.


jbtrystram (approved):
LGTM

dustymabe (approved):
LGTM

PR #3004, opened 2024-05-17
[rawhide] overrides: pin systemd-255.5-1.fc41

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,34 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  systemd:
+    evr: 255.5-1.fc41
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1735
+      type: pin
+  systemd-container:
+    evr: 255.5-1.fc41
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1735
+      type: pin
+  systemd-libs:
+    evr: 255.5-1.fc41
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1735
+      type: pin
+  systemd-pam:
+    evr: 255.5-1.fc41
+    metadata:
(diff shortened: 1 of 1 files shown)


dustymabe:
fun - apparently a bunch of other packages now require systemd >= 256 so it's not going to be as easy..

we either need to figure out a way to disable the check temporarily OR pin a bnuch of other packages too

cgwalters:
(Or, don't try to gate things like this in FCOS only, but move checks like this into fedora gating itself)

dustymabe:
> (Or, don't try to gate things like this in FCOS only, but move checks like this into fedora gating itself)

(these comments aren't helpful)

jlebon:
Reading the depsolve tea leaves, I think adding pins on lvm2-2.03.23-1.fc40.x86_64 and device-mapper-1.02.197-1.fc40 should do it.

Skipping the check would be nice, but requires some cosa work.

> (Or, don't try to gate things like this in FCOS only, but move checks like this into fedora gating itself)

I'm confused by this comment. I think you know we already know this would be valuable and that we already did a bunch of work in that area.

manifest-lock.overrides.yaml
```diff
@@ -8,4 +8,44 @@
...
+  lvm2:
+    evr: 2.03.23-1.fc40
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1735
+      type: pin
+  device-mapper:
+    evr: 1.02.197-1.fc40
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1735
+      type: pin
```
jlebon: If you grep the lockfile for those EVRs, you'll see there's a bunch of other subpackages we also need to add.

jlebon: Ideally, the workflow would support giving multiple Koji/Bodhi URLs. For now, one way to do this is to run the `ci/overrides.py pin` command locally for each one. (You can see how the `add-overrides.yml` action uses it.)

jbtrystram: Ok, added a couple through the correct script (see commit message) and now it builds :) 
Thanks for the hint @jlebon !

jlebon (approved):

PR #3009, opened 2024-05-29
Enable composefs

jbtrystram:
rebased https://github.com/coreos/fedora-coreos-config/pull/2856 on rawhide

See: https://github.com/coreos/fedora-coreos-config/pull/2856#issuecomment-2137635204
See: https://github.com/coreos/fedora-coreos-tracker/issues/1718

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -51,3 +51,10 @@
   streams:
     - rawhide
     - branched
+- pattern: ext.config.kdump.crash
+  tracker: https://bugzilla.redhat.com/show_bug.cgi?id=2284097
+  snooze: 2024-09-01
+  warn: true
+  streams:
+    - rawhide
+    - branched
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -30,6 +30,8 @@ conditional-include:
   # Wifi firmwares will be dropped in F41
   - if: releasever < 41
     include: wifi-firmwares.yaml
+  - if: releasever >= 41
+    include: composefs.yaml
 
 ostree-layers:
   - overlay/15fcos
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/ostree-cmdline.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/ostree-cmdline.sh
@@ -11,7 +11,10 @@ set -euo pipefail
 case "${1:-unset}" in
     start)
         treepath="$(echo /sysroot/ostree/boot.1/*/*/0)"
-        echo "$(cat /proc/cmdline) ostree=${treepath#/sysroot}" > /tmp/cmdline
+        # ostree-prepare-root requires /etc and /var to be writeable for composeFS
+        # which cannot happen in the live ISO. Disable composeFS there
+        # https://github.com/coreos/fedora-coreos-config/pull/3009#issuecomment-2235923719
+        echo "$(cat /proc/cmdline) ostree=${treepath#/sysroot} ostree.prepare-root.composefs=0" > /tmp/cmdline
         mount --bind /tmp/cmdline /proc/cmdline
         ;;
     stop)
--- a/overlay.d/08composefs/README.md
+++ b/overlay.d/08composefs/README.md
@@ -1,0 +1,1 @@
+Enable composefs by default; more in https://ostreedev.github.io/ostree/composefs/
--- a/manifests/composefs.yaml
+++ b/manifests/composefs.yaml
@@ -1,0 +1,3 @@
+# Enable composefs by default.
+ostree-layers:
+  - overlay/08composefs
--- a/overlay.d/08composefs/usr/lib/ostree/prepare-root.conf
+++ b/overlay.d/08composefs/usr/lib/ostree/prepare-root.conf
@@ -1,0 +1,2 @@
+[composefs]
+enabled = true
--- a/tests/kola/root-reprovision/autosave-xfs/test.sh
+++ b/tests/kola/root-reprovision/autosave-xfs/test.sh
@@ -25,7 +25,7 @@ fi
 journalctl -u ignition-ostree-transposefs-autosave-xfs.service --grep=agcount
 ok "autosaved XFS on large disk"
 
-eval $(xfs_info / | grep -o 'agcount=[0-9]*')
+eval $(xfs_info /sysroot | grep -o 'agcount=[0-9]*')
 expected=4
 if [ "$agcount" -gt "$expected" ]; then
     fatal "expected agcount of at most ${expected}, got ${agcount}"
--- a/tests/kola/root-reprovision/luks/autosave-xfs/test.sh
+++ b/tests/kola/root-reprovision/luks/autosave-xfs/test.sh
@@ -27,7 +27,7 @@ if [ -z "${AUTOPKGTEST_REBOOT_MARK:-}" ]; then
     fi
     ok "autosaved XFS on large disk"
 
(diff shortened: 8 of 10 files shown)


jbtrystram:
Why is it always kdump :sob: 
I only see `May 29 15:31:15.686509 kdump.service[1598]: kdump: mkdumprd: failed to make kdump initrd`
in the journal. I don't see any clue about the filesystem being read-only 
I'll investigate more locally.

jbtrystram:
After chatting a bit with Coiby Xu I filed https://bugzilla.redhat.com/show_bug.cgi?id=2284097

jbtrystram:
Maybe related to the kdump failure : https://issues.redhat.com/browse/RHEL-35885

jbtrystram:
Closing for #3009 (against testing-devel)

jbtrystram:
Tested this today locally : 
```
=== RUN   ext.config.kdump.crash
--- PASS: ext.config.kdump.crash (153.11s)
```

jbtrystram:
Interesting failures (but making total sense) : 
` ext.config.root-reprovision.luks.autosave-xfs`:
```
++ xfs_info /
++ grep -o 'agcount=[0-9]*'
xfs_info: cannot open /: Is a directory
 /usr/local/bin/kola-runext-test.sh: line 32: agcount: unbound variable
```
I suppose this is because `/` is no longer an XFS filesystem 

`ext.config.root-reprovision.raid1`:
```
findmnt -nvr / -o SOURCE
srcdev=none
```
Same here, `/` is not mounted from the raid device but from the composeFS overlay

jlebon:
Pushed some fixes which should help.

jbtrystram:
Thanks !

jbtrystram:
Live ISO fails to boot with :
```
[    6.036086] ostree-prepare-root[761]: ostree-prepare-root: sysroot.readonly=true requires /sysroot to be writable at this point, the cmdline should contain rw but not ro, if that is not the case this is likely the issue
```

From what I understand it's because in the composeFS path, ostree-prepare-root want to mount `/etc/` and `/var` as writeable , and requires sysroot to be writeable to do that ? 

See https://github.com/ostreedev/ostree/blob/4b96359e106a832dacdec9f3ef19dc730783e7a3/src/switchroot/ostree-prepare-root.c#L505


I guess the quickfix option here would be to disable composeFS on the live ISO, since it's already a read-only system anyway. We could  add the `ostree.prepare-root.composefs=0` karg to https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator

@cgwalters  WDYT ?

Let's try [`49c6b4a` (#3009)](https://github.com/coreos/fedora-coreos-config/pull/3009/commits/49c6b4a6bf30e9f86fddfc570472991a10f21d16)

jbtrystram:
This didn't work : 
```
[    5.554586] ostree-prepare-root[775]: composefs: mounted successfully
[    5.555464] ostree-prepare-root[775]: ostree-prepare-root: sysroot.readonly=true requires /sysroot to be writable at this point, the cmdline should contain rw but not ro, if that is not the case this is likely the issue
```

okay, I just see now that this karg support was merged last week and not released yet

cgwalters:
> I guess the quickfix option here would be to disable composeFS on the live ISO, since it's already a read-only system anyway. We could add the ostree.prepare-root.composefs=0 karg to https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator

Yes, or have the liveiso actually be built from a derived image/commit which drops the composefs config.

jlebon:
I think we can use the karg in the short-term, but it's also awkward to work around this as a built-in karg in our official ISOs. Note that ISO kargs are highly visible because there are APIs to query and modify them. (And on that point, nothing actually prevents a user from deleting that karg. We could say "don't do that", but that's awkward UX.)

I think this is basically another instance of https://github.com/ostreedev/ostree/issues/1921 (which I've just retitled); so then ostree-prepare-root would detect that this is a live environment and just not even try to set up composefs there.

manifests/fedora-coreos.yaml
```diff
@@ -30,6 +30,8 @@ conditional-include:
   # Wifi firmwares will be dropped in F41
   - if: releasever < 41
     include: wifi-firmwares.yaml
+  - if: stream == "rawhide"
```
jlebon: I'm optimistic we can land this in f41. Let's do `releasever < 41` here?

kola-denylist.yaml
```diff
@@ -51,3 +51,9 @@
   streams:
     - rawhide
     - branched
+- pattern: ext.config.kdump.crash
+  tracker: https://bugzilla.redhat.com/show_bug.cgi?id=2284097
+  snooze: 2024-09-01
+  warn: true
+  streams:
+    - rawhide
```
jlebon, line 59: So then, we'd have to add `branched` here.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/ostree-cmdline.sh
```diff
@@ -11,7 +11,7 @@ set -euo pipefail
 case "${1:-unset}" in
     start)
         treepath="$(echo /sysroot/ostree/boot.1/*/*/0)"
-        echo "$(cat /proc/cmdline) ostree=${treepath#/sysroot}" > /tmp/cmdline
+        echo "$(cat /proc/cmdline) ostree=${treepath#/sysroot} ostree.prepare-root.composefs=0" > /tmp/cmdline
```
jlebon, line 17: A bit self-referential, but let's link to https://github.com/coreos/fedora-coreos-config/pull/3009#issuecomment-2235923719 here as way to explain why we're adding this karg.

overlay.d/08composefs/README.md
```diff
@@ -0,0 +1,2 @@
+Enable composefs by default; more in https://ostreedev.github.io/ostree/composefs/
+For now rawhide only.
```
jlebon:
```suggestion
```

jlebon (commented):
Nice, this looks sane to me. Obviously, the ostree-prepare-root cmdline bit is a massive hack, but it's not new either.

Once those changes are done and CI is happy, we can retarget the PR against testing-devel instead of creating a new one.

jlebon:
Sorry, I messed up the suggestion in https://github.com/coreos/fedora-coreos-config/pull/3009#discussion_r1733143847. It should've been `>= 41`. To not waste another round-trip, I just fixed that. While there, I also cleaned up the commits a bit, merging the two commits that were both called e.g. "Enable composefs for 41+", and finally rebased onto the latest rawhide. Once CI is green here, let's retarget testing-devel and get this in.

jbtrystram:
Thanks @jlebon ! I am a bit embarrassed I didn't caught it 😅

jlebon (approved):

jlebon:
CI happy on rawhide. Retargeted testing-devel and approved. Will let someone else also stamp and merge.

travier (approved):

jbtrystram:
:tada:

kola-denylist.yaml
```diff
@@ -51,3 +51,10 @@
   streams:
     - rawhide
     - branched
+- pattern: ext.config.kdump.crash
+  tracker: https://bugzilla.redhat.com/show_bug.cgi?id=2284097
```
dustymabe, line 55: can you open an FCOS issue tracker ticket for this?

jbtrystram: https://github.com/coreos/fedora-coreos-tracker/issues/1791

PR #3014, opened 2024-06-04
s390x: Move copying of 01-secex.ign config from systemd generator to service

nikita-dubrovskii:
Fedora 40 during early startup (initramfs) mounts rootfs as read-only, so systemd-generators fail to modify it:
```
Welcome to Fedora CoreOS 40.20240524.dev.0 dracut-101-1.fc40 (Initramfs)!
[    1.857142] systemd[1]: No hostname configured, using default hostname.
[    1.857175] systemd[1]: Hostname set to <localhost>.
[    1.857230] systemd[1]: Initializing machine ID from random generator.
[    2.010194] systemd[1]: bpf-lsm: LSM BPF program attached
[    2.014388] ln:
[    2.014407] failed to create symbolic link '/var/log'
[    2.014414] : Read-only file system
[    2.014415]
[    2.038544] cp:
[    2.038553] cannot create regular file '/usr/lib/ignition/base.d/01-secex.ign'
```

This moves all config related logic to one place.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -150,7 +150,6 @@ fi
 if [[ "${secure_execution}" = "1" ]]; then
     mkdir -p /run/coreos/
     touch /run/coreos/secure-execution
-    cp /usr/lib/coreos/01-secex.ign /usr/lib/ignition/base.d/01-secex.ign
     # Add dropins to disable Ignition logging for all stages
     stages=("fetch-offline" "fetch" "kargs" "disks" "mount" "files")
     for s in "${stages[@]}"; do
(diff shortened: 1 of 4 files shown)


nikita-dubrovskii:
I'm still looking for the change and reason why rootfs is `ro` now, this PR could serve as a fix for now

jlebon:
This is probably more fallout from systemd v253 now sandboxing generators so they can't write outside the expected paths (see e.g. https://github.com/coreos/fedora-coreos-tracker/issues/1402).

So I think the fix here actually is to do the Ignition config injection as part of a systemd service instead of this generator. We could create a new service for it, or add it to `coreos-secex-ignition-decrypt.sh` which is already conditional on secex and runs before `ignition-fetch-offline.service`. Might be worth renaming the unit and script to e.g. `coreos-secex-ignition-prepare.{service,sh}` since it wouldn't be about just decrypting anymore.

jlebon (approved):

PR #3015, opened 2024-06-06
manifests: Prepare for CentOS Stream 9 -> 10 transition

travier:
manifests: Move NetworkManager-team/teamd to EL9 shared manifest

libteam support is removed in CentOS Stream 10. Move packages related to
this support to a Enterprise Linux 9 only manifest to prepare for that
removal.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1727

---

manifests: Move runc to EL9 shared manifest

runc is not included in CentOS Stream 10. Move the runc package to an
Enterprise Linux 9 only manifest to prepare for that removal.

---

manifests: Move GPU firmwares & kdump tools to EL10 shared manifest

Move packages to an Enterprise Linux 10 only manifest to prepare
for the package split changes:
- GPU firmwares are split into their own package in CentOS Stream 10
- kdump tools are split between kexec-utils, makedumpfile and
  kdump-utils

---

See also: https://github.com/openshift/os/pull/1498

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -13,6 +13,7 @@ include:
   # See https://github.com/coreos/bootupd
   - bootupd.yaml
   - shared-el9.yaml
+  - shared-el10.yaml
 ostree-layers:
   - overlay/05core
   - overlay/08nouveau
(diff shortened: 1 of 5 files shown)


travier:
Draft as some changes have not been confirmed yet.

travier:
Another option for the sgdisk transition to avoid having to set a custom path in the Ignition RPM and in the script here is to install `/usr/libexec/ignition-sgdisk` to `/usr/sbin/sgdisk` only in the initramfs via a dracut module, as we will need a custom dracut module either way to install this vendored version.

jlebon:
The only real use of sgdisk in this repo is for the UUID restamp and moving the GPT backup header. We should be able to do those with `sfdisk`, which supports `--relocate gpt-bak-std`. For the stamping, there's `--disk-id`, but it doesn't natively support generating a random one, so we'd have to generate it ourselves and pass it in.

Overall that use can easily be dropped since it's much more scoped and easier to verify that it's equivalent to the sgdisk bit. Not sure it's worth trying to piggy back on Ignition's sgdisk.

travier:
OK. I don't want to do the gdisk -> fdisk migration in FCOS now so I'll remove that change and I'll install sgdisk in the PATH only in the initramfs for now in SCOS

travier:
I think those changes are ready and should be a NOP for FCOS.

dustymabe:
LGTM - can we merge next Monday once we get the F41/beta `next` built?

travier:
Alright, I've rebased this to re-run the tests. Let's get this in now that we're good with F41?

dustymabe (approved):
LGTM

PR #3019, opened 2024-06-14
overlays: drop `coreos-multipath-trigger.service`

jlebon:
This service was needed in the past to make multipath + LUKS work well. The underlying bug seems to have been fixed now as I can no longer reproduce it in Fedora or RHEL 9.4. Conveniently, this also works around a bug in which that service would sometimes hang because of a bug[[1]] in systemd which is still outstanding in RHEL 9.

Drop it.

We don't have any tests for this yet. Multipath + LUKS currently doesn't work but should be fixed soon[[2]]. A test will be added as part of that work.

[1]: https://github.com/systemd/systemd/issues/29863
[2]: https://github.com/coreos/fedora-coreos-tracker/issues/1728

Fixes: https://issues.redhat.com/browse/OCPBUGS-29325


jlebon:
> Conveniently, this also works around a bug in which that service would sometimes hang because of a bug[https://github.com/systemd/systemd/issues/29863] in systemd which is still outstanding in RHEL 9.

For this reason, we should backport this to at least 4.14. I haven't checked if the bug that motivated this service is still in 9.2 or not, but it doesn't matter because we in effect don't support multipath + LUKS anyway since we've regressed a long time ago (all active streams are affected).

mike-nguyen (approved):
LGTM

PR #3010, opened 2024-06-03
[rawhide] Revert "override: pin packages depending on systemd 256" 

jbtrystram:
This reverts commit https://github.com/jbtrystram/fedora-coreos-config/commit/06a9dfbdf3720409a2b60eb9c20ab349c7c9bd65.

Revert "overrides: pin systemd-255.5-1.fc41"

This reverts commit https://github.com/jbtrystram/fedora-coreos-config/commit/6f6aa3900d6170a4bef57d328c5df44f0c3dfc44.

The SELinux policies have been updated in https://github.com/fedora-selinux/selinux-policy/pull/1979

PR #3020, opened 2024-06-14
[rawhide] overrides: pin selinux-policy-40.20-1.fc41

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -54,6 +54,16 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1735
       type: pin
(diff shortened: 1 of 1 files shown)


gursewak1997:
@aaradhak  Can you add a comment in this PR as to why we need this in addition to the existing pins for the issue?

jlebon:
I don't think we need this. Current rawhide builds are passing tests with the pinned systemd. Moving to systemd v256 (i.e. merging https://github.com/coreos/fedora-coreos-config/pull/3010) is what requires a new selinux-policy (i.e. fixing https://bugzilla.redhat.com/show_bug.cgi?id=22924720).

aaradhak:
yes, this is not required. Thanks for closing it.

PR #3021, opened 2024-06-15
[rawhide] overrides: pin container-selinux-2.231.0-6.fc41

coreosbot-releng:


jlebon:
Hmm, I didn't see container-selinux be implicated here. I don't think we need this as per https://github.com/coreos/fedora-coreos-config/pull/3020#issuecomment-2176877467. Feel free to reopen with more info if you have more information.

PR #3032, opened 2024-06-18
tests/kola: adapt for new pure-RHEL variants

jlebon:
A big part of the new variants added in
https://github.com/openshift/os/pull/1445 is that we only minimally
modify `/etc/os-release`. This means that e.g. `ID` is still `rhel` and
`VERSION_ID` is e.g. `9.4` for the `rhel-9.4` variant. We do still
inject `VARIANT` and `VARIANT_ID` though.

Adapt these library functions here to handle this.

### DIFF
--- a/tests/kola/networking/default-network-behavior-change/test.sh
+++ b/tests/kola/networking/default-network-behavior-change/test.sh
@@ -160,7 +160,7 @@ if [ "$ID" == "fedora" ]; then
     else
         fatal "fail: not operating on expected OS version"
     fi
-elif [[ "${ID_LIKE}" =~ "rhel" ]]; then
+elif [[ "${ID}" = "rhel" ]] || [[ "${ID_LIKE}" =~ "rhel" ]]; then
     # For the version comparison use string substitution to remove the
     # '.` from the version so we can use integer comparison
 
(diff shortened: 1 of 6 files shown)


tests/kola/networking/default-network-behavior-change/test.sh
```diff
@@ -160,7 +160,7 @@ if [ "$ID" == "fedora" ]; then
     else
         fatal "fail: not operating on expected OS version"
     fi
-elif [[ "${ID_LIKE}" =~ "rhel" ]]; then
+elif [[ "${ID}" = "rhel" ]] || [[ "${ID_LIKE}" =~ "rhel" ]]; then
```
marmijo, line 163: Looks like we've mostly used `==` for comparisons like these. I'm fine leaving it as is though because it's identical behavior in bash.

marmijo (approved):
LGTM, but I left one comment

tests/kola/networking/default-network-behavior-change/test.sh
jlebon: Yeah, it's kind of a mess. Would've been nice if bash only supported one kind. I don't think we have any consistency overall in this codebase one way or the other.

PR #3039, opened 2024-06-25
tests/kdump.crash: enable more debugging in kdump service

jbtrystram:
Recent kdump updates started to fail without any error, so this
    will increase the verbosity of the kdump service.
    This will increase the logs size but logs are discarded on
    successful tests so that's fine.
    
    See https://github.com/rhkdump/kdump-utils/issues/15#issuecomment-2188528819

### DIFF
--- a/tests/kola/kdump/crash/config.bu
+++ b/tests/kola/kdump/crash/config.bu
@@ -12,3 +12,8 @@ systemd:
   units:
     - name: kdump.service
       enabled: true
+      dropins:
+        - name: debug.conf
+          contents: |
+            [Service]
+            Environment="debug=1"


tests/kola/kdump/crash/config.bu
```diff
@@ -12,3 +12,10 @@ systemd:
   units:
     - name: kdump.service
       enabled: true
+storage:
```
jlebon: The more canonical way to do this is via the systemd knob also. See the dropins examples in https://coreos.github.io/butane/examples/#systemd-units.

jbtrystram: Thanks ! Somehow I didn't think of it

jbtrystram: fixed

tests/kola/kdump/crash/config.bu
```diff
@@ -12,3 +12,8 @@ systemd:
   units:
     - name: kdump.service
       enabled: true
+      dropins:
+        - name: /etc/systemd/system/kdump.service.d/debug.conf
```
jlebon:
```suggestion
        - name: debug.conf
```

jlebon (approved):
LGTM, assuming you've verified the dropin works as intended.

PR #3042, opened 2024-06-27
overlay/15fcos: fix aleph file and update bootloader for Secure Boot nodes

### LINKED ISSUE #543: Boot fails with "vmlinuz has invalid signature" or "bad shim signature, you need to load the kernel first"
# Current workaround

**See https://github.com/fedora-silverblue/issue-tracker/issues/543#issuecomment-2048350047**

---

# Original issue text

**Describe the bug**
Trying to rebase an existing SB39 to SB40 fails to boot showing `vmlinuz-6.8.1-300.fc40.x86.x64 has invalid signature. you need to load the kernel first` on an Thinkpad T495. I was not able to reproduce the issue on my desktop systems.

coming from deployment:
```
fedora:fedora/39/x86_64/silverblue
                  Version: 39.20240328.0 (2024-03-28T00:39:32Z)
               BaseCommit: a1ac8885d05ca13728d92c2bcf1ded67a7ddb409d657e446a808397366a463b1
             GPGSignature: Valid signature by E8F23996F23218640CB44CBE75CF5AC418B8E74C
```

going for:
```
fedora:fedora/40/x86_64/silverblue
                  Version: 40.20240328.n.0 (2024-03-28T08:09:28Z)
               BaseCommit: f726d0be3361a42a8ac175b08851de67a2d97c9c01ee130a3abcd32720120f9c
```

**What I've tried so far**
- cleanup + `rpm-ostree initramfs --enable` + rebase
- `cleanup -r` (removing all other deployments) + rebase
- rebase to `fedora:fedora/40/x86_64/testing/silverblue` instead, same error except error message points to vmlinuz-6.8.2

**To Reproduce**
1. `rpm-ostree rebase fedora:fedora/40/x86_64/silverblue`
2. `systemctl reboot`

### LINKED ISSUE #1724: aleph file extra attributes gives bootupctl status error
There is some history here, but the aleph file for a specific release of Fedora CoreOS on the `testing` and `next` streams had an extra field in the aleph file:

```
cat /sysroot/.coreos-aleph-version.json 
{
        "build": "39.20231204.1.0",
        "version": "39.20231204.1.0",
        "ref": "fedora/x86_64/coreos/next",
        "ostree-commit": "3b5484230c1f1299bdab9d52b3663468db482dc37d80bc511bd1866c8b88fac8",
        "imgid": "fedora-coreos-39.20231204.1.0-qemu.x86_64.qcow2"
}
sudo bootupctl status
 Component EFI
   Installed: grub2-efi-x64-1:2.06-109.fc39.x86_64,shim-x64-15.6-2.x86_64
   Update: Available: grub2-efi-x64-1:2.06-119.fc40.x86_64,shim-x64-15.8-3.x86_64
 No components are adoptable.
 error: duplicate field `version` at line 3 column 11
```

As far as I can tell this only affected one release of `next` and `testing`:

- `next`
    - 39.20231119.1.0 -> good
    - 39.20231204.1.0 -> bad
    - 39.20240104.1.0 -> good
- `testing`
    - 39.20231119.2.0 -> good
    - 39.20231204.2.1 -> bad
    - 39.20240104.2.0 -> good
- `stable`
    - 39.20231101.3.0 -> good
    - 39.20231119.3.0 -> good
    - 39.20231204.3.3 -> good
    - 39.20240104.3.0 -> good
    - 39.20240112.3.0 -> good

We halted the `20231217` releases because we detected this problem:

- https://github.com/coreos/fedora-coreos-streams/issues/827#issuecomment-1863238034
- https://github.com/coreos/fedora-coreos-streams/issues/830#issuecomment-1863239036
- https://github.com/coreos/fedora-coreos-streams/issues/831#issuecomment-1863238782

We reverted the original change that caused this behavior in https://github.com/coreos/coreos-assembler/pull/3686

I guess I had thought that the problematic commit hadn't actually gone into any releases but apparently it had gone out in the previous set of `testing` and `next` releases (`39.20231204.2.1` and `39.20231204.1.0`).

jbtrystram:
overlay/15fcos: ensure valid aleph file

Due to an ordering mishap, some builds have both a `version` and a
`build` field. This causes bootupctl to fail while parsing the file.

Detect this case, and fix the aleph if necessary by removing the `build`
field.

This should be removed after the next barrier release.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1724

---

overlay/15fcos: upgrade bootloader for Secure Boot-enabled systems

The 6.9 kernel won't boot on systems installed prior to F39, as the shim
is too old.

Add a systemd unit that updates the bootloader on those machines.
Manually handle systems with mirrored ESPs.

See also: https://github.com/coreos/fedora-coreos-tracker/issues/1752
Fixes: https://github.com/fedora-silverblue/issue-tracker/issues/543

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -6,3 +6,6 @@ enable fwupd-refresh.timer
 # Check if wifi firmwares are missing when NetworkManager-wifi is installed
 # https://github.com/coreos/fedora-coreos-tracker/issues/1575
 enable coreos-check-wireless-firmwares.service
+# Strip extraneous field in aleph files to avoid bootupctl failing
+# https://github.com/coreos/fedora-coreos-tracker/issues/1724
+enable coreos-fix-aleph-file.service
--- a/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
+++ b/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
@@ -9,3 +9,6 @@ enable coreos-check-wireless-firmwares.service
 # Strip extraneous field in aleph files to avoid bootupctl failing
 # https://github.com/coreos/fedora-coreos-tracker/issues/1724
 enable coreos-fix-aleph-file.service
+# Upgrade bootloader on Secure Boot-enabled nodes to avoid
+# https://github.com/coreos/fedora-coreos-tracker/issues/1752
+enable coreos-bootupctl-update-secureboot.service
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-bootupctl-update-secureboot.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-bootupctl-update-secureboot.service
@@ -1,0 +1,20 @@
+# Remove after the next barrier release
+
+[Unit]
+Description=Update Bootloader for Secure Boot-enabled Systems
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1752
+ConditionSecurity=uefi-secureboot
+
+# make sure to run after the aleph file is fixed
+# see https://github.com/coreos/fedora-coreos-tracker/issues/1724
+After=coreos-fix-aleph-file.service
+Requires=coreos-fix-aleph-file.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-update-bootloader
+RemainAfterExit=yes
+MountFlags=slave
+
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-aleph-file.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-aleph-file.service
@@ -1,0 +1,14 @@
+# Remove after the next barrier release
+
+[Unit]
+Description=Fix CoreOS Aleph File
+Documentation=https://github.com/coreos/fedora-coreos-tracker/issues/1724
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-fix-aleph-file
+RemainAfterExit=yes
+MountFlags=slave
+
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
+++ b/overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
@@ -1,0 +1,28 @@
+#!/usr/bin/bash
+
+# This script removes the extra `version` field
+# which was shipped in a couple of builds
+# when switching to the `build` field
+# To be removed after the next barrier release.
+# see https://github.com/coreos/fedora-coreos-tracker/issues/1724 for more details
+
+set -euo pipefail
+
+ALEPH_FILE=/sysroot/.coreos-aleph-version.json
+
+if ! jq -e 'has("build") and has("version")' ${ALEPH_FILE}; then
+    echo "Aleph file does not require fixing"
+    exit
+fi
+
+echo "Aleph file is invalid; fixing"
+
+# remount /sysroot as writable
+mount -o rw,remount /sysroot
+
+# remove field "build"
+fixed_aleph=$(jq 'del(.build)' ${ALEPH_FILE})
+
+echo "$fixed_aleph" > ${ALEPH_FILE}
+
+echo "Aleph file is fixed"
--- a/overlay.d/15fcos/usr/libexec/coreos-update-bootloader
+++ b/overlay.d/15fcos/usr/libexec/coreos-update-bootloader
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script update the bootloader using bootupd
+# and also detect RAID-1 setups as those requires
+# extra steps
+
+if [ -e /dev/disk/by-label/EFI-SYSTEM ]; then
+    echo "Found ESP; calling 'bootupctl update'"
+    bootupctl update
+    exit
+fi
+
+# handle RAID case manually since bootupd doesn't support it
+# https://github.com/coreos/bootupd/issues/132
+i=1
+while true; do
+    if [ ! -e /dev/disk/by-label/esp-$i ]; then
+        break
+    fi
+    echo "Found ESP (replica $i); updating"
+    mount /dev/disk/by-label/esp-$i /boot/efi
+    cp -rp /usr/lib/bootupd/updates/EFI /boot/efi
+    umount /boot/efi
+    i=$((i+1))
+done
+sync


overlay.d/15fcos/usr/lib/systemd/system/coreos-bootupctl-update-secureboot.service
```diff
@@ -0,0 +1,18 @@
+# Remove after the next barrier release
+# https://github.com/coreos/fedora-coreos-tracker/issues/1752
+
+[Unit]
+Description=Update Bootloader for secureboot-enabled systems
+ConditionSecurity=uefi-secureboot
+ConditionFirmware=uefi
+# make sure to run after the aleph file is fixed
+# see https://github.com/coreos/fedora-coreos-tracker/issues/1724
+After=coreos-fix-aleph-file.service
```
travier, line 10: It should likely also require it

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-aleph-file.service
```diff
@@ -0,0 +1,13 @@
+# Remove after the next barrier release
+# https://github.com/coreos/fedora-coreos-tracker/issues/1724
+
+[Unit]
+Description=Remove extra attribute from aleph file
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-fix-aleph-file
+RemainAfterExit=yes
```
travier, line 10: This needs to run in its own mount namespace to be able to remount `/sysroot` RW safely.

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
```diff
@@ -0,0 +1,16 @@
...
+
+# This script removes the extra `version` field
+# which was shipped in a couple of builds
+# To be removed after the next barrier release.
+# see https://github.com/coreos/fedora-coreos-tracker/issues/1724 for more details
+
+set -euo pipefail
+
+ALEPH_FILE=/sysroot/.coreos-aleph-version.json
+
```
travier, line 12: Needs a `mount -o rw,remount /sysroot`

travier (dismissed):

travier:
secureboot -> Secure Boot

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-aleph-file.service
jbtrystram: related thought : once we move to composeFS, we won't be able to do that kind of fixes anymore I guess

jbtrystram:
```suggestion
RemainAfterExit=yes
RequireMountFor=/sysroot
```

Is that what you mean ? I don't think so but i don't find any hints in `man systemd.unit`

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
jbtrystram: should I add another `mount -o ro,remount /sysroot` once done ? 

And a stamp in `var` to make sure it runs once ? (not needed if we remove it right after the barrier though)

overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
```diff
@@ -6,3 +6,7 @@ enable fwupd-refresh.timer
 # Check if wifi firmwares are missing when NetworkManager-wifi is installed
 # https://github.com/coreos/fedora-coreos-tracker/issues/1575
 enable coreos-check-wireless-firmwares.service
+# Strip extraneous field in aleph files to avoid bootupctl failing
+# https://github.com/coreos/fedora-coreos-tracker/issues/1724
+enable coreos-fix-aleph-file.service
+
```
jlebon: Minor: extraneous line

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-aleph-file.service
jlebon: He means `MountFlags=slave`.

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
```diff
@@ -0,0 +1,16 @@
...
+# remove field "version"
+fixed_aleph=$(jq 'del(.version)' ${ALEPH_FILE})
```
jlebon: The version key is the canonical one between the two, and the *only* one in newer builds. So we definitely don't want to remove it unconditionally. Let me write up a more detailed flow.

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
jlebon: I would structure the logic like this:

- check if the aleph file has *both* a `build` key *and* a `version` key
- if not, exit
- if yes, remount /sysroot rw
- regenerate aleph, removing `build` key

jlebon: There's no need to remount ro once we're in a private mount namespace. I would skip the stamp file as well, since the code is idempotent and it will only be present on one barrier release.

overlay.d/15fcos/usr/lib/systemd/system/coreos-bootupctl-update-secureboot.service
```diff
@@ -0,0 +1,18 @@
+# Remove after the next barrier release
+# https://github.com/coreos/fedora-coreos-tracker/issues/1752
+
+[Unit]
+Description=Update Bootloader for secureboot-enabled systems
```
jlebon: Secure Boot-enabled

overlay.d/15fcos/usr/lib/systemd/system/coreos-bootupctl-update-secureboot.service
```diff
@@ -0,0 +1,18 @@
...
+ConditionSecurity=uefi-secureboot
+ConditionFirmware=uefi
```
jlebon: I guess that's redundant, but meh... sure!

overlay.d/15fcos/usr/lib/systemd/system/coreos-bootupctl-update-secureboot.service
```diff
@@ -0,0 +1,18 @@
...
+# make sure to run after the aleph file is fixed
+# see https://github.com/coreos/fedora-coreos-tracker/issues/1724
+After=coreos-fix-aleph-file.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/bootupctl update
```
jlebon: So yeah, as discussed, we should probably have this call a script at this point so it can handle the RAID1 case too.

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-aleph-file.service
travier: Yes, it's `MountFlags` sorry.

We will still be able to do that kind of fixes with composefs as this is `/sysroot`, the "real root" which is always writable by definition. Only `/`, the "composefs root" will be truly read only.

jlebon:
In the script, I think let's just have two separate paths entirely for RAID1 and non-RAID1. You can detect if a system is using RAID1 or not by checking if there is a `/dev/disk/by-label/esp-1` and `/dev/disk/by-label/esp-2`. So then:

- if not RAID: just do `bootupctl update` and exit
- if RAID: do https://github.com/coreos/fedora-coreos-tracker/issues/1485#issuecomment-1533350911

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-aleph-file.service
jbtrystram: but then the signature won't match the underlying file ?

jlebon: `/sysroot` is a mount of the root of the filesystem, not part of `/` (and so not covered by the composefs image/signature).

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
jbtrystram: looking at https://github.com/coreos/coreos-assembler/blob/38235218542c39a29255933d3cea64e91754e85f/src/create_disk.sh#L398  the only one present in newer builds should be `build` ?

Also https://github.com/coreos/coreos-assembler/pull/3686/files

jlebon: `create_disk.sh` is mostly dead code at this point in favour of osbuild. The stage that creates the aleph is at https://github.com/osbuild/osbuild/blob/92e5e437947e52ec2cfb463d50e403b1317b4215/stages/org.osbuild.ostree.aleph#L22-L27

jbtrystram:
> In the script, I think let's just have two separate paths entirely for RAID1 and non-RAID1. You can detect if a system is using RAID1 or not by checking if there is a `/dev/disk/by-label/esp-1` and `/dev/disk/by-label/esp-2`. So then:
> 
>     * if not RAID: just do `bootupctl update` and exit
> 
>     * if RAID: do [bootupd fails on mirrored boot disks fedora-coreos-tracker#1485 (comment)](https://github.com/coreos/fedora-coreos-tracker/issues/1485#issuecomment-1533350911)

This won´t support RAID-1 with more than two disks though

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
jbtrystram: updated, let me know what you think

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
jbtrystram: okay that make more sense now :) updated

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
```diff
@@ -0,0 +1,24 @@
...
+# This script removes the extra `version` field
+# which was shipped in a couple of builds
+# To be removed after the next barrier release.
+# see https://github.com/coreos/fedora-coreos-tracker/issues/1724 for more details
+
+set -euo pipefail
+
+ALEPH_FILE=/sysroot/.coreos-aleph-version.json
+
+if ! $(jq 'has("build")' ${ALEPH_FILE}); then
```
jlebon: This needs to check whether *both* the `build` and `version` keys are present.

Basically, the history here is:
- old builds had just `build`
- some builds had both `build` and `version`
- new builds have just `version`

We're trying to fix the set in the middle.

jbtrystram:
Sorry for the multiples force pushes, I'm getting burn by the trailing whitespaces !

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
```diff
@@ -0,0 +1,21 @@
...
+set -euo pipefail
+
+# This script update the bootloader using bootupd
+# and also detect RAID-1 setups as those requires
+# extra steps
+
+
+# bootupd don't support RAID-1 setups
+# https://github.com/coreos/fedora-coreos-tracker/issues/1485
+if test -f /dev/disk/by-label/esp-1 && test -f /dev/disk/by-label/esp-2; then
```
jlebon: They're block devices, not files, so it'd be `-b`, but I'd just do `-e`.

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
```diff
@@ -0,0 +1,21 @@
...
+  mount /dev/disk/by-label/esp-1 /boot/efi
+  cp -rp /usr/lib/bootupd/updates/EFI /boot/efi
+  umount /boot/efi
+  mount /dev/disk/by-label/esp-2 /boot/efi
+  cp -rp /usr/lib/bootupd/updates/EFI /boot/efi
+  umount /boot/efi
```
jlebon: This should probably instead be a loop e.g.

```
i=1
while true; do
  if [ ! -e /dev/disk/by-label/esp-$i" ]; then
    break
  fi
  mount /dev/disk/by-label/esp-$i /boot/efi
  cp -rp /usr/lib/bootupd/updates/EFI /boot/efi
  umount /boot/efi
  i=$((i+1))
done
```

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+# Regular case
+bootupctl update
```
jlebon: This shouldn't run in the RAID case.

OK, how about we reflow this file as:

```bash
if [ -e /dev/disk/by-label/EFI-SYSTEM ]; then
    bootupctl update
    exit
fi

# handle RAID case manually since bootupd doesn't support it
# https://github.com/coreos/bootupd/issues/132
while true; do
    ...
done
```

overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset
```diff
@@ -6,3 +6,9 @@ enable fwupd-refresh.timer
 # Check if wifi firmwares are missing when NetworkManager-wifi is installed
 # https://github.com/coreos/fedora-coreos-tracker/issues/1575
 enable coreos-check-wireless-firmwares.service
+# Strip extraneous field in aleph files to avoid bootupctl failing
+# https://github.com/coreos/fedora-coreos-tracker/issues/1724
+enable coreos-fix-aleph-file.service
+# Upgrade bootloader on secureboot nodes to avoid
```
travier:
```suggestion
# Upgrade bootloader on Secure Boot-enabled nodes to avoid
```

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
```diff
@@ -0,0 +1,24 @@
...
```
travier: I think this should also check that it has a `version`?

https://github.com/coreos/fedora-coreos-config/pull/3042/files#r1657176704

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script update the bootloader using bootupd
+# and also detect RAID-1 setups as those requires
+# extra steps
+
+
+# bootupd don't support RAID-1 setups
```
travier:
```suggestion
# bootupd doesn't support RAID-1 setups
```

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
```diff
@@ -0,0 +1,21 @@
...
+# https://github.com/coreos/fedora-coreos-tracker/issues/1485
+if test -f /dev/disk/by-label/esp-1 && test -f /dev/disk/by-label/esp-2; then
+  mount /dev/disk/by-label/esp-1 /boot/efi
+  cp -rp /usr/lib/bootupd/updates/EFI /boot/efi
```
travier: This is missing `EFI/BOOT` but maybe we don't have that at all in the CoreOS case.

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
```diff
@@ -0,0 +1,21 @@
...
+  umount /boot/efi
+  mount /dev/disk/by-label/esp-2 /boot/efi
+  cp -rp /usr/lib/bootupd/updates/EFI /boot/efi
+  umount /boot/efi
+fi
```
travier, line 12: Probably needs a `sync` at the end

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
jbtrystram: I forgot an exit there yeah. thanks for the suggestion

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
jbtrystram: This is taken from https://github.com/coreos/fedora-coreos-tracker/issues/1485#issuecomment-1533350911 which worked for the user on FCOS

travier: We do, but I wonder if we need it. If we have an older bootloader this means that we've already booted at least once so the EFI entry for Fedora should exists already and thus we should not be using the `BOOT` default path?

Better be safe maybe?

overlay.d/15fcos/usr/libexec/coreos-fix-aleph-file
jbtrystram: ok thanks for the clarification, I thought `build` was just a blip that was introduced then removed.

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
jlebon: Not sure I follow. This is copying everything in `/usr/lib/bootupd/updates/EFI/` to `/boot/efi`. On an RHCOS I have laying around:

```
[root@cosa-devsh EFI]# ls -la /usr/lib/bootupd/updates/EFI
total 0
drwxr-xr-x. 4 root root  32 Jan  1  1970 .
drwxr-xr-x. 3 root root  50 Jan  1  1970 ..
drwx------. 2 root root  42 Jan  1  1970 BOOT
drwx------. 2 root root 122 Jan  1  1970 redhat
```

So it would copy both `BOOT` and `redhat`.

jlebon (approved):
Tested this pretty extensively. Secure Boot/non-Secure boot, RAID/non-RAID, old builds, new builds, builds affected by https://github.com/coreos/fedora-coreos-tracker/issues/1724.

LGTM

jbtrystram:
Thanks @jlebon for pushing it through!

overlay.d/15fcos/usr/libexec/coreos-update-bootloader
travier: Ah yes, my bad

travier:
👍🏻

PR #3046, opened 2024-07-01
denylist: snooze ssh.key until new afterburn release

marmijo:
Extend the snooze on the `coreos.ignition.ssh.key` test failure for another month until a new release of Afterburn can be performed that will include the fix for the failure: https://github.com/coreos/afterburn/pull/1074.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1553


aaradhak (approved):
LGTM

PR #3047, opened 2024-07-02
fasttrack openssh-9.6p1-1.fc40.4

### LINKED ISSUE #1754: CVE-2024-6387: OpenSSH 9.8: regreSSHion: RCE in OpenSSH's server, on glibc-based Linux systems
See:
- https://lwn.net/Articles/980211/
- https://www.qualys.com/2024/07/01/cve-2024-6387/regresshion.txt

> We discovered a vulnerability (a signal handler race condition) in OpenSSH's server (sshd): if a client does not authenticate within LoginGraceTime seconds (120 by default, 600 in old OpenSSH versions), then sshd's SIGALRM handler is called asynchronously, but this signal handler calls various functions that are not async-signal-safe (for
example, syslog()). This race condition affects sshd in its default configuration.

They only have working exploits for i686 right now.

The configuration workarounds are not ideal unfortunately: https://lwn.net/ml/all/4f270df5-2b24-979d-c03f-6d8f3b9d007d@mindrot.org/

travier:
Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1754

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -39,3 +39,21 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1752
       type: pin
+  openssh:
+    evr: 9.6p1-1.fc40.4
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2024-dc89a2e1bf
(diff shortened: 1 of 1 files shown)


cverna (approved):
LGTM

PR #3048, opened 2024-07-02
kola/kernel-replace: adapt for RHCOS and SCOS

c4rt0:
There were instances where [kola/switch-kernel test](https://github.com/coreos/coreos-assembler/issues/1245) would fail.
It was [recommended](https://github.com/coreos/coreos-assembler/issues/1245#issuecomment-2083044615) to remove it ([PR](https://github.com/coreos/coreos-assembler/pull/3825)) and generalize this test, so that it can work with FCOS, RHCOS and SCOS.

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -5,9 +5,6 @@
 ##   # We've seen some OOM when 1024M is used:
 ##   # https://github.com/coreos/fedora-coreos-tracker/issues/1506
 ##   minMemory: 2048
-##   # This test only runs on FCOS due to a problem with skopeo copy on
-##   # RHCOS. See: https://github.com/containers/skopeo/issues/1846
-##   distros: fcos
 ##   # Needs internet access as we fetch files from koji
 ##   # We add the "reprovision" tag here even though we aren't
 ##   # reprovisioning as a hack so that in our pipeline the test
@@ -43,13 +40,28 @@ set -euxo pipefail
 cd $(mktemp -d)
 
 # TODO: It'd be much better to test this via a registry
-image_dir=/var/tmp/fcos
-image=oci:$image_dir
-image_pull=ostree-unverified-image:$image
-tmp_imagedir=/var/tmp/fcos-tmp
-arch=$(arch)
-kver="6.2.9-300.fc38.${arch}"
 
+# define OS ID in order to assign appropriate kernel later
+OS_ID=$(. /etc/os-release; echo $ID)
+image_dir=/var/tmp/coreos
+image=oci:${image_dir}
+image_pull=ostree-unverified-image:$image
+tmp_imagedir=${image_dir}-tmp
+arch=$(arch)
+case "$OS_ID" in
+  rhcos|scos|rhel|centos)
+    kver="5.14.0-472.el9.${arch}"
+    url="https://mirror.stream.centos.org/9-stream/BaseOS/${arch}/os/Packages/kernel"
+    ;;
+  fedora)
+    kver="6.9.8-200.fc40.${arch}"
+    url="https://kojipkgs.fedoraproject.org//packages/kernel/6.9.8/200.fc40/${arch}/kernel"
+    ;;
+  *)
+    echo "Unknown OS_ID: ${OS_ID}"
+    exit 1
+    ;;
+esac
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
     # Take the existing ostree commit, and export it to a container image, then rebase to it.
@@ -80,16 +92,16 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   1)
     # Setup
     # copy the OCI dir to containers-storage for a local build
-    skopeo copy $image containers-storage:localhost/fcos
+    skopeo copy $image containers-storage:localhost/coreos
     rm "${image_dir}" -rf
     td=$(mktemp -d)
     cd ${td}
     version=$(rpm-ostree --version | grep Version)
-cat > Dockerfile << EOF
-FROM localhost/fcos
-RUN rpm-ostree cliwrap install-to-root /
+    cat > Containerfile << EOF
+FROM localhost/coreos
 RUN rpm-ostree override replace \
-    https://koji.fedoraproject.org/koji/buildinfo?buildID=2178613 && \
+    $url-{,core-,modules-,modules-core-,modules-extra-}$kver.rpm && \
+    rpm-ostree cleanup -m && \
     ostree container commit
 EOF
     # Older podman found in RHEL8 blows up without /etc/resolv.conf
@@ -101,12 +113,12 @@ EOF
         3.*) touched_resolv_conf=1; touch /etc/resolv.conf;;
       esac
     fi
-    podman build --net=host -t localhost/fcos-derived --squash .
+    podman build --net=host -t localhost/coreos-derived --squash .
     if test "${touched_resolv_conf}" -eq 1; then
       rm -vf /etc/resolv.conf
     fi
     derived=oci:$image_dir:derived
-    skopeo copy containers-storage:localhost/fcos-derived $derived
+    skopeo copy containers-storage:localhost/coreos-derived $derived
     rpm-ostree --version
     rpm-ostree rebase ostree-unverified-image:$derived
     ostree container image list --repo=/ostree/repo
@@ -118,7 +130,14 @@ EOF
     if test "$un" != "$kver"; then
       echo "Expected kernel $kver but found $un"
       exit 1
+    else
+      echo "Kernel switch to $un was successful"
     fi
     test -f /usr/lib/modules/$kver/initramfs.img
     test -f /usr/lib/modules/$kver/vmlinuz
+    ;;
+  *)
+    echo "Unknown AUTOPKGTEST_REBOOT_MARK: ${AUTOPKGTEST_REBOOT_MARK:-}"
+    exit 1
+    ;;
 esac


tests/kola/rpm-ostree/kernel-replace
```diff
@@ -5,9 +5,7 @@
 ##   # We've seen some OOM when 1024M is used:
 ##   # https://github.com/coreos/fedora-coreos-tracker/issues/1506
 ##   minMemory: 2048
-##   # This test only runs on FCOS due to a problem with skopeo copy on
-##   # RHCOS. See: https://github.com/containers/skopeo/issues/1846
-##   distros: fcos
+##   distros: "fcos rhcos scos"
```
jbtrystram: just omit the tag would work from what i understand ?

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,10 +41,13 @@ set -euxo pipefail
...
 # TODO: It'd be much better to test this via a registry
-image_dir=/var/tmp/fcos
+
+OS_ID=$(grep '^ID=' /etc/os-release | cut -d "=" -f2)
+
+image_dir=/var/tmp/$OS_ID
 image=oci:$image_dir
 image_pull=ostree-unverified-image:$image
-tmp_imagedir=/var/tmp/fcos-tmp
+tmp_imagedir=/var/tmp/$OS_ID-tmp
```
jbtrystram: shellcheck [will complain](https://www.shellcheck.net/wiki/SC2086) :)

jbtrystram (dismissed):
Shellcheck will complain other than this it looks good. 
Could you run the test locally on RHCOS ?

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,10 +41,13 @@ set -euxo pipefail
 cd $(mktemp -d)
 
 # TODO: It'd be much better to test this via a registry
-image_dir=/var/tmp/fcos
+
+OS_ID=$(grep '^ID=' /etc/os-release | cut -d "=" -f2)
```
jlebon: How we name the image dir and built images is purely cosmetic. Instead of this, we could s/fcos/coreos/ instead to make it more generic.

c4rt0: That's true. If however I will completely remove it, in [here](https://github.com/coreos/fedora-coreos-config/pull/3048/files#diff-ffd616bc9607b8abd660319a3679b1afa2d949dc673e752ffd58a6160eaa7a1eR106) we would have podman building `localhost/coreos-derived` - it feels odd, but is it? 
I changed the [image_dir](https://github.com/coreos/fedora-coreos-config/pull/3048/files#diff-ffd616bc9607b8abd660319a3679b1afa2d949dc673e752ffd58a6160eaa7a1eR48) according to this suggestion.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Would it? I'm  not sure to be completely honest.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Modified  👍🏻

c4rt0:
> Could you run the test locally on RHCOS ?

Just tried it now, it's failing. I'm now investigating:

```
=== RUN   ext.config.shared.rpm-ostree.kernel-replace
systemctl status kola-runext.service:
× kola-runext.service
     Loaded: loaded (/etc/systemd/system/kola-runext.service; static)
     Active: failed (Result: exit-code) since Fri 2024-07-05 18:34:49 UTC; 399ms ago
   Duration: 36.717s
    Process: 1266 ExecStart=/usr/local/bin/kola-runext-kernel-replace (code=exited, status=1/FAILURE)
   Main PID: 1266 (code=exited, status=1/FAILURE)
        CPU: 33.418s

Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1270]: ++ grep '^ID=' /etc/os-release
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + OS_ID=rhcos
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + image_dir=/var/tmp/coreos
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + image=oci:/var/tmp/coreos
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + image_pull=ostree-unverified-image:oci:/var/tmp/coreos
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + tmp_imagedir=/var/tmp/coreos-tmp
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1273]: ++ arch
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + arch=x86_64
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + kver=6.2.9-300.fc38.x86_64
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + case "${AUTOPKGTEST_REBOOT_MARK:-}" in
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1266]: + skopeo copy oci:/var/tmp/coreos containers-storage:localhost/rhcos
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1274]: time="2024-07-05T18:34:13Z" level=info msg="Not using native diff for overlay, this may cause degraded performance for building images: kernel has CONFIG_OVERLAY_FS_REDIRECT_DIR enabled"
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1274]: Getting image source signatures
Jul 05 18:34:13 qemu0 kola-runext-kernel-replace[1274]: Copying blob sha256:5d033c3086bc3b8a75fe007c50fa48d18cdea48560bd6d6b2b359d79397bb167
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1274]: Copying config sha256:5b8ccce2c709b5c14f75df3761da4eccfcef89d39165711f14c71196df75b28c
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1274]: Writing manifest to image destination
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + rm /var/tmp/coreos -rf
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1307]: ++ mktemp -d
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + td=/tmp/tmp.q2EfPgCVdW
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + cd /tmp/tmp.q2EfPgCVdW
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1310]: ++ grep Version
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1309]: ++ rpm-ostree --version
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + version=' Version: '\''2024.3'\'''
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + cat
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + touched_resolv_conf=0
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + test '!' -f /etc/resolv.conf
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1266]: + podman build --net=host -t localhost/rhcos-derived --squash .
Jul 05 18:34:48 qemu0 podman[1313]: 2024-07-05 18:34:48.253248775 +0000 UTC m=+0.050040248 system refresh
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1313]: STEP 1/3: FROM localhost/rhcos
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1313]: STEP 2/3: RUN rpm-ostree cliwrap install-to-root /
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1345]: Successfully enabled cliwrap for /
Jul 05 18:34:48 qemu0 kola-runext-kernel-replace[1313]: STEP 3/3: RUN rpm-ostree override replace     https://koji.fedoraproject.org/koji/buildinfo?buildID=2178613 &&     ostree container commit
Jul 05 18:34:49 qemu0 kola-runext-kernel-replace[1397]: Downloading https://koji.fedoraproject.org/koji/buildinfo?buildID=2178613...done
Jul 05 18:34:49 qemu0 kola-runext-kernel-replace[1397]: (rpm-ostree override replace:2): libdnf-WARNING **: 18:34:49.701: failed to read RPM: /tmp/.tmpysF0mq/13.rpm: not a rpm, skipping
Jul 05 18:34:49 qemu0 kola-runext-kernel-replace[1397]: error: Invalid RPM file: /tmp/.tmpysF0mq/13.rpm
Jul 05 18:34:49 qemu0 podman[1313]: 2024-07-05 18:34:49.787637333 +0000 UTC m=+1.584428810 image build  
Jul 05 18:34:49 qemu0 kola-runext-kernel-replace[1313]: Error: building at STEP "RUN rpm-ostree override replace     https://koji.fedoraproject.org/koji/buildinfo?buildID=2178613 &&     ostree container commit": while running runtime: exit status 1
Jul 05 18:34:49 qemu0 systemd[1]: kola-runext.service: Main process exited, code=exited, status=1/FAILURE
Jul 05 18:34:49 qemu0 systemd[1]: kola-runext.service: Failed with result 'exit-code'.
Jul 05 18:34:49 qemu0 systemd[1]: kola-runext.service: Consumed 33.418s CPU time.
--- FAIL: ext.config.shared.rpm-ostree.kernel-replace (171.99s)
        cluster.go:162: Error: Unit kola-runext.service exited with code 1
        cluster.go:162: 2024-07-05T18:34:50Z cli: Unit kola-runext.service exited with code 1
        harness.go:1265: kolet failed: : kolet run-test-unit failed: Process exited with status 1
FAIL, output in tmp/kola
```

tests/kola/rpm-ostree/kernel-replace
jlebon: > That's true. If however I will completely remove it, in [here](https://github.com/coreos/fedora-coreos-config/pull/3048/files#diff-ffd616bc9607b8abd660319a3679b1afa2d949dc673e752ffd58a6160eaa7a1eR106) we would have podman building `localhost/coreos-derived` - it feels odd, but is it?

It's all CoreOS, so it's appropriate. :) There's not much value in doing extra work just to make temporary filenames more specific than they need to be.

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -80,13 +82,13 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
...
+    skopeo copy $image containers-storage:localhost/${OS_ID}
     rm "${image_dir}" -rf
     td=$(mktemp -d)
     cd ${td}
     version=$(rpm-ostree --version | grep Version)
-cat > Dockerfile << EOF
-FROM localhost/fcos
+cat > Containerfile << EOF
+FROM localhost/${OS_ID}
 RUN rpm-ostree cliwrap install-to-root /
```
jlebon: We should drop this now as per https://github.com/openshift/openshift-docs/pull/77066.

jlebon:
Trying to replace the RHCOS kernel with a Fedora one in theory should work, but I think we can get closer to what users are likely to do by replacing with a CentOS Stream kernel instead. So basically:
- FCOS: replace with Fedora kernel
- RHCOS and SCOS: replace with CentOS Stream kernel

For Fedora, the test currently pins to a specific Koji build, which is stale now. We should bump it. Once we have dnf in all the streams, we should probably change that actually to just dynamically always take the latest kernel that's != to the kernel embedded in the container image (similarly for RHCOS/SCOS too).

tests/kola/rpm-ostree/kernel-replace
c4rt0: I tried to drop this, and replace with the `rpm-ostree override replace` from centos, but it's failing for now during download. I will get back to it after the weekend.

c4rt0:
I ran this on both `rhcos` (4.17) and the latest `fcos` build. Test passes as expected.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Removed the tags, it works indeed :)

tests/kola/rpm-ostree/kernel-replace
c4rt0: Thanks for this suggestion @jlebon. I think I implemented it as expected, I would be grateful for some feedback here though.

c4rt0:
> Trying to replace the RHCOS kernel with a Fedora one in theory should work, but I think we can get closer to what users are likely to do by replacing with a CentOS Stream kernel instead. So basically:
> 
>     * FCOS: replace with Fedora kernel
> 
>     * RHCOS and SCOS: replace with CentOS Stream kernel

That's now done.

> For Fedora, the test currently pins to a specific Koji build, which is stale now. We should bump it. 


That's also done.

> Once we have dnf in all the streams, we should probably change that actually to just dynamically always take the latest kernel that's != to the kernel embedded in the container image (similarly for RHCOS/SCOS too).

I can look into this in the future.

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -5,9 +5,7 @@
 ##   # We've seen some OOM when 1024M is used:
 ##   # https://github.com/coreos/fedora-coreos-tracker/issues/1506
 ##   minMemory: 2048
-##   # This test only runs on FCOS due to a problem with skopeo copy on
-##   # RHCOS. See: https://github.com/containers/skopeo/issues/1846
-##   distros: fcos
+##   # distros: "fcos rhcos scos"
```
jlebon: Drop

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,13 +41,18 @@ set -euxo pipefail
 cd $(mktemp -d)
 
 # TODO: It'd be much better to test this via a registry
-image_dir=/var/tmp/fcos
-image=oci:$image_dir
+
+# define OS ID in order to assign appropriate kernel later
+OS_ID=$(grep '^ID=' /etc/os-release | cut -d "=" -f2 | cut -d '"' -f2)
```
jlebon: `/etc/os-release` is meant to be shell-script compatible. Could just source it (`. /etc/os-release`) and use `$ID` below, or use e.g. `OS_ID=$(. /etc/os-release; echo $ID)`.

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,13 +41,18 @@ set -euxo pipefail
...
+image_dir=/var/tmp/coreos
+image=oci:${image_dir}
 image_pull=ostree-unverified-image:$image
-tmp_imagedir=/var/tmp/fcos-tmp
+tmp_imagedir=${image_dir}-tmp
 arch=$(arch)
 kver="6.2.9-300.fc38.${arch}"
-
+echo $(uname -r)
+echo "AUTOPKGTEST_REBOOT_MARK: ${AUTOPKGTEST_REBOOT_MARK:-}"
```
jlebon: Leftover debugging?

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,13 +41,18 @@ set -euxo pipefail
...
+# define OS ID in order to assign appropriate kernel later
+OS_ID=$(grep '^ID=' /etc/os-release | cut -d "=" -f2 | cut -d '"' -f2)
+
+image_dir=/var/tmp/coreos
+image=oci:${image_dir}
 image_pull=ostree-unverified-image:$image
-tmp_imagedir=/var/tmp/fcos-tmp
+tmp_imagedir=${image_dir}-tmp
 arch=$(arch)
 kver="6.2.9-300.fc38.${arch}"
```
jlebon: This needs to be updated. It needs to match the target kernel version we're installing. I would probably do something like:

```bash
case "$ID" in
rhcos|scos|rhel|centos)
  kver=...
  url=...
  ;;
fedora)
  kver=...
  url=...
  ;;
esac
```

(This also fixes looking for the regular `rhel`/`centos` IDs so it works with the new variants in https://github.com/openshift/os/pull/1445.)

And then lower down, you don't have to be conditional when rendering the Containerfile:

```bash
cat Containerfile << EOF
FROM localhost/coreos
RUN rpm-ostree override replace \
    $url-{,core-,modules-,modules-core-,modules-extra-}$kver.x86_64.rpm && \
    rpm-ostree cleanup -m && \
    ostree container commit
EOF
```

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -115,10 +128,18 @@ EOF
     ;;
   2)
     un=$(uname -r)
-    if test "$un" != "$kver"; then
-      echo "Expected kernel $kver but found $un"
+    if test "$un" == "$kver"; then
```
jlebon: And flip this back.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Ok, this is definitely a much better solution.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Yes! Dropping, thanks.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Sure.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Again, much better solution than a previous one. 🙏🏻

c4rt0:
Again, latest changes successfully passed test both on `fcos` and `rhcos`. Thank you for all of the suggestions!

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -118,7 +127,15 @@ EOF
     if test "$un" != "$kver"; then
       echo "Expected kernel $kver but found $un"
       exit 1
+    else
+     echo "Kernel switch to $un was successfull"
+     exit 0
```
jlebon: This is skipping some more tests below.

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,13 +40,25 @@ set -euxo pipefail
...
-kver="6.2.9-300.fc38.${arch}"
-
+# kver="6.2.9-300.fc38.${arch}"
+case "$OS_ID" in
+  rhcos|scos|rhel|centos)
+    kver="5.14.0-472.el9.${arch}"
+    url="https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/kernel"
+    ;;
+  fedora)
+    kver="6.10.0-0.rc7.58.fc41.${arch}"
```
jlebon: Probably safer to pin on a non-rawhide version. E.g. let's use the golden kernel version from f40?

```
$ koji latest-pkg f40 kernel
Build                                     Tag                   Built by
----------------------------------------  --------------------  ----------------
kernel-6.8.5-301.fc40                     f40                   jforbes
```

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,13 +40,25 @@ set -euxo pipefail
...
+    url="https://kojipkgs.fedoraproject.org//packages/kernel/6.10.0/0.rc7.58.fc41/x86_64/kernel"
```
jlebon: This URL and the CentOS one hardcode x86_64.

jlebon (commented):
Suggested commit title:

> kola/kernel-replace: adapt for RHCOS and SCOS

For the commit message, I think in general we should avoid just linking to some other URL that has all the context. Let's inline that context, and provide the URL only as an additional reference in the footer.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Good point, I will use the [kernel-6.9.8-200.fc40](https://koji.fedoraproject.org/koji/search?terms=kernel-6.9.8-200.fc40&type=build&match=exact). I hope that's the one you're referring to.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Removed.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Updated

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -118,7 +126,14 @@ EOF
     if test "$un" != "$kver"; then
       echo "Expected kernel $kver but found $un"
       exit 1
+    else
+      echo "Kernel switch to $un was successfull"
```
jlebon: Minor/optional:

```suggestion
      echo "Kernel switch to $un was successful"
```

jlebon (commented):
LGTM overall. Thoughts on suggested commit title in https://github.com/coreos/fedora-coreos-config/pull/3048#pullrequestreview-2170022913 ? I think mentioning `OS_ID` there isn't really capturing it.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Thanks :)

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -43,13 +40,24 @@ set -euxo pipefail
...
-
+case "$OS_ID" in
+  rhcos|scos|rhel|centos)
+    kver="5.14.0-472.el9.${arch}"
+    url="https://mirror.stream.centos.org/9-stream/BaseOS/${arch}/os/Packages/kernel"
+    ;;
+  fedora)
+    kver="6.9.8-200.fc40.${arch}"
+    url="https://kojipkgs.fedoraproject.org//packages/kernel/6.9.8/200.fc40/${arch}/kernel"
+    ;;
```
jbtrystram, line 59: I would add a default case statement to error early if $OS_ID happen to be unknown

jbtrystram (dismissed):
LGTM overall. I am a bit concerned by the hard-coded kernel package URLs, won't they go stale eventually, breaking the test? 

I don't have a good alternative to suggest though, except adding an item to the rebase checklist..

tests/kola/rpm-ostree/kernel-replace
c4rt0: Good idea, thanks.

c4rt0:
> LGTM overall. I am a bit concerned by the hard-coded kernel package URLs, won't they go stale eventually, breaking the test?
> 
> I don't have a good alternative to suggest though, except adding an item to the rebase checklist..

I agree, @jlebon also talked about it here: https://github.com/coreos/fedora-coreos-config/pull/3048#issuecomment-2211282873. Let's wait with that until we have DNF in all of the streams.

tests/kola/rpm-ostree/kernel-replace
c4rt0: Implemented.

jbtrystram (approved):

jlebon (approved):

PR #3051, opened 2024-07-03

coreosbot-releng:


jlebon:
We can't merge this until the base lockfiles are updated with the new openssh.

jlebon (approved):

PR #3052, opened 2024-07-03
Manifests: includes kdump subpackages in f40

jbtrystram:
Kdump split out into subpackages in rawhide a while back [1] But this split was just pushed to F40 as well with [2].

Also include the subpackages for releases > 40

[1] https://github.com/coreos/fedora-coreos-config/commit/437dc2a3f9a8207eb639e487500466a762d8888a
[2] https://bodhi.fedoraproject.org/updates/FEDORA-2024-d70dfd026e

----

@coiby was the push of the sub-packages to F40 intended ? It's quite unusual to do that in an active release


manifests/user-experience.yaml
```diff
@@ -34,6 +34,8 @@ packages:
   # kdump support
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
   - kexec-tools
+  - makedumpfile
+  - kdump-utils
```
jlebon: This file is [inherited by RHCOS](https://github.com/openshift/os/blob/2467e5e2029dd5e387a8e92ac3b8555f56471130/common.yaml#L8), so we need to check if the split has happened there.

If not, then we'll need to put this in `fedora-coreos-base.yaml` instead for now (you can add a comment that it should be `user-experience.yaml` but it's here because the split hasn't happened in RHEL yet).

manifests/fedora-coreos.yaml
```diff
@@ -33,9 +33,6 @@ conditional-include:
   - if: releasever == 39
     # Checks for breaking changes that came with Podman v5.
     include: podman-v5.yaml
```
jlebon: Aside: looks like we have a bunch of f39-related cruft we could remove from here. Feel free to do that in this PR as a second commit (since you have to repush anyway).

jlebon:
Also, super minor: s/Manifests/manifests/ in the commit title

manifests/user-experience.yaml
jbtrystram: good catch ! This is indeed not in c9s/rhcos

manifests/fedora-coreos.yaml
```diff
@@ -20,19 +20,6 @@ conditional-include:
   - if: basearch != "s390x"
     # And remove some cruft from grub2
     include: grub2-removals.yaml
-  - if: releasever == 39
-    # passwd RPM was obsoleted by shadow-utils in F40+, but we need to keep
-    # including it in F39. Remove this when we are on F40+ on all streams.
-    include: passwd.yaml
```
jlebon, line 26: Need to delete this file.

manifests/fedora-coreos.yaml
```diff
@@ -20,19 +20,6 @@ conditional-include:
...
-  - if: releasever == 39
-    # No longer need CNI plugins in f40 since Podman dropped
-    # support for CNI networking entirely in Podmanv5
-    # Remove this once we have Podman v5 in f39.
-    # xref: https://github.com/coreos/fedora-coreos-tracker/issues/1629
-    include: cni-plugins.yaml
```
jlebon, line 32: Need to delete this file.

jlebon (commented):
Missing lockfile updates for aarch64.

For the second commit title, how about:

> manifests: cleanup f39 conditional includes

jlebon (approved):
Second "fixup" commit should be squashed into first commit. LGTM otherwise!

PR #3053, opened 2024-07-05
denylist: add ext.kdump.crash for aarch64 and remove for ppc64le

jbtrystram:
See https://github.com/rhkdump/kdump-utils/issues/22

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -16,9 +16,3 @@
   warn: true
   arches:
     - ppc64le
-- pattern: ext.config.kdump.crash
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1750
-  warn: true
-  snooze: 2024-07-15
-  arches:
-    - ppc64le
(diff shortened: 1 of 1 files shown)


kola-denylist.yaml
```diff
@@ -22,3 +22,5 @@
   snooze: 2024-07-15
   arches:
     - ppc64le
+    # see https://github.com/rhkdump/kdump-utils/issues/22
+    - aarch64
```
jlebon, line 24: Since it has its own tracker, let's make this a separate denylist entry. We might be able to get rid of the ppc64le one: https://github.com/coreos/fedora-coreos-tracker/issues/1750#issuecomment-2210949765

jbtrystram:
Needs https://github.com/coreos/fedora-coreos-config/pull/3054

yasminvalim (approved):
lgtm

jlebon (approved):
LGTM. Were you able to verify that ppc64le now passes?

jbtrystram:
> LGTM. Were you able to verify that ppc64le now passes?

--- PASS: ext.config.kdump.crash (273.04s)
PASS, output in tmp/kola

jlebon (approved):

PR #3054, opened 2024-07-05
Revert barrier logic to update bootloader on Secure Boot systems and unpin kernel

jlebon:

See also: https://github.com/coreos/fedora-coreos-tracker/issues/1752


jbtrystram (approved):
LGTM

PR #3056, opened 2024-07-09
overrides: pin selinux-policy-40.20-1.fc40

coreosbot-releng:


gursewak1997 (approved):

jlebon:
CI failure is from RPM diff detecting a downgrade, which is expected here.

PR #3057, opened 2024-07-10
[rawhide] overrides: pin json-glib and selinux-policy

jlebon:
A lot of things are pinned right now because of a bunch of SELinux issues.

As a result, we haven't had a new rawhide build in almost a month, and this is starting to affect the reliability of Bodhi testing. Let's pin the problematic packages for now to get a build out.


yasminvalim (approved):
LGTM ✅

jlebon:
CI failing on downgrade, which is expected.

PR #3060, opened 2024-07-15
Denylist: add ext.config.shared.podman.rootless-systemd

c4rt0:
The test currently fails to fetch F38 metadata

Tracker issue: coreos/fedora-coreos-tracker#1760
Also see: https://pagure.io/releng/issue/12204


marmijo:
IIUIC, this test is only failing on RHCOS 4.14. I'll close this in favor of: https://github.com/openshift/os/pull/1550

PR #3061, opened 2024-07-15
Updates for the RHCOS Pipeline migration to an ITUP Cluster 

marmijo:
Use the fedora.repo file defined in fedora-coreos-config to set up the container for the `podman.rootless-systemd` and `ntp.chrony.dhcp-propagation` tests. This will force packages to be downloaded from dl.fedoraproject.org, as specified in the FCOS file. The ITUP cluster, being used by the RHCOS pipeline, requires all outbound connections to be specified in a Firewall Egress file, and this will ensure the same connection will always be used.


jlebon (approved):
This is related to https://github.com/openshift/os/issues/1546. In the future I think we shouldn't be using Fedora content in these containers when testing RHCOS/SCOS.

PR #3065, opened 2024-07-16
denylist: add kdump.crash.ssh for aarch64

jbtrystram:
This test was recently added in https://github.com/coreos/coreos-assembler/pull/3829 but hit the same issue as `ext.kdump.crash`.
This should be fixed in kexec-tools-2.0.28-12.fc40, so extending the snooze for 5 days only as it should hit stable soon.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -17,8 +17,14 @@
   arches:
     - ppc64le
 - pattern: ext.config.kdump.crash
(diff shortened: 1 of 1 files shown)


c4rt0 (commented):
Only snoozing it for 4 days? 
edit: sorry - I only now see it's intended for such short period of time.

c4rt0 (approved):
/lgtm

PR #3067, opened 2024-07-16
overrides: fast-track kexec-tools-2.0.28-12.fc40

coreosbot-releng:


jbtrystram (approved):

jlebon (approved):
Thanks!

PR #3072, opened 2024-07-22
Draft: Test selinux-policy-41.10-1.fc41

zpytela:

PR #3079, opened 2024-08-01
manifest-lock.overrides: Drop selinux-policy & systemd pins

travier:
manifest-lock.overrides: Drop selinux-policy pin

---

manifest-lock.overrides: Drop systemd 255 pin

---

Let's see if we've made enough progress on:
- https://github.com/coreos/fedora-coreos-tracker/issues/1735
- https://bugzilla.redhat.com/show_bug.cgi?id=2292472


travier:
This is an updated version of https://github.com/coreos/fedora-coreos-config/pull/3010

c4rt0 (approved):
/lgtm

travier:
With this, if I run `cosa run --devshell-console`, then I get a shell, and I can also ssh to the system from itself.
If I run `cosa run --devshell`, kola can not connect to the system via ssh.

travier:
Ah, I had a local override. Doing a new build.

aaradhak:
Recently we did a test with the updated copr build for selinux-policy that failed , adding here for reference: https://github.com/coreos/fedora-coreos-config/pull/3072

marmijo:
Closing this with some follow up info:
- the selinux-policy issue was resolved: https://github.com/coreos/fedora-coreos-tracker/issues/1735
  - pin removed from `rawhide`
    -  https://github.com/coreos/fedora-coreos-config/pull/3102
  - pin removed from `branched`
    -  https://github.com/coreos/fedora-coreos-config/pull/3092
  
-  A systemd-256 issue is still ongoing, so the pin on systemd-255 is still necessary
    -  potentially: https://github.com/coreos/fedora-coreos-tracker/issues/1775

PR #3080, opened 2024-08-10
overrides: unpin selinux-policy-40.20-1.fc40

c4rt0:
With the latest build of selinux-policy (selinux-policy-40.27-1.fc40), the issue with `no serial console login` is fixed.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1758

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -19,13 +19,3 @@ packages:
     metadata:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2024-eeaf090672
       type: fast-track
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Thanks for testing this! To confirm, e.g. `cosa run -c` works as expected?

Edit: OK I see https://github.com/coreos/fedora-coreos-tracker/issues/1758#issuecomment-2281911372 now. So actually, on top of just unpinning, we need to fast-track the selinux-policy in that Bodhi update. So in effect, this becomes just changing the version string in the overrides file and changing `pin` to `fast-track` (and adding a `bodhi` key).

PR #3094, opened 2024-08-14
denylist: snooze ext.config.extensions.package for aarch64

c4rt0:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1776

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -16,3 +16,8 @@
   warn: true
   arches:
     - ppc64le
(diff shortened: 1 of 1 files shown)


jlebon (approved):

jlebon:
Actually, testing-devel aarch64 isn't failing on this, because it pulls a different vim version (this test is sensitive to the vim-* packages in the base).

So open this PR, but directly against the stable branch instead.

PR #3093, opened 2024-08-13
[rawhide] manifest: Rawhide is now F42

marmijo:
Fedora 41 branched from rawhide, so rawhide is now F42

PR #3102, opened 2024-08-19
[rawhide] overrides: remove selinux-policy pin

marmijo:
The issue related to this pin seems to have been resolved and keeping the pin is causing build failures in rawhide.

Wait for https://github.com/coreos/fedora-coreos-config/pull/3101 to merge so CI can run smoothly past the build stage.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -14,16 +14,6 @@ packages:
     metadata:
       reason: https://bugzilla.redhat.com/show_bug.cgi?id=2297094
       type: pin
-  selinux-policy:
-    evra: 41.2-1.fc41.noarch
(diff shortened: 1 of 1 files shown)


aaradhak (approved):

gursewak1997 (approved):

marmijo:
CI is failing: 
```
FAIL: ext.config.files.validate-symlinks
fatal 'Error: /usr/lib/bootc/storage symlink to /sysroot/ostree/bootc/storage which does not exist'
```
I think this failure is related to the fact that rawhide now has some f42 packages. I think I need to close this and merge the commit into https://github.com/coreos/fedora-coreos-config/pull/3093 to debug further.

PR #3105, opened 2024-08-20
tests/kernel-replace: use `dnf repoquery` to get kernel version

jlebon:
When testing from the c9s repos, don't hardcode a kernel version since it'll eventually cycle out of the mirrors. Instead, just query the repo to get the latest kernel.

Since we need the info we queried across reboots and we don't want to query it each time (and possibly get different answers), persist it in `/var` instead.

Ideally we'd do this for FCOS too, but we need to wait until the f41 rebase, which will bring dnf. (Or... could do it in a toolbox, but that seems overkill.) But anyway for now, the kernel from the stable Fedora repo shouldn't ever be GC'ed.

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -48,20 +48,37 @@ image=oci:${image_dir}
 image_pull=ostree-unverified-image:$image
 tmp_imagedir=${image_dir}-tmp
 arch=$(arch)
-case "$OS_ID" in
-  rhcos|scos|rhel|centos)
-    kver="5.14.0-472.el9.${arch}"
-    url="https://mirror.stream.centos.org/9-stream/BaseOS/${arch}/os/Packages/kernel"
-    ;;
-  fedora)
-    kver="6.9.8-200.fc40.${arch}"
-    url="https://kojipkgs.fedoraproject.org//packages/kernel/6.9.8/200.fc40/${arch}/kernel"
-    ;;
-  *)
-    echo "Unknown OS_ID: ${OS_ID}"
-    exit 1
-    ;;
-esac
+
+# we only want to query repos once but need the info on multiple boots, so write out state to /var
+if [ ! -f /var/kola-kernel.evr ]; then
+  case "$OS_ID" in
+    rhcos|scos|rhel|centos)
+      c9s_mirror="https://mirror.stream.centos.org/9-stream/BaseOS/${arch}/os"
+      # we're probably already on CentOS and running the latest kernel. so
+      # here we need to instead pick whatever the latest that's *not* the same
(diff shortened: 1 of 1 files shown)


tests/kola/rpm-ostree/kernel-replace
```diff
@@ -48,20 +48,37 @@ image=oci:${image_dir}
...
+if [ ! -f /var/kola-kernel.evr ]; then
+  case "$OS_ID" in
+    rhcos|scos|rhel|centos)
+      c9s_mirror="https://mirror.stream.centos.org/9-stream/BaseOS/${arch}/os"
+      # we're probably already on CentOS and running the latest kernel. so
+      # here we need to instead pick whatever the latest that's *not* the same
+      kver=$(dnf repoquery kernel --qf '%{EVR}' --repofrompath=tmp,"${c9s_mirror}" --disablerepo '*' --enablerepo tmp | \
+                grep -v "$(rpm -q kernel --qf '%{EVR}')" | tail -n1)
+      kver="${kver}.${arch}"
+      url="${c9s_mirror}/Packages/kernel"
```
marmijo, line 62: I'm not super familiar with this, but I seem to get a 404 when I visit this URL. I see kernel packages located one level up `../Packages`.

tests/kola/rpm-ostree/kernel-replace
```diff
@@ -48,20 +48,37 @@ image=oci:${image_dir}
...
+      ;;
+    fedora)
+      # to get the version to use: koji latest-pkg f40 kernel
+      # XXX: once we can rely on dnf in FCOS, use repoquery also here
+      kver="6.8.5-301.fc40.${arch}"
+      url="https://kojipkgs.fedoraproject.org//packages/kernel/6.8.5/301.fc40/${arch}/kernel"
```
marmijo, line 68:
```suggestion
      url="https://kojipkgs.fedoraproject.org/packages/kernel/6.8.5/301.fc40/${arch}/kernel"
```

marmijo: Same here too. I get a "Not Found" and see kernel packages one level up `../${arch}`

tests/kola/rpm-ostree/kernel-replace
jlebon: Yeah, it's weird but it's actually appended to in the Containerfile. So e.g. the full URL would be `.../Packages/kernel-core-$kver.rpm`. Ideally eventually here we should just change things to have the Containerfile install from the repo directly using dnf instead of dealing with URLs.

tests/kola/rpm-ostree/kernel-replace
jlebon: Similarly here.

marmijo (approved):
LGTM

aaradhak (approved):

jlebon:
4.17 backport in https://github.com/coreos/fedora-coreos-config/pull/3106

PR #3106, opened 2024-08-20
[rhcos-4.17] tests/kernel-replace: use `dnf repoquery` to get kernel version

jlebon:
When testing from the c9s repos, don't hardcode a kernel version since it'll eventually cycle out of the mirrors. Instead, just query the repo to get the latest kernel.

Since we need the info we queried across reboots and we don't want to query it each time (and possibly get different answers), persist it in `/var` instead.

Ideally we'd do this for FCOS too, but we need to wait until the f41 rebase, which will bring dnf. (Or... could do it in a toolbox, but that seems overkill.) But anyway for now, the kernel from the stable Fedora repo shouldn't ever be GC'ed.

(cherry picked from commit 64c3a1cc3b38f6776d4ff45874572a284cc9f8f8)

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -48,20 +48,37 @@ image=oci:${image_dir}
 image_pull=ostree-unverified-image:$image
 tmp_imagedir=${image_dir}-tmp
 arch=$(arch)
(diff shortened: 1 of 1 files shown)


aaradhak (approved):
LGTM

marmijo (approved):

PR #3108, opened 2024-08-21
[rhcos-4.17] tests/files/validate-symlinks: add bootc/storage symlink to allowlist

jlebon:
This symlink isn't broken on a system deployed via bootc, but it
it is broken otherwise. Add the symlink to the allowlist for the test
so it will pass.

fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1782
(cherry picked from commit 88a473fc7230b064cc38e6632aa7155a00e6961e)


jlebon:
See https://github.com/openshift/os/pull/1581#issuecomment-2302136938, specifically:

> We can get https://github.com/coreos/fedora-coreos-config/pull/3108 in once/if that bootc release lands in 9.4.

PR #3109, opened 2024-08-21
[branched] Ignore json-glib broken symlink and unpin

travier:
See: https://bugzilla.redhat.com/show_bug.cgi?id=2297094

---

Will have to open the same for Rawhide as well.

PR #3116, opened 2024-08-21
[branched] overrides: fast-track systemd-256.5-1.fc41

jlebon:
systemd v256 added a new userdb functionality where SSH authorized
keys can be part of a User Record. To make this transparently
work with sshd authentication, an sshd config dropin that sets an
`AuthorizedKeysCommand` directive was added.

Unfortunately, it was added with a higher priority than intended,
which meant that it overrode the `AuthorizedKeysCommand` directive from
`ssh-key-dir`, which is how our `~/.ssh/authorized_keys.d/` magic works
today with Ignition and Afterburn. So the end result is that this broke
SSH which of course broke kola too.

This is tracked in upstream systemd at:

https://github.com/systemd/systemd/issues/33648

The dropin was recently reverted in Fedora:

https://src.fedoraproject.org/rpms/systemd/c/38291e13c1dec15618b7d09e4217d10076897cdf?branch=f41

Fast-track the latest f41 systemd build with that change.

We'll need to keep an eye on the conversation there to make sure that
the final solution doesn't re-break FCOS, but we would notice it pretty
quickly too.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1775

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -14,78 +14,39 @@ packages:
     metadata:
       reason: https://bugzilla.redhat.com/show_bug.cgi?id=2297094
       type: pin
-  device-mapper:
-    evr: 1.02.197-1.fc40
-    metadata:
-      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1735
-      type: pin
-  device-mapper-event:
-    evr: 1.02.197-1.fc40
-    metadata:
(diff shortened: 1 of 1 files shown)


marmijo (dismissed):
LGTM

travier (approved):

travier:
Let's merge this one first and then I'll rebase https://github.com/coreos/fedora-coreos-config/pull/3109

PR #3123, opened 2024-08-26
tests/kola: use latest fedora:latest on tests that require it

marmijo:
Both `ntp.chrony.dhcp-propagation` and `podman.rootless-systemd` use the Fedora container to set up the environment for the test. Always use the `fedora:latest` container so we never use EOL Fedora releases in kola tests.

### DIFF
--- a/tests/kola/ntp/data/ntplib.sh
+++ b/tests/kola/ntp/data/ntplib.sh
@@ -20,7 +20,7 @@ ntp_test_setup() {
     # run podman commands to set up dnsmasq server
     pushd "$(mktemp -d)"
     cat <<EOF >Dockerfile
-FROM quay.io/fedora/fedora:40
+FROM quay.io/fedora/fedora:latest
 RUN rm -f /etc/yum.repos.d/*.repo \
 && curl -L https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/fedora.repo -o /etc/yum.repos.d/fedora.repo
 RUN dnf -y install systemd dnsmasq iproute iputils \
--- a/tests/kola/podman/rootless-systemd
+++ b/tests/kola/podman/rootless-systemd
@@ -35,7 +35,7 @@ set -euxo pipefail
 #       https://github.com/coreos/coreos-assembler/issues/1645
 cd $(mktemp -d)
 cat <<EOF > Containerfile
-FROM quay.io/fedora/fedora:40
+FROM quay.io/fedora/fedora:latest
 RUN rm -f /etc/yum.repos.d/*.repo \
 && curl -L https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/fedora.repo -o /etc/yum.repos.d/fedora.repo
 RUN dnf -y update \


aaradhak (approved):

dustymabe:
hmm. I feel like we pin on purpose (since moving tags move at times which could be inopportune for us). Don't we usually bump these when we switch to a new Fedora major?

I guess another way to phrase this line of questioning.. what problem are we solving?

jlebon:
@dustymabe See the commit message in https://github.com/openshift/os/pull/1583 for details.

But yeah, also not sure about this even though I agree it's a logical follow-up to https://github.com/openshift/os/pull/1583. One difference from that PR though that motivated me there is that in the extensions case, we're doing something trivial with the Fedora image: installing `createrepo_c` and generating repodata. It felt to me like something unlikely to break.

The tests here OTOH are doing more complex things like dnsmasq, httpd, and running systemd as init in the container... which I can definitely imagine breaking. And we don't want e.g. a broken dnsmasq or systemd update in the latest Fedora to suddenly break all our streams.

I guess easiest short-term is to add a new `fedora-archive.repo` file that points to the archived location to work around the current issues.

I still think the cleanest approach for these tests though is to eventually use content from the same compose we're building from (https://github.com/openshift/os/issues/1546).

dustymabe:
> @dustymabe See the commit message in [openshift/os#1583](https://github.com/openshift/os/pull/1583) for details.
> 
> But yeah, also not sure about this even though I agree it's a logical follow-up to [openshift/os#1583](https://github.com/openshift/os/pull/1583). One difference from that PR though that motivated me there is that in the extensions case, we're doing something trivial with the Fedora image: installing `createrepo_c` and generating repodata. It felt to me like something unlikely to break.

Even there maybe we should be using CentOS or RHEL instead. I could foresee a case where create_repo adopted a new feature or default that older clients couldn't handle.

> 
> The tests here OTOH are doing more complex things like dnsmasq, httpd, and running systemd as init in the container... which I can definitely imagine breaking. And we don't want e.g. a broken dnsmasq or systemd update in the latest Fedora to suddenly break all our streams.

correct.. or even something as simple as glibc updates in newer Fedora containers not being compatible with older branched OS. I vaguely remember cases where older OS couldn't run newer containers.

> 
> I guess easiest short-term is to add a new `fedora-archive.repo` file that points to the archived location to work around the current issues.



> 
> I still think the cleanest approach for these tests though is to eventually use content from the same compose we're building from ([openshift/os#1546](https://github.com/openshift/os/issues/1546)).


👍

jlebon:
> > But yeah, also not sure about this even though I agree it's a logical follow-up to [openshift/os#1583](https://github.com/openshift/os/pull/1583). One difference from that PR though that motivated me there is that in the extensions case, we're doing something trivial with the Fedora image: installing `createrepo_c` and generating repodata. It felt to me like something unlikely to break.
> 
> Even there maybe we should be using CentOS or RHEL instead.

That actually used to be the case, but we moved to Fedora because of https://github.com/openshift/os/issues/1000. We might be able to revisit this now. (Obviously something we'd need to figure out too for https://github.com/openshift/os/issues/1546.)

travier:
I've just realized that we actually have such a potentially breaking change coming in Fedora 41: https://discussion.fedoraproject.org/t/f41-change-proposal-upgrade-systems-to-createrepo-c-1-0-and-change-repositories-metadata-settings/108893

So https://github.com/openshift/os/pull/1583 might be a risk as well.

marmijo:
> I guess easiest short-term is to add a new `fedora-archive.repo` file that points to the archived location to work around the current issues.

I opened https://github.com/coreos/fedora-coreos-config/pull/3128 as a first attempt at that.

marmijo:
Closing because it was decided that https://github.com/coreos/fedora-coreos-config/pull/3128 would be the solution here

PR #3125, opened 2024-08-27
[rawhide] disable systemd-ssh-generator for s390x

jbtrystram:
selinux-policy denies access to /proc/sysinfo which is required by the ssh-generator for s390x architecture.
Desabling this generator until selinux-policy get an update as all the tests fails due to this: kola check for badness in the console output.

See https://github.com/coreos/fedora-coreos-tracker/issues/1786

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -9,3 +9,8 @@ include: image-base.yaml
 # and enable it by default for rawhide!
 deploy-via-container: true
 container-imgref: "ostree-remote-registry:fedora:quay.io/fedora/fedora-coreos:rawhide"
+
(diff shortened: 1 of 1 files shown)


travier:
> Desabling

Disabling

travier:
I may be mistaken but I don't think we have arch conditionals for this file. It's not an rpm-ostree manifest.

dustymabe:
> I may be mistaken but I don't think we have arch conditionals for this file. It's not an rpm-ostree manifest.

I think you are right.

dustymabe (approved):
LGTM

PR #3126, opened 2024-08-27
[branched] Disable systemd-ssh-generator for s390x

jbtrystram:
selinux-policy denies access to /proc/sysinfo which is required by the ssh-generator for s390x architecture.
Desabling this generator until selinux-policy get an update as all the tests fails due to this: kola check for badness in the console output.

See https://github.com/coreos/fedora-coreos-tracker/issues/1786


dustymabe (approved):
LGTM

PR #3130, opened 2024-08-29
tests: verify `bootupctl status` output is not empty

HuijingHei:
See https://github.com/coreos/bootupd/issues/694

### DIFF
--- a/tests/kola/boot/bootupd
+++ b/tests/kola/boot/bootupd
@@ -22,11 +22,21 @@ case "$(arch)" in
     aarch64|x86_64)
         # on these arches, we always expect state files to exist
         check_state_file=1
+        # aarch64 (and x86_64) uses uefi firmware, should check grub2-efi
+        evr=$(rpm -q grub2-common --qf '%{EVR}')
+        if ! bootupctl status | grep "grub2-efi-.*-${evr}"; then
+            fatal "bootupctl status output should include grub2-efi package version"
+        fi
         ;;
     ppc64le)
         # ppc64le has it if built by osbuild, otherwise not
         if [ -e /sysroot/.aleph-version.json ]; then
             check_state_file=1
+        fi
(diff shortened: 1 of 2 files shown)


tests/kola/boot/bootupd
```diff
@@ -22,12 +22,21 @@ case "$(arch)" in
     aarch64|x86_64)
         # on these arches, we always expect state files to exist
         check_state_file=1
+        # aarch64 (and x86_64) uses uefi firmware, should check grub2-efi
+        if ! bootupctl status | grep "grub2-efi"; then
```
jlebon: Ahh I see, you'd have to make this additionally conditional to check for either grub2-efi-x64 or -aa64 to be able to do a NEVRA check. Hmm, how about we get the EVR for `grub2-common`, which is on both, and then do `grep "grub2-efi-.*-${evr}"`?

tests/kola/upgrade/extended/test.sh
```diff
@@ -145,7 +145,7 @@ ok "Reached version: $version"
 if vereq $version $target_version; then
     ok "Fully upgraded to $target_version"
     # log bootupctl information for inspection where available
-    [ -f /usr/bin/bootupctl ] && /usr/bin/bootupctl status
+    [ -f /usr/bin/bootupctl ] && (/usr/bin/bootupctl status | grep "CoreOS aleph version")
```
jlebon: I think the intent here was to have that output go to the test logs if we ever wanted to inspect it. Maybe to still do that, let's instead do `| tee out.txt` and then as a separate command below do `grep` on that file?

jlebon (commented):
Nice! Thank you so much for tackling this.

tests/kola/boot/bootupd
HuijingHei: SGTM, thanks for the suggestion.

travier (approved):

PR #3135, opened 2024-09-04
tests/upgrade: fix permission error; add logging

dustymabe:
With e726f1f we get a permission denied when this test runs in the
kola-upgrade job in the pipeline:

```
kola-runext-test.sh[1840]: + /usr/bin/bootupctl status
kola-runext-test.sh[1841]: + tee bootupctl.txt
kola-runext-test.sh[1841]: tee: bootupctl.txt: Operation not permitted
```

Let's rework it to not need to `tee` to write to a file and also
remove the "where available" part since we should have bootupd
in all of our newly built CoreOS versions.

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -217,6 +217,10 @@ fi
 set -o pipefail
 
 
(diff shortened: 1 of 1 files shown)


HuijingHei (approved):
LGTM

PR #3139, opened 2024-09-04
tests/reboot: reboot 10 times

dustymabe:
This just confirms there aren't somehow issues with rebooting multiple
times.

### DIFF
--- a/tests/kola/reboot/test.sh
+++ b/tests/kola/reboot/test.sh
@@ -9,46 +9,58 @@ set -xeuo pipefail
 # shellcheck disable=SC1091
 . "$KOLA_EXT_DATA/commonlib.sh"
 
-# /var
-varsrc=$(findmnt -nvr /var -o SOURCE)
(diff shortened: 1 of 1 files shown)


dustymabe:
I'm not convinced we should do this but I had made changes to the code last week and am interested in what others think.

jlebon:
Is this motivated by some recent regression?

dustymabe:
> Is this motivated by some recent regression?

when investigating https://github.com/coreos/fedora-coreos-tracker/issues/1756 i was initially suspicious (since the failure always seemed to occur with secureboot enabled) that there was some fragility in rebooting uefi-secure machines many times so I modified this test to help me prove or disprove that theory.

jbtrystram:
It feels a bit weird to test more than once IMO. Do we trust our stuff to be indempotent ?

dustymabe:
I'll close this out. I mostly posted it up because I had written the code and wanted to see if anyone else thought it was a good idea. The code will be here in case we ever want it in the future.

PR #3140, opened 2024-09-04
[rhcos-4.16] tests/kola: use FCOS defined fedora-archive.repo to set up containers

marmijo:
This is a backport of: https://github.com/coreos/fedora-coreos-config/pull/3061
- 48530410c95836ae6a5b1c352d029e6bbbb26d2d (tests/ntp: use the FCOS defined fedora.repo to set up container)
- b9f3487e71c187742900d33622c996c079cea3df (tests/podman/rootless-systemd: use the FCOS defined fedora.repo to set up container)

There are two tests that use the fedora container to set up and run the test environment. Use the fedora.repo file defined in fedora-coreos-config to set up that container. This will force packages to be downloaded from dl.fedoraproject.org, as specified in the FCOS file. The ITUP cluster, being used by the RHCOS pipeline, requires all outbound connections to be specified in a Firewall Egress file, and this will ensure the same connection will always be used.

---
This is also a backport of https://github.com/coreos/fedora-coreos-config/pull/3146

The fedora-archive.repo file now contains both EOL and non-EOL repo locations[1]. This means we can change the two kola tests that use the fedora container to use fedora-archive.repo as the only repo configuration file. This reduces the maintenance burden because now we don't have to change this curl statement when fedora versions reach EOL.

### DIFF
--- a/tests/kola/ntp/data/ntplib.sh
+++ b/tests/kola/ntp/data/ntplib.sh
@@ -22,7 +22,7 @@ ntp_test_setup() {
     cat <<EOF >Dockerfile
 FROM quay.io/fedora/fedora:40
 RUN rm -f /etc/yum.repos.d/*.repo \
-&& curl -L https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/fedora.repo -o /etc/yum.repos.d/fedora.repo
+&& curl -L https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/fedora-archive.repo -o /etc/yum.repos.d/fedora-archive.repo
 RUN dnf -y install systemd dnsmasq iproute iputils \
 && dnf clean all \
 && systemctl enable dnsmasq
--- a/tests/kola/ntp/data/ntplib.sh
+++ b/tests/kola/ntp/data/ntplib.sh
@@ -21,6 +21,8 @@ ntp_test_setup() {
     pushd "$(mktemp -d)"
     cat <<EOF >Dockerfile
 FROM quay.io/fedora/fedora:40
+RUN rm -f /etc/yum.repos.d/*.repo \
+&& curl -L https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/fedora.repo -o /etc/yum.repos.d/fedora.repo
 RUN dnf -y install systemd dnsmasq iproute iputils \
 && dnf clean all \
 && systemctl enable dnsmasq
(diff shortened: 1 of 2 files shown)


dustymabe:
We should consider populating the archive repo here now as well.. I think it will just ignore it since no repo actually exists in those locations, but it future proofs us from having to add it later.

If it does work we should update the files in the `testing-devel` branch and add that change to the backport here for 4.15 and 4.16.

marmijo:
> We should consider populating the archive repo here now as well.. I think it will just ignore it since no repo actually exists in those locations, but it future proofs us from having to add it later.

I tested this in the ITUP cluster and `dnf` failed because the repo metadata does not exist in the archive location for F40. I was able to get it to work by setting `skip_if_unavailable=True` in `fedora-archive.repo`, but we'd have to do that for  `fedora.repo` as well to avoid that issue when F40 content is archived, and I'm sure that has broader implications.

dustymabe:
> but we'd have to do that for `fedora.repo` as well

I think I'd be fine with that since we already define our own `fedora.repo`. cc @jlebon 

i.e. I don't think there should be a case where neither of them are available.

jlebon:
Instead of `skip_if_unavailable`, we can also set multiple baseurls instead. See e.g. https://github.com/coreos/fedora-coreos-config/blob/2e10ea850b0bf634c18c02f1802373150e9b1753/fedora.repo#L6-L7.

The advantage of this is that it succeeds if at least one of the two is valid, otherwise it fails. `skip_if_unavailable` in general in CI systems is not great because it means that you can get non-deterministic results.

Edit: to clarify, what I'm suggesting here is to have the baseurls in the `fedora-archive.repo` be e.g.:

```
baseurl=https://dl.fedoraproject.org/pub/fedora/linux/releases/$releasever/Everything/$basearch/os/
        https://dl.fedoraproject.org/pub/fedora-secondary/releases/$releasever/Everything/$basearch/os/
        https://dl.fedoraproject.org/pub/archive/fedora/linux/releases/$releasever/Everything/$basearch/os/
        https://dl.fedoraproject.org/pub/archive/fedora-secondary/releases/$releasever/Everything/$basearch/os/

```

marmijo:
> Edit: to clarify, what I'm suggesting here is to have the baseurls in the fedora-archive.repo be e.g.:

So in that case, we would _only_ use `fedora-archive.repo` for the two kola tests and the extensions container (and any other future use of the fedora container)? Otherwise we'd still hit the issue where, once e.g. F40 content is archived, then `dnf` will fail on `fedora.repo`.

marmijo:
I opened https://github.com/coreos/fedora-coreos-config/pull/3145 to handle that.

marmijo:
/hold
I'll rework this PR to include https://github.com/coreos/fedora-coreos-config/pull/3146, which resulted from https://github.com/coreos/fedora-coreos-config/pull/3145

marmijo:
/unhold
this is ready for review

dustymabe (approved):
LGTM

PR #3145, opened 2024-09-05
fedora-archive.repo: use both EOL and non-EOL locations

marmijo:
Include both the EOL and non-EOL repo locations in the baseurl list to simplify maintenance. This update benefits two kola tests and the RHCOS extensions container by simplifying the repo configuration into a single file. As Fedora versions used by those containers reach EOL, we wont have to remove and replace the `fedora.repo` file. Instead, we can just use this file for container setup.


dustymabe (approved):
LGTM

PR #3146, opened 2024-09-05
tests/kola: use fedora-archive.repo when setting up fedora container

marmijo:
The `fedora-archive.repo` file now contains both EOL and non-EOL repo locations[1]. This means we can change the two kola tests that use the fedora container to use `fedora-archive.repo` as the only repo configuration file. This reduces the maintenance burden because now we don't have to change this curl statement when fedora versions reach EOL.

[1] https://github.com/coreos/fedora-coreos-config/pull/3145


dustymabe (approved):
LGTM

PR #3147, opened 2024-09-05
tests/setuid: add userhelper to RHCOS allow list

jlebon:
I'd like to add subscription-manager in RHCOS, which unfortunately currently pulls in the usermode stack for authentication. That includes the well-known userhelper setuid binary.

Add it to the RHCOS allow list.

See also: https://github.com/openshift/os/pull/1563


dustymabe (approved):
LGTM

PR #3138, opened 2024-09-04
[rhcos-4.12] tests/kola: use fedora-archive.repo for EOL Fedora containers

marmijo:
Use the fedora-archive.repo file defined in fedora-coreos-config to set up the EOL containers in tests that use it. This will force packages to be downloaded from `https://df.fedoraproject.org`, as specified in the repo file. The ITUP cluster, being used by the RHCOS pipeline, requires all outbound connections to be specified in a Firewall Egress file, and this will ensure the same connection will always be used when downloading the archived fedora content.

See: https://github.com/coreos/fedora-coreos-config/pull/3128
This is a backport of https://github.com/coreos/fedora-coreos-config/pull/3136
- **Note**: `ntp.chrony.dhcp-propagation` does not exist in `rhcos-4.12`

PR #3149, opened 2024-09-06
[rhcos-4.12] kola/dhcp-propagation: use fedora-archive.repo to set up container

marmijo:
This should have been part of https://github.com/coreos/fedora-coreos-config/pull/3138. I thought the dhcp-propagation test had been introduced in 4.13, but it had just been moved to another directory (https://github.com/coreos/fedora-coreos-config/commit/3d5ec3563db2a6d4a901ad90398a21244c778f3e).

---
Use the fedora-archive.repo file defined in fedora-coreos-config to set up the EOL containers in tests that use it. This will force packages to be downloaded from `https://df.fedoraproject.org`, as specified in the repo file. The ITUP cluster, being used by the RHCOS pipeline, requires all outbound connections to be specified in a Firewall Egress file, and this will ensure the same connection will always be used when downloading the archived fedora content.

see: https://github.com/coreos/fedora-coreos-config/pull/3128

### DIFF
--- a/tests/kola/chrony/dhcp-propagation
+++ b/tests/kola/chrony/dhcp-propagation
@@ -49,6 +49,8 @@ test_setup() {
     NTPHOSTIP=$(getent hosts time-c-g.nist.gov | cut -d ' ' -f 1)
     cat <<EOF >Dockerfile
 FROM registry.fedoraproject.org/fedora:36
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

PR #3150, opened 2024-09-09
overlay.d: add 07fix-selinux-labels overlay

jbtrystram:
/boot/efi and /sysroot dir and subfiles are unlabeled_t since
40.20240504.3.0.
This is likely due to a missing step in the OSBuild pipeline as this
started with coreos/fedora-coreos-tracker#1653.

This should be removed after the next barrier release, if the newly
produced images are fixed.

See coreos/fedora-coreos-tracker#1771
And coreos/fedora-coreos-tracker#1772


travier:
See https://docs.fedoraproject.org/en-US/fedora-silverblue/troubleshooting/#_running_restorecon for why you should never run `restorecon -R` on `/sysroot`.

It's also part of the reasons that we made `/sysroot` RO by default on Atomic Desktops. There should be more links in the change page.

dustymabe:
Probably going to need to make this more complex unfortunately: https://github.com/coreos/fedora-coreos-tracker/issues/1772#issuecomment-2338849028

travier:
Let's focus on `/boot` first as that's what blocking bootupd. We can fix `/sysroot` afterwards.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,41 @@
...
+
+echo "Relabeling ostree repo dirmeta and dirtreefiles"
+dirtree_files=$(mktemp)
+grep '/ostree/repo/objects|.dirtree|.dirmeta' "$unlabeled" > "$dirtree_files"
+while read -r f ; do
+  chcon -h -v -t root_t "$f"
+done < "$dirtree_files"
+
+echo "Check the repository for consistency"
+ostree fsck
```
travier: That's likely going to take a lot of time and disk I/O so I don't think we can afford do that.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,41 @@
...
+echo "Relabeling the /boot mount point"
+/usr/bin/chcon -t boot_t "$root_mount_point"/boot
+
+unlabeled=$(mktemp)
+echo "$unlabeled"
+# list oll the files with their selinux labels
+find "$root_mount_point" -context '*:unlabeled_t:*' > "$unlabeled"
+
+to_relabel=$(mktemp -d)
+# filter out ostree repo objects
```
travier: Instead of filtering, I would select only the files that we know we can safely relabel instead.

travier: Ideally we would call restorecon on those files instead of hardcoding a label in the script. But that might not work.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/bash
+set -euxo pipefail
+
+echo "Mounting boot partition"
+boot_mount_point=$(mktemp --directory)
+/usr/bin/mount /dev/disk/by-partlabel/boot "$boot_mount_point"
+echo "Fixing label for efi mount point"
+/usr/bin/chcon -t boot_t "$boot_mount_point"/efi
```
dustymabe: we might want to make this conditional since some architectures don't use efi. i.e. this file might not exist?

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/bash
+set -euxo pipefail
+
+echo "Mounting boot partition"
+boot_mount_point=$(mktemp --directory)
+/usr/bin/mount /dev/disk/by-partlabel/boot "$boot_mount_point"
+echo "Fixing label for efi mount point"
```
dustymabe:
```suggestion
echo "Fixing label for efi mount point on the /boot filesystem"
```

overlay.d/15fcos/usr/libexec/coreos-fix-labels
dustymabe: let's also make these relabelings conditional on if the labels are wrong. i.e. when someone looks at the logs for this migration script I'd like to know if the changes were necessary or not.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,41 @@
...
+/usr/bin/chcon -t boot_t "$boot_mount_point"/efi
+ 
+
+root_mount_point=$(mktemp --directory)
+echo "Mounting sysroot partition"
+/usr/bin/mount /dev/disk/by-partlabel/root "$root_mount_point"
+
+echo "Relabeling the /boot mount point"
```
dustymabe: Let's make this consistent with the other message 

```suggestion
echo "Fixing label for the /boot mount point on the root filesystem"
```

overlay.d/15fcos/usr/libexec/coreos-fix-labels
dustymabe: IIUC we can't just call restorecon without having everything mounted in the right place, which we could do. i.e. on a running system `sudo restorecon -vr /var/ /etc/ /usr/ /boot/` but that doesn't really handle some of the unlabeled files under `/sysroot/` or `/osree/` for which there isn't really any policy, which is kind of why we are in the mess we are in now.

I was thinking we could just find all files under `/` with `unlabeled_t` and make them `root_t` but I think @jlebon thought that wouldn't be safe.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
dustymabe: was recommended by @jlebon over in https://github.com/coreos/fedora-coreos-tracker/issues/1772#issuecomment-2338849028

overlay.d/15fcos/usr/libexec/coreos-fix-labels
jbtrystram: > Ideally we would call restorecon on those files instead of hardcoding a label in the script. But that might not work.

there is no policy for files under `/sysroot` so that would not work

overlay.d/15fcos/usr/libexec/coreos-fix-labels
jbtrystram: It's fairly fast on a fresh VM (less than 10s) but i don't know how it scale on a machine that went through updates. Note that this will only run on machine installed between now and February 1st of this year

overlay.d/15fcos/usr/libexec/coreos-fix-labels
jbtrystram: > we might want to make this conditional since some architectures don't use efi. i.e. this file might not exist?

I pushed a change that just relabelle the whole `boot` FS as boot_t as I found a couple of other unlabeled files in there as well

jbtrystram: > let's also make these relabelings conditional on if the labels are wrong. i.e. when someone looks at the logs for this migration script I'd like to know if the changes were necessary or not.

I also added that at the beginning

jbtrystram:
[Here](https://jbtrystram.fedorapeople.org/fixed_cleaned.txt) is what the root filesystem looks like after this script ran. 
For comparison,[ this is extracted](https://jbtrystram.fedorapeople.org/good_cleaned.txt) from a good filesystem
the only file i am curious about is `/ostree/deploy/fedora-coreos/deploy/$COMMIT.0.origin`

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,41 @@
...
+
+unlabeled=$(mktemp)
+echo "$unlabeled"
+# list oll the files with their selinux labels
+find "$root_mount_point" -context '*:unlabeled_t:*' > "$unlabeled"
+
+to_relabel=$(mktemp -d)
+# filter out ostree repo objects
+grep -v '/ostree/deploy/fedora-coreos/deploy/' "$unlabeled" > "$to_relabel"/unlabeled.1
+grep -v '/ostree/deploy/fedora-coreos/var/' "$to_relabel"/unlabeled.1 > "$to_relabel"/unlabeled.2
```
HuijingHei: Need to update to `/ostree/deploy/.*/deploy` (and also `/ostree/deploy/.*/var`) if we want to also apply it to `rhcos`, as the name is specific to fcos, see [comment](https://github.com/coreos/fedora-coreos-tracker/issues/1772#issuecomment-2338849028)

overlay.d/15fcos/usr/lib/systemd/system/coreos-fix-selinux-labels.service
```diff
@@ -0,0 +1,20 @@
+# Remove after the next barrier release
+
+[Unit]
+Description=Fix incorrect selinux labels under /boot and /sysroot
```
jlebon:
```suggestion
Description=CoreOS Fix SELinux Labels
```

overlay.d/15fcos/usr/libexec/coreos-fix-labels
jbtrystram: good catch, thanks

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,48 @@
...
+current_label=$(getfattr -n security.selinux --absolute-name --only-values /sysroot/ostree)
+
+if [ "$current_label" != "system_u:object_r:unlabeled_t:s0" ]; then
+  echo "This CoreOS installation is properly labeled. Exiting"
+  exit 0
+fi
+
+echo "Mounting boot partition"
+boot_mount_point=$(mktemp --directory)
+/usr/bin/mount /dev/disk/by-partlabel/boot "$boot_mount_point"
```
jlebon: Instead of this, we can just remount the boot partition in-place rw (`mount -o rw,remount /boot`).

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,48 @@
...
+echo "Fixing label for files on the /boot filesystem"
+/usr/bin/chcon -h -v -R -t boot_t "$boot_mount_point"
```
jlebon: This seems more aggressive than it needs to be. Isn't it just `/boot/efi`, the `.root_uuid` and the `bootuuid.cfg`?

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,48 @@
...
+
+root_mount_point=$(mktemp --directory)
```
jlebon: Similarly here. That way we don't leak directories.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,48 @@
...
+echo "Mounting sysroot partition"
+/usr/bin/mount /dev/disk/by-partlabel/root "$root_mount_point"
+
+echo "Fixing the label for the /boot mount point on the root filesystem"
+/usr/bin/chcon -t boot_t "$root_mount_point"/boot
+
+unlabeled=$(mktemp)
+echo "$unlabeled"
+# list oll the files with their selinux labels
+find "$root_mount_point" -context '*:unlabeled_t:*' > "$unlabeled"
```
jlebon: Hmm... this will crawl the entire sysroot. Quite expensive.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
```diff
@@ -0,0 +1,48 @@
...
+
+to_relabel=$(mktemp -d)
+# filter out ostree repo objects
+grep -v '/ostree/deploy/fedora-coreos/deploy/' "$unlabeled" > "$to_relabel"/unlabeled.1
+grep -v '/ostree/deploy/fedora-coreos/var/' "$to_relabel"/unlabeled.1 > "$to_relabel"/unlabeled.2 
+grep -v '/ostree/repo/objects' "$to_relabel"/unlabeled.2 > "$to_relabel"/unlabeled
+
+while read -r f; do
+  chcon -h -v -t root_t "$f"
```
jlebon: Shouldn't hardcode `root_t` here. Or actually, should use `restorecon` instead of `chcon`.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
jlebon: Older versions are pruned, so the repo shouldn't keep growing as a node updates.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
jlebon: This only works though if `/boot/efi` isn't mounted (which is the default, but someone might have changed it). So using a temporary directory seems fine.

overlay.d/15fcos/usr/libexec/coreos-fix-labels
travier: So what do we do if it fails?

jlebon: Currently, the service will fail, which will make it show up in the motd via clhm at least. I guess we could say something like "The system is corrupted. Please save your data and re-install.", but I don't really expect corruption to happen here if we're careful enough. It's more a reflex anytime you fiddle with the objects dir directly.

dustymabe: > "The system is corrupted. Please save your data and re-install."

This sounds pretty harsh.. Maybe we could direct them to re-install if easy or open an issue if not?

jlebon: Or we just tell them to open an issue. E.g. "The OSTree repo is corrupted. This shouldn't have happened. Please file an issue at ..." I think regardless of whether they need to reinstall or not, we should know if it happens in case e.g. we want to stop the rollout.

I guess there's one case here, which is if their repo was _already_ corrupted before this script even ran. We could run fsck before doing the migration, but twice is a bit much. But anyway, logs should show if the corrupted objects are ones that we relabeled.

dustymabe:
I've updated this PR to fix existing FCOS and RHCOS systems and to include the tests from https://github.com/coreos/fedora-coreos-config/pull/3172 (which I'll now close in favor of this PR for those tests).

overlay.d/15fcos/usr/libexec/coreos-fix-labels
dustymabe: updated to have `maxdepth` as discussed in OOB session

overlay.d/15fcos/usr/libexec/coreos-fix-labels
dustymabe: fixed in latest upload

dustymabe:
I think I've tested this enough now that I'm comfortable with it merging if it passes review.

HuijingHei (dismissed):
Nice and neat, overall LGTM.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,149 @@
...
+context=$(matchpathcon -n "/")
+print_header "Changing context of unlabeled files to ${context}"
+(
+    find "/sysroot/ostree/repo/refs"           \
+         "/sysroot/.aleph-version.json"        \
+         "/sysroot/.coreos-aleph-version.json" \
+         /sysroot/ostree/repo/{.lock,config}   \
+         /sysroot/ostree/boot*                 \
+                     -context '*:unlabeled_t:*';
+    find "/sysroot/" -maxdepth 7 -context '*:unlabeled_t:*'                             \
```
jbtrystram: This is going to go in the users' data in  `/sysroot/ostree/deploy/*/var` directory. should we be concerned about relabelling files theres ?

jbtrystram (commented):
Thanks @dustymabe for reworking that. Well documentend and easy to review ! 
I have just one concern about relabelling stuff in `var`

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,149 @@
...
+        \(                                                                              \
+           -type l -or -type d -or                                                      \
+           \( -type f -regex '.*\.\(commitmeta\|commit\|dirmeta\|dirtree\|origin\)' \)  \
```
jbtrystram:
```suggestion
           \( -type f -regex '.*\.\(commitmeta\|commit\|dirmeta\|dirtree\|origin\)\$' \)  \
```
I would add a $ at the end here so we are sure to match files extensions ( just to be extra safe)

dustymabe: IIUC the current regex will only match suffixes because it doesn't have a `.*` at the end of the pattern.

dustymabe: ok. I tested with and without it and the results are the same so I'll just add it.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: added a `-prune` to forcefully skip the var dir

jbtrystram:
I can't approved my own PR, but LGTM !

HuijingHei (dismissed):

dustymabe:
I'll leave this open for a few more hours for further reviews.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,156 @@
...
+}
+
+get_context() {
+    path=$1
+    getfattr -n security.selinux --absolute-name --only-values "${path}" | \
+        tr -d '\0' # Trim the null byte from the ouptut to prevent bash warning
+}
+
+path_unlabeled() {
+    test -e "$1" || return 1 # no exist so not unlabeled
```
jlebon, line 30: Hmm, isn't this a success case? I.e. shouldn't we `return 0` here?

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,156 @@
...
+    if [ "$(get_context "$1")" == "system_u:object_r:unlabeled_t:s0" ]; then
+        return 0
+    else
+        return 1
+    fi
```
jlebon, line 35: Minor/optional: that's

```suggestion
    [ "$(get_context "$1")" == "system_u:object_r:unlabeled_t:s0" ]
```

:)

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,156 @@
...
+if ! path_unlabeled "/sysroot"; then
+    # We don't need to go further with the other fixes since
+    # this system doesn't appear to be affected by #1771,#1772.
+    echo "coreos-fix-selinux-labels finished successfully" > /var/lib/coreos-fix-selinux-labels.stamp
+    exit 0
+fi
+
+print_header "Mounting boot partition separately to check shadowed /boot/efi"
+boot_mount_point=$(mktemp --directory)
+mount -v /dev/disk/by-partlabel/boot "$boot_mount_point"
```
jlebon: `by-label/boot` is more canonical.

And... actually I think `by-partlabel/boot` will not work at all in RAID1.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,156 @@
...
+    /usr/bin/chcon -h -v "${context}" "${boot_mount_point}/efi"
+fi
+umount -v "$boot_mount_point"
+rmdir "$boot_mount_point"
+
+
+# The underlying /boot directory on the root filesystem can be wrong
+print_header "Mounting root partition separately to check shadowed /boot"
+root_mount_point=$(mktemp --directory)
+mount -v /dev/disk/by-partlabel/root "$root_mount_point"
```
jlebon: Similarly here.

jlebon: But actually, we don't really need this mount in the first place. The directory is available at `/sysroot/boot`. It's not shadowed.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,156 @@
...
+    find "/sysroot/" -maxdepth 7 -path /sysroot/ostree/deploy/*/var -prune -o                \
+        \(                                                                                   \
+            -context '*:unlabeled_t:*'                                                       \
+            \(                                                                               \
+                -type l -or -type d -or                                                      \
+                \( -type f -regex '.*\.\(commitmeta\|commit\|dirmeta\|dirtree\|origin\)$' \) \
+            \)                                                                               \
+            -print0                                                                          \
+        \)
+) | xargs --null -I{} chcon -v -h "${context}" {} | \
```
jlebon: You almost always want `-r` for `xargs`.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
```diff
@@ -0,0 +1,156 @@
...
+    wc -l | xargs -I{} echo "Relabeled {} files to ${context}"
```
jlebon: Should we instead print the output to the stamp file? And then `wc -l` that. That might be useful for debugging in the future if we relabeled (or didn't relabel) something we should/shouldn't have.

overlay.d/README.md
```diff
@@ -10,6 +10,17 @@ This overlay matches `fedora-coreos-base.yaml`; core Ignition+ostree bits.
...
+07fix-selinux-labels
+--------------------
+
+Fix incorrect SELinux labels in /boot and /sysroot
+- https://github.com/coreos/fedora-coreos-tracker/issues/1772
+- https://github.com/coreos/fedora-coreos-tracker/issues/1771
+We need this for both FCOS and RHCOS and it needs to live for
+some time (not just a single FCOS barrier release) so that we
+can ensure RHCOS 4.16 aleph nodes and some early 4.17 aleph
+nodes have been fixed.
```
jlebon: In FCOS it can be a single barrier release. In RHCOS, it will need to live for the duration of 4.18, which is the next EUS release. We can remove it in 4.19 once it opens.

tests/kola/selinux/file-context-policy-match
```diff
@@ -0,0 +1,52 @@
...
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+# Add the `-not -regex '.*.0/var/mnt'` here to have an exception for
+# the /sysroot/ostree/deploy/*/deploy/*.0/var/mnt directory existing
+# and being unlabeled. We don't know why this file is getting created
+# early in boot (sometime in initrd teardown or very early systemd
+# startup in the real root), but the behavior is fixed in systemd
+# v254+ so this should no longer be a problem on RHCOS once we move
+# to EL10 and we can remove that from the `find` command below.
+unlabeled="$(find /sysroot -context '*unlabeled_t*' -not -regex '.*.0/var/mnt' -print0 | xargs --null -I{} ls -ldZ '{}')"
```
jlebon, line 19:
```suggestion
unlabeled="$(find /sysroot -context '*unlabeled_t*' -not -regex '.*.0/var/mnt' -print0 | xargs -r --null -I{} ls -ldZ '{}')"
```

tests/kola/selinux/file-context-policy-match
```diff
@@ -0,0 +1,52 @@
...
+if [ -n "${unlabeled}" ]; then
```
jlebon, line 20: Let's print here what we found?

tests/kola/upgrade/extended/test.sh
```diff
@@ -154,6 +154,49 @@ move-to-cgroups-v2() {
...
 }
 
+selinux-sanity-check() {
+    # First make sure the migrations/fix script has finished (if it is going
+    # to run) before doing the checks
+    systemd-run --wait --property=After=coreos-fix-selinux-labels.service \
+        echo "Waited for coreos-fix-selinux-labels.service to finish"
+    # Verify SELinux labels are sane. Migration scripts should have cleaned
+    # up https://github.com/coreos/fedora-coreos-tracker/issues/1772
+    unlabeled="$(find /sysroot -context '*unlabeled_t*' -print0 | xargs --null -I{} ls -ldZ '{}')"
```
jlebon, line 164:
```suggestion
    unlabeled="$(find /sysroot -context '*unlabeled_t*' -print0 | xargs -r --null -I{} ls -ldZ '{}')"
```

:)

tests/kola/upgrade/extended/test.sh
```diff
@@ -154,6 +154,49 @@ move-to-cgroups-v2() {
...
+    fi
+    mislabeled="$(restorecon -vnr /var/ /etc/ /usr/ /boot/)"
+    if [ -n "${mislabeled}" ]; then
+        # Exceptions for files that could be wrong (sometimes upgrades are messy)
+        #   Would relabel /var/lib/cni from system_u:object_r:var_lib_t:s0 to system_u:object_r:container_var_lib_t:s0
+        #   Would relabel /etc/selinux/targeted/semanage.read.LOCK from system_u:object_r:semanage_trans_lock_t:s0 to system_u:object_r:selinux_config_t:s0
+        #   Would relabel /etc/selinux/targeted/semanage.trans.LOCK from system_u:object_r:semanage_trans_lock_t:s0 to system_u:object_r:selinux_config_t:s0
+        #   Would relabel /etc/systemd/journald.conf.d from system_u:object_r:etc_t:s0 to system_u:object_r:systemd_conf_t:s0
+        #   Would relabel /etc/systemd/journald.conf.d/forward-to-console.conf from system_u:object_r:etc_t:s0 to system_u:object_r:systemd_conf_t:s0
+        #   Would relabel /boot/lost+found from system_u:object_r:unlabeled_t:s0 to system_u:object_r:lost_found_t:s0' ']'
```
jlebon, line 176: Hmm, doesn't the migration script fix this one?

tests/kola/upgrade/extended/test.sh
```diff
@@ -154,6 +154,49 @@ move-to-cgroups-v2() {
...
+        fatal "Some unlabeled files were found"
+    fi
+    mislabeled="$(restorecon -vnr /var/ /etc/ /usr/ /boot/)"
+    if [ -n "${mislabeled}" ]; then
+        # Exceptions for files that could be wrong (sometimes upgrades are messy)
+        #   Would relabel /var/lib/cni from system_u:object_r:var_lib_t:s0 to system_u:object_r:container_var_lib_t:s0
+        #   Would relabel /etc/selinux/targeted/semanage.read.LOCK from system_u:object_r:semanage_trans_lock_t:s0 to system_u:object_r:selinux_config_t:s0
+        #   Would relabel /etc/selinux/targeted/semanage.trans.LOCK from system_u:object_r:semanage_trans_lock_t:s0 to system_u:object_r:selinux_config_t:s0
+        #   Would relabel /etc/systemd/journald.conf.d from system_u:object_r:etc_t:s0 to system_u:object_r:systemd_conf_t:s0
+        #   Would relabel /etc/systemd/journald.conf.d/forward-to-console.conf from system_u:object_r:etc_t:s0 to system_u:object_r:systemd_conf_t:s0
```
jlebon, line 175: I suspect what happened there is that the policy changed over time and the labels they should have changed but in a backwards compatible way.

jlebon (commented):
Thank you so much for working on this!

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: ehh. yeah I had it that way originally, but it reads better (to me at least) to have the scaffolding around it

overlay.d/README.md
dustymabe: do we guarantee that upgrading nodes have to go in between each minor? (i.e. would 4.16->4.19 ever be allowed)?

tests/kola/selinux/file-context-policy-match
dustymabe: the test is `set -x` so it all gets in the output anyway. maybe we shouldn't make it `set -x`?

tests/kola/upgrade/extended/test.sh
dustymabe: it does, but I think some older nodes (like upgrading from f34 where the migration script should NOOP) had this one unlabeled and I just decided to make it an exception.

tests/kola/upgrade/extended/test.sh
dustymabe: correct. that's what I suspect too

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: there may be some opportunity here for improving the name of this function

basically what we want to know is if any action should be taken. If the file doesn't exist then it's certainly not in the state of being unlabeled and needing to be fixed.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: the `-I{}` does the same thing:

```
[dustymabe@hattop ~]$ mkdir foo
[dustymabe@hattop ~]$ cd foo
[dustymabe@hattop foo]$ ls | xargs ls -l
total 0
[dustymabe@hattop foo]$ ls | xargs -r ls -l
[dustymabe@hattop foo]$ ls | xargs -I{} ls -l {}
[dustymabe@hattop foo]$ touch file
[dustymabe@hattop foo]$ ls | xargs -r ls -l
-rw-r--r--. 1 dustymabe dustymabe 0 Sep 27 13:58 file
[dustymabe@hattop foo]$ ls | xargs -I{} ls -l {}
-rw-r--r--. 1 dustymabe dustymabe 0 Sep 27 13:58 file
```

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: maybe.. It's not a small number of lines that would get printed to the file and I was wondering if we wanted to waste the space for that.

Let me know what you think.

tests/kola/selinux/file-context-policy-match
dustymabe: `-I{}` handles this.

tests/kola/upgrade/extended/test.sh
dustymabe: `-I{}` handles this

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: Will fix

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: I think this is leftover from an iteration of the script when I wasn't doing `mount -v -o remount,rw /sysroot` at the top so `sysroot/boot` would have been read-only.

I'll rework it.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
jlebon: Nice, TIL.

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
jlebon: I don't think it'd be nearly large enough to be a concern space-wise. I guess it's more whether we think it'd be useful in the future? Esp. since we don't log it to the journal. OK not to also.

overlay.d/README.md
jlebon: Control planes, yes. AIUI, worker nodes can skip, but the most they'd skip is to the next EUS as part of an EUS-to-EUS upgrade (so e.g. it couldn't go from 4.17 to 4.19, but it could go from 4.16 to 4.18).

tests/kola/selinux/file-context-policy-match
jlebon: Cool either way. I guess I'd go with consistency with the majority of other tests.

overlay.d/README.md
dustymabe: do you have a suggestion for an update to the comment that would make things clearer?

tests/kola/selinux/file-context-policy-match
dustymabe: I'll leave it like it is

overlay.d/07fix-selinux-labels/usr/libexec/coreos-fix-selinux-labels
dustymabe: ok let me think about how to get some useful info in there.

overlay.d/README.md
dustymabe: updated it to say to remove in the 4.19 cycle

dustymabe (approved):
All code review comments have been addressed.

PR #3154, opened 2024-09-10
Revert "overrides: fast-track selinux-policy-41.15-1.fc41"

jbtrystram:
This reverts commit 47a33bddc6aa91ee73b6fb48f44531dd48a57e4b.
 selinux-policy-41.15-1.fc41 was pushed to stable

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,16 +8,4 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages:
-  selinux-policy:
-    evra: 41.15-1.fc41.noarch
-    metadata:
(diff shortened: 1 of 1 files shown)


c4rt0 (approved):
/lgtm

dustymabe:
CI failed because it detected a downgrade. there is some lag (at least during `branched` status of a Fedora pre-release) where an update is marked as `stable` in bodhi but hasn't yet been pushed to the yum repos.

I'd say just re-run CI on this later today or tomorrow and it should pass.

jlebon:
Hmm, won't the graduation workflow take care of this for us when it's ready?

dustymabe:
> Hmm, won't the graduation workflow take care of this for us when it's ready?

it should, but I wasn't sure if that worked properly for `branched` or not. Probably

jlebon:
Yup, it does. If you look at a recent run, you can see the job matrix for every branch it runs for.

dustymabe:
ok then we'll just close this and let the bot do its thing when the package is in the repo.

PR #3155, opened 2024-09-10
coreos-boot-edit: relabel rdcore files

### LINKED ISSUE #1770: `coreos-boot-edit.sh` creates `/boot/.root_uuid` as `unlabeled_t`
### Describe the bug

https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh likely runs before we've loaded the SELinux policy and writes `/boot/.root_uuid` which ends up not being labeled (`unlabeled_t`).

### Reproduction steps

Boot FCOS. Run:

```
ls -alhZ /boot
```

### Expected behavior

All files are correctly labeled.

### Actual behavior

`/boot/.root_uuid` is `unlabeled_t`.

### System details

N/A

### Butane or Ignition config

```yaml
None
```

### Additional information

_No response_

jlebon:
The `rdcore bind-boot` command write files to the bootfs but currently doesn't relabel them. Let's just relabel it from this side for now. In the future we could look at having `rdcore` call `setfiles` like Ignition does, or better, make `coreos-relabel` a more public API.

This fixes https://github.com/coreos/fedora-coreos-tracker/issues/1770 for new installs.

Refrained from adding tests for this. I think instead what we need is once all these relabeling issues are fixed, a test that verifies that *everything* is labeled.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
@@ -16,8 +16,7 @@ karg() {
 
 # Mount /boot. Note that we mount /boot but we don't unmount it because we
 # are run in a systemd unit with MountFlags=slave so it is unmounted for us.
-bootmnt=/mnt/boot_partition
-mkdir -p ${bootmnt}
+bootmnt=/sysroot/boot
 bootdev=/dev/disk/by-label/boot
 mount -o rw ${bootdev} ${bootmnt}
 
@@ -44,3 +43,9 @@ fi
 # 4. it adds GRUB bootuuid.cfg dropins so that GRUB selects the boot filesystem
 #    by UUID
 rdcore bind-boot /sysroot ${bootmnt}
+
+# relabel files rdcore created; ideally in the future rdcore does this itself
+coreos-relabel /boot/.root_uuid
+if [ -e /sysroot/boot/grub2/bootuuid.cfg ]; then
+    coreos-relabel /boot/grub2/bootuuid.cfg
+fi


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
```diff
@@ -44,3 +43,9 @@ fi
 # 4. it adds GRUB bootuuid.cfg dropins so that GRUB selects the boot filesystem
 #    by UUID
 rdcore bind-boot /sysroot ${bootmnt}
```
dustymabe, line 45: is this running before the policy is loaded and that's why the created files don't have labels?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
```diff
@@ -44,3 +43,9 @@ fi
...
+
+# relabel files rdcore created; ideally in the future rdcore does this itself
+coreos-relabel /boot/.root_uuid
+if [ -e /sysroot/boot/grub2/bootuuid.cfg ]; then
+    coreos-relabel /boot/grub2/bootuuid.cfg
+fi
```
dustymabe, line 51: if the files that rdcore created are pretty static and we know the context the files should have I wonder if we should just hardcode it in `rdcore`. i.e. rather than have `rdcore` call `coreos-relabel` or some other command that consults the policy we just set the context directly when creating the files?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
jlebon: Correct.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
jlebon: Hmm, in general I don't think it's a good idea to hardcode specific labels.

dustymabe: I would typically agree, just wondering if in this specific instance (a few files in a known location) if it would be advantageous (i.e. to avoid the layering violations and also the complexity of an alternative more comprehensive solution in `rdcore`).

dustymabe: one other option here could be to have the package that delivers `rdcore` deliver a tmpfiles.d entry to request a relabel.

this assumes that there aren't selinux denials that happen between when this is run and tmpfiles that would invalidate this as an option (maybe it does).

dustymabe (approved):
LGTM - thanks for digging in to this

PR #3157, opened 2024-09-11
gcp-test: update check to run instance with confidential type `sev-snp`

HuijingHei:
Based on https://github.com/coreos/coreos-assembler/pull/3871

### DIFF
--- a/tests/kola/platforms/gcp/confidential-vm-nvme-symlink
+++ b/tests/kola/platforms/gcp/confidential-vm-nvme-symlink
@@ -8,10 +8,11 @@
 
 # See https://issues.redhat.com/browse/OCPBUGS-7582
 # https://github.com/coreos/fedora-coreos-tracker/issues/1457
+# https://github.com/coreos/coreos-assembler/issues/3556
 #
 # Force this test to not run by default unless named specifically
 # or `--tag confidential` is passed to `kola run`, also requires
-# `--gcp-machinetype n2d-standard-2 --gcp-confidential-vm`
+# `--gcp-machinetype n2d-standard-2 --gcp-confidential-type sev_snp`
 #
 # It will create confidential instance on GCP with 1 nvme persistent disk
 # and 1 local ssd disk, then check the new udev rules make effect.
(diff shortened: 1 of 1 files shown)


yasminvalim (dismissed):
/lgtm

dustymabe:
it feels weird to me that this test is named `nvme-symlink` but runs on confidential VMs. I'd never guess it from the name. Maybe `confidential-vm-nvme-symlink` would work better?

dustymabe (dismissed):
LGTM

HuijingHei:
> it feels weird to me that this test is named `nvme-symlink` but runs on confidential VMs. I'd never guess it from the name. Maybe `confidential-vm-nvme-symlink` would work better?

Updated, thanks!

tests/kola/platforms/gcp/confidential-vm-nvme-symlink
```diff
@@ -8,10 +8,11 @@
 
 # See https://issues.redhat.com/browse/OCPBUGS-7582
 # https://github.com/coreos/fedora-coreos-tracker/issues/1457
+# https://github.com/coreos/coreos-assembler/issues/3556
 #
 # Force this test to not run by default unless named specifically
 # or `--tag confidential` is passed to `kola run`, also requires
-# `--gcp-machinetype n2d-standard-2 --gcp-confidential-vm`
+# `--gcp-machinetype n2d-standard-2 --gcp-confidential-type SEV_SNP`
```
travier:
```suggestion
# `--gcp-machinetype n2d-standard-2 --gcp-confidential-type sev_snp`
```
Nit but that's what's documented in the kola help. Only change this if another round is needed

travier (dismissed):

tests/kola/platforms/gcp/confidential-vm-nvme-symlink
HuijingHei: Good catch, updated, thanks!

dustymabe (approved):
LGTM

dustymabe:
when this merges we'll need to cherry-pick it to `testing` so that our next `stable` build will have successful gcp tests.

PR #3158, opened 2024-09-13
[Rawhide]  overrides: pin `podman`, `shawdow-utils` and `skopeo`

gursewak1997:
Regressions in these packages fails the toolbox creation. Pinning the packages till then.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1793


dustymabe (approved):
LGTM

PR #3161, opened 2024-09-13
denylist: update denial to include next/next-devel

dustymabe:
We're moving next over to F41 so we'll need to apply this denial there too.

https://github.com/coreos/fedora-coreos-tracker/issues/1791


jlebon (approved):

dustymabe:
merging early since this won't be tested by CI here.

PR #3162, opened 2024-09-14
rawhide: overrides: pin toolbox in rawhide

gursewak1997:
Regressions in toolbox package fails the creation of toolbox Pinning the package till then.
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1793


dustymabe (approved):
LGTM

PR #3164, opened 2024-09-15
denylist: deny ext.config.extensions.package in f41 for now

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/1795


nirik:
I'm not sure I fully understand this, but temp dropping that test to workaround the infra repodata problem sound fine.

dustymabe:
skipping CI on this since it won't test this change anyway.

PR #3179, opened 2024-09-24
denylist: drop denial for `ignition.ssh.key`

gursewak1997:
After fast tracking to `afterburn-5.7.0-1`, we can remove this entry from the kola-denylist.


dustymabe (approved):

dustymabe:
When the next set of kola-azure tests run successfully we can close the issue.

PR #3181, opened 2024-09-26
test: extend bootupd test to include testing adoption and updates

HuijingHei:
See https://github.com/coreos/fedora-coreos-tracker/issues/1788#issuecomment-2326473398

---
denylist: deny `ext.config.boot.bootupd-validate` until we rollout F41

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -16,3 +16,10 @@
      - rawhide
      - next-devel
      - next
+- pattern: ext.config.boot.bootupd-validate
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1788
+  snooze: 2024-11-18
+  streams:
+    - stable
+    - testing
+    - testing-devel
--- a/tests/kola/boot/bootupd-validate
+++ b/tests/kola/boot/bootupd-validate
@@ -1,0 +1,87 @@
+#!/bin/bash
+## kola:
+##   # bootupd does not support bootloader update on s390x
+##   architectures: "! s390x"
+##   description: Extend bootupd test to include testing adoption and updates.
+
+# See https://github.com/coreos/fedora-coreos-tracker/issues/1788#issuecomment-2326473398
+# Steps:
+# 1) Only x64 and aarch64 have esp device
+#  - Overwrite an existing file in the ESP
+#  - Verify that `bootupctl validate` fails as expected
+# 2) Remove /boot/bootupd-state.json
+# 3) Run `bootupctl adopt_and_update`
+# 4) Verify that validate is successful and results are expected
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+overwrite=
+# only x64 and aarch64 have esp device
+overwrite_file() {
+    case "$(arch)" in
+        x86_64|aarch64)
+            local esp_dev=$(realpath /dev/disk/by-partlabel/EFI-SYSTEM)
+            if [ ! -b "${esp_dev}" ]; then
+                fatal "can not find ${esp_dev}"
+            fi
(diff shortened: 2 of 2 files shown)


HuijingHei:
The adoption for `BIOS` is implemented with `bootupd >= 0.2.22`, but we do not have the new build yet.

HuijingHei:
The CI will not run this test. Do testing manually based on `next-devel` image passed.
```
$ kola run  -E ./fedora-coreos-config --qemu-image data/fedora-coreos-41.20241017.10.0-qemu.x86_64.qcow2 ext.fedora-coreos-config.boot.bootupd-validate
=== RUN   ext.fedora-coreos-config.boot.bootupd-validate
--- PASS: ext.fedora-coreos-config.boot.bootupd-validate (58.27s)
PASS, output in _kola_temp/qemu-2024-10-18-0615-58
```

tests/kola/boot/bootupd-validate
```diff
@@ -0,0 +1,87 @@
...
+            ;;
+        *)
+            echo "skipped validate"
+            ;;
+    esac
+}
+
+overwrite_file
+if [ -n "${overwrite}" ] && bootupctl validate 2>&1; then
+    fatal "bootupctl validate failed as expected"
```
jbtrystram: I think the message should be `"bootupctl validate did not fail as expected" ?

jbtrystram (dismissed):
I am a bit unsure about the overwrite failure message, otherwise LGTM.

tests/kola/boot/bootupd-validate
HuijingHei: Agree with you, update the message, thanks!

kola-denylist.yaml
```diff
@@ -17,3 +17,11 @@
...
     - next-devel
     - next
+- pattern: ext.config.boot.bootupd-validate
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1788
+  snooze: 2024-11-18
+  streams:
+    - stable
+    - testing
+    - testing-devel
+    - next
```
travier: This should work on next now as it's running F41 and we've pushed the updated bootupd there.

travier:
Otherwise looks good but I only gave it a quick look. Will take another look next week.

kola-denylist.yaml
HuijingHei: Current next is `41.20241006.1.1` which has old `bootupd-0.2.20-2.fc41`, but next-devel `41.20241017.10.0` which has new bootupd-0.2.24-1.fc41, so we will not run `next` until move to `next-devel`, is this right?

travier: ~~Let's hold until we have the right version in next at least~~

travier: Ignore my previous comment.

HuijingHei: Sure, remove `next`, thanks!

HuijingHei:
The test is passed using next `fedora-coreos-41.20241020.1.0-qemu.x86_64.qcow2`.
```
$ kola run  -E ./fedora-coreos-config --qemu-image data/fedora-coreos-41.20241020.1.0-qemu.x86_64.qcow2 ext.fedora-coreos-config.boot.bootupd-validate
=== RUN   ext.fedora-coreos-config.boot.bootupd-validate
--- PASS: ext.fedora-coreos-config.boot.bootupd-validate (58.43s)
PASS, output in _kola_temp/qemu-2024-10-22-0150-41
```

jmarrero (approved):
lgtm

travier:
This fails in the RHCOS pipeline: https://github.com/openshift/os/pull/1644#issuecomment-2462331993

@HuijingHei can you take a look and either make something specific for RHCOS or keep the test only for FCOS for now? Thanks

PR #3182, opened 2024-09-26
[rawhide] overrides: pin `dracut-102-2.fc41` and `clevis-20-2.fc41`

marmijo:
see: https://github.com/coreos/fedora-coreos-tracker/issues/1803

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,39 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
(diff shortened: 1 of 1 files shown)


dustymabe:
this gets us past the multipath problem, but the next problem is `ext.config.security.coreos-update-ca-trust` is failing.

I'll go ahead and merge this but we won't get a successful rawhide until we look at that one too.

dustymabe (approved):

marmijo:
> this gets us past the multipath problem, but the next problem is ext.config.security.coreos-update-ca-trust is failing.

Thanks, I'm looking into it now.

marmijo:
https://github.com/coreos/fedora-coreos-config/pull/3183 should resolve that failure

PR #3183, opened 2024-09-26
tests/update-ca-trust: validate new `ca-bundle.crt` location

marmijo:
A recent update to the ca-certificates package in Fedora removed the `openssl_format_trust_bundle` ca-bundle to improve the startup speed of OpenSSL. Certificates passed into `/etc/pki/ca-trust/source/anchors` are picked up and added to `/etc/pki/tls/certs/ca-bundle.crt` after running `update-ca-trust extract`, so let's change the location here to the new cert bundle.
see: https://github.com/coreos/fedora-coreos-tracker/issues/1804

### DIFF
--- a/tests/kola/security/coreos-update-ca-trust/test.sh
+++ b/tests/kola/security/coreos-update-ca-trust/test.sh
@@ -12,7 +12,7 @@ set -xeuo pipefail
 if ! systemctl show coreos-update-ca-trust.service -p ActiveState | grep ActiveState=active; then
     fatal "coreos-update-ca-trust.service not active"
 fi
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

dustymabe:
I was going to ask if this would still work for !F42 too, but CI on this PR will tell us that :)

marmijo:
I tested it on testing-devel locally, so it should work. (from the issue I opened):

> I actually tested this in testing-devel and found the supplied cert in the new file, which confirms that this was the case before this change in rawhide. IIUIC, /etc/pki/tls/certs/ca-bundle.crt is a default location where OpenSSL looks for certificates.

PR #3185, opened 2024-09-28
[rhcos-4.16] overlay.d: add 07fix-selinux-labels overlay

dustymabe:
/boot/efi and /sysroot dir and subfiles are unlabeled_t since
40.20240504.3.0. This is likely due to some missing scaffolding
in the OSBuild software and definitions that we started using in
[1]. These issues [2] [3] were addressed in [4] for new image
builds, but we still need to fix upgrading systems, which we
do here in this migration script.

Note that we also fix a few files in /boot that were left
unlabeled by `rdcore` [5] while we are in here.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1653.
[2] https://github.com/coreos/fedora-coreos-tracker/issues/1771
[3] https://github.com/coreos/fedora-coreos-tracker/issues/1772
[4] https://github.com/coreos/coreos-assembler/pull/3885
[5] https://github.com/coreos/fedora-coreos-tracker/issues/1770

Co-authored-by: Dusty Mabe <dusty@dustymabe.com>                    
(cherry picked from commit 2e355fdd6494612489a488cd6299944e032204f4)

### DIFF
--- a/manifests/shared-workarounds.yaml
+++ b/manifests/shared-workarounds.yaml
@@ -1,2 +1,4 @@
 # This manifest is a list of shared workarounds that are needed in both Fedora CoreOS
 # and downstreams (i.e. Red Hat CoreOS).
+ostree-layers:
+  - overlay/07fix-selinux-labels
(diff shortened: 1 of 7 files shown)


HuijingHei (approved):
LGTM

PR #3187, opened 2024-10-01
35coreos-multipath: explicitly depend on multipath module

jlebon:
The multipath module has been included by default so we didn't need to do this, but a bug has recently made it stop being included:

https://github.com/coreos/fedora-coreos-tracker/issues/1803#issuecomment-2384671129

We should have been declaring that dependency here anyway.


jlebon:
This probably also fixes https://github.com/coreos/fedora-coreos-tracker/issues/1803.

dustymabe (approved):
LGTM

marmijo (approved):

PR #3190, opened 2024-10-03
[next-devel] overrides: fast-track grub packages: 2.12-10.fc41

ravanelli:
- VMWare OVA Fails To Boot due grub update See: https://github.com/coreos/fedora-coreos-tracker/issues/1802

### DIFF
--- a/manifest-lock.overrides.ppc64le.yaml
+++ b/manifest-lock.overrides.ppc64le.yaml
@@ -1,0 +1,23 @@
+# This lockfile should be used to pin to a package version (`type: pin`) or to
+# fast-track packages ahead of Bodhi (`type: fast-track`). Fast-tracked
+# packages will automatically be removed once they are in the stable repos.
+#
+# IMPORTANT: YAML comments *will not* be preserved. All `pin` overrides *must*
+# include a URL in the `metadata.reason` key. Overrides of type `fast-track`
+# *should* include a Bodhi update URL in the `metadata.bodhi` key and a URL
+# in the `metadata.reason` key, though it's acceptable to omit a `reason`
+# for FCOS-specific packages (ignition, afterburn, etc.).
+
+packages:
+  grub2-ppc64le:
+    evra: 1:2.12-10.fc41.ppc64le
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2024-7d58433dd5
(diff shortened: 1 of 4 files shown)


ravanelli:
@dustymabe sorry, it has been a while I do not pin a package in next-devel :grin:, I don't recall if I can open a PR directly to next-devel

dustymabe:
If we do merge this we'll need to account for some grub packages being different for different architectures.

I'm not sure we should merge it or not, though. Since it's only in `next` right now and we've already shipped it and the impact is small (i.e. new systems on VMWare) then maybe we should wait and see if the GRUB devs can figure out the problem soon rather than reverting (i.e. downgrading GRUB in the next `next` release) for now.

dustymabe:
I think there are a few ppc64le grub specific packages too aren't there?

ravanelli:
> I think there are a few ppc64le grub specific packages too aren't there?

Ahh, Indeed! fixed!

manifest-lock.overrides.ppc64le.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+packages:
+  grub2-ppc64le:
+    evra: 1:2.12-10.fc41.ppc64le
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2024-7d58433dd5
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1802
+      type: fast-track
+  grub2-ppc64le-modules:
+    evra: 1:2.12-10.fc41.ppc64le
```
dustymabe: I think this one is `noarch` for some reason:

```suggestion
    evra: 1:2.12-10.fc41.noarch
```

```
$ grep -A 1 grub2- manifest-lock.ppc64le.json 
    "grub2-common": {
      "evra": "1:2.12-9.fc41.noarch",
--
    "grub2-ppc64le": {
      "evra": "1:2.12-9.fc41.ppc64le",
--
    "grub2-ppc64le-modules": {
      "evra": "1:2.12-9.fc41.noarch",
--
    "grub2-tools": {
      "evra": "1:2.12-9.fc41.ppc64le",
--
    "grub2-tools-minimal": {
      "evra": "1:2.12-9.fc41.ppc64le",
```

dustymabe (approved):
LGTM

PR #3191, opened 2024-10-04
tests/kola: drop files.unlabeled-contexts test

dustymabe:
I didn't know about this test existing when I wrote the selinux.file-context-policy-match test in 01c70bd, which is more comprehensive. Let's drop this test in favor of that one.


jbtrystram (approved):
LGTM

PR #3192, opened 2024-10-04
tests/kola: add exceptions in upgrade test SELinux checks

dustymabe:
These are failing in the extended upgrade tests and we need to add them as exceptions for now while we work out the fixes.

- https://github.com/coreos/fedora-coreos-tracker/issues/1806
- https://github.com/coreos/fedora-coreos-tracker/issues/1808

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -174,6 +174,9 @@ selinux-sanity-check() {
         #   Would relabel /etc/systemd/journald.conf.d from system_u:object_r:etc_t:s0 to system_u:object_r:systemd_conf_t:s0
         #   Would relabel /etc/systemd/journald.conf.d/forward-to-console.conf from system_u:object_r:etc_t:s0 to system_u:object_r:systemd_conf_t:s0
         #   Would relabel /boot/lost+found from system_u:object_r:unlabeled_t:s0 to system_u:object_r:lost_found_t:s0' ']'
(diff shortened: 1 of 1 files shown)


tests/kola/upgrade/extended/test.sh
```diff
@@ -174,17 +174,26 @@ selinux-sanity-check() {
...
+        #       - 39.20230916.1.1->41.20240928.10.1
+        #       - https://github.com/fedora-selinux/selinux-policy/commit/3ba70ae27d067f7edc0a52ff722511c5ada724f2
         declare -A exceptions=(
            ['/var/lib/cni']=1
            ['/etc/selinux/targeted/semanage.read.LOCK']=1
            ['/etc/selinux/targeted/semanage.trans.LOCK']=1
            ['/etc/systemd/journald.conf.d']=1
            ['/etc/systemd/journald.conf.d/forward-to-console.conf']=1
            ['/boot/lost+found']=1
+           ['/var/lib/systemd/home']=1
```
aaradhak, line 187: Just for my understanding - I assume that by adding this it ensures that the script doesn't falsely flag an error during SELinux policy updates?

aaradhak (approved):
LGTM

tests/kola/upgrade/extended/test.sh
dustymabe: it just means that if that file shows up in the list of ones that would be relabeled (i.e. it doesn't match the current policy) then we allow it because of some known SELinux policy change that happened.

PR #3197, opened 2024-10-07
[next] overrides: pin podman-5.2.4-1.fc41

coreosbot-releng:


dustymabe:
I did the "pin" workflow here because there hasn't been a bodhi update created for F41 for https://koji.fedoraproject.org/koji/buildinfo?buildID=2563135 yet, but I'm sure it will happen soon. We'll make sure to `fast-track` when we do it on `next-devel`.

dustymabe (approved):
LGTM

marmijo (approved):

PR #3205, opened 2024-10-14
tests/kola: make podman pasta test clean up

dustymabe:
Since this is a non-exclusive test let's clean up a little bit and at least not leave the container hanging around. This caused a conflict with the selinux.file-context-policy-match test because there were some directories in container storage that restorecon was warning about.


jbtrystram (approved):
lgtm

PR #3206, opened 2024-10-15
Revert "denylist: add kdump.crash for rawhide"

### LINKED ISSUE #1791: kdump fails with composeFS enabled
kdump gets confused with the RO filesystem. 

https://bugzilla.redhat.com/show_bug.cgi?id=2284097

Fix : https://github.com/rhkdump/kdump-utils/pull/28

Denylisted in https://github.com/coreos/fedora-coreos-config/pull/3009

jbtrystram:
This reverts commit e8625af2e7ecd4423d8f138eff196a76a8e9a5d5.

kdump-utils updates fixed this issue.
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1791

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -10,3 +10,10 @@
   warn: true
   arches:
     - ppc64le
+- pattern: kdump.crash.ssh
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1813
+  streams:
+    - rawhide
+    - branched
+    - next-devel
+    - next
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -10,12 +10,3 @@
   warn: true
   arches:
     - ppc64le
-- pattern: ext.config.kdump.crash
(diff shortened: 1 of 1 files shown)


kola-denylist.yaml
```diff
@@ -10,3 +10,11 @@
   warn: true
   arches:
     - ppc64le
+- pattern: kdump.crash.ssh
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1813
```
dustymabe, line 14: if we know it will fail then I'd prefer to add a snooze here (OR remove the `warn: true`).

The way it is configured now the test will always run and for now will always report a warning to us. I'd rather not see the warning if we know it is going to fail.

jbtrystram: good point!

dustymabe (approved):

dustymabe:
merging early since CI won't test this change since this PR is against `testing-devel` but the denylist entry is for other streams.

PR #3207, opened 2024-10-16
[next] overrides: fast-track sssd-2.10.0-1.fc41

coreosbot-releng:


dustymabe:
I think we'd want to target `next-devel` with this and not `next`?

aaradhak:
Ok I had the doubt earlier if it should be next or next-devel. I ll open another PR with target as next-devel. Will close this PR for now.

next-devel PR —> https://github.com/coreos/fedora-coreos-config/pull/3208

PR #3216, opened 2024-10-22
overrides: fast-track kdump-utils-1.0.48-1.fc40

coreosbot-releng:

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -10,10 +10,3 @@
   warn: true
   arches:
     - ppc64le
-- pattern: kdump.crash.ssh
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1813
-  streams:
-    - rawhide
-    - branched
-    - next-devel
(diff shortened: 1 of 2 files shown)


dustymabe:
let's open a fast-track for `next-devel` too and just drop the denylist entry here completely?

jbtrystram:
> let's open a fast-track for `next-devel` too and just drop the denylist entry here completely?

https://github.com/coreos/fedora-coreos-config/pull/3219

dustymabe:
the denylist itself has a single source of truth (on the `testing-devel` branch) so you'll need to remove the entry here in this PR.

also if PRs are to the main development branch (which is `testing-devel`) we don't usually prefix the PR title with the branch name.

dustymabe:
> the denylist itself has a single source of truth (on the `testing-devel` branch) so you'll need to remove the entry here in this PR.

fixed in latest push

dustymabe (approved):

PR #3225, opened 2024-10-24
tests/kola: add SELinux systemd /var/cache exceptions in upgrade test

dustymabe:
https://github.com/coreos/fedora-coreos-tracker/issues/1819#issuecomment-2436280393


dustymabe:
@c4rt0 - using your new knowledge of how to run the upgrade test can you review and test this? Run the test once without this change to make sure you can reproduce the problem and then apply this change and see that it works.

travier (approved):

travier:
Not tested but LGTM

PR #3237, opened 2024-10-30
tests/kola: remove dtb exceptions in upgrade test SELinux checks

### LINKED ISSUE #1808: aarch64: upgrade causes SELinux mislabeled dtb files in /boot/ostree
On upgrades it appears files in dtb files in /boot on aarch64 systems have the wrong context. It looks like the context matches where the files are copied from rather than the paths they are being copied to.

When booted from the fresh `40.20240825.3.0` `stable` image (i.e. the release from a few weeks back):

```
Fedora CoreOS 40.20240825.3.0
[core@cosa-devsh ~]$
[core@cosa-devsh ~]$ sudo restorecon -vnr /boot/ostree/
[core@cosa-devsh ~]$
```

After upgrade to latest stable:

```
Fedora CoreOS 40.20240906.3.0
[core@cosa-devsh ~]$
State: idle
  DriverState: active; periodically polling for updates (last checked Fri 2024-10-04 16:07:17 UTC)
Deployments:
_ fedora:fedora/aarch64/coreos/stable
                  Version: 40.20240906.3.0 (2024-09-23T18:11:19Z)
                   Commit: aad86a6390dbcf8caab4390e14854b2d6ddfa2b6885d5d3dfc7511bd431ba0ae

  fedora:fedora/aarch64/coreos/stable
                  Version: 40.20240825.3.0 (2024-09-10T00:49:47Z)
                   Commit: a0e2520be57c71a0d459fee831b1e9eb0f9aea02b4d05d2fe7b8f66b8678471a



[core@cosa-devsh ~]$ sudo restorecon -vnr /boot/ostree/ | wc -l
1866


[core@cosa-devsh ~]$ sudo restorecon -vnr /boot/ostree/ | head
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/cn9132-db-B.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/armada-3720-uDPU.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/cn9130-crb-B.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/armada-7040-mochabin.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/armada-3720-gl-mv1000.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/armada-7040-db.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/cn9131-db-B.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0
Would relabel /boot/ostree/fedora-coreos-3bdd09e36aecabd4d88e93b696c885903cde80cbdadc32861bc59881a26cf953/dtb/marvell/armada-3720-espressobin-v7-emmc.dtb from system_u:object_r:modules_object_t:s0 to system_u:object_r:boot_t:s0

```

So they are `modules_object_t` versus the policy expecting them to be `boot_t`. Not sure if this is a real problem or not.

Note I think this happens even on upgrades where there is no kernel update.

HuijingHei:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1808

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -162,10 +162,34 @@ wait-for-coreos-fix-selinux-labels() {
         echo "Waited for coreos-fix-selinux-labels.service to finish"
 }
 
+# Check if the rollback deployment has the dtb copy fix, which
+# means that the dtb files should have the correct SELinux labels.
+# https://github.com/coreos/fedora-coreos-tracker/issues/1808
+#
+# NOTE: we can drop this once the newest barrier release for all
+# streams is newer than 41.20241028.x.x.
+has_dtb_cp_fix() {
+    # The dtb copy issue was only ever an issue ever on aarch64
+    [ "$(arch)" != 'aarch64' ] && return 0
+    # We have the dtb copy fix if the rollback deployment is newer than
+    # when the fixed ostree was included. It should be fixed in the
+    # next round of releases after 41.20241028. Note 41.20241028.0.0
+    # is not a real build and uses `0` for the stream identifier, but
+    # should sort accordingly.
+    previous=$(rpm-ostree status --json | jq -r '.deployments[] | select(.booted == false).version')
+    if ! vergt $previous '41.20241028.0.0'; then
+        return 1
+    else
+        return 0
+    fi
+}
+
 selinux-sanity-check() {
     # First make sure the migrations/fix script has finished if this is the boot
     # where the fixes are taking place.
     wait-for-coreos-fix-selinux-labels
+    # Check to see if we have the dtb copy fix
+    has_dtb_cp_fix || add_dtb_exception='true'
     # Verify SELinux labels are sane. Migration scripts should have cleaned
     # up https://github.com/coreos/fedora-coreos-tracker/issues/1772
     unlabeled="$(find /sysroot -context '*unlabeled_t*' -print0 | xargs --null -I{} ls -ldZ '{}')"
@@ -206,8 +230,9 @@ selinux-sanity-check() {
             # Add in a few temporary glob exceptions
             # https://github.com/coreos/fedora-coreos-tracker/issues/1806
             [[ "${path}" =~ /etc/selinux/targeted/active/ ]] && continue
-            # https://github.com/coreos/fedora-coreos-tracker/issues/1808
-            [[ "${path}" =~ /boot/ostree/.*/dtb ]] && continue
+            if [ "${add_dtb_exception}" == 'true' ]; then
+                [[ "${path}" =~ /boot/ostree/.*/dtb ]] && continue
+            fi
             if [[ "${exceptions[$path]:-noexception}" == 'noexception' ]]; then
                 echo "Unexpected mislabeled file found: ${path}"
                 found="1"


dustymabe:
I'm not really sure if we can remove these exceptions yet. After an upgrade there are still two deployments around. Does the OSTree fix fix the old deployment too such that we guarantee all files in `/boot/` have the right labels?

You could try applying this patch to the upgrade test and then running it to see what the results are:

https://github.com/coreos/fedora-coreos-config/blob/7ef7e4044069e8c46fffb3d949809bd9ecac47e4/tests/kola/upgrade/extended/test.sh#L22-L28

so you would `buildfetch` an older build and start the test with that.

HuijingHei:
> Does the OSTree fix fix the old deployment too such that we guarantee all files in `/boot/` have the right labels?

Unluckily no, sorry that I just did testing using the fixed ostree and ignore the old deployment which already has the wrong label, in this case the old deployment will keep the wrong label as ostree will not copy `xattrs` for devicetree refer to https://github.com/ostreedev/ostree/pull/3323.

dustymabe:
> > Does the OSTree fix fix the old deployment too such that we guarantee all files in `/boot/` have the right labels?
> 
> Unluckily no, sorry that I just did testing using the fixed ostree and ignore the old deployment which already has the wrong label.

I think that's fine and expected. It just means we can't remove this exception for now.

OR maybe we could `rpm-ostree cleanup -r` first before doing the check. maybe that would work.

dustymabe:
Thinking about this a little more can we make it conditional? Would you mind testing something like on aarch64?

```patch
diff --git a/tests/kola/upgrade/extended/test.sh b/tests/kola/upgrade/extended/test.sh
index 4c880d96..3bca64aa 100755
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -162,10 +162,34 @@ wait-for-coreos-fix-selinux-labels() {
         echo "Waited for coreos-fix-selinux-labels.service to finish"
 }
 
+# Check if the rollback deployment has the dtb copy fix and if not
+# delete the rollback deployment.
+# https://github.com/coreos/fedora-coreos-tracker/issues/1808
+#
+# NOTE: we can drop this once the newest barrier release for all
+# streams is newer than 41.20241028.x.x.
+drop_rollback_if_needed() {
+    # The dtb copy issue was only ever an issue ever on aarch64
+    [ "$(arch)" != 'aarch64' ] && return
+    # We only need to drop if the rollback deployment is older than
+    # when the fixed ostree was included. It should be fixed in the
+    # next round of releases after 41.20241028. Note 41.20241028.0.0
+    # is not a real build and uses `0` for the stream identifier, but
+    # should sort accordingly.
+    previous=$(rpm-ostree status --json | jq -r '.deployments[] | select(.booted == false).version')
+    if ! vergt $previous '41.20241028.0.0'; then
+        echo "Dropping rollback deployment because it could have mislabeled dtb files"
+        rpm-ostree cleanup -r
+    fi
+}
+
 selinux-sanity-check() {
     # First make sure the migrations/fix script has finished if this is the boot
     # where the fixes are taking place.
     wait-for-coreos-fix-selinux-labels
+    # Next if the dtb copy fix wasn't present in the previous deployment
+    # then we'll delete that deployment before proceeding
+    drop_rollback_if_needed
     # Verify SELinux labels are sane. Migration scripts should have cleaned
     # up https://github.com/coreos/fedora-coreos-tracker/issues/1772
     unlabeled="$(find /sysroot -context '*unlabeled_t*' -print0 | xargs --null -I{} ls -ldZ '{}')"
```

HuijingHei:
> Would you mind testing something like on aarch64?

You are correct, the dtb files are only on aarch64, will try to find out how to run upgrade tests on aarch64, ~much appreciated if any pointer? Thanks!~

HuijingHei:
Do testing with this PR failed when running from `41.20241024.1.0` to latest `41.20241027.1.0` (which does not include the fixed ostree), maybe in this case we should not cleanup the rollback and still need the dtb exceptions?
steps:
```
# git clone -b remove-selinux-exception https://github.com/HuijingHei/fedora-coreos-config config

# cosa buildfetch --build=41.20241024.1.0 --artifact=qemu --arch=aarch64
# cosa decompress --build=41.20241024.1.0
# cosa kola run --build=41.20241024.1.0 --tag extended-upgrade -E ./config/
```

dustymabe:
> from `41.20241024.1.0` to latest `41.20241027.1.0` (which does not include the fixed ostree)

right. it won't work until the next builds on the prod streams, but you can test against the non-prod streams now, which is what all future tests will do:

```
$ cat tmp/target_stream.bu 
variant: fcos
version: 1.0.0
storage:
  files:
    - path: /etc/target_stream
      mode: 0644
      contents:
        inline: |
          next-devel
$ cosa kola run --build=41.20241024.1.0 --tag extended-upgrade --append-butane tmp/target_stream.bu --debug
```

This ^^ should work without your latest commit.

tests/kola/upgrade/extended/test.sh
```diff
@@ -162,6 +162,27 @@ wait-for-coreos-fix-selinux-labels() {
...
+drop_rollback_if_needed() {
+    # The dtb copy issue was only ever an issue ever on aarch64
+    [ "$(arch)" != 'aarch64' ] && return
+    # We only need to drop if the rollback deployment is older than
+    # when the fixed ostree was included. It should be fixed in the
+    # next round of releases after 41.20241028. Note 41.20241028.0.0
+    # is not a real build and uses `0` for the stream identifier, but
+    # should sort accordingly.
+    previous=$(rpm-ostree status --json | jq -r '.deployments[] | select(.booted == false).version')
+    if (! vergt $previous '41.20241028.0.0') && vergt $version '41.20241028.0.0'; then
```
HuijingHei: If previous version is older and current still does not include the fixed ostree, will not cleanup the rollback, not sure if this makes sense.

dustymabe: can you see if this works: https://github.com/coreos/fedora-coreos-config/pull/3237#issuecomment-2451728555

tests/kola/upgrade/extended/test.sh
```diff
@@ -206,8 +227,9 @@ selinux-sanity-check() {
             # Add in a few temporary glob exceptions
             # https://github.com/coreos/fedora-coreos-tracker/issues/1806
             [[ "${path}" =~ /etc/selinux/targeted/active/ ]] && continue
+            # Do not need this when the fixed ostree was included
             # https://github.com/coreos/fedora-coreos-tracker/issues/1808
-            [[ "${path}" =~ /boot/ostree/.*/dtb ]] && continue
+            verlt $version '41.20241028.0.0' && [[ "${path}" =~ /boot/ostree/.*/dtb ]] && continue
```
HuijingHei: Still need the dtb exception if the upgraded version does not include the fixed ostree.

tests/kola/upgrade/extended/test.sh
HuijingHei: With the patch in https://github.com/coreos/fedora-coreos-config/pull/3237#issuecomment-2450507016, failed but not related the code, seems the [next-devel](https://builds.coreos.fedoraproject.org/prod/streams/next-devel/builds/builds.json) latest is `41.20241027.10.0`, maybe should use `testing-devel`

tests/kola/upgrade/extended/test.sh
dustymabe: > Still need the dtb exception if the upgraded version does not include the fixed ostree.

This test gets run when new builds happen. Since the new OSTree RPM is already in `testing-devel` there should be no new builds that happen without the fixed OSTree RPM.

tests/kola/upgrade/extended/test.sh
dustymabe: yes! sorry. we disalbed `next-devel` earlier this week and I guess it didn't have the new OSTree RPM yet. Can you try with `testing-devel`?

HuijingHei: Yes, tried `testing-devel`, no lucky. 
I think it is because when doing upgrade based on older version (with unfixed ostree), which will make the two deployments both with wrong label, and cleanup the rollback only remove the previous deployment.

dustymabe: ahh. I see, yes. in that case dropping the rollback deployment doesn't really give us anything.

what we can do is still apply the exceptions but only in case that it's needed. something like:

```patch
diff --git a/tests/kola/upgrade/extended/test.sh b/tests/kola/upgrade/extended/test.sh
index 0e069e5a..39bc3278 100755
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -162,10 +162,34 @@ wait-for-coreos-fix-selinux-labels() {
         echo "Waited for coreos-fix-selinux-labels.service to finish"
 }
 
+# Check if the rollback deployment has the dtb copy fix, which
+# means that the dtb files should have the correct SELinux labels.
+# https://github.com/coreos/fedora-coreos-tracker/issues/1808
+#
+# NOTE: we can drop this once the newest barrier release for all
+# streams is newer than 41.20241028.x.x.
+has_dtb_cp_fix() {
+    # The dtb copy issue was only ever an issue ever on aarch64
+    [ "$(arch)" != 'aarch64' ] && return 0
+    # We have the dtb copy fix if the rollback deployment is newer than
+    # when the fixed ostree was included. It should be fixed in the
+    # next round of releases after 41.20241028. Note 41.20241028.0.0
+    # is not a real build and uses `0` for the stream identifier, but
+    # should sort accordingly.
+    previous=$(rpm-ostree status --json | jq -r '.deployments[] | select(.booted == false).version')
+    if ! vergt $previous '41.20241028.0.0'; then
+        return 0
+    else
+        return 1
+    fi
+}
+
 selinux-sanity-check() {
     # First make sure the migrations/fix script has finished if this is the boot
     # where the fixes are taking place.
     wait-for-coreos-fix-selinux-labels
+    # Check to see if we have the dtb copy fix
+    has_dtb_cp_fix || add_dtb_exception='true'$
     # Verify SELinux labels are sane. Migration scripts should have cleaned
     # up https://github.com/coreos/fedora-coreos-tracker/issues/1772
     unlabeled="$(find /sysroot -context '*unlabeled_t*' -print0 | xargs --null -I{} ls -ldZ '{}')"
@@ -207,7 +231,9 @@ selinux-sanity-check() {
             # https://github.com/coreos/fedora-coreos-tracker/issues/1806
             [[ "${path}" =~ /etc/selinux/targeted/active/ ]] && continue
             # https://github.com/coreos/fedora-coreos-tracker/issues/1808
-            [[ "${path}" =~ /boot/ostree/.*/dtb ]] && continue
+            if [ "${add_dtb_exception}" == 'true' ]; then
+                [[ "${path}" =~ /boot/ostree/.*/dtb ]] && continue
+            fi
             if [[ "${exceptions[$path]:-noexception}" == 'noexception' ]]; then
                 echo "Unexpected mislabeled file found: ${path}"
                 found="1"
```

HuijingHei:
> > Does the OSTree fix fix the old deployment too such that we guarantee all files in `/boot/` have the right labels?
> 
> Unluckily no, sorry that I just did testing using the fixed ostree and ignore the old deployment which already has the wrong label, in this case the old deployment will keep the wrong label as ostree will not copy `xattrs` for devicetree refer to [ostreedev/ostree#3323](https://github.com/ostreedev/ostree/pull/3323).

Sorry I misunderstood this, if booting os with unfixed ostree, then upgrade, then dtb files in both deployments will have the wrong labels. If in this case we get the fixed ostree, if do another upgrade, the old deployment will keep the wrong label, not sure if we have such upgrade test, for example upgrade A -> B -> C.

tests/kola/upgrade/extended/test.sh
HuijingHei: If update to following, test is passed.
```
+    if ! vergt $previous '41.20241028.0.0'; then
+        return 1
+    else
+        return 0
+    fi
```

dustymabe (approved):
LGTM

HuijingHei:
```
[coreos-assembler]$ cosa kola run --build=41.20241024.1.0 --tag extended-upgrade --append-butane tmp/target_stream.bu --debug -E ./config/
--- PASS: ext.config.upgrade.extended (350.46s)
PASS, output in tmp/kola
```

PR #3238, opened 2024-10-30
dnf5: use new syntax

nikita-dubrovskii:
Fedora 41 uses dnf5, which doesn't support `--set-enabled` and `--spec` switches:
```
Unknown argument "--set-enabled" for command "config-manager"
Unknown argument "--spec" for command "builddep".
```

### DIFF
--- a/ci/buildroot/install-buildroot.sh
+++ b/ci/buildroot/install-buildroot.sh
@@ -35,7 +35,7 @@ rm -rf "${tmpd:?}"/*
 echo "Installing build dependencies from canonical spec files"
 specs=$(grep -v '^#' "${dn}"/buildroot-specs.txt)
 (cd "${tmpd}" && echo "${specs}" | xargs curl -L --remote-name-all)
-(cd "${tmpd}" && find . -type f -print0 | xargs -0 dnf -y builddep --spec)
+(cd "${tmpd}" && find . -type f -print0 | xargs -0 dnf -y builddep)
 rm -rf "${tmpd:?}"/*
 
 echo "Installing test dependencies from canonical upstream files"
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

ci/buildroot/install-buildroot.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 
 dnf -y install dnf-plugins-core
 # We want to avoid a 7 day cycle for e.g. new ostree etc.
-dnf config-manager --set-enabled updates-testing
+dnf config-manager setopt updates-testing.enabled=1
```
c4rt0, line 8: I first thought it should be `--setopt`. According to [docs](https://dnf5.readthedocs.io/en/latest/dnf5_plugins/config-manager.8.html) I was wrong, so this lgtm.

ci/buildroot/install-buildroot.sh
```diff
@@ -35,7 +35,7 @@ rm -rf "${tmpd:?}"/*
 echo "Installing build dependencies from canonical spec files"
 specs=$(grep -v '^#' "${dn}"/buildroot-specs.txt)
 (cd "${tmpd}" && echo "${specs}" | xargs curl -L --remote-name-all)
-(cd "${tmpd}" && find . -type f -print0 | xargs -0 dnf -y builddep --spec)
+(cd "${tmpd}" && find . -type f -print0 | xargs -0 dnf -y builddep)
```
c4rt0, line 38: Honestly I don't know too much about this, but based on @jlebon [commit](https://github.com/jlebon/rpm-ostree/commit/0ba1ebafce94418c80440a1ca09e725ca2a72ae1) it makes sense.

c4rt0 (approved):
Not tested, but after a little bit of a research this /lgtm

PR #3239, opened 2024-11-01
07fix-selinux-labels: fix service ordering

dustymabe:
This fixes a mistake in b182027. Before/After directives need to go in the [Unit] section, not [Service].


ravanelli (approved):
/lgtm

gursewak1997 (approved):

PR #3177, opened 2024-09-23
Inherit from fedora-bootc's tier-x on Fedora 42+

jlebon:
There is a new tier-x in the fedora-bootc project whose goal is to
provide a common base that all variants (including tier-1) can share.

Move FCOS over to use this new tier, but only starting from Fedora 42.

This is a profound change and the start of an exciting new future! This
formalizes our relationship to other image-mode variants, encouraging us
to innovate and solve problems together in a more direct way.

Put more practically, e.g. bug fixes, new features, or temporary
workarounds that concern all/most tier-x derivatives should probably be
carried out at the tier-x level rather than the CoreOS level.

Eventually, this inheritance will be made even more explicit by having
FCOS be built `FROM` the tier-x image. For now, we share at the manifest
level, which is a stepping stone towards that goal.

Patches to actually dedupe our manifests with tier-x will follow. Though
note there is no change in the resulting package set here.

PR #3246, opened 2024-11-08
manifest: Add bootc

jmarrero:
follows https://github.com/coreos/fedora-coreos-config/pull/2758

https://fedoraproject.org/wiki/Changes/DNFAndBootcInImageModeFedora

We should ship this along with dnf5 for 41.

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -159,6 +159,8 @@ packages:
   - iptables-legacy
   # NIC firmware we've traditionally shipped but then were split out of linux-firmware in Fedora
   - qed-firmware # https://github.com/coreos/fedora-coreos-tracker/issues/1746
+  # https://fedoraproject.org/wiki/Changes/DNFAndBootcInImageModeFedora
+  - bootc
 
 
 # - irqbalance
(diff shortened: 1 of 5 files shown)


cgwalters:
I think you also need to manually update the lockfile or so?

We should also xref here https://src.fedoraproject.org/rpms/rpm-ostree/c/e31f656abc0d451f0ffdd2a3afd60944796c2246?branch=rawhide which is already in rawhide.

dustymabe:
I'd think this would need to go in a normal package list and not the top level manifest.

manifests/fedora-coreos-base.yaml
```diff
@@ -159,6 +159,8 @@ packages:
   - iptables-legacy
   # NIC firmware we've traditionally shipped but then were split out of linux-firmware in Fedora
   - qed-firmware # https://github.com/coreos/fedora-coreos-tracker/issues/1746
+  # https://fedoraproject.org/wiki/Changes/DNFAndBootcInImageModeFedora
+  - bootc
```
jlebon, line 163: This intersects with https://github.com/coreos/fedora-coreos-config/pull/3177. No need to worry about it here. I'll move this line to the most appropriate manifest in my next rebase of that PR.

jlebon (approved):

jlebon:
Once this merges, can you open a PR to remove it from the rawhide `manifest.yaml`?

jmarrero:
> Once this merges, can you open a PR to remove it from the rawhide `manifest.yaml`?

will do

PR #3248, opened 2024-11-08
overrides: Pin `kdump-utils-1.0.48-1.fc42` for ppc64le

gursewak1997:
Pin `kdump-utils-1.0.48-1.fc42` for arch-specific overrides for only ppc64le.
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1830


dustymabe:
open PR against `rawhide` branch?

PR #3253, opened 2024-11-13
overlay.d & tests: Add alternatives migration and test and remove iptables-legacy

travier:
overlay.d: Add empty statoverride config files

---

overlay.d & tests: Add alternatives migration and test

- Add an overlay with the migration logic for alternatives
- Add a test for the migration script

This should make sure that the system is setup properly and that the
migration script will do the right thing on older systems.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1818
See: https://github.com/coreos/fedora-coreos-tracker/issues/677
See: https://docs.fedoraproject.org/en-US/fedora-coreos/alternatives/

 - Remove iptables-legacy starting with Fedora 43


tests/kola/files/alternatives
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+## kola:
+##   exclusive: true
```
HuijingHei: We might do not need this, refer to https://github.com/coreos/coreos-assembler/blob/main/docs/kola/external-tests.md, `When the exclusive key is not provided, tests are marked exclusive: true by default.`

tests/kola/files/alternatives
```diff
@@ -0,0 +1,40 @@
...
+fi
+if ! test -d "/etc/alternatives"; then
+    fatal "Error: '/etc/alternatives' is missing"
+fi
+if ! test -d "/etc/alternatives-admindir"; then
+    fatal "Error: '/etc/alternatives-admindir' is missing"
+fi
+
+# Re-create the setup form an older FCOS node
+alternatives --admindir /etc/alternatives-admindir --set iptables /usr/sbin/iptables-legacy
```
HuijingHei: On ostree system, the admindir should be `/etc/alternatives-admindir` by default, do we still need the `--admindir` option?

travier: We indeed don't need it for this one. I added it here to be explicit but we can also remove it.

tests/kola/files/alternatives
```diff
@@ -0,0 +1,46 @@
...
+if ! test -d "/etc/alternatives"; then
+    fatal "Error: '/etc/alternatives' is missing"
+fi
+if ! test -d "/etc/alternatives-admindir"; then
+    fatal "Error: '/etc/alternatives-admindir' is missing"
+fi
+
+# To test the migration we will re-create the setup from an older FCOS node
+
+# First, reset iptables to the lefacy backend
```
aaradhak: looks like a typo here in the comment - "legacy"

travier:
I've now added the migration script to this PR and a test for it. Not fully tested yet.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -euo pipefail
+# set -x
+
+main() {
+    # Should never happen as systemd checks this, but just in case
+    if [[ ! -d "/var/lib/alternatives" ]]; then
```
HuijingHei: On silverblue, check `/var/lib/alternatives` is broken link, should we care about this?
```
$ ls /var/lib/alternatives  -al
lrwxrwxrwx. 1 root root 26 Oct 16 07:16 /var/lib/alternatives -> ../../usr/lib/alternatives
```

travier: Indeed, if we want to have a single migration script for all variants then we should take this one into account.

HuijingHei (dismissed):
Overall LGTM.

PeaceRebel:
This test was failing after rpm-ostree update including https://github.com/coreos/rpm-ostree/pull/5368. Updated the path to `/usr/bin` instead of `/usr/sbin`

overlay.d/07fix-selinux-labels/statoverride
```diff
@@ -0,0 +1,2 @@
+# Config file for overriding permission bits on overlay files/dirs
```
jlebon: See https://github.com/coreos/fedora-coreos-config/commit/53cfd24445593f47e8bf9bde2fa6cb9b0340450c. Let's no longer use statoverride files.

Those are all empty, so we can just drop them.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
c4rt0: I came across this on `Silverblue` today as well, the /var/lib/alternatives link is still broken

c4rt0 (commented):
This is the first time I come across the use of `alternatives` for iptables, regardless...

/lgtm

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,23 @@
...
+        echo "Skipped /var/lib/alternatives as it is not a directory"
+        exit 0
+    fi
+
+    # We can safely directly try to remove the directory as rmdir will fail on
+    # a non-empty directory
+    rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty"
```
dustymabe: If this fails and the directory continues to exist, what are the consequences?

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,23 @@
...
+
+    # Do the migration, explicitely using the new configuration directory to
```
dustymabe:
```suggestion
    # Do the migration, explicitly using the new configuration directory to
```

tests/kola/files/alternatives
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+## kola:
+##   description: Verify that the alternatives config is properly migrated and test the migration
```
dustymabe, line 3: Are we planning to run this migration on RHCOS too?

If not, we should limit this test to FCOS only.

tests/kola/files/alternatives
```diff
@@ -0,0 +1,79 @@
...
+    fatal "Error: Found '/var/lib/alternatives' which should not exists"
+fi
+if ! test -d "/etc/alternatives"; then
+    fatal "Error: '/etc/alternatives' is missing"
+fi
+if ! test -d "/etc/alternatives-admindir"; then
+    fatal "Error: '/etc/alternatives-admindir' is missing"
+fi
+
+# To test the migration we will re-create the setup from an older FCOS node
```
dustymabe, line 30: Maybe instead of "re-create the setup" we should consider adding some verification of this migration to [the extended upgrade test](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/upgrade/extended/test.sh).

overlay.d/50alternatives/usr/systemd/system/coreos-alternatives-migration.service
```diff
@@ -0,0 +1,12 @@
...
+ConditionPathExists=/var/lib/alternatives
+ConditionPathIsDirectory=/var/lib/alternatives
+
+[Service]
+ExecStart=/usr/libexec/coreos-alternatives-migration
+Type=oneshot
+RemainAfterExit=yes
+
+[Install]
+WantedBy=basic.target
```
dustymabe: I feel like switching out the `iptables` implementation mid boot might be problematic. Is there a way we can make sure this "migration" runs early enough that nothing would have been leveraging `iptables` yet?

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
PeaceRebel: It should use the default dir as mentioned [here](https://github.com/fedora-sysv/chkconfig/blob/a14c6340ac39e0bf87b0d049adc8d0c05bef79a5/alternatives.8#L217). I'm not sure why we created `/etc/alternatives-admindir` or if I am missing some  information.

PeaceRebel: Looks like there is some inconsistency between the man page and the code. https://github.com/fedora-sysv/chkconfig/blob/a14c6340ac39e0bf87b0d049adc8d0c05bef79a5/alternatives.c#L1517

PeaceRebel: > If this fails and the directory continues to exist, what are the consequences?

It should keep using `/var/lib/altenatives`. Should we check `isOstree` first [here](https://github.com/fedora-sysv/chkconfig/blob/a14c6340ac39e0bf87b0d049adc8d0c05bef79a5/alternatives.c#L1514-L1517)?

overlay.d/50alternatives/usr/systemd/system/coreos-alternatives-migration.service
PeaceRebel: @dustymabe Is `sysinit.target` good enough for this?

tests/kola/files/alternatives
PeaceRebel: I have added a check in this test.

overlay.d/50alternatives/usr/lib/systemd/system/coreos-alternatives-migration.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Migrate systems to fixed alternatives configuration
+ConditionPathExists=/var/lib/alternatives
+ConditionPathIsDirectory=/var/lib/alternatives
```
dustymabe: I'm guessing these conditions will hopefully ensure that the migration only runs once. Is that correct?

If so let's add a comment here to that effect saying the directory gets deleted as part of the migration.

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,20 @@ selinux-sanity-check() {
 
 ok "Reached version: $version"
 
+verify-alternatives-migration() {
+    # Verify iptables migration
+    if [[ $(alternatives --admindir /etc/alternatives-admindir --display iptables | grep -c -E 'link currently points to /usr/(bin|sbin)/iptables-nft') != "1" ]]; then
+        systemctl status coreos-alternatives-migration.service
```
dustymabe: This service won't run on every boot (i.e. think about when the migration happened 2 or 3 "upgrades" ago, this `verify-alternatives-migration` only runs on the final fully upgraded system) so I don't think we can rely on getting `status` output here.

dustymabe: what we may be able to do, though is query the journal for messages related to the migration.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
dustymabe, line 1: Can we add a few log statements to this script so we have a little more of a trail in the journal acknowledging the migration is taking place and if it was successful.

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,20 @@ selinux-sanity-check() {
 
 ok "Reached version: $version"
 
+verify-alternatives-migration() {
+    # Verify iptables migration
+    if [[ $(alternatives --admindir /etc/alternatives-admindir --display iptables | grep -c -E 'link currently points to /usr/(bin|sbin)/iptables-nft') != "1" ]]; then
```
dustymabe: I've gone around to a few older systems I have and I don't see any of them that don't say:

```
link currently points to /usr/sbin/iptables-nft
```

Is this a proper test that the migration happened?

overlay.d/50alternatives/usr/systemd/system/coreos-alternatives-migration.service
dustymabe: I'm not 100% sure. @jlebon or @travier might have a good answer

dustymabe (commented):
This is looking good. A few comments (mostly clarifying questions).

I really don't feel comfortable merging this without a review from @travier though. He should be back next week I think.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
dustymabe: I'm not sure honestly. cc @travier

overlay.d/50alternatives/usr/systemd/system/coreos-alternatives-migration.service
jlebon: A foolproof way to do this is to do it from the initramfs. Probably not worth the churn though. I think `sysinit.target` is probably fine.

tests/kola/upgrade/extended/test.sh
PeaceRebel: Oops. I added that for debugging.

tests/kola/upgrade/extended/test.sh
PeaceRebel: If I run the script on a new/old system, the path changes to /usr/bin

dustymabe: I think what I'm trying to point out isn't the `bin` vs `sbin`, but I'm asking the question: in what case would this test fail?

i.e. on my systems (even older deployed ones) they all already say link currently points to nft.

IOW is this a proper test that a migration has taken place?

PeaceRebel: Hmm, this validates the final state we want, doesn't it? I could be misunderstanding something here.

It doesn't check if the migration ran or not. That could be done with a journal log check, but it only works if upgrades are from version with the older alternatives admindir.

dustymabe: > Hmm, this validates the final state we want, doesn't it? I could be misunderstanding something here.

Yes. I think it's fine to just check the final state of the fully upgraded system to verify it's in the state we want. i.e. whether a migration was necessary or not shouldn't matter here.

Here's an example. I just started a `35.20211119.3.0` system (note iptables-nft wasn't [default for new nodes until F36](https://github.com/coreos/fedora-coreos-tracker/issues/676#issuecomment-1137829912)) and let it upgrade all the way to F42 (latest). Here's what I see:

```
localhost login: core
Password: 
Fedora CoreOS 42.20250526.3.0

[core@localhost ~]$ sudo bootupctl status | grep aleph
Running as unit: bootupd.service
[WARN  bootupd::bios] Get error: Found multiple parent devices /dev/vda and /dev/vdb; not currently supported
CoreOS aleph version: 35.20211119.3.0

[core@localhost ~]$ alternatives --admindir /etc/alternatives-admindir --display iptables
iptables - status is manual.
 link currently points to /usr/sbin/iptables-nft
/usr/bin/iptables-legacy - priority 10
 follower ip6tables: /usr/bin/ip6tables-legacy
 follower ip6tables-restore: /usr/bin/ip6tables-legacy-restore
 follower ip6tables-save: /usr/bin/ip6tables-legacy-save
 follower iptables-restore: /usr/bin/iptables-legacy-restore
 follower iptables-save: /usr/bin/iptables-legacy-save
/usr/bin/iptables-nft - priority 10
 follower ip6tables: /usr/bin/ip6tables-nft
 follower ip6tables-restore: /usr/bin/ip6tables-nft-restore
 follower ip6tables-save: /usr/bin/ip6tables-nft-save
 follower iptables-restore: /usr/bin/iptables-nft-restore
 follower iptables-save: /usr/bin/iptables-nft-save
Current `best' version is /usr/bin/iptables-legacy.
```

So this test will pass (and passes today) prior to any migration script running (i.e. this PR merging). 

> 
> It doesn't check if the migration ran or not. That could be done with a journal log check, but it only works if upgrades are from version with the older alternatives admindir.

Right, but my question here is more about the need for the migration. If this test passes on a pre-migrated system that has been upgraded from a really old starting point, then what is it really testing?

i.e. considering this test is called `verify-alternatives-migration()` how do we more fully test the migration?

overlay.d/50alternatives/usr/lib/systemd/system/coreos-alternatives-migration.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Migrate systems to fixed alternatives configuration
+# This directory is deleted by the script
```
dustymabe: A little more would help those who might not see the relationship:

```suggestion
# This directory is deleted by the script, which means the "migration" will only ever run once.
```

travier: agree

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
travier: Let's keep this for FCOS only and we'll do the migration for Atomic Desktop independently.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
travier: So, the idea is that we update what iptables point to to use the nft variant, regardless of the state of the config or migration.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1818#issuecomment-2474312151

tests/kola/upgrade/extended/test.sh
travier: We should get this result without having to override the config directory:
```suggestion
    if [[ $(alternatives --display iptables | grep -c -E 'link currently points to /usr/(bin|sbin)/iptables-nft') != "1" ]]; then
```

travier: If we want to test the other path of this migration (the case where an admin had set some config manually), we would have to create a fake config in `/var/lib/alternatives`. See the second part of the kola test: https://github.com/coreos/fedora-coreos-config/pull/3253/files#diff-17d63b84cd29575ee973f0cf86cd6d68c4613ec1ae14f56674eee4e69a06dc54R65

We probably should not do that however as that would mean duplicating the test and I don't think anybody did that (but it's here just in case).

tests/kola/upgrade/extended/test.sh
```diff
@@ -256,6 +269,8 @@ if vereq $version $target_version; then
     fi
     # One last check!
     selinux-sanity-check
+    #One more last check
```
travier:
```suggestion
    # One more last check!
```
:D

travier (commented):
Mostly LGTM. Some nits.

travier (commented):
Mostly LGTM. Some nits.

tests/kola/upgrade/extended/test.sh
travier: Ah, we should also check that `/var/lib/alternatives` does not exists anymore here. This is the part that should fail today.

dustymabe: ok on that "upgraded from F35" system I see:

```
[core@localhost ~]$ alternatives --display iptables
[core@localhost ~]$
```

Does that tell you anything?

travier: If you get that results, it means that the migration did not happen? This is a system without this migration script, right?

travier: Can you do `ls -l /var/lib/alternatives`?

overlay.d/50alternatives/usr/systemd/system/coreos-alternatives-migration.service
travier: Does sysinit guarantees us that `/var` is mounted? we need that to be the case. Maybe we add a Requires mount for `/var`?

PeaceRebel: IIUC, `/var` should be mounted as this target is after `local-fs.target`

tests/kola/upgrade/extended/test.sh
dustymabe: > If you get that results, it means that the migration did not happen? This is a system without this migration script, right?

Correct. The F35 upgraded system does not include this PR. So that is expected then.

What I wanted to do was transform the test here into something that tested the migration happened. The command from before worked whether the migration happened or not.

PeaceRebel: > If we want to test the other path of this migration (the case where an admin had set some config manually), we would have to create a fake config in /var/lib/alternatives. See the second part of the kola test: https://github.com/coreos/fedora-coreos-config/pull/3253/files#diff-17d63b84cd29575ee973f0cf86cd6d68c4613ec1ae14f56674eee4e69a06dc54R65

> We probably should not do that however as that would mean duplicating the test and I don't think anybody did that (but it's here just in case).

I've added this check now.

PeaceRebel: > What I wanted to do was transform the test here into something that tested the migration happened. The command from before worked whether the migration happened or not.

I've added a log to the migration script. Should checking if this show up in the journal logs be enough?

overlay.d/50alternatives/usr/systemd/system/coreos-alternatives-migration.service
travier: Just to be sure, let's add `RequiresMountsFor=/var` in the `[Unit]` section.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,29 @@
...
+        exit 0
+    fi
+
+    # We can safely directly try to remove the directory as rmdir will fail on
+    # a non-empty directory
+    rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty"
+
+    # Do the migration, explicitly using the new configuration directory to
+    # ignore /var/lib/alternatives if it still exists
+    alternatives --admindir /etc/alternatives-admindir --set iptables /usr/bin/iptables-nft
```
travier:
```suggestion
    alternatives --admindir /etc/alternatives-admindir --set iptables /usr/bin/iptables-nft \
      || (echo "Setting up iptables-nft alternatives failed" ; exit 1)
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,29 @@
...
+main() {
+    # Should never happen as systemd checks this, but just in case
+    if [[ ! -d "/var/lib/alternatives" ]]; then
+        echo "Skipped /var/lib/alternatives as it is not a directory"
+        exit 0
+    fi
+
+    # We can safely directly try to remove the directory as rmdir will fail on
+    # a non-empty directory
+    rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty"
```
travier:
```suggestion
    rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty"
    if [[ ! -d "/var/lib/alternatives" ]]; then
        echo "Successfully removed /var/lib/alternatives"
    fi
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,29 @@
...
+
+    # Do the migration, explicitly using the new configuration directory to
+    # ignore /var/lib/alternatives if it still exists
+    alternatives --admindir /etc/alternatives-admindir --set iptables /usr/bin/iptables-nft
+    migration_err=$?
```
travier:
```suggestion
```

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,24 @@ selinux-sanity-check() {
...
+    # Verify iptables migration
+    if [[ $(alternatives --display iptables | grep -c -E 'link currently points to /usr/(bin|sbin)/iptables-nft') != "1" ]]; then
+        fatal "Error: migration did not set iptables to nft backend"
+    fi
+    if [[ $(iptables --version | grep -c "nf_tables") != "1" ]]; then
+        fatal "Error: iptables not reset to nftables backend"
+    fi
+
+    ok "alternatives migration verification passed."
+
```
travier:
```suggestion
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
travier: We added those.

travier:
I think we are good here. LGTM. I can not approve this PR as I'm the one that opened it initially so we'll need an approve from someone else.

travier:
@dustymabe can you take a last look at this one and we'll merge it? Thanks

tests/kola/files/alternatives
```diff
@@ -0,0 +1,79 @@
...
+# - https://github.com/coreos/fedora-coreos-tracker/issues/1818
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+if test -e "/var/lib/alternatives"; then
+    ls -al "/var/lib/alternatives"
+    fatal "Error: Found '/var/lib/alternatives' which should not exists"
```
c4rt0:
```suggestion
    fatal "Error: Found '/var/lib/alternatives' which should not exist"
```

HuijingHei:
Overall the script looks good to me. Thank you for the work!
Did testing that upgrade from 40.20240424.20.1, and check that the service skipped as `/var/lib/alternatives` is not directory, and it is symlink by default on FCOS, just curious in which case we would make the service running.

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,23 @@ selinux-sanity-check() {
...
+    # Verify /var/lib/alternatives dir is removed
+    if [ -d /var/lib/alternatives ]; then
+        fatal "Error: migration didn't remove /var/lib/alternatives"
+    fi
+
+    # Verify iptables migration
+    if [[ $(alternatives --display iptables | grep -c -E 'link currently points to /usr/(bin|sbin)/iptables-nft') != "1" ]]; then
+        fatal "Error: migration did not set iptables to nft backend"
+    fi
+    if [[ $(iptables --version | grep -c "nf_tables") != "1" ]]; then
```
c4rt0, line 263: Just wondering, as in all of the examples above we are using the `grep -c` for count of occurrences. If anywhere for any reason the count would result in number higher than 1 this would fail. Should we not use the `=~` operator instead?
I mean something like:
```
if ! [[ $(alternatives --display iptables) =~ "link currently points to /usr/bin/iptables-legacy" ]]; then
    fatal "Failed to set alternatives configuration to legacy backend"
fi
```

c4rt0 (dismissed):
Some questions, but overall :

/lgtm

tests/kola/upgrade/extended/test.sh
PeaceRebel: There should only be one entry for iptables-nft, isn't? That's what I have observed from the current usage of `alternatives`.

c4rt0: As far as I understand it yes - there should be only one.
I asked the question as running the below returns 3 results:

```
[core@cosa-devsh bin]$ ls | grep "iptables-nft"
...
iptables-nft
iptables-nft-restore
iptables-nft-save
```

As I saw `grep -c` multiple times in this PR I was just curious if that's something of interest.

travier: Output from `iptables --version` is a single line:

```
$ iptables --version
iptables v1.8.11 (nf_tables)
```

travier:
> Overall the script looks good to me. Thank you for the work! Did testing that upgrade from 40.20240424.20.1, and check that the service skipped as `/var/lib/alternatives` is not directory, and it is symlink by default on FCOS, just curious in which case we would make the service running.

Hum somehow I had missed that:

```
State: idle
  DriverState: inactive
Deployments:
● fedora:fedora/x86_64/coreos/stable
                  Version: 40.20240416.3.1 (2024-05-07T03:03:33Z)
                   Commit: fb67df533ff58f5770b735496e624752517f48d27e29ec0f5c47aad145afda71
[core@cosa-devsh ~]$ ls -alh /var/lib/alternatives
lrwxrwxrwx. 1 root root 26 Jun 25 15:39 /var/lib/alternatives -> ../../usr/lib/alternatives
[core@cosa-devsh ~]$ ls -alh /usr/lib/alternatives/
total 24K
drwxr-xr-x.  2 root root  108 Jan  1  1970 .
drwxr-xr-x. 42 root root 4.0K Jan  1  1970 ..
-rw-r--r--.  2 root root  584 Jan  1  1970 arptables
-rw-r--r--.  2 root root  114 Jan  1  1970 cifs-idmap-plugin
-rw-r--r--.  2 root root  277 Jan  1  1970 ebtables
-rw-r--r--.  2 root root  581 Jan  1  1970 iptables
-rw-r--r--.  2 root root   69 Jan  1  1970 libnssckbi.so.x86_64
[core@cosa-devsh ~]$ ls -alh /etc/alternatives/
total 16K
drwxr-xr-x.  2 root root 4.0K May  7  2024 .
drwxr-xr-x. 85 root root 8.0K Jun 25 15:39 ..
lrwxrwxrwx.  1 root root   23 May  7  2024 arptables -> /usr/sbin/arptables-nft
lrwxrwxrwx.  1 root root   33 May  7  2024 arptables-helper -> /usr/libexec/arptables-nft-helper
lrwxrwxrwx.  1 root root   31 May  7  2024 arptables-restore -> /usr/sbin/arptables-nft-restore
lrwxrwxrwx.  1 root root   28 May  7  2024 arptables-save -> /usr/sbin/arptables-nft-save
lrwxrwxrwx.  1 root root   39 May  7  2024 cifs-idmap-plugin -> /usr/lib64/cifs-utils/cifs_idmap_sss.so
lrwxrwxrwx.  1 root root   22 May  7  2024 ebtables -> /usr/sbin/ebtables-nft
lrwxrwxrwx.  1 root root   30 May  7  2024 ebtables-restore -> /usr/sbin/ebtables-nft-restore
lrwxrwxrwx.  1 root root   27 May  7  2024 ebtables-save -> /usr/sbin/ebtables-nft-save
lrwxrwxrwx.  1 root root   23 May  7  2024 ip6tables -> /usr/sbin/ip6tables-nft
lrwxrwxrwx.  1 root root   31 May  7  2024 ip6tables-restore -> /usr/sbin/ip6tables-nft-restore
lrwxrwxrwx.  1 root root   28 May  7  2024 ip6tables-save -> /usr/sbin/ip6tables-nft-save
lrwxrwxrwx.  1 root root   22 May  7  2024 iptables -> /usr/sbin/iptables-nft
lrwxrwxrwx.  1 root root   30 May  7  2024 iptables-restore -> /usr/sbin/iptables-nft-restore
lrwxrwxrwx.  1 root root   27 May  7  2024 iptables-save -> /usr/sbin/iptables-nft-save
lrwxrwxrwx.  1 root root   34 May  7  2024 libnssckbi.so.x86_64 -> /usr/lib64/pkcs11/p11-kit-trust.so
```

This is weird, I don't remember seeing that before. We would have to remove that symlink as well here to migrate those systems. I'm letting one update to see what the result it.

travier:
OK, I've booted a `40.20240416.3.1` image and had it run the updates and I end up with a broken symlink for `/var/lib/alternatives`:

```
[core@cosa-devsh ~]$ ls -alh /var/lib/alternatives 
lrwxrwxrwx. 1 root root 26 Jun 25 15:45 /var/lib/alternatives -> ../../usr/lib/alternatives
```

So this one is "easy", we can just remove it. We'll have to loosen the unit conditions and update the migration script.

travier:
Thanks again @HuijingHei for catching that.

HuijingHei:
And sorry for the late question, should do testing earlier.

As the service is to migrate existing to `iptables-nft`, how about checking if `iptables --version` is legacy, then run `alternatives --admindir /etc/alternatives-admindir --set iptables /usr/bin/iptables-nft` ? Also if `/var/lib/alternatives` is broken symlink, then remove it.

Edit: or maybe `/usr/sbin/alternatives --set iptables /usr/sbin/iptables-nft` refer to [doc](https://docs.fedoraproject.org/en-US/fedora-coreos/alternatives/#_using_alternatives_commands)?

PeaceRebel:
> OK, I've booted a `40.20240416.3.1` image and had it run the updates and I end up with a broken symlink for `/var/lib/alternatives`:
> 
> ```
> [core@cosa-devsh ~]$ ls -alh /var/lib/alternatives 
> lrwxrwxrwx. 1 root root 26 Jun 25 15:45 /var/lib/alternatives -> ../../usr/lib/alternatives
> ```
> 
> So this one is "easy", we can just remove it. We'll have to loosen the unit conditions and update the migration script.

@travier I removed the `ConditionPathIsDirectory` from the service and modified the migration script to remove `/var/lib/alternatives` if it's a symlink.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,33 @@
...
+# set -x
+
+main() {
+    if [[ ! -d "/var/lib/alternatives" ]]; then
+        echo "Skipped /var/lib/alternatives as it is not a directory"
+        exit 0
+    fi
+
+    if [ -L "/var/lib/alternatives" ]; then
+        sudo unlink "/var/lib/alternatives"
```
travier: No sudo needed here, we are already running as root.

```suggestion
        unlink "/var/lib/alternatives"
```

travier: We can probably use rm directly here instead of unlink.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,33 @@
...
+        echo "unlinked /var/lib/alternatives"
```
travier:
```suggestion
        echo "Removed /var/lib/alternatives symlink"
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,33 @@
...
+    else
+        # We can safely directly try to remove the directory as rmdir will fail on
+        # a non-empty directory
+        sudo rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty or doesn't exist"
```
travier:
```suggestion
        rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty or doesn't exist"
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,33 @@
...
+    fi
+
+    if [[ ! -d "/var/lib/alternatives" ]]; then
+        echo "successfully removed /var/lib/alternatives"
+    fi
+
+    # Do the migration, explicitly using the new configuration directory to
+    # ignore /var/lib/alternatives if it still exists
+    sudo alternatives --admindir /etc/alternatives-admindir --set iptables /usr/bin/iptables-nft \
```
travier:
```suggestion
    alternatives --admindir /etc/alternatives-admindir --set iptables /usr/bin/iptables-nft \
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,33 @@
...
+        sudo unlink "/var/lib/alternatives"
+        echo "unlinked /var/lib/alternatives"
+    else
+        # We can safely directly try to remove the directory as rmdir will fail on
+        # a non-empty directory
+        sudo rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty or doesn't exist"
+    fi
+
+    if [[ ! -d "/var/lib/alternatives" ]]; then
+        echo "successfully removed /var/lib/alternatives"
```
travier:
```suggestion
    if [[ ! -e "/var/lib/alternatives" ]]; then
        echo "Successfully removed /var/lib/alternatives"
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -euo pipefail
+# set -x
+
+main() {
+    if [[ ! -d "/var/lib/alternatives" ]]; then
+        echo "Skipped /var/lib/alternatives as it is not a directory"
+        exit 0
+    fi
```
travier: We also need to remove this check now

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,23 @@ selinux-sanity-check() {
 
 ok "Reached version: $version"
 
+verify-alternatives-migration() {
+    # Verify /var/lib/alternatives dir is removed
```
travier, line 254: We should now also check that there are no symlink here as well

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
PeaceRebel: Oops. I added sudo to check it on a running vm. Missed to remove it before committing.

travier:
> As the service is to migrate existing to `iptables-nft`, how about checking if `iptables --version` is legacy, then run `alternatives --admindir /etc/alternatives-admindir --set iptables /usr/bin/iptables-nft` ? Also if `/var/lib/alternatives` is broken symlink, then remove it.

I'm not sure I follow. Isn't that what the migration script is doing in this PR?

Edit: The "remove symlink" part has just been added indeed, but the "switch to the iptables-nft backend" one is already there.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
HuijingHei: Better to check if `iptables --version` is legacy before the command, and we do not set again if it is already nft.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
HuijingHei: Also check if it is broken like `if [ -L "$file" ] && [ ! -e "$file" ]; then`

PeaceRebel: @HuijingHei Isn't -L enough for this?

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
PeaceRebel: changed to `-e`, hope this is enough.

tests/kola/upgrade/extended/test.sh
PeaceRebel: changed to `-e`, hope this is enough

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
HuijingHei: -L means we remove if it is symlink, in my testing after upgrade it is always a broken symlink. I am OK to just keep -L.

PeaceRebel: IIUC, -L will be true even if it broken symlink.

travier: While theoretically possible, I think it's very unlikely that someone would have replaced the symlink with another one. In both cases, removing the symlink should not lead to configuration loss (the directory pointed by the symlink will stay there), so it should be fine to just remove it directly.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
travier: We could do that but why bother? The result would be the same. So if we just do it, we're fairly sure that it happened?

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -euo pipefail
+# set -x
+
+main() {
+    if [ -L "/var/lib/alternatives" ]; then
```
travier:
```suggestion
    if [[ -L "/var/lib/alternatives" ]]; then
```

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,28 @@
...
+        rm "/var/lib/alternatives"
+        echo "Removed /var/lib/alternatives symlink"
+    else
+        # We can safely directly try to remove the directory as rmdir will fail on
+        # a non-empty directory
+        rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty or doesn't exist"
```
travier: The directory not existing should never happen as we condition the unit on it existing.

```suggestion
        rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty"
```

tests/kola/files/alternatives
```diff
@@ -0,0 +1,79 @@
...
+##   distros: fcos
+# See
+# - https://github.com/coreos/fedora-coreos-tracker/issues/1818
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+if test -e "/var/lib/alternatives"; then
```
travier:
```suggestion
if [[ -e "/var/lib/alternatives" ]]; then
```

tests/kola/files/alternatives
```diff
@@ -0,0 +1,79 @@
...
+    ls -al "/var/lib/alternatives"
+    fatal "Error: Found '/var/lib/alternatives' which should not exist"
+fi
+if ! test -d "/etc/alternatives"; then
```
travier:
```suggestion
if [[ ! -d "/etc/alternatives" ]]; then
```

tests/kola/files/alternatives
```diff
@@ -0,0 +1,79 @@
...
+    fatal "Error: '/etc/alternatives' is missing"
+fi
+if ! test -d "/etc/alternatives-admindir"; then
```
travier:
```suggestion
if [[ ! -d "/etc/alternatives-admindir" ]]; then
```

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,23 @@ selinux-sanity-check() {
...
+    if [ -e /var/lib/alternatives ]; then
```
travier:
```suggestion
    if [[ -e /var/lib/alternatives ]]; then
```

travier (commented):
Thanks, LGTM

HuijingHei (dismissed):
LGTM

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,28 @@
...
+# set -x
+
+main() {
+    if [[ -L "/var/lib/alternatives" ]]; then
+        rm "/var/lib/alternatives"
+        echo "Removed /var/lib/alternatives symlink"
+    else
+        # We can safely directly try to remove the directory as rmdir will fail on
+        # a non-empty directory
+        rmdir "/var/lib/alternatives" || echo "Warning: /var/lib/alternatives is not empty"
```
dustymabe, line 13: Since this can successfully fail it means the `/var/lib/alternatives` dir can stick around for future boots, which means `coreos-alternatives-migration.service` will run every boot (i.e. `ConditionPathExists=/var/lib/alternatives`)

Maybe you already know that and we're OK with it, but figured I'd point it out.

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,23 @@ selinux-sanity-check() {
...
 ok "Reached version: $version"
 
+verify-alternatives-migration() {
+    # Verify /var/lib/alternatives dir is removed
+    if [[ -e /var/lib/alternatives ]]; then
+        fatal "Error: migration didn't remove /var/lib/alternatives"
+    fi
+
+    # Verify iptables migration
+    if [[ $(alternatives --display iptables | grep -c -E 'link currently points to /usr/(bin|sbin)/iptables-nft') != "1" ]]; then
```
dustymabe, line 260: why in some of the commands we specify `--admindir` and some of them we don't?

dustymabe (commented):
considering this comment from the description of https://github.com/coreos/fedora-coreos-tracker/issues/1818

> As this could potentially be a breaking change, we might want to do it only starting with Fedora 42.

Should we only roll this in for Fedora 43?

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -euo pipefail
+# set -x
+
+main() {
+    if [[ -L "/var/lib/alternatives" ]]; then
+        rm "/var/lib/alternatives"
+        echo "Removed /var/lib/alternatives symlink"
+    else
```
dustymabe, line 10: lets make this check specifically if it's a directory:


```suggestion
    else if [[ -d "/var/lib/alternatives" ]]; then
```

otherwise you get a weird error if the directory didn't exist to begin with (because it tries to `rmdir` a directory that doesn't exist). 


```
rmdir: failed to remove '/var/lib/alternatives': No such file or directory
Warning: /var/lib/alternatives is not empty
Successfully removed /var/lib/alternatives
Setting up iptables-nft alternatives succeeded
```

dustymabe:
I fail to see how the `ext.config.firewall.iptables-legacy` test passes with the introduction of this code. Most likely a race condition?

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
travier: Indeed. I think it's fine. The goal was to have this in a barrier and then remove it, so this would not run that often.

tests/kola/upgrade/extended/test.sh
travier: We specify `--admindir` when we want to bypass any "local state" on the node (i.e. whether `/var/lib/alternatives` exists or not) and still do operations (such as symlink changes) using the up to date config.

When we don't specify it, we want to make sure that the migration has been successfully completed and that the new config folder is used automatically.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
travier: OK but it's very unlikely that this would happen as we condition the unit on this path existing and it's either a symlink or a dir. If it's something else then it's unexpected.

travier:
> I fail to see how the `ext.config.firewall.iptables-legacy` test passes with the introduction of this code. Most likely a race condition?

Indeed this is weird. We should remove this test as part of this change.

travier:
> considering this comment from the description of [coreos/fedora-coreos-tracker#1818](https://github.com/coreos/fedora-coreos-tracker/issues/1818)
> 
> > As this could potentially be a breaking change, we might want to do it only starting with Fedora 42.
> 
> Should we only roll this in for Fedora 43?

I think I was thinking more about the `-legacy` removal as a breaking change but the more it goes the less likely I think it matters. Kubernetes is moving to nft now and the rest of Fedora has removed the legacy one a long time ago so it's becoming less and less likely to be an issue.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
dustymabe: So the later code in this script is OK to run multiple times then?

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
dustymabe: ahh. ok. I was manually testing this out on a system (i.e. running the script without the unit) and that's why I encoutered this problem.

we can leave it as is

dustymabe:
> > I fail to see how the `ext.config.firewall.iptables-legacy` test passes with the introduction of this code. Most likely a race condition?
> 
> Indeed this is weird. We should remove this test as part of this change.

Ok, now I understand how it passes. Newly deployed systems don't have `/var/lib/alternatives` at all:


```
Fedora CoreOS 42.20250705.dev.0
Discuss: https://discussion.fedoraproject.org/tag/coreos

[core@cosa-devsh ~]$ 
[core@cosa-devsh ~]$ 
[core@cosa-devsh ~]$ ls -l /var/lib/alternatives
ls: cannot access '/var/lib/alternatives': No such file or directory
```

so the migration service (and thus script) won't run. So the symlinks manually placed there by Ignition won't be updated. So those people won't be migrated with this script, which may be what we want (i.e. if they explicitly chose to use legacy iptables maybe we don't stop them here?)

dustymabe:
> > considering this comment from the description of [coreos/fedora-coreos-tracker#1818](https://github.com/coreos/fedora-coreos-tracker/issues/1818)
> > > As this could potentially be a breaking change, we might want to do it only starting with Fedora 42.
> > 
> > 
> > Should we only roll this in for Fedora 43?
> 
> I think I was thinking more about the `-legacy` removal as a breaking change but the more it goes the less likely I think it matters. Kubernetes is moving to nft now and the rest of Fedora has removed the legacy one a long time ago so it's becoming less and less likely to be an issue.

I think the problem for me here is that maybe we think it's not a big deal, but if we're wrong we're probably breaking something pretty fundamental. I'd almost prefer we do this migration and also remove the iptables-legacy packages in the F43 rebase. i.e. just do them both at the same time? Or maybe that isn't possible? 

Doing it on the F43 boundary means we can at least get a lot of people doing testing on it before hand before we roll it out to the more stable streams. We can also include it in our regular communications about the major rebase.

overlay.d/50alternatives/usr/libexec/coreos-alternatives-migration
travier: Yes, it will set the iptables to the nftable alternative on every boot, which is useless but harmless. Note that this is very unlikely to be triggered as it requires admins to have manually added alteratives configs (not just symlinks) to nodes.

travier:
I'm OK with delaying to the F43 rebase. I find very unlikely that someone would have used the legacy backend intentionally. I think we should have dropped the documentation for reverting to the legacy backend when we updated to the nft one (we kept it as an example for how to configure alternatives).

travier:
If we delay to F43 then we should merge https://github.com/coreos/fedora-coreos-config/pull/3573 in this one as well and remove the legacy test.

dustymabe:
> If we delay to F43 then we should merge #3573 in this one as well and remove the legacy test.

This sounds good to me! @PeaceRebel are you comfortable doing that?

PeaceRebel:
> This sounds good to me! @PeaceRebel are you comfortable doing that?

Yeah, I have made the changes accordingly. What do we do with the extended upgrade test?

overlay.d/50alternatives/usr/lib/systemd/system-preset/50-alternatives-migration.preset
```diff
@@ -0,0 +1,2 @@
+# Enable alternatives migration script
```
travier: Let's name this file `50-coreos-alternatives-migration.preset` to be explicit

travier (commented):
Can you add the changes from https://github.com/coreos/fedora-coreos-config/pull/3573 here as well? Thanks

PeaceRebel:
> Can you add the changes from #3573 here as well? Thanks

Added it here and closed the other PR.

dustymabe:
Let's squash the last two commits together.

Also there is some [post script](https://github.com/coreos/fedora-coreos-config/blob/e44375b071707719c7b4e1e4ff035f454f59f912/manifests/fedora-coreos-base.yaml#L74-L86) we should clean up. Let's just leave the changes all in fedora-coreos-base and add a new conditional_include in there? 


Suggestion:

```diff
diff --git a/manifests/fedora-coreos-base.yaml b/manifests/fedora-coreos-base.yaml
index 05b2927a..98e7b24b 100644
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -27,6 +27,28 @@ ostree-layers:
   - overlay/25azure-udev-rules
   - overlay/30lvmdevices
 
+conditional-include:
+  - if: releasever < 43
+    include:
+      packages:
+        # iptables-legacy was in <43 but excluded from 43+
+        # https://github.com/coreos/fedora-coreos-tracker/issues/1818
+        - iptables-legacy
+      postprocess:
+        # Default to iptables-nft. Otherwise, legacy wins. We can drop this once/if we
+        # remove iptables-legacy. This is needed because alternatives don't work
+        # https://github.com/coreos/fedora-coreos-tracker/issues/677
+        # https://github.com/coreos/fedora-coreos-tracker/issues/676
+        - |
+          #!/usr/bin/bash
+          set -eux -o pipefail
+          ln -sf /usr/sbin/ip6tables-nft         /etc/alternatives/ip6tables
+          ln -sf /usr/sbin/ip6tables-nft-restore /etc/alternatives/ip6tables-restore
+          ln -sf /usr/sbin/ip6tables-nft-save    /etc/alternatives/ip6tables-save
+          ln -sf /usr/sbin/iptables-nft          /etc/alternatives/iptables
+          ln -sf /usr/sbin/iptables-nft-restore  /etc/alternatives/iptables-restore
+          ln -sf /usr/sbin/iptables-nft-save     /etc/alternatives/iptables-save
+
 # Be minimal
 recommends: false
 
@@ -71,20 +93,6 @@ postprocess:
     set -eux -o pipefail
     systemctl mask dnsmasq.service
 
-  # Default to iptables-nft. Otherwise, legacy wins. We can drop this once/if we
-  # remove iptables-legacy. This is needed because alternatives don't work
-  # https://github.com/coreos/fedora-coreos-tracker/issues/677
-  # https://github.com/coreos/fedora-coreos-tracker/issues/676
-  - |
-    #!/usr/bin/bash
-    set -eux -o pipefail
-    ln -sf /usr/sbin/ip6tables-nft         /etc/alternatives/ip6tables
-    ln -sf /usr/sbin/ip6tables-nft-restore /etc/alternatives/ip6tables-restore
-    ln -sf /usr/sbin/ip6tables-nft-save    /etc/alternatives/ip6tables-save
-    ln -sf /usr/sbin/iptables-nft          /etc/alternatives/iptables
-    ln -sf /usr/sbin/iptables-nft-restore  /etc/alternatives/iptables-restore
-    ln -sf /usr/sbin/iptables-nft-save     /etc/alternatives/iptables-save
-
   # sudo prefers its config files to be mode 440, and some security scanners
   # complain if /etc/sudoers.d files are world-readable.
   # https://bugzilla.redhat.com/show_bug.cgi?id=1981979
@@ -164,9 +172,6 @@ packages:
   - console-login-helper-messages-motdgen
   # i18n
   - kbd
-  # In F35+ need `iptables-legacy` package
-  # See https://github.com/coreos/fedora-coreos-tracker/issues/676#issuecomment-928028451
-  - iptables-legacy
   # NIC firmware we've traditionally shipped but then were split out of linux-firmware in Fedora
   - qed-firmware # https://github.com/coreos/fedora-coreos-tracker/issues/1746
   # Include udev rules for NVMe backed Azure Instances
```

PeaceRebel:
> Let's just leave the changes all in fedora-coreos-base and add a new conditional_include in there?

Including the alternates overlay changes too? @dustymabe

dustymabe:
> > Let's just leave the changes all in fedora-coreos-base and add a new conditional_include in there?
> 
> Including the alternates overlay changes too? @dustymabe

No sorry.. Just the last commit where we are moving the iptables-legacy package (and also the postprocess for setting nft as the default, which won't be needed once we remove the legacy package).

PeaceRebel:
> No sorry.. Just the last commit where we are moving the iptables-legacy package (and also the postprocess for setting nft as the default, which won't be needed once we remove the legacy package).

Ok. Updated.

dustymabe:
This LGTM, but now the alternatives test (that we are also adding in this PR) can't do the full test because there is no iptables-legacy to set it back to so we can test the migration. 

Not sure what to do about that.

Maybe we'll just have to settle for some manual testing?

PeaceRebel:
> This LGTM, but now the alternatives test (that we are also adding in this PR) can't do the full test because there is no iptables-legacy to set it back to so we can test the migration.
> 
> Not sure what to do about that.
> 
> Maybe we'll just have to settle for some manual testing?

We can probably do the same verification as we do in the extended upgrade test. I guess the upgrade test needs to be updated as well to do the verification if release version is >= 43

travier:
> This LGTM, but now the alternatives test (that we are also adding in this PR) can't do the full test because there is no iptables-legacy to set it back to so we can test the migration.
> 
> Not sure what to do about that.
> 
> Maybe we'll just have to settle for some manual testing?

We can overlay the package and apply live in the test as well.

PeaceRebel:
Modified the alternatives test to skip on version before 43 and overlay iptables-legacy.

tests/kola/files/alternatives
```diff
@@ -0,0 +1,89 @@
...
+# See
+# - https://github.com/coreos/fedora-coreos-tracker/issues/1818
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+# This test is only valid in version 43 or later
+if [ $(. /etc/os-release && echo "$VERSION_ID") -le 43 ]; then
```
travier: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/data/commonlib.sh#L53

tests/kola/upgrade/extended/test.sh
```diff
@@ -244,6 +244,23 @@ selinux-sanity-check() {
 
 ok "Reached version: $version"
 
+verify-alternatives-migration() {
```
travier, line 247: Let's check the Fedora versions here before we run this check: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/data/commonlib.sh#L53

PeaceRebel: Done

travier (commented):
OK, this should be good for F43 now. Thanks Bipin!

travier:
Requesting an (hopefully) final review from @dustymabe :)

PeaceRebel:
~~@dustymabe, dropping a reminder.~~
I see the test is failing. I'll take a look tomorrow.

PeaceRebel:
> ~@dustymabe, dropping a reminder.~ I see the test is failing. I'll take a look tomorrow.

Fixed it. @dustymabe , can you take a final look?

dustymabe (approved):
LGTM

Issue #3254, opened 2024-11-14
Migrate ci/overrides.py to dnf5


jlebon:
It's still using the old dnf4 APIs. We'll need to do this if we want to eventually be able to stop having to pull it in dnf4.

dustymabe:
fixed in https://github.com/coreos/fedora-coreos-config/pull/3428

PR #3255, opened 2024-11-14
workflows/add-overrides: install python3-dnf

jlebon:
F41 moved on to dnf5, which doesn't pull in Python bindings anymore. But anyway, the Python API in libdnf5 is different and we'd need to migrate `overrides.py`.

For now at least to unbreak things, just re-install back python3-dnf.

See also: https://github.com/coreos/fedora-coreos-config/issues/3254


prestist (approved):
lgtm, thank you for fixing this!

PR #3265, opened 2024-11-18
manifests/fedora-coreos: move from `rojig` to `metadata`

jlebon:
Now that cosa supports looking at the `metadata` key, let's finally stop abusing the `rojig` key.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3948

dustymabe (approved):
LGTM

PR #3267, opened 2024-11-19
[rawhide] overrides: pin clevis-21-6.fc42

coreosbot-releng:


dustymabe (approved):
LGTM

PR #3269, opened 2024-11-21
[rawhide] overrides: pin kernel-6.12.0-0.rc7.59.fc42

coreosbot-releng:


dustymabe:
There was a final (non-rc) kernel that was built but we never picked up in a `rawhide` FCOS build. Let's try to see if that one passes the test (I'd like to get some testing with the final 6.12 kernel).

https://github.com/coreos/fedora-coreos-config/pull/3270

If that one fails tests we can merge this one.

PR #3282, opened 2024-11-26
denylist: add multipath.partition

c4rt0:
multipath.partition fails after the recent kernel upgrade.

For more details ...

See: https://github.com/coreos/fedora-coreos-tracker/issues/1840

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -6,3 +6,7 @@
   warn: true
   arches:
     - ppc64le
+- pattern: multipath.partition
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1840
+  streams:
+     - rawhide


c4rt0:
Sorry for the noise with so many forced pushes. This should be it now.

kola-denylist.yaml
```diff
@@ -6,3 +6,5 @@
   warn: true
   arches:
     - ppc64le
+- pattern: multipath.partition
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1840
```
dustymabe, line 10: let's limit to `rawhide` for now since that is the only place we are seeing it.

c4rt0: done

dustymabe (approved):
LGTM

PR #3285, opened 2024-11-29
[Rawhide] overrides: Remove `kdump-utils` pin

gursewak1997:
The issue is fixed now. Let's remove the pin from Rawhide.


jlebon (approved):
CI should be fixed now.

gursewak1997:
Ci should be fixed by https://github.com/coreos/fedora-coreos-config/pull/3286

PR #3288, opened 2024-12-02
[rawhide] overrides: Remove `kdump-utils` pin in rawhide

gursewak1997:
Issue is fixed now. Let's drop the pin.


dustymabe (approved):
LGTM

PR #3290, opened 2024-12-02
overrides: pin qemu-user-static-x86

nikita-dubrovskii:
qemu `9.2.0` fails with `segfault`. Pin it until https://github.com/coreos/fedora-coreos-tracker/issues/1846 gets fixed

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -29,3 +29,8 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1836
       type: pin
+  qemu-user-static-x86:
+    evr: 2:9.1.1-2.fc42
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1846
+      type: pin


jbtrystram (approved):
lgtm

manifest-lock.overrides.yaml
```diff
@@ -29,3 +29,8 @@ packages:
     metadata:
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1836
       type: pin
+  qemu-user-static-x86:
+    evr: 2:9.1.1-2.fc42
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1846
+      type: pin
```
dustymabe, line 36: do all architectures include this RPM? should it be an architecture specific override? Maybe it doesn't matter.

nikita-dubrovskii: Just checked: all, but x86.

dustymabe: then I guess if CI passes on this PR we should be good :)

PR #3292, opened 2024-12-03
[rawhide] overrides: remove clevis pin

aaradhak:
The kola test ext.config.var-mount.luks seems to PASS when tested with the latest clevis pkg update `clevis-21-8.fc42.x86_64`. By unpinning this clevis pkg, the latest clevis pkg could be fetched and tested.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1836


dustymabe (approved):
LGTM

PR #3293, opened 2024-12-03
image-base.yaml: compress applehv with gzip, hyperv with zip

jlebon:
These platforms are already compressed using those compressors today, but we're moving that policy out of cosa and into here since it's not a requirement from the platform itself and more an opinion (which others may want to differ from).

This uses a new `platform-compressor` key which `cosa compress` will learn to read.


dustymabe (approved):
LGTM

travier (approved):

PR #3294, opened 2024-12-03
tests: add/use common definition for container repo/tag to use

dustymabe:
This will make it easier to update our tests to use newer Fedora
containers when we are ready.


gursewak1997 (approved):
lgtm

PR #3297, opened 2024-12-04
tests: fix boot.bootupd-validate on rhel-9.6

jlebon:
In the el-only variants, there is no `RHEL_VERSION` since the RHEL version _is_ `VERSION_ID`.

Add a new `get_rhel_ver` commonlib function to abstract this away and use that in the test.

Don't bother checking for the `VARIANT_ID`. We only care about the coreos variant in the tests in this repo. (And once we start moving tests to fedora-bootc, we would want to remove these checks anyway.)

### DIFF
--- a/tests/kola/boot/bootupd-validate
+++ b/tests/kola/boot/bootupd-validate
@@ -84,7 +84,7 @@ validate() {
 # should remove this until https://issues.redhat.com/browse/RHEL-66584 is fixed
 set_selinux_permissve() {
     source /etc/os-release
-    if is_scos || ([ "${RHEL_VERSION}" == "9.6" ] && [ "${VARIANT_ID}" == "coreos" ] ); then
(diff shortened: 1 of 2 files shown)


tests/kola/boot/bootupd-validate
```diff
@@ -84,7 +84,7 @@ validate() {
 # should remove this until https://issues.redhat.com/browse/RHEL-66584 is fixed
 set_selinux_permissve() {
     source /etc/os-release
-    if is_scos || ([ "${RHEL_VERSION}" == "9.6" ] && [ "${VARIANT_ID}" == "coreos" ] ); then
+    if is_scos || [ "$(get_rhel_ver)" = "9.6" ]; then
```
dustymabe, line 87:
```suggestion
    if is_scos || [ "$(get_rhel_ver)" == "9.6" ]; then
```

I tend to prefer the `==` variant.

dustymabe (approved):

PR #3315, opened 2025-01-07
[rhcos-4.15] tests/kola: handle flakes in quadlet test

aaradhak:
This is a backport of https://github.com/coreos/fedora-coreos-config/commit/24c939e7a7c069dd20e8d72a2a2541df079f16ce

This test was flaking on ppc64le a lot. After debugging we were able to see that the `podman inspect systemd-test` was running before the container was actually started, but the call didn't fail because `systemd-test` matched the name of the volume and the network resource. So we need to do a few things here:

- make sure we don't try to check things before the volume and network has been created or the container has been started.
- limit the inspect call for the container to only containers by doing `podman container inspect` instead of generic `podman inspect`.


dustymabe (approved):
/lgtm

PR #3316, opened 2025-01-13
[rawhide] overrides: fast-track systemd-257.2-4.fc42

coreosbot-releng:

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -10,32 +10,38 @@
 
 packages:
   systemd:
-    evr: 257-1.fc42
+    evr: 257.2-4.fc42
     metadata:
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

dustymabe:
testiso failure in CI: 

```
Unexpected string from completion channel: coreos-installer-test-entered-emergency-target expected: iscsi-boot-ok
```

might want to see if you can reproduce locally

aaradhak:
I will check kola testiso test and see if it is consistent locally.

aaradhak:
Looks like this test failure is not consistent. I am not seeing the same issue when tested locally. I have replayed the CI job to see if the same error occurs again

PR #3325, opened 2025-01-20
[rawhide] overrides: Drop systemd-257-1.fc42 pin

travier:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1857
See: https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin


dustymabe:
Obsoleted by https://github.com/coreos/fedora-coreos-config/pull/3318

PR #3326, opened 2025-01-20
tests/kola: Add lockdown LSM test

travier:
See: https://bugzilla.redhat.com/show_bug.cgi?id=2333706

---

Will fail until https://bugzilla.redhat.com/show_bug.cgi?id=2333706 is fixed.

### DIFF
--- a/tests/kola/security/lockdown
+++ b/tests/kola/security/lockdown
@@ -1,0 +1,24 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   tags: secure-boot
+##   architectures: x86_64
+##   description: Verify that the lockdown LSM is set to integrity when booted using Secure Boot
+#
+# See https://bugzilla.redhat.com/show_bug.cgi?id=2333706
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+lockdown_state="$(cat "/sys/kernel/security/lockdown")"
+
+if [[ "$(mokutil --sb)" == "SecureBoot enabled" ]]; then
+    if [[ "${lockdown_state}" == "none [integrity] confidentiality" ]]; then
+        ok "lockdown LSM set to integrity on a Secure Boot system"
+    else
+        fatal "lockdown LSM not set to integrity on a Secure Boot system"
+    fi
+else
+    fatal "system is not running with Secure Boot enabled"
+fi
--- a/tests/kola/security/lockdown_no_sb
+++ b/tests/kola/security/lockdown_no_sb
@@ -1,0 +1,23 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   architectures: x86_64 aarch64
+##   description: Verify that the lockdown LSM is set to none when booted without Secure Boot enabled
+#
+# See https://bugzilla.redhat.com/show_bug.cgi?id=2333706
+
+set -xeuo pipefail
+
+. $KOLA_EXT_DATA/commonlib.sh
+
+lockdown_state="$(cat "/sys/kernel/security/lockdown")"
+
+if [[ "$(mokutil --sb)" != "SecureBoot enabled" ]]; then
+    if [[ "${lockdown_state}" == "[none] integrity confidentiality" ]]; then
+        ok "lockdown LSM set to none on a non Secure Boot system"
+    else
+        fatal "lockdown LSM not set to none on a non Secure Boot system"
+    fi
+else
+    fatal "system is running with Secure Boot enabled"
+fi
--- a/tests/kola/security/data/commonlib.sh
+++ b/tests/kola/security/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh


tests/kola/security/lockdown
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   description: Verify that the lockdown LSM is set to integrity.
```
jlebon: This needs some special tagging so it only runs on Secure Boot, but I don't think we have something yet, so it'll likely need some kola work.

And also `architectures: x86_64`.

travier:
Will rebase / update once https://bodhi.fedoraproject.org/updates/FEDORA-2025-cca2fcc70c lands in FCOS.

travier:
Updated to account for both Secure Boot and non-SB runs. This should work on all architectures but running it only on x86_64 and aarch64 makes sense.

> This needs some special tagging so it only runs on Secure Boot, but I don't think we have something yet, so it'll likely need some kola work.

Hum, do you mean that we should add a special kola tag so that it runs this test in a SB enabled VM? We indeed don't run the general kola tests with SB enabled AFAIK.

travier:
Created https://github.com/coreos/coreos-assembler/issues/4112 to track the work in kola

tests/kola/security/lockdown_no_sb
```diff
@@ -0,0 +1,23 @@
...
+lockdown_state="$(cat "/sys/kernel/security/lockdown")"
+
+if [[ "$(mokutil --sb)" != "SecureBoot enabled" ]]; then
+    if [[ "${lockdown_state}" == "[none] integrity confidentiality" ]]; then
+        ok "lockdown LSM set to none on a non Secure Boot system"
+    else
+        fatal "lockdown LSM not set to none on a non Secure Boot system"
+    fi
+else
+  fatal "system is running with secure boot enabled"
```
travier:
```suggestion
    fatal "system is running with secure boot enabled"
```

tests/kola/security/lockdown_no_sb
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   architectures: x86_64 aarch64
+##   description: Verify that the lockdown LSM is set to integrity when booted using Secure Boot
```
travier:
```suggestion
##   description: Verify that the lockdown LSM is set to none when booted without Secure Boot enabled
```

tests/kola/security/lockdown
```diff
@@ -0,0 +1,24 @@
...
+lockdown_state="$(cat "/sys/kernel/security/lockdown")"
+
+if [[ "$(mokutil --sb)" == "SecureBoot enabled" ]]; then
+    if [[ "${lockdown_state}" == "none [integrity] confidentiality" ]]; then
+        ok "lockdown LSM set to integrity on a Secure Boot system"
+    else
+        fatal "lockdown LSM not set to integrity on a Secure Boot system"
+    fi
+else
+    fatal "system is not running with secure boot enabled"
```
travier:
```suggestion
    fatal "system is not running with Secure Boot enabled"
```

tests/kola/security/lockdown_no_sb
```diff
@@ -0,0 +1,23 @@
...
+lockdown_state="$(cat "/sys/kernel/security/lockdown")"
+
+if [[ "$(mokutil --sb)" != "SecureBoot enabled" ]]; then
+    if [[ "${lockdown_state}" == "[none] integrity confidentiality" ]]; then
+        ok "lockdown LSM set to none on a non Secure Boot system"
+    else
+        fatal "lockdown LSM not set to none on a non Secure Boot system"
+    fi
+else
+    fatal "system is running with secure boot enabled"
```
travier:
```suggestion
    fatal "system is running with Secure Boot enabled"
```

travier (commented):
Small nits but LGTM. You can wait to push the fixes as the CI will fail until we get the PR in COSA.

PeaceRebel:
Merge after https://github.com/coreos/coreos-assembler/pull/4114

tests/kola/security/lockdown
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   tags: needs-secureboot
+##   architectures: x86_64 aarch64
```
dustymabe: does aarch64 in Fedora support secureboot now? I know for a long time releng didn't have that set up.

tests/kola/security/lockdown
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
dustymabe, line 1: we'll need to denylist this test for all centos stuff in the rhel-coreos-config.

we could add a distros: entry here but I prefer to keep all the "overrides" for the secureboot in centos problem in the same place.

tests/kola/security/lockdown
PeaceRebel: I haven't tried it. @travier any idea?

tests/kola/security/lockdown
PeaceRebel: Sure, I'll take a look at this.

PeaceRebel: @dustymabe Here's the PR for this. https://github.com/coreos/rhel-coreos-config/pull/15

tests/kola/security/lockdown
HuijingHei: Perhaps not yet, see https://issues.redhat.com/browse/RHEL-1927

PeaceRebel: @HuijingHei I don't seem to have the permissions to see that issue.

PeaceRebel: I'll remove `aarch64`. We can add it back if required.

tests/kola/security/lockdown_no_sb
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   architectures: x86_64 aarch64
+##   description: Verify that the lockdown LSM is set to none when booted without Secure Boot enabled
```
HuijingHei, line 5: IMU, we do not test this using uefi firmware right? Does this mean we will add another tag with secure disabled? Or maybe add kola json supported key like `firmware: uefi-secure/uefi`? (see [kolajson](https://github.com/coreos/coreos-assembler/blob/main/docs/kola/external-tests.md#kolajson))

PeaceRebel: @HuijingHei  By default most tests run without secure boot. This test will be for all cases where secure boot is not supported or is disabled.  I don't think another tag is required.

HuijingHei: Hmm, but we start VM using BIOS (not UEFI) by default, and secure boot is for UEFI.

PeaceRebel: I think we'll need one more similar test to cover those two cases. But I'm not sure if we need to test these negative cases. We need to validate that the lockdown LSM is set to integrity if secure boot is supported and enabled, and this is done in the other test from this PR.

travier: The Secure Boot disabled & BIOS cases are indeed conflated in this one because there are no differences. I agree that technically there should be another case with UEFI & Secure Boot disabled, but I don't think it's worth it. One could argue that this BIOS test is not useful either as it only tests for a weak setting thus it has no security impact.

HuijingHei: SGTM.

jbtrystram (approved):
LGTM

PR #3327, opened 2025-01-20
tests/kola/files/setuid: Ignore both bin & sbin paths

travier:
See: https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin


dustymabe:
LGTM - Optional to add a comment above the `sbin` ones to mention we can drop those when RHEL has done the bin/sbin merge.

dustymabe (dismissed):

travier:
Add notes about dropping the sbin entries

dustymabe (approved):
LGTM

PR #3328, opened 2025-01-21
tests/kola: add new test to verify composefs is on

nikita-dubrovskii:
We have `composefs` turned on in FCOS. We should have a test that verifies it.
```
$ cosa kola run ext.config.composefs.enabled
=== RUN   ext.config.composefs.enabled
--- PASS: ext.config.composefs.enabled (25.06s)
```

### DIFF
--- a/tests/kola/composefs/enabled
+++ b/tests/kola/composefs/enabled
@@ -1,0 +1,29 @@
+#!/bin/bash
+## kola:
+##   # We now have composefs turned on in CoreOS.
+##   description: Verify that composefs is on.
+
+set -euo pipefail
+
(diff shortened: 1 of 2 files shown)


jbtrystram (approved):
LGTM

dustymabe (approved):
LGTM

tests/kola/composefs/enabled
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+## kola:
+##   # We now have composefs turned on in CoreOS.
+##   description: Verify that composefs is on.
```
jlebon, line 4: This needs `exclusive: false`, no?

PR #3333, opened 2025-01-23
[testing] deploy container images by default

jbtrystram:
Switch boot images to use OCI for updates. This is a step towards bootable containers support and bootc rebase.

See https://fedoraproject.org/wiki/Changes/CoreOSOstree2OCIUpdates 
Requires https://github.com/coreos/zincati/pull/1241 
See https://github.com/coreos/fedora-coreos-tracker/issues/1823

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -2,3 +2,8 @@
 # similarly to manifest.yaml. Unlike image-base.yaml, which is shared by all
 # streams.
 include: image-base.yaml
+
+# Move testing to use OCI images by default
+# https://github.com/coreos/fedora-coreos-tracker/issues/1823
+deploy-via-container: true
+container-imgref: "ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos:testing"


dustymabe:
dumb questions.. is this PR just for testing? why open it against the `testing` branch?

jbtrystram:
> dumb questions.. is this PR just for testing? why open it against the `testing` branch?

because the container imgref points to the `testing` tag, and each stream have it's own tag, so I don't want this to flow into stable through promotion.

But i guess i could do it in testing-devel first yeah

travier:
I think this should be for `next-devel` first

dustymabe:
> I think this should be for `next-devel` first

yes. it would start with `branched` and `next-devel`

jbtrystram:
> > I think this should be for `next-devel` first
> 
> yes. it would start with `branched` and `next-devel`

PR for branched : https://github.com/coreos/fedora-coreos-config/pull/3363
PR for next-devel : https://github.com/coreos/fedora-coreos-config/pull/3364

PR #3334, opened 2025-01-23
tests/kola: check only for rhcos based on 9.4 in composefs.enabled test

nikita-dubrovskii:
Only 9.4 doesn't support composefs.
There are other issues on ppc64le and kernel-64k in el9, so this test is denylisted in openshift/os repo.


dustymabe (approved):
LGTM

PR #3335, opened 2025-01-27
[rawhide] overrides: pin rsync-3.4.0-1.fc42

coreosbot-releng:


dustymabe (approved):
LGTM

dustymabe:
There is a new rawhide failure that I'm chasing down now in https://github.com/coreos/rpm-ostree/issues/5261

As for this I'm just going to go ahead and merge it since it clearly allows the build to complete now and we're just failing tests.

PR #3336, opened 2025-01-28
rework kernel-replace test; denylist it on rawhide

dustymabe:
The kernel replace test was using a fairly old kernel (hardcoded) so I decided to update it to more dynamically choose a kernel to use and also took the opportunity to rework the rest of the test to modernize it a bit.

I also denylist the test here on rawhide because of https://github.com/coreos/fedora-coreos-tracker/issues/1870

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -6,3 +6,9 @@
   warn: true
   arches:
     - ppc64le
+- pattern: ext.config.rpm-ostree.kernel-replace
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

There is a bug in rpm-ostree with calling rpm-ostree override-replace on the new path. However if you call `dnf downgrade -y kernel*.rpm` it should work.

PR #3338, opened 2025-01-28
tests/kola: disable yum repos for kernel-replace test

dustymabe:
Even though all the rpms that we are replacing are local rpm-ostree
will still reach out to the repos during the operation. Let's just
disable all the repos to prevent this from happening.

This was causing issues in our RHCOS pipeline because we have more
strict egress rules there and it was trying to hit the CentOS mirrors
when testing on c9s since the yum repos are there in c9s by default.

xref: https://github.com/openshift/os/issues/1673

----

also a second commit in there to fix a copy/paste error from 6f86533

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -102,6 +102,10 @@ build_derived_ociarchive() {
 
     cat > Containerfile << EOF
 FROM $imagespec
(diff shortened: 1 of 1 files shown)


marmijo (approved):
LGTM

aaradhak (approved):

c4rt0 (commented):
It took me a while to wrap my head around this.

/lgtm

PR #3341, opened 2025-01-29
tests/kola: name all el9 kernel subpackages in kernel-replace

dustymabe:
Depending on if we are overriding a newer or older kernel `dnf download` may download all of them or just a subset. Let's just force it to download them all so we don't get differing behavior.


c4rt0 (approved):
Ok, now that makes sense! Thanks @dustymabe !

/lgtm

PR #3342, opened 2025-01-30
live-iso: add erofs support

nikita-dubrovskii:


overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -55,22 +55,31 @@ EOF
 
 isoroot=$(getarg coreos.liveiso= ||:)
 
+# By default we use squashfs
+rootfs_fstype=squashfs
+rootfs_path=/root.squashfs
```
dustymabe: root.squashfs and root.erofs would never both exist on the same media so I think it would be easier in the future to remove the squashfs path if we structured it such that root.erofs was the default here.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -58,19 +58,20 @@ isoroot=$(getarg coreos.liveiso= ||:)
...
-# Verifies that we have the right root.squashfs, or downloads it if needed
+# Verifies that we have the right ${rootfs_path}, or downloads it if needed
 After=coreos-livepxe-rootfs.service
 Before=initrd-root-fs.target
 
 [Mount]
-What=/root.squashfs
+What=${rootfs_path}
 Where=/sysroot
-Type=squashfs
```
dustymabe, line 73: Not sure if there are any consequences for removing `Type=` but it sounds good to me if there aren't any. I'm sure the kernel just autodetects here.

dustymabe:
the code LGTM - I'll test it out today.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
nikita-dubrovskii: Yes, there is no need for that

nikita-dubrovskii:
PXE installation doesn't work: 
```
[    1.631722] ls:                                                                                                                                             
[    1.631725] cannot access '/root.*fs'  
```

Fixing ...

nikita-dubrovskii:
Now PXE works, here are logs

`root.squashfs`

```
[  OK  ] Mounted run-ephemeral.mount - /run/ephemeral.                                                                                                         
[    4.833554] coreos-livepxe-rootfs[806]: Creating symlink to /root.squashfs                                                                                  
[  OK  ] Finished coreos-livepxe-rootfs.ser…ce - Acquire Live PXE rootfs Image.                                                                                
[    4.835793] systemd[1]: Mounting sysroot.mount - /sysroot...                                                                                                
[  OK  ] Mounted sysroot.mount - /sysroot.                                                                                                                     
```

`root.erofs`:

```
[  OK  ] Mounted run-ephemeral.mount - /run/ephemeral.                                                                                                         
[    9.394641] coreos-livepxe-rootfs[780]: Creating symlink to /root.erofs                                                                                     
[  OK  ] Finished coreos-livepxe-rootfs.ser…ce - Acquire Live PXE rootfs Image.
[    9.398447] systemd[1]: Mounting sysroot.mount - /sysroot...                                                                                                                                    
[    9.480587] erofs: (device loop1): mounted with root inode @ nid 208.                                                                                       
[  OK  ] Mounted sysroot.mount - /sysroot.                                                                                                                     
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -58,19 +58,19 @@ isoroot=$(getarg coreos.liveiso= ||:)
 if [ -z "${isoroot}" ]; then
     # In this case, the rootfs is already unpacked into the initrd, or we need
     # to retrieve it
+
```
dustymabe: delete added newline?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
```diff
@@ -79,6 +79,11 @@ elif [[ -n "${rootfs_url}" ]]; then
         fi
         exit 1
     fi
+
+    # we support both squashfs and erofs, let's create the symlink for sysroot.mount
+    sysroot=$(ls /root.*fs)
+    echo "Creating symlink to ${sysroot}"
+    ln -s "${sysroot}" /sysroot.img
```
dustymabe: This needs to happen in both the `if` and `elif` cases here.

I'd say let's just move it to below the `else` at the bottom of the file.

This actually exposes a gap in our CI. We have an option on `kola testiso` to test this but we don't exercise it anywhere:

```
[dustymabe@media fcos]$ cosa kola testiso pxe* --pxe-append-rootfs 
kola -p qemu testiso pxe* --pxe-append-rootfs --output-dir tmp/kola
Detected development build; disabling signature verification
Running test: pxe-offline-install.bios
FAIL: pxe-offline-install.bios (57.89s)
    entered emergency.target in initramfs
Running test: pxe-offline-install.4k.uefi
FAIL: pxe-offline-install.4k.uefi (1m40.222s)
    entered emergency.target in initramfs
Running test: pxe-online-install.bios
FAIL: pxe-online-install.bios (58.15s)
    entered emergency.target in initramfs
Running test: pxe-online-install.4k.uefi
FAIL: pxe-online-install.4k.uefi (1m39.835s)
    entered emergency.target in initramfs
2025-02-10T21:24:05Z cli: harness: test suite failed
```

dustymabe: I'm not the biggest fan of `sysroot.img` here because it's a squash or erofs and those don't typically have that suffix.. maybe we modify it slightly:

In the other file we use the var `sysrootfsimg` to refer to the `rootfs.img` (the file that contains `root.squashfs` embedded in it) so maybe we just use `sysrootfs` here with no suffix:

```bash
    sysrootfs=$(ls /root.*fs)
    echo "Creating symlink '/sysrootfs -> /${sysrootfs}'"
    ln -s "${sysrootfs}" /sysrootfs

File: overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator [hunk 5]
```diff
@@ -58,19 +58,19 @@ isoroot=$(getarg coreos.liveiso= ||:)
...
     cat >"${UNIT_DIR}/sysroot.mount" <<EOF
 # Automatically generated by live-generator
 
 [Unit]
 DefaultDependencies=false
-# Verifies that we have the right root.squashfs, or downloads it if needed
+# Verifies that we have the right sysroot.img, or downloads it if needed
```
dustymabe, line 66: as mentioned in the other file I think `sysrootfs` would be more appropriate here:

```suggestion
# Verifies that we have the right sysrootfs, or downloads it if needed
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
jlebon: Fine either way, but I personally consider `.img` idiomatic for image-like artifacts like the ones we're dealing with here, so `sysroot.img` seems more appropriate and informative than `sysrootfs` to me.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
jlebon: I don't think it's the kernel doing the auto-detection here, but util-linux. I think in general it's better to specify the type you expect to not rely on it getting it right and just be more strict. One way that I think should work here is to make this an env var and then add an `EnvironmentFile` that we populate from `coreos-livepxe-rootfs.sh`.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
nikita-dubrovskii: What's about `sysrootfs.img` ?

nikita-dubrovskii: Is it by design, that `-m` and `--qemu-memory` kola options are not supported by `testiso` ? For `cosa kola testiso pxe* --pxe-append-rootfs`  had to hardcode more memory  ( `builder.MemoryMiB = 8192`).

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
dustymabe: While we can write out an override from `coreos-livepxe-rootfs.sh` for this particular mount where we can set the `Type` that won't be so easy for the `sysroot.mount` from ~line 150 because we don't know whether the file inside the `rootfs.img` is erofs or squashfs at that point in time.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
nikita-dubrovskii: fixed that in https://github.com/coreos/coreos-assembler/pull/4012/commits/6328c761ca10315a4d4f04e11b74da998dd699d1

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
dustymabe: For setting the What= and Type= here this seems to work:

```diff
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
index 61b4661f..d62443ab 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
@@ -88,6 +88,20 @@ else
 fi
 
 # we support both squashfs and erofs, let's create the symlink for sysroot.mount
-sysroot=$(ls /root.*fs)
-echo "Creating symlink '/sysrootfs.img -> ${sysroot}'"
-ln -s "${sysroot}" /sysrootfs.img
+if [ -f /root.squashfs ]; then
+    fs='/root.squashfs'
+    fstype='squashfs'
+elif [ -f /root.erofs ]; then
+    fs='/root.erofs'
+    fstype='erofs'
+else
+    echo "No root.squashfs or root.erofs exists" >&2
+    exit 1
+fi
+
+# Let sysroot.mount know what path and type to use for mounting
+echo "Updating /rootfs.env with WHAT=${fs} TYPE=${fstype}"
+cat >/rootfs.env <<EOF
+WHAT=${fs}
+TYPE=${fstype}
+EOF
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
index 9c052e5a..013a8d23 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
@@ -58,18 +58,19 @@ isoroot=$(getarg coreos.liveiso= ||:)
 if [ -z "${isoroot}" ]; then
     # In this case, the rootfs is already unpacked into the initrd, or we need
     # to retrieve it
-    cat >"${UNIT_DIR}/sysroot.mount" <<EOF
+    cat >"${UNIT_DIR}/sysroot.mount" <<'EOF'
 # Automatically generated by live-generator
 
 [Unit]
 DefaultDependencies=false
-# Verifies that we have the right sysrootfs.img, or downloads it if needed
+# Verifies that we have the right rootfs image, or downloads it if needed
 After=coreos-livepxe-rootfs.service
 Before=initrd-root-fs.target
-
 [Mount]
-What=/sysrootfs.img
+EnvironmentFile=/rootfs.env
+What=${WHAT}
 Where=/sysroot
+Type=${TYPE}
 Options=loop
 EOF
 else
```

but I'm not opposed to leaving it the way it is either.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
jlebon: > What's about `sysrootfs.img` ?

Still prefer the cleanliness of `sysroot.img`, but WFM. :)

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
jlebon: > For setting the What= and Type= here this seems to work:

Ahh yes, I do like the explicitness of just templating the `What` also actually.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-rootfs.sh
nikita-dubrovskii: Done

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
nikita-dubrovskii: So, should i add Dusty's patch on top ? Does it make sense to have this explicity just in 1 place, while omitting the `else` case?

jlebon: Hmm yeah, I think it would require another systemd service that runs after we mount the ISO but before we mount the sysroot. OK, let's leave it this way for now and revisit this later.

dustymabe (approved):
LGTM

jlebon (approved):

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator
```diff
@@ -148,9 +147,8 @@ ${sysrootrequiresmountsfor}
 [Mount]
 What=${sysrootfsimg}
 Where=/sysroot
-Type=squashfs
-# Offset of the squashfs within the rootfs cpio.  Assumes newc format
-# and that a file named "root.squashfs" is the first member.  This offset
+# Offset of the root.erofs within the rootfs cpio.  Assumes newc format
+# and that a file named "root.[squash|ero]fs" is the first member.  This offset
```
jlebon, line 151:
```suggestion
# Offset of the sysroot image within the rootfs cpio.  Assumes newc format
# and that the sysroot image is the first member.  This offset
```

PR #3343, opened 2025-01-30
tests/quadlet: Fix network name retrieval in podman inspect

marmijo:
A recent Podman release updated `podman inspect` to properly return the NetworkID instead of the name `[1]`. Update the retrieval of the network name to use the first key in the dictionary, since there is only one network.

`[1]`: https://github.com/containers/podman/issues/24910


dustymabe (approved):
LGTM

Nice work!

Issue #44, opened 2019-01-25
Audit kernel command line


bgilbert:
Verify that everything on the kernel command line needs to be there.

jlebon:
I think we're in a good position now wrt our kernel cmdline and controlling it.

PR #3348, opened 2025-02-03
Build via Containerfile and derive from fedora-bootc

jlebon:
This builds Fedora CoreOS using a Containerfile. It uses the [FROM scratch flow](https://docs.fedoraproject.org/en-US/bootc/building-from-scratch/) to do a base compose.

This is structured in a way that we can build FCOS using _both_ `podman build` _and_ `cosa build`. This allows us to make the cutover much smoother. So then, we could turn this on in e.g. rawhide first and let it percolate down.

All the heavy lifting is done in the `build-rootfs` script. The idea is that this script is shared by both FCOS and RHCOS. Random notes:
1. It's in Python. This is fine, because the builder image is separate from the target rootfs. The `bootc-base-imagectl` script itself is in Python and so can only be used from the `:standard` image.
2. Packages are extracted from the manifest, and fed to `bootc-base-imagectl` (see
   https://gitlab.com/fedora/bootc/base-images/-/merge_requests/178) so it's part of a single compose.
3. We keep respecting the in-git lockfiles. In the future, how lockfiles work will change (see discussions in https://github.com/coreos/fedora-coreos-tracker/issues/1861).
4. What you see overall is a lot of cosa-isms and rpm-ostree-isms being lifted and carried right into the build script. Things like overlays, versioning, postprocessing, os-release mutating, etc...

Rechunking is expected to be done as a secondary step. Once we have https://github.com/containers/buildah/issues/5952, we can inline it back into the Containerfile.

---

Requires:
- [ship bootc-base-imagectl in Fedora (#44)](https://gitlab.com/fedora/bootc/base-images/-/issues/44) (currently to test this, you need to build fedora-bootc locally yourself)
- https://github.com/coreos/rpm-ostree/pull/5391
- Some bits of [base-imagectl: support --cachedir; use in Containerfile (!180)](https://gitlab.com/fedora/bootc/base-images/-/merge_requests/180)

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,50 @@
+# DO NOT EDIT. This Containerfile is produced by the concatenation of:
+# - Containerfile.advisory: contains this advisory
+# - Containerfile.args: contains stream-specific build args
+# - Containerfile.base: actual build logic
+# Rebuild it using `cat Containerfile.* > Containerfile`.
+
+### Containerfile.args
+
+# This is the developer default version. In pipelines, this is driven by versionary.
+ARG VERSION="42"
+# XXX: This uses the Konflux-built version until it takes over from pungi and
+# shows up at the official quay.io/fedora/fedora-bootc endpoint since the latter
+# doesn't yet have bootc-base-imagectl.
+# https://gitlab.com/fedora/bootc/base-images/-/issues/44
+# XXX: Note also this should be a digested pull that gets bumped.
+# https://gitlab.com/fedora/bootc/tracker/-/issues/34
+ARG BUILDER_IMG=quay.io/bootc-devel/fedora-bootc-42-standard:latest
+ARG MANIFEST=manifest.yaml
+
+### Containerfile.base
+
+FROM ${BUILDER_IMG} as builder
+
+ARG VERSION
+ARG MANIFEST
+
+# useful if you're hacking on rpm-ostree/bootc-base-imagectl
+# COPY rpm-ostree /usr/bin/
+# COPY bootc-base-imagectl /usr/libexec/
+
+# Note: once we can rely on https://github.com/coreos/rpm-ostree/pull/5391,
+# add this bit to the RUN command to make the developer path less painful.
+# --mount=type=cache,rw,id=coreos-build-cache,target=/cache
+RUN --mount=type=bind,target=/run/src /run/src/build-rootfs "${MANIFEST}" "${VERSION}" /target-rootfs
+
+FROM scratch
+ARG VERSION
+COPY --from=builder /target-rootfs/ /
+RUN <<EOF
+set -xeuo pipefail
+for script in /usr/libexec/coreos-postprocess-*; do
+  $script; rm $script
+done
+EOF
+
+LABEL containers.bootc=1
+LABEL ostree.bootable=1
+LABEL org.opencontainers.image.version=$VERSION
+STOPSIGNAL SIGRTMIN+3
+CMD ["/sbin/init"]
--- a/Containerfile.base
+++ b/Containerfile.base
@@ -1,0 +1,31 @@
+### Containerfile.base
+
+FROM ${BUILDER_IMG} as builder
+
+ARG VERSION
+ARG MANIFEST
+
+# useful if you're hacking on rpm-ostree/bootc-base-imagectl
+# COPY rpm-ostree /usr/bin/
+# COPY bootc-base-imagectl /usr/libexec/
+
+# Note: once we can rely on https://github.com/coreos/rpm-ostree/pull/5391,
+# add this bit to the RUN command to make the developer path less painful.
+# --mount=type=cache,rw,id=coreos-build-cache,target=/cache
+RUN --mount=type=bind,target=/run/src /run/src/build-rootfs "${MANIFEST}" "${VERSION}" /target-rootfs
+
+FROM scratch
+ARG VERSION
+COPY --from=builder /target-rootfs/ /
+RUN <<EOF
+set -xeuo pipefail
+for script in /usr/libexec/coreos-postprocess-*; do
+  $script; rm $script
+done
+EOF
+
+LABEL containers.bootc=1
+LABEL ostree.bootable=1
+LABEL org.opencontainers.image.version=$VERSION
+STOPSIGNAL SIGRTMIN+3
+CMD ["/sbin/init"]
--- a/build-rootfs
+++ b/build-rootfs
@@ -1,0 +1,236 @@
+#!/usr/bin/python3
+
+# This script is called by the Containerfile to build FCOS. Here's what it does at a high-level:
+# 1. It gathers the list of FCOS-specific packages using the manifests.
+# 2. It gathers the list of FCOS-specific overlays using the manifests.
+# 3. It runs `bootc-base-imagectl rebuild`, passing in the packages and overlays.
+# 4. It injects various metadata (e.g. image.json, live/ bits, and platforms.json).
+# 5. It runs the postprocess scripts defined in the manifest.
+
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import yaml
+
+
+ARCH = os.uname().machine
+CONTEXTDIR = '/run/src'
+
+
+def main():
+    manifest_name = sys.argv[1]
+    version = sys.argv[2]
+    target_rootfs = sys.argv[3]
+
+    manifest_path = os.path.join(CONTEXTDIR, manifest_name)
+    manifest = get_treefile(manifest_path)
+
+    packages = list(manifest['packages'])
+
+    locked_nevras = get_locked_nevras()
+    if locked_nevras:
+        inject_pool_repo(locked_nevras)
+
+    packages.extend(locked_nevras)
+    overlays = gather_overlays(manifest)
+    build_rootfs(target_rootfs, manifest_path, packages, overlays)
+
+    inject_live(target_rootfs)
+    inject_image_json(target_rootfs, manifest_path)
+    inject_platforms_json(target_rootfs)
+
+    if version != "":
+        inject_version_info(target_rootfs, manifest['mutate-os-release'], version)
+    inject_postprocess_scripts(target_rootfs, manifest)
+
+
+def get_treefile(manifest_path):
+    with tempfile.NamedTemporaryFile(suffix='.json', mode='w') as tmp_manifest:
+        # This ensures that the treefile represents only the CoreOS bits and
+        # doesn't recurse into fedora-bootc. We can drop this once we've fully
+        # cut over to derivation.
+        json.dump({
+            "variables": {
+                "deriving": True
+            },
+            "include": manifest_path
+        }, tmp_manifest)
+        tmp_manifest.flush()
+        data = subprocess.check_output(['rpm-ostree', 'compose', 'tree',
+                                        '--print-only', tmp_manifest.name])
+    return json.loads(data)
+
+
+def build_rootfs(target_rootfs, manifest_path, packages, overlays):
+    with tempfile.NamedTemporaryFile(mode='w') as argsfile:
+        for pkg in packages:
+            argsfile.write(f"--install={pkg}\n")
+        for overlay in overlays:
+            argsfile.write(f"--add-dir={overlay}\n")
+        argsfile.flush()
+        cache_arg = ['--cachedir=/cache'] if os.path.isdir('/cache') else []
+        subprocess.check_call(["/usr/libexec/bootc-base-imagectl",
+                               "--args-file", argsfile.name, "build-rootfs",
+                               "--manifest", 'fedora-minimal-plus',
+                               target_rootfs] + cache_arg)
+
+
+def inject_postprocess_scripts(rootfs, manifest):
+    # since we have the derive-only manifest handy, just inject a script now
+    # that we can execute in the second stage. we could of course use e.g.
+    # bwrap to run those now, but... it's just cleaner to use multi-stage build
+    # semantics
+    for i, script in enumerate(manifest.get('postprocess', [])):
+        with open(os.path.join(rootfs, f'usr/libexec/coreos-postprocess-{i}'), mode='w') as f:
+            os.fchmod(f.fileno(), 0o755)
+            f.write(script)
+
+
+def get_locked_nevras():
+    lockfile_path = os.path.join(CONTEXTDIR, f"manifest-lock.{ARCH}.json")
+    overrides_path = os.path.join(CONTEXTDIR, "manifest-lock.overrides.yaml")
+    overrides_arch_path = os.path.join(CONTEXTDIR, f"manifest-lock.overrides.{ARCH}.yaml")
+
+    locks = {}
+    for path in [lockfile_path, overrides_path, overrides_arch_path]:
+        if os.path.exists(path):
+            with open(path) as f:
+                if path.endswith('.yaml'):
+                    data = yaml.safe_load(f)
+                else:
+                    data = json.load(f)
+                # this essentially re-implements the merge semantics of rpm-ostree
+                locks.update({pkgname: v['evra'] if 'evra' in v else v['evr']
+                              for (pkgname, v) in data['packages'].items()})
+    return [f'{k}-{v}' for (k, v) in locks.items()]
+
+
+def inject_pool_repo(locked_nevras):
+    # When adding the pool, we only want to _filter in_ locked packages;
+    # matching `lockfile-repos` semantics. This is abusing pretty hard the
+    # `includepkgs=` semantic but... it works.
+    shutil.copyfile(os.path.join(CONTEXTDIR, "fedora-coreos-pool.repo"),
+                    "/etc/yum.repos.d/pool.repo")
+    packages = ','.join(locked_nevras)
+    with open("/etc/yum.repos.d/pool.repo", 'a') as f:
+        f.write(f"\nincludepkgs={packages}\n")
+
+
+# This re-implements rpm-ostree's mutate-os-release to preserve the historical
+# /usr/lib/os-release API, but we may in the future completely sever off of that
+# and only rely on org.opencontainers.image.version as argued in:
+# https://gitlab.com/fedora/bootc/base-images/-/issues/40
+# https://gitlab.com/fedora/bootc/base-images/-/issues/46
+def inject_version_info(rootfs, base_version, version):
+    os_release_path = os.path.join(rootfs, 'usr/lib/os-release')
+    with open(os_release_path) as f:
+        from collections import OrderedDict
+        os_release = OrderedDict([line.strip().split('=', 1) for line in f])
+
+    for key in ['VERSION', 'PRETTY_NAME']:
+        os_release[key] = os_release[key].replace(base_version, version)
+    os_release['OSTREE_VERSION'] = f"'{version}'"
+    os_release['IMAGE_VERSION'] = f"'{version}'"
+
+    with open(os_release_path, mode='w', encoding='utf-8') as f:
+        for (k, v) in os_release.items():
+            f.write(f'{k}={v}\n')
+
+
+# This re-implements cosa's overlay logic.
+def gather_overlays(manifest):
+    overlays = []
+    for layer in manifest.get('ostree-layers', []):
+        assert layer.startswith('overlay/')
+        overlays.append(os.path.join(CONTEXTDIR, 'overlay.d', layer[len('overlay/'):]))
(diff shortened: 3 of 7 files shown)


jlebon:
This is still in the early phase. It doesn't work yet.

Also note we're using tier-1 for now since there's no tier-x image available yet (https://gitlab.com/fedora/bootc/base-images/-/issues/25).

Edit: now using a tier-x image I built myself.

jlebon:
OK cool, got a successful build with this and https://github.com/coreos/rpm-ostree/pull/5274 which I can even rebase to and boot successfully. We're not rebuilding the initramfs yet nor adding overlays. Those shouldn't be hard though.

jlebon:
Updated this now and updated the PR description with more details! Still working on it (notably, going over preliminary diffs between a cosa-built tree and this).

image-default.yaml
```diff
@@ -0,0 +1,34 @@
+# This file contains defaults for image.yaml
+
+bootfs: "ext4"
+rootfs: "xfs"
```
cgwalters: Can you also link here https://github.com/osbuild/bootc-image-builder/pull/932

cgwalters (commented):
Looks generally sane to me!

Containerfile
```diff
@@ -0,0 +1,27 @@
+ARG VERSION="42"
+ARG MANIFEST=manifest.yaml
+
+FROM quay.io/fedora/fedora-bootc:42 as builder
+ARG VERSION
+ARG MANIFEST
+# # XXX: for hacking
+COPY rpm-ostree /usr/bin/
+COPY bootc-base-imagectl /usr/libexec/
+RUN --mount=type=cache,rw,id=cache,target=/cache \
```
dustymabe:
```suggestion
RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
```

build-rootfs
dustymabe, line 1: suggestion.. to reduce confusion, since this is coreos specific for now, rename to `coreos-build-rootfs`.

build-rootfs
```diff
@@ -0,0 +1,239 @@
...
+                                        '--print-only', tmp_manifest.name])
+    return json.loads(data)
+
+
+def build_rootfs(target_rootfs, manifest_path, packages):
+    inject_passwd_groups_hack()
+    with tempfile.NamedTemporaryFile(mode='w') as argsfile:
+        for pkg in packages:
+            argsfile.write("--install\n")
+            argsfile.write(f"{pkg}\n")
```
dustymabe:
```suggestion
            argsfile.write(f"--install\n{pkg}\n")
```

or would `argsfile.write(f"--install={pkg}\n")` work?

build-rootfs
```diff
@@ -0,0 +1,239 @@
...
+# This re-implements cosa's overlay logic.
+def inject_overlays(rootfs, manifest):
+    for layer in manifest.get('ostree-layers', []):
+        assert layer.startswith('overlay/')
+        path = os.path.join(CONTEXTDIR, 'overlay.d', layer[len('overlay/'):])
+        shutil.copytree(path, rootfs, dirs_exist_ok=True)
+
+    # inject live/ bits
+    target_path = os.path.join(rootfs, 'usr/share/coreos-assembler/live')
+    shutil.copytree(os.path.join(CONTEXTDIR, "live"), target_path)
```
dustymabe: I guess we could put this in a proper overlay to not have to special case this.

manifests/fedora-coreos-base.yaml
```diff
@@ -102,6 +100,26 @@ postprocess:
     ConditionPathExists=
     EOF
 
+  # sudo prefers its config files to be mode 440, and some security scanners
+  # complain if /etc/sudoers.d files are world-readable.
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1981979
```
dustymabe: Should we mention what "overlay" this is from?

manifests/fedora-coreos-base.yaml
```diff
@@ -102,6 +100,26 @@ postprocess:
...
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    chmod 440 /etc/sudoers.d/coreos-sudo-group
+
+  - |
```
dustymabe: Would be nice to have a comment for this block.

manifests/fedora-coreos-base.yaml
```diff
@@ -102,6 +100,26 @@ postprocess:
...
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    ln -sf multi-user.target /usr/lib/systemd/system/default.target
+
+  - |
```
dustymabe: Would be nice to have a comment for this block.

manifests/fedora-coreos-base.yaml
```diff
@@ -102,6 +100,26 @@ postprocess:
...
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    if [ -f /run/.containerenv ]; then
+      grep sudo /usr/lib/group >> /etc/group
+    fi
```
dustymabe: Should we add these three new postprocess scripts in a manifest that is shared with RHCOS/SCOS?

manifests/fedora-coreos.yaml
```diff
@@ -89,8 +93,14 @@ postprocess:
...
 
     EOF
-    cat /usr/etc/rpm-ostreed.conf >> /tmp/rpm-ostreed.conf
-    cp /tmp/rpm-ostreed.conf /usr/etc/rpm-ostreed.conf
+    conf_file=/usr/etc/rpm-ostreed.conf
+    # if we're deriving, then the config is in /etc already
+    if [ -f /run/.containerenv ]; then
+      conf_file=/etc/rpm-ostreed.conf
+    fi
+
```
dustymabe: Maybe delete this newline

manifests/fedora-coreos.yaml
```diff
@@ -134,13 +144,27 @@ postprocess:
         done
     done
 
+  - |
```
dustymabe, line 172: Would be nice to have a comment for this block.

manifests/fedora-coreos.yaml
```diff
@@ -134,13 +144,27 @@ postprocess:
...
-  # We don't ship man(1) or info(1)
-  - usr/share/info
-  - usr/share/man
-  # Drop text docs too
-  - usr/share/doc
+    # Nuke broken alternatives links
+    rm -f /etc/alternatives/*-man
+
+  # This replicates `etc-group-members` above but for the container-native case.
+  # We can nuke that other one once that's the only path we support.
```
dustymabe: The comment for the other one is more comprehensive :)

dustymabe:
Havne't had a chance to test this out yet. Some comments in the mean time.

manifests/fedora-coreos-base.yaml
jlebon: Maybe. For a first pass, I'm just keeping all these "postprocess moves" in the same manifest as the equivalent compose key they come from to ensure I don't mess up any inheritance in the process.

build-rootfs
jlebon: Yeah, true. It's trivial too to just inject it so will just do that for now to avoid the cosa churn.

build-rootfs
jlebon: Hmm, I like simple names. :) Having a coreos denotation seems unnecessary given that it lives in this repo.

jlebon:
Updated for comments!

This now requires https://gitlab.com/fedora/bootc/base-images/-/merge_requests/195.

Split prep patches in https://github.com/coreos/fedora-coreos-config/pull/3534.

Dropped the `WIP` prefix to signal this is nearing completion.

jlebon:
One thing worth highlighting is that if you look at the `build-rootfs` script, there's only a few keys that we look at currently:
- `packages`
- `ostree-layers`
- `mutate-os-release`
- `postprocess`

Once we switch to the container-native flow, every key other than those could be nuked. At that point, manifests are just a way of organizing those keys into a useful hierarchy to make sharing across streams and across RHCOS/SCOS easier, but it does make it easier to switch to a different way of doing that if we wanted to.

cgwalters:
Thanks again for all of the work on this! It's a giant milestone towards the original vision I had here...been a long, long journey and many parts have been much harder than I thought. But I do believe (hope) that this will pay itself back in a lot of ways. 

I mean, one thing *for sure* that is so much dramatically nicer is that use case of "hey I just want to drop in this one override for one package" is now just `FROM <base> RUN dnf -y upgrade foo` whereas before...so much bespoke tooling.

If there's anything more I (or the team) can do to help let me know!

jlebon:
OK, updated and dropped out of draft!

The new GHA will fail until the next Konflux run that picks up https://gitlab.com/fedora/bootc/base-images/-/merge_requests/195.

jlebon:
> Thanks again for all of the work on this! It's a giant milestone towards the original vision I had here...been a long, long journey and many parts have been much harder than I thought. But I do believe (hope) that this will pay itself back in a lot of ways.
> 
> I mean, one thing _for sure_ that is so much dramatically nicer is that use case of "hey I just want to drop in this one override for one package" is now just `FROM <base> RUN dnf -y upgrade foo` whereas before...so much bespoke tooling.
> 
> If there's anything more I (or the team) can do to help let me know!

Thanks! I think we should be able to get this in soon once people are back from conferences and can do a final review. Then we can work towards turning it on in e.g. rawhide to start.

The main blocker I see that's not obvious CoreOS-related things like cosa/pipeline work is https://gitlab.com/fedora/bootc/tracker/-/issues/34. Without that, we can't have semantically reproducible manifests.

build-rootfs
dustymabe: I agree with you that in the context of this file it doesn't seem necessary to name it more specifically, but the cache type mounts are all stored under `/var/tmp` and having it be named something more specific can be useful there.

Containerfile
```diff
@@ -0,0 +1,36 @@
...
+# COPY rpm-ostree /usr/bin/
+# COPY bootc-base-imagectl /usr/libexec/
+
+# Note: once we can rely on https://github.com/coreos/rpm-ostree/pull/5391,
+# add this bit to the RUN command to make the developer path less painful.
+# --mount=type=cache,rw,id=coreos-build-cache,target=/cache
+RUN --mount=type=bind,target=/run/src /run/src/build-rootfs "${MANIFEST}" "${VERSION}" /target-rootfs
+
+FROM scratch
+ARG VERSION
```
dustymabe, line 37: dumb question.. why do we have `ARG VERSION` 3 times in this file? I guess the context gets reset and naming it again is useful?

Containerfile
```diff
@@ -0,0 +1,36 @@
+ARG VERSION="42"
```
dustymabe: Considering this file will get synced to other branches of this repo should we not be hardcoding `42` here and also in the `FROM` command below this but rather pull that from the manifest somehow.

build-rootfs
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/python3
+
```
dustymabe, line 2: optional: might be nice to have a description of what this file is for (what it does) here.

dustymabe (dismissed):
LGTM

Containerfile
jlebon: Pulling it from the manifest would mean that `podman build` is no longer the primary interface. The base issue I think is that (AFAIK) there's no way to set things like labels and `FROM` images using dynamically derived information from the build itself.

What I was thinking there, even though I don't like it, is to make use of buildah's `#include` support. So then we could have `Containerfile` be like `manifest.yaml`; it doesn't get synced across branches and can contain stream-specific information, and `Containerfile.shared` which does all the heavy lifting.

So e.g.:

```Containerfile
# Containerfile
ARG VERSION="42" # default if building directly, but driven by versionary in the pipeline
ARG BUILDER="quay.io/bootc-devel/fedora-bootc-42-standard@sha256:..." # bumped by bumper
ARG MANIFEST=manifest.yaml
#include Containerfile.shared
```

```Containerfile
# Containerfile.shared

FROM ${BUILDER} as builder
...
```

You can also see how this would make it easy to implement variants in r-c-c.

The downside with this is that this makes this even more podman ecosystem-specific. Though I think we're willing to do that for https://github.com/containers/buildah/pull/5975 anyway once that lands to be able to inline rechunking.

Hmm, let me play with this.

Containerfile
jlebon: There are scoping rules with `ARG` that required me to redefine them when I was hacking on this, but yeah looking at the final state I think I should be able to get rid of some of them.

jlebon: OK no, we can't get rid of any of the `ARG`s. Each one is needed to scope in the global `ARG` defined at the top.

Containerfile
jlebon: The immediate problem with this is that it requires `cpp` which isn't in FCOS and so not in our builders either. While `cpp` is in cosa, it would be quite unfortunate more generally to require users to have `cpp`.

So moved away from that and just switched to requiring us to concatenate the pieces ourselves and committing that. I'm not really happy with this, but so far it looks like the least terrible option.

Containerfile.base
```diff
@@ -0,0 +1,31 @@
+### Containerfile.base
+
+FROM ${BUILDER_IMG} as builder
+
+ARG VERSION
+ARG MANIFEST
+
+# useful if you're hacking on rpm-ostree/bootc-base-imagectl
+# COPY rpm-ostree /usr/bin/
+# COPY bootc-base-imagectl /usr/libexec/
```
cgwalters, line 10: FWIW I have switched to trying to always build "our" projects (like rpm-ostree and bootc) into a container image by default instead of my toolbox. Which means that then the expected flow here is to override `${BUILDER_IMG}`, not `COPY`.

Containerfile
dustymabe: > So moved away from that and just switched to requiring us to concatenate the pieces ourselves and committing that. I'm not really happy with this, but so far it looks like the least terrible option.

Yeah. This isn't great, but I'm happy with at least unblocking ourselves for now.

cgwalters: Up to you all of course but my lean on things like this is to just have a plain `podman build` do something useful (it's ok if the version numbers aren't what's expected) but I would expect to have a `Justfile` or something equivalent that helps me to trivially locally what a more "production/orchestrated" pipeline would do.

jlebon: Definitely possible, yeah. I think it might eventually happen, though I want to try to avoid adding a `Makefile`/`Justfile` into this repo, and always keeping `podman build` the canonical entrypoint since it works as a "choke point" through which we must ensure all enhancements are implemented.

Dynamic label information (like versions) is one of those that you kinda have to drive from the outside (though https://github.com/containers/buildah/pull/5975 in theory would allow it), but there really shouldn't be that many such cases.

Containerfile.base
jlebon: I like that, though iteration speed-wise it seems less efficient.

cgwalters: The [incremental caching in bootc](https://github.com/bootc-dev/bootc/blob/9493f2c8ec2dface58e4e5f9b3de9b097b440814/hack/Containerfile#L23-L25) works just fine! It definitely works less well for autotools based things :cry:

dustymabe:
For some reason in my local desktop I can't get this to build. I'm seeing the following when I `podman build --security-opt=label=disable --cap-add=all --device /dev/fuse -t localhost/fedora-coreos .` locally:

```
Creating empty machine-id
error: Executing compose install: Subprocess failed: ExitStatus(unix_wait_status(256))
p/dracut/dracut.29iPSx/initramfs /usr/lib/dracut/modules.d/35coreos-live/coreos-livepxe-persist-osmet.service /usr/lib/systemd/system/coreos-livepxe-persist-osmet.service
cp: setting attributes for '/tmp/dracut/dracut.29iPSx/initramfs/lib/modules/6.14.9-300.fc42.x86_64//kernel/drivers/block/loop.ko.xz': Operation not supported
dracut-install: ERROR: 'cp --reflink=auto --sparse=auto --preserve=mode,xattr,timestamps,ownership -fL /lib/modules/6.14.9-300.fc42.x86_64//kernel/drivers/block/loop.ko.xz /tmp/dracut/dracut.29iPSx/initramfs/lib/modules/6.14.9-300.fc42.x86_64//kernel/drivers/block/loop.ko.xz' failed with 1
dracut-install: dracut_install 'loop' ERROR
dracut-install: ERROR: installing 'loop'
dracut[E]: FAILED:  /usr/lib/dracut/dracut-install -D /tmp/dracut/dracut.29iPSx/initramfs --kerneldir /lib/modules/6.14.9-300.fc42.x86_64/ -m loop
dracut[F]: installkernel failed in module coreos-live
```

that being said, @jlebon isn't able to reproduce. I'm happy to merge this in and debug my setup later.

dustymabe (approved):

jbtrystram:
FYI i was able to build that locally but i had to override the from with : 
`quay.io/bootc-devel/fedora-bootc-42-standard:git-f7d48642c044ccfa89f9c0b9c18871f0f1ed1e8d`

jlebon:
> FYI i was able to build that locally but i had to override the from with : `quay.io/bootc-devel/fedora-bootc-42-standard:git-f7d48642c044ccfa89f9c0b9c18871f0f1ed1e8d`

@jbtrystram Hmm weird, the latest `quay.io/bootc-devel/fedora-bootc-42-standard` should've worked. What error were you seeing?

jbtrystram:
bootc-base-imagebuilder doesn't know about `--add-dir` on latest :

```
 [~] podman run --rm -it --pull=always quay.io/bootc-devel/fedora-bootc-42-standard:latest grep add-dir /usr/libexec/bootc-base-imagectl                            
Trying to pull quay.io/bootc-devel/fedora-bootc-42-standard:latest...
Getting image source signatures
Copying blob c22a22f9973e skipped: already exists  
Copying config f2fd0d907f done   | 
Writing manifest to image destination
 [~]                                                                                                                                                                
 [~] podman run --rm -it --pull=always quay.io/bootc-devel/fedora-bootc-42-standard:git-f7d48642c044ccfa89f9c0b9c18871f0f1ed1e8d grep add-dir /usr/libexec/bootc-base-imagectl
Trying to pull quay.io/bootc-devel/fedora-bootc-42-standard:git-f7d48642c044ccfa89f9c0b9c18871f0f1ed1e8d...
Getting image source signatures
Copying blob a975329ad19c skipped: already exists  
Copying config 3da1278f65 done   | 
Writing manifest to image destination
    build_rootfs.add_argument("--add-dir", help='Copy dir contents into the target', action='append', default=[], metavar='DIR')
```

which is surprising given that `latest` image have been updated recently, but i haven't looked at the konflux definitions, maybe latest is only updated on a git tag or a manual release ?

jbtrystram:
ah, looking at the labels on `latest` I see : 
```json
"Labels": {
        "architecture": "x86_64",
        "bootc.diskimage-builder": "quay.io/centos-bootc/bootc-image-builder",
        "build-date": "2025-06-04T14:01:42",
        "containers.bootc": "1",
        "io.buildah.version": "1.39.0-dev",
        "org.opencontainers.image.version": "43",
        "vcs-ref": "08254c0ba435bed9f8871a0c6b0ed806599ed41c",
        "vcs-type": "git"
}
```

haven't been rebuilt since june 04th

jlebon:
Really odd, it seems like the image was rolled back for some reason. But anyway, with https://gitlab.com/fedora/bootc/base-images/-/merge_requests/201 we should be able to use the canonical bootc images instead. I'll open a PR to switch to that once that change is picked up.

cgwalters:
```
$ podman run --rm -ti quay.io/fedora/fedora-bootc:42 /usr/libexec/bootc-base-imagectl --help                                                        2 06/11/2025 09:29:50 AM
usage: bootc-base-imagectl [-h] [--args-file FILE] {build-rootfs,rechunk,list} ...
```

(And thank you for that!!!)

mmartinv:
> ah, looking at the labels on `latest` I see :
> 
> ```json
> "Labels": {
>         "architecture": "x86_64",
>         "bootc.diskimage-builder": "quay.io/centos-bootc/bootc-image-builder",
>         "build-date": "2025-06-04T14:01:42",
>         "containers.bootc": "1",
>         "io.buildah.version": "1.39.0-dev",
>         "org.opencontainers.image.version": "43",
>         "vcs-ref": "08254c0ba435bed9f8871a0c6b0ed806599ed41c",
>         "vcs-type": "git"
> }
> ```
> 
> haven't been rebuilt since june 04th

This is basically my fault: Yesterday I was trying to clean up the failed releases and I accidentally triggered some of them which, for what it seems, were quite old. This should have been fixed with the latest builds from yesterday's compose

PR #3351, opened 2025-02-06
[rawhide] manifest: Rawhide is now F43

travier:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1851

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -2,7 +2,7 @@ variables:
   stream: rawhide
   prod: false
 
-releasever: 42
+releasever: 43
 
 repos:
   - fedora-rawhide


aaradhak (approved):
LGTM

dustymabe (approved):

dustymabe:
merging this as all that was failed was a `warn: true` test and I think merging this may resolve that failure.

dustymabe:
The kernel-replace test failed here because it's trying to override replace with newest kernel in the N-1 version of Fedora. Right now for rawhide N=43 and N-1=42, but there is no F43 kernel in the rawhide repos yet so it's the same kernel that's latest in the F42 and in the rawhide repos so the test is failing.

Will fast-track the F43 kernel and it should resolve this.

dustymabe:
https://github.com/coreos/fedora-coreos-config/pull/3353

PR #3355, opened 2025-02-07
overlay.d/15fcos: add a migration script to move to OCI images

jbtrystram:
to simplify testing for https://github.com/coreos/fedora-coreos-tracker/issues/1823
ship a script to fake the ostree origin to appear like it's on an
OCI deployement.
    
Just ship the migration script for now, without the systemd unit, to  allow testing.

### DIFF
--- a/overlay.d/15fcos/usr/libexec/coreos-oci-rebase
+++ b/overlay.d/15fcos/usr/libexec/coreos-oci-rebase
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+# This is a migration script to move FCOS to OCI transport
+#
+# Users that have disabled Zincati or use a non default ostree remote
+# won't be migrated.
+#
+# see https://fedoraproject.org/wiki/Changes/CoreOSOstree2OCIUpdates
+# and https://github.com/coreos/fedora-coreos-tracker/issues/1823
+
+CINCINNATI_URL="${CINCINNATI_URL:-https://raw-updates.coreos.fedoraproject.org/v1/graph}"
+
+# Save the rpm-ostree status.
+echo "Saving rpm-ostree status."
+status=$(rpm-ostree status --json --booted)
+
+# Maybe the machine is already on an OCI deployment
+booted_imgref=$(jq -r '.deployments[0]."container-image-reference"' <<< "$status")
+
+if [ "$booted_imgref" != "null" ]; then
+    echo "The booted deployment is already an OCI container."
+    exit 0
+fi
+
+# check if the origin was changed
+origin=$(jq -r '.deployments[0].origin' <<< "$status" | cut -d ':' -f 1)
+origin_url=$(ostree remote show-url "$origin")
+if [ "$origin_url" != "https://ostree.fedoraproject.org" ]; then
+    echo "ERROR: The OSTree origin is not matching the default Fedora CoreOS."
+    exit 0
+fi
+
+# fetch the SHA checksum of the matching OCI image for the booted deployment
+version=$(jq -r '.deployments[0].version' <<< "$status")
+stream=$(jq -r '.deployments[0]."base-commit-meta"."fedora-coreos.stream"' <<< "$status")
+checksum=$(jq -r '.deployments[0].checksum' <<< "$status")
+arch=$(arch)
+cin_url="$CINCINNATI_URL?basearch=$arch&stream=$stream&oci=true"
+
+# Grab the OCI update graph that matches our stream and arch.
+echo "Fetching cincinnati update graph for stream $stream on $arch"
+cincinnati_graph=$(curl "$cin_url" -s)
+imgref=$(echo "$cincinnati_graph" | jq --arg VERSION "$version" -r '.nodes[]  | select(.version==$VERSION) | .payload')
+if [ "$imgref" == "" ]; then
+    echo "ERROR: Could not find the current deployment in the update graph from cincinnati."
+    echo "   If it is proxied, you can override the cincinnati url with an environment varable."
+    echo "   e.g. CINCINNATI_URL=https://custom-cincinnati-address.com/v1/graph /usr/libexec/coreos-oci-rebase"
+    exit 0
+fi
+
+echo "Found OCI image $imgref in the update graph that matches the local deployment version."
+digest=$(echo "$imgref" | cut -d '@' -f 2)
+
+# Proceed with the migration by writing an override status file,
+# so at the next update, Zincati will pull the OCI image.
+
+echo "Writing a status override file for the booted deployment in /run/zincati/booted-status-override.json"
+cat > /run/zincati/booted-status-override.json << EOF
+{
+    "booted": true,
+    "container-image-reference": "ostree-remote-registry:fedora:$imgref",
+    "container-image-reference-digest": "$digest",
+    "base-commit-meta": {
+        "fedora-coreos.stream": "$stream"
+    },
+    "checksum": "$checksum",
+    "version": "$version"
+}
+EOF
+
+echo "Zincati will rebase to an OCI image for the next update."


jbtrystram:
There is at least one missing bit here : I found out that when faking the origin file, `rpm-ostree --status` doesn't pick up the change immediately.
Something like `rpm-ostree useoverlay` does it, but obviously it's not great. I tried to restart the service rpm-ostreed, found a `refresh` subcommand, but no luck.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,99 @@
...
+write_motd () {
+
+# Change the output color to yellow
+warn=$(echo -e '\033[0;33m')
+# No color
+nc=$(echo -e '\033[0m')
+
+motd_path=/run/motd.d/40-fcos-oci-rebase.motd
+
+cat << EOF > "${motd_path}"
```
dustymabe: I think we should have a separate script that writes out the MOTD and it should run on every boot unless the user has neutered it by touching a stamp file.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,99 @@
...
+} >> "$tmpfile"
+
+mount -o remount,rw /sysroot
+
+# Replace the origin with our crafted one
+cp "$tmpfile" /ostree/deploy/fedora-coreos/deploy/"$checksum".0.origin
+
+
+# Restart Zincati
+systemctl restart zincati
```
dustymabe: this happening here leads to temporary problems.. basically if there is an update available `$now` (which I guess is unlikely but could be the case when older systems are jumping through this new barrier we are going to create) then zincati will do it's thing and rpm-ostree will too, but if rpm-ostreed hasn't timed out and deactivated since we wrote the custom origin file then it will complain:

```
Feb 27 22:01:40 cosa-devsh rpm-ostree[3917]: Txn Deploy on /org/projectatomic/rpmostree1/fedora_coreos failed: Cannot look up version while tracking a container image reference
```

Maybe we should restart `rpm-ostreed` here too just to make sure it's picked up the new custom origin?

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
dustymabe: basically we can roll out the MOTD script later (i.e. like a month in to F42 or something).

dustymabe:
One thing I notice that isn't great is it still says `CustomOrigin` even after the update to the container from the registry:

```
State: idle
Deployments:
● quay.io/fedora/fedora-coreos@sha256:40c0e515ed93cb8a9d5a773fb5a206b13e2edda02d92c6011a84399ceb31ed03
             CustomOrigin: Fedora CoreOS testing stream
                  Version: 41.20250215.1.0 (2025-02-17T11:44:02Z)

  ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
             CustomOrigin: Fedora CoreOS testing stream
                  Version: 41.20250130.1.0 (2025-01-31T20:20:28Z)

```

is that expected?

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,99 @@
...
+#
+# see https://fedoraproject.org/wiki/Changes/CoreOSOstree2OCIUpdates
+# and https://github.com/coreos/fedora-coreos-tracker/issues/1823
+
+
+# Maybe the machine is already on an OCI deployment
+booted_imgref=$(rpm-ostree status --json --booted | jq -r '.deployments[0]."container-image-reference"')
+
+if [ "$booted_imgref" != "null" ]; then
+    echo "The booted deployement is already an OCI container."
```
dustymabe:
```suggestion
    echo "The booted deployment is already an OCI container."
```

dustymabe:
Also on the system I see a bunch of these messages. Not sure if we can do anything about them:

```
Feb 27 22:08:58 cosa-devsh rpm-ostree[3591]: failed to query container image base metadata: Reading manifest data from commit: Missing ostree.manifest-digest metadata on merge commit
```

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,99 @@
...
+# so at the next update, Zincati will pull the OCI image
+
+# get the currently booted ostree checksum
+checksum=$(rpm-ostree status --booted --json | jq -r '.deployments[0].checksum')
+# fetch the SHA checksum of the matching OCI image for the booted deployment
+version=$(rpm-ostree status --booted --json | jq -r '.deployments[0].version')
+stream=$(rpm-ostree status --booted --json | jq -r '.deployments[0]."base-commit-meta"."fedora-coreos.stream"')
+arch=$(arch)
+cincinnati_url="https://raw-updates.coreos.fedoraproject.org/v1/graph?basearch=$arch&stream=$stream&oci=true"
+imgref=$(curl "$cincinnati_url" -s | jq --arg VERSION "$version" -r '.nodes[]  | select(.version==$VERSION) | .payload')
```
dustymabe: We should make a failure here nonfatal. IIUC the reason for the curl is more cosmetic than anything else (so `rpm-ostree status` shows something not unexpected), but if it fails for whatever reason I think we should continue the migration.

Things that could cause this to fail could be complicated network environments like proxies, or maybe wifi where the NIC hasn't come up fully yet or something.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
jbtrystram: Interesting scenario, I didn't thought of that. 
This definitely requires more testing in these corner cases.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
jbtrystram: It's not only cosmetic : Zincati will use that digest to find $currentNode in the graph. 
https://github.com/coreos/zincati/blob/fb745a46c980827959763bcaaecdb4363a35321f/src/cincinnati/mod.rs#L231

jbtrystram: So i guess if the update server is proxied we need to pull that info from the Zincati config

jbtrystram:
> Also on the system I see a bunch of these messages. Not sure if we can do anything about them:
> 
> ```
> Feb 27 22:08:58 cosa-devsh rpm-ostree[3591]: failed to query container image base metadata: Reading manifest data from commit: Missing ostree.manifest-digest metadata on merge commit
> ```

I guess that's a byproduct of faking out the origin file

jbtrystram:
> One thing I notice that isn't great is it still says `CustomOrigin` even after the update to the container from the registry:
> ...
> is that expected?

Yes. Because otherwise we cannot communicate which stream the node is following, as the digest does not show that information.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
jbtrystram: I added a knob and checked for failure

jbtrystram:
@dustymabe @jlebon I tweaked this script adding some logging, handling failures and allowing to override the cincinnati address for proxied environments.

I also stop rpm-ostree and zincati while overwriting the origin file to avoid a race or conflict.

jbtrystram:
I just found out `ostree admin set-origin`. [manpage](https://ostreedev.github.io/ostree/man/ostree-admin-set-origin.html)
So i experimented with: 
```
ostree admin set-origin fedora-coreos ostree-remote-image:fedora:registry:$imgref --index 0 -s custom-url=$imgref  -s custom-description="Fedora CoreOS testing stream"
```
Instead of rewriting the origin file manually. 

The command does not errors out but I don't see any changes ?

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
dustymabe: Should be handed by the update with the stop + restart.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
dustymabe: > It's not only cosmetic : Zincati will use that digest to find $currentNode in the graph. https://github.com/coreos/zincati/blob/fb745a46c980827959763bcaaecdb4363a35321f/src/cincinnati/mod.rs#L231

ouch.. I wish there was a fallback where it would find it based on version.. Obviously that would only be useful during this transition period.

One thing we could do is hardcode in this script 5-10 possible `ostree commit -> container hash` mappings and use that. Would help us get past this transitionairy period and not require user manual interaction.

dustymabe:
> > One thing I notice that isn't great is it still says `CustomOrigin` even after the update to the container from the registry:
> > ...
> > is that expected?
> 
> Yes. Because otherwise we cannot communicate which stream the node is following, as the digest does not show that information.

Hmm. So basically we are changing the origin file forever here and what's reported isn't necessarily what's true any more. i.e. if someone runs this script today it will end up with:

```
[origin]
container-image-reference=ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
custom-url=quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
custom-description=Fedora CoreOS testing stream
```

i.e. `custom-description=Fedora CoreOS testing stream` isn't true anymore if the user rebases to `stable` or `next`, right?

Maybe we need to find some way to sanitize the origin file once we're done with all of this?

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,91 @@
...
+} >> "$tmpfile"
+
+mount -o remount,rw /sysroot
+
+echo "Stopping rpm-ostree and zincati."
+systemctl stop rpm-ostreed zincati
+
+echo "Rewriting the origin file for booted deployment: $checksum."
+# Replace the origin with our crafted one
+mv "$tmpfile" /ostree/deploy/fedora-coreos/deploy/"$checksum".0.origin
```
dustymabe: We need to be careful here not to change too much:

Before Script:

```
[core@cosa-devsh ~]$ ls -lZ /ostree/deploy/fedora-coreos/deploy/c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c.0.origin 
-rw-r--r--. 1 root root system_u:object_r:root_t:s0 50 Aug  1  2022 /ostree/deploy/fedora-coreos/deploy/c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c.0.origin
```

After Script:

```
[core@cosa-devsh ~]$ ls -lZ /ostree/deploy/fedora-coreos/deploy/c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c.0.origin 
-rw-------. 1 root root unconfined_u:object_r:user_tmp_t:s0 332 Mar  3 14:19 /ostree/deploy/fedora-coreos/deploy/c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c.0.origin
```

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
jbtrystram: > ouch.. I wish there was a fallback where it would find it based on version.. Obviously that would only be useful during this transition period.

I tried that initially but it's not possible with skopeo, because it does not get the ecpected digest when inspecting multi-arch manifests. So we'd have to pull the whole manifest using the version tag and inspect it.
See https://github.com/containers/skopeo/issues/1554

> One thing we could do is hardcode in this script 5-10 possible mappings....

So we'd had to wait for the next 5-10 release before enabling the barrier.. And also, since it's a barrier only one is enough I think. 

Embedding mapping for this "experimental" migration script is weird imho. But we can add a knob to provide the digest as an argument

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
jbtrystram: good point ! I'll update to fix permissions and selinux context

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
dustymabe: > > ouch.. I wish there was a fallback where it would find it based on version.. Obviously that would only be useful during this transition period.
> 
> I tried that initially but it's not possible with skopeo, because it does not get the ecpected digest when inspecting multi-arch manifests. So we'd have to pull the whole manifest using the version tag and inspect it. See [containers/skopeo#1554](https://github.com/containers/skopeo/issues/1554)

No. I mean we allow the version of the local system to be used as a substitute to OCI digest. We wouldn't be using `skopeo`, we'd just be indexing into the graph using version string instead of the digest.

but like I said, I think this would be a lot of work for not much gain.. and think the below option would be better

> 
> > One thing we could do is hardcode in this script 5-10 possible mappings....
> 
> So we'd had to wait for the next 5-10 release before enabling the barrier.. And also, since it's a barrier only one is enough I think.
> 
> Embedding mapping for this "experimental" migration script is weird imho. But we can add a knob to provide the digest as an argument


I don't think so. We wouldn't have to wait.. this migration script runs on existing systems. We'd just keep adding to the mappings as we did new releases.

dustymabe:
Talked a little more with @jlebon today. We don't feel we need to block the releases tomorrow on getting this in because people can just download/run the script if they want. At the same time if we feel this is ready we also won't block it from going into the release either since it would only get run if someone knew about it and executed it.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,91 @@
...
+
+echo "Found matching OCI image: $imgref."
+
+# Empty the current origin
+tmpfile=$(mktemp)
+sed -e '/^refspec=/d' \
+    -e '/^baserefspec=/d' /ostree/deploy/fedora-coreos/deploy/"$checksum".0.origin > "$tmpfile"
+
+{
+    echo "container-image-reference=ostree-remote-image:fedora:registry:$imgref"
```
dustymabe:
```suggestion
    echo "container-image-reference=ostree-remote-registry:fedora:$imgref"
```

shorter form that is equivalent.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,91 @@
...
+# If Zincati is disabled, do nothing as the user takes responsibility for updates.
+if ! systemctl is-enabled --quiet zincati; then
+    echo "Zincati is disabled, not migrating. Please rebase manually."
+    exit 1
+fi
+
+# Save the rpm-ostree status then stop zincati and rpm-ostree.
+status=$(mktemp)
+echo "Saving rpm-ostree status."
+rpm-ostree status --json --booted > "$status"
```
dustymabe: We don't really need a tmpfile here IIUC:

```suggestion
echo "Saving rpm-ostree status."
status=$(rpm-ostree status --json --booted)
```

then in all cases below change `< "$status"` to `<<< "$status"`

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
jbtrystram: i am not sure how we can work out the mapping because we don't know the digest until we create the image, and the digest should go in the image as a mapping value. or each release we add update the mapping for the previous release, but then the fallback is to rollback a version but to the OCI image ..

jbtrystram:
> Hmm. So basically we are changing the origin file forever here and what's reported isn't necessarily what's true any more. i.e. if someone runs this script today it will end up with:
> 
> ```
> [origin]
> container-image-reference=ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
> custom-url=quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
> custom-description=Fedora CoreOS testing stream
> ```
> 
> i.e. `custom-description=Fedora CoreOS testing stream` isn't true anymore if the user rebases to `stable` or `next`, right?
> 
> Maybe we need to find some way to sanitize the origin file once we're done with all of this?

I updated the script to insert $stream instead of the hardcoded "testing" in the custom description, it was a leftover, thanks for catching that. 

However i am not sure about "changing the origin file forever". Wouldn't a manual rebase create a new origin file with the deployement ? I don't think rpm-ostree will forward custom origin fields to a new origin.
I am going to test that

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
jbtrystram: fixed

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
dustymabe: Yes. I think you are right.

jbtrystram:
With https://github.com/coreos/zincati/pull/1273 changes , dropping the following in `run/zincati/booted-status-override.json` triggers the migration : 
```json
             {
                "booted": true,
                "container-image-reference": "ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos:$STREAM",
                "container-image-reference-digest" : "sha256:$DIGEST",
                "base-commit-meta": {
                    "fedora-coreos.stream": "$STREAM"
                },
                "checksum": "oci-rebase",
                "version": "$VERSION"
            }
```

Without having to change the origin file so the `rpm-ostree status` will stay true.

```
[root@cosa-devsh core]# rpm-ostree status
State: idle
  DriverState: active; update staged: 41.20250302.2.0; reboot delayed due to active user sessions
Deployments:
  ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:testing
                   Digest: sha256:7bb94d1516fc6e368eb9cf07a4fe23ca7378e2e1ddb76c6a1902f75cb5c9203a
                  Version: 41.20250302.2.0 (2025-03-04T16:52:41Z)
                     Diff: 147 upgraded

● fedora:fedora/x86_64/coreos/testing
                  Version: 41.20241215.2.0 (2024-12-17T00:07:38Z)
                   Commit: 0e93f07d8d11856eb0773bde01757337320b26eb823cb5faff2c89d2849edb0a
             GPGSignature: Valid signature by 466CF2D8B60BC3057AA9453ED0622462E99D6AD1

```

jbtrystram:
Ok updated the script to do that and tested.
It's way less invasive now.



> Also on the system I see a bunch of these messages. Not sure if we can do anything about them:
> ```
> Feb 27 22:08:58 cosa-devsh rpm-ostree[3591]: failed to query container image base metadata: Reading manifest data from commit: Missing ostree.manifest-digest metadata on merge commit
> ```

This does not happens anymore.

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,77 @@
...
+digest=$(echo "$imgref" | cut -d '@' -f 2)
+
+# Proceed with the migration by writing an override status file,
+# so at the next update, Zincati will pull the OCI image.
+
+echo "Writing a status override file for the booted deployment in /run/zincati/booted-status-override.json"
+cat > /run/zincati/booted-status-override.json << EOF
+{
+    "booted": true,
+    "container-image-reference": "ostree-remote-image:fedora:registry:$imgref",
```
dustymabe: why not use the shorter version here?

```suggestion
    "container-image-reference": "ostree-remote-registry:fedora:$imgref",
```

dustymabe:
ok with the latest updates from the https://github.com/coreos/zincati/pull/1273 I'm seeing this error when running the migration script as part of an `ExecStartPre=`


```
Mar 13 13:45:39 cosa-devsh systemd[1]: zincati.service: start operation timed out. Terminating.
Mar 13 13:45:39 cosa-devsh systemd[1]: zincati.service: Failed with result 'timeout'.
Mar 13 13:45:39 cosa-devsh systemd[1]: Failed to start zincati.service - Zincati Update Agent.
Mar 13 13:45:49 cosa-devsh systemd[1]: zincati.service: Scheduled restart job, restart counter is at 1.
Mar 13 13:45:49 cosa-devsh systemd[1]: Starting zincati.service - Zincati Update Agent...
Mar 13 13:45:49 cosa-devsh coreos-oci-rebase[4105]: Saving rpm-ostree status.
Mar 13 13:45:49 cosa-devsh coreos-oci-rebase[4105]: Fetching cincinnati update graph for stream next on x86_64
Mar 13 13:45:50 cosa-devsh coreos-oci-rebase[4105]: Found OCI image quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918 in the update graph that matches the local deployment version.
Mar 13 13:45:50 cosa-devsh coreos-oci-rebase[4105]: Writing a status override file for the booted deployment in /run/zincati/booted-status-override.json
Mar 13 13:45:50 cosa-devsh coreos-oci-rebase[4105]: Zincati will rebase to an OCI image for the next update.
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::cli::agent] starting update agent (zincati 0.0.29)
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::cincinnati] Cincinnati service: https://updates.coreos.fedoraproject.org
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::cli::agent] agent running on node 'c26ec99392314d71a26afd4a59238a18', in update group 'default'
Mar 13 13:45:50 cosa-devsh zincati[4136]: [DEBUG zincati::metrics] started metrics service on Unix-domain socket '/run/zincati/public/metrics.promsock'
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::update_agent::actor] registering as the update driver for rpm-ostree
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Loaded sysroot 
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Locked sysroot 
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Initiated txn Deploy for client(dbus:1.87 unit:zincati.service uid:981): /org/projectatomic/rpmostree1/fedora_coreos
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Process [pid: 4156 uid: 981 unit: zincati.service] connected to transaction progress
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Txn Deploy on /org/projectatomic/rpmostree1/fedora_coreos successful
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Unlocked sysroot
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Process [pid: 4156 uid: 981 unit: zincati.service] disconnected from transaction progress
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: In idle state; will auto-exit in 61 seconds
Mar 13 13:45:50 cosa-devsh zincati[4136]: [DEBUG zincati::update_agent::actor] no other local finalized deployments found; no update targets will be excluded.
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::update_agent::actor] initialization complete, auto-updates logic enabled
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::strategy] update strategy: immediate
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::update_agent::actor] reached steady state, periodically polling for updates
Mar 13 13:45:50 cosa-devsh systemd[1]: Started zincati.service - Zincati Update Agent.
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::cincinnati] current release detected as not a dead-end
Mar 13 13:45:50 cosa-devsh zincati[4136]: [DEBUG zincati::cincinnati] MOTD updated with no dead-end state
Mar 13 13:45:50 cosa-devsh zincati[4136]: [INFO  zincati::update_agent::actor] target release '41.20250302.1.0' selected, proceeding to stage it
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Loaded sysroot 
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Locked sysroot 
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Initiated txn Rebase for client(dbus:1.95 unit:zincati.service uid:981): /org/projectatomic/rpmostree1/fedora_coreos
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Process [pid: 4203 uid: 981 unit: zincati.service] connected to transaction progress
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Txn Rebase on /org/projectatomic/rpmostree1/fedora_coreos failed: Old and new refs are equal: ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:next
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Unlocked sysroot
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: Process [pid: 4203 uid: 981 unit: zincati.service] disconnected from transaction progress
Mar 13 13:45:50 cosa-devsh rpm-ostree[3183]: In idle state; will auto-exit in 61 seconds
Mar 13 13:45:50 cosa-devsh zincati[4136]: [ERROR zincati::update_agent::actor] failed to stage deployment: rpm-ostree deploy failed:
Mar 13 13:45:50 cosa-devsh zincati[4136]:     error: Old and new refs are equal: ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:next
Mar 13 13:45:50 cosa-devsh zincati[4136]:     
Mar 13 13:46:52 cosa-devsh systemd[1]: rpm-ostreed.service: Deactivated successfully.
Mar 13 13:46:52 cosa-devsh systemd[1]: rpm-ostreed.service: Consumed 22.362s CPU time, 452.1M memory peak.
```


This is the file that got created:

```
[core@cosa-devsh ~]$ cat /run/zincati/booted-status-override.json 
{
    "booted": true,
    "container-image-reference": "ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918",
    "container-image-reference-digest" : "sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918",
    "base-commit-meta": {
        "fedora-coreos.stream": "next"
    },
    "checksum": "c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c",
    "version": "41.20250130.1.0"
}
```

jbtrystram:
> ok with the latest updates from the [coreos/zincati#1273](https://github.com/coreos/zincati/pull/1273) I'm seeing this error when running the migration script as part of an `ExecStartPre=`
> ...

Just for reference this was fixed with https://github.com/coreos/zincati/commit/2ff39a6d2def3f2af7873dc25eeeee296f744acf

dustymabe:
Let's reference the (newly created) https://github.com/coreos/fedora-coreos-tracker/issues/1890 issue in commit messages for this.

jbtrystram:
> Let's reference the (newly created) [coreos/fedora-coreos-tracker#1890](https://github.com/coreos/fedora-coreos-tracker/issues/1890) issue in commit messages for this.

updated commit message

overlay.d/15fcos/usr/libexec/coreos-oci-rebase
```diff
@@ -0,0 +1,77 @@
...
+# see https://fedoraproject.org/wiki/Changes/CoreOSOstree2OCIUpdates
+# and https://github.com/coreos/fedora-coreos-tracker/issues/1823
+
+CINCINNATI_URL="${CINCINNATI_URL:-https://raw-updates.coreos.fedoraproject.org/v1/graph}"
+
+# If Zincati is disabled, do nothing as the user takes responsibility for updates.
+if ! systemctl is-enabled --quiet zincati; then
+    echo "Zincati is disabled, not migrating. Please rebase manually."
+    exit 0
+fi
```
dustymabe: If we implement this to run as part of the zincati startup then this won't be needed since it would only run if zincati was enabled and starting anyway.

jbtrystram: That's a good catch ! Removed

dustymabe (approved):

dustymabe (approved):

PR #3356, opened 2025-02-12
test: verify FCOS runs on `TDX_CAPABLE` instances in GCP

HuijingHei:
See https://github.com/coreos/fedora-coreos-tracker/issues/1814

Use type `c3-standard-4` with only 1 disk that uses the NVMe interface, as Intel TDX `c3-standard-4-lssd` is not supported yet. (Refer to https://cloud.google.com/confidential-computing/confidential-vm/docs/supported-configurations#limitations)


tests/kola/data/commonlib.sh
dustymabe, line 1: hmm. I'm thinking if these functions aren't useful for any other tests then maybe we should create a new `gcp-lib.sh` for it and store it closer to those tests instead of storing it up here in the highlest level library.

for an example see https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/ntp/data/ntplib.sh and https://github.com/coreos/fedora-coreos-config/blob/d10529174cc29095e1a7a2166294c7c303ace9c7/tests/kola/ntp/chrony/dhcp-propagation#L28

tests/kola/data/commonlib.sh
```diff
@@ -137,3 +137,45 @@ vergt() {
...
+assert_confidential_type_sev_snp() {
+    local cvm_type=$(systemd-detect-virt --cvm)
+    [ "${cvm_type}" == "sev-snp" ]
+}
+
+# Verify the instance is Confidential VM with TDX type
+assert_confidential_type_tdx() {
+    local cvm_type=$(systemd-detect-virt --cvm)
+    [ "${cvm_type}" == "tdx" ]
+}
```
dustymabe: Seems like these functions could just be a single function with an argument:

```
assert_confidential_type_match() {
    local cvm_type=$(systemd-detect-virt --cvm)
    local expected=$1
    [ "${cvm_type}" == "${expected} ]
}

assert_confidential_type_match "sev-snp"
assert_confidential_type_match "tdx"
```

tests/kola/platforms/gcp/confidential-vm-tdx-nvme-symlink
```diff
@@ -0,0 +1,52 @@
...
+    eval $(/usr/lib/udev/google_nvme_id -d "${device}")
+    if [ ! -n "${ID_SERIAL_SHORT:-}" ]; then
+        fatal "can not get nvme ${device} ID_SERIAL_SHORT"
+    fi
+
+    local link="/dev/disk/by-id/google-${ID_SERIAL_SHORT}"
+    if ! ls -l "${link}"; then
+        fatal "can not find ${device} symlink ${link}"
+    fi
+}
```
dustymabe: Can this function be put in the library?

tests/kola/data/commonlib.sh
```diff
@@ -137,3 +137,45 @@ vergt() {
...
+
+# NVME_INFO=$(nvme list-subsys -o json)
+# check instance has 1 nvme disk
+assert_one_nvme_disk() {
```
dustymabe: since these are now inf a different file I'd prefer for `NVME_INFO` to be passed in as an argument.

```
local nvme_info="$1"
```

tests/kola/data/commonlib.sh
HuijingHei: Good suggestion, put `assert_confidential_type_match()` to commonlib.sh as it might be used on other platform like AWS, and others are copied to `gcp-lib.sh`.

tests/kola/data/commonlib.sh
HuijingHei: Updated, thanks!

tests/kola/data/commonlib.sh
```diff
@@ -137,3 +137,10 @@ vergt() {
 vergte() {
     vereq "$1" "$2" || vergt "$1" "$2"
 }
+
+# Verify the instance is Confidential VM type that matches expected
+assert_confidential_type_match() {
+    local cvm_type=$(systemd-detect-virt --cvm)
+    local expected=$1
+    [ "${cvm_type}" == "${expected}" ]
+}
```
dustymabe, line 146: I assume you left this here because it would be useful outside of GCP confidential compute tests?

dustymabe (approved):
LGTM

tests/kola/data/commonlib.sh
HuijingHei: Yes, it might be also useful if running Confidential VM on other platform, for example Azure.

PR #3362, opened 2025-02-17
[branched]: Enable 'erofs' by default for the rootfs in the Live ISO/PXE artifacts

nikita-dubrovskii:
(cherry picked from commit a8735437384c9d4484924695f1638db68d8b9887)


dustymabe (approved):
LGTM

PR #3368, opened 2025-02-19
overlay/05core: coreos-ignition dracut: use vendored sgdisk for el10

jbtrystram:
sgdisk is dropped from RHEL10/c10s, so we vendored it in ignition in [1]. Condiontionnaly include in the initramfs the vendored binary for EL10 variants only.

[1] https://gitlab.com/redhat/centos-stream/rpms/ignition/-/commit/6e48bd528cce5e227352fd375f95393322dcfef0

See https://issues.redhat.com/browse/RHEL-56080

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -23,8 +23,17 @@ install() {
         lsblk \
         sed \
         grep \
-        sgdisk \
         uname
+
+
+    # In some cases we had to vendor gdisk in Ignition.
+    # If this is the case here use that one.
+    # See https://issues.redhat.com/browse/RHEL-56080
+    if [ -f /usr/libexec/ignition-sgdisk ]; then
+        inst /usr/libexec/ignition-sgdisk /usr/sbin/sgdisk
+    else
+        inst sgdisk
+    fi
 
     # For IBM SecureExecution
     if [[ $(uname -m) = s390x ]]; then


dustymabe:
I'm trying to figure out why this needs to go here instead of in `openshift/os`. https://github.com/openshift/os/pull/1498 already has a commit that does this.

jbtrystram:
> I'm trying to figure out why this needs to go here instead of in `openshift/os`. [openshift/os#1498](https://github.com/openshift/os/pull/1498) already has a commit that does this.

because the commit there does not work without removing `sgdisk` from here (in overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh ) 

One solution would be to maintain a copy of this whole file in openshift/os (and then I don't even know which one would override the other) so I thought handling this here would be cleaner. 

Maybe there is a better solution i am not aware of though

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -23,9 +23,20 @@ install() {
...
+    # as it's vendored there.
+    # See https://issues.redhat.com/browse/RHEL-56080
+    source /usr/lib/os-release
+    if [ "$PLATFORM_ID" = "platform:el10" ]; then
+        inst /usr/libexec/ignition-sgdisk /usr/sbin/sgdisk
+    else
+        inst sgdisk
+    fi
+
+
```
travier: Whitespace here :)

jbtrystram: fixed

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
```diff
@@ -23,9 +23,19 @@ install() {
...
+    # In RHEL10 / C10s sgdisk got dropped so we get sgdisk from ignition
+    # as it's vendored there.
+    # See https://issues.redhat.com/browse/RHEL-56080
+    source /usr/lib/os-release
+    if [ "$PLATFORM_ID" = "platform:el10" ]; then
+        inst /usr/libexec/ignition-sgdisk /usr/sbin/sgdisk
+    else
+        inst sgdisk
+    fi
```
dustymabe: Maybe cleaner here would be to not do a version check here but rather

```suggestion
    # In some cases we had to vendor gdisk in Ignition. If we detect
    # that case here then use that one.
    # See https://issues.redhat.com/browse/RHEL-56080
    source /usr/lib/os-release
    if [ -f /usr/libexec/ignition-sgdisk ]; then
        inst /usr/libexec/ignition-sgdisk /usr/sbin/sgdisk
    else
        inst sgdisk
    fi
```

dustymabe (approved):
LGTM

travier (approved):

PR #3369, opened 2025-02-19
[rawhide] overrides: pin grub2-2.12-23.fc43

dustymabe:
The new grub2-2.12-24.fc43 seems to cause initramfs unpack failure. https://github.com/coreos/fedora-coreos-tracker/issues/1878


marmijo (approved):
LGTM

PR #3370, opened 2025-02-20
35coreos-live: detect rootfs filesystem type

nikita-dubrovskii:
As we transition from SquashFS to EROFS, the 'Type' field was removed from autogenerated systemd units, which may cause failures in EROFS auto-detection.
This PR detects the rootfs fstype by reading the magic number and generates an environment file (/rootfs.env) for those units.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1877


dustymabe:
We won't need this PR if we [get the util-linux fix in](https://github.com/coreos/fedora-coreos-tracker/issues/1877#issuecomment-2669816544).

nikita-dubrovskii:
Ok, let's keep it open until util-linux fix gets merged, and than close.

jlebon:
I still would rather we did something like this, but yeah as it is, it's also just auto-detecting the format based on magic which is no different than libblkid. Ideally instead, it should directly flow from the canonical filename.

jlebon:
The funny thing here is that we control all the media involved, so it would be trivial actually to just bake in a marker in the live initramfs that makes it clear which format it is. Just like today we have the stream hash file which verifies the `rootfs.img` since we know exactly what to expect.

dustymabe:
Closing this considering https://github.com/coreos/fedora-coreos-tracker/issues/1877#issuecomment-2683414623

PR #3371, opened 2025-02-21
overlay/05core: ignition-ostree dracut: use vendored sgdisk for el10

jbtrystram:
Follow-up of 11cce8c139408d2caf3099b15b195cbf728b1624 See https://github.com/coreos/fedora-coreos-config/pull/3368


HuijingHei (approved):
LGTM

HuijingHei:
Maybe later need some change for https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh#L30 as it directly call `sgdisk`

dustymabe (approved):

PR #3124, opened 2024-08-27
tests/kola: adding the external kola test for luks + multipath case

Poorna-Gottimukkula1:
Adding the external kola test for luks + multipath validation
ref: https://issues.redhat.com/browse/OCPBUGS-13123
https://github.com/coreos/fedora-coreos-tracker/issues/1728

PR #3375, opened 2025-02-24
tests/kola/root-reprovision/luks: don't use `root` label for LUKS device

jlebon:
LUKS devices can have labels, separate from names, and this shows up as the "filesystem label" and e.g. competes for `/dev/disk/by-label/root` for the real root.

The Butane LUKS sugar doesn't do that. The storage docs don't do that. So don't do that in our tests here either.

Likely fixes the testing issues in
https://github.com/coreos/fedora-coreos-config/pull/3124.


dustymabe (approved):
LGTM

PR #3378, opened 2025-02-26
Add the `azure-vm-utils` package to support Azure NVMe Disks

marmijo:
Add the `azure-vm-utils` package to include udev rules to support certain VM types (e.g. Standard_M16bds_v3, Standard_M16bs_v3, Standard_L8s_v4) that present managed disks as NVMe devices instead of traditional SCSI disks. Without these rules, LUN-based disk detection fails in the Azure Disk CSI driver, causing volume mount failures.

Add the azure provided udev rules[1] through the `azure-vm-utils` package. This package was recently built for Fedora 41, so let's also fast-track it in testing-devel to start consuming it sooner.

`[1]`: https://github.com/Azure/azure-vm-utils/blob/6fb0ae1d047c0f06e19aab8f1e27e2cd588a6f9f/udev/80-azure-disk.rules

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,9 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  azure-vm-utils:
+    evr: 0.5.1-1.fc41
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2025-b24d671634
+      type: fast-track
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -168,7 +168,10 @@ packages:
   - iptables-legacy
   # NIC firmware we've traditionally shipped but then were split out of linux-firmware in Fedora
   - qed-firmware # https://github.com/coreos/fedora-coreos-tracker/issues/1746
-
+  # Include udev rules for NVMe backed Azure Instances
+  # see: https://issues.redhat.com/browse/COS-3124
+  # This package is not available in RHCOS yet
+  - azure-vm-utils
 
 # - irqbalance
 #   - This thing is crying out to be pulled into systemd, but that hasn't happened


overlay.d/25azure-udev-rules/usr/lib/dracut/modules.d/25azure-udev-rules/module-setup.sh
```diff
@@ -14,4 +14,5 @@ depends() {
 # called by dracut
 install() {
     inst_rules 68-azure-sriov-nm-unmanaged.rules
+    inst_rules 80-azure-disk.rules
```
dustymabe: is this needed in the >=42 case?

manifests/azure-vm-utils.yaml
```diff
@@ -0,0 +1,5 @@
+# azure-vm-utils provides Azure disk udev rules that support NVMe backed Azure Managed Disks.
+# Certain VM types (e.g. Standard_M16bds_v3, Standard_L8s_v4) present managed disks as NVMe 
+# devices instead of traditional SCSI disks.
+packages:
+  - azure-vm-utils
```
dustymabe: 2 questions:

- do we know what else this package provides beyond the udev rules? Is all of it pretty safe/benign?
- what does it bring in as far as dependencies are concerned?

overlay.d/25azure-udev-rules/usr/lib/udev/rules.d/80-azure-disk.rules
dustymabe, line 1: given our proximity to F42 I think I'd be OK just not including this in <F42 as it makes things simpler.

I know we need to do this downstream for a period of time, but would prefer to just have the rules live in the overlay there and never live here. 

alternatively, we could message the package maintainer to ask if they'd build the RPM for F41.

dustymabe (commented):
Thanks for working on this..

Were we able to reproduce the error case and verify adding the rules here fixes it?

manifests/azure-vm-utils.yaml
marmijo: > do we know what else this package provides beyond the udev rules? Is all of it pretty safe/benign?

It looks like a pretty light package right now, but it's intended to be a home for core utilities, udev rules, and other configuration to support Linux VMs on Azure. Right now the disk udev rules and a utility to discover NVMe devices are included.

> what does it bring in as far as dependencies are concerned?

The package doesn't have any dependencies that I can see:
https://src.fedoraproject.org/rpms/azure-vm-utils/blob/f42/f/azure-vm-utils.spec

I built and tested FCOS with the changes from this PR and `azure-vm-utils` was the only additional package introduced in the build.

overlay.d/25azure-udev-rules/usr/lib/udev/rules.d/80-azure-disk.rules
marmijo: It makes sense to me to just include ~~this~~ the package in >=42. I split out the overlay part into a separate commit in case we wanted to do that. 
I agree that we can just keep the overlay downstream until it's no longer needed there.

> alternatively, we could message the package maintainer to ask if they'd build the RPM for F41.

That'd be fine as well because it would be cleaner to just include the package unconditionally, but it doesn't seem like there's a immediate need for this functionality here so we _could_ just do this in >=42.

overlay.d/25azure-udev-rules/usr/lib/dracut/modules.d/25azure-udev-rules/module-setup.sh
marmijo: No it's not. Let's just drop the overlay altogether and just include the package (see comments on 80-azure-disk.rules)

marmijo:
> Were we able to reproduce the error case and verify adding the rules here fixes it?

I wasn't able to reproduce the exact issue, but I think I can confirm this fixes it. The issue is related to the Azure Disk CSI Driver not being able to perform LUN-based disk discovery without these udev rules. see: https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/2777

I was able to confirm that the udev rules properly map NVMe devices in `/dev/disk/azure` and also sets `AZURE_DISK_TYPE` and `AZURE_DISK_LUN` to their proper values. These values are not set without the udev rules.

```
[core@marmijo-Standard_M16bds_v3 ~]$ udevadm info /dev/nvme0n2
M: nvme0n2
...
... (output truncated)
...
E: AZURE_DISK_TYPE=data
E: AZURE_DISK_LUN=0
```

dustymabe:
> I was able to confirm that the udev rules properly map NVMe devices in `/dev/disk/azure` and also sets `AZURE_DISK_TYPE` and `AZURE_DISK_LUN` to their proper values. These values are not set without the udev rules.

IOW we have enough to write a simple test for this?

overlay.d/25azure-udev-rules/usr/lib/udev/rules.d/80-azure-disk.rules
dustymabe: do you want to send an email to them? address should be able to be lifted from the git log in distgit for the RPM.

marmijo:
> IOW we have enough to write a simple test for this?

Looks like it! :+1:

overlay.d/25azure-udev-rules/usr/lib/udev/rules.d/80-azure-disk.rules
marmijo: Will do!

marmijo: The maintainer responded to my email and added the package in fedora 41: https://bodhi.fedoraproject.org/updates/FEDORA-2025-b24d671634

dustymabe (approved):
LGTM

PR #3381, opened 2025-02-28
tests/manual: trim saved state in coreos-builds-bisect.py

dustymabe:
The statefile that tracks the progress of a bisect will today have all the builds from a given stream in it when really it only needs to keep track of the range of builds between the known good and known bad entries. Let's trim it to just that range of builds.


nikita-dubrovskii (approved):
LGTM

PR #3385, opened 2025-03-03
[next-devel] image: enable deploy-via-container for generated images

dustymabe:
Now that we are moving to deploy updates via OCI we'll also change the images we bake to be "deploy-via-container" as well.

This is part of https://github.com/coreos/fedora-coreos-tracker/issues/1823

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -3,6 +3,11 @@
 # streams.
 include: image-base.yaml
 
+# Deploy via container now
+# https://github.com/coreos/fedora-coreos-tracker/issues/1823
+deploy-via-container: true
+container-imgref: "ostree-remote-registry:fedora:quay.io/fedora/fedora-coreos:next-devel"
+
 # Enable 'erofs' by default for the rootfs in the Live ISO/PXE artifacts
 live-rootfs-fstype: "erofs"
 live-rootfs-fsoptions: "-zlzma,level=6 -Eall-fragments,fragdedupe=inode -C1048576 --quiet"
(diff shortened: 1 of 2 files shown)


image.yaml
```diff
@@ -3,6 +3,11 @@
 # streams.
 include: image-base.yaml
 
+# Deploy via container now
+# https://github.com/coreos/fedora-coreos-tracker/issues/1823
+deploy-via-container: true
+container-imgref: "ostree-remote-registry:fedora:quay.io/fedora/fedora-coreos:next-devel"
```
dustymabe, line 9: Considering https://github.com/coreos/fedora-coreos-config/pull/3355 should we be using this instead:

```suggestion
container-imgref: "ostree-remote-registry:fedora:registry:quay.io/fedora/fedora-coreos:next-devel"
```

jbtrystram:
Duplicate of https://github.com/coreos/fedora-coreos-config/pull/3364 ?

image.yaml
dustymabe: actually [it turns out](https://coreos.github.io/rpm-ostree/container/#url-format-for-ostree-native-containers) `ostree-remote-registry:fedora:` is equivalent to `ostree-remote-image:fedora:registry:` so the shorter form is actually better.

dustymabe:
> Duplicate of #3364 ?

It is. Do you want to update it to the [shorter form](https://github.com/coreos/fedora-coreos-config/pull/3385#discussion_r1978199470) and close this?

marmijo:
This can be closed now. https://github.com/coreos/fedora-coreos-config/pull/3396 implemented the change in `next-devel`

PR #3394, opened 2025-03-12
overrides.ppc64le: pin grub 2 to 1:2.12-15.fc41

jbtrystram:
See https://github.com/coreos/fedora-coreos-tracker/issues/1886


jbtrystram:
Tested this in a debug pod build and it does not fix the test.

PR #3399, opened 2025-03-17
overlay.d/40grub: remove 70_coreos-user.cfg

champtar:
bootupd sources user.cfg since before it allows configs.d https://github.com/coreos/bootupd/commit/1475f2587d7fb652f21b07f8defb4581d6c50754


HuijingHei (approved):
/lgtm

cgwalters:
Thanks so much for all the work on this!!

I guess the main thing that could probably use some human validation is that the instructions for setting a GRUB password still work. But AFAICS your analysis is correct.

HuijingHei:
Verify [GRUB password](https://docs.fedoraproject.org/en-US/fedora-coreos/grub-password/) works with latest fcos that includes this patch (without `source $prefix/70_coreos-user.cfg` in the grub.cfg)

PR #3406, opened 2025-03-19
overlay.d/40grub: rename 40_coreos-ignition.cfg -> 05_ignition.cfg

champtar:
ignition grub config needs to be before blscfg as it sets the ignition_firstboot variable that is part of the kargs.

To help align bootupd config ordering with legacy grub config, rename to 05_ignition.cfg which is before 10_blscfg.cfg.


dustymabe:
hold this - i'm looking into it.

dustymabe:
we should probably do something like this instead:

- https://github.com/coreos/fedora-coreos-config/pull/3407 


and if we need to update the name of the static grub config fragment we should do it [upstream](https://github.com/coreos/ignition/blob/fc324e12230b036ce9d44f64346780121431ff27/grub2/ignition.cfg)

champtar:
> we should probably do something like this instead:
> 
>     * [overlay.d: drop 40grub overlay #3407](https://github.com/coreos/fedora-coreos-config/pull/3407)
> 
> 
> and if we need to update the name of the static grub config fragment we should do it [upstream](https://github.com/coreos/ignition/blob/fc324e12230b036ce9d44f64346780121431ff27/grub2/ignition.cfg)

PR already created upstream: https://github.com/coreos/ignition/pull/2037
Yes this file should be deleted, but it might take a good amount of time for the fixed ignition package to be available

dustymabe:
> PR already created upstream: [coreos/ignition#2037](https://github.com/coreos/ignition/pull/2037)

Nice.

> Yes this file should be deleted, but it might take a good amount of time for the fixed ignition package to be available

Hmm IMO I think we should just do backports and get it in now. otherwise we'll forget.

prestist:
> Hmm IMO I think we should just do backports and get it in now. otherwise we'll forget.

Agreed, we can add a patch to the packaging pr [here](https://src.fedoraproject.org/rpms/ignition/pull-request/132#)

champtar:
Fine with me, https://github.com/coreos/ignition/pull/2037 is ready anyway

dustymabe:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/3407

PR #3407, opened 2025-03-19
overlay.d: drop 40grub overlay

dustymabe:
The user.cfg was already dropped in f9a2278 and the Ignition config we were providing here was only meant to be temporary since Ignition is now shipping the required config as part of the `ignition-grub` subpackage. Let's include the subpackage and stop shipping any static configs here, which means we can drop the overlay too.


HuijingHei:
Before we drop this, might make sure the downstream ignition include the [change](https://src.fedoraproject.org/rpms/ignition/pull-request/132) which is needed by https://github.com/openshift/os/blob/master/overlay.d/40grub ?

dustymabe:
Indeed. we'll make sure that is the case before we merge this

dustymabe:
This is now ready for review

champtar (approved):
LGTM

prestist (approved):
LGTM, thank you for doing this!

PR #3408, opened 2025-03-19
denylist: skip console warnings on ppc

dustymabe:
We are getting a bunch of kernel warnings on ppc64le and it's blocking us from getting new updates for all architectures (i.e. bump-lockfile won't pass). Let's skip console warnings for now on ppc64le.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1889

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -6,3 +6,7 @@
   warn: true
   arches:
     - ppc64le
+- pattern: skip-console-warnings
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1889
+  arches:
+    - ppc64le


marmijo (approved):
I didn't know about `skip-console-warnings` until today!
LGTM, but should a snooze or `warn:true` be added to this?

dustymabe:
> but should a snooze or `warn:true` be added to this?

Looking at the [code for this](https://github.com/coreos/coreos-assembler/blob/e61093c60b4c4ec274b603d2ed54849e9a9c6be1/mantle/kola/harness.go#L454-L457) the loop gets `continue`d before snooze or warn is processed, so I don't think so.

PR #3413, opened 2025-03-23
[rawhide] overrides: pin dracut-103-3.fc42

dustymabe:
This new dracut-105-2.fc43 update is causing multipath tests to fail. See https://github.com/coreos/fedora-coreos-tracker/issues/1898


prestist (approved):
LGTM

PR #3425, opened 2025-03-31
tests/kola/ignition: add azure blob tests

prestist:
Test new Azure priv fetching support in Ignition 2.21.0.

Used the gcp platform tests as a skeleton fits pretty well..

Added the fixtures to the fedora coreos subscription


prestist:
Tested by making a local fcos image and uploading it t azure. Test passes when using the new --azure-credentials argument for cosa kola.

marmijo (approved):
Thanks for working on this! LGTM

dustymabe (approved):
LGTM

We now have three test.sh files that are mostly the same content. Might be a good followup sometime to move that boilerplate into a library.

PR #3428, opened 2025-04-01
ci/overrides.py: switch to libdnf5

nikita-dubrovskii:
F41 switched to dnf5, which has different Python API.
This PR breaks dependency on unsupported `hawkey Python bindings` and `python3-dnf`.

Issue: https://github.com/coreos/fedora-coreos-config/issues/3254

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -51,34 +51,36 @@ def main():
     parser = argparse.ArgumentParser(description='Manage overrides.')
     # "dest" to work around https://bugs.python.org/issue29298
     subcommands = parser.add_subparsers(title='subcommands', required=True,
-            dest='command')
+                                        dest='command')
 
-    fast_track = subcommands.add_parser('fast-track',
-            description='Fast-track Bodhi updates.')
+    fast_track = subcommands.add_parser(
+        'fast-track', description='Fast-track Bodhi updates.')
     fast_track.add_argument('update', nargs='+',
-            help='ID or URL of Bodhi update to fast-track')
-    fast_track.add_argument('-r', '--reason',
-            help='URL explaining the reason for the fast-track')
+                            help='ID or URL of Bodhi update to fast-track')
+    fast_track.add_argument(
+        '-r',
+        '--reason',
+        help='URL explaining the reason for the fast-track')
     fast_track.add_argument('--ignore-dist-mismatch', action='store_true',
-            help='ignore mismatched Fedora major version')
+                            help='ignore mismatched Fedora major version')
     fast_track.set_defaults(func=do_fast_track)
 
     pin = subcommands.add_parser('pin', description='Pin source RPMs.')
     pin.add_argument('nvr', nargs='+',
-            help='NVR of SRPM to pin')
+                     help='NVR of SRPM to pin')
     pin.add_argument('-r', '--reason', required=True,
-            help='URL explaining the reason for the pin')
+                     help='URL explaining the reason for the pin')
     pin.add_argument('--ignore-dist-mismatch', action='store_true',
-            help='ignore mismatched Fedora major version')
+                     help='ignore mismatched Fedora major version')
     pin.set_defaults(func=do_pin)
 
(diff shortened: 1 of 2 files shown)


ci/overrides.py
```diff
@@ -326,18 +326,22 @@ def graduate_lockfile(base, fn):
...
+                                     arch=nevra.get_arch()).latest().run()
 
     if len(pkgs) == 0:
         # Odd... the only way I can imagine this happen is if we fast-track a
         # brand new package from Koji which hasn't hit the updates repo yet.
         # Corner-case, but let's be nice.
-        eprint(f"couldn't find package {nevra.name}; assuming not graduated")
+        eprint(f"couldn't find package {nevra.get_name()}; assuming not graduated")
         return False
+    nevra = (nevra.get_version(), nevra.get_release())
```
dustymabe: we re-use the variable here. how is it being changed?

Might be more clear if we use two different variable names to highlight the difference.

dustymabe: i.e. this could be:

```suggestion
    nevra_tuple = (nevra.get_version(), nevra.get_release())
```

dustymabe (commented):
Thanks for working on this!

Does it run without any dnf 4 packages installed?

nikita-dubrovskii:
> Thanks for working on this!
> 
> Does it run without any dnf 4 packages installed?

I don't know how to test it properly. So i've only tested it by re-using this code in dummy unittest locally. I suspect, that without dnf4 it would fail: we use : 
```
import dnf  

def get_dnf_base(treefile):
    base = dnf.Base()
```

ci/overrides.py
```diff
@@ -326,19 +326,23 @@ def graduate_lockfile(base, fn):
...
+        eprint(f"couldn't find package {nevra.get_name()}; assuming not graduated")
         return False
+    nevra_tuple = (nevra.get_version(), nevra.get_release())
 
-    nevra_latest = hawkey.split_nevra(str(pkgs[0]))
-    return nevra_latest >= nevra
+    nevra_latest = rpm.Nevra.parse(
+        str(pkgs[0]), rpm.VectorNevraForm(1, rpm.Nevra.Form_NEVRA))[0]
+    nevra_latest_tuple = (nevra_latest.get_version(), nevra_latest.get_release())
+    return nevra_latest_tuple >= nevra_tuple
```
jlebon: Hmm, this is very subtle. Are you sure comparison works as we expect here? RPM version comparison and not string comparison.

I think likely what we want here is just 

```
return nevra_latest >= nevra
```

or did that not work?

jlebon (commented):
We should also revert https://github.com/coreos/fedora-coreos-config/pull/3255 as part of this.

jlebon:
Also, can you add a commit message with context and e.g. link to the issue? (In general, it should be rare that a commit does not deserve a message.)

ci/overrides.py
nikita-dubrovskii: Nope, `lindnf5` is low-level wrapper around dnf5 C++ code.

nikita-dubrovskii:
> We should also revert #3255 as part of this.

Ok, than `def get_dnf_base(treefile):` requires rework as well:

```
$ rpm -qf /usr/lib/python3.13/site-packages/dnf/__init__.py
python3-dnf-4.23.0-1.fc41.noarch
```

Will add more commits to this PR

dustymabe:
Wow. This turned out to be more work than I thought it would be. Thanks @nikita-dubrovskii for working on it.

dustymabe (approved):
There's a lot in here and I mostly skimmed this but if you tested it I'm willing to go with it.

Do we need to hold this for any reason until say FCOS has switched to F42?

jlebon (approved):
Nice work!

dustymabe:
@nikita-dubrovskii merge when ready

nikita-dubrovskii:
> @nikita-dubrovskii merge when ready

I know on Friday it's a bad habit to merge, but let's do it )

PR #3433, opened 2025-04-04
manifests: Drop 'fips-mode-setup' usage for EL10

### LINKED ISSUE #1665: Drop `40rhcos-fips` dracut module
See https://fedoraproject.org/wiki/Changes/RemoveFipsModeSetup. The idea there is that the `fips=1` karg becomes the source of truth. This is something that should work in image-mode RHEL/CentOS Stream already and we just inherit from that. On the MCO side, long-term on-cluster layering should do the equivalent of https://gitlab.com/fedora/bootc/examples/-/blob/main/fips/Containerfile?ref_type=heads to enable it at layering time. Short-term, at the very least, we should be able to get rid of [this bit](https://github.com/openshift/os/blob/b65935f461493fd3315edbfcc72e01ff43097790/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh#L67-L113) since calling `fips-mode-setup` should no longer be necessary. See also conversation in https://github.com/coreos/ignition/issues/1323.

joelcapitao:
The 'fips-mode-setup' script have been removed in EL10 since [1][2], but is still needed for EL9.
So we are dropping its usage in a postprocess script for EL10 only.

Partially closes https://github.com/openshift/os/issues/1665

[1] https://gitlab.com/redhat/centos-stream/rpms/crypto-policies/-/commit/67e22dbc3721d1d17505bff85228b465cd5ca225
[2] https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/commit/ff7551bbb3011f17cb8c6aac03f64682dde14c21

### DIFF
--- a/manifests/shared-el10.yaml
+++ b/manifests/shared-el10.yaml
@@ -6,3 +6,13 @@ packages:
   # These should be with kexec-tools in the user-experience manifest
   - makedumpfile
   - kdump-utils
+
+postprocess:
+  # The 'fips-mode-setup' script have been removed in EL10 since [1][2], but is still needed
+  # for EL9, so we drop its usage here.
+  # [1] https://gitlab.com/redhat/centos-stream/rpms/crypto-policies/-/commit/67e22dbc3721d1d17505bff85228b465cd5ca225
+  # [2] https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/commit/ff7551bbb3011f17cb8c6aac03f64682dde14c21
+  - |
+    #!/usr/bin/env bash
+    set -euxo pipefail
+    sed -i "/.fips-mode-setup --enable.*/d" /usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh


dustymabe:
Not a big fan of this change as is. I assume something like this done inside the script would also work?


```patch
diff --git a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
index ac91292..923757b 100755
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
@@ -69,6 +69,13 @@ finish() {
         fatal "FIPS mode is not enabled."
     fi
 
+    # on EL10 fips-mode-setup was removed and these steps are
+    # no longer needed. Please delete this function and sysroot_bwrap
+    # when EL9 is no longer supported.
+    if [ ! -e /sysroot/usr/bin/fips-mode-setup ]; then
+        return 0
+    fi
+
     # If we're running from a live system, then set things up so that the dracut fips
     # module will find the kernel binary.  TODO change dracut to look in /usr/lib/modules/$(uname -r)
     # directly.

```

dustymabe:
also.. IIUC this PR should be against the `openshift/os` repo.

joelcapitao:
> Not a big fan of this change as is. I assume something like this done inside the script would also work?
> 
> ```diff
> diff --git a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
> index ac91292..923757b 100755
> --- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
> +++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
> @@ -69,6 +69,13 @@ finish() {
>          fatal "FIPS mode is not enabled."
>      fi
>  
> +    # on EL10 fips-mode-setup was removed and these steps are
> +    # no longer needed. Please delete this function and sysroot_bwrap
> +    # when EL9 is no longer supported.
> +    if [ ! -e /sysroot/usr/bin/fips-mode-setup ]; then
> +        return 0
> +    fi
> +
>      # If we're running from a live system, then set things up so that the dracut fips
>      # module will find the kernel binary.  TODO change dracut to look in /usr/lib/modules/$(uname -r)
>      # directly.
> ```

Yeah, this approach sounds better, I was not satisfied by the postscript TBH.
Let me put your code in https://github.com/openshift/os/pull/1498 and close this one

PR #3380, opened 2025-02-27
tests/networking: update `nm-ifcfg-rh-plugin` and `team-dhcp-via-ignition`  working on el9

HuijingHei:
See https://github.com/openshift/os/pull/1498

PR #3434, opened 2025-04-04
exttests/team-dhcp-via-ignition and nm-ifcg: disable for RHEL 10

jbtrystram:
Teamd is no longer supported in rhel 10.
Also add a helper for rhel10.

See https://github.com/coreos/fedora-coreos-tracker/issues/1727

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -80,6 +80,13 @@ is_rhcos9() {
     { [ "${ID}" == "rhel" ] && [ "${VERSION_ID%%.*}" -eq 9 ]; }
 }
 
+# rhcos10 && SCOS 10
+is_el10() {
+    source /etc/os-release
+    { [ "${ID}" == "rhel" ] && [ "${VERSION_ID%%.*}" -eq 10 ]; } || \
+    { [ "${ID}" == "centos" ] && [ "${VERSION_ID%%}" -eq 10 ]; }
+}
+
 # scos
 is_scos() {
     source /etc/os-release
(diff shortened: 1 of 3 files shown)


dustymabe (dismissed):
LGTM

tests/kola/data/commonlib.sh
```diff
@@ -80,6 +80,13 @@ is_rhcos9() {
     { [ "${ID}" == "rhel" ] && [ "${VERSION_ID%%.*}" -eq 9 ]; }
 }
 
+# rhcos10 && SCOS 10
+is_el10() {
+    source /etc/os-release
+    { [ "${ID}" == "rhcos" ] && [ "${RHEL_VERSION%%.*}" -eq 10 ]; } || \
```
jlebon: This needs to be

```suggestion
    { [ "${ID}" == "rhel" ] && [ "${VERSION_ID%%.*}" -eq 10 ]; } || \
```

We don't modify `/etc/os-release` anymore in the new model. (See https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/split-rhcos-into-layers.md#etcos-release).

tests/kola/data/commonlib.sh
```diff
@@ -80,6 +80,13 @@ is_rhcos9() {
...
+    { [ "${ID}" == "centos" ] && [ "${VERSION_ID%%}" -eq 10 ]; }
```
jlebon, line 87: Hmm, though note this is inconsistent with `is_rhcos9` which doesn't check the scos/centos case.

jbtrystram: yeah that's why i named it `el10`

tests/kola/data/commonlib.sh
jbtrystram: ah yes, thanks

jbtrystram:
While i was there I also updated the nm-ifcfg test as it's removed citing the same teamd ticket (see network-manager commit)

jbtrystram:
hmm I just saw https://github.com/coreos/fedora-coreos-config/pull/3380 which is definitely fancier.

PR #3435, opened 2025-04-04
workflows/add-overrides: install python3-libdnf5

dustymabe:
We need the libdnf5 python3 library for ab42e75 and because dnf5 is no longer python based the python3 libraries don't get installed by default so we need to name them here.


marmijo (approved):
LGTM

aaradhak (approved):

dustymabe:

PR #3439, opened 2025-04-07
tests: drop some cruft from the net behavior change test

dustymabe:
We don't need to build or test for RHEL8 or < Fedora 37 any longer so drop the conditionals here for those and also clean up the list of config variations at the top.


HuijingHei (approved):
LGTM

PR #3440, opened 2025-04-07
workflows/remove-graduated-overrides: Install python3-libdnf5

aaradhak:
The github [action](https://github.com/coreos/fedora-coreos-config/actions/runs/14316272652/job/40123032450) 'Remove Graduates Overrides' has been failing due to missing libdnf5 pkg. The installation of libdnf5 pkg would address this failure. The python3 libraries don't get installed by default so we need to name them here.


dustymabe (approved):

dustymabe:
skipping CI since it won't test this.

PR #3442, opened 2025-04-07
[rawhide] lockfiles: drop graduated overrides 🎓

coreosbot-releng:


dustymabe (approved):

dustymabe:
The CI failure here is because a warning was output. I opened https://github.com/coreos/fedora-coreos-config/pull/3444 to exnted the snooze on that test.

Merging this since it otherwise passed tests.

PR #3444, opened 2025-04-08
denylist: extend snooze for toolbox

dustymabe:
We are waiting for a fixed kernel.
https://github.com/coreos/fedora-coreos-tracker/issues/1908


aaradhak (approved):
LGTM

PR #3450, opened 2025-04-11
ci/overrides.py: don't fail with `KeyError` when manifest has no `repos`

nikita-dubrovskii:
I've noticed, that `manifest.yaml` in some [branches](https://github.com/coreos/fedora-coreos-config/blob/next/manifest.yaml) doesn't contain the `repos` list, which leads to an error:
```
  File "fedora-coreos-config/./ci/overrides.py", line 295, in setup_repos
    for repo in treefile['repos']:                                                                                                                             
                ~~~~~~~~^^^^^^^^^    
KeyError: 'repos'       
```

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -292,7 +292,7 @@ def setup_repos(base, treefile):
         repo.disable()
 
     eprint("Enabled repos:")
-    for repo in treefile['repos']:
+    for repo in treefile.get('repos', []):
         query = libdnf5.repo.RepoQuery(base)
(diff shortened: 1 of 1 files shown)


dustymabe:
Interesting. How did you discover this bug?

dustymabe (approved):

nikita-dubrovskii:
> Interesting. How did you discover this bug?

Was missing `c++`, so today was hacking on `dnf5`. And during tests on different branches discovered this

PR #3451, opened 2025-04-11
image.yaml: switch to EROFS; deploying via container

dustymabe:
These changes are ones that we are adopting for F42+ and since F42 GA is a GO for Tuesday 04/15 we'll go ahead and make the change for `testing-devel` now.

- https://github.com/coreos/fedora-coreos-tracker/issues/1823
- https://github.com/coreos/fedora-coreos-tracker/issues/1852


dustymabe:
I'm going to include this in a bigger "switching to F42" PR. Closing

PR #3452, opened 2025-04-11
Move to Fedora Linux 42

dustymabe:
This also has some F42 related changes in it and also a bunch of cleanup. See individual commit messages.

Also note CI won't pass on this right now because F42 isn't officially released which means the repos aren't populated yet. So this won't be able to merge until Tuesday.


cgwalters (commented):
LGTM and thanks for doing this!

dustymabe:
this should pass CI now that F42 has shipped - marking as ready for review!

marmijo (approved):
LGTM

image-base.yaml
```diff
@@ -33,3 +33,12 @@ platform-compressor:
     # For DigitalOcean you can upload a qcow2 compressed in either gzip or bzip2
     # https://docs.digitalocean.com/products/custom-images/how-to/upload/
     digitalocean: gzip
+
+# Move to use OCI images by default
+# https://github.com/coreos/fedora-coreos-tracker/issues/1823
+deploy-via-container: true
+container-imgref: "ostree-remote-registry:fedora:quay.io/fedora/fedora-coreos:{stream}"
```
aaradhak, line 40: I tried checking the link quay.io/fedora/fedora-coreos:{stream} with different streams and it was directing to the same page. Just wanted to confirm if that is what is expected here

aaradhak:
LGTM otherwise

image-base.yaml
dustymabe: `:{stream}` is the tag inside the OCI registry repo. For example we have `:stable`, `:testing`, `:next`, etc..

PR #3453, opened 2025-04-11
denylist: add `skip-console-warnings` to openstack

marmijo:
Add the pattern to openstack to ignore the kernel warnings we've been seeing on that platform: https://github.com/coreos/fedora-coreos-tracker/issues/1924

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -16,3 +16,9 @@
   warn: true
   streams:
     - rawhide
+- pattern: skip-console-warnings
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1924
+  platforms:
+    - openstack
+  streams:
+    - rawhide


dustymabe:
Interesting.. I didn't know (or did know at one point and forgot) we can limit denials to a specific platform. That's great.

Should we limit it to a specific arch and also a specific stream too?

dustymabe:
We should also spot check [the code](https://github.com/coreos/coreos-assembler/blob/677fe262a8c5edfb2437f53acda9179a7bdac6eb/mantle/kola/harness.go#L429-L452) to make sure it will actually do what we want here.

marmijo:
From the source code, it looks like it should work for a single platform.

> Should we limit it to a specific arch and also a specific stream too?

Good catch, I have seen only the `ostree.hotfix` test fail this way on `aarch64`, so I'll leave it arch independent for now.  I'll update the stream though.

dustymabe (approved):
LGTM

dustymabe:
CI won't test this change.

PR #3459, opened 2025-04-15
[next-devel] image.yaml: drop OCI and erofs customizations

dustymabe:
These are now part of image-base.yaml [1] so we can drop the settings here.

[1] https://github.com/coreos/fedora-coreos-config/commit/5d3d7deac2216c81dbafdb54622caddf84931f81


aaradhak (approved):
LGTM

PR #3461, opened 2025-04-16
OCPEDGE-1752: [TNF] Add podman-etcd resource agent to the expected broken symlinks in RHCOS

jaypoulz:
In https://github.com/openshift/os/pull/1796 we introduce a symbolic link for the purpose of allowing a pacemaker resource agent to be introduced into OCP via the MCO when installing Two Node OpenShift with Fencing (TNF). The kola test that verifies symlinks fails, so we overrode this test entirely for those builds. The patch updates the test to be more granular so that we can reenable it in RHCOS builds until the symlink is removed with the completion of https://issues.redhat.com/browse/OCPEDGE-1706.


jlebon (approved):
Thanks!

PR #3464, opened 2025-04-17
ci/overrides.py: Simplify NEVRA comparison logic

nikita-dubrovskii:
libdnf5 Python bindings now provide `cmp_nevra` fucntion, this patch uses it to simplify the existing comparison logic.

Issue: https://github.com/coreos/fedora-coreos-config/issues/3465


jlebon (approved):
Nice, LGTM! I didn't check but feel free to merge if the needed dnf version has made it into stable.

PR #3473, opened 2025-04-22
denylist: snooze root-reprovision on rawhide for ppc64le

c4rt0:
A number of root-reprovision tests is failing now on rawhide. Let's snooze them until the issue is resolved.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1926


dustymabe (approved):
LGTM

PR #3475, opened 2025-04-22
coreos-boot-mount-generator: honor `boot=` karg better in multipath case

jlebon:
As described in openshift/os#1787, we need to
loosen up our restriction that only a single `boot` and `root` labeled
filesystem exists, at least past the first boot to start.

There is code already that injects `boot` and `root` kargs to specify
the exact UUID we want so this is feasible.

There's currently two ways in which the boot device can be specified:
- the `/run/coreos/bootfs_uuid` file exists; this is written by rdcore
  but is only available on first boot
- a `boot=` karg, which supports the "dracut convention" (i.e.
  `boot=LABEL=` or `boot=UUID=`, etc.), exists; this is usually written
  by rdcore but is only available on subsequent boots

If neither of those are specified, then we just default to
`by-label/boot`.

But currently, if multipath is enabled, we override all that and just
always use `by-label/dm-mpath-boot`. This is not great, because a UUID
might've been specified, and that's much stronger than a label. It also
directly conflicts with the mentioned higher level goal of supporting
multiple `boot` labels.

Instead, rework multipath support so that it respects both `boot=` kargs
and `bootfs_uuid` files. When possible, we use the multipathed version
of the specified UUID/label written by our udev rules because it's less
racy. Otherwise, we just use the non-multipathed one.

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
+++ b/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
@@ -61,21 +61,26 @@ EOF
     add_wants "${unit_name}"
 }
 
-# Copied from
+# Copied and adapted from
 # https://github.com/dracutdevs/dracut/blob/9491e599282d0d6bb12063eddbd192c0d2ce8acf/modules.d/99base/dracut-lib.sh#L586
 # rather than sourcing it.
 label_uuid_to_dev() {
-    local _dev
+    local _dev mpath prefix
     _dev="${1#block:}"
+    mpath=$(karg rd.multipath)
+    prefix=
+    if [ -n "${mpath}" ] && [ "${mpath}" != 0 ]; then
+        prefix=dm-mpath-
+    fi
     case "$_dev" in
         LABEL=*)
-            echo "/dev/disk/by-label/$(echo "${_dev#LABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
+            echo "/dev/disk/by-label/${prefix}$(echo "${_dev#LABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
             ;;
         PARTLABEL=*)
             echo "/dev/disk/by-partlabel/$(echo "${_dev#PARTLABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
             ;;
         UUID=*)
-            echo "/dev/disk/by-uuid/$(echo "${_dev#UUID=}" | tr "[:upper:]" "[:lower:]")"
+            echo "/dev/disk/by-uuid/${prefix}$(echo "${_dev#UUID=}" | tr "[:upper:]" "[:lower:]")"
             ;;
         PARTUUID=*)
             echo "/dev/disk/by-partuuid/$(echo "${_dev#PARTUUID=}" | tr "[:upper:]" "[:lower:]")"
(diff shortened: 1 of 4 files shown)


tests/kola/multipath/test.sh
dustymabe, line 1: how does this test overlap with: 

```
[dustymabe@media fcos]$ cosa kola list | grep -P 'multipath|mpath'
ext.config.root-reprovision.luks.multipath                              [qemu]                                          [all]           [all]           [external reprovision]
multipath.day1                                                          [qemu]                                          [all]           [all]           []
multipath.day2                                                          [qemu]                                          [all]           [all]           []
multipath.partition
```

could we name this test something more specific?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -61,47 +61,56 @@ EOF
...
     _dev="${1#block:}"
+    mpath=$(karg rd.multipath)
+    prefix=
+    if [ -n "${mpath}" ] && [ "${mpath}" != 0 ]; then
+        prefix=dm-mpath-
+    fi
     case "$_dev" in
         LABEL=*)
-            echo "/dev/disk/by-label/$(echo "${_dev#LABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
+            echo "/dev/disk/by-label/$prefix$(echo "${_dev#LABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
```
dustymabe: my OCD wants:

```suggestion
            echo "/dev/disk/by-label/${prefix}$(echo "${_dev#LABEL=}" | sed 's,/,\\x2f,g;s, ,\\x20,g')"
```

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -61,47 +61,56 @@ EOF
...
             bootdev="$(label_uuid_to_dev "$bootkarg")";;
         /dev/*) bootdev=$bootkarg;;
         *) echo "Unknown boot karg '${bootkarg}'; falling back to ${bootdev}";;
     esac
+elif [ -n "${mpath}" ] && [ "${mpath}" != 0 ]; then
+    bootdev=/dev/disk/by-label/dm-mpath-boot
+    # This is used for the first boot only
+    if [ -f /run/coreos/bootfs_uuid ]; then
+        bootdev=/dev/disk/by-uuid/dm-mpath-$(cat /run/coreos/bootfs_uuid)
+    fi
```
dustymabe, line 113: so basically we're just making it so that on subsequent boots (since on subsequent boots `boot=UUID=` should exist), we'll go ahead and rely on the UUID from there, so we decrease the priority of this path to lower than if bootkarg exists?

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
```diff
@@ -61,47 +61,56 @@ EOF
...
 # https://github.com/dracutdevs/dracut/blob/9491e599282d0d6bb12063eddbd192c0d2ce8acf/modules.d/99base/dracut-lib.sh#L586
 # rather than sourcing it.
 label_uuid_to_dev() {
-    local _dev
+    local _dev mpath prefix
     _dev="${1#block:}"
+    mpath=$(karg rd.multipath)
+    prefix=
+    if [ -n "${mpath}" ] && [ "${mpath}" != 0 ]; then
+        prefix=dm-mpath-
```
dustymabe, line 73: I'm guessing the thing that prefixes the `/dev/disk/by-` symlinks with `dm-mpath-` isn't something we control, right?

dustymabe (dismissed):
Some suggestions. Overall LGTM.

tests/kola/multipath/test.sh
jlebon: Good call. Renamed! This test uses the new-ish `:mpath` and `primaryDisk` notations, which means actually all those tests can be rewritten as external tests.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
jlebon: Yeah, exactly. And also even on first boot, we use the bootfs_uuid file rather than ignore it.

overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator
jlebon: It is (it's the referenced 90-coreos-device-mapper.rules file). But we can't really change it now because it's basically an API.

dustymabe: ok. I was just trying to understand if it was a coreos-ism or if it was generic.

dustymabe (dismissed):

dustymabe:
The `test -d /boot/ostree` test on second boot seems to be failing in CI.

jlebon:
> The `test -d /boot/ostree` test on second boot seems to be failing in CI.

OK weird, it looks like that's failing because `/boot` isn't mounted. I thought it was because the kola service was somehow racing with `boot.mount`, but no, for some weird reason `/boot` is getting mounted and then _unmounted_...

jlebon:
OK, I still think this is somehow a race between the kola service and `boot.mount`. Let's see if the latest push helps.

tests/kola/multipath/resilient/test.sh
```diff
@@ -0,0 +1,47 @@
...
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+# we may race with the /boot mount
+if ! is_service_active boot.mount; then
+  fatal "boot.mount failed to activate"
+fi
```
dustymabe, line 19: Potentially a little more direct:

```suggestion
systemd-run --wait --property=After=boot.mount \
    echo "Waited for boot.mount"      
```

dustymabe (approved):

PR #3476, opened 2025-04-22
some riscv enablement 

dustymabe:
There is an effort to build all of Fedora using the RISCV architecture.
This comment adds the yum repo files and also conditionalizes some
manifest items to make it possible to build for `riscv64`.

There are still some other pieces that need to land before builds
will succeed (like bootupd support for riscv64), but this is a start
in the right direction.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1931

---

inline the 16disable-zincati overlay

Now that we have support for inlining manifests [1] in
conditional-include blocks we can drop the 16disable-zincati
and just write out the file in a postprocess.

This just makes it so there's much less scaffolding that
needs to be in place for something simple like this.

[1] https://github.com/coreos/rpm-ostree/pull/5351

### DIFF
--- a/manifests/user-experience.yaml
+++ b/manifests/user-experience.yaml
@@ -20,17 +20,6 @@ packages:
   # Improved MOTD experience
   - console-login-helper-messages-issuegen
   - console-login-helper-messages-profile
-  # kdump support
-  # https://github.com/coreos/fedora-coreos-tracker/issues/622
-  # The makedumpfile and kdump-utils RPMs were broken out in
-  # Fedora and EL10+. To be able to use the same package list
-  # Across EL9 + Fedora + EL10 let's just name paths for now.
-  # We can go back to just specifying the RPM names when we
-  # no longer support EL9.
-  - kexec-tools
-  - /usr/share/makedumpfile # makedumpfile RPM
-  - /usr/bin/kdumpctl       # kdump-utils RPM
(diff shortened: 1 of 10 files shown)


manifests/user-experience.yaml
```diff
@@ -22,7 +22,15 @@ packages:
   - console-login-helper-messages-profile
   # kdump support
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
+  # The makedumpfile and kdump-utils RPMs were broken out in
+  # Fedora and EL10+. To be able to use the same package list
+  # Across EL9 + Fedora + EL10 let's just name paths for now.
+  # We can go back to just specifying the RPM names when we
+  # no longer support EL9.
   - kexec-tools
+  - /usr/share/makedumpfile # makedumpfile RPM
```
jlebon: Hmm, shouldn't this be `/usr/bin`?

dustymabe: We just so happen to be spanning the `/usr/bin/` `/usr/sbin/` merge too, which is why I chose `/usr/share` as the prefix for this one.

Would you like a comment to that effect?

jlebon: It means that we would have to also dowload the filelists, but it looks like we're already doing that today.

jlebon (approved):

PR #3477, opened 2025-04-23
overrides: fast-track moby-engine, containerd

dustymabe:
In Fedora 42 we've had reports of occasional core dumps with Docker and we think we've finally got an update that will fix it.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1918


HuijingHei (approved):
LGTM

PR #3478, opened 2025-04-23
[testing] overrides: fast-track moby-engine, containerd

dustymabe:
In Fedora 42 we've had reports of occasional core dumps with Docker and we think we've finally got an update that will fix it.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1918


dustymabe:
created this PR in case we decide we want to spin a new `testing` at least we won't have to wait on CI.

jlebon (approved):

PR #3486, opened 2025-04-28
denylist: drop toolbox test denial on rawhide

dustymabe:
The fixed kernel shipped in rawhide so we can remove this entry. See https://github.com/coreos/fedora-coreos-tracker/issues/1908#issuecomment-2835865801


marmijo (dismissed):
LGTM

dustymabe:
updated this to extend another denial due to flaky CI. 

can I get another review?

marmijo (approved):

PR #3490, opened 2025-04-30
denylist: extend snooze for kdump.crash*

marmijo:
These tests are still failing. Let's extend the snooze again so we can continue to test the latest RC kernels in rawhide.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1925


dustymabe (approved):
LGTM

PR #3494, opened 2025-05-05
[rawhide] overrides: drop s390x kernel-6.15.0-0.rc2 pin now that rc4 is available

nikita-dubrovskii:
Finally, `kernel-6.15.0-0.rc4.20250502gitebd297a2affa.39.fc43` with all required fixes is available in the rawhide repositories.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1936


jbtrystram (approved):
lgtm

PR #3496, opened 2025-05-05
workflows: bump coreos/rhel-coreos-config submodule

jlebon:
As part of https://github.com/openshift/os/issues/1775, the RHCOS layer is now defined at https://github.com/coreos/rhel-coreos-config.

Tweak the f-c-c submodule bumper workflow to target that repo instead. Keep targeting openshift/os for the rhcos-* branches. Add `rhcos-4.18` while we're here. Finally, rename the workflow to make it less tied to openshift/os.

### DIFF
--- a/.github/workflows/bump-rhcos-submodule.yml
+++ b/.github/workflows/bump-rhcos-submodule.yml
@@ -1,4 +1,4 @@
-name: Sync to openshift/os
+name: Sync to coreos/rhel-coreos-config
 on:
   # We could do push: branches: [testing-devel] but that would restart
   # downstream CI a lot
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-rhcos-submodule.yml
```diff
@@ -14,6 +14,7 @@ on:
         type: choice
         options:
           - testing-devel
+          - rhcos-4.18
```
dustymabe, line 17: we should just delete all of these, though?

syncing to rhel-coreos-config doesn't make sense for the branches below

dustymabe: ahh I see. It does both now.

dustymabe (approved):
LGTM

jlebon:
Follow-up in https://github.com/coreos/fedora-coreos-config/pull/3497.

PR #3498, opened 2025-05-07
[ppc] kola-denylist: extend snooze for root-reprovision.* until 2025-05-12

nikita-dubrovskii:
Original snooze expired, but the kernel fix has not yet been merged, causing the pipeline to break. This commit extends the snooze by one week.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1926


HuijingHei (approved):
LGTM

travier (approved):

PR #3507, opened 2025-05-14
denylist: drop coreos.boot-mirror, extend other snoozes

marmijo:
The coreos.boot-mirror test has been passing since the snooze expired, so we can remove it now and close:
https://github.com/coreos/fedora-coreos-tracker/issues/1932

The kdump.crash and root-reprovision.* tests are still failing, so let's extend the snooze while the issues are investigated.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1925
xref: https://github.com/coreos/fedora-coreos-tracker/issues/1926


dustymabe (approved):
LGTM

PR #3506, opened 2025-05-14
[rhcos-4.18] coreos-boot-mount-generator: honor boot= karg better in multipath case 

joelcapitao:
Backporting https://github.com/coreos/fedora-coreos-config/pull/3475 to the rhcos-4.18 branch.

(cherry picked from commit https://github.com/coreos/fedora-coreos-config/commit/33421b36)
(cherry picked from commit https://github.com/coreos/fedora-coreos-config/commit/0661e4ed)

PR #3508, opened 2025-05-16
manifests/fedora-coreos-base: apply dracut fix in postcript

joelcapitao:
We are awaiting some dracut patches to land in Fedora mirror for [1]. We could wait for it, but we'd need it also for older rhcos branches and those branches won't have the new dracut. So we decided to carry part of the dracut-ng patch [2] in a post-script. This will be backported in older rhcos branches, and will be reverted on testing-devel once new dracut 107 available in mirror.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1937
[2] https://github.com/dracut-ng/dracut-ng/pull/1306

### DIFF
--- a/manifests/shared-el.yaml
+++ b/manifests/shared-el.yaml
@@ -13,3 +13,16 @@ conditional-include:
     include: shared-el10.yaml
   - if: osversion == "centos-10"
     include: shared-el10.yaml
+
+# ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
+# See also the version of this in fedora-coreos.yaml
+postprocess:
+  # We apply part of the patch https://github.com/dracut-ng/dracut-ng/pull/1306 here
+  # while awaiting for the new dracut release to land.
+  # This script will start to fail once dracut 107 release available in repo, signaling
+  # to remove this postscript.
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1937
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    sed -i '/^ExecStart=\/usr\/sbin\/mpathconf --enable$/,${s//ExecStart=\/usr\/sbin\/mpathconf --enable --user_friendly_names n/;b};$q1' /usr/lib/dracut/modules.d/90multipath/multipathd-configure.service
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -16,5 +16,3 @@
     - openstack
   streams:
     - rawhide
-- pattern: ext.config.multipath.resilient
-  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1937
--- a/tests/kola/root-reprovision/luks/multipath/test.sh
+++ b/tests/kola/root-reprovision/luks/multipath/test.sh
@@ -22,7 +22,7 @@ set -xeuo pipefail
 
 # Check if the child device is part of the parent device
 srcdev=$(findmnt -nvr /sysroot -o SOURCE)
-parent_device="/dev/mapper/mpatha"
+parent_device="/dev/mapper/$(multipath -l -v 1)"
 
 if ! lsblk -pno NAME "$parent_device" | grep -qw "$srcdev"; then
     fatal "$srcdev is NOT a child of $parent_device."


dustymabe (changes requested):
Will add review here in a bit.

manifests/fedora-coreos-base.yaml
```diff
@@ -95,6 +95,16 @@ postprocess:
     ConditionPathExists=
     EOF
 
+  # We apply part of the patch https://github.com/dracut-ng/dracut-ng/pull/1306 here
+  # while awaiting for the new dracut release to land.
+  # This script will start to fail once dracut 107 release available in repo, signaling
```
dustymabe: Are we pretty sure of this? i.e. can we test that if we put the updated file in say overrides/rootfs and then try to do a build it will fail?

I'm pretty confident if we were using `patch` it would fail, but not sure about `sed` here.

dustymabe (changes requested):
Sorry for the long wait for the review here. I wanted to put some of the groundwork in place for you to be able to implement my suggestion.

This patch is something we want to carry in FCOS and RHCOS and is also something that we foresee we will drop progressively over time as it enters our various streams (i.e. we'll drop it from `rawhide` first, then `next-devel`/`next`, eventually we'll drop it from el10 and maybe never drop it from el9??

Anyway we need to set ourselves up here for success. Let's move this into a manifest that is shared with el9/10.

I should have made this easier with https://github.com/coreos/fedora-coreos-config/pull/3511 If you don't mind review that and test it out (both an FCOS build and downstream el9/10 build would be useful).

manifests/fedora-coreos-base.yaml
joelcapitao: Good point, I missed a `$`

Below two unit tests:
```
# first one
$ echo "ExecStart=/usr/sbin/mpathconf --enable" > multipathd-configure.service
$ sed -i '/^ExecStart=\/usr\/sbin\/mpathconf --enable$/,${s//ExecStart=\/usr\/sbin\/mpathconf --enable --user_friendly_names n/;b};$q1' multipathd-configure.service ; echo $?
0
$ cat multipathd-configure.service
ExecStart=/usr/sbin/mpathconf --enable --user_friendly_names n

# second one
$ sed -i '/^ExecStart=\/usr\/sbin\/mpathconf --enable$/,${s//ExecStart=\/usr\/sbin\/mpathconf --enable --user_friendly_names n/;b};$q1' multipathd-configure.service ; echo $?
1
$ cat multipathd-configure.service
ExecStart=/usr/sbin/mpathconf --enable --user_friendly_names n
```

I agree the `sed` command is not straightforward

joelcapitao:
> Sorry for the long wait for the review here. I wanted to put some of the groundwork in place for you to be able to implement my suggestion.
> 
> This patch is something we want to carry in FCOS and RHCOS and is also something that we foresee we will drop progressively over time as it enters our various streams (i.e. we'll drop it from `rawhide` first, then `next-devel`/`next`, eventually we'll drop it from el10 and maybe never drop it from el9??
> 
> Anyway we need to set ourselves up here for success. Let's move this into a manifest that is shared with el9/10.
> 
> I should have made this easier with #3511 If you don't mind review that and test it out (both an FCOS build and downstream el9/10 build would be useful).

Understood, I'll rebase my PR on top of #3511 and put the post-script in `shared-el.yaml` manifest

manifests/fedora-coreos-base.yaml
joelcapitao: I gave a shot with `patch`:

```
    #!/usr/bin/env bash
    set -xeuo pipefail
    patch -p1 <<'EOF'
    --- a/usr/lib/dracut/modules.d/90multipath/multipathd-configure.service
    +++ b/usr/lib/dracut/modules.d/90multipath/multipathd-configure.service
    @@ -15,7 +15,7 @@ Type=oneshot
     RemainAfterExit=yes
     # mpathconf requires /etc/multipath to already exist
     ExecStartPre=-/usr/bin/mkdir -p /etc/multipath
    -ExecStart=/usr/sbin/mpathconf --enable
    +ExecStart=/usr/sbin/mpathconf --enable --user_friendly_names n

     [Install]
     WantedBy=sysinit.target
    EOF
```

it's ok locally as ad-hoc script, but not in `cosa build` because `patch` is not available
```
Executing `postprocess` inline script '11'
+ patch -p1
/usr/bin/rpmostree-postprocess-inline-11: line 3: patch: No such file or directory
error: Installing packages: Executing inline postprocessing script: bwrap(/usr/bin/rpmostree-postprocess-inline-11): Child process stopped by signal 0
failed to execute cmd-build: exit status 1
```

so I'm going to keep the `sed` command, except if you want to add `patch` in base image ?

manifests/shared-el.yaml
```diff
@@ -13,3 +13,16 @@ conditional-include:
...
+postprocess:
+  # We apply part of the patch https://github.com/dracut-ng/dracut-ng/pull/1306 here
+  # while awaiting for the new dracut release to land.
+  # This script will start to fail once dracut 107 release available in repo, signaling
+  # to remove this postscript.
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1937
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    sed -i '/^ExecStart=\/usr\/sbin\/mpathconf --enable$/,${s//ExecStart=\/usr\/sbin\/mpathconf --enable --user_friendly_names n/;b};$q1' /usr/lib/dracut/modules.d/90multipath/multipathd-configure.service
```
dustymabe, line 28: Testing this it looks to do what we want, but man is it hard to understand (I partially blame the `sed` man page because it doesn't describe things in the most clear way).

A simple version of the above is:

```
sed -i 's|^ExecStart=/usr/sbin/mpathconf --enable$|ExecStart=/usr/sbin/mpathconf --enable --user_friendly_names n|' /usr/lib/dracut/modules.d/90multipath/multipathd-configure.service
```

but that won't give us the error exit code that we need for this to fail when the patched file comes in.

I think I'll just throw this into the "trust @jcapiitao" bucket.

dustymabe:
Can we add some text to the body of the 2nd commit. A nice description of why we can re-enable this test would be useful.

dustymabe:
CI failures here:

```
[2025-05-20T13:14:34.721Z] --- FAIL: multipath.day2 (84.86s)
[2025-05-20T13:14:34.721Z]         multipath.go:155: mount /sysroot has non-multipath source /dev/mapper/0x150a7519b1eaaa71p4
[2025-05-20T13:15:01.175Z] --- FAIL: multipath.day1 (36.71s)
[2025-05-20T13:15:01.175Z]         multipath.go:155: mount /sysroot has non-multipath source /dev/mapper/0x79b7cca15d2f4a68p4
```

which is interesting. @jlebon would you have expected multipath regressions as a result of https://github.com/dracut-ng/dracut-ng/pull/1306 ?

jlebon:
> CI failures here:
> 
> ```
> [2025-05-20T13:14:34.721Z] --- FAIL: multipath.day2 (84.86s)
> [2025-05-20T13:14:34.721Z]         multipath.go:155: mount /sysroot has non-multipath source /dev/mapper/0x150a7519b1eaaa71p4
> [2025-05-20T13:15:01.175Z] --- FAIL: multipath.day1 (36.71s)
> [2025-05-20T13:15:01.175Z]         multipath.go:155: mount /sysroot has non-multipath source /dev/mapper/0x79b7cca15d2f4a68p4
> ```
> 
> which is interesting. @jlebon would you have expected multipath regressions as a result of [dracut-ng/dracut-ng#1306](https://github.com/dracut-ng/dracut-ng/pull/1306) ?

I think it's more that [the test](https://github.com/coreos/coreos-assembler/blob/7d6fa376fc9f73625487adbb9386785bb09f1bb2/mantle/kola/tests/misc/multipath.go#L154) needs to be adapted so it doesn't verify that we're on multipath based on whether the device is `/dev/mapper/mpath*`. Instead do something [like this](https://github.com/coreos/fedora-coreos-config/blob/ebcd73e573596e7004598157abb12b47a5e5ca72/tests/kola/multipath/resilient/test.sh#L25).

Aside: a lot (probably all) of those multipath tests could nowadays be moved to external tests.

dustymabe:
> I think it's more that [the test](https://github.com/coreos/coreos-assembler/blob/7d6fa376fc9f73625487adbb9386785bb09f1bb2/mantle/kola/tests/misc/multipath.go#L154) needs to be adapted so it doesn't verify that we're on multipath based on whether the device is `/dev/mapper/mpath*`. Instead do something [like this](https://github.com/coreos/fedora-coreos-config/blob/ebcd73e573596e7004598157abb12b47a5e5ca72/tests/kola/multipath/resilient/test.sh#L25).

That seems reasonable, but is it possible here we're going to break people who somehow made assumptions based on what the device names were in the past?

dustymabe:
> Aside: a lot (probably all) of those multipath tests could nowadays be moved to external tests.

I would love that

manifests/shared-el.yaml
joelcapitao: Dunno if you read my previous comment https://github.com/coreos/fedora-coreos-config/pull/3508#discussion_r2097423201 about this.
I'd prefer using the `patch` command I put in this comment TBH, but cannot in current state. 
One solution would be to use `rpm-ostree compose postprocess/commit` and install `patch` for postprocess only, but might be overkill I'd say.

joelcapitao:
> Can we add some text to the body of the 2nd commit. A nice description of why we can re-enable this test would be useful.

Done in last patch set.


> > which is interesting. @jlebon would you have expected multipath regressions as a result of [dracut-ng/dracut-ng#1306](https://github.com/dracut-ng/dracut-ng/pull/1306) ?
> 
> I think it's more that [the test](https://github.com/coreos/coreos-assembler/blob/7d6fa376fc9f73625487adbb9386785bb09f1bb2/mantle/kola/tests/misc/multipath.go#L154) needs to be adapted so it doesn't verify that we're on multipath based on whether the device is `/dev/mapper/mpath*`. Instead do something [like this](https://github.com/coreos/fedora-coreos-config/blob/ebcd73e573596e7004598157abb12b47a5e5ca72/tests/kola/multipath/resilient/test.sh#L25).
> 
> Aside: a lot (probably all) of those multipath tests could nowadays be moved to external tests.

I proposed a PR to do so https://github.com/coreos/coreos-assembler/pull/4113 (i.e still in go tests though).
I added a task to move them to external test in my TODO list.

With my cosa PR, it's passing the verification step, but still fails while rebooting:
```
Timed out waiting for device dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device - /dev/disk/by-label/dm-mpath-boot.
Warning: /dev/disk/by-label/dm-mpath-root does not exist
```

manifests/fedora-coreos-base.yaml
dustymabe: I had not seen this before. Thanks for the explanation.

manifests/shared-el.yaml
dustymabe: I had not seen it, but I have now :).

I'd say we can mark this resolved.

dustymabe:
@jlebon with `user_friendly_names=no` we are going to need to adapt all of our tests to not hardcode in `mpatha`, correct?

jlebon:
> > I think it's more that [the test](https://github.com/coreos/coreos-assembler/blob/7d6fa376fc9f73625487adbb9386785bb09f1bb2/mantle/kola/tests/misc/multipath.go#L154) needs to be adapted so it doesn't verify that we're on multipath based on whether the device is `/dev/mapper/mpath*`. Instead do something [like this](https://github.com/coreos/fedora-coreos-config/blob/ebcd73e573596e7004598157abb12b47a5e5ca72/tests/kola/multipath/resilient/test.sh#L25).
> 
> That seems reasonable, but is it possible here we're going to break people who somehow made assumptions based on what the device names were in the past?

The problem is that those names were not stable to begin with and so not recommended (in favour of the WWID, but the problem with _that_ is that it's node-specific). I guess it is reliable if you only have a single multipathed device (in which case you're guaranteed I think to get `/dev/mapper/mpatha`) and don't plan to ever connect another multipathed device.

That said, a really important factor is that upgrading nodes will not get this change because the `/etc/multipath.conf` generated here [is propagated into the real root](https://github.com/coreos/fedora-coreos-config/blob/efa9dd5ec548396f58abf6a1c05a1f8f04923355/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-propagate-multipath-conf.service). So any existing nodes that did hardcode `/dev/mapper/mpath*` paths in some settings will "keep working" (the symlinks will exist, but this changes nothing about them being unreliable).

That said, there's definitely an argument to be made for not backporting this change because it's too significant to change mid-release (even if it would only take effect with new boot media). I think the workaround if you _don't_ want friendly names then on those older releases is to add an `/etc/multipath.conf` in your Ignition config equivalent to what `mpathconf --enable --user_friendly_names n` would output (and thus neuter the propagation logic). We should sanity-check this flow though before recommending it to users.

cc @bmarzins in case I got some details wrong.

jlebon:
> @jlebon with `user_friendly_names=no` we are going to need to adapt all of our tests to not hardcode in `mpatha`, correct?

Yes, indeed. Which... I think those previously fell under:

> I guess it is reliable if you only have a single multipathed device (in which case you're guaranteed I think to get /dev/mapper/mpatha) and don't plan to ever connect another multipathed device.

But at the same time, in realistic scenarios you very often do have multiple multpath devices and our provisioning flows should be able to handle those cases without having to jump through hoops. I.e. those should not be treated in a special way.

Anyway, arguing with myself a bit here but I think this is still the right change to make overall even though it might/likely will cause discomfort.

dustymabe:
so that means we need to update all the uses of `mpatha` in

- `ext.config.root-reprovision.luks.multipath` from f-c-c
- `mantle/kola/tests/misc/multipath.go` from COSA

Is that correct?

dustymabe:
> That said, there's definitely an argument to be made for not backporting this change because it's too significant to change mid-release (even if it would only take effect with new boot media).

So should we kill https://github.com/coreos/fedora-coreos-config/pull/3506 (the backport to 4.18 that we were holding up to pull in this patch) then?

Note that what we land here will slip straight into `rhel-9.6` stream which will affect OCP 4.19.

bmarzins:
> > > I think it's more that [the test](https://github.com/coreos/coreos-assembler/blob/7d6fa376fc9f73625487adbb9386785bb09f1bb2/mantle/kola/tests/misc/multipath.go#L154) needs to be adapted so it doesn't verify that we're on multipath based on whether the device is `/dev/mapper/mpath*`. Instead do something [like this](https://github.com/coreos/fedora-coreos-config/blob/ebcd73e573596e7004598157abb12b47a5e5ca72/tests/kola/multipath/resilient/test.sh#L25).
> > 
> > 
> > That seems reasonable, but is it possible here we're going to break people who somehow made assumptions based on what the device names were in the past?
> 
> The problem is that those names were not stable to begin with and so not recommended (in favour of the WWID, but the problem with _that_ is that it's node-specific). I guess it is reliable if you only have a single multipathed device (in which case you're guaranteed I think to get `/dev/mapper/mpatha`) and don't plan to ever connect another multipathed device.

That depends on what you mean when you say "stable". Multipath stores the mapping between user_friendly_name and WWID in /etc/multipath/bindings, and once there is a mapping there, multipath will always use it unless there already is a device with that name. That's actually the origin of the problem that  dracut-ng/dracut-ng#1306 is trying to fix. Since there is no /etc/multipath/bindings in the initramfs, the devices get their user_friendly_name there by the order that they are discovered. However once you switch-root, multipath can't always switch a device to the user_friendly_name it is supposed to have, since it might be in use by another device. In this case it will fail back to using the WWID as the name. 

As long as you only have one multipath device, it will always be named mpatha. If you later add another device, the original device will still always be named mpatha, unless the other device gets named mpatha in the initramfs, in which case the original device will be named by its WWID for that boot.

> That said, a really important factor is that upgrading nodes will not get this change because the `/etc/multipath.conf` generated here [is propagated into the real root](https://github.com/coreos/fedora-coreos-config/blob/efa9dd5ec548396f58abf6a1c05a1f8f04923355/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-propagate-multipath-conf.service). So any existing nodes that did hardcode `/dev/mapper/mpath*` paths in some settings will "keep working" (the symlinks will exist, but this changes nothing about them being unreliable).
> 
> That said, there's definitely an argument to be made for not backporting this change because it's too significant to change mid-release (even if it would only take effect with new boot media). I think the workaround if you _don't_ want friendly names then on those older releases is to add an `/etc/multipath.conf` in your Ignition config equivalent to what `mpathconf --enable --user_friendly_names n` would output (and thus neuter the propagation logic). We should sanity-check this flow though before recommending it to users.

Since the multipath user_friendly_names are supposed to be persistent (in RHEL, we tell users that they must remake their initramfs whenever a new multipath device is added, to propagate /etc/multipath/bindings to the initramfs and avoid the name mismatch issue) users may well use that specific name in their settings. As long as you don't change /etc/multipath.conf on existing nodes, you should be safe. But having multipath devices suddenly named differently on new nodes may well confuse users, so I'm not sure that you should make this change mid-release.

jlebon:
@bmarzins Thanks for the additional context!

We actually do have facilities to inject back the bindings into the initramfs as well. Though I'm not sure of the payoff of trying to wire that through in all the right places. It just seems to me like user friendly names should be avoided in general anyway.

> That said, a really important factor is that upgrading nodes will not get this change because the /etc/multipath.conf generated here [is propagated into the real root](https://github.com/coreos/fedora-coreos-config/blob/efa9dd5ec548396f58abf6a1c05a1f8f04923355/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-propagate-multipath-conf.service). So any existing nodes that did hardcode /dev/mapper/mpath* paths in some settings will "keep working" (the symlinks will exist, but this changes nothing about them being unreliable).

In discussing this with @dustymabe, we realized it's a little more nuanced than that. Upgrading nodes will indeed have a carried over `/etc/multipath.conf`, but _not in the initramfs_. Which means that any node using `root=/dev/mapper/mpatha4` for example will fail after upgrade because the `/etc/multipath.conf`	generated in the initramfs will now turn off those symlinks.

However, before this latest push to drop the `root=` karg requirement, it's been pretty well documented that you _must_ use `root=/dev/disk/by-label/dm-mpath-root` (see e.g. https://github.com/openshift/os/blob/master/docs/faq.md#q-does-rhcos-support-multipath-on-the-primary-disk and https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/installing_on_bare_metal/user-provisioned-infrastructure#rhcos-enabling-multipath_installing-bare-metal).

So in practice, there should be no or few systems using `root=/dev/mapper/mpatha4`. The fix there is simple of course if you do hit this: change your `root` karg, or better, use its UUID.

> So should we kill #3506 (the backport to 4.18 that we were holding up to pull in this patch) then?

Yeah, let's not backport it.

> Note that what we land here will slip straight into `rhel-9.6` stream which will affect OCP 4.19.

Yeah, I think doing this in 4.19+ is fine. We're fast-tracking a patch which will land in 9.6 anyway eventually.

jlebon:
@jcapiitao What's the status here? Can we rebase?

jlebon:
I think for 4.19 (but too late for GA), we should get this in, ensure the new test is no longer flaky, and then drop the `root=/dev/disk/by-label/dm-mpath-root` recommendation.

joelcapitao:
> @jcapiitao What's the status here? Can we rebase?

it should pass the tests now, let me rebase it and see how it goes

dustymabe:
From https://github.com/coreos/fedora-coreos-config/pull/3508#issuecomment-2899383374 

> so that means we need to update all the uses of `mpatha` in
> 
> * `ext.config.root-reprovision.luks.multipath` from f-c-c
> * `mantle/kola/tests/misc/multipath.go` from COSA
> 
> Is that correct?

I think we still need to update `ext.config.root-reprovision.luks.multipath` (hence CI failure here), right?

joelcapitao:
It now needs https://github.com/coreos/coreos-assembler/pull/4151

joelcapitao:
It finally passed CI :)

jlebon (approved):
Woohoo, thanks!

dustymabe (approved):
LGTM

PR #3511, opened 2025-05-19
manifests: create a shared-el.yaml manifest

dustymabe:
This manifest will include any misc shared bits with our enterprise linux downstream builds. It will dynamically pick up el9 or el10 specific snippets using conditional includes.

This is prep for adding snippets that we can share with both el9 and el10 and not just one or the other.


joelcapitao:
LGTM

joelcapitao:
LGTM

joelcapitao (approved):

PR #3513, opened 2025-05-19
[rawhide] overrides: unpin dnf5-5.2.11.0-1.fc43

aaradhak:
The dnf5 rpm can be unpinned as we have refactored systemd preset handling by moving enablement code to the minimal configuration. Also, the dnf-makecache.timer has been disabled by default for minimal images.

Ref:
- https://gitlab.com/fedora/bootc/base-images/-/merge_requests/181
- https://github.com/coreos/fedora-coreos-config/pull/3509
- https://github.com/coreos/fedora-coreos-tracker/issues/1896#issuecomment-2890876017

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,21 +9,6 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
-  dnf5:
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

PR #3514, opened 2025-05-19
feat: warn user on login if mount points are incorrect

### LINKED ISSUE #1954: Warn if /sysroot is not read only or if the root mountpoint is not a composefs mount
See: https://github.com/coreos/fedora-coreos-tracker/issues/1232

Add a warning if we find that /sysroot is not read only.

Roshan-R:
Closes https://github.com/coreos/fedora-coreos-tracker/issues/1954

### DIFF
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-warn-invalid-mounts.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-warn-invalid-mounts.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Warns if mount sources or options for critical system filesystems are incorrect
+Before=systemd-user-sessions.service
+ConditionPathExists=!/run/ostree-live
+
+[Service]
+Type=oneshot
+# Service should only run once at boot - set RemainAfterExit=yes.
+RemainAfterExit=yes
+ExecStart=/usr/libexec/coreos-warn-invalid-mounts
+
+[Install]
+WantedBy=multi-user.target
--- a/overlay.d/15fcos/usr/libexec/coreos-warn-invalid-mounts
+++ b/overlay.d/15fcos/usr/libexec/coreos-warn-invalid-mounts
@@ -1,0 +1,40 @@
+#!/usr/bin/bash
+
+. /usr/lib/console-login-helper-messages/motd.defs
+
+OUTDIR="${RUN_SNIPPETS}"
+
+warnings=""
+
+root_object=$(findmnt / --json | jq --arg ROOT '/' -r '.filesystems[] | select(.target==$ROOT)')
+
+# Check if the root filesystem source is "composefs"
+root_source=$(echo "$root_object" | jq -r '.source')
+if [[ "$root_source" != "composefs" ]]; then
+    warnings+="Warning: Root mountpoint ('/') is not a 'composefs' mount."$'\n'
+fi
+
+# Check if / is mounted read-only
+root_options=$(echo "$root_object" | jq -r '.options' | tr ',' '\n' | grep -c '^ro$')
+if [[ "$root_options" -ne 1 ]]; then
+    warnings+="Warning: Root mountpoint ('/') is not mounted read-only"$'\n'
+fi
+
+# Check if /sysroot is mounted read-only
+sysroot_options=$(findmnt /sysroot --json | jq --arg SYSROOT '/sysroot' -r '.filesystems[] | select(.target==$SYSROOT) | .options')
+if [[ $(echo "$sysroot_options" | tr ',' '\n' | grep -c '^ro$') -ne 1 ]]; then
+    warnings+="Warning: '/sysroot' is not mounted read-only"$'\n'
+fi
+
+# Remove trailing newline
+warnings="${warnings%$'\n'}"
+
+# Format final message with banner if warnings exist
+if [[ -n "$warnings" ]]; then
+    message=$'##########################################################################\n'
+    message+="$warnings"$'\n'
+    message+="To disable this warning, use:"$'\n'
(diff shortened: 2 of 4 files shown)


overlay.d/15fcos/usr/libexec/coreos-warn-invalid-mounts
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/bash
+
+# Copyright (c) 2020 Fedora CoreOS Authors. All rights reserved.
+# Copyright (c) 2013 The CoreOS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+# Modified from the CoreOS repository: https://github.com/coreos/init/blob/master/scripts/sshd_keygen
```
jbtrystram: We usually don't have this kind of license headers

overlay.d/15fcos/usr/libexec/coreos-warn-invalid-mounts
```diff
@@ -0,0 +1,40 @@
...
+
+set -e
+
+. /usr/lib/console-login-helper-messages/libutil.sh
+. /usr/lib/console-login-helper-messages/issue.defs
+
+OUTDIR="${RUN_SNIPPETS}"
```
jbtrystram, line 5: This variable is not set. 
```
OUTDIR=/run/motd.d/
```

overlay.d/15fcos/usr/libexec/coreos-warn-invalid-mounts
```diff
@@ -0,0 +1,40 @@
...
+
+message=""
+
+# Check if the root filesystem source is "composefs"
+root_source=$(findmnt / --json | jq -r '.filesystems[0].source')
```
jbtrystram: A more deterministic check would be : `findmnt / --json | jq --arg ROOT '/' -r '.filesystems[] | select(.target==$ROOT) | .source'`

jbtrystram (changes requested):
I made a couple of suggestion, but the overall approach LGTM
Make sure to update the overlay README as well

overlay.d/15fcos/usr/libexec/coreos-warn-invalid-mounts
Roshan-R: the OUTDIR is populated with the RUN_SNIPPETS that is sourced from the /usr/lib/console-login-helper-messages/issue.defs file having this logic inside it

```bash
private_issue_snippets_path="/run/${PKG_NAME}/issue.d"
# Since util-linux 2.35, `agetty` has been able to display files
# from additional locations like `/run/issue` and `/run/issue.d`
# if `/etc/issue` does not exist.
# We are currently dropping issue snippets in `/etc/issue.d` (instead
# of `/run/issue.d`) because of a systemd tmpfile that makes removal of
#  `/etc/issue` virtually impossible.
# https://github.com/systemd/systemd/commit/062666c7c452711f3e38efdec0c2b952ced64d06
# In the future, we would ideally drop into `/run/issue.d`; this would
# require either users configure the `--issue-file` option in `agetty`,
# or not use `/etc/issue`.
public_issue_snippets_path="/etc/issue.d"
if [ "${USE_PUBLIC_RUN_DIR}" == "false" ]; then
    RUN_SNIPPETS="${private_issue_snippets_path}"
else
    RUN_SNIPPETS="${public_issue_snippets_path}"
fi
```

overlay.d/15fcos/usr/libexec/coreos-warn-invalid-mounts
```diff
@@ -0,0 +1,34 @@
...
+message=""
+
+# Check if the root filesystem source is "composefs"
+root_source=$(findmnt / --json | jq --arg ROOT '/' -r '.filesystems[] | select(.target==$ROOT) | .source')
+if [[ "$root_source" != "composefs" ]]; then
+    message+="Warning: Root mountpoint ('/') is not a 'composefs' mount."$'\n'
+fi
+
+# Check if / is mounted read-only
+root_options=$(findmnt / --json | jq -r '.filesystems[].options')
```
jbtrystram: use the same `jq` as above, or better yet, save the output for the whole root object then reuse that

jbtrystram:
I think we should also add a line explaining how to disable this warning, some people may want to not run composefs for some reason.

overlay.d/15fcos/usr/lib/systemd/system/coreos-warn-invalid-mounts.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Warns if mount sources or options for critical system filesystems are incorrect
+Before=systemd-user-sessions.service
```
jbtrystram, line 3: I think the liveiso is not composefs, and isn't writeable by definition so it's worth skipping entirely with e.g. :
`ConditionPathExists=!/run/ostree-live`

travier (approved):

Roshan-R:
How to test the PR:

```
# 1. Run using cosa and composefs=no flag

cosa run --kargs ostree.prepare-root.composefs=no -c

# 2. This only triggered two of the three warnings, for the other remount /sysroot to writable

sudo mount -o remount,rw /sysroot

# 3. Restart the service 

sudo systemctl restart coreos-warn-invalid-mounts.service

# 4. Re-login to shell using exit 

exit
```

Warning should be like this 

```
##########################################################################
Warning: Root mountpoint ('/') is not a 'composefs' mount.
Warning: Root mountpoint ('/') is not mounted read-only
Warning: '/sysroot' is not mounted read-only
To disable this warning, use:
  sudo systemctl disable coreos-warn-invalid-mounts.service
##########################################################################
```

jbtrystram (approved):
LGTM
Thanks !

PR #3516, opened 2025-05-21
tests/kdump.crash: bump kdump.service activation timeout

marmijo:
We've been seeing this kdump.service activation hit the timeout on the Fedora CoreOS aarch64 builder. 240s seems to be enough to allow the service to activate on the builder, so let's increase the timeout to unblock the FCOS pipeline.

### DIFF
--- a/tests/kola/kdump/crash/test.sh
+++ b/tests/kola/kdump/crash/test.sh
@@ -19,9 +19,9 @@ set -xeuo pipefail
 
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
-      # use 120s for this since kdump can take a while to build its initramfs,
+      # use 240s for this since kdump can take a while to build its initramfs,
       # especially if the system is loaded
-      if ! is_service_active kdump.service 120; then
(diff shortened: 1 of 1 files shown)


aaradhak (approved):
LGTM

tests/kola/kdump/crash/test.sh
```diff
@@ -19,9 +19,9 @@ set -xeuo pipefail
 
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   "")
-      # use 120s for this since kdump can take a while to build its initramfs,
+      # use 240s for this since kdump can take a while to build its initramfs,
```
dustymabe, line 22: 120 to 240 is quite the jump. In your testing what value was this bumping up to in real world runs (i.e. when it would pass)?

marmijo: You're right. I added some print statements to the test and saw the service took anywhere between 130 to 170 seconds to activate. Although, the timeout of 240s was still exceeded when other jobs were actively using the aarch64 builder.

dustymabe (approved):

PR #3517, opened 2025-05-21
tests/commonlib: use systemd-run to wait for services

dustymabe:
Rather than busy looping we can use systemd-run to start a transient unit that will wait for a service to be started.

For now we still need to rely on /bin/timeout to implement the timeout because providing --property=JobTimeoutSec=${timeout} gave me: Cannot set property JobTimeoutUSec, or unknown property.


marmijo (approved):
LGTM

aaradhak (approved):

PR #3520, opened 2025-05-22
make postprocess more consistent; drop preset from postprocess

dustymabe:
```
manifests: make postprocess preamble more consistent

We always want set -x so we can see more information if things either
fail during a postprocess or if things in the final image somehow don't
turn out how we inpsect. The logs will help us figure out what is going
on.
```

---

```
manifests: drop preset-all from postprocess

This is done in bootc now as of [1] and we've bumped
the bootc submodule since then so we can drop it here now.

[1] https://gitlab.com/fedora/bootc/base-images/commit/4888be4e199583664fe32264f16ff9e166372c34
```

### DIFF
--- a/manifests/coreos-bootc-minimal-plus.yaml
+++ b/manifests/coreos-bootc-minimal-plus.yaml
@@ -25,8 +25,8 @@ opt-usrlocal: var
 
 postprocess:
   - |
-    #!/usr/bin/env bash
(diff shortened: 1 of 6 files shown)


jlebon (approved):
Yeah, makes sense. I actually have a similar patch for fedora-bootc as well pending to debug an unrelated thing.

PR #3526, opened 2025-05-30
denylist: drop kola reprovision tests for power from denylist

Poorna-Gottimukkula1:
Kola tests are now passing since the fix released in 6.15.0-0.rc7, so we can remove it now and close:
https://github.com/coreos/fedora-coreos-tracker/issues/1926

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -16,12 +16,5 @@
     - openstack
   streams:
     - rawhide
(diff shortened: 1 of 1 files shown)


c4rt0 (approved):
/lgtm

Based on the positive results in:
https://github.com/coreos/fedora-coreos-tracker/issues/1926#issuecomment-2918827549

dustymabe (approved):
LGTM

PR #3527, opened 2025-05-30
[rawhide] overrides: unpin rpm-4.20.1-3.fc43

aaradhak:
The fix for kola tests (ostree.unlock, ostree.hotfix) which failed with a "no digest" error after a system rpm upgrade to rpm-5.99.90-2.fc43 has merged. We can now unpin the rpm-4.20.1-3.fc43

Ref: https://github.com/coreos/coreos-assembler/pull/4098


dustymabe (approved):
LGTM

PR #3529, opened 2025-06-02
manifests/fedora-coreos: allow python in F43+ for now

dustymabe:
`nfs-utils-coreos` grew a dependency for `nfs-utils`, which requires
`python`. Let's allow it in rawhide and hopefully the dependency
will get cleaned up before it reaches production FCOS streams.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1942#issuecomment-2881075898


dustymabe:
Note this commit also adds a test to warn us if any more packages add a python dependency.

marmijo (approved):
LGTM

PR #3530, opened 2025-06-02
[rawhide] overrides: drop nfs-utils-coreos pin in rawhide

dustymabe:
For now we loosened the requirement of "no python" in Fedora 43+ so we can drop this pin.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1942#issuecomment-2881075898

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,11 +9,6 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
-  nfs-utils-coreos:
-    evr: 1:2.8.2-1.rc8.fc43
-    metadata:
-      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1942
-      type: pin
   selinux-policy:
     evra: 41.41-1.fc43.noarch
     metadata:


marmijo (dismissed):
LGTM

dustymabe:
Sigh `ext.config.networking.network-online-service` is failing because `network-online.target` is getting pulled in by `rpc-statd-notify.service` which is getting pulled in by `nfs-client.target` which is enabled by default via presets.

dustymabe:
Ideas for how to workaround the network-online.target problem:

- neuter the network-online Wants from rpc-statd-notify.service in a postprocess
- denylist the network-online test for rawhide
- specifically to disable nfs-client.target via butane/ignition for the `ext.config.networking.network-online-service` test

I consider all of these to be temporary solutions

jlebon:
Hmm, yeah I think this one should be OK because it's a leaf node and doesn't actually block bootup (which is the primary thing that test is trying to guard against).

That said, given that
- `network-online.target` is in general discouraged,
- `nfs-utils` is in the default set of installed packages for a lot of variants, and
- NFS seems like a good example of what _should be able to_ withstand dynamic network (though to be fair, here we're talking about the client side, not server side),

it would be good to get rid of that dependency. Can we go with

> specifically to disable nfs-client.target via butane/ignition for the `ext.config.networking.network-online-service` test

and file a bug against nfs-utils to discuss this? The advantage over denylisting is that we still keep an eye out for other packages that may introduce it back.

dustymabe:
@jlebon did the first part in https://github.com/coreos/fedora-coreos-config/pull/3532

aaradhak (approved):

dustymabe:
> @jlebon did the first part in #3532

and the second part in https://bugzilla.redhat.com/show_bug.cgi?id=2383585

PR #3531, opened 2025-06-03
tests/python: fix test to allow python in f43+

jbtrystram:
[1] introduced python dependencies and a test to make sure no python was introduced in earlier releases. The test intended to be skip for versions above 43 but use `verlt`. Update to use `vergte` which will skip the test for f43 and above, as intended.

[1] e70fd15dc08ec1ca2234f513275dd18ca0430206

### DIFF
--- a/tests/kola/python/test.sh
+++ b/tests/kola/python/test.sh
@@ -9,7 +9,7 @@ set -xeuo pipefail
 # shellcheck disable=SC1091
 . "$KOLA_EXT_DATA/commonlib.sh"
 
-if verlt "$(get_fedora_ver)" 43; then
+if vergte "$(get_fedora_ver)" 43; then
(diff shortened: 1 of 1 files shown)


tests/kola/python/test.sh
```diff
@@ -9,7 +9,7 @@ set -xeuo pipefail
 # shellcheck disable=SC1091
 . "$KOLA_EXT_DATA/commonlib.sh"
 
-if verlt "$(get_fedora_ver)" 43; then
+if vergte "$(get_fedora_ver)" 43; then
```
dustymabe, line 12: I think this is/was working as expected. Looking at the logs from the pipeline failure it might be because https://github.com/coreos/fedora-coreos-config/pull/3530 hasn't merged yet that the test is failing.

PR #3534, opened 2025-06-04
Prep patches for container-native path

jlebon:
These commits were split out of https://github.com/coreos/fedora-coreos-config/pull/3348.

### DIFF
--- a/image-default.yaml
+++ b/image-default.yaml
@@ -1,0 +1,36 @@
+# This file contains defaults for image.yaml and historically lived in
+# coreos-assembler.
+
+# related: https://github.com/osbuild/bootc-image-builder/pull/932
+bootfs: "ext4"
+rootfs: "xfs"
+# Add arguments here that will be passed to e.g. mkfs.xfs
+rootfs-args: ""
+# Set to "true" to enable composefs
+composefs: false
+
+# Additional default kernel arguments injected into disk images
+extra-kargs: []
+
+# Can also be oci-chunked
+ostree-format: oci-chunked-v1
+# Inject io.openshift.build.version-display-names for OpenShift CVO
+ostree-container-inject-openshift-cvo-labels: false
+# True if we should use `ostree container image deploy`
+deploy-via-container: false
+
+# Set this to a target container reference, e.g. ostree-unverified-registry:quay.io/example/os:latest
+# container-imgref: ""
+
+# Format used when generating a squashfs image.  Can also be e.g. gzip or lz4
+squashfs-compression: zstd
+
+# Defaults for VMware OVA, matching historical behavior
+vmware-hw-version: 13
+vmware-os-type: rhel7_64Guest
+vmware-secure-boot: true
+
+# Defaults for AWS
+aws-imdsv2-only: true
+aws-volume-type: "gp3"
+aws-x86-boot-mode: "uefi-preferred"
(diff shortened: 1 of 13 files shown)


jlebon:
I'll do a final sanity-check with `cosa diff` to make sure this is effectively a no-op.

jlebon:
> I'll do a final sanity-check with `cosa diff` to make sure this is effectively a no-op.

Done! :heavy_check_mark:

image-default.yaml
jbtrystram, line 1: Those defaults are really different than what we use in FCOS. If we need to carry this in f-c-c, why not make the defaults closer to what we use ?

manifests/fedora-coreos-base.yaml
```diff
@@ -34,11 +34,11 @@ ignore-removed-users:
...
 ignore-removed-groups:
   - root
+# Add the sudo group to /etc/group
+# This is re-implemented below for the container-native flow. Nuke this once
+# that's the only path we support.
 etc-group-members:
-  - wheel
   - sudo
-  - systemd-journal
-  - adm
```
jbtrystram, line 42: I suppose these groups are not longer used ?

manifests/fedora-coreos.yaml
```diff
@@ -165,13 +165,22 @@ postprocess:
         done
     done
 
+  # Delete documentation. (Note that fedora-bootc doesn't explicitly set
+  # `documentation`, and neither do we, so the rpm-ostree default is true. We
+  # historically have effectively done `the equivalent of `documentation: false`
```
jbtrystram, line 170: Nit : extra backtick ? 
```suggestion
  # historically have effectively done the equivalent of `documentation: false`
```

manifests/fedora-coreos.yaml
```diff
@@ -165,13 +165,22 @@ postprocess:
         done
     done
 
+  # Delete documentation. (Note that fedora-bootc doesn't explicitly set
```
jbtrystram, line 168: Is there an issue or a discussion so I could learn the reason for that ? Is there something more than "using less space" ?

jbtrystram (approved):
LGTM ! 
I left a couple of questions

image-default.yaml
jlebon: They're also used by RHCOS. It seems safer as part of this move to just directly keep untouched the file as it was in cosa. Once we do the same in RHCOS, we can drop the cosa copy, and then we can simplify the defaults.

manifests/fedora-coreos-base.yaml
jlebon: They're already inherited by fedora-bootc: https://gitlab.com/fedora/bootc/base-images/blob/e3b9fbd6bafad97cd7f8aac25672490a9afe5ced/minimal/postprocess-conf.yaml#L21.

I guess I could add a comment, though it feels weird to have a comment about something that's _not_ there. Presumably this could've gone away when we did the switch to inherit from tier-x.

manifests/fedora-coreos.yaml
jlebon: I think we do, but I can't find it right now. But basically yes mostly space. We decided not to ship `man`/`info` and we don't ship that there's no point in shipping the docs.

manifests/fedora-coreos.yaml
jlebon: Good catch! To not waste good CI, I'll fix that up as part of https://github.com/coreos/fedora-coreos-config/pull/3348!

PR #3535, opened 2025-06-04
tests: mark lockdown test as exclusive

dustymabe:
Mark as exclusive since we are requesting secureboot and that won't take effect when we're running non-exclusively.

Partial fix for https://github.com/coreos/fedora-coreos-tracker/issues/1966


aaradhak (approved):
LGTM

PR #3536, opened 2025-06-05
overlay.d: Update ignition.firstboot services for soft-reboots

jmarrero:
https://github.com/coreos/coreos-assembler/pull/4119 surfaced that this service would fail on soft-reboots, it's non fatal but would make the Kola tests fail. It looks like originally this was set to fail instead of risking not running it:
https://github.com/coreos/fedora-coreos-config/blob/20feb176f19c3142b7256c1eb5bf1cb7c53b29b9/overlay.d/05core/usr/libexec/coreos-ignition-firstboot-complete#L14

### DIFF
--- a/overlay.d/05core/usr/lib/systemd/system/coreos-ignition-firstboot-complete.service
+++ b/overlay.d/05core/usr/lib/systemd/system/coreos-ignition-firstboot-complete.service
@@ -3,6 +3,10 @@ Description=CoreOS Mark Ignition Boot Complete
 Documentation=https://docs.fedoraproject.org/en-US/fedora-coreos/
 ConditionKernelCommandLine=ignition.firstboot
 ConditionPathExists=!/run/ostree-live
+# This condition is required for cases where a soft-reboot is issued on the
+# firstboot. Soft reboot do not change the kernel or its command-line arguments,
+# which would cause this service to fail after the soft-reboot completes.
+ConditionPathExists=/boot/ignition.firstboot
 RequiresMountsFor=/boot
 
 [Service]
--- a/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
+++ b/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
@@ -11,7 +11,7 @@ After=afterburn-sshkeys.target
 # before Ignition/Afterburn started logging structured data don't
 # get misleading messages. Also handles the case where the journal
 # gets rotated and no longer has the structured log messages.
-ConditionKernelCommandLine=ignition.firstboot
+ConditionFirstBoot=true
 # Run before user sessions to avoid reloading agetty
 Before=systemd-user-sessions.service
 


cgwalters (commented):
LGTM (but I suggested this so I won't officially approve)

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-firstboot-complete.service
```diff
@@ -3,6 +3,7 @@ Description=CoreOS Mark Ignition Boot Complete
 Documentation=https://docs.fedoraproject.org/en-US/fedora-coreos/
 ConditionKernelCommandLine=ignition.firstboot
 ConditionPathExists=!/run/ostree-live
+ConditionPathExists=!/boot/ignition.firstboot
```
jlebon: Hmm, shouldn't the polarity be flipped here?

jlebon (commented):
I presume `ConditionFirstBoot=true` units do not run again in a soft reboot on the first boot since `/etc/machine-id` is now present. Correct?

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-firstboot-complete.service
cgwalters: Ahhh...yes I think so

jmarrero: flipped and it also prevents the failure of the test.

jmarrero:
> I presume `ConditionFirstBoot=true` units do not run again in a soft reboot on the first boot since `/etc/machine-id` is now present. Correct?

/etc/machine-id is there:
```
$ cosa run
Fedora CoreOS 42.20250606.dev.0
Discuss: https://discussion.fedoraproject.org/tag/coreos

Last login: Fri Jun  6 13:51:17 2025
core@cosa-devsh:~$ cat /etc/machine-id 
a77f9d521de9495585ba4c92ee22dcfb
```
```
core@cosa-devsh:~$ sudo systemctl soft-reboot 
Broadcast message from root@localhost on pts/1 (Fri 2025-06-06 13:51:51 UTC):
The system will reboot now!

[EVENT | QEMU guest is ready for SSH] [ [0;32m  OK   [0m] Started  [0;1;39mzincati.service [0m - Zincati Update Agent.Fedora CoreOS 42.20250606.dev.0
Discuss: https://discussion.fedoraproject.org/tag/coreos

Last login: Fri Jun  6 13:51:54 2025
core@cosa-devsh:~$ cat /etc/machine-id 
a77f9d521de9495585ba4c92ee22dcfb
```

However, I don't see anything in the code or service that mentions /etc/machine-id... Not sure if you mean that it should not run because /etc/machine-id means that is not longer the first boot.

overlay.d/05core/usr/lib/systemd/system/coreos-ignition-firstboot-complete.service
```diff
@@ -3,6 +3,7 @@ Description=CoreOS Mark Ignition Boot Complete
 Documentation=https://docs.fedoraproject.org/en-US/fedora-coreos/
 ConditionKernelCommandLine=ignition.firstboot
 ConditionPathExists=!/run/ostree-live
+ConditionPathExists=/boot/ignition.firstboot
```
jlebon, line 9: I think this deserves a comment about why we have both `ConditionKernelCommandLine=ignition.firstboot` and this one.

jlebon (commented):
One comment, but looks fine overall. That said, one could imagine other systemd units that use `ignition.firstboot` out there instead of `ConditionFirstBoot` that will also hit this.

For this specific unit, I think the current conditionals here are more appropriate than `ConditionFirstBoot=`, but everything else should probably use that.

OK and we have one actually in this very repo: https://github.com/coreos/fedora-coreos-config/blob/112a7851c6d19a2a0e1a23f371cc329efcbd31ec/overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service#L14. Can you adapt that one?

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
```diff
@@ -12,6 +12,10 @@ After=afterburn-sshkeys.target
 # get misleading messages. Also handles the case where the journal
 # gets rotated and no longer has the structured log messages.
 ConditionKernelCommandLine=ignition.firstboot
+# This condition is required for cases where a soft-reboot is issued on the
+# firstboot. Soft reboot do not change the kernel or its command-line arguments,
+# which would cause this service to fail after the soft-reboot completes.
+ConditionPathExists=/boot/ignition.firstboot
```
jlebon: Not quite. What I mean is that this unit should use `ConditionFirstBoot=true` instead of `ConditionKernelCommandLine=` + `ConditionPathExists=`.

cgwalters: Hum. Looking there are some other uses of `ConditionFirstBoot=true` but not many

```
$ podman run --rm -ti quay.io/fedora/fedora-coreos:stable grep -r ConditionFirstBoot /usr/lib/systemd/system
/usr/lib/systemd/system/systemd-homed-firstboot.service:ConditionFirstBoot=yes
/usr/lib/systemd/system/coreos-update-ca-trust.service:ConditionFirstBoot=true
/usr/lib/systemd/system/coreos-update-ca-trust.service:# All services which use ConditionFirstBoot=yes should use
/usr/lib/systemd/system/ignition-delete-config.service:ConditionFirstBoot=true
/usr/lib/systemd/system/first-boot-complete.target:ConditionFirstBoot=yes
/usr/lib/systemd/system/zincati.service:# created so that we don't cause ConditionFirstBoot=true units to run twice
/usr/lib/systemd/system/systemd-firstboot.service:ConditionFirstBoot=yes
/usr/lib/systemd/system/afterburn-firstboot-checkin.service:ConditionFirstBoot=yes
$
```

I think I agree with you it'd be better to do so, *however* a general issue with `ConditionFirstBoot` is it doesn't handle the case where the system is interrupted (poweroff/crash) during first boot. Those services will then later fail to reconcile. 

Dunno that it matters though in practice.

jlebon (approved):

overlay.d/15fcos/usr/lib/systemd/system/coreos-check-ssh-keys.service
jlebon: A lot of background on that topic. See discussions in e.g. https://github.com/systemd/systemd/issues/4511. This is why we do this here for example: https://github.com/coreos/fedora-coreos-config/blob/bfe3327d7df792ec315c6e839cfef76007473195/overlay.d/05core/usr/lib/systemd/system/coreos-update-ca-trust.service#L9-L16

jlebon: Which, presumably we should also do for `coreos-check-ssh-keys.service`, but yeah I'm not too worried about it.

PR #3538, opened 2025-06-06
dracut: Ignore ignition modules if building kdump initrd

### LINKED ISSUE #1832: Configure kdump to exclude some coreos modules in generated kdump initramfs
There are some low memory cases where kdump is failing to start.

The initramfs generated for kdump includes all the modules used for CoreOS, including Ignition, which make it larger than needed (i.e. Ignition shouldn't and won't run on the kdump boot). So there is opportunity for us to reduce the size (and thus memory requirements) through configuration.

This request originated in https://issues.redhat.com/browse/OCPBUGS-44368

Johan-Liebert1:
Fixes: coreos/fedora-coreos-tracker#1832

Also needs 
https://github.com/coreos/afterburn/pull/1216
~~https://github.com/coreos/ignition/pull/2090~~

Kdump initrd size before patch 66M, after patch 39M

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh
@@ -2,6 +2,12 @@
 # -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
 # ex: ts=8 sw=4 sts=4 et filetype=sh
 
+check() {
+    if [[ $IN_KDUMP == 1 ]]; then
+        return 1
+    fi
+}
+
 depends() {
     echo systemd network ignition coreos-live
 }
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/module-setup.sh
@@ -1,3 +1,9 @@
+check() {
+    if [[ $IN_KDUMP == 1 ]]; then
+        return 1
+    fi
+}
+
 depends() {
     # We need the rdcore binary
     echo rdcore
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/module-setup.sh
@@ -2,6 +2,12 @@
 # -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
 # ex: ts=8 sw=4 sts=4 et filetype=sh
 
+check() {
+    if [[ $IN_KDUMP == 1 ]]; then
+        return 1
+    fi
+}
+
 depends() {
     echo multipath
 }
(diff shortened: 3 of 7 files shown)


Johan-Liebert1:
hmmm, to generate vmcore, we still need `crashkernel=256M`. No vmcore is generated with just 128M

Johan-Liebert1:
The entire unpacked initrd comes in at 99M

`kdumpctl estimate` also recommends at least 256M of reserved memory

```
Reserved crashkernel:    256M
Recommended crashkernel: 256M

Kernel image size:   46M
Kernel modules size: 11M
Initramfs size:      38M
Runtime reservation: 64M
Large modules:
    xfs: 3670016
    kvm: 1482752

```

jlebon:
OK actually, there is a more explicit check we could do here: [mkdumprd sets `IN_KDUMP=1`](https://github.com/rhkdump/kdump-utils//blob/1d40ea3a186c65e009f8cca775e6bb07c39fd80f/mkdumprd#L22), which can then be checked in the dracut modules to know if the kdump initrd is being built ([example](https://github.com/rhkdump/kdump-utils//blob/1d40ea3a186c65e009f8cca775e6bb07c39fd80f/dracut/99earlykdump/module-setup.sh#L11), [example](https://github.com/rhkdump/kdump-utils//blob/1d40ea3a186c65e009f8cca775e6bb07c39fd80f/dracut/99kdumpbase/module-setup.sh#L32)).

So let's use that instead.

Johan-Liebert1:
Thanks for the review. Updated PR to use `$IN_KDUMP`.

Another module for consideration is `99emergency-shell-setup`. This seems to be partially tied to ignition, so we could also skip this one?

jbtrystram:
We should also have a test checking that the modules are not present in the kdump initramfs, in order to not regress in case upstream kexec-tools change that `IN_KDUMP` variable IMO

jbtrystram:
> Another module for consideration is `99emergency-shell-setup`. This seems to be partially tied to ignition, so we could also skip this one?

Yes we can, we already have the needed bits to extract the journal in kdump tests. However this module is really small, so I am not expecting a big size difference

jlebon (approved):
Thanks!

jlebon:
> We should also have a test checking that the modules are not present in the kdump initramfs, in order to not regress in case upstream kexec-tools change that `IN_KDUMP` variable IMO

Hmm, you think? I would consider that part of the API, but maybe not. cc @licliu @coiby

licliu:
Hi @jlebon, IN_KDUMP is a long-standing variable that is unlikely to change, but we only use it in kdump maintained dracut modules. For other dracut modules, we use dracut hostonly check to avoid including unnecessary modules into the kdump initramfs.

I think any correct implementation is acceptable, I just prefer to use the method that is already in use. And I think it would be useful to add a test, sometimes modules are added to the initramfs as dependencies, and the test could provide some useful information for debugging.

BTW, we also maintain a list of modules that are explicitly excluded when building the kdump initramfs  https://github.com/rhkdump/kdump-utils/blob/1d40ea3a186c65e009f8cca775e6bb07c39fd80f/99-kdump.conf. But for you, it may be more flexible to maintain related code in a repository that you are familiar with.

Johan-Liebert1:
Thanks for merging this. As suggested by @jbtrystram, I'll create another PR adding tests

PR #3547, opened 2025-06-10
Containerfile: switch to quay.io/fedora/fedora-bootc

jlebon:
Now that we have bootc-base-imgectl in the canonical fedora-bootc image, we can just use that instead. That image also has all the arches which would be a blocker to turning this on for rawhide.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -8,13 +8,9 @@
 
 # This is the developer default version. In pipelines, this is driven by versionary.
 ARG VERSION="42"
-# XXX: This uses the Konflux-built version until it takes over from pungi and
-# shows up at the official quay.io/fedora/fedora-bootc endpoint since the latter
-# doesn't yet have bootc-base-imagectl.
-# https://gitlab.com/fedora/bootc/base-images/-/issues/44
-# XXX: Note also this should be a digested pull that gets bumped.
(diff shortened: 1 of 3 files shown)


Containerfile
```diff
@@ -8,13 +8,9 @@
...
 ARG VERSION="42"
-# XXX: This uses the Konflux-built version until it takes over from pungi and
-# shows up at the official quay.io/fedora/fedora-bootc endpoint since the latter
-# doesn't yet have bootc-base-imagectl.
-# https://gitlab.com/fedora/bootc/base-images/-/issues/44
-# XXX: Note also this should be a digested pull that gets bumped.
+# XXX: This should be a digested pull that gets bumped.
 # https://gitlab.com/fedora/bootc/tracker/-/issues/34
-ARG BUILDER_IMG=quay.io/bootc-devel/fedora-bootc-42-standard:latest
+ARG BUILDER_IMG=quay.io/fedora/fedora-bootc:42
```
dustymabe, line 13: Is this the standard image that includes `python`?

dustymabe (approved):
LGTM

Containerfile
jlebon: Yup!

PR #3548, opened 2025-06-11
tests/kola: limit `platforms.azure.nvme` test to x86_64

marmijo:
Azure NVMe disk support is only available with x86_64 sizes, so the test should only run on that architecture.


dustymabe (approved):
LGTM

PR #3550, opened 2025-06-16
Konflux: build FCOS using podman path

jbtrystram:
Signed-off-by: konflux-fedora <konflux@no-reply.konflux-ci.dev>


jlebon:
Note https://github.com/coreos/fedora-coreos-config/pull/3551 is proposing to use the `-v $PWD:/mnt` trick to work around https://github.com/containers/buildah/issues/5952. A similar trick was implemented in Konflux that we should be able to use: https://github.com/konflux-ci/build-definitions/pull/2269

PR #3551, opened 2025-06-16
Various patches for RHCOS container-native path enablement

jlebon:

---

Requires:
- [x] [Various tweaks as part of RHCOS container-native path work (!209) · Merge requests · fedora / bootc / Base Images · GitLab](https://gitlab.com/fedora/bootc/base-images/-/merge_requests/209)
- [x] [Drop `fedora-` prefix from canonical manifests (!210) · Merge requests · fedora / bootc / Base Images · GitLab](https://gitlab.com/fedora/bootc/base-images/-/merge_requests/210)


jlebon:
This should be ready to go now!

build-rootfs
```diff
@@ -108,12 +108,15 @@ def get_locked_nevras():
     return [f'{k}-{v}' for (k, v) in locks.items()]
 
 
-def inject_pool_repo(locked_nevras):
+def inject_any_pool_repo(locked_nevras):
```
dustymabe: totally optional and only if we do another push:

I think naming `inject_pool_repo_if_exists(locked_nevras):` might be clearer.

build-rootfs
```diff
@@ -134,7 +134,13 @@ def inject_version_info(rootfs, base_version, version):
     os_release_path = os.path.join(rootfs, 'usr/lib/os-release')
     with open(os_release_path) as f:
         from collections import OrderedDict
-        os_release = OrderedDict([line.strip().split('=', 1) for line in f])
+        os_release = OrderedDict()
+        for line in f:
+            line = line.strip()
+            if len(line) == 0 or line.startswith('#'):
```
dustymabe, line 141: do we need to consider lines with spaces before the `#` starts?

argfile.conf
dustymabe, line 1: I think it would be way more obvious here if we left this file named `Containerfile.args`.

Alternatively something like `container-build-argfile.conf` would be good too. Right now it's not super clear what it is just from the name.

dustymabe (dismissed):
Looks mostly good to me. Haven't tested this myself but don't block on that because I don't know when I'll get around to it. Some suggestions inline.

build-rootfs
jlebon: We `strip()` just above this which should take care of that.

argfile.conf
jlebon: > I think it would be way more obvious here if we left this file named `Containerfile.args`.

Hmm, I'd be worried people might think it's a Containerfile since `Containerfile.$x` is a common pattern too.

> Alternatively something like `container-build-argfile.conf` would be good too. Right now it's not super clear what it is just from the name.

`argfile.conf` [is suggested by buildah](https://github.com/containers/buildah/blob/e6375b3c282c199d340e8df53a001cd85dfcae51/docs/buildah-build.1.md?plain=1#L79) itself. But yeah, since it's not commonplace, I'm OK with being more explicit.

OK, changed it to `build-args.conf` to not make it too wordy. Let me know what you think!

jlebon:
Updated for comments!

dustymabe (approved):
LGTM

PR #3552, opened 2025-06-17
platforms: add console settings for applehv

dustymabe:
See the documentation in:
https://github.com/crc-org/vfkit/pull/323

### DIFF
--- a/platforms.yaml
+++ b/platforms.yaml
@@ -14,6 +14,16 @@
 #
 # All architectures, platforms, and fields are optional.
 aarch64:
+  # https://github.com/crc-org/vfkit/pull/323/files
+  # XXX update the above link to actual docs when it merges
+  applehv:
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=tty0
+      - console=hvc0
   aws:
     # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/connect-to-serial-console.html#sc-considerations
     # It appears there's no screenshot support on ARM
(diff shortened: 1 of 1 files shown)


platforms.yaml
```diff
@@ -14,6 +14,15 @@
...
 aarch64:
+  # https://github.com/crc-org/vfkit/pull/323/files
+  # XXX update the above link to actual docs when it merges
+  applehv:
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=hvc0
```
dustymabe: @cfergeau - do we want just this OR do we want some combination of console= arguments?

Also, should we modify the grub console= kernel args for x86_64 applehv images too?

baude: we need the GUI and serial consoles to both work ideally.

baude: I guess GUI == tty0

cfergeau: `console=hvc0` works on intel as well, so you can add it to the x86_64 image too, see https://github.com/crc-org/vfkit/pull/323#discussion_r2151942532

cfergeau: > we need the GUI and serial consoles to both work ideally.
> I guess GUI == tty0

@vyasgun do you know which parameters should be used to achieve this?

baude: I think it is just multiple declarations of `console=`

dustymabe:
@baude @cfergeau updated this PR.

jlebon (approved):

platforms.yaml
vyasgun: `console=hvc0` is required for getting boot logs on the host file and `console=tty0` is required for getting logs on the gui window. I have added this to the [boot messages document](https://github.com/crc-org/vfkit/pull/323/files#diff-82056274162c4579095370946492028b65db835273af54ba6976c3bf629f122bR54)

dustymabe:
@vyasgun after this week you can update that document to say FCOS already sets the console kernel arguments. I'd prefer if that open PR merges first though and a separate later PR update the text, pointing to this PR.

dustymabe:
@baude @vyasgun can you test out these images?

- https://builds.coreos.fedoraproject.org/prod/streams/testing-devel/builds/42.20250619.20.1/aarch64/fedora-coreos-42.20250619.20.1-applehv.aarch64.raw.gz
- https://builds.coreos.fedoraproject.org/prod/streams/testing-devel/builds/42.20250619.20.1/x86_64/fedora-coreos-42.20250619.20.1-applehv.x86_64.raw.gz

PR #3554, opened 2025-06-20
platforms.yaml: add serial console to proxmoxve

b-:
https://github.com/coreos/fedora-coreos-tracker/issues/1652#issuecomment-2992991313

### DIFF
--- a/platforms.yaml
+++ b/platforms.yaml
@@ -53,14 +53,6 @@ aarch64:
     kernel_arguments:
       - console=ttyAMA0,115200n8
       - console=tty0
-  packet:
-    # https://metal.equinix.com/developers/docs/resilience-recovery/serial-over-ssh/#limitations
-    grub_commands:
-      - serial --speed=115200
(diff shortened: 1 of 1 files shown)


dustymabe:
the diff here LGTM, but can you break out the deletion of packet to a separate commit?

b-:
> the diff here LGTM, but can you break out the deletion of packet to a separate commit?

done. sorry for the repeated force-pushing!

dustymabe:
Thanks!

I didn't realize there were two entries for packet (makes sense since they supported aarch64 and x86_64). Can you delete the aarch64 entry in that first commit too?

b-:
> Thanks!
> 
> I didn't realize there were two entries for packet (makes sense since they supported aarch64 and x86_64). Can you delete the aarch64 entry in that first commit too?

Done! Sorry, I missed that one there!

dustymabe (approved):
LGTM

PR #3559, opened 2025-06-23
overrides: Pin glibc to 2.41-5.fc42

travier:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1974

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,24 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  glibc:
+    evr: 2.41-5.fc42
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1974
+      type: pin
+  glibc-common:
+    evr: 2.41-5.fc42
+    metadata:
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1974
+      type: pin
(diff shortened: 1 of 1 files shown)


jlebon:
Shoudl we hold on this to see if it'll get reverted?

dustymabe:
> Shoudl we hold on this to see if it'll get reverted?

I think if we can reach out to them directly to see and we get a new build today then I'd be OK with that. Otherwise we need to ship so pinning would be acceptable.

dustymabe (approved):

dustymabe:
> I think if we can reach out to them directly to see and we get a new build today then I'd be OK with that.

This is probably unlikely. I'm going to set this to auto-merge.

dustymabe:
```
[2025-06-23T17:59:56.654Z] Downgraded:
[2025-06-23T17:59:56.654Z]   glibc 2.41-6.fc42 -> 2.41-5.fc42
[2025-06-23T17:59:56.654Z]   glibc-common 2.41-6.fc42 -> 2.41-5.fc42
[2025-06-23T17:59:56.654Z]   glibc-gconv-extra 2.41-6.fc42 -> 2.41-5.fc42
[2025-06-23T17:59:56.654Z]   glibc-minimal-langpack 2.41-6.fc42 -> 2.41-5.fc42
[2025-06-23T17:59:56.654Z] + grep -q Downgraded tmp/diff.txt
[2025-06-23T17:59:56.654Z] + echo 'Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.'
```

Thanks CI. We'll ignore since this was desired.

PR #3563, opened 2025-06-23
manifests/fedora-coreos: advance OCI migration to `testing` stream

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/1890#issuecomment-2881068761


jbtrystram (approved):
LGTM

PR #3564, opened 2025-06-25
platforms: add console support for Oracle Cloud Infrastructure

alaviss:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1967

### DIFF
--- a/platforms.yaml
+++ b/platforms.yaml
@@ -53,6 +53,14 @@ aarch64:
     kernel_arguments:
       - console=ttyAMA0,115200n8
       - console=tty0
+  oraclecloud:
+    # https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/enablingserialconsoleaccess.htm
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=ttyAMA0,115200
   qemu:
     # The kernel successfully autodetects a serial console, but we still
     # want GRUB to use one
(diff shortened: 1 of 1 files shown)


dustymabe (dismissed):
LGTM

platforms.yaml
```diff
@@ -53,6 +53,14 @@ aarch64:
...
       - console=ttyAMA0,115200n8
       - console=tty0
+  oraclecloud:
+    # https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/enablingserialconsoleaccess.htm
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=ttyAMA0,115200n8
```
dustymabe: where did you come up with the `n8` here? I don't see it mentioned specifically in the linked docs, so just curious.

dustymabe: and another reason I ask is because the x86_64 entry added below specifically doesn't have it.

dustymabe:
Thanks for working on this

platforms.yaml
alaviss: I booted the official Alma 9 image and found it in `/proc/cmdline`, so I figured it might be some undocumented quirk.

dustymabe: I'd be interested to know if it works without the `n8` just for consistency with x86_64 and the docs, but it's not necessarily a blocker here.

alaviss: I've confirmed that it works without `n8` and fixed the file accordingly

dustymabe (approved):

PR #3566, opened 2025-06-26
overrides: fast-track rust-bootupd-0.2.28-2.fc42

coreosbot-releng:
Requested by @HuijingHei via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -9,6 +9,11 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
+  bootupd:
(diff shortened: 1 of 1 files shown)


dustymabe:
Native podman build GHA needs to be updated to handle overrides I think?

```
error: Installing packages: Packages not found: bootupd-0.2.28-2.fc42
```

cc @jlebon

dustymabe (approved):
LGTM

PR #3567, opened 2025-06-26
overrides: fast-track glibc-2.41-8.fc42

coreosbot-releng:


dustymabe:
For the CI failure see https://github.com/coreos/fedora-coreos-config/issues/3568

dustymabe (approved):
LGTM

Issue #3568, opened 2025-06-27
Native podman build GHA needs to be updated to handle overrides I think?


dustymabe:
Native podman build GHA needs to be updated to handle overrides I think?

```
error: Installing packages: Packages not found: bootupd-0.2.28-2.fc42
```

cc @jlebon

_Originally posted by @dustymabe in https://github.com/coreos/fedora-coreos-config/issues/3566#issuecomment-3008963478_

jbtrystram:
it's already handling those : 

https://github.com/coreos/fedora-coreos-config/blob/a39f95471f57af8360f0a7e9872cf7c09faa9532/build-rootfs#L96-L99

Then create a `.repo` file with the values :
https://github.com/coreos/fedora-coreos-config/blob/a39f95471f57af8360f0a7e9872cf7c09faa9532/build-rootfs#L115-L126


The package needs to be in `coreos-pool` though, i am not sure where rpm-ostree pull the override from by default ?

HuijingHei:
Check [moby-engine-28.3.0-1.fc42](https://koji.fedoraproject.org/koji/buildinfo?buildID=2738710) is not in `coreos-pool` yet, and it does not have tag `coreos-pool`, so the job will fail if the build is not included in the repo (which is actually `fedora-coreos-pool`), is this right?

dustymabe:
Correct. The package isn't in `coreos-pool` because it's not actually officially apart of any coreos stream yet (that's what this PR is trying to do). 

Where we handle this in the `cosa` case is here:

https://github.com/coreos/fedora-coreos-config/blob/9a568544fc88011b4d7f627946da4a5ff73c2882/.cci.jenkinsfile#L20

jlebon:
Yeah, I think let's just move `download-overrides.py` from cosa to f-c-c? That'd require updating the pipeline too.

dustymabe:
OR (probably easier), just have the GHA download it from the cosa git repo

HuijingHei:
Seems the cosa `download-overrides.py` just download the packages using `koji`, maybe we also need to create repo and let rpm-ostree pick them like what we did in cosa (https://github.com/coreos/coreos-assembler/blob/main/src/cmdlib.sh#L447)?

jlebon:
> OR (probably easier), just have the GHA download it from the cosa git repo

[...]

PR #3569, opened 2025-06-27
overrides: fast-track moby-engine-28.3.0-1.fc42

coreosbot-releng:


dustymabe:
The GHA for container native podman build is going to fail: https://github.com/coreos/fedora-coreos-config/issues/3568

PR #3570, opened 2025-06-27
05core: fix expansion of OPTIONS variable when creating sysroot.mount

nikita-dubrovskii:


fedora-bootc
dustymabe, line 1: did you intend to do a bootc submodule bump?

nikita-dubrovskii: by mistake. should be fixed now

dustymabe:
I think the behavior here was intentional. We don't want to expand the `$OPTIONS` from the environment here. The `OPTIONS=` are defined in `coreos-rootflags.service` and are written to the envfile `/run/coreos-rootflags` and pulled into this unit at runtime by `EnvironmentFile=/run/coreos-rootflags`.

https://github.com/coreos/fedora-coreos-config/blob/9a568544fc88011b4d7f627946da4a5ff73c2882/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-rootflags.service#L19-L25

nikita-dubrovskii:
> I think the behavior here was intentional.

Thx for the explanation! Will close PR than

PR #3572, opened 2025-06-30
05core: Add weak dependency on coreo-multipath-wait.target for sysroot.mount

nikita-dubrovskii:
This fixes a race condition where the symlink `/dev/disk/by-label/root` was not updated in time by udev/multipathd for the device-mapper target `root`. As a result, the `sysroot.mount` unit occasionally failed to mount the root filesystem.

The issue was first observed on s390x builders under high system load, where `ext.config.multipath.resilient` would intermittently fail.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -89,6 +89,7 @@ if test -n "$(cmdline_arg ostree)" && test -z "$(cmdline_arg root)" && ! dracut_
 [Unit]
 Before=initrd-root-fs.target
 After=coreos-rootflags.service
+After=coreos-multipath-wait.target
 [Mount]
 EnvironmentFile=/run/coreos-rootflags
 What=/dev/disk/by-label/root
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
@@ -1,9 +1,11 @@
 [Unit]
-Description=CoreOS Wait For Multipathed Boot
+Description=CoreOS Wait For Multipathed Boot & Root
 DefaultDependencies=false
 After=dracut-cmdline.service
 Requires=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
 After=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
+Requires=dev-disk-by\x2dlabel-dm\x2dmpath\x2droot.device
+After=dev-disk-by\x2dlabel-dm\x2dmpath\x2droot.device
 Requires=multipathd.service
 After=multipathd.service
 


jbtrystram (approved):
LGTM

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -89,6 +89,7 @@ if test -n "$(cmdline_arg ostree)" && test -z "$(cmdline_arg root)" && ! dracut_
 [Unit]
 Before=initrd-root-fs.target
 After=coreos-rootflags.service
+After=coreos-multipath-wait.target
```
jlebon, line 92: Hmm, AIUI we're seeing the failure on the second boot though, where this is not used.

> ```
> # Note that for new machines this is *only* useful on first boot because
> # coreos-boot-edit.service will add the root= and rootflags= kargs to the
> # kernel command line for subsequent boots.
> ```

(Notably, notice the `test -z "$(cmdline_arg root)"`.)

So this should have no effect.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-multipath-wait.target
```diff
@@ -1,9 +1,11 @@
 [Unit]
-Description=CoreOS Wait For Multipathed Boot
+Description=CoreOS Wait For Multipathed Boot & Root
 DefaultDependencies=false
 After=dracut-cmdline.service
 Requires=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
 After=dev-disk-by\x2dlabel-dm\x2dmpath\x2dboot.device
+Requires=dev-disk-by\x2dlabel-dm\x2dmpath\x2droot.device
+After=dev-disk-by\x2dlabel-dm\x2dmpath\x2droot.device
```
jlebon, line 8: It's not a given that this will exist. If e.g. you have LUKS on top of multipath, it will not.

jlebon (commented):
Looking at the test failure logs, I'm not actually seeing sysroot getting mounted at all. But `coreos-propagate-multipath-conf.service` clearly has [`After=initrd-root-fs.target`](https://github.com/coreos/fedora-coreos-config/blob/55fa641e4960745aa026f9c29baf2ec604428493/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-propagate-multipath-conf.service#L6) which is by definition ordered after the auto-generated `sysroot.mount`.

And actually, I'm not even seeing something like:

> Reached target initrd-root-fs.target - Initrd Root File System.

Hmm, I wonder if somehow something changed in the systemd-fstab-generator's `sysroot.mount`? If you can reproduce this locally, can you sanity-check that it still has `Before=initrd-root-fs.target`?

I mean, I guess a more heavy-handed approach is to just do

```diff
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-propagate-multipath-conf.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-multipath/coreos-propagate-multipath-conf.service
@@ -3,7 +3,8 @@ Description=CoreOS Propagate Multipath Configuration
 Before=initrd.target

 # we write to the rootfs, so run after it's ready
-After=initrd-root-fs.target
+After=ostree-prepare-root.service
+Requires=ostree-prepare-root.service

 # we only propagate if multipath wasn't configured via Ignition
 After=ignition-files.service
```

The `initrd-root-fs.target` interface was meant to be more generic, but in the end that's really what we care about.

nikita-dubrovskii:
@jlebon , 
>Looking at the test failure logs, I'm not actually seeing sysroot getting mounted at all. 

Here it is, on subsequent boot:
```
[    0.124018] Kernel command line: rw mitigations=auto,nosmt ostree=/ostree/boot.1/fedora-coreos/962bc5726d2e17c164cbcebc267d644699e2bb1be024dd13383d71352b40bacf/0 ignition.platform.id=qemu rd.multipath=default root=UUID=394b4215-e868-48e8-ae44-9022a9697aa7 rw rootflags=prjquota boot=UUID=f1769418-791a-4ce8-af8c-c6f1a28c9edc
...
[    2.781559] multipathd[321]: sdd [8:48]: path added to devmap 0xcadf6fadb3ee446d
[    2.853163] multipathd[321]: sdb [8:16]: path added to devmap 0x000000000000000b

[  OK  ] Found device dev-disk-by\x2dlabel-…- /dev/disk/by-label/dm-mpath-boot.
[    3.012431] systemd[1]: Reached target coreos-multipath-wait.target - CoreOS Wait For Multipathed Boot.
[  OK  ] Reached target coreos-multipath-wa…- CoreOS Wait For Multipathed Boot.
[  OK  ] Reached target cryptsetup-pre.target - Local Encrypted Volumes (Pre).
[  OK  ] Reached target cryptsetup.target - Local Encrypted Volumes.
...
[    3.137674] systemd[1]: Finished systemd-fsck-root.service - File System Check on /dev/disk/by-uuid/394b4215-e868-48e8-ae44-9022a9697aa7.
[    3.139605] systemd[1]: Mounting sysroot.mount - /sysroot...
         Mounting sysroot.mount - /sysroot...
[    3.450666] mount[806]: mount: /sysroot: fsconfig system call failed: /dev/sdd4: Can't open blockdev.
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
nikita-dubrovskii: Indeed! Checked with patched `coreos-multipath-wait.target` (`--multiply 50`) and it woked!

nikita-dubrovskii:
>Hmm, I wonder if somehow something changed in the systemd-fstab-generator's sysroot.mount? If you can reproduce this locally, can you sanity-check that it still has Before=initrd-root-fs.target?

It has: 
```
# /run/systemd/generator/sysroot.mount
# Automatically generated by systemd-fstab-generator
[Unit]
Documentation=man:fstab(5) man:systemd-fstab-generator(8)
SourcePath=/proc/cmdline
Before=initrd-root-fs.target
Requires=systemd-fsck-root.service
After=systemd-fsck-root.service
```

jlebon:
> ```
> [    3.139605] systemd[1]: Mounting sysroot.mount - /sysroot...
>          Mounting sysroot.mount - /sysroot...
> [    3.450666] mount[806]: mount: /sysroot: fsconfig system call failed: /dev/sdd4: Can't open blockdev.
> ```

OK interesting. So this looks like a clearer race between I guess multipathd taking ownership of root and systemd trying to mount sysroot. The assumption this whole time was that it happened at the same time for boot and root. Possibly e.g. the udev database hasn't been updated yet or something?

This becomes tricky to fix. I think probably instead at this point we should just fix `rdcore rootmap` to, when the rootfs is directly on multipath, instead of injecting `root=UUID=...`, inject `root=/dev/disk/by-id/dm-mpath-$UUID`. That would not only fix this, but actually drop the need for `coreos-multipath-wait.target` post-first boot at all.

nikita-dubrovskii:
closing, fix comes with https://github.com/coreos/coreos-installer/pull/1677

PR #3574, opened 2025-07-03
manifests/fedora-coreos-base: drop ssh-host-keys-migration code

dustymabe:
This forced code provided by openssh-server to run on every boot for a specific reason [1]. We've had plenty of barrier releases so we can drop this code now.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1473


jlebon (approved):

jbtrystram (approved):
LGTM

PR #3576, opened 2025-07-05
overrides: fast-track sudo-1.9.17-2.p1.fc42

coreosbot-releng:


dustymabe:
Container Native GHA failure tracked in https://github.com/coreos/fedora-coreos-config/issues/3568

PR #3581, opened 2025-07-06
ci: add step in GHA `container-native.yml` to download overrides packages

### LINKED ISSUE #3568: Native podman build GHA needs to be updated to handle overrides I think?

HuijingHei:
The new step will download override packages and create local repo, which will be injected in the repo by next step.
Fixes https://github.com/coreos/fedora-coreos-config/issues/3568

### DIFF
--- a/.github/workflows/container-native.yml
+++ b/.github/workflows/container-native.yml
@@ -24,6 +24,25 @@ jobs:
           sudo apt install -y crun/testing podman/testing skopeo/testing
           # Something is confused in latest GHA here
           sudo rm /var/lib/containers -rf
+      - name: Download override packages and create yum repo
+        run: |
+          podman run --rm -v $PWD:/run/src -w /run/src quay.io/fedora/fedora:latest sh -c '
+            set -xeuo pipefail
+            dnf -y install koji createrepo_c python3-dnf python3-PyYAML && dnf clean all
+            overridesdir="rpmoverrides"
+
+            # prepare directory to download override packages
+            [ -d ${overridesdir} ] && rm -rf ${overridesdir}
+            mkdir -p "${overridesdir}"
+
+            curl -LO https://raw.githubusercontent.com/coreos/coreos-assembler/refs/heads/main/src/download-overrides.py
+            python3 download-overrides.py --downloaddir "./${overridesdir}" --lockfiledir "./"
+
+            # create local yum repo
+            if [[ -n $(ls "${overridesdir}/"*.rpm 2> /dev/null) ]]; then
+                cd "${overridesdir}" && createrepo_c .
+            fi
+          '
       - name: Build
         # Note: we should be able to drop the `-v $PWD:/run/src` once
         # https://github.com/containers/buildah/issues/5952 is fixed.
--- a/build-rootfs
+++ b/build-rootfs
@@ -33,6 +33,7 @@ def main():
     locked_nevras = get_locked_nevras()
     if locked_nevras:
         inject_pool_repo_if_exists(locked_nevras)
+        create_and_inject_local_rpm_overrides_repo_if_needed(locked_nevras)
 
     packages.extend(locked_nevras)
     overlays = gather_overlays(manifest)
@@ -47,6 +48,24 @@ def main():
     if version != "":
         inject_version_info(target_rootfs, manifest['mutate-os-release'], version)
     inject_postprocess_scripts(target_rootfs, manifest)
+
+
+def create_and_inject_local_rpm_overrides_repo_if_needed(locked_nevras):
+    rpmoverrides_dir = os.path.join(CONTEXTDIR, 'rpmoverrides')
+    if not os.path.exists(rpmoverrides_dir):
+         return
+
+    # create local-overrides.repo
+    packages = ','.join(locked_nevras)
+    with open("/etc/yum.repos.d/local-overrides.repo", 'w') as f:
+        f.write(f"""
+[local-overrides]
+name=local-overrides
+baseurl=file://{rpmoverrides_dir}/
+gpgcheck=0
+cost=500
+includepkgs={packages}
+""")
 
 
 def get_treefile(manifest_path):


Containerfile.override
```diff
@@ -0,0 +1,51 @@
...
+
+RUN <<EORUN
+set -xeuo pipefail
+workdir=/run/src
+overridesdir="${workdir}/overrides"
+tmp_overridesdir=tmpoverrides
+
+# prepare directories to download override packages
+mkdir -p ${tmp_overridesdir} builds src/config
+cp manifest-lock.overrides.yaml src/config
```
HuijingHei: `download-overrides.py` requires `builds` and `src/config`
https://github.com/coreos/coreos-assembler/blob/35032cc1bc591710ef6ef4bfe5823b388835a522/src/download-overrides.py#L44 & https://github.com/coreos/coreos-assembler/blob/35032cc1bc591710ef6ef4bfe5823b388835a522/src/download-overrides.py#L48, should I update in coreos-assembler to remove the requirements and just add info if not existing?

jbtrystram: > should I update in coreos-assembler to remove the requirements and just add info if not existing?

Yes. I think the general rule of thumb we should have is that we should go and ajust tooling to support the container native flow wherever needed rather than hacking workarounds. Because the `cosa build` path will be deprecated anyway

HuijingHei: SGTM. Thank you!

HuijingHei:
This PR requires https://github.com/coreos/coreos-assembler/pull/4183.

build-rootfs
```diff
@@ -32,7 +32,7 @@ def main():
 
     locked_nevras = get_locked_nevras()
     if locked_nevras:
-        inject_pool_repo_if_exists(locked_nevras)
+        inject_overrides_repo_if_exists()
```
dustymabe: I don't think you meant to delete `inject_pool_repo_if_exists(locked_nevras)` here.

dustymabe:
I think in this PR we are duplicating some code that we don't really need to because build-rootfs already has some handling for locked nevras already.

What if instead of trying to build up the includepkgs in the `podman build` we just let the build-rootfs script do that for us?

```diff
diff --git a/build-rootfs b/build-rootfs
index 05769f5e..c4b50e55 100755
--- a/build-rootfs
+++ b/build-rootfs
@@ -32,7 +32,8 @@ def main():
 
     locked_nevras = get_locked_nevras()
     if locked_nevras:
-        inject_overrides_repo_if_exists()
+        inject_pool_repo_if_exists(locked_nevras)
+        create_and_inject_local_rpm_overrides_repo_if_needed(locked_nevras)
 
     packages.extend(locked_nevras)
     overlays = gather_overlays(manifest)
@@ -49,12 +50,23 @@ def main():
     inject_postprocess_scripts(target_rootfs, manifest)
 
 
-def inject_overrides_repo_if_exists():
-    overrides_repo = os.path.join(CONTEXTDIR, "local-overrides.repo")
-    if not os.path.exists(overrides_repo):
+def create_and_inject_local_rpm_overrides_repo_if_needed(locked_nevras):
+    rpmoverrides_dir = os.path.join('.', 'rpmoverrides')
+    if not os.path.exists(rpmoverrides_dir)
         return
-    shutil.copyfile(overrides_repo, "/etc/yum.repos.d/local-overrides.repo")
-
+    # When adding the local overrides, we only want to _filter in_ locked packages;
+    # matching `lockfile-repos` semantics. This is abusing pretty hard the
+    # `includepkgs=` semantic but... it works.
+    packages = ','.join(locked_nevras)
+    with open("/etc/yum.repos.d/local-overrides.repo", 'w') as f:
+        f.write(f"""
+[local-overrides]
+name=local-overrides
+baseurl=file://${rpmoverrides_dir}/
+gpgcheck=0
+cost=500
+includepkgs=${packages}
+""")
 
 def get_treefile(manifest_path):
     with tempfile.NamedTemporaryFile(suffix='.json', mode='w') as tmp_manifest:
```

build-rootfs
HuijingHei: Nice catch! You are correct, I should not delete it, just narrowly think about the overrides packages. Your change looks good to me, and the local-override repo will win as it has high priority with cost 500. Thank you for the kind review!

Containerfile.override
```diff
@@ -0,0 +1,30 @@
...
+python3 download-overrides.py --downloaddir "./${overrides}" --lockfiledir "./"
+
+arch=$(arch)
+# Based on https://github.com/coreos/coreos-assembler/blob/35032cc1bc591710ef6ef4bfe5823b388835a522/src/cmdlib.sh#L446,
+# and simplified to create local repo
+if [[ -n $(ls "${overridesdir}/"*.rpm 2> /dev/null) ]]; then
+    (cd "${overridesdir}" && rm -rf .repodata && createrepo_c .)
+    dnf repoquery  --repofrompath=tmp,"file://${overridesdir}/" \
+        --disablerepo '*' --enablerepo tmp --refresh --latest-limit 1 \
+        --arch "$arch,noarch" --quiet
```
dustymabe: What's the point of this command? Just to print out what's in the repo? and thus verify the repo creation worked?

Containerfile.override
```diff
@@ -0,0 +1,30 @@
...
+mkdir -p "${overrides}"
+
+curl -LO https://raw.githubusercontent.com/coreos/coreos-assembler/refs/heads/main/src/download-overrides.py
+python3 download-overrides.py --downloaddir "./${overrides}" --lockfiledir "./"
+
+arch=$(arch)
+# Based on https://github.com/coreos/coreos-assembler/blob/35032cc1bc591710ef6ef4bfe5823b388835a522/src/cmdlib.sh#L446,
+# and simplified to create local repo
+if [[ -n $(ls "${overridesdir}/"*.rpm 2> /dev/null) ]]; then
+    (cd "${overridesdir}" && rm -rf .repodata && createrepo_c .)
```
dustymabe: Is this in a subshell for a reason? (i.e. the `(` and `)`?

.github/workflows/container-native.yml
```diff
@@ -24,6 +24,8 @@ jobs:
           sudo apt install -y crun/testing podman/testing skopeo/testing
           # Something is confused in latest GHA here
           sudo rm /var/lib/containers -rf
+      - name: Download override packages
```
dustymabe:
```suggestion
      - name: Download override packages and create yum repo
```

Containerfile.override
HuijingHei: Not needed anymore, remove this.

Containerfile.override
HuijingHei: Not needed, remove the subshell ().

dustymabe (dismissed):
LGTM

Containerfile.override
```diff
@@ -0,0 +1,27 @@
+FROM quay.io/fedora/fedora:42
+
+WORKDIR /run/src
+USER root
+
+RUN dnf -y install koji createrepo_c python3-createrepo_c python3-dnf python3-PyYAML && dnf clean all
+COPY download-overrides.py .
```
dustymabe: I think we need to remove this line since we are `curl`ing it below.

```suggestion
```

Containerfile.override
```diff
@@ -0,0 +1,27 @@
...
+overrides="rpmoverrides"
+overridesdir="${workdir}/${overrides}"
+
+# prepare directory to download override packages
+mkdir -p "${overrides}"
+
+curl -LO https://raw.githubusercontent.com/coreos/coreos-assembler/refs/heads/main/src/download-overrides.py
+python3 download-overrides.py --downloaddir "./${overrides}" --lockfiledir "./"
+
+arch=$(arch)
```
dustymabe: dead code?

dustymabe:
I was about to merge but noticed the `Container-native/Podman Build` GHA was failing and I think we need to do https://github.com/coreos/fedora-coreos-config/pull/3581#discussion_r2193434172 to get it to pass.

Containerfile.override
dustymabe: i.e. do we use $arch anywhere?

Containerfile.override
```diff
@@ -0,0 +1,27 @@
+FROM quay.io/fedora/fedora:42
```
jlebon: Wait, sorry. Why are we modeling this as a Container build? This could just be a podman run that does the same things, no?

I'd like to avoid adding another top-level Containerfile for this, especially for a niche case like this. I consider top-level Containerfiles special since they have high visibility and form a sort of soft API for any git repos which output container images. It's quite obvious when you have just one, but once you have _two_, the questions becomes what does each one do?

At the very least, let's just move it to the `ci/` dir. But yeah, ideally unless there's a reason I'm missing (some kind of smart caching?), let's just do this in a e.g. `podman run -v $PWD:/run/src quay.io/fedora/fedora:42 /run/src/ci/download-overrides` instead.

dustymabe: Yeah. I didn't realize the Containerfile was in the root directory of the repo.

I also agree we can just do it in a `podman run`. Maybe a `podman run <<EOF/EOF` even so we can embed the script in the GHA yaml.

Containerfile.override
HuijingHei: Nice catch, remove the dead code.

Containerfile.override
HuijingHei: Sorry to forget to remove the debug code.

Containerfile.override
HuijingHei: > I also agree we can just do it in a podman run. Maybe a podman run <<EOF/EOF even so we can embed the script in the GHA yaml.

Great catch, thank you @jlebon and @dustymabe for the nice suggestion!
Update to `podman run --rm -v $PWD:/run/src -w /run/src quay.io/fedora/fedora:42 sh -c '<script>'`

.github/workflows/container-native.yml
```diff
@@ -24,6 +24,25 @@ jobs:
           sudo apt install -y crun/testing podman/testing skopeo/testing
           # Something is confused in latest GHA here
           sudo rm /var/lib/containers -rf
+      - name: Download override packages and create yum repo
+        run: |
+          podman run --rm -v $PWD:/run/src -w /run/src quay.io/fedora/fedora:42 sh -c '
```
dustymabe: Let's just use the unversioned moving tag here so we never have to update it. 


```suggestion
          podman run --rm -v $PWD:/run/src -w /run/src quay.io/fedora/fedora:latest sh -c '
```

dustymabe (dismissed):
LGTM one suggestion

dustymabe (approved):

.github/workflows/container-native.yml
```diff
@@ -24,6 +24,25 @@ jobs:
...
+            [ -d ${overridesdir} ] && rm -rf ${overridesdir}
+            mkdir -p "${overridesdir}"
+
+            curl -LO https://raw.githubusercontent.com/coreos/coreos-assembler/refs/heads/main/src/download-overrides.py
+            python3 download-overrides.py --downloaddir "./${overridesdir}" --lockfiledir "./"
+
+            # create local yum repo
+            if [[ -n $(ls "${overridesdir}/"*.rpm 2> /dev/null) ]]; then
+                cd "${overridesdir}" && createrepo_c .
+            fi
```
jlebon, line 44: Hmm, here, we're only creating repodata if there are RPMs, but in `build-rootfs`, we're conditional on the directory just existing, which it always will. So wouldn't this fail if there are no overrides?

I guess the easiest fix is to just nuke the directory in an `else` block.

build-rootfs
```diff
@@ -49,6 +50,24 @@ def main():
     inject_postprocess_scripts(target_rootfs, manifest)
 
 
+def create_and_inject_local_rpm_overrides_repo_if_needed(locked_nevras):
```
jlebon, line 53: Cool with this, but another way we could structure that is to have the GHA generate the repo file and then the podman build can use [the existing secrets API](https://github.com/coreos/fedora-coreos-config/blob/3986c33a5099b1a95df25ca92b303cb688a82868/Containerfile#L26) to inject its definition. (I.e. `podman build --secret id=yumrepos,src=rpmoverrides/overrides.repo ...`.)

HuijingHei: LGTM, the change is in https://github.com/coreos/fedora-coreos-config/pull/3587, not sure if my understanding is correct.

PR #3583, opened 2025-07-07
build-rootfs: temporarily inject our passwd/group files

jlebon:
We want to keep our passwd/group files as canonical for now to eliminate
any changes in our currently fixed UIDs/GIDs as part of the rebase to
the bootc base images.

Obviously not ideal. The medium-term fix for this is to convert our
passwd and group files to sysusers entries instead, and then force
bootc-base-imagectl to treat sysusers as canonical.

See also: https://github.com/coreos/rpm-ostree/pull/5427

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -13,6 +13,8 @@ FROM ${BUILDER_IMG} as builder
 
 ARG VERSION=overridden
 ARG MANIFEST=overridden
+# XXX: see inject_passwd_group() in build-rootfs
+ARG PASSWD_GROUP_DIR
 
 # useful if you're hacking on rpm-ostree/bootc-base-imagectl
 # COPY rpm-ostree /usr/bin/
--- a/Containerfile
+++ b/Containerfile
@@ -6,15 +6,13 @@
 # Note: we should be able to drop the `-v $PWD:/run/src` once
 # https://github.com/containers/buildah/issues/5952 is fixed.
 
-# Overridden by argfile.conf. The values here are invalid on purpose.
-ARG VERSION=overridden
+# Overridden by argfile.conf. The value here in invalid on purpose.
 ARG BUILDER_IMG=overridden
-ARG MANIFEST=overridden
 
 FROM ${BUILDER_IMG} as builder
 
-ARG VERSION
-ARG MANIFEST
+ARG VERSION=overridden
+ARG MANIFEST=overridden
 
 # useful if you're hacking on rpm-ostree/bootc-base-imagectl
 # COPY rpm-ostree /usr/bin/
--- a/build-rootfs
+++ b/build-rootfs
@@ -67,6 +67,9 @@ def get_treefile(manifest_path):
 
 
 def build_rootfs(target_rootfs, manifest_path, packages, overlays, nodocs):
+    passwd_group_dir = os.getenv('PASSWD_GROUP_DIR')
+    if passwd_group_dir is not None:
+        inject_passwd_group(os.path.join(CONTEXTDIR, passwd_group_dir))
     with tempfile.NamedTemporaryFile(mode='w') as argsfile:
         for pkg in packages:
             argsfile.write(f"--install={pkg}\n")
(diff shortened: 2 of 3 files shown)


travier (dismissed):
LGTM. Not tested

build-rootfs
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
```
dustymabe: Wish there was a long option for this because the short option requires people to look it up if they don't know :(

alternatively setting PYTHONUNBUFFERED env var will do it I think, but it's not very ergonomic.

No need to make any changes here, just me thinking out loud.

Containerfile
```diff
@@ -6,10 +6,8 @@
 # Note: we should be able to drop the `-v $PWD:/run/src` once
 # https://github.com/containers/buildah/issues/5952 is fixed.
 
-# Overridden by argfile.conf. The values here are invalid on purpose.
-ARG VERSION=overridden
+# Overridden by argfile.conf. The value here in invalid on purpose.
 ARG BUILDER_IMG=overridden
-ARG MANIFEST=overridden
```
dustymabe, line 12: Did this commit solve any problems or was it purely cosmetic? For completeness, I kind of like it the way it was.

Containerfile
```diff
@@ -13,6 +13,8 @@ FROM ${BUILDER_IMG} as builder
 
 ARG VERSION
 ARG MANIFEST
+# XXX: see inject_passwd_group() in build-rootfs
+ARG PASSWD_GROUP_DIR
```
dustymabe, line 17: Why not:

```
ARG PASSWD_GROUP_DIR=overridden
```

I guess it's because it would set a value where we have conditional logic later that does different things based on set or unset.

dustymabe: I guess another question here.. should we make this conditional, i.e. should we require `$PASSWD_GROUP_DIR` to be set in `build-rootfs` and not allow it to be empty?

Containerfile
dustymabe: alternatively we can just add the "decorations" to the lines below at around line 14 and drop them here since they aren't used for the initial `FROM`

dustymabe (commented):
No blockers, just a few comments.

build-rootfs
travier: Agree. I had to look it up as well.

jlebon: I dropped this actually and just used `flush=True` on the print statement. I _think_ it was leaving my terminal in a weird state if I interrupted it at the wrong time.

Containerfile
jlebon: Yeah, good idea. Kept the `overridden` bit below.

It was mostly cosmetic, but also I don't want to give the impression that that list is complete since that's more `build-args.conf`.

Otherwise, when adding new build args, we'll always have to add it in three places (the args files, at the top of the Containerfile, and in the right scope in the Containerfile). This reduces it to two.

Containerfile
jlebon: I'm hoping to be able to drop that hack in FCOS much sooner than RHCOS, which would require it being conditional.

Containerfile
```diff
@@ -6,15 +6,15 @@
 # Note: we should be able to drop the `-v $PWD:/run/src` once
 # https://github.com/containers/buildah/issues/5952 is fixed.
 
-# Overridden by argfile.conf. The values here are invalid on purpose.
-ARG VERSION=overridden
+# Overridden by argfile.conf. The value here in invalid on purpose.
```
dustymabe, line 9:
```suggestion
# Overridden by build-args.conf. The value here is invalid on purpose.
```

Containerfile
```diff
@@ -6,15 +6,15 @@
...
 ARG BUILDER_IMG=overridden
-ARG MANIFEST=overridden
 
 FROM ${BUILDER_IMG} as builder
```
dustymabe, line 13: should we mention (comment) above these that they are overridden by build-args.conf too?

dustymabe (approved):
LGTM, some suggestions about comments.

Containerfile
jlebon: Ahh yes. Got another PR coming soon, so will fix it there to not waste CI.

Containerfile
jlebon: I thought about that but since they're so close to each other and use the same `overridden` pattern, I thought it was clear enough that the same rationale carries over.

PR #3584, opened 2025-07-08
Add name as label in container image

PeaceRebel:
meta.json needs name field when cosa imports the oci image built with the Containerfile

Related to: https://github.com/coreos/coreos-assembler/pull/4164

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -34,6 +34,7 @@ RUN --mount=type=bind,target=/run/src,rw \
 
 FROM oci-archive:./out.ociarchive
 ARG VERSION
+ARG NAME=overridden
 # Need to reference builder here to force ordering. But since we have to run
 # something anyway, we might as well cleanup after ourselves.
 RUN --mount=type=bind,from=builder,target=/var/tmp \
@@ -51,5 +52,6 @@ EOF
 LABEL containers.bootc=1
 LABEL ostree.bootable=1
 LABEL org.opencontainers.image.version=$VERSION
+LABEL com.coreos.osname=$NAME
 STOPSIGNAL SIGRTMIN+3
 CMD ["/sbin/init"]
--- a/build-args.conf
+++ b/build-args.conf
@@ -9,3 +9,4 @@ BUILDER_IMG=quay.io/fedora/fedora-bootc:42
 MANIFEST=manifest.yaml
 # XXX: see inject_passwd_group() in build-rootfs
 PASSWD_GROUP_DIR=manifests
+NAME=fedora-coreos
--- a/build-args.conf
+++ b/build-args.conf
@@ -2,7 +2,7 @@
 # Pass to buildah/podman using `--build-arg-file`.
 
 # This is the developer default version. In pipelines, this is driven by versionary.
-VERSION="42"
+VERSION=42
 # XXX: This should be a digested pull that gets bumped.
 # https://gitlab.com/fedora/bootc/tracker/-/issues/34
 BUILDER_IMG=quay.io/fedora/fedora-bootc:42


travier (dismissed):

jbtrystram (dismissed):

build-args.conf
```diff
@@ -2,8 +2,9 @@
 # Pass to buildah/podman using `--build-arg-file`.
 
 # This is the developer default version. In pipelines, this is driven by versionary.
-VERSION="42"
+VERSION=42
```
dustymabe, line 5: Was this change necessary?

If not and it was purely for consistency or cosmetic benefit maybe we put it in a separate commit?

PeaceRebel: Yeah, the double quotes are kept as is in the label. When using this arg as buildid it was weird. I thought changing it here is the better option

dustymabe: ahh. good to know. So it is making an effective change. Could you add a separate commit?

I bet this is the same underlying problem that [I ran in to recently](https://github.com/jlebon/coreos-pipeline-assistant/commit/0233a6aa4f7a22ad05aec754e4a700b2e0e1126c). Maybe a similar commit message here linking to the upstream issues would be useful.

dustymabe:
For some reason when I try to build this locally I'm getting an error, but I don't know why. Trying to debug it now.

PeaceRebel:
> For some reason when I try to build this locally I'm getting an error, but I don't know why. Trying to debug it now.

Hmm... I just tried. Build succeeds for me.

dustymabe:
ok I worked around my local build problem..

The resulting built image doesn't have labels set right:

```
+ rm /usr/libexec/coreos-postprocess-8
+ for script in /usr/libexec/coreos-postprocess-*
+ /usr/libexec/coreos-postprocess-9
+ chmod 440 /etc/sudoers.d/coreos-sudo-group
+ rm /usr/libexec/coreos-postprocess-9
--> dc14be76e2c1
[2/2] STEP 6/11: LABEL containers.bootc=1
--> 9f7ac75395c7
[2/2] STEP 7/11: LABEL ostree.bootable=1
--> 87cc3b6ad87f
[2/2] STEP 8/11: LABEL org.opencontainers.image.version=$VERSION
--> f0e1d33b76e9
[2/2] STEP 9/11: LABEL name=$NAME
--> 8f53e847a464
[2/2] STEP 10/11: STOPSIGNAL SIGRTMIN+3
--> e03094cd5435
[2/2] STEP 11/11: CMD ["/sbin/init"]
[2/2] COMMIT localhost/fedora-coreos
--> f3397aa39a92
Successfully tagged localhost/fedora-coreos:latest
f3397aa39a92fe2a645104325500b2feb82c80af1cffe740cecd5acf57be7fa3

[core@cosa-devsh config]$ podman inspect localhost/fedora-coreos:latest | jq .[0].Labels
{
  "containers.bootc": "1",
  "io.buildah.version": "1.40.1",
  "name": "",
  "org.opencontainers.image.version": "42",
  "ostree.bootable": "1",
  "ostree.commit": "5ca75bb87b87481bf69e6807c0e8502032519188bf3b6c83709795037b209912",
  "ostree.final-diffid": "sha256:12787d84fa137cd5649a9005efe98ec9d05ea46245fdc50aecb7dd007f2035b1"
}

```

I think you need to play around with your scoping of the NAME arg. i.e. add it to line 37 under the `ARG VERSION`.

dustymabe (dismissed):
needs some changes for the change to have an effect

Containerfile
```diff
@@ -51,5 +52,6 @@ EOF
 LABEL containers.bootc=1
 LABEL ostree.bootable=1
 LABEL org.opencontainers.image.version=$VERSION
+LABEL name=$NAME
```
jlebon: Let's have a label a bit more specific here. E.g. `com.coreos.osname`? OK, or how about using a predefined OCI spec label for this: `org.opencontainers.image.title`. (This abuses the spec a bit because that field is supposed to be human-readable, but it's not really that far from the human version either.)

PeaceRebel:
> needs some changes for the change to have an effect

Yes, found it.

jlebon (approved):

PR #3585, opened 2025-07-09
tests: update multipath.resilent to support full root path

nikita-dubrovskii:
Related changes: https://github.com/coreos/coreos-installer/pull/1677


jlebon (approved):

jlebon:
Can you sanity-check that all multipath-related tests pass with https://github.com/coreos/coreos-installer/pull/1677 ? (E.g. `kola run *multipath*`). We might need to adapt others.

PR #3589, opened 2025-07-10
build-rootfs: add workaround for RHEL-97826

jlebon:
bootc owns files in `/usr/share/doc` which are correctly marked as documentation, but because of a bug in rpm-ostree, the subdirectories are still installed even with `documentation: false`. This throws off `bootc container lint`.

This is fixed in newer bootc, but not RHEL 9.6.

Work around this there by checking if documentation is turned off whether the issue is fixed, and if not, just "pass through" the docs into the installed target via another overlay. This then will only activate when building RHCOS 9.6. (Also c9s currently until bootc is rebased there.)

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -6,7 +6,7 @@
 # Note: we should be able to drop the `-v $PWD:/run/src` once
 # https://github.com/containers/buildah/issues/5952 is fixed.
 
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
Looks sane to me

dustymabe (approved):

PR #3592, opened 2025-07-10
[stable] overrides: pin moby-engine-28.3.0-1.fc42

coreosbot-releng:


jlebon:
I guess this is technically a fast-track, but it doesn't matter for stable anyway since it gets stomped over.

dustymabe (approved):
LGTM

marmijo (approved):

PR #3598, opened 2025-07-15
Containerfile: run postprocess scripts in build stage

jlebon:
To have optimal layers, let's run our postprocess scripts in the build stage _before_ chunking occurs. That way, the chunking operates on the final flattened tree and we don't carry around a fat layer at the end with lots of whiteouts.

For example, we nuke documentation as a postprocess script, which meant that we were actually still carrying all the docs in the lower layers. (On that topic, we should switch to `documentation: false` really.) But also e.g. we recompile the SELinux policy, edit a bunch of files, etc.

This also fixes a bug in 29ac01e6 ("Containerfile: include rechunking step"), in which I forgot to remove a `COPY` statement that copies the whole rootfs on top of the rechunked image. Ouch.

This uses bwrap to run the scripts. Another way to do this of course is to go back to what we had before 29ac01e6 (where we copied the rootfs from the previous stage into a scratch rootfs and then ran the postprocess scripts) and _then_ rechunk. The huge advantage doing it the bwrap way is that it avoids that expensive full rootfs copy.

With this change, the image size in the container-native and non-native paths are comparable.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -41,14 +41,6 @@ RUN --mount=type=bind,from=builder,target=/var/tmp \
     --mount=type=bind,target=/run/src,rw \
       rm /run/src/out.ociarchive
 
-COPY --from=builder /target-rootfs/ /
-RUN <<EOF
-set -xeuo pipefail
-for script in /usr/libexec/coreos-postprocess-*; do
-  $script; rm $script
-done
-EOF
-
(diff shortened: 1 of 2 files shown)


dustymabe (approved):
I didn't test this myself, but LGTM otherwise.

PR #3601, opened 2025-07-17
denylist: mark ext.config.rpm-ostree.kernel-replace as 'warn:true'

nikita-dubrovskii:
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1984


nikita-dubrovskii:
Proper fix was merged - https://github.com/coreos/coreos-installer/pull/1682

PR #3603, opened 2025-07-18
manifests: remove selinux workaround for F41+

joelcapitao:


dustymabe:
Any context you could add to this one? I'm not sure we can remove them.

If we are able to we'd also want to remove the yaml file too.

joelcapitao:
> Any context you could add to this one? I'm not sure we can remove them.
> 
> If we are able to we'd also want to remove the yaml file too.

This was a test to check if the latest `selinux-policy` rpm was fixing this w/o the workaround.

PR #3606, opened 2025-07-21
Use sysusers only at compose-time for container-native path

jlebon:
fedora-bootc today defines its own passwd/group files which were originally based on the Fedora CoreOS ones (which in turn came from Fedora Atomic).

The main problem with that is that we have our own passwd/group files which we want to be able to use. There is no interface for passing this to bootc-base-imagectl and it would be awkward to add one. Hence the `PASSWD_GROUP_DIR` hack.

More recently, `bootc-base-imagectl` learned a new `--sysusers` option in which we can opt-out of the centralized passwd/group file and have full control over UID allocation via sysusers dropins. Use it. Also use the hidden `--nobody-99` option for backwards compatibility.

All the entries in our passwd/group files are already present in our sysusers dropins, so this is in fact functionally equivalent. Notably this does not change anything about nss-altfiles. The entries in `/usr/lib/{passwd,group}` remain the exact same.

This allows us to stop using the `PASSWD_GROUP_DIR` hack for FCOS at least. For RHCOS, we'll have to keep it for RHEL 9.6. But because buildah prints a warning if a build arg is undefined, set the default value to "none".

Note this does not affect the legacy cosa build path. The in-tree passwd/group files are still used there.

For more information, see:
- https://github.com/coreos/rpm-ostree/pull/5427
- https://gitlab.com/fedora/bootc/base-images/-/merge_requests/242
- https://gitlab.com/fedora/bootc/base-images/-/merge_requests/243

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -21,7 +21,7 @@ FROM ${BUILDER_IMG} as builder
 ARG VERSION=overridden
 ARG MANIFEST=overridden
 # XXX: see inject_passwd_group() in build-rootfs
-ARG PASSWD_GROUP_DIR
+ARG PASSWD_GROUP_DIR=none
 ARG STRICT_MODE=0
 
 COPY . /src
--- a/build-rootfs
+++ b/build-rootfs
@@ -112,8 +112,8 @@ def inject_yumrepos():
 
 
 def build_rootfs(target_rootfs, manifest_path, packages, locked_nevras, overlays, repos, nodocs):
-    passwd_group_dir = os.getenv('PASSWD_GROUP_DIR')
-    if passwd_group_dir is not None:
+    passwd_group_dir = os.getenv('PASSWD_GROUP_DIR', 'none')
+    if passwd_group_dir != 'none':
         inject_passwd_group(os.path.join(SRCDIR, passwd_group_dir))
     with tempfile.NamedTemporaryFile(mode='w') as argsfile:
         for pkg in packages:
@@ -130,6 +130,9 @@ def build_rootfs(target_rootfs, manifest_path, packages, locked_nevras, overlays
         if locked_nevras and lock_arg_supported():
             for locked_nevra in locked_nevras:
                 argsfile.write(f"--lock={locked_nevra}\n")
+        if passwd_group_dir == 'none':
+            argsfile.write("--sysusers\n")
+            argsfile.write("--nobody-99\n")
         argsfile.flush()
         cache_arg = []
         if os.path.isdir('/cache') and rpm_ostree_has_cachedir_fix():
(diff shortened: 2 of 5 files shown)


jlebon:
Hmm, CI is failing on

```
11:18:38  Traceback (most recent call last):
11:18:38    File "/usr/lib/coreos-assembler/cmd-prune", line 193, in <module>
11:18:38      oci_manifest = os.path.join(build_dir, meta['images']['oci-manifest']['path'])
11:18:38                                             ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
```

which I think is because https://builds.coreos.fedoraproject.org/prod/streams/testing-devel/builds/42.20250721.20.0/x86_64/meta.json doesn't have an `oci-manifest` artifact. Except... I'm not even sure how that's possible since it's create by [`cosa build` itself](https://github.com/coreos/coreos-assembler/blob/4a735bd7154e24e24f4be5c41f935d130bef58e4/src/cmd-build#L564-L581) together with the `ostree` entry.

jlebon:
OK right, it's because we `cosa buildupload --artifact=ostree` first, which will scrub un-uploaded artifacts from `meta.json`: https://github.com/coreos/coreos-assembler/blob/4a735bd7154e24e24f4be5c41f935d130bef58e4/src/cmd-buildupload#L177-L181.

I think it would make sense to also upload the OCI manifest, but I guess the prune code should be tolerant to this anyway: https://github.com/coreos/coreos-assembler/pull/4236

travier (dismissed):
LGTM but not tested. We would need to double check that nothing changes.

jlebon:
Rebased this now!

I've removed the change to `build-args.conf` so that we can keep the status quo for f43-based builds if we prefer. But once we merge this we can remove `PASSWD_GROUP_DIR` from `build-args.conf` for rawhide and then let it happen as part of f44 and down.

If one does a `cosa diff --ostree`, the main thing that jumps out is that before you had entries like this in `/usr/etc/gshadow`:

```
input:::
```

And now it looks slightly different:

```
input:!*::
```

Those are mostly equivalent though if you read `gshadow(5)`.

> This field may be empty, in which case only the group members can gain the group permissions.

> If the password field contains some string that is not a valid result of crypt(3), for instance ! or *, users will not be able to use a unix password to access the group (but group members do not need the password).

Either way, users can't directly access the group using a group password.

jlebon:
Hmm, though I do see also this in `/usr/etc/gshadow`:

```diff
-root:*::0:99999:7:::
+root:!unprovisioned:::::::
```

which is weird. One has passwords disabled entirely, whereas the other has it just locked. Will look at squashing that.

Edit: comes from https://github.com/systemd/systemd/blob/1f0acba5121cab17fe447c1e0aebfabf177ce1e8/src/basic/user-util.h#L146-L149 I think.

PR #3612, opened 2025-07-21
build-rootfs: inject our repo definitions into build stage

jlebon:
So far, we've been using the default repos in the bootc base image. But e.g. those repo files use mirrors and so are subject to lag/inconsistencies across runs.

In the end, we want our repo files to be canonical.

Do this by nuking any default repo file and injecting our own. Also use the new `bootc-base-imagectl build-rootfs --repo` switch to control enablement.

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -101,7 +101,9 @@ def build_rootfs(target_rootfs, manifest_path, packages, overlays, repos, nodocs
             for repo in repos:
                 argsfile.write(f"--repo={repo}\n")
         argsfile.flush()
-        cache_arg = ['--cachedir=/cache'] if os.path.isdir('/cache') else []
+        cache_arg = []
+        if os.path.isdir('/cache') and rpm_ostree_has_cachedir_fix():
+            cache_arg = ['--cachedir=/cache']
         subprocess.check_call(["/usr/libexec/bootc-base-imagectl",
                                "--args-file", argsfile.name, "build-rootfs",
                                "--manifest", 'minimal-plus',
@@ -133,6 +135,18 @@ def workaround_rhel_97826(argsfile):
     shutil.copytree(f'/{basedir}', os.path.join(tmpd.name, basedir), symlinks=True)
     argsfile.write(f'--add-dir={tmpd.name}\n')
     return tmpd
+
+
+def rpm_ostree_has_cachedir_fix():
+    # we can only use --cachedir if we have rpm-ostree >= 2025.9 which has
+    # https://github.com/coreos/rpm-ostree/pull/5391
+    out = subprocess.check_output(['rpm-ostree', '--version'], encoding='utf-8')
+    data = yaml.safe_load(out)
+    version_str = data['rpm-ostree']['Version']
+    # ideally, we could use `packaging.version`, but that's not in centos-bootc
+    # but conveniently, Python list comparisons do the right thing here
+    version = [int(c) for c in version_str.split('.')]
+    return version >= [2025, 9]
 
 
 # We want to keep our passwd/group as canonical for now. We should be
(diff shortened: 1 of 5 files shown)


jlebon:
This requires https://gitlab.com/fedora/bootc/base-images/-/merge_requests/248.

build-rootfs
```diff
@@ -66,7 +71,19 @@ def get_treefile(manifest_path):
     return json.loads(data)
 
 
-def build_rootfs(target_rootfs, manifest_path, packages, overlays, nodocs):
+def inject_yumrepos():
+    for repo in glob.glob('/etc/yum.repos.d/*.repo'):
+        if os.path.basename(repo) == 'secret.repo':
+            continue
```
dustymabe: could probably use a comment as to why this is special

build-rootfs
```diff
@@ -66,7 +71,19 @@ def get_treefile(manifest_path):
...
+        os.unlink(repo)
```
dustymabe, line 80: could probably use a comment here.. basically we're draining all existing repos first, but may be worth calling that out

build-rootfs
```diff
@@ -66,7 +71,19 @@ def get_treefile(manifest_path):
...
+    for repo in glob.glob(f'{CONTEXTDIR}/*.repo'):
+        # the pool repo is special and handled separately (in inject_pool_repo_if_exists)
+        if os.path.basename(repo) == 'fedora-coreos-fedora-coreos-pool.repo':
```
dustymabe: Is it not just `fedora-coreos-pool.repo`?

```suggestion
        if os.path.basename(repo) == 'fedora-coreos-pool.repo':
```

dustymabe:
> This requires https://gitlab.com/fedora/bootc/base-images/-/merge_requests/248.

rebased this PR now that the bootc submodule has been bumped to include that ^^ in https://github.com/coreos/fedora-coreos-config/pull/3623

dustymabe:
CI is failing because the repo files we have are kind of specific to being run inside COSA where we have the `distribution-gpg-keys` RPM installed:

```
error: Installing packages: Updating rpm-md repo 'fedora': Failed to download gpg key for repo 'fedora': Curl error (37): Could not read a file:// file for file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-42-primary [Couldn't open file /usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-42-primary]
```

jlebon:
The larger problem with this of course is that we don't have that bootc-base-imagectl patch in RHEL 9.6. Of course, it matters less there because we solely rely on the secret-injected repo file.

So I think I need to make this more conditional so that it only takes effect in FCOS.

dustymabe:
> build-rootfs: only use --cachedir if rpm-ostree >= v2025.9
>
> I.e. we can't use it if we don't have
> https://github.com/coreos/rpm-ostree/pull/5391.

i.e. this breaks downstream?

jlebon:
> > build-rootfs: only use --cachedir if rpm-ostree >= v2025.9
> > I.e. we can't use it if we don't have
> > [coreos/rpm-ostree#5391](https://github.com/coreos/rpm-ostree/pull/5391).
> 
> i.e. this breaks downstream?

Sorry, I should've made this more explicit. But yeah, it's not in RHEL 9.6 (and likely will never be).

build-rootfs
```diff
@@ -164,17 +194,13 @@ def get_locked_nevras():
     return [f'{k}-{v}' for (k, v) in locks.items()]
 
 
-def inject_pool_repo_if_exists(locked_nevras):
-    srcrepo = os.path.join(CONTEXTDIR, "fedora-coreos-pool.repo")
-    if not os.path.exists(srcrepo):
-        return
-
+def modify_pool_repo_if_exists(locked_nevras):
```
dustymabe, line 211: The change in this commit is kind of dropping the `if_exists` part of this function.

Are we now assuming it exists? Is that safe?

dustymabe (approved):
LGTM

build-rootfs
jlebon: It doesn't hurt because FCOS is the only one that has lockfiles and we only do this if those exist. But yeah, the function name should've been updated.

Will fix that in the next patch set. For now at least, will get this in to not waste CI!

PR #3613, opened 2025-07-24
konflux: onboard all streams

joelcapitao:
See individual commits.

### DIFF
--- a/.tekton/base/base/fedora-coreos.yaml
+++ b/.tekton/base/base/fedora-coreos.yaml
@@ -1,0 +1,50 @@
+apiVersion: tekton.dev/v1
+kind: PipelineRun
+metadata:
+  name: fedora-coreos
+  namespace: coreos-tenant
+  labels:
+    appstudio.openshift.io/application: fedora-coreos
+    appstudio.openshift.io/component: fedora-coreos
+    pipelines.appstudio.openshift.io/type: build
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/coreos/fedora-coreos-config?rev={{revision}}
+    build.appstudio.redhat.com/commit_sha: '{{revision}}'
+    build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    pipelinesascode.tekton.dev/max-keep-runs: "3"
+  creationTimestamp: null
+spec:
+  params:
+  - name: git-url
+    value: '{{source_url}}'
+  - name: revision
+    value: '{{revision}}'
+  - name: output-image
+    value: quay.io/konflux-fedora/coreos-tenant/fedora-coreos-{{target_branch}}
+  - name: dockerfile
+    value: Containerfile
+  - name: path-context
+    value: .
+  - name: privileged-nested
+    value: true
+  - name: build-args
+    value:
+    - VERSION=42
+    - BUILDER_IMG=quay.io/fedora/fedora-bootc:42
+    - MANIFEST=manifest.yaml
+    - PASSWD_GROUP_DIR=manifests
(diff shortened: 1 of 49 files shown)


joelcapitao:
Needs https://gitlab.com/fedora/bootc/tekton-catalog/-/merge_requests/2 to be merged and bundle generated.

jbtrystram (dismissed):
LGTM

jbtrystram (approved):
LGTM

ci/generate-tekton-pipelinerun
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
+REPO_ROOT="$(realpath "${SCRIPT_DIR}/..")"
+
+if ! command -v "kustomize" &> /dev/null; then
+    echo "Please install kustomize to run this script:"
```
jlebon, line 7: Looks like this is packaged in Fedora. Let's just add it to cosa?

.tekton/base/base/fedora-coreos.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - name: path-context
+    value: .
+  - name: privileged-nested
+    value: true
+  - name: build-args
+    value:
+    - VERSION=42
+    - BUILDER_IMG=quay.io/fedora/fedora-bootc:42
+    - MANIFEST=manifest.yaml
+    - PASSWD_GROUP_DIR=manifests
```
jlebon, line 35: We shouldn't need to relist these build args here or anywhere in these manifests. The source of truth is `build-args.conf`. Does the task not support build arg files? Can we enhance it to support it?

jlebon (commented):
Overall looks sane in a quick scan. Thanks for doing this!

I wonder if there's a way to try to get to rid of the kustomize step. Isn't there any "built-in templating" possible based on the ref being built?

.tekton/base/base/fedora-coreos.yaml
joelcapitao: https://github.com/coreos/fedora-coreos-config/pull/3645 is addressing this

ci/generate-tekton-pipelinerun
joelcapitao: Good point, I'm gonna add it.

joelcapitao:
> Overall looks sane in a quick scan. Thanks for doing this!
> 
> I wonder if there's a way to try to get to rid of the kustomize step. Isn't there any "built-in templating" possible based on the ref being built?

I may have misunderstood the question but kustomize is itself kind of the built-in templating k8s-native tool for customizing the k8s configuration files i.e kubectl/oc supports it.
Though, we could use a pure template engine to achieve the same e.g Helm
Running kustomize has the advantage of sanitizing the configuration files.

ci/generate-tekton-pipelinerun
joelcapitao: https://github.com/coreos/coreos-assembler/pull/4247

jlebon:
> > Overall looks sane in a quick scan. Thanks for doing this!
> > I wonder if there's a way to try to get to rid of the kustomize step. Isn't there any "built-in templating" possible based on the ref being built?
> 
> I may have misunderstood the question but kustomize is itself kind of the built-in templating k8s-native tool for customizing the k8s configuration files i.e kubectl/oc supports it.

Rewording my question: with https://github.com/coreos/fedora-coreos-config/pull/3645 now, what in the pipeline configuration needs to differ across streams? It's just the ref being built, no? Is there anything we can leverage that doesn't require templating but just default behaviour based on the ref being built?

PR #3614, opened 2025-07-24
[rawhide] overrides: pin shadow-utils-4.17.4-4.fc43

coreosbot-releng:


marmijo (approved):
LGTM

PR #3615, opened 2025-07-24
overlay.d/20platform-chrony: add support for Oracle Cloud

alaviss:
Similar to AWS/GCP, OCI runs an NTP implementation at metadata IP.

Ref: https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/configuringntpservice.htm

### DIFF
--- a/tests/kola/ntp/chrony/coreos-platform-chrony-config
+++ b/tests/kola/ntp/chrony/coreos-platform-chrony-config
@@ -1,7 +1,7 @@
 #!/bin/bash
 ## kola:
 ##   exclusive: false
-##   platforms: "aws azure gcp qemu"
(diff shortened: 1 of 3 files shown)


dustymabe:
This LGTM. Could you also update tests/kola/ntp/chrony/coreos-platform-chrony-config just in case we are ever able to run kola tests in OracleCloud we can test that this is getting set in our automated tests.

tests/kola/ntp/chrony/coreos-platform-chrony-config
dustymabe, line 1: Could you add `oraclecloud` to line 4?

dustymabe (approved):
LGTM

PR #3616, opened 2025-07-24
denylist: rawhide: remove `skip-console-warnings` for openstack

marmijo:
It's difficult to reproduce this locally, and the current openstack builds don't show any console warnings. Let's remove this entry because the issue is most likely resolved.


dustymabe (approved):
LGTM

PR #3621, opened 2025-07-25
OCPBUGS-56597: Work around missing multipath WWIDs population in dracut

ravanelli:
- This commit adds a temporary workaround for a bug where dracut fails to populate
/etc/multipath/wwids from the kernel command line, which breaks:

ExecStartPre=-/sbin/multipath -A

This issue affects both FCOS and RHCOS and is
tracked in RHEL-96106.

- It adds a Systemd override to apply the fix for multipathd.service

### DIFF
--- a/overlay.d/05custom-multipath-fix/dracut/90custom-multipath-fix/module-setup.sh
+++ b/overlay.d/05custom-multipath-fix/dracut/90custom-multipath-fix/module-setup.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+check() {
+    require_binaries multipath || return 1
+}
+
+depends() {
+    echo dm
+    return 0
+}
+
+install() {
+    inst_multiple multipath
+    inst_hook pre-mount 90 "$moddir/run-multipath.sh"
+}
--- a/overlay.d/05custom-multipath-fix/dracut/90custom-multipath-fix/run-multipath.sh
+++ b/overlay.d/05custom-multipath-fix/dracut/90custom-multipath-fix/run-multipath.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+/sbin/multipath -A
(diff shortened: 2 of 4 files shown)


ravanelli:
It depends on https://github.com/coreos/coreos-installer/pull/1676

jlebon:
AIUI, we don't need this in Fedora since dracut v107 which is in f43 and f42. From the RHEL ticket:

> Upstream dracut-ng used its own copy of multipathd.service until 5e87b68c ("refactor(multipath): remove custom multipathd.service"), which is included in dracut-107.

which is https://github.com/dracut-ng/dracut-ng/commit/5e87b68c.

So... I think this needs to happen in coreos/rhel-coreos-config instead?

I think also it would be cleaner to structure this as a postprocess script which just checks if the patch is present and if not, then edit dracut's multipathd.service.

ravanelli:
> AIUI, we don't need this in Fedora since dracut v107 which is in f43 and f42. From the RHEL ticket:
> 
> > Upstream dracut-ng used its own copy of multipathd.service until 5e87b68c ("refactor(multipath): remove custom multipathd.service"), which is included in dracut-107.
> 
> which is [dracut-ng/dracut-ng@5e87b68c](https://github.com/dracut-ng/dracut-ng/commit/5e87b68c).
> 
> So... I think this needs to happen in coreos/rhel-coreos-config instead?
> 
> I think also it would be cleaner to structure this as a postprocess script which just checks if the patch is present and if not, then edit dracut's multipathd.service.

Umm interesting because I tested it using FCOS 42 and it does fail with the same issue in rhcos without it.

jlebon:
Hmm, do you know if you tested it before or after https://github.com/coreos/fedora-coreos-config/pull/3593 ? Possibly it wasn't in FCOS yet?

ravanelli:
> Hmm, do you know if you tested it before or after #3593 ? Possibly it wasn't in FCOS yet?

Aa that's it, I didn't see we have it before. I'm closing it and will add it only for rhcos so. Thanks @jlebon

PR #3628, opened 2025-07-29
Containerfile: Make use of cosa directories

PeaceRebel:
To go with making cosa build using podman/buildah, use the tmp and cache directories from cosa build working directory.

Wanted by: https://github.com/coreos/coreos-assembler/pull/4242

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -22,24 +22,25 @@ ARG PASSWD_GROUP_DIR
 
 # Note: once we can rely on https://github.com/coreos/rpm-ostree/pull/5391,
 # add this bit to the RUN command to make the developer path less painful.
-# --mount=type=cache,rw,id=coreos-build-cache,target=/cache
+#
 RUN --mount=type=secret,id=yumrepos,target=/etc/yum.repos.d/secret.repo \
+    --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
     --mount=type=secret,id=contentsets \
     --mount=type=bind,target=/run/src \
         /run/src/build-rootfs "${MANIFEST}" "${VERSION}" /target-rootfs
-RUN --mount=type=bind,target=/run/src,rw \
+RUN --mount=type=bind,target=/run/tmp \
       rpm-ostree compose build-chunked-oci \
         --bootc --format-version=1 --rootfs /target-rootfs \
-        --output oci-archive:/run/src/out.ociarchive
+        --output oci-archive:/run/tmp/out.ociarchive
 
-FROM oci-archive:./out.ociarchive
+FROM oci-archive:./tmp/out.ociarchive
 ARG VERSION
 ARG NAME=overridden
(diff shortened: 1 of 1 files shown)


Containerfile
```diff
@@ -22,24 +22,25 @@ ARG PASSWD_GROUP_DIR
...
 # add this bit to the RUN command to make the developer path less painful.
-# --mount=type=cache,rw,id=coreos-build-cache,target=/cache
+#
 RUN --mount=type=secret,id=yumrepos,target=/etc/yum.repos.d/secret.repo \
+    --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
     --mount=type=secret,id=contentsets \
     --mount=type=bind,target=/run/src \
         /run/src/build-rootfs "${MANIFEST}" "${VERSION}" /target-rootfs
-RUN --mount=type=bind,target=/run/src,rw \
+RUN --mount=type=bind,target=/run/tmp \
```
travier, line 31:
```suggestion
RUN --mount=type=bind,target=/run/tmp,rw \
```

Containerfile
```diff
@@ -22,24 +22,25 @@ ARG PASSWD_GROUP_DIR
 
 # Note: once we can rely on https://github.com/coreos/rpm-ostree/pull/5391,
 # add this bit to the RUN command to make the developer path less painful.
-# --mount=type=cache,rw,id=coreos-build-cache,target=/cache
```
jbtrystram, line 25: have this landed in all streams ? If so let's delete the whole comment.

jbtrystram (changes requested):

Containerfile
PeaceRebel: The release that contains these changes is in stable repo now.

PeaceRebel:
These changes makes use of directories in cosa workdir. I'll try to get it working without these.

jbtrystram:
you will need to rebase as https://github.com/coreos/fedora-coreos-config/pull/3636 went in

dustymabe:
Putting this on hold as it might not be needed: https://github.com/coreos/coreos-assembler/pull/4242#issuecomment-3140417738

jlebon:
Hmm, this patch doesn't/shouldn't really have any effect wrt the outside world. It's still only mounting the build context directory, just to a differently named directory inside the build namespace.

jlebon:
But for more context (pun intended), the build context mount is just for the rechunk hack until https://github.com/containers/buildah/issues/5952 is fixed. If you look at [the PR](https://github.com/containers/buildah/pull/5975) that will fix that, the implementation is that the build context overlay persists across stages. The `--mount` line currently in that Containerfile is not currently load-bearing (because it's shadowed by the `-v $PWD:/srv` bind-mount), but _will be_ once the buildah issue is fixed (and we can remove the `-v` bit from the instructions).

jlebon:
I don't think we'll need this one.

PR #3633, opened 2025-07-29
tests/coreos-update-ca-trust: change location to check existance

prestist:
With the latest changes on rawhide fedora no longer has the file located in /etc/pki/tls/certs/ca-bundle.crt.  Instead check the extracted location /etc/pki/ca-trust/extracted/pem to verify existence.

https://fedoraproject.org/wiki/Changes/dropingOfCertPemFile
42 and rawhide pass testing.


jbtrystram (approved):
LGTM

marmijo (approved):
LGTM

dustymabe:
Since the test runs on RHCOS as well can you confirm that the updates to the test will continue to pass there?

Otherwise the changes LGTM

prestist:
> Since the test runs on RHCOS as well can you confirm that the updates to the test will continue to pass there?
> 
> Otherwise the changes LGTM

Sure I can test RHCOS, ok it passes in c9s as well.

PR #3636, opened 2025-07-29
Containerfile: add caching for local podman build

dustymabe:
Now that rpm-ostree-2025.9-1.fc42 is out we can leverage the work done upstream [1] to make cacheing in the container build workflow more practical. Let's enable it by default!

This leverages the code in build-rootfs that auto adds --cachedir to the bootc-base-imagectl call.

[1] https://github.com/coreos/rpm-ostree/pull/5391


jbtrystram (approved):

jlebon:
Follow up to this in https://github.com/coreos/fedora-coreos-config/pull/3612.

PR #3644, opened 2025-07-31
Konflux: generate rpm-lockfile-prototype input file

jbtrystram:
The resulting config file will be fed to `rpm-lockfile-protype` which in turns will generate the hermeto config.
This will allow us to enable hermetic builds in Konflux.

Run with :
```
 ./ci/generate_rpm_lockfile_config manifest.yaml
````

This creates `rpm.in.yaml`.

The intent is to trigger this in the `bump-lockfile` job of jenkins.

Requires https://github.com/konflux-ci/rpm-lockfile-prototype/pull/107 And https://github.com/konflux-ci/rpm-lockfile-prototype/pull/104

### DIFF
--- a/ci/generate_rpm_lockfile_config
+++ b/ci/generate_rpm_lockfile_config
@@ -1,0 +1,130 @@
+#!/usr/bin/python
+
+import argparse
+import os
+import re
+import sys
+import yaml
+
+# Correctly import get_treefile from the specified path
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+from manifest_utils import get_treefile, get_locked_nevras
+sys.path.pop(0)
+
+
+WORKDIR = os.getcwd()
+ARCH = os.uname().machine
+
+
+def filter_repofile(repos, locked_nevras, repo_path):
+    if not os.path.exists(repo_path):
+        print(f"Error: {repo_path} not found. Cannot inject includepkg filter.")
+        return
+
+    include_str = ','.join(locked_nevras)
+
+    with open(repo_path, 'r') as f:
+        repofile= f.read()
+
+    # We use a regex that looks for [reo_name] on a line by itself,
+    # possibly with whitespace.
+    sections = re.split(r'^\s*(\[.+\])\s*', repofile, flags=re.MULTILINE)
+
+    new_content = sections[0]  # part before any section
+    keep = False
+
+    # sections will be [before, section1_name, section1_content, section2_name, section2_content, ...]
+    for i in range(1, len(sections), 2):
+        name = sections[i]
+        # ignore repos that don't match the repos we want to use
+        if name.strip("[]") in repos:
+            repodef = sections[i+1]
+            if 'includepkgs=' not in repodef:
+                # We only keep the repo definition that we edited
+                # to avoid accidentaly taking in other packages
+                # from a repofile already having an includepkgs
+                # directive.
+                keep = True
+                new_content += name + '\n'
+                repodef += f"includepkgs={include_str}\n"
+                new_content += repodef
+
+    if keep:
+        with open(repo_path, 'w') as f:
+            f.write(new_content)
+
+    return keep
+
+
+def build_rpm_lockfile_config(packages, repo_files):
+    """
+    Augments package names in rpm_in_data with version numbers from locks.
+    Populates contentOrigin and repofiles.
+    """
+    # Initialize the structure for rpm_lockfile_input, similar to write_rpms_input_file
+    # This ensures consistency whether it comes from a manifest or directly
+    rpm_lockfile_config = {
+        'contentOrigin': {
+            'repofiles': repo_files
+        },
+        'installWeakDeps': False,
+        'context': {
+            'bare': True,
+        },
+        'packages': packages
+    }
+
+    return rpm_lockfile_config
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate rpm-lockfile-prototype input file from rpm-ostree manifest and augment with version overrides."
+    )
+
+    parser.add_argument(
+        'manifest',
+        help='Path to the rpm-ostree manifest (e.g., fedora-coreos-config/manifest.yaml)'
+    )
+    parser.add_argument(
+        '--output',
+        default='./rpms.in.yaml',
+        help="Path for the final rpm-lockfile-protoype config file. (default: './rpms.in.yaml')"
+    )
+
+    args = parser.parse_args()
+    contextdir = os.path.dirname(os.path.join(WORKDIR, args.manifest))
+
+    manifest_path = os.path.join(contextdir, os.path.basename(args.manifest))
+    # 1. Get initial manifest data from get_treefile
+    manifest_data = get_treefile(manifest_path, deriving=False)
+    repos = manifest_data.get('repos', [])
+    repos += manifest_data.get('lockfile-repos', [])
+
+    packages = manifest_data.get('packages', [])
+
+    # 2 Get locked NEVRAs from manifest-lock.overrides.yaml
+    locks = get_locked_nevras(contextdir, ARCH)
+
+    # Gather list of *.repo files
+    repofiles = [file for file in os.listdir(contextdir) if file.endswith('.repo')]
+    relevant_repofiles = []
+
+    # 3 inject filters to relevant repofiles
+    if locks:
+        locked_nevras = [f'{k}-{v}' for (k, v) in locks.items()]
+        for repofile in repofiles:
+            if filter_repofile(repos, locked_nevras, os.path.join(contextdir, repofile)):
+                print(f"\u2705 filter injected in: {os.path.join(contextdir, repofile)}")
+                relevant_repofiles.append(repofile)
+
+    # 4 generate rpm-lockfile-prototype input file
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: ci/generate_rpm_lockfile_config [hunk 1]
```diff
@@ -0,0 +1,127 @@
...
+        description="Generate rpm-lockfile-prototype input file from rpm-ostree manifest and augment with version overrides."
+    )
+
+    parser.add_argument(
+        'manifest',
+        help='Path to the rpm-ostree manifest (e.g., fedora-coreos-config/manifest.yaml)'
+    )
+    parser.add_argument(
+        '--output',
+        default='./rpm.in.yaml',
```
joelcapitao: can we use `rpms.in.yaml` as default value which is the one used by `rpm-lockfile-protoype` [1] ?

[1] https://github.com/konflux-ci/rpm-lockfile-prototype?tab=readme-ov-file#whats-the-input_file

ci/generate_rpm_lockfile_config
```diff
@@ -0,0 +1,127 @@
...
+        help="Path for the final rpm-lockfile-protoype config file. (default: './rpm.in.yaml')"
```
joelcapitao: ditto

ci/generate_rpm_lockfile_config
```diff
@@ -0,0 +1,127 @@
...
+        # ignore repos that don't match the repos we want to use
+        if name.strip("[]") in repos:
+            repodef = sections[i+1]
+            new_content += name + '\n'
+            if 'includepkgs=' not in repodef:
+                repodef += f"includepkgs={include_str}\n"
+
+            new_content += repodef
+            # remember to keep that repofile
+            keep = True
```
joelcapitao: To be run only once, `keep = True` can be put in the `if` block that initializes `new_content` with `includepkgs=`

ci/generate_rpm_lockfile_config
```diff
@@ -0,0 +1,127 @@
...
+
+    if keep:
+        with open(repo_path, 'w') as f:
+            f.write(new_content)
+
+    return keep
+
+
+def build_rpm_lockfile_config(packages, repo_files):
```
joelcapitao, line 59: To simplify, I'd rather get rid of this function and write the `rpms.in.yaml` ourself with same config. The python script will then only load the file, update the `packages` list, and write it. That's how we achieve it in coreos-assembler [1].
But we can let it as-is, see how it goes and adjust if needed 

[1] https://github.com/coreos/coreos-assembler/blob/babe36483937879ee8b57fe5a5001d4b5561f756/ci/hermetic/update_package_list#L84-L129

ci/generate_rpm_lockfile_config
```diff
@@ -0,0 +1,127 @@
...
+    """
+    # Initialize the structure for rpm_lockfile_input, similar to write_rpms_input_file
+    # This ensures consistency whether it comes from a manifest or directly
+    rpm_lockfile_config = {
+        'contentOrigin': {
+            'repofiles': repo_files
+        },
+        'installWeakDeps': False,
+        'context': {
+            'bare': True,
```
joelcapitao, line 72: I'm not sure of the `bare` usage here, I think we want to base libdnf resolution on the bootc base image e.g: https://github.com/coreos/coreos-assembler/blob/babe36483937879ee8b57fe5a5001d4b5561f756/rpms.in.yaml#L22

To clarify, using `bare: True` works but we might resolve (thus fetch) more packages than necessary. We'll see when testing it with Konflux whether we can optimize or not.

joelcapitao (changes requested):
See my inline comments
Also, I'd have split into two commits 1. the migration of `get_locked_nevras` and `get_treefile` into `ci/manifest_utils.py` and 2. the specific `generate_rpm_lockfile_config` code.

ci/generate_rpm_lockfile_config
jbtrystram: wooops :)

ci/generate_rpm_lockfile_config
jbtrystram: that will end up running the same amount of times anyway. The check for if `includepkgs` already exist is just to avoid generating an invalid repofile (two includepkgs would be illegal).
Hmm writing that i realize that moving the `keep` to there would protect us from an edge case where an repofile already have an `includepkgs`, and we'd silently keep it without knowing the content. 
thanks !

ci/generate_rpm_lockfile_config
jbtrystram: My idea was to run `rpm-lockfile-prototype` at the same time as we generate the input file so we only commit the resulting lock. The input file get re-generated everytime, because the repos list must be dynamic for each stream so there isn't much common ground anyway

ci/generate_rpm_lockfile_config
jbtrystram: That would not work because we are not layering on top of the base bootc image. We pull the manifest from it, and re-download everything.
This is necessary because we might want to pin a package that is in the base image, and as we want to control the version of all the RPMs we ship we re-download everything.
The base bootc image is just here to provide a manifest and the `bootc` and `rpm-ostree` binaries essentially.

ci/generate_rpm_lockfile_config
joelcapitao: Right, I misunderstood the compute process

ci/generate_rpm_lockfile_config
joelcapitao: Okk, I was not aware of this nuance, thank you for the clarification here. I need to hack it more on bootc and the derivation that is taking place in our bakery :) 
So yeah it makes lot of sense to pull everything from scratch then. :+1:

ci/generate_rpm_lockfile_config
joelcapitao: Right, I was missing the fact that the list of repos is generated dynamically (i.e only the relevant repo files are injected).

joelcapitao (dismissed):
Ideally, we should 
1. copy/paste the source repo files to a temporary ones, compute the latter ones with `filter_repofile`, and define them in `rpms.in.yaml`.
2. run `rpm-lockfile-prototype rpms.in.yaml -o rpms.lock.yaml`
3. remove `rpms.in.yaml` (as it's a dynamic generated file)

all in the script.
That way we'd only have one change i.e: `rpms.lock.yaml`. The current behavior is changing the repo files.
But this can be achieve in a followup path

jbtrystram:
Splitted in two commits.

jbtrystram:
> Ideally, we should
> 
>     1. copy/paste the source repo files to a temporary ones, compute the latter ones with `filter_repofile`, and define them in `rpms.in.yaml`.
> 
>     2. run `rpm-lockfile-prototype rpms.in.yaml -o rpms.lock.yaml`
> 
>     3. remove `rpms.in.yaml` (as it's a dynamic generated file)
> 
> 
> all in the script. That way we'd only have one change i.e: `rpms.lock.yaml`. The current behavior is changing the repo files. But this can be achieve in a followup path

all of this will happen in the pipeline where we throw the clone away at the end. I don't expect devs to locally build hermeto lockfiles, so i don't think it's worth the complexity. The Jenkins job will just commit the expected file and ignore the repofiles changes.

jbtrystram:
Moving this to draft as i haven't had time to test it in the pipeline yet, only locally, and i want to be extra-careful not breaking the `bump-lockfile` job.

joelcapitao:
> all of this will happen in the pipeline where we throw the clone away at the end. I don't expect devs to locally build hermeto lockfiles, so i don't think it's worth the complexity. The Jenkins job will just commit the expected file and ignore the repofiles changes.

Indeed, there's no point in bothering with it then

jlebon:
Instead of being run by the bump-lockfile job, wouldn't it be cleaner to have it happen as part of `cosa fetch --update-lockfile`? Though maybe with a `--with-konflux` switch as well as start (which we'd add to the bump-lockfile job).

We'll probably want a CI check as well to sanity-check they're in sync.

I don't understand the repo file handling here. What does a resulting Konflux lockfile look like?

jbtrystram:
> Instead of being run by the bump-lockfile job, wouldn't it be cleaner to have it happen as part of `cosa fetch --update-lockfile`? Though maybe with a `--with-konflux` switch as well as start (which we'd add to the bump-lockfile job).

Good idea, i'll look into that.

> I don't understand the repo file handling here. What does a resulting Konflux lockfile look like?

The overall process is : 
We look at manifests to know what repo will be used.
Unfortunately the repo names don't necessarily match the filenames so we open all repo files to keep only the
repo definitions we want and inject the `nevra` values in a `includepkg` section. 
The resulting repo file is passed to rpm-lockfile protype along the list of packages through the [rpm-lockfile-prototype input file](https://github.com/konflux-ci/rpm-lockfile-prototype#whats-the-input_file).
Rpm-lockfile-protype does the resolution using our crafted repofile, so it will respect the locks. 
The result is an hermeto (ex-cachi2) lockfile that hermeto turns into a local rpm repo and inject it into the container build, shadowing existing repos.

[hermeto lockfile doc](https://hermetoproject.github.io/hermeto/rpm/#rpm-lockfile-format)

jbtrystram:
Ok i moved this to cosa : https://github.com/coreos/coreos-assembler/pull/4298

PR #3649, opened 2025-08-03
denylist: snooze docker tests for now on rawhide

dustymabe:
There is a behavior change in the 6.17 kernel that needs investigation.
- https://github.com/coreos/fedora-coreos-tracker/issues/1998


buckaroogeek:
Seems reasonable to skip the tests for now. Docker will still be broken. Still looking for answers on ip_tables module and the 6.17 kernel

PR #3653, opened 2025-08-05
Enable `bootloader-update.service` since F43

HuijingHei:
`bootloader-update.service` is shipped in `rust-bootupd-0.2.26-3.fc42` (https://src.fedoraproject.org/rpms/rust-bootupd/c/28a98663eabd21cb9e9ff334700d554d36c0b0aa?branch=rawhide) but it is disabled by default.
Track issue: https://github.com/coreos/fedora-coreos-tracker/issues/1468

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -52,6 +52,16 @@ conditional-include:
         - python
         - python3
         - python3-libs
+  # We will enable bootloader-update.service on F43+.
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1468#issuecomment-2996654547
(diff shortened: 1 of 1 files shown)


dustymabe:
If this is part of https://fedoraproject.org/wiki/Changes/AutomaticBootloaderUpdatesBootc then maybe we should enable it in bootc upstream with a conditional includes [here](https://gitlab.com/fedora/bootc/base-images/-/blob/f17929f6b0b1c369c61c5900076617a05a073cda/minimal/systemd-presets.yaml) and then we'll inherit that?

cc @travier

HuijingHei:
Thank you Dusty for the pointer, close this as will do in https://gitlab.com/fedora/bootc/base-images/-/merge_requests/257

PR #3654, opened 2025-08-06
manifests/fedora-coreos: advance OCI migration to all streams

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/1890. We are now ready for this in our `stable` stream.


marmijo (approved):
LGTM

PR #3658, opened 2025-08-08
Use systemd-run instead of machinectl shell

keszybz:


dustymabe:
Thanks for the PR here. I do think it's still worth chasing down the `machinectl` bug, but I'm OK with moving forward with this PR as is.


That being said this still doesn't pass the test. `toolbox --help` works, but `toolbox create --assumeyes` does not:

```
[root@qemu0 ~]# systemd-run -q --wait --uid=core -t toolbox --help
toolbox - Tool for interactive command line environments on Linux

Common commands are:
create    Create a new Toolbx container
enter     Enter an existing Toolbx container
list      List all existing Toolbx containers and images

Go to https://containertoolbx.org/ for further information.
```

```
[root@qemu0 ~]# systemd-run -q --wait --uid=core -t toolbox create --assumeyes
Error: failed to create runtime directory toolbox: mkdir toolbox: read-only file system
```

One of the nice things about `machinectl` is that it brings up your login session (not sure of the fully correct term here) and thus some env vars might be set that aren't otherwise. I'm not sure if `systemd-run` is missing that stuff.

keszybz:
Indeed. So I think we really want `machinectl shell`. I changed the patch to use a less intrusive approach. (I didn't run the tests locally, just the standalone commands… I'm relying on the CI for this.)

dustymabe:
> Indeed. So I think we really want `machinectl shell`. I changed the patch to use a less intrusive approach.

Thanks!

 (I didn't run the tests locally, just the standalone commands… I'm relying on the CI for this.)

> CI here won't help because it runs against `testing-devel` (F42 based for now), but I can test it locally.


I am just wondering.. Are we just working around a bug? Or are we expected to have to make changes like this to scripts in the future?

keszybz:
> Are we just working around a bug?

That is a good question. I made it also in the upstream systemd bug. The man page says that `systemd-run` should be used for scripting, but it does *not* say that `machinectl shell` is not suitable. In general systemd tools automatically disable coloring and other output intended for the terminal emulator when not connected to a terminal. But in this case, the output *is* connected to a terminal so that code is not effective. I looked at the code to figure out how this could be changed, but it's not all obvious. So I think it's reasonable to apply the workaround for now so that we can push the systemd update with passing CI, and figure this out upstream asynchronously.

dustymabe (approved):
This passes in a local test run against `rawhide`.

PR #3659, opened 2025-08-08
versionary: support `--dev`

jlebon:
Add a `--dev` switch which will be used to replace the functionality
lost by `automatic-version-prefix` after moving away from directly
running rpm-ostree composes to the container native path.


dustymabe (approved):
LGTM

PR #3663, opened 2025-08-12
Remove qcom dtbs on aarch64 again

dustymabe:
According to Peter Robinson the qcom dtbs are the least likely to actually be used because they are for chromebooks or phones. Let's drop these for now to help us with our space in /boot problem on ARM.

https://github.com/coreos/fedora-coreos-tracker/issues/2004


aaradhak (approved):
LGTM

PR #3666, opened 2025-08-14
tests: bump kernel-replace to 20G

jlebon:
We've been hitting ENOSPC on this test all over the place and have been
adding workarounds a bit everywhere. Hit it again today as part of the
f43 container-native work.

Let's just bump its disk size.


dustymabe (approved):
LGTM

PR #3667, opened 2025-08-18
overrides: pin selinux-policy-42.4-1.fc42

coreosbot-releng:
Requested by @dustymabe via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).

PR #3668, opened 2025-08-18
tests: clean up rollback before testing for unlabeled files

dustymabe:
In the case where the name of a label gets changed then the rollback deployment will show files as unlabeled_t because the currently loaded policy (i.e. the upgraded policy) doesn't know about the old label. Since we are more concerned about the upgraded system let's just focus on finding unlabeled files there and drop the rollback deployment.

See https://github.com/coreos/fedora-coreos-tracker/issues/2007#issuecomment-3197248482

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -160,32 +160,16 @@ move-to-cgroups-v2() {
     fi
 }
 
-# We need to drop the rollback deployment. During upgrade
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

keszybz (approved):
LGTM.

dustymabe:
CI here is complaining about a downgraded package but that was intentional as part of https://github.com/coreos/fedora-coreos-config/pull/3667

I'll merge this.

PR #3672, opened 2025-08-19
overrides: drop selinux-policy pin

dustymabe:
We decided to change the scope of our selinux labels check to not consider the rollback deployment in be78a69 so we can drop this pin.

This reverts commit 34c50b271e08a493d5aa74c4068b37d967d1d873.


marmijo (approved):
LGTM

PR #3675, opened 2025-08-21
konflux: add Tekton pipelines for fcos-buildroot builds

Roshan-R:
Add new Tekton pipelines to build the fcos-buildroot image on push and pull requests to the testing-devel.
These pipelines only run when files in ci/buildroot change. Other pipelines are set to ignore changes in this path.

### DIFF
--- a/.tekton/branched/on-pull-request/fedora-coreos-branched-on-pull-request.yaml
+++ b/.tekton/branched/on-pull-request/fedora-coreos-branched-on-pull-request.yaml
@@ -14,7 +14,7 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "branched"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "branched" && !("ci/buildroot/**".pathChanged())
   creationTimestamp: null
 spec:
   params:
--- a/.tekton/branched/on-pull-request/fedora-coreos-branched-on-pull-request.yaml
+++ b/.tekton/branched/on-pull-request/fedora-coreos-branched-on-pull-request.yaml
@@ -14,7 +14,7 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "branched" && !("ci/buildroot/**".pathChanged())
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "branched" && !(files.all.all(f, f.matches("ci/buildroot/")))
   creationTimestamp: null
 spec:
   params:
--- a/.tekton/fcos-buildroot/fcos-buildroot-pull-request.yaml
+++ b/.tekton/fcos-buildroot/fcos-buildroot-pull-request.yaml
@@ -1,19 +1,21 @@
 apiVersion: tekton.dev/v1
 kind: PipelineRun
 metadata:
-  name: fedora-coreos-testing-on-push
+  name: fcos-buildroot-on-pull-request
   namespace: coreos-tenant
   labels:
-    appstudio.openshift.io/application: fedora-coreos-testing
-    appstudio.openshift.io/component: fedora-coreos-testing
+    appstudio.openshift.io/application: fcos-buildroot
+    appstudio.openshift.io/component: fcos-buildroot
     pipelines.appstudio.openshift.io/type: build
   annotations:
     build.appstudio.openshift.io/repo: https://github.com/coreos/fedora-coreos-config?rev={{revision}}
-    build.appstudio.redhat.com/cancel-in-progress: "false"
+    pipelinesascode.tekton.dev/cancel-in-progress: "true"
     build.appstudio.redhat.com/commit_sha: '{{revision}}'
+    build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && target_branch == "testing" && !("ci/buildroot/**".pathChanged())
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
+      == "testing-devel" && (".tekton/fcos-buildroot-pull-request.yaml".pathChanged() || "ci/buildroot/**".pathChanged())
   creationTimestamp: null
 spec:
   params:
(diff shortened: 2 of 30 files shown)


.tekton/fcos-buildroot-push.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: fcos-buildroot-on-push
+  namespace: coreos-tenant
+  labels:
+    appstudio.openshift.io/application: fcos-buildroot
+    appstudio.openshift.io/component: fcos-buildroot
+    pipelines.appstudio.openshift.io/type: build
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/Roshan-R/fedora-coreos-config?rev={{revision}}
+    pipelinesascode.tekton.dev/cancel-in-progress: "false"
+    pipelinesascode.tekton.dev/on-path-changed: "ci/buildroot"
```
joelcapitao: this line will be ignored because `on-cel-expression` is used
See below:
> If you have a [CEL expression](https://pipelinesascode.com/docs/guide/matchingevents/#matching-pipelinerun-by-path-change) the on-path-change annotation will be ignored.

from https://pipelinesascode.com/docs/guide/matchingevents/

.tekton/base/base/fedora-coreos.yaml
```diff
@@ -12,6 +12,7 @@ metadata:
     build.appstudio.redhat.com/commit_sha: '{{revision}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
+    pipelinesascode.tekton.dev/on-path-changed-ignore: ci/buildroot
```
joelcapitao: this line will be ignored because `on-cel-expression` is defined afterward with kustomize
See below:
> If you have a [CEL expression](https://pipelinesascode.com/docs/guide/matchingevents/#matching-pipelinerun-by-path-change) the on-path-change annotation will be ignored.

from https://pipelinesascode.com/docs/guide/matchingevents/

You need to ignore it in the `on-cel-expression` e.g https://github.com/coreos/fedora-coreos-config/blob/03bcf76410c0f0af8143dd841e6dcdb7f2be31f3/.tekton/testing-devel/on-pull-request/kustomization.yaml#L27-L29

.tekton/fcos-buildroot-pull-request.yaml
```diff
@@ -0,0 +1,45 @@
...
+kind: PipelineRun
+metadata:
+  name: fcos-buildroot-on-pull-request
+  namespace: coreos-tenant
+  labels:
+    appstudio.openshift.io/application: fcos-buildroot
+    appstudio.openshift.io/component: fcos-buildroot
+    pipelines.appstudio.openshift.io/type: build
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/Roshan-R/fedora-coreos-config?rev={{revision}}
```
joelcapitao: s/Roshan-R/coreos/

joelcapitao (changes requested):
I left some comments, but the overall looks good to me
Also, could we move the two `fcos-buildroot-*.yaml` files into a subdirectory e.g `fcos-buildroot` ?

Roshan-R:
> I left some comments, but the overall looks good to me Also, could we move the two `fcos-buildroot-*.yaml` files into a subdirectory e.g `fcos-buildroot` ?

Should I create the on-push and on-pull-request folders like the other directories?

jbtrystram (changes requested):
can you properly rebase on top of tesing devel and remove the extra commits ?

joelcapitao:
> > I left some comments, but the overall looks good to me Also, could we move the two `fcos-buildroot-*.yaml` files into a subdirectory e.g `fcos-buildroot` ?
> 
> Should I create the on-push and on-pull-request folders like the other directories?

We could but I don't think it worth it for `fcos-buildroot`. We can drop the two files as is

Roshan-R:
@jbtrystram @jcapiitao PR should be okay to merge now. The build fails due to a provision error

joelcapitao:
/retest 
Ralph Bean from the Konflux team fixed the temporary issue, let's kick the pipeline

.tekton/fcos-buildroot/fcos-buildroot-pull-request.yaml
```diff
@@ -0,0 +1,45 @@
...
+  name: fcos-buildroot-on-pull-request
+  namespace: coreos-tenant
+  labels:
+    appstudio.openshift.io/application: fcos-buildroot
+    appstudio.openshift.io/component: fcos-buildroot
+    pipelines.appstudio.openshift.io/type: build
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/coreos/fedora-coreos-config?rev={{revision}}
+    pipelinesascode.tekton.dev/cancel-in-progress: "true"
+    pipelinesascode.tekton.dev/on-path-changed: "ci/buildroot"
```
joelcapitao: this line can be removed

.tekton/fcos-buildroot/fcos-buildroot-pull-request.yaml
```diff
@@ -0,0 +1,45 @@
...
+    build.appstudio.redhat.com/commit_sha: '{{revision}}'
+    build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
+    build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    pipelinesascode.tekton.dev/max-keep-runs: "3"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
+      == "testing-devel" && ".tekton/fcos-buildroot-pull-request.yaml".pathChanged() && "ci/buildroot".pathChanged()
```
joelcapitao: I think we should add `"ci/buildroot/***".pathChanged()` to match all the files but I'm not sure TBH, have you tested it already ?

Also, it should be `(".tekton/fcos-buildroot-pull-request.yaml".pathChanged() || "ci/buildroot/***".pathChanged())` otherwise the pipeline would trigger only when both the tekton file and the files in ci/buildroot are updated

joelcapitao (changes requested):

.tekton/fcos-buildroot/fcos-buildroot-pull-request.yaml
joelcapitao: The doc is not clear for the `/***` see https://pipelinesascode.com/docs/guide/matchingevents/

Roshan-R: Looks like it uses it uses [this](https://github.com/gobwas/glob) library to handle glob patterns, so I think `"ci/buildroot/**".pathChanged()` should be enough. Let me test it and confirm

Roshan-R: I was able to confirm that the pipeline get triggered when something inside ci/buildroot changes by using `"ci/buildroot/**".pathChanged()`: https://github.com/Roshan-R/fedora-coreos-config/pull/6

jbtrystram (dismissed):
LGTM

.tekton/fcos-buildroot/fcos-buildroot-pull-request.yaml
joelcapitao: Thank you

.tekton/branched/on-pull-request/fedora-coreos-branched-on-pull-request.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "branched"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "branched" && !("ci/buildroot/**".pathChanged())
```
joelcapitao: Currently, if there are changes on both `ci/buildroot/` and somewhere else, then the pipeline is not triggered at all c.f https://github.com/coreos/fedora-coreos-config/pull/3695
We want to always trigger pipeline except if changes are scoped to `ci/buildroot/` only

joelcapitao (changes requested):

.tekton/branched/on-pull-request/fedora-coreos-branched-on-pull-request.yaml
joelcapitao: The last commit i.e https://github.com/coreos/fedora-coreos-config/pull/3675/commits/f6fb7b0c300f255fc2e3f6459bfa8e5c6b3b1ca1 does the trick :+1:

joelcapitao (approved):
LGTM
I like the approach i.e using `files.all.all(f, f.matches(""))`

joelcapitao:
I don't know why `container native` GHA is failing whereas Konflux pipeline is not.. They both build the same image, hence running the same `rpm-ostree compose rootfs` command.
Something is not right, but it's not related to this PR

PR #3676, opened 2025-08-21
[branched] overrides: pin git-core-2.50.1-1.fc43

coreosbot-releng:


dustymabe:
closing in favor of #3681

PR #3677, opened 2025-08-21
[rawhide] overrides: fast-track git-core-2.51.0-2.fc44

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/2011

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,10 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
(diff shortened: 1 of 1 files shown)


dustymabe:
switched this to a fast-track with the fix

jlebon (approved):

jlebon:
Hmm, CI failures seem legit but unrelated. I think we need to adjust some of the tests for the new NFS packaging.

dustymabe:
> Hmm, CI failures seem legit but unrelated. I think we need to adjust some of the tests for the new NFS packaging.

I'll handle those over in https://github.com/coreos/fedora-coreos-config/pull/3682

Merging this for now.

PR #3680, opened 2025-08-21
manifests: adjust NFS package specification

dustymabe:
In F44 the NFS package was adjusted [1] such that there is now a nfs-client-utils package and the specific nfs-utils-coreos package (i.e. a package specific to coreos) is no longer needed.

[1] https://src.fedoraproject.org/rpms/nfs-utils/c/4b4eb333ac24c2e393ca1022c29752024b864d82?branch=rawhide

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -55,6 +55,16 @@ conditional-include:
         - python
         - python3
         - python3-libs
+  # In F43 and older pull in the nfs-utils-coreos package
+  - if: releasever <= 43
+    include:
+      packages:
+        - nfs-utils-coreos
+  # In F44 we can use the nfs-client-utils package, which obsoletes nfs-utils-coreos
+  - if: releasever >= 44
+    include:
+      packages:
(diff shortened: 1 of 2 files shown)


manifests/fedora-coreos.yaml
```diff
@@ -55,6 +55,16 @@ conditional-include:
         - python
         - python3
         - python3-libs
+  # In F43 and older pull in the nfs-utils-coreos package
+  - if: releasever <= 43
+    include:
+      packages:
+        - nfs-utils-coreos
+  # In F44 we can use the nfs-client-utils package, which obsoloted nfs-utils-coreos
```
jlebon: Minor: obsoletes

jlebon (dismissed):

manifests/fedora-coreos.yaml
dustymabe: I'll wait for CI to finish here - then force push to fix the comment and merge.

dustymabe: ok CI passed (ignoring konflux)

PR #3687, opened 2025-08-25
manifests/fedora-coreos: switch exclude-packages to a postprocess

### LINKED ISSUE #3678: Find equivalent for exclude-packages in container native build flow
We have a list of defined packages we don't want to get pulled in:

https://github.com/coreos/fedora-coreos-config/blob/faae99664b426628ede56f6427add875e29ed904/manifests/fedora-coreos.yaml#L200-L219

This recently triggered and lead us to take action: https://github.com/coreos/fedora-coreos-tracker/issues/2011

We need some equivalent behavior in the container native workflow, either halting the build-via-buildah, or in CI kola tests. 

To stress the importance of this if https://github.com/coreos/fedora-coreos-pipeline/pull/1194 had merged yesterday we'd have missed this and would have been shipping perl until someone noticed it.

dustymabe:
Since we don't currently have this functionality in the build-via-container path let's convert this to a postprocess.

Resolves: https://github.com/coreos/fedora-coreos-config/issues/3678

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -51,6 +51,8 @@ conditional-include:
   # https://github.com/coreos/fedora-coreos-tracker/issues/1942#issuecomment-2881075898
   - if: releasever < 43
     include:
+      # Note we can still use exclude-packages in F42 because we aren't
+      # building via container there.
       exclude-packages:
         - python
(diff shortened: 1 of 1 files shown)


marmijo (approved):
LGTM

PR #3689, opened 2025-08-27
build-rootfs: rework lockfile handling

jlebon:
We weren't quite matching the cosa semantics of lockfiles, and this
was causing us trouble. Notably, doing `dnf install -y` for all locked
packages doesn't work for autolocked packages because those lockfiles
may contain packages that don't exist at all on other arches.

It also meant that e.g. even if removing a package from the manifest,
you'd still have it installed if you didn't also remove it from the
lockfiles.

Really, we need the `packages` list to be canonical and the lockfiles to
only act as filters.

There is a new `--lock` switch now in bootc-base-imagectl that we can
use for that (which underneath just uses rpm-ostree lockfiles, like we
do now). So let's use that.

This also allows us to drop the hack around modifying the pool repo by
adding `includepkgs` manually. As well as the caveat around local RPM
overrides handling.

---

Requires: https://gitlab.com/fedora/bootc/base-images/-/merge_requests/279 and https://github.com/coreos/rpm-ostree/pull/5475

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -191,14 +191,14 @@ def prepare_local_rpm_overrides(rootfs):
     pkglist = subprocess.check_output(['dnf', 'repoquery', f'--repofrompath=overrides,file://{overrides_repo}',
                                        '--repo=overrides', '--latest-limit=1', f'--arch={ARCH},noarch',
                                        '--qf', 'pkg: %{name} %{evr} %{arch}\n'], encoding='utf-8')
-    lockfile = {"packages": {}}
+    lockfile = {}
     for line in pkglist.splitlines():
         if not line.startswith("pkg: "):
             continue
         _, name, evr, arch = line.strip().split()
-        lockfile["packages"][name] = {"evra": f"{evr}.{arch}"}
+        lockfile[name] = {"evra": f"{evr}.{arch}"}
 
-    if len(lockfile['packages']) == 0:
+    if len(lockfile) == 0:
         return None
 
     with open('/etc/yum.repos.d/overrides.repo', 'w') as f:
@@ -211,7 +211,7 @@ cost=500
 priority=1
 ''')
 
-    print("Injected", len(lockfile['packages']), 'package overrides')
+    print("Injected", len(lockfile), 'package overrides')
     return lockfile
 
 
@@ -248,7 +248,7 @@ def get_locked_nevras(local_overrides):
         # want to support dumb `koji download-build` flows. For everything else
         # that's unlocked, we rely on the overrides repo having priority=1.
         locks.update({pkgname: v['evra'] if 'evra' in v else v['evr']
-                      for (pkgname, v) in local_overrides['packages'].items()
+                      for (pkgname, v) in local_overrides.items()
                       if pkgname in locks})
     return [f'{k}-{v}' for (k, v) in locks.items()]
 
--- a/build-rootfs
+++ b/build-rootfs
@@ -40,13 +40,10 @@ def main():
         repos += ['overrides']
 
     locked_nevras = get_locked_nevras(local_overrides)
-    if locked_nevras:
-        modify_pool_repo(locked_nevras)
 
-    packages.extend(locked_nevras)
     overlays = gather_overlays(manifest)
     nodocs = (manifest.get('documentation') is False)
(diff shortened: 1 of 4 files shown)


build-rootfs
```diff
@@ -125,6 +125,14 @@ def repo_arg_supported():
     return '--repo REPO' in help
 
 
+def lock_arg_supported():
+    # Detect if we have https://gitlab.com/fedora/bootc/base-images/-/merge_requests/279.
+    # If not, then we can't use `--lock`. That should only happen in RHCOS,
+    # where we only use this for autolocking and not base lockfile management.
```
dustymabe, line 131: so in the case where we are just using lockfiles for autolocking there's no problem?

dustymabe (dismissed):
LGTM - haven't tested it

dustymabe:
rebased on top of latest `testing-devel`.

buildroot-prep
```diff
@@ -0,0 +1,17 @@
...
+
+case "$VERSION_ID" in
+    43) urls=(https://kojipkgs.fedoraproject.org//packages/rpm-ostree/2025.10/3.fc43/"$arch"/rpm-ostree-{,libs-}2025.10-3.fc43."$arch".rpm);;
+    44) urls=(https://kojipkgs.fedoraproject.org//packages/rpm-ostree/2025.10/3.fc44/"$arch"/rpm-ostree-{,libs-}2025.10-3.fc44."$arch".rpm);;
+    *) exit 0;;
+esac
+
+if rpm -q "rpm-ostree-2025.10-1.fc43.$arch"; then
+    sudo dnf install -y "${urls[@]}"
+fi
```
dustymabe, line 17: I think a better model here would be to just continue to use our continuous repos when we need overrides like this.

This LGTM for now, though. I'll open a followup PR.

dustymabe:
hmm. Something isn't working right here. I did local testing on the `rawhide` branch (i.e. verifying local overrides worked) before merging https://gitlab.com/fedora/bootc/base-images/-/merge_requests/279 but then I rebased this PR and am doing local tests and it's pulling in f43 and f44 packages.

yes I did change the code locally to pull in a F44 rpm-ostree (because the update isn't built for F42). 

```patch
[dustymabe@media fedora-coreos-config (PR3689 *%)]$ git diff
diff --git a/buildroot-prep b/buildroot-prep
index 678015fc..1553c6e9 100755
--- a/buildroot-prep
+++ b/buildroot-prep
@@ -9,9 +9,8 @@ arch=$(uname -m)
 case "$VERSION_ID" in
     43) urls=(https://kojipkgs.fedoraproject.org//packages/rpm-ostree/2025.10/3.fc43/"$arch"/rpm-ostree-{,libs-}2025.10-3.fc43."$arch".rpm);;
     44) urls=(https://kojipkgs.fedoraproject.org//packages/rpm-ostree/2025.10/3.fc44/"$arch"/rpm-ostree-{,libs-}2025.10-3.fc44."$arch".rpm);;
+    42) urls=(https://kojipkgs.fedoraproject.org//packages/rpm-ostree/2025.10/3.fc44/"$arch"/rpm-ostree-{,libs-}2025.10-3.fc44."$arch".rpm);;
     *) exit 0;;
 esac
 
-if rpm -q "rpm-ostree-2025.10-1.fc43.$arch"; then
-    sudo dnf install -y "${urls[@]}"
-fi
+sudo dnf install -y "${urls[@]}"
```

dustymabe:
> hmm. Something isn't working right here.

and now after I added instrumentation I can't reproduce.. i.e. it seems to be working right.

I'd propose we at least build the rpm-ostree update for F42 too so that our `container native` GHA doesn't permafail.

dustymabe (approved):

buildroot-prep
jlebon: Yeah, using the continuous repo SGTM. I initially was implementing this a different way to have it be a generic hotpatch mechanism usable by RHCOS/SCOS too but then changed it.

build-rootfs
jlebon: Just to answer this one: without `--lock`, we'll regress on autolocking in RHCOS/SCOS. There's ways to hack around it, the easiest being to just use `buildroot-prep` to pull in rpm-ostree from 9.7, though I think we could do a hack in code too. Meanwhile, ART at least would continue to catch consistency issues.

PR #3698, opened 2025-08-28
buildroot-prep: fixups to get it to work for rawhide

dustymabe:
In rawhide $releasever evaluates to `rawhide` but we need it to be `44`. Let's use the VERSION_ID here.

Also install `patch` if we need to patch. This passed in CI on 5040a32 because I guess the 42 bootc image had been built in the past day.

Fixes 5040a32


aaradhak (approved):
LGTM

PR #3699, opened 2025-08-29
Add org.opencontainers.image.title OCI label

jlebon:
Aside from the OS codename like `fedora-coreos`, we also have a human-friendly version that's used for e.g. cloud uploads. We want to retain that in the buildah path.

Let's use the standard org.opencontainers.image.title label for this purpose.


dustymabe (approved):
LGTM

dustymabe (approved):

travier:
I merged everything but branched & next-devel:
- https://github.com/coreos/fedora-coreos-config/pull/3701
- https://github.com/coreos/fedora-coreos-config/pull/3702

PR #3700, opened 2025-08-29
[rawhide] build-args: add description

dustymabe:
Add the description that is specific to the `rawhide` stream. See https://github.com/coreos/fedora-coreos-config/pull/3699 for why this is useful.


jlebon (approved):

travier (approved):

travier:
Merging as this should not be breaking CI

PR #3705, opened 2025-08-29
build-rootfs: fix lockfile-repos handling

jlebon:
I mistakenly removed the pool repo hack in 534fa61e ("build-rootfs: rework lockfile handling") thinking it was no longer needed, but it is.

While we're here, make lockfile-repos handling more correct by only enabling them if we have locked packages and if we know how to add excludes to the repo so only those are available through there.

A more generic solution to this that wouldn't require repo file fiddling would have been to set that includepkgs in `/etc/dnf/dnf.conf` directly since you can change settings for specific repos there, but rpm-ostree disables that config file in its submoduled libdnf.

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -31,7 +31,7 @@ def main():
 
     packages = list(manifest['packages'])
 
-    repos = manifest.get('repos', []) + manifest.get('lockfile-repos', [])
+    repos = manifest.get('repos', [])
     if repos:
         inject_yumrepos()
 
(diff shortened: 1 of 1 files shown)


dustymabe:
This LGTM. Previously `rawhide` was pulling a package from F42 (it was in the pool) that had a higher NVR than the `rawhide` version of the package:

```
[2025-08-29T00:23:29.026Z]   zlib-ng-compat-2.2.5-2.fc42.x86_64 (fedora-coreos-pool)
```

but now it looks right:

```
  zlib-ng-compat-2.2.5-1.fc43.x86_64 (fedora-rawhide)
```

dustymabe (approved):

travier (approved):
Untested but looks reasonable.

travier:
Konflux build failures appear to be unrelated but I'm not sure and I don't know how to re-trigger them.

PR #3717, opened 2025-09-03
tests/validate-symlinks: Add broken symlinks from nfs-client-utils

PeaceRebel:
Broken links:
	1. /etc/nfsmount.conf.d/10-nfsv3.conf
	2. /etc/nfsmount.conf.d/10-nfsv4.conf

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/2012

Remove the test from denylist

### DIFF
--- a/tests/kola/files/validate-symlinks
+++ b/tests/kola/files/validate-symlinks
@@ -18,6 +18,8 @@ set -xeuo pipefail
 # - The symlinks in /usr/lib/firmware are often broken and are typically not an
 #   issue, so let's skip this location altogether.
 # - json-glib -> https://bugzilla.redhat.com/show_bug.cgi?id=2297094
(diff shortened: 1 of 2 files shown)


tests/kola/files/validate-symlinks
```diff
@@ -32,6 +32,8 @@ list_broken_symlinks_skip=(
     '/usr/share/rhel/secrets/etc-pki-entitlement'
     '/usr/share/rhel/secrets/redhat.repo'
     '/usr/share/rhel/secrets/rhsm'
+    '/etc/nfsmount.conf.d/10-nfsv3.conf'
+    '/etc/nfsmount.conf.d/10-nfsv4.conf'
```
travier, line 38: Let's add a comment at the top of the list there with the other ones linking to the issue for reference.

PeaceRebel: Done

travier (approved):

PR #3719, opened 2025-09-03
manifests: Exclude python again

travier:
We cleared up the dependency chain for the nfs-utils-coreos package in F43 so we can exclude Python again.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1942
Reverts: e70fd15d manifests/fedora-coreos: allow python in F43+ for now

---

Depends on https://src.fedoraproject.org/rpms/nfs-utils/pull-request/27


jlebon (approved):

jbtrystram (approved):
LGTM
Even though that's not helping the bootc install to-filesystem change :)

travier:
> LGTM Even though that's not helping the bootc install to-filesystem change :)

How so?

jbtrystram:
Having python in the image would allow using it as the buildroot

jbtrystram:
> > LGTM Even though that's not helping the bootc install to-filesystem change :)
> 
> How so?

Having python in the image would allow using it as the buildroot for osbuild tasks.

travier:
Will need https://bodhi.fedoraproject.org/updates/FEDORA-2025-ebd8554ce9 in branched/next

travier:
Rawhide has the fixed packages (the new split) so this should be good to go

dustymabe:
additional cleanup in https://github.com/coreos/fedora-coreos-config/pull/3731

> Even though that's not helping the bootc install to-filesystem change :)

indeed. I was hoping it would go through for that reason (among others)

PR #3723, opened 2025-09-04
konflux: enable hermetic builds

jbtrystram:
Using coreos/coreos-assembler#4298

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -21,6 +21,8 @@ import yaml
 ARCH = os.uname().machine
 SRCDIR = '/src'
 INPUTHASH = '/run/inputhash'
+HERMETIC_REPO = '/etc/yum.repos.d/cachi2.repo'
+IS_HERMETIC = os.path.exists(HERMETIC_REPO)
 
 
 def main():
@@ -51,8 +53,9 @@ def main():
         # NEVRAs to appear there. For lack of a generic solution for any repo
         # there, we only special-case the one place where we know we use this.
         if lockfile_repos == ['fedora-coreos-pool']:
-            modify_pool_repo(locked_nevras)
-            repos += lockfile_repos
+            if not IS_HERMETIC:
+                modify_pool_repo(locked_nevras)
+                repos += lockfile_repos
         elif len(lockfile_repos) > 0:
             raise Exception(f"unknown lockfile-repo found in {lockfile_repos}")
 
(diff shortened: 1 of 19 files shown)


build-rootfs
```diff
@@ -88,11 +88,16 @@ def inject_yumrepos():
         if os.path.basename(repo) == 'secret.repo':
             # this is a supported podman secret to inject repo files; see Containerfile
             continue
+        # cachi2 is an injected repo by konflux for hermetic build.
+        # We do nothing in that case.
+        if os.path.basename(repo) == 'cachi2.repo':
```
joelcapitao: We can use this `cachi2.repo` check to define whether or not we are in a hermetic build. This would save us changing the pipeline definition c.f https://github.com/konflux-ci/build-definitions/pull/2449.
We'd just have to add `HERMETIC=1` before continuing.

build-rootfs
```diff
@@ -88,11 +88,16 @@ def inject_yumrepos():
...
+            continue
         os.unlink(repo)
 
     # and now inject our repos
-    for repo in glob.glob(f'{CONTEXTDIR}/*.repo'):
-        shutil.copy(repo, "/etc/yum.repos.d")
+    # TODO conditional on hermetic
```
joelcapitao: This might be fix alongside https://github.com/coreos/fedora-coreos-config/pull/3723/files#r2352228001

.containerignore
```diff
@@ -17,3 +17,4 @@
 !/overlay.d/
 !/overrides/
 !/platforms.yaml
+!/content-sets.json
```
joelcapitao: this can be removed since https://gitlab.com/fedora/bootc/tekton-catalog/-/commit/b1a9bb60a0651116ceeea63e136b0327e3c44367

joelcapitao: wait, it's still not available in our CI because of https://github.com/coreos/fedora-coreos-config/pull/3797#issuecomment-3350497576

jbtrystram:
requires https://github.com/coreos/fedora-coreos-pipeline/pull/1245

joelcapitao:
See https://github.com/coreos/fedora-coreos-config/commit/c0abc9cd871327349951ab59994d83466e2740fe#r171804321 the glib2 URL in `rpms.lock.yaml` returns 404.

https://github.com/coreos/coreos-assembler/pull/4381 should fix the issue we are hitting

buildroot-prep
```diff
@@ -8,8 +8,11 @@ set -euo pipefail
 arch=$(uname -m)
 . /etc/os-release
 
-cp /src/fedora-coreos-continuous.repo /etc/yum.repos.d
-
+# cachi2 is the repo Konflux injects when hermetic build is enabled and
+# is self-sufficient to pull all the required RPMs.
+if [ ! -f "/etc/yum.repos.d/cachi2.repo" ]; then
+    cp /src/fedora-coreos-continuous.repo /etc/yum.repos.d
+fi
```
jbtrystram, line 15: IMHO we should add a parameter to this whole script and pass the arg from `build-rootfs` so we only have to hardcode this cachi2 path once.

jbtrystram: wait no that does not work.  Nevermind. Konflux should set an env var for the whole build

dustymabe, line 15: not copying the fedora-coreos-continuous repo here means the step below with `--repo fedora-coreos-continuous` will fail.

I wonder if the buildroot-prep stuff is just never going to work with hermetic?

OR - better, let's just fold calling buildroot-prep into build-rootfs as a another function that gets called (which shells out to the script). This means it can run after we call `inject_yumrepos()` and can use the hermeto repo to pull the RPM from if we want.

joelcapitao: Good point, let's give a try.
But note that the line `-repo fedora-coreos-continuous` is not run at all since a while now. Actually, the whole script is noop at the moment. But we still need to support what you proposed for future need.

joelcapitao: https://github.com/coreos/fedora-coreos-config/pull/3973

PR #3724, opened 2025-09-05
Containerfile: COPY build context

jlebon:
I like the efficiency of being able to just mount the build context instead of having to copy it in. For example, I often have git worktrees in my git checkout and I don't want `COPY` to dumbly copy those too (though I then realized we could just use a `.containerignore` for it).

However, a down side of mounting is that we render layer caching even more useless because it can't know which files from the mount we're actually using.

More recently, an issue around setgid leaking into the context dir (because OpenShift) came up (see
https://github.com/coreos/coreos-assembler/pull/4277) and it's just not acceptable to chmod the real build context directory.

Of course, `build-rootfs` could e.g. copy the overlay dirs into a temporary location and then adjust the perms there, but at that point, we're quite close to just doing a `COPY` upfront anyway.

So let's just switch to that, strip the bits as we want, but include a `.containerignore` to only copy what we need.

This then limits the build context mounting to just the rechunk and `FROM oci-archive:` hack. Which... hopefully one day we could normalize that.

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -1,0 +1,18 @@
+# by default, ignore everything
+*
+
+# only add what we actually need; this provides more accurate layer caching and
+# makes the COPY faster
+!/*.repo
+!/build-args.conf
+!/build-rootfs
+!/buildroot-prep
+!/image-base.yaml
+!/image-default.yaml
+!/image.yaml
+!/live/
+!/manifest-lock.*
+!/manifests/
+!/manifest.yaml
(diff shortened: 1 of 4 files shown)


jbtrystram (approved):
LGTM

dustymabe:
This appears to have broken `overrides/rpm` working for me.

This means that all of our bodhi updates testing is not testing what we think it is.

PR #3732, opened 2025-09-09
konflux: update pipeline reference

joelcapitao:
This new reference contains [1] which is not merged yet. Once merged we'll move back to official pipeline reference. We need this to unblock Konflux implementation.

[1] https://gitlab.com/fedora/bootc/tekton-catalog/-/merge_requests/4


jbtrystram (approved):
LGTM

PR #3733, opened 2025-09-09
[rawhide] drop grub2 fast-track, systemd pin

dustymabe:


dustymabe:
CI fails on rawhide because rawhide has been moved to building via container.. I'm also not sure why, but the links for CI are against the staging openshift cluster :(

aaradhak (approved):

aaradhak:
I had deployed the CI stg cluster today. Looks like the reason for CI to be against the staging openshift cluster. I am not sure what config to change to not affect the prod CI. I will try to look into it

PR #3737, opened 2025-09-11
configure container signature verification ; enable for `next*` and mechanical streams

dustymabe:
This includes the configuration for container signature checking
for our quay.io/fedora/fedora-coreos containers and also a migration
to move systems over to using a refspec that tells OSTree not to check
OSTree commit signatures, but to check signatures on container pulls instead.

See individual commit messages. Part of https://github.com/coreos/fedora-coreos-tracker/issues/2029


dustymabe:
untested at this point. getting the code out there for review, while I try to kick off testing in the background

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
```diff
@@ -0,0 +1,30 @@
...
+  "transports": {
+    "docker-daemon": {
+      "": [
+        {
+          "type": "insecureAcceptAnything"
+        }
+      ]
+    },
+    "docker": {
+      "quay.io/fedora": [
```
jlebon: Is there a reason this isn't `quay.io/fedora/fedora-coreos`?

overlay.d/35container-signing-migration/usr/lib/systemd/system/coreos-container-signing-migration-motd.service
```diff
@@ -0,0 +1,12 @@
+# This service is used for printing a message if
+# the container signing migration failed or was opted out.
+[Unit]
+Description=Warn based on container signing migration status
```
jlebon: Title Case

overlay.d/35container-signing-migration/usr/lib/systemd/system/coreos-container-signing-migration-motd.service
```diff
@@ -0,0 +1,12 @@
...
+Before=systemd-user-sessions.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-container-signing-migration-motd
+RemainAfterExit=yes
```
jlebon, line 10: Minor: space

overlay.d/35container-signing-migration/usr/libexec/coreos-container-signing-migration
```diff
@@ -0,0 +1,85 @@
...
+    echo "ERROR: The booted container imgref doesn't match the Fedora CoreOS default."
+    exit 1
+fi
+
+# Determine the updated imgref to use. THIS IS THE MIGRATION. Example:
+# from: ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:testing
+#   to: ostree-image-signed:registry:quay.io/fedora/fedora-coreos:testing
+#
+# https://github.com/coreos/rpm-ostree/blob/6f386be3d474aeb24fdc7f672d38151f95069223/docs/container.md?plain=1#L53
+updated_imgref=$(sed 's,ostree-remote-image:fedora:docker://,ostree-image-signed:registry:' <<< "${booted_imgref}")
```
dustymabe:
```suggestion
updated_imgref=$(sed 's,ostree-remote-image:fedora:docker://,ostree-image-signed:registry:,' <<< "${booted_imgref}")
```

will fix this in the next upload

overlay.d/35container-signing-migration/usr/lib/systemd/system/zincati.service.d/010-container-signing-migration.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+StateDirectory=coreos-container-signing-migration
+ExecStartPre=-/usr/libexec/coreos-container-signing-migration
+# wait for 3 min
+TimeoutStartSec=300
```
jlebon: Hmm, why do we need this?

overlay.d/35container-signing-migration/usr/libexec/coreos-container-signing-migration
```diff
@@ -0,0 +1,85 @@
...
+checksum=$(jq -r '.deployments[0].checksum' <<< "$status")
+digest=$(jq -r '.deployments[0]."container-image-reference-digest"' <<< "$status")
+
+# Proceed with the migration by writing an override status file,
+# so at the next update, Zincati will use container signing.
+echo "Writing a status override file for the booted deployment in /run/zincati/booted-status-override.json"
+cat > /run/zincati/booted-status-override.json << EOF
+{
+    "booted": true,
+    "container-image-reference": "ostree-unverified-image:$updated_imgref",
```
jlebon: Hmm, confused by the `ostree-unverified-image:` prefix. Doesn't `updated_imgref` already have the prefix?

jlebon: OK, looking at [the Zincati code](https://github.com/coreos/zincati/blob/c47f43edb1c35204d948d16c2520f185455ec244/src/rpm_ostree/actor.rs#L92), I'm unsure whether Zincati will do the right thing. Since in this migration we just want to change the prefix but not the OCI repo, it might not know to rebase and will try to do a deploy (thus keeping the unverified prefix).

So... we might need to fall back to tweaking the origin file.

overlay.d/35container-signing-migration/usr/libexec/coreos-container-signing-migration-motd
```diff
@@ -0,0 +1,73 @@
...
+${MIGRATION_OPT_OUT}, then restarting zincati.service"
+
+failed_message="
+The migration to OCI signed images for updates failed. Check the logs of
+zincati.service for more details.
+
+The system will likely fail to continue updating in the near future.
+
+When ready, the migration can be retried by deleting
+${MIGRATION_FAILED}, then restarting zincati.service"
```
jlebon: Should we link to e.g. https://github.com/coreos/fedora-coreos-tracker/issues/2029 in these messages?

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
dustymabe: For sure. It should be. Will update! Thanks

overlay.d/35container-signing-migration/usr/lib/systemd/system/coreos-container-signing-migration-motd.service
dustymabe: I delete the space above `[Service]` will be in the next upload.

overlay.d/35container-signing-migration/usr/lib/systemd/system/zincati.service.d/010-container-signing-migration.conf
dustymabe: I copied this from the previous iteration of the oci migration and that one had to reach out to the network, so I think we had to extend the timeout there in case it took a while for the network to come up. I'll delete this here.

overlay.d/35container-signing-migration/usr/libexec/coreos-container-signing-migration
dustymabe: this was a typo. it should just be:

```
"container-image-reference": "$updated_imgref",
```

overlay.d/35container-signing-migration/usr/libexec/coreos-container-signing-migration-motd
dustymabe: sounds like a good idea

overlay.d/35container-signing-migration/usr/libexec/coreos-container-signing-migration
jbtrystram: > OK, looking at [the Zincati code](https://github.com/coreos/zincati/blob/c47f43edb1c35204d948d16c2520f185455ec244/src/rpm_ostree/actor.rs#L92), I'm unsure whether Zincati will do the right thing. Since in this migration we just want to change the prefix but not the OCI repo, it might not know to rebase and will try to do a deploy (thus keeping the unverified prefix).

A trick could be to write the version tag or the full digest in the override status. This will lead Zincati to rebase because then the refspec wouldn't match. 
And since the override is only for Zincati, that value would not even be surfaced anywhere (except Zincati logs maybe)

dustymabe: latest upload handles this case by doing exactly what JB suggested.

dustymabe:
Notes for testing:

```
# 1. make overlay compatible as a systemd sysext

mkdir -p overlay.d/17fcos-container-signing/usr/lib/extension-release.d/
echo "ID='_any'" > overlay.d/17fcos-container-signing/usr/lib/extension-release.d/extension-release.17fcos-container-signing
mkdir -p overlay.d/35container-signing-migration/usr/lib/extension-release.d/
echo "ID='_any'" > overlay.d/35container-signing-migration/usr/lib/extension-release.d/extension-release.35container-signing-migration


# 2. Define a butane config to pick up those files from the overlay and write them out as sysext. Note had to disable selinux because running things on sysext was getting blocked

$ cat config.bu
variant: fcos
version: 1.6.0
passwd:
  users:
    - name: core
      password_hash: xyz
systemd:
  units:
    - name: systemd-sysext.service
      enabled: true
    - name: zincati.service
      enabled: false
storage:
  trees:
    - local: ./17fcos-container-signing/etc/containers/registries.d/
      path: /etc/containers/registries.d
    - local: ./17fcos-container-signing
      path: /var/lib/extensions/17fcos-container-signing
    - local: ./35container-signing-migration
      path: /var/lib/extensions/35container-signing-migration
  files:
    # SELinux->permissive because there are some denials running
    # things from a sysext.
    - path: /etc/selinux/config
      mode: 0644
      overwrite: true
      contents:
        inline: |
          SELINUXTYPE=targeted
          SELINUX=permissive

#3. Convert to Ignition

cat config.bu | butane --pretty --strict --files-dir overlay.d/ > config.ign


#4. Launch VM using the Ignition config using the last `stable` stream release of F41

cosa buildfetch --force --build=42.20250818.1.0 --artifact=qemu
cosa decompress --build=42.20250818.1.0
cosa run --build=42.20250818.1.0 --ignition=config.ign --ignition-direct --devshell=false --usernet


#5. Once the machine is up log in as `core` user and daemon-reload to pick up sysext changes, etc.

sudo mkdir /etc/systemd/system/zincati.service.d
sudo ln -s /dev/null /etc/systemd/system/zincati.service.d/010-oci-migration.conf
sudo systemctl daemon-reload
sudo systemctl stop rpm-ostreed
sudo /usr/bin/zincati-update-now
```

Inspired by https://github.com/coreos/fedora-coreos-config/pull/3458#issuecomment-2898576929

dustymabe:
I'll continue to do more testing on this today, but I think it's close. Marking as ready for review.

overlay.d/35container-signing-migration/usr/libexec/coreos-container-signing-migration
```diff
@@ -0,0 +1,95 @@
...
+MIGRATION_OPT_OUT=/var/lib/coreos-container-signing-migration/opt-out.stamp
+MIGRATION_FAILED=/var/lib/coreos-container-signing-migration/failed.stamp
+
+if [ -e $MIGRATION_OPT_OUT ]; then
+    echo "${MIGRATION_OPT_OUT} to opt out of migration exists, doing nothing."
+    exit 0
+fi
+
+# create the failed stamp file as a precaution
+touch $MIGRATION_FAILED
```
travier, line 21: Won't that fail if the folder does not exists?

dustymabe: I think the `StateDirectory=coreos-container-signing-migration` in `010-container-signing-migration.conf` handles that part.

travier: Ah indeed, I had missed that. But that means that's set for the entire Zincati service. But should be fine.

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
```diff
@@ -0,0 +1,30 @@
+{
+  "default": [
+    {
+      "type": "reject"
+    }
+  ],
+  "transports": {
```
travier, line 7: I think you need to define all transports here otherwise they will be blocked.

But maybe we don't care as it's not really something we expect people to do:

```
        "containers-storage": {
            "": [
                {
                    "type": "insecureAcceptAnything"
                }
            ]
        },
        "oci": {
            "": [
                {
                    "type": "insecureAcceptAnything"
                }
            ]
        },
        "oci-archive": {
            "": [
                {
                    "type": "insecureAcceptAnything"
                }
            ]
        }
```

travier (dismissed):
One nit but LGTM. Not tested

travier:
The question this raises is: How do we get out of this workaround? Hopefully at some point either bootc/rpm-ostree do things differently or we ask the container team to support different ways to configure the policy.

dustymabe:
> The question this raises is: How do we get out of this workaround?

I'll be verbose here to make sure I understand you.

The "workaround" being the bindmount of our custom policy of /etc/containers/policy.json. I think the good thing about this workaround is that we own the pieces. So when we no longer need it we just drop it from our overlay. Now how would we possibly not need this workaround in the future?

> or we ask the container team to support different ways to configure the policy.

I'd like to start here. We need to take time to properly discuss this with that team to see if these use cases could be considered. If that fails, then we can fall back to support in bootc I think.

Either way since the pieces of the workaround here are being delivered in /usr/ they are easy for us to control.

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
dustymabe: ahh ok. I think maybe I was confused a bit by the default policy that has:

```
[core@cosa-devsh ~]$ cat /etc/containers/policy.json 
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports":
        {
            "docker-daemon":
                {
                    "": [{"type":"insecureAcceptAnything"}]
                }
        }
}
```

is the 

```
            "docker-daemon":
                {
                    "": [{"type":"insecureAcceptAnything"}]
                }
        }
```

bit redundant?

dustymabe: I'll add the suggestions you made because I do think they should be there.

travier: I think you still need the `docker-daemon` part as well to be exhaustive.

travier: i.e. my suggestion is on top of what's already there

dustymabe: yes. Agree

dustymabe: added in latest upload.

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
```diff
@@ -0,0 +1,51 @@
...
```
jlebon, line 7: > I think you need to define all transports here otherwise they will be blocked.

Hmm, doesn't that allow `ostree-signed-image` too liberally? I.e. I think then people would be able to use that prefix even if it's not signed. Currently the dumb `policy.json` checker in bootc wouldn't block it, but in the future it would (because it'd tell skopeo directly to always enforce signatures).

To be clear, I think we can say this is a bug in bootc, though it feels funny making it easier to fall into that illusion.

jlebon: I think for now keeping what you had before and requiring people to mask the dropin if they want to rebase is OK (I guess they could technically also use `bootc switch` instead, which wouldn't have the bind mount).

dustymabe: > Hmm, doesn't that allow ostree-signed-image too liberally?

Could you enumerate some cases here? i.e. what are the cases and what do you think we should allow and shouldn't allow?

dustymabe: Here's one that I would want to continue to support without having to do an override:

```
State: idle                                                                                                           
Deployments:                                                                                                          
● ostree-image-signed:docker://quay.io/fedora/fedora-coreos:testing-devel                                             
                   Digest: sha256:289da93e6afa3dfc10d71c0e3a4f72bc67a53243fac9063bf7c09d4e61459738                    
                  Version: 42.20250912.dev.0 (2025-09-12T16:33:25Z)

[core@cosa-devsh ~]$ sudo rpm-ostree rebase ostree-unverified-image:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive                                                   
Pulling manifest: ostree-unverified-image:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive                                                                             
error: Preparing import: Fetching manifest: failed to invoke method OpenImage: Running image oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive: is rejected by policy.  
                                                                                                 
[core@cosa-devsh ~]$ sudo mkdir /etc/systemd/system/rpm-ostreed.service.d                                             
[core@cosa-devsh ~]$ sudo ln -s /dev/null /etc/systemd/system/rpm-ostreed.service.d/coreos-containers-policy.conf     
                                                                                                                                        
[core@cosa-devsh ~]$ sudo cp /usr/lib/coreos/coreos-containers-policy.json /etc/containers/policy.json                                                                                                                                      
[core@cosa-devsh ~]$ sudo vi /etc/containers/policy.json                                                              
                                                                                                  
[core@cosa-devsh ~]$ sudo systemctl stop rpm-ostreed                                                                                                                                                                                        
[core@cosa-devsh ~]$ sudo rpm-ostree rebase ostree-unverified-image:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive                                                   
Pulling manifest: ostree-unverified-image:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive                                                                             
Importing: ostree-unverified-image:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive (digest: sha256:050857f345ed74240faa1d853f255d6a12f20548509d7148e60e830ccc90889c)
ostree chunk layers needed: 65 (980.2 MB)                                                                             
custom layers needed: 1 (179 bytes)                                                                                                                                                                                                         
[0/66] Fetching ostree chunk c0d9d8b9f9a3272fd9d (42.1 MB)... done
```

dustymabe: ok I think I see what you mean.. basically the current updated policy (with timothee's suggestion) allows:

```
[core@cosa-devsh ~]$ sudo rpm-ostree rebase ostree-image-signed:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive                                                       
Pulling manifest: ostree-image-signed:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive                                                                                 
Importing: ostree-image-signed:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive (digest: sha256:050857f345ed74240faa1d853f255d6a12f20548509d7148e60e830ccc90889c)      
ostree chunk layers needed: 65 (980.2 MB)                                                                             
custom layers needed: 1 (179 bytes)                                                                                                                                                                                                         
[0/66] Fetching ostree chunk c0d9d8b9f9a3272fd9d (42.1 MB)... done

State: idle
Deployments:
  ostree-image-signed:oci-archive:/var/mnt/workdir/builds/44.20250912.dev.0/x86_64/fedora-coreos-44.20250912.dev.0-ostree.x86_64.ociarchive
                   Digest: sha256:050857f345ed74240faa1d853f255d6a12f20548509d7148e60e830ccc90889c
                  Version: 44.20250912.dev.0 (2025-09-12T15:34:56Z)
                     Diff: 431 upgraded, 3 downgraded, 4 removed, 16 added

● ostree-image-signed:docker://quay.io/fedora/fedora-coreos:testing-devel
                   Digest: sha256:289da93e6afa3dfc10d71c0e3a4f72bc67a53243fac9063bf7c09d4e61459738
                  Version: 42.20250912.dev.0 (2025-09-12T16:33:25Z)

```

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
```diff
@@ -0,0 +1,51 @@
...
+        }
+      ]
+    },
+    "docker": {
+      "quay.io/fedora/fedora-coreos": [
+        {
+          "type": "signedBy",
+          "keyType": "GPGKeys",
+          "keyPaths": [
+            "/etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-41-primary",
```
jlebon: Hmm, is there a reason for this one? Doesn't hurt of course.

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
```diff
@@ -34,6 +34,11 @@
       ]
     },
     "docker": {
+      "": [
+        {
+          "type": "insecureAcceptAnything"
+        }
+      ],
```
jlebon, line 41: This technically should be part of the previous commit.

overlay.d/35container-signing-migration/usr/lib/systemd/system/coreos-container-signing-migration-motd.service
jlebon: I more meant adding a space before `[Install]`. :)

jlebon (dismissed):

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
dustymabe: good catch - my fixup went in the wrong place

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
dustymabe: removed

overlay.d/17fcos-container-signing/usr/lib/coreos/coreos-containers-policy.json
dustymabe: fixed I think

dustymabe:
Added a commit here to rework the upgrade test (in conjuction with https://github.com/coreos/fedora-coreos-pipeline/pull/1223) to actually test container signing (which will make us keep the policy.json updated with the gpgkeys we need).

overlay.d/35container-signing-migration/usr/lib/systemd/system/coreos-container-signing-migration-motd.service
dustymabe: oops. either way not worth an extra CI cycle. If I do another push I'll fix it.

tests/kola/upgrade/extended/test.sh
```diff
@@ -319,18 +319,28 @@ systemd-run --wait --property=After=coreos-fix-selinux-labels.service \
...
-    rpm-ostree cleanup --rollback
-    # Pull the ociarchive from the builds dir here because the
-    # containers aren't pushed to quay until the release job is run
-    # and that hasn't happened yet.
-    curl -L -o /srv/update.ociarchive \
-        "https://builds.coreos.fedoraproject.org/prod/streams/${target_stream}/builds/${target_version}/${arch}/fedora-coreos-${target_version}-ostree.${arch}.ociarchive"
-    rpm-ostree rebase "ostree-unverified-image:oci-archive:/srv/update.ociarchive"
-    rm /srv/update.ociarchive
+
+    inspect=$(skopeo inspect --retry=3 -n docker://quay.io/fedora/fedora-coreos:${target_version})
```
jlebon: Hmm, I get:

```
FATA[0000] unknown flag: --retry                        
```

Did you mean `--retry-times`?

jlebon: Shouldn't the logic here be to check whether the image was pushed or not? As is, I think the `skopeo` command will fail if the image wasn't pushed, and because of `set -e`, we'll exit the script.

dustymabe: Fixed

dustymabe: This should have been `target_stream`. Fixed now.

jlebon (approved):

dustymabe:
This passed CI before the last upload and the last upload only changed a test that CI here doesn't cover. Merging.

PR #3738, opened 2025-09-11
build-rootfs: handle repos correctly in prod case

jlebon:
On prod streams, we _only_ have the lockfile repos. This made obvious a slight logic bug here. We want to inject our repo files if there's either a non-empty `repos` key or a non-empty `lockfile-repos` key.


dustymabe (approved):
LGTM

untested

PR #3743, opened 2025-09-14
[branched] switch to `ostree-image-signed` refspec for built images

dustymabe:
Part of https://github.com/coreos/fedora-coreos-tracker/issues/2029

Also drop the earlier customizations that shouldn't have been there any longer.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -2,3 +2,7 @@
 # similarly to manifest.yaml. Unlike image-base.yaml, which is shared by all
 # streams.
 include: image-base.yaml
+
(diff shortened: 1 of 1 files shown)


jlebon (approved):

jlebon:
CI is failing because it's not using the buildah path. I think at this point we should probably drop the pipecfg env var approach and instead have a marker of some kind in the config repo (e.g. some file, or setting a variable in `manifest.yaml`?) that tells cosa to default to the buildah path even without the env var. (But obviously still being able to opt out.) We want local developers, CI and prod to default to the same thing.

PR #3744, opened 2025-09-14
[next-devel] Re-activate next-devel branch; switch to ostree-image-signed refspec

dustymabe:


build-args.conf
```diff
@@ -2,7 +2,7 @@
 # Pass to buildah/podman using `--build-arg-file`.
 
 # This is the developer default version. In pipelines, this is driven by versionary.
-VERSION=42
+VERSION=42.dev
```
jlebon, line 5: Ahh yeah, nice catch.

jlebon (approved):

PR #3746, opened 2025-09-14
[stable] tree: promote changes from testing at aa81785bc15537bcb147384f892d59c3d24a47eb

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/1123.


dustymabe (approved):
LGTM - the check diffs is complaining about build-args.conf. I imagine we'll have to add an exception for the difference there in that file on each stream.

jlebon (approved):

marmijo (approved):
LGTM

PR #3747, opened 2025-09-14
[next] tree: promote changes from next-devel at f42c6ac541aba137b1336f22965d00d7f4c03320

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/1124.


dustymabe (approved):
LGTM - the check diffs is complaining about build-args.conf. I imagine we'll have to add an exception for the difference there in that file on each stream.

jlebon (approved):

marmijo (approved):
LGTM

PR #3754, opened 2025-09-15
[rawhide] manifest.yaml: default to buildah path

jlebon:
We switched rawhide over to the buildah path in the pipeline, but we want anyone/anything building rawhide to do the same. cosa recently learned to read `metadata.build_with_buildah` for this.

Make use of it.


dustymabe (approved):
LGTM

jbtrystram (approved):

PR #3757, opened 2025-09-16
s390x: treat EXIT_NO_PERMIT_CPI from cpi.service as success under Secure Execution

nikita-dubrovskii:
Since commit https://github.com/ibm-s390-linux/s390-tools/commit/ce9c518b977925cc4c9eb92a3e508762fd57f551
CPI no longer sends data for Secure Execution guests. This causes `cpi.service` to exit with `EXIT_NO_PERMIT_CPI`,
which `systemd` interprets as a failure. Adjust the unit to consider this exit code successful.

Issue: https://github.com/coreos/rhel-coreos-config/issues/64

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/01-secex.ign
@@ -76,6 +76,15 @@
       {
         "mask": true,
         "name": "autovt@.service"
+      },
+      {
+        "dropins": [
+          {
+            "contents": "# CPI may exit with code 6 (EXIT_NO_PERMIT_CPI)if permission is not granted\n[Service]\nSuccessExitStatus=6",
+            "name": "secex.conf"
+          }
+        ],
+        "name": "cpi.service"
       }
     ]
   }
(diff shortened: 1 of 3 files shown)


nikita-dubrovskii:
Tested on FCOS with secex: 
```
[core@localhost ~]$ systemctl status cpi.service
● cpi.service - Apply Control Program Identification (CPI)
     Loaded: loaded (/usr/lib/systemd/system/cpi.service; enabled; preset: enabled)
             └─10-timeout-abort.conf
             /etc/systemd/system/cpi.service.d
             └─secex.conf
     Active: active (exited) since Tue 2025-09-16 15:01:05 UTC; 1min 50s ago
 Invocation: 5f0576d811674b5893a976f2c4b8acf9
    Process: 5192 ExecStart=/lib/s390-tools/cpictl -e (code=exited, status=6)
   Main PID: 5192 (code=exited, status=6)
   Mem peak: 2.6M
        CPU: 62ms
```

jlebon:
Can we just completely mask it? Is there any point in having the unit still run?

hbrueckner:
> Can we just completely mask it? Is there any point in having the unit still run?

There is work on-going for this change in the s390-tools package to solve this problem. So let's keep that change for now.

HuijingHei (approved):

jlebon (approved):

PR #3758, opened 2025-09-16
build-rootfs: add com.coreos.inputhash label

jlebon:
Right now, there is no change detection in our builds. We just always do a new build, which is wasteful.

Let's re-implement a concept similar to rpm-ostree's inputhash, which just hashes all the relevant inputs. We then add a label to the final image with that input. On the cosa side, we can compare the hash to the previous build to know if to no-op.

One major difference from rpm-ostree's inputhash is that it could know the hash right after doing the depsolve (to have the final list of RPMs) and thus avoid a full build. It's possible to do this here too though it'd require bootc-base-imagectl and rpm-ostree changes. We'd need to also define an API for having a `podman build` actually signal "no-op"; e.g. writing a file in the build context and erroring the build... awkward.

Much more interesting and related to this is reproducible builds; if our builds were reproducible, we wouldn't have to worry about this because we'd just build the exact same artifact everytime. There's some work required to get there though, and likely we'd have to rework how we calculate our versions, since that's a dynamic value which affects the rootfs and OCI labels (e.g. always have the same version for the set of inputs; see also discussions in https://github.com/coreos/fedora-coreos-tracker/issues/2015).

---

Requires rpm-ostree v2025.11 in the buildroot.

### DIFF
--- a/buildroot-prep
+++ b/buildroot-prep
@@ -8,11 +8,12 @@ set -euo pipefail
 arch=$(uname -m)
 . /etc/os-release
 
-# Fast-track backport of https://github.com/coreos/rpm-ostree/pull/5475.
-# Comment this out if not used to not unnecessarily pay for repo metadata.
 cp /src/fedora-coreos-continuous.repo /etc/yum.repos.d
-if [ "${VERSION_ID}" == 42 ]; then
+
+# NOTE: try to remove anything that queries repos here once it's no longer
+# needed so that we don't unnecessarily pay for repo metadata.
+
+# make sure we have https://github.com/coreos/rpm-ostree/pull/5454
+if ! rpm-ostree compose build-chunked-oci -h | grep -q -- --label; then
     sudo dnf update rpm-ostree -y --repo fedora-coreos-continuous --releasever "$VERSION_ID"
-else
-    sudo dnf install rpm-ostree-2025.10 --allow-downgrade -y --repo fedora-coreos-continuous --releasever "$VERSION_ID"
 fi
--- a/Containerfile
+++ b/Containerfile
@@ -45,7 +45,8 @@ RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
 RUN --mount=type=bind,target=/run/src,rw \
       rpm-ostree experimental compose build-chunked-oci \
         --bootc --format-version=1 --rootfs /target-rootfs \
-        --output oci-archive:/run/src/out.ociarchive
+        --output oci-archive:/run/src/out.ociarchive \
+        --label com.coreos.inputhash=$(cat /run/inputhash)
 
 FROM oci-archive:./out.ociarchive
 ARG VERSION
--- a/build-rootfs
+++ b/build-rootfs
@@ -8,6 +8,7 @@
 # 5. It runs the postprocess scripts defined in the manifest.
 
 import glob
+import hashlib
 import json
 import os
 import shutil
@@ -19,6 +20,7 @@ import yaml
 
 ARCH = os.uname().machine
 SRCDIR = '/src'
+INPUTHASH = '/run/inputhash'
 
 
 def main():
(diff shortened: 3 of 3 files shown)


dustymabe (dismissed):

dustymabe:
> Requires rpm-ostree v2025.11 in the buildroot.

Mind mentioning why exactly? I think that's blocked on https://github.com/coreos/fedora-coreos-tracker/issues/2030

jlebon:
> > Requires rpm-ostree v2025.11 in the buildroot.
> 
> Mind mentioning why exactly? I think that's blocked on [coreos/fedora-coreos-tracker#2030](https://github.com/coreos/fedora-coreos-tracker/issues/2030)

For https://github.com/coreos/rpm-ostree/pull/5454.

dustymabe:
> > > Requires rpm-ostree v2025.11 in the buildroot.
> > 
> > 
> > Mind mentioning why exactly? I think that's blocked on [coreos/fedora-coreos-tracker#2030](https://github.com/coreos/fedora-coreos-tracker/issues/2030)
> 
> For [coreos/rpm-ostree#5454](https://github.com/coreos/rpm-ostree/pull/5454).

but was backported by you into f42 (2025.10-2) f43 (2025.10-3) f44 (2025.10-3), so we're not actually blocked?

jlebon:
> but was backported by you into f42 (2025.10-2) f43 (2025.10-3) f44 (2025.10-3), so we're not actually blocked?

Ahh, that's a different patch (https://github.com/coreos/rpm-ostree/pull/5475).

jbtrystram:
I feel this is doing the same thing as lockfiles.
If we have lockfiles committed, isn't the git commit a unique input hash ?

We already have the git commit hash attached to the image with konflux builds

jlebon:
> I feel this is doing the same thing as lockfiles. If we have lockfiles committed, isn't the git commit a unique input hash ?

Yeah, I should've mentioned this in the message.

First, not all streams are locked currently.

Second, on locked streams, we have more files than _just_ those that serve as inputs to the OCI image (e.g. tests). So we'd still end up churning more than necessary if we relied on just the git commit. At least, we'd need smarter triggering logic which understands the input set and filters the git log based on that.

The advantage of this approach is that it doesn't require any smart triggering logic and works for both kinds of streams.

I think this interacts with Konflux in two ways:
- Once we switch over to importing from Konflux, we could in theory rely on the CEL expressions for the locked streams but not the unlocked ones. We'll need Konflux to keep building the unlocked streams on a timer. The no-op'ing if nothing changed could happen in either Konflux or Jenkins after importing.
- Even if we do hermetic builds for currently unlocked streams, while I think we could craft really complete CEL expressions to get a good first past filtering, I think there's still value in having change detection implemented in the lower layers. Not only because it's bound to be more accurate, but also so that it works in other scenarios, like local development and CI outside of Konflux.

But anyway, to re-iterate my commit message though, to me the gold standard is still reproducible builds. Then we would get near-perfect change detection.

dustymabe (dismissed):
LGTM

jlebon:
OK had to adapt to https://github.com/coreos/fedora-coreos-config/commit/27f85ca7ec98c1acdd897d3d13b028b734e01d32 as well. But tested working well locally now!

One annoying thing for the local dev experience is that you're still paying for the cost of rechunking and the second stage `FROM` (which imports the intermediate OCI archive into containers-storage) even if it'll end up being a no-op. We could have a thing like `ARG ABORT_ON_INPUTHASH` where basically we touch a file and `exit 1` to stop the build if the hash is unchanged, but `cosa build-with-buildah` knows to treat it as a no-op in that case. Definitely hacky though.

buildroot-prep
```diff
@@ -8,11 +8,12 @@ set -euo pipefail
 arch=$(uname -m)
 . /etc/os-release
 
-# Fast-track backport of https://github.com/coreos/rpm-ostree/pull/5475.
-# Comment this out if not used to not unnecessarily pay for repo metadata.
 cp /src/fedora-coreos-continuous.repo /etc/yum.repos.d
```
dustymabe, line 11: should we comment out including the continuous repo when we're not actually doing any dnf operations?

jlebon: It's harmless to have, so just left it in there.

dustymabe (approved):

dustymabe:
LGTM

PR #3756, opened 2025-09-16
[next-devel] Start tracking Fedora 43

marmijo:
ref: https://github.com/coreos/fedora-coreos-tracker/issues/1935

PR #3760, opened 2025-09-16
Containerfile: add STRICT_MODE build arg

jlebon:
This re-implements rpm-ostree's `--lockfile-strict`, in which we want to ensure that only locked packages were installed.

We can't do this the same way as rpm-ostree did, i.e. as excludes pre-compose, but we can trivially check the result in post.

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -62,6 +62,9 @@ def main():
 
     if version != "":
         inject_version_info(target_rootfs, manifest['mutate-os-release'], version)
+    strict_mode = os.getenv('STRICT_MODE')
+    if strict_mode == '1':
+        verify_strict_mode(target_rootfs, locked_nevras)
     run_postprocess_scripts(target_rootfs, manifest)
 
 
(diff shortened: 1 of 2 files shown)


build-rootfs
```diff
@@ -62,6 +62,9 @@ def main():
 
     if version != "":
         inject_version_info(target_rootfs, manifest['mutate-os-release'], version)
+    strict_mode = os.getenv('STRICT_MODE')
```
dustymabe, line 65: does `ARG STRICT_MODE=0` set the `STRICT_MODE` an env variable?

dustymabe (approved):
LGTM

build-rootfs
jlebon: Yup, exactly.

jlebon:
The container-native flow passed. CoreOS CI won't test this here since testing-devel is still using the legacy path. Let's get the patches in and CI will test it in https://github.com/coreos/fedora-coreos-config/pull/3756.

PR #3765, opened 2025-09-17
denylist: add snooze for coreos.unique.boot.failure on aarch64

dustymabe:
Something changed in F43+ to cause this test to be very flaky. This is documented over in https://github.com/coreos/fedora-coreos-tracker/issues/2019

Let's add it to the denylist while we investigate.


marmijo (approved):
LGTM

PR #3767, opened 2025-09-17
buildroot-prep: pin rpm-ostree to 2025.10

dustymabe:
2025.11 (which is now in rawhide) suffers from an issue with SELinux labels. Let's "pin" in the container used as the buildroot here.

See https://github.com/coreos/fedora-coreos-tracker/issues/2030

### DIFF
--- a/buildroot-prep
+++ b/buildroot-prep
@@ -16,10 +16,3 @@ if [ "${VERSION_ID}" == 42 ]; then
 else
     sudo dnf install rpm-ostree-2025.10 --allow-downgrade -y --repo fedora-coreos-continuous --releasever "$VERSION_ID"
 fi
-
-# fast-track https://gitlab.com/fedora/bootc/base-images/-/merge_requests/279
-help=$(/usr/libexec/bootc-base-imagectl build-rootfs -h)
-if ! grep -q -- --lock <<< "$help"; then
-    echo "Patching bootc-base-imagectl"
-    sudo dnf install -y patch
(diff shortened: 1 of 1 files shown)


dustymabe:
also added a commit to drop a backport that's not longer needed.

buildroot-prep
```diff
@@ -11,12 +11,4 @@ arch=$(uname -m)
...
-sudo dnf update rpm-ostree -y --repo fedora-coreos-continuous --releasever "$VERSION_ID"
-
-# fast-track https://gitlab.com/fedora/bootc/base-images/-/merge_requests/279
-help=$(/usr/libexec/bootc-base-imagectl build-rootfs -h)
-if ! grep -q -- --lock <<< "$help"; then
-    echo "Patching bootc-base-imagectl"
-    sudo dnf install -y patch
-    curl -L https://gitlab.com/fedora/bootc/base-images/-/commit/35467bdbdcc01a8ad6b42c33c935ffe5c8c74395.patch | patch -p1 -d /usr/libexec
-fi
+sudo dnf install rpm-ostree-2025.10 --allow-downgrade -y --repo fedora-coreos-continuous --releasever "$VERSION_ID"
```
jlebon: On f42, it's just no-op'ing instead of pulling in 2025.10-2 which has the lockfile backport.

> Package "rpm-ostree-2025.10-1.fc42.x86_64" is already installed.

dustymabe: fixed

jlebon (approved):

HuijingHei (approved):
LGTM

PR #3770, opened 2025-09-18
[next-devel] lockfiles: drop graduated overrides 🎓

coreosbot-releng:


manifest-lock.overrides.yaml
```diff
@@ -39,24 +39,6 @@ packages:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2025-a9062c828c
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1935#issuecomment-3298693537
       type: fast-track
-  crun:
```
dustymabe, line 42: I think this one has some arch specific subpackages so we'll need to add a commit here to fixup the arch specific overrides files

dustymabe: fixed

jlebon (approved):

PR #3771, opened 2025-09-18
build-rootfs: mutate /usr/lib/initrd-release in initrd

jlebon:
In the legacy path, rpm-ostree handles `mutate-os-release` before
calling dracut. So the initrd-release in the initramfs (built from
os-release) contains the updated values.

In the buildah path, we handle version mutation after the rpm-ostree
compose done by bootc-base-imagectl and so the initrd doesn't have the
updated values.

I find that versioning information in the initrd useful because systemd
will print it out on the console (the "Welcome to ..." string), and a
lot of our bug reports are from failures in the initramfs.

So this patch restores this information for the buildah path, but does
it in a subtle way to avoid having to regenerate the initramfs just
for that: we create a temporary dracut module and then delete it in
postprocessing.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -57,11 +57,11 @@ RUN --mount=type=bind,from=builder,target=/var/tmp \
     --mount=type=bind,target=/run/src,rw \
       rm /run/src/out.ociarchive
 
-LABEL containers.bootc=1
-LABEL ostree.bootable=1
-LABEL org.opencontainers.image.version=$VERSION
-LABEL com.coreos.osname=$NAME
-LABEL org.opencontainers.image.title=$DESCRIPTION
-LABEL org.opencontainers.image.description=$DESCRIPTION
+LABEL containers.bootc=1 \
+      ostree.bootable=1 \
+      org.opencontainers.image.version=$VERSION \
+      com.coreos.osname=$NAME \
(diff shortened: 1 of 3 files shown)


dustymabe:
Do we not need some similar changes for bootc base images? Just wondering if there may be a better place for it.

jlebon:
The bootc base images don't today mutate os-release.

dustymabe:
I think this looks reasonable, but should we consider just making this an actual overlay rather than something done more dynamically. If someone rebuilds the initramfs client side it wouldn't have these changes?

dustymabe (approved):

jlebon:
> I think this looks reasonable, but should we consider just making this an actual overlay rather than something done more dynamically.

Yeah, though it's different in two ways:
- we'd have to mutate the overlay in build-rootfs to inject the version
- we delete the overlay in postprocessing

That would make it a weird overlay unlike the others. And I fully recognize that this is a hack. So I'm happy to keep it self-contained to `build-rootfs` at least.

> If someone rebuilds the initramfs client side it wouldn't have these changes?

Yeah, that's covered. I mentioned this in the code:

> ```
> # we can nuke this from the rootfs now; any dracut regeneration from this
> # point on (e.g. in derived builds, or client side) will be able to see the
> # os-release changes done by inject_version_info
> ```

PR #3772, opened 2025-09-19
test/files/setgid: add file to exception list for EL9

joelcapitao:
This reverts commit 4eb342e7f8f2da5a3eee73d108fd8afb80330d10.

That patch was included in Fedora, but it isn't in EL9 and likely won't be backported.

### DIFF
--- a/tests/kola/files/setgid
+++ b/tests/kola/files/setgid
@@ -14,6 +14,12 @@ list_setgid_files=(
     '/usr/bin/write'
     '/usr/libexec/utempter/utempter'
 )
+# Drop '/usr/libexec/openssh/ssh-keysign' after
+# https://src.fedoraproject.org/rpms/openssh/c/b615362fd0b4da657d624571441cb74983de6e3f?branch=rawhide
+# landed in EL9 (it's in Fedora and EL10 already).
+if match_maj_ver "9"; then
+    list_setgid_files+=('/usr/libexec/openssh/ssh-keysign')
+fi
 
 unknown_setgid_files=""
 while IFS= read -r -d '' e; do


jbtrystram (commented):
Another way of doing that would be to append `ssh-keysign` to the list only for EL9 . 
Helper here : https://github.com/coreos/fedora-coreos-config/blob/27f85ca7ec98c1acdd897d3d13b028b734e01d32/tests/kola/data/commonlib.sh#L101-L105

joelcapitao:
> Another way of doing that would be to append `ssh-keysign` to the list only for EL9 . Helper here :
> 
> https://github.com/coreos/fedora-coreos-config/blob/27f85ca7ec98c1acdd897d3d13b028b734e01d32/tests/kola/data/commonlib.sh#L101-L105

Thank you for pointing me to this helper !

tests/kola/files/setgid
```diff
@@ -14,6 +14,12 @@ list_setgid_files=(
     '/usr/bin/write'
     '/usr/libexec/utempter/utempter'
 )
+# Drop '/usr/libexec/openssh/ssh-keysign' after
+# https://src.fedoraproject.org/rpms/openssh/c/b615362fd0b4da657d624571441cb74983de6e3f?branch=rawhide
+# landed in EL9 (it's in Fedora and EL10 already).
+if match_maj_ver "9"; then
```
jlebon, line 20: This should probably be


```suggestion
if ! is_fcos && match_maj_ver "9"; then
```

It works as is because on FCOS that helper will `exit 1`, but it does print an error in the logs and makes this logic here look suspicious.

jlebon (approved):
One comment but LGTM as is to get the submodule bump unblocked faster.

Thanks for looking at this!

PR #3773, opened 2025-09-19
40ignition-ostree: bind mount stateroot_var_path rw

champtar:
From man mount(8):

    Since util-linux 2.39, mount may use the new kernel mount API if
    it is available. This new kernel interface provides a more precise
    way to work with mountpoint attributes. For example, the -o
    bind,rw operation will create a read-write node even if the
    original node was read-only.

ostree-prepare-root is creating a rw bind mount at stateroot_var_path because in the past, if a mount unit had 'Options=bind,rw', the 'rw' option was ignored.

Remount /sysroot/var rw so we can remove the stateroot_var_path rw
bind mount from ostree-prepare-root.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-var.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-var.sh
@@ -41,7 +41,9 @@ do_mount() {
     fi
 
     echo "Mounting $stateroot_var_path"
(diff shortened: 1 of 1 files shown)


champtar:
If this code also need to work with EL 9, we can make 2 mount calls instead of 1
EDIT: done

champtar:
All checks ok now

cgwalters:
Offhand looks sane to me, thank you!

jlebon (approved):
Thanks!

PR #3776, opened 2025-09-21
tests/kola: rework reboot-preparation to reduce timeouts

dustymabe:
Change it so that the autopkg-reboot-prepare only gets called once all the downloading from the network is done. Else we'll hit the 120 second timeout on reboot often because sometimes it takes a long time to download all the content for a major version rebase.


jbtrystram (approved):
Nice !

PR #3782, opened 2025-09-23
[rawhide] image.yaml: drop container-imgref override

dustymabe:
The image-base.yaml was updated to use this as the default value in b7e75bd so we don't need to carry this stream specific override here any longer.


jbtrystram (approved):
LGTM

PR #3787, opened 2025-09-24
tests/upgrade: add in SELinux label exception for <F43

dustymabe:
xref: https://github.com/coreos/fedora-coreos-tracker/issues/2030#issuecomment-3329932294

Also drop another exception that is no longer needed.


aaradhak (approved):
LGTM

PR #3791, opened 2025-09-26
#3790 - Fix coreos-disk-contains-fs.sh to handle device mappers

### LINKED ISSUE #3790: coreos-boot-disk link not working with multipath on early boot
On a scenario where a device mapper is used (ie: multipath), coreos-boot-disk is linked with the "physical" disk(ie: sda). Because of this the installation fails when try to changing partitions as coreos-boot-disk should be linked to the device mapper (ie: dm-0) to be allowed to partition the disk.

There is also a issue on ignition side that impacts the changes on partitions and also needs to be included to get this issue fixed
  - [2128](https://github.com/coreos/ignition/issues/2128) 
  - [PR](https://github.com/coreos/ignition/pull/2129) 

### To reproduce the issue run the command below

`cosa run --qemu-size +35G --qemu-multipath --kargs 'rd.multipath=default rd.debug rw' --debug --devshell-console --butane tmp/butane.yaml`

butane file contents:
```
variant: fcos
version: 1.6.0
kernel_arguments:
  should_exist:
    - rd.debug
    - rd.udev.log_level=debug
    - udev.log_level=debug
    - rd.systemd.log_level=debug
    - systemd.log_level=debug
storage:
  disks:
  - device: /dev/disk/by-id/coreos-boot-disk
    wipe_table: false
    partitions:
    - number: 4
      label: root
      size_mib: 20480
      resize: true
    - number: 5
      label: var
      size_mib: 0
  filesystems:
    - path: /var
      device: /dev/disk/by-partlabel/var
      format: ext4
      with_mount_unit: true
      wipe_filesystem: true
      label: var
```

tlbueno:
Change on coreos-disk-contains-fs.sh to also handle device mappers which has a different /sys/block structure when searching for partition labels

This fix is using sfdisk --json /dev/"${disk}" to looks into partitions because it can be used on usual disks and device mappers.

fixes #3790


tlbueno:
This fix is using kpartx -a -v /dev/"${disk}" to update the partitions of the device as it could take so long to show up and was causing the igntion-disk service to timeout.
I tried to move the 80-coreos-boot-disk.rules to be executed earlier by renaming it to 30-coreos-boot-disk and updated the required files as well as moved it to 99-coreos-boot-disk.rules but there were no effect on getting the partitions loaded quickly.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -1,20 +1,81 @@
...
+    fi
+else
+    echo "Checking standard block devices"
+    block_details_path="/sys/block/${disk}"
+    check_partitions "${block_details_path}" "${disk}" "${label}" "${is_dm}"
+fi
+
+echo "Label ${label} not found on block device ${disk}, returning 1"
 exit 1
+
```
joelcapitao: There is an extra line at EOF

tlbueno: Thanks I was trying to find out where it was. :-D

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -1,20 +1,80 @@
...
+        if [[ "${MPATH_DEVICE_READY:-}" == 1 ]]; then
+            echo "Block device ${disk} is multipath ready"
+        else
+            echo "Block device ${disk} is not multipath ready"
+        fi
+        block_details_path="/sys/block/${disk}/holders"
+        files=("${block_details_path}"/*)
+        if [[ ${#files[@]} -eq 0 ]]; then
+            echo "Partitions of block device ${disk} are not ready yet, forcing it"
+            kpartx -a -v /dev/"${disk}" || true
```
jlebon: I'm very wary of calling `kpartx` here and effectively taking over from the normal paths through which those partition devices are created (which I think is another udev rule shipped by multipath-tools).

Hmm, I think we should probably chat with @bmarzins about what we're trying to do and see if he has any suggestions.

tlbueno: I agree with you. I will reach @bmarzins to get all of us together and discuss about that.

tlbueno: I was testing it a bit more and tried a couple of things here. I tried use udevadm settle but it blocks the script for so long causing a timeout. Then I tried to move the udev rule to 99- instead of 80- and looks like it worked. In my previous test I tried 90- and for some reason it fail as 80-. I am going to push the update for review and do some more additional testing.

jlebon:
It's not clear right now whether re-ordering is the right way to do this because we may actually just be racing with the rule that uses `kpartx` to create the partition devices.

Instead I'm thinking we add another systemd unit which has `WantedBy=coreos-multipath-wait.target` (so it's only active in a multipath boot) and does `udevadm trigger /path/to/multipath/device` (which you can get at that point by looking at the parent of `/dev/disk/by-label/dm-mpath-boot`).

jlebon:
Actually, @bmarzins mentioned a more obvious solution, which is to just parse the partition table. E.g. basically change `coreos-disk-contains-fs.sh` to call `sfdisk --json $dev` and check (with a `jq` query) if one of the partitions has the desired label. That works for both multipath and non-multipath.

This is slightly different in check technically because we're no longer checking the filesystem label but the partition label. But for our use case here that's fine.

tlbueno:
> Actually, @bmarzins mentioned a more obvious solution, which is to just parse the partition table. E.g. basically change `coreos-disk-contains-fs.sh` to call `sfdisk --json $dev` and check (with a `jq` query) if one of the partitions has the desired label. That works for both multipath and non-multipath.
> 
> This is slightly different in check technically because we're no longer checking the filesystem label but the partition label. But for our use case here that's fine.

Looks a better solution. I will work on that and update there once completed. I am also start to do some tests with raid-1 as we suppose to support the same with that.

jlebon:
> I am also start to do some tests with raid-1 as we suppose to support the same with that.

Sorry, I know I'm the one who mentioned RAID earlier when we started looking at this issue, but actually I don't think we need to/even can support `/dev/disk/by-id/coreos-boot-disk` there. The RAID is done individually for each filesystem and not the parent block device, so there isn't a single "boot disk".

bmarzins:
> Sorry, I know I'm the one who mentioned RAID earlier when we started looking at this issue, but actually I don't think we need to/even can support `/dev/disk/by-id/coreos-boot-disk` there. The RAID is done individually for each filesystem and not the parent block device, so there isn't a single "boot disk".

Yeah, I noticed that coreos-disk-contains-fs.sh doesn't handle the case where the device itself directly contains the target filesystem. This knocks out LVM and MD (without weird workarounds), as well as filesystems on whole disks. It makes sense if you're using the symlink to point to the device where you should add other system partitions. But LVM should be able to give you that same ability. Instead of knowing the disk with the boot partition, you would just need to know the boot Volume Group, so you could use it to allocate new Logical Volumes. Does CoreOS not support systems built with LVM? That would get you RAID for free. Thin provisioning too, which might be useful on a bare-metal setup. Or maybe I'm just showing my CoreOS ignorance here.

tlbueno:
PR updated to include thea kola test and use sfdisk to discover the partition labels.
Regarding RAID, as @jlebon mentioned we will not support in this case
@jlebon what about the LVM mentioned by @bmarzins ?

PS: I just tried the https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_advanced_examples that has a raid1 to ensure that still working.

tlbueno:
added ext.config.multipath.custom-partition into kola-denylist until the ignition fix is released

bmarzins:
@tlbueno, just to make sure I understand this: If multipath is enabled on the kernel command line, then the boot disk must be a multipath disk for the "disk/by-id/coreos-boot-disk" symlink to get set. I'm assuming that you did it this way so that you won't end up picking one of the underlying path devices as the boot disk if multipath doesn't immediately claim the path device, since IIUC multipath won't do this on the first boot. And this restriction (the boot disk must be a multipath device if you enable multipathing in the kernel command line, otherwise you don't get the symlink set) is o.k. for you. Correct?

tlbueno:
> @tlbueno, just to make sure I understand this: If multipath is enabled on the kernel command line, then the boot disk must be a multipath disk for the "disk/by-id/coreos-boot-disk" symlink to get set. I'm assuming that you did it this way so that you won't end up picking one of the underlying path devices as the boot disk if multipath doesn't immediately claim the path device, since IIUC multipath won't do this on the first boot. And this restriction (the boot disk must be a multipath device if you enable multipathing in the kernel command line, otherwise you don't get the symlink set) is o.k. for you. Correct?

Yes, that is my understanding and how I did the fix. 

But now I am thinking about another scenario. What if I have a multipath device but not in the boot disk. 

Let's suppose that I would like to keep the usual partitioning on boot disk and use multipath for additional disks. I am not sure if that scenario is valid but I will need to dig into that a bit to check if in that case the fix will work.

tlbueno:
@jlebon @bmarzins I think that I found a corner case based on @bmarzins comment above.

If I want to keep the boot disk as a single disk but I want a multipath in additional disks the fix fails.

Like the butane example below. I need to have the kernel argument rd.multipath=default to enable the multipathd but since the script is looking only for device mappers, it fails. Any idea about how we could get this scenario working? 
```
variant: fcos
version: 1.6.0
storage:
  disks:
  - device: /dev/disk/by-id/coreos-boot-disk
    wipe_table: false
    partitions:
    - number: 4
      label: root
      size_mib: 10240
      resize: true
    - label: var
      size_mib: 0
      number: 5
  - device: /dev/disk/by-id/wwn-0xx000000000000000b
    wipe_table: true
    partitions:
    - number: 1
      label: containers
      size_mib: 0
  filesystems:
    - path: /var
      device: /dev/disk/by-partlabel/var
      format: ext4
      with_mount_unit: true
      wipe_filesystem: true
      label: var
    - path: /var/lib/containers
      device: /dev/disk/by-partlabel/containers
      format: ext4
      with_mount_unit: true
      wipe_filesystem: true
      label: containers
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -5,16 +5,22 @@ set -euo pipefail
...
-# during execution of udev rules on disks 'lsblk' returns empty fields
-for pt in /sys/block/$disk/*; do
-    name=$(basename $pt)
-    if [[ "$name" =~ ${disk}p?[[:digit:]] ]] && [[ -e "/sys/block/$disk/$name/start" ]];
-    then
-        eval $(udevadm info --query=property -n /dev/$name | grep -e ID_FS_LABEL -e PARTNAME)
-        if [[ "${ID_FS_LABEL:-}" == "$label" ]] || [[ "${PARTNAME:-}" == "$label" ]]; then
-            exit 0
-        fi
+echo "Running coreos-disk-contains-fs ${disk} ${label}"
```
jlebon: Let's get rid of the print statements.

Hmm, though I guess udev doesn't surface those anyway to the journal, so this is useful for when running the script directly for testing? Is that the reasoning?

What'd be nice is logging that respects udev's debug logging so that if one uses e.g. `rd.udev.log_priority=debug` we see these messages. Does udev already surface the stdout of helper scripts in debug mode?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -5,16 +5,22 @@ set -euo pipefail
...
+    if [[ ! -f /sys/block/${disk}/dm/name ]]; then
+        echo "Block device ${disk} is not a device mapper, returning 1"
+        exit 1
     fi
-done
+fi
 
+echo "Checking for partitions on ${disk} with label ${label}"
+has_label=$(sfdisk "/dev/${disk}" --json | jq '.partitiontable.partitions | any(.name == "'"${label}"'")')
+if [[ "${has_label}" == true ]]; then
```
jlebon: Minor/optional:

```suggestion
if sfdisk "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
jlebon, line 1: I guess we should rename this script now. E.g. `coreos-disk-contains-partition-name`?

tests/kola/multipath/custom-partition/test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+## kola:
+##   # This uses appendKernelArgs and multipath, which is QEMU only
+##   platforms: qemu
+##   description: Verify if multipath works with a custom partitioning
+##   appendKernelArgs: "rd.multipath=default rd.debug rd.udev.log_level=debug udev.log_level=debug rd.systemd.log_level=debug systemd.log_level=debug"
```
jlebon: This is tempting, but I personally would skip the debug logging kargs here. First it will drastically increase the logs we'd have to go through when doing a first pass triage of why the test failed. It's totally fine IMO for the second pass engineer to try to reproduce locally and enable debug logging there. Second, it might mask [Heisenbugs](https://en.wikipedia.org/wiki/Heisenbug).

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
```diff
@@ -5,16 +5,22 @@ set -euo pipefail
...
-    name=$(basename $pt)
-    if [[ "$name" =~ ${disk}p?[[:digit:]] ]] && [[ -e "/sys/block/$disk/$name/start" ]];
-    then
-        eval $(udevadm info --query=property -n /dev/$name | grep -e ID_FS_LABEL -e PARTNAME)
-        if [[ "${ID_FS_LABEL:-}" == "$label" ]] || [[ "${PARTNAME:-}" == "$label" ]]; then
-            exit 0
-        fi
+echo "Running coreos-disk-contains-fs ${disk} ${label}"
+
+if  [[ $(grep -q rd.multipath /proc/cmdline; echo $?) -eq 0 ]]; then
```
jlebon: This isn't strict enough. It's not doing a word boundary check and also not checking that the karg value isn't `0`. I would just copy/paste the logic in the `coreos-boot-mount-generator` that checks this karg too (including the `karg` helper function).

jlebon (commented):
Some comments but overall this looks good to me. Love how much cleaner the new approach is!

> This knocks out LVM and MD (without weird workarounds), as well as filesystems on whole disks.

Yeah, we don't really care about these use cases in this context.

> If I want to keep the boot disk as a single disk but I want a multipath in additional disks the fix fails.

Yeah, we today define `rd.multipath=default` as implying that the boot disk is multipathed. If you're only using multipath for a secondary disk, it's currently awkward but doable (see e.g. [this test](https://github.com/coreos/coreos-assembler/blob/0c3493c6e3efc4a602574adeb8b9d04b2c0c8df6/mantle/kola/tests/misc/multipath.go#L37-L103)). I think in the future we could try to loosen this but it needs to be done carefully.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-fs.sh
tlbueno: yes, I used it with the rd.debug flag. I will remove it for now.

kola-denylist.yaml
```diff
@@ -12,19 +12,24 @@
...
   platforms:
     - azure
 - pattern: coreos.unique.boot.failure
   tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2019
-  snooze: 2025-10-13
+  snooze: 2025-09-29
   warn: true
   arches:
     - aarch64
+    - x86_64
```
jlebon, line 27: Wait weird, why is this in this PR?

kola-denylist.yaml
```diff
@@ -12,19 +12,24 @@
...
+- pattern: ext.config.multipath.custom-partition
```
jlebon: I guess this is to have it gated until we have new Ignition?

I think usually what we do is we just hold the patch until we actually have the new package land.

jlebon (commented):
Can we squash it down to one commit?

kola-denylist.yaml
tlbueno: Oh, that is because it was failing not only on aarch but also on x86_64. I think it needs be moved to another PR.

kola-denylist.yaml
tlbueno: yes, that was the reason.
So, let's keep it open until the ignition is released.

tlbueno:
> Can we squash it down to one commit?

done!

jlebon (dismissed):
Nice work!

HuijingHei:
Build FCOS with ignition-2.24.0-1.fc42, run test passed.
```
[coreos-assembler]$ kola run ext.config.multipath.custom-partition
=== RUN   ext.config.multipath.custom-partition
--- PASS: ext.config.multipath.custom-partition (67.67s)
PASS, output in tmp/kola/qemu-2025-10-15-1418-86748
```

tests/kola/multipath/custom-partition/test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+if ! cat /proc/cmdline | grep -q rd.multipath=default; then
+    fatal "kernel argument rd.multipath=default not defined"
+fi
+
+boot_src=$(findmnt -nvr -o SOURCE /boot)
+dm_mpath=$(udevadm info --query=property "${boot_src}" | grep DM_MPATH | cut -d"=" -f2)
```
HuijingHei: Copy @jbtrystram 's comment https://github.com/coreos/fedora-coreos-config/pull/3832#discussion_r2444801026:
you can replace this with a less complex `multipath -l -v 1`

prestist (approved):

PR #3796, opened 2025-09-26
feat(pipeline): enable ppc64le and s390x architectures

Roshan-R:
Now that PowerPC (ppc64le) and s390x architectures are supported in the Fedora Konflux cluster, this commit enables them across all Fedora CoreOS Tekton pipeline configurations.


jbtrystram (approved):
LGTM

jlebon:
Nice! This is huge.

PR #3814, opened 2025-10-07

marmijo:
Some packages in F42 are sorting as newer than F43 due to the juggle of packages in between beta and final freeze. Fast track them in F43 so they don't show as downgraded. Today they are:

- `libdrm 2.4.126-1.fc42 -> 2.4.125-2.fc43`
- `libjcat 0.2.5-1.fc42 -> 0.2.3-2.fc43`
- `skopeo 1:1.20.0-3.fc42 -> 1:1.20.0-2.fc43`
- `vim-data 2:9.1.1818-1.fc42 -> 2:9.1.1775-1.fc43`
- `vim-minimal 2:9.1.1818-1.fc42 -> 2:9.1.1775-1.fc43`

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -37,6 +37,18 @@ packages:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2025-1a9b66e130
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1935#issuecomment-3298693537
       type: fast-track
+  libdrm:
+    evr: 2.4.126-1.fc43
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2025-551d41aa4b
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1935#issuecomment-3298693537
+      type: fast-track
+  libjcat:
+    evr: 0.2.5-1.fc43
+    metadata:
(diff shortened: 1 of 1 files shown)


manifest-lock.overrides.yaml
```diff
@@ -49,3 +61,21 @@ packages:
       bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2025-1a9b66e130
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/1935#issuecomment-3298693537
       type: fast-track
+  skopeo:
+    evr: 1:1.20.0-2.fc43
+    metadata:
+      bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2025-e5464b6660
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/1935#issuecomment-3298693537
+      type: fast-track
```
dustymabe, line 69: This is already in F43, but it shows as a downgrade right now because there is indeed a newer NVR in F42. I think we need to work with the team to get a new build for F43.

That being said adding the fast-track here doesn't hurt and the graduated overrides job will just remove it later so LGTM

dustymabe (approved):

manifest-lock.overrides.yaml
marmijo: Thanks, didn't catch that! I can try to reach out to the maintainer to request a new F43 build.

PR #3825, opened 2025-10-13
[testing] tree: promote changes from testing-devel at ed41a6dc8723a8ff39e9f1a62696559e83ec5801

coreosbot-releng:
Triggered by @aaradhak in https://github.com/coreos/fedora-coreos-streams/issues/1139.


marmijo:
check promotion diffs is failing because of the expected difference in the `DESCRIPTION` field of build-args.conf.
I opened https://github.com/coreos/fedora-coreos-config/pull/3827 to try to ignore the stream-specific diff.

marmijo (approved):

PR #3827, opened 2025-10-13
ci/promotion-diff: ignore DESCRIPTION diff in build-args.conf

marmijo:
The `DESCRIPTION` line in build-args.conf is stream-specific and expected to differ between branches. Filter it out before comparing trees so the promotion diff check doesn't fail for the expected difference.


aaradhak (approved):
LGTM

dustymabe (approved):

PR #3833, opened 2025-10-15
[next-devel] manifest: add fedora-candidate-compose to list of repos for F43

dustymabe:
Late in the Fedora major release process adding this repo will allow us to pull in candidate RPMs that are proposed to fix release blocking or freeze exception issues. We will be able to test out these proposed fixes sooner and also be more prepared for releasing our final `next` release once the release is deemed "Go".

See https://github.com/coreos/fedora-coreos-config/pull/2705

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -15,5 +15,5 @@ repos:
   # GitHub Action.
   - fedora-next
   - fedora-next-updates
-  #- fedora-candidate-compose
+  - fedora-candidate-compose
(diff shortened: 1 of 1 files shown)


aaradhak (approved):
LGTM

dustymabe:
Skipping CI since it won't test this change (i.e. lockfiles weren't updated so contentset won't change).

Will kick off a bump-lockfile run right after this.

PR #3836, opened 2025-10-17
fedora-archive.repo: add archives mirrors

jbtrystram:
Add another set of mirrors for archives in case dl.fedoraproject.org/pub/archive timeouts or flake

We've had the pipeline fail for older releases on that so a set of extra mirrors can help

### DIFF
--- a/fedora-archive.repo
+++ b/fedora-archive.repo
@@ -14,6 +14,8 @@ baseurl=https://dl.fedoraproject.org/pub/fedora/linux/releases/$releasever/Every
         https://dl.fedoraproject.org/pub/fedora-secondary/releases/$releasever/Everything/$basearch/os/
         https://dl.fedoraproject.org/pub/archive/fedora/linux/releases/$releasever/Everything/$basearch/os/
         https://dl.fedoraproject.org/pub/archive/fedora-secondary/releases/$releasever/Everything/$basearch/os/
+        https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/$releasever/Everything/$basearch/os/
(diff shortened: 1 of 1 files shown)


dustymabe:
Hmm. didn't know about a second site. Are there any docs for it?

travier:
I don't think that's the same thing. Those archives are for EOL Fedora releases, not for archived updates for current releases.

dustymabe:
> I don't think that's the same thing. Those archives are for EOL Fedora releases, not for archived updates for current releases.

I think we are using archives for EOL releases here so he is in the right context. From the file:

https://github.com/coreos/fedora-coreos-config/blob/4b6d1fc3727c94524a2b54a08dc987e8cafd8428/fedora-archive.repo#L1-L9

jbtrystram:
> Hmm. didn't know about a second site. Are there any docs for it?

I discovered it by following the archive link that was in this .repo file for secondary arch and found this [readme](https://dl.fedoraproject.org/pub/fedora-secondary/releases/38/README)
If you look 1 level above there is nothing there, but only for secondary arches.

dustymabe:
right, but it feels a bit awkward to have a link to both

- https://dl.fedoraproject.org/pub/archive/fedora-secondary/
- https://archives.fedoraproject.org/pub/archive/fedora-secondary/

i.e. I can almost guarantee they are just the same exact thing. In other words, if one of them fails they will probably both fail?

In the cases where you saw a failure did a rerun succeed (i.e. the automatic rerun that we do not a manually initiated rerun)?

dustymabe:
closing as stale. re-open if this is still relevant

PR #3839, opened 2025-10-18

marmijo:
F43 is now in final freeze. This means that packages in F42 will sort as newer than those in F43. We'll prevent this 'downgrade' by fast-tracking any pacakge that would violate a 'no-downgrade' rule.

Today they are:
```
linux-firmware
moby-engine
hwdata
nmstate
selinux-policy
stalld
```

see: https://github.com/coreos/fedora-coreos-tracker/issues/1935#issuecomment-3298693537


marmijo:
/remove-hold

I fixed some errors in the overrides files, this should be good to go.

dustymabe:
LGTM

PR #3840, opened 2025-10-20
kola-denylist.yaml: add x86_64 on coreos.unique.boot.failure

tlbueno:
Since it is failing on x86_64 as well I am adding it to the deny list.


marmijo (approved):
LGTM

aaradhak (approved):

PR #3848, opened 2025-10-24
[next] tree: promote changes from next-devel at c7d4a30154b45770252ae6181d3fc70f5e60570f

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/1156.


dustymabe:
added a second commit here to pin on the builder image digest (from konflux built bootc images since those don't get garbage collected) so that when we promote this to `testing` and to `stable` things remain consistent.

This is manually doing https://github.com/coreos/fedora-coreos-tracker/issues/2037 for now.

marmijo:
LGTM. promotion diffs is failing for the new addition, but that's expected.

marmijo (approved):

PR #3849, opened 2025-10-24
[testing] tree: promote changes from next at e7ec23371a288438b0a187e3e07c2ec7bd18f9b2

dustymabe:
This promotes `testing` to F43 and does the promotion directly from the `next` branch where the Fedora GA content was recently built.

It also switches to container building by default, which is part of the Fedora 43 changes.


marmijo (approved):
LGTM!

PR #3851, opened 2025-10-24
[next-devel] build-args: pin on a specific bootc builder image

dustymabe:
This will guaruantee when things are promoted later we have have consistency when those builds are done.


dustymabe:
Note I pinned on an older image for now so we can watch to see if https://github.com/coreos/fedora-coreos-pipeline/pull/1250 works.

marmijo (approved):
LGTM

PR #3858, opened 2025-10-29
build-rootfs: support postprocess directly

dustymabe:
This builds on the upstream change [1] to support being able to pass
in post-process scripts directly so that we can make changes during
the initial compose (prior to dracut running), which is important
in some cases.

[1] https://gitlab.com/fedora/bootc/base-images/-/merge_requests/315

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -57,12 +57,13 @@ def main():
     nodocs = (manifest.get('documentation') is False)
     recommends = manifest.get('recommends')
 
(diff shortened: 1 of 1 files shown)


dustymabe:
Closing this in favor of https://github.com/coreos/fedora-coreos-config/pull/3869.

See the discussion in https://gitlab.com/fedora/bootc/base-images/-/merge_requests/315

PR #3862, opened 2025-10-30
35coreos-ignition: find sfdisk path

HuijingHei:
Thank you Dusty for pointing out the root cause in
https://github.com/coreos/fedora-coreos-config/pull/3862#issuecomment-3467947591

For centos stream and rhcos, the tests that want to use
"/dev/disk/by-id/coreos-boot-disk" failed with
```
ignition[728]: disks: createPartitions: op(1): [failed]   waiting
for devices [/dev/disk/by-id/coreos-boot-disk]: device unit
dev-disk-by\x2did-coreos\x2dboot\x2ddisk.device timeout
```
e.g. `ext.config.shared.ignition.stable-boot`

Detail logs:
`/usr/lib/udev/coreos-disk-contains-partition-name: line 28: sfdisk: command not found`

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
@@ -25,9 +25,23 @@ if  [[ -n ${multipath} ]]; then
     fi
 fi
 
-if sfdisk "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
+# On Fedora (with /bin_/sbin merge), sfdisk is located at /usr/bin/sfdisk,
+# so udev helper scripts find it automatically via PATH. On RHEL/RHCOS 9.6,
+# sfdisk resides in /usr/sbin, but udev worker processes reset PATH to a
+# minimal environment (/usr/local/bin:/usr/bin), causing scripts that call
+# 'sfdisk' without the full path to fail. So we need this to find the full
+# sfdisk path.
+# See https://github.com/coreos/fedora-coreos-config/pull/3862#issuecomment-3467947591
+sfdisk_cmd=''
+for f in /usr/bin/sfdisk /usr/sbin/sfdisk; do
+    if [ -x "${f}" ]; then
+        sfdisk_cmd="${f}"
+        break
+    fi
+done
+
+if [ -n "${sfdisk_cmd}" ] && ${sfdisk_cmd} "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
     exit 0
 fi
 
 exit 1
-


joelcapitao:
This would fix https://github.com/coreos/fedora-coreos-config/pull/3791 but I'm wondering why it's working on Fedora ? Are we adding `sfdisk` somewhere else ?

HuijingHei:
Might misunderstand, we already have sfdisk in [rhel-coreos-config](https://github.com/coreos/rhel-coreos-config/blob/main/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh#L58) and [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh#L79).
Do testing with this change, no lucky.
```
# journalctl -u systemd-udevd.service | grep coreos-disk-contains-partition-name
Oct 30 09:59:57 localhost (udev-worker)[811]: vda: /run/udev/rules.d/80-coreos-boot-disk.rules:6 PROGRAM=="coreos-disk-contains-partition-name $name boot": Running command "
coreos-disk-contains-partition-name vda boot"
Oct 30 09:59:57 localhost (udev-worker)[811]: vda: Starting 'coreos-disk-contains-partition-name vda boot'
Oct 30 09:59:57 localhost (udev-worker)[811]: vda: 'coreos-disk-contains-partition-name vda boot'(err) '/usr/lib/udev/coreos-disk-contains-partition-name: line 28: sfdisk: c
ommand not found'
Oct 30 09:59:57 localhost (udev-worker)[811]: vda: Process 'coreos-disk-contains-partition-name vda boot' failed with exit code 1.
```

dustymabe:
I think maybe this is a `$PATH` issue. On Fedora (where the [bin/sbin merge](https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin) has already happened)`lsinitrd` shows:

```
-rwxr-xr-x   1 root     root       111072 Jan  1  1970 usr/bin/sfdisk
```

Where on RHEL/RHCOS 9.6:

```
-rwxr-xr-x   1 root     root       106656 Jan  1  1970 usr/sbin/sfdisk
```

If it is a path issue we could just modify the script to "find" the sfdisk to use (search bin, then sbin) and then use that.

HuijingHei:
> I think maybe this is a `$PATH` issue. On Fedora (where the [bin/sbin merge](https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin) has already happened)`lsinitrd` shows:
> 
> ```
> -rwxr-xr-x   1 root     root       111072 Jan  1  1970 usr/bin/sfdisk
> ```
> 
> Where on RHEL/RHCOS 9.6:
> 
> ```
> -rwxr-xr-x   1 root     root       106656 Jan  1  1970 usr/sbin/sfdisk
> ```
> 
> If it is a path issue we could just modify the script to "find" the sfdisk to use (search bin, then sbin) and then use that.

Nice catch, thank you! Update to find the sfdisk path.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
```diff
@@ -25,9 +25,12 @@ if  [[ -n ${multipath} ]]; then
     fi
 fi
 
-if sfdisk "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
-    exit 0
+sfdisk_bin=$(find /usr/{,s}bin -mindepth 1 -maxdepth 1 -name sfdisk)
```
dustymabe: I feel like calling `find` here is overkill.

```
sfdisk_cmd=''
for f in /usr/bin/fdisk /usr/sbin/fdisk; do
    if [ -x "${f}" ]; then
        sfdisk_cmd="${f}"
        break
    fi
done
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
```diff
@@ -25,9 +25,12 @@ if  [[ -n ${multipath} ]]; then
...
+if [ -z "${sfdisk_bin}" ]; then
+    echo "Could not find sfdisk"
+    exit 1
 fi
 
-exit 1
+eval $(${sfdisk_bin} "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")') && exit 0
```
dustymabe: why do we need to change to using `eval` here? Can we leave it the way it was?

dustymabe:
> 35coreos-ignition: get sfdisk path

Maybe `find sfdisk path` would be better

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
HuijingHei: I thought it causes ci/validate error, do we need to update the regrex (https://github.com/coreos/fedora-coreos-config/blob/testing-devel/ci/validate#L35) to skip this?
```
> ci/validate
  File "/home/hhei/github/fedora-coreos-config/ci/validate", line 56, in <module>
    validate_shell(rootfd, name)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/hhei/github/fedora-coreos-config/ci/validate", line 40, in validate_shell
    raise Exception(f'Possible unbracketed conditional in {name}')
Exception: Possible unbracketed conditional in coreos-disk-contains-partition-name.sh
```

HuijingHei: No need to change the regrex. Update to 
```
if [ -n "${sfdisk_cmd}" ] && ${sfdisk_cmd} "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
    exit 0
fi
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
```diff
@@ -25,9 +25,16 @@ if  [[ -n ${multipath} ]]; then
     fi
 fi
 
-if sfdisk "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
+sfdisk_cmd=''
+for f in /usr/bin/sfdisk /usr/sbin/sfdisk; do
```
dustymabe, line 36: my only suggestion here is to add a comment on why we need this because it's not super clear :)

the context is in the commit message, but if someone were to come in and try to clean this up they might not look there.

dustymabe (dismissed):
code LGTM

tlbueno:
This is also related to #3790

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
HuijingHei: Updated, thank you!

tlbueno (approved):
lgtm

PR #3863, opened 2025-10-31
Add a Justfile

cgwalters:
I'm trying to spread this as a pattern, see also the Justfile in bootc.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -3,6 +3,8 @@
 #     podman build --security-opt=label=disable --cap-add=all --device /dev/fuse \
 #         --build-arg-file build-args.conf -v $PWD:/run/src . -t localhost/fedora-coreos
 #
+# (This is also in Justfile - if you change the above also change that)
+#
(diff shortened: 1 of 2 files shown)


jlebon:
This has crossed my mind multiple times while hacking on this. One thing I value a lot is the idea that "all you need to get started is podman", and I personally like that we document it just that way for passers-by. If you want full developer convenience, that's today still cosa. E.g. `cosa build` _is_ that wrapper.

(I realize you've kept the Containerfile comment with the `podman` command, but again I think we need to refine a bit who we're targeting given `cosa build`.)

(Also, I think we already mentioned we could get rid of `--device /dev/fuse` and ideally eventually `-v $PWD:/run/src`. We just need a `--privileged` flag to replace `--security-opt=label=disable --cap-add=all` :laughing: )

cgwalters:
Just for general visibility I'm continuing the push on containers-as-native-devenv with Justfile also over here https://github.com/coreos/rpm-ostree/pull/5525

PR #3864, opened 2025-10-31
tests/kola: Add exceptions in upgrade test SELinux checks

aaradhak:
The `/var/opt/kola*` files are failing in the extended upgrade tests. Since they are only test artifacts, we will be adding an exception for these mislabeled-files check.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/2052

### DIFF
--- a/tests/kola/upgrade/extended/test.sh
+++ b/tests/kola/upgrade/extended/test.sh
@@ -196,6 +196,9 @@ selinux-sanity-check() {
         # - Would relabel /var/lib/systemd/random-seed from system_u:object_r:init_var_lib_t:s0 to system_u:object_r:random_seed_t:s0
         #       - 42.20250526.1.0 -> 42.20250609.1.0
         #       - https://github.com/coreos/fedora-coreos-tracker/issues/1965#issuecomment-2959831808
+        # - Would relabel /var/opt/kola* from unconfined_u:object_r:var_t:s0 to unconfined_u:object_r:usr_t:s0
+        #       - 42.20250410.2.0 -> 43.20251031.20.0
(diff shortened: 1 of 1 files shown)


tests/kola/upgrade/extended/test.sh
```diff
@@ -196,6 +196,11 @@ selinux-sanity-check() {
         # - Would relabel /var/lib/systemd/random-seed from system_u:object_r:init_var_lib_t:s0 to system_u:object_r:random_seed_t:s0
         #       - 42.20250526.1.0 -> 42.20250609.1.0
         #       - https://github.com/coreos/fedora-coreos-tracker/issues/1965#issuecomment-2959831808
+        # - Would relabel /var/opt/kola from unconfined_u:object_r:var_t:s0 to unconfined_u:object_r:usr_t:s0
+        # - Would relabel /var/opt/kola/extdata from unconfined_u:object_r:var_t:s0 to unconfined_u:object_r:usr_t:s0
+        # - Would relabel /var/opt/kola/extdata/commonlib.sh from unconfined_u:object_r:var_t:s0 to unconfined_u:object_r:usr_t:s0
```
dustymabe:
```suggestion
        # - Would relabel /var/opt/kola* from unconfined_u:object_r:var_t:s0 to unconfined_u:object_r:usr_t:s0
```

Simplify.. it down a bit.. gets the point across without being too verbose.

dustymabe (dismissed):

tests/kola/upgrade/extended/test.sh
aaradhak: I will update it

aaradhak: I have updated it

dustymabe (approved):

PR #3865, opened 2025-11-01
Fix sfdisk path on coreos-disk-contains-partition-name.sh

### LINKED ISSUE #3790: coreos-boot-disk link not working with multipath on early boot

tlbueno:
Use sfdisk full path as /usr/sbin is not part of the default path on older releases

fixes https://github.com/coreos/fedora-coreos-config/issues/3790

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-disk-contains-partition-name.sh
@@ -25,7 +25,7 @@ if  [[ -n ${multipath} ]]; then
     fi
 fi
 
-if sfdisk "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
+if /usr/sbin/sfdisk "/dev/${disk}" --json | jq -e '.partitiontable.partitions | any(.name == "'"${label}"'")'; then
     exit 0
 fi
 


dustymabe:
this should be taken care of by https://github.com/coreos/fedora-coreos-config/pull/3862

tlbueno:
Thanks @dustymabe going to close it and I am tracking #3862.

PR #3867, opened 2025-11-04
Add STREAM build arg and generic com.coreos.stream label for image builds

Roshan-R:
This PR adds support for labeling CoreOS-derived images with their stream during build time. 

Introduces a new `STREAM` arg in the `Containerfile` and `build-args.conf` so the image can be labeled with `fedora-coreos.stream=$STREAM` directly during the build. This removes the temporary labeling hack used in `cmd-build-with-buildah`. 
See: [coreos-assembler#4337](https://github.com/coreos/coreos-assembler/issues/4337)

Also adds a generic com.coreos.stream label for both FCOS and RHCOS images. This helps Jenkins jobs importing Konflux-built OCI images via cosa import determine the stream.
See: [COS-3604](https://issues.redhat.com/browse/COS-3604)

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -48,7 +48,7 @@ RUN --mount=type=bind,target=/run/src,rw \
         --bootc --format-version=1 --rootfs /target-rootfs \
         --output oci-archive:/run/src/out.ociarchive \
         --label com.coreos.inputhash=$(cat /run/inputhash) \
-        --label fedora-coreos.stream=$STREAM
+        --label com.coreos.stream=$STREAM
 
 FROM oci-archive:./out.ociarchive
 ARG VERSION
--- a/Containerfile
+++ b/Containerfile
@@ -20,6 +20,7 @@ FROM ${BUILDER_IMG} as builder
 
 ARG VERSION=overridden
 ARG MANIFEST=overridden
+ARG STREAM=overridden
 # XXX: see inject_passwd_group() in build-rootfs
 ARG PASSWD_GROUP_DIR
 ARG STRICT_MODE=0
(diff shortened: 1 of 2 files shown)


joelcapitao (dismissed):
Thank you Roshan.
This is the first step to get rid of [1] and get back to the self-contained path.
The next step is to update the Jenkins pipeline to consume it. Currently, it's handled by feeding `STREAM=$BRANCH` with [2].

[1] https://gitlab.com/fedora/bootc/tekton-catalog/-/blob/main/tasks/prepare-build-context/prepare-build-context.yaml#L87-89
[2] https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/blob/main/cluster/kfluxfedorap01/coreos-tenant/fedora-coreos/releaseplans/base/releaseplan.yaml#L34

Containerfile
```diff
@@ -46,7 +47,9 @@ RUN --mount=type=bind,target=/run/src,rw \
       rpm-ostree experimental compose build-chunked-oci \
         --bootc --format-version=1 --rootfs /target-rootfs \
         --output oci-archive:/run/src/out.ociarchive \
-        --label com.coreos.inputhash=$(cat /run/inputhash)
+        --label com.coreos.inputhash=$(cat /run/inputhash) \
+        --label fedora-coreos.stream=$STREAM \
```
dustymabe: to be completely honest I don't like the `fedora-coreos.stream` label here and would prefer we dropped it and only used the `com.coreos.stream` label.

@jbtrystram is zincati a user of this? Could we update it to handle the more generic label? @jlebon do you know of any other consumers?

jbtrystram: > to be completely honest I don't like the `fedora-coreos.stream` label here and would prefer we dropped it and only used the `com.coreos.stream` label.

+1 

> @jbtrystram is zincati a user of this? Could we update it to handle the more generic label? 

Yes. Roshan started looking at this already

dustymabe (approved):
LGTM - we'll need to open PRs on all the other branches here too to update build-args.conf STREAM=

jbtrystram (approved):

joelcapitao (approved):

PR #3868, opened 2025-11-05
Update bootc image with renovate

joelcapitao:
Konflux is designed to be event-driven, more precisely commit-driven
only, whereas Jenkins offers more flexible triggering i.e: time-driven
or event-driven.
For clarification, that's only for the builds. For the test and release
Konflux pipelines, they can be triggered with CronJobs by taking the
latest completed build.
The underlying reason is to make reproducible builds possible, all the
information should be self-contained in the repo. By reproducible I
mean having the same build result everywhere e.g: CI or locally, at
any give time (within a defined time window) from the same commit.
But this is an issue when the software depends on external dependencies
like RPM repos. RPM lockfiles fixes this, ensuring we have the updated
package periodically committed in the repo, thus triggering a new build.

However, we don't enable rpm-ostree lockfiles for rawhide, thus we
cannot generate Konflux lockfiles by derivation as we do for development
streams e.g: testing-devel. For clarification, Konflux RPM lockfiles is
only required for hermetic build purpose, and we don't want to enable
it for rawhide for now. In another words, having rpm-ostree lockfiles
would be enough.

This patch is taking another path, which is to bump the digest of the
bootc base image with Renovate. This is a cheap way to have periodic
build for rawhide. This does not solve the reproducibility, which
can be fixed by pulling RPMs from the compose repos instead of the
mirrors.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -34,6 +34,33 @@
       "rebaseWhen": "behind-base-branch"
     },
     {
+      "matchPackageNames": [
+        "/^quay.io/bootc-devel/fedora-bootc*/"
+      ],
+      "matchBaseBranches": [
+        "rawhide"
+      ],
+      "schedule": ["before 5am"],
+      "enabled": true,
+      "automerge": true,
+      "automergeType": "branch",
+      "groupName": "bootc base image",
+      "branchPrefix": "renovate/bump-bootc-base-image/",
+      "group": {
+        "branchTopic": "{{{baseBranch}}}",
+        "commitMessageTopic": "{{{groupName}}}"
+      },
+      "commitMessageTopic": "bootc base image",
+      "prBodyColumns": [
+        "Package",
+        "Change",
+        "Notes"
+      ],
+      "prBodyTemplate": "{{{header}}}{{{table}}}{{{notes}}}{{{changelogs}}}{{{configDescription}}}{{{controls}}}{{{footer}}}",
+      "recreateWhen": "always",
+      "rebaseWhen": "behind-base-branch"
+    },
+    {
       "matchDatasources": [
         "docker"
       ],
--- a/renovate.json
+++ b/renovate.json
@@ -15,6 +15,8 @@
         "testing-devel"
       ],
       "enabled": true,
+      "automerge": true,
+      "automergeType": "branch",
       "groupName": "Konflux references",
       "branchPrefix": "renovate/konflux/references/",
       "group": {


dustymabe:
> This does not solve the reproducibility,

I'm not sure what problem this PR solves. Can you be specific about what problem we are solving by doing this?

I'd much prefer not having more PRs get opened even if they do get automerged.

joelcapitao:
> > This does not solve the reproducibility,
> 
> I'm not sure what problem this PR solves. Can you be specific about what problem we are solving by doing this?
> 
> I'd much prefer not having more PRs get opened even if they do get automerged.

`This is a cheap way to have periodic build for rawhide` - to just have daily commit/builds.
But it's a draft now, because circling back with JB this morning, we came up with a better solution to have a daily builds AND have reproducibility.
The solution is:

1. Use compose repos e.g https://kojipkgs.fedoraproject.org/compose/rawhide/ in `fedora-rawhide.repo`.
2. Ask renovate to update the URL without opening PRs.

What do you think ?

dustymabe:
> What do you think ?

not sure that really does what we want either. There's definitely a lot of nuance here.

TBH if we want reproducibility in rawhide we might as well just run the bump-lockfile job for it?

Basically if we're going to have git be the "source of truth for everything" in `rawhide` then I don't want things to get committed that haven't passed tests.

git either locks everything (our prod and dev streams today) OR it shouldn't lock anything IMO.

Now there is some good reason to have some reproducibility in `rawhide`. Our bodhi tests. They currently use `--autolock` today to get back some reproducibility and we could pull the container digest from the previous build too to achieve the same goal.

Happy to discuss this more in a call if you like.

joelcapitao:
> They currently use --autolock today to get back some reproducibility and we could pull the container digest from the previous build too to achieve the same goal.

Got it, I think it's related to this discussion https://github.com/coreos/fedora-coreos-releng-automation/issues/181
which makes me wonder, why not adding full lockfiles to rawhide in the first place ?

> TBH if we want reproducibility in rawhide we might as well just run the bump-lockfile job for it?

So yeah, we considered that path with JB but that would be more complex to implement and resource-intensive (i.e running the tests for rawhide). But I'm ok with that path as well. As you said, this would be an enhancement of the current "intermediate approach" of lockfiles for rawhide.

I guess we have a consensus ? cc/ @jbtrystram

dustymabe:
> > They currently use --autolock today to get back some reproducibility and we could pull the container digest from the previous build too to achieve the same goal.
> 
> Got it, I think it's related to this discussion [coreos/fedora-coreos-releng-automation#181](https://github.com/coreos/fedora-coreos-releng-automation/issues/181) which makes me wonder, why not adding full lockfiles to rawhide in the first place ?

You answered your question below. For `rawhide` where we don't need 100% perfection when it comes to our builds it's less efficient to run a full build and suite of tests and then commit to git, which then triggers a real build, which does a lot of the same work that had just been done.

We decided for rawhide that it's good enough to just let it fail in a real build and investigate it there.

> 
> > TBH if we want reproducibility in rawhide we might as well just run the bump-lockfile job for it?
> 
> So yeah, we considered that path with JB but that would be more complex to implement and resource-intensive (i.e running the tests for rawhide). But I'm ok with that path as well. As you said, this would be an enhancement of the current "intermediate approach" of lockfiles for rawhide.
> 
> I guess we have a consensus ? cc/ @jbtrystram

I really just don't understand why konflux is so inflexible it can't handle this use case. We really can't do a new build without requiring a new git commit?

jbtrystram:
> I'm not sure what problem this PR solves. Can you be specific about what problem we are solving by doing this?

Joel already kinda answered this but in a nutshell : We want daily rebuilds of rawhide, and if there is no new commits since the last build, Konflux build will stop early because `no new commits`.
Konflux cron jobs only work for release tasks (i.e. taking the last good build and promoting it to a released state.).

So if we want daily builds, we need to commit something to the repo. Rather than pushing a dummy commit writing a date to a text file we came up with this solution of bumping the repo URL to the daily compose. 
This would have given us reproducible builds as a bonus point (the compose repo is available for 10 days).


What I wasn't aware until now is that rawhide packages are tagged into `coreos-pool`. In this case, committing the full lockfile works for us, I agree. 

To address the load on Jenkins we could add a knob to `bump-lockfile` that skips tests. for mechanical streams. That's what we started to do before pivoting to this PR.

jbtrystram:
> For `rawhide` where we don't need 100% perfection when it comes to our builds it's less efficient to run a full build and suite of tests and then commit to git, which then triggers a real build, which does a lot of the same work that had just been done.
> 
> We decided for rawhide that it's good enough to just let it fail in a real build and investigate it there.

So couldn't we rethink this and skip the tests that have already been done ? if we commit lockfiles and those have already passed the QEMU tests, and PRs do the same (but only x86, that's a big limitation), we could have `releases` (i.e. the `build` job in Jenkins) only do the liveiso and testiso tests then.
Since konflux does not throw away builds that would work . I am sidetracking a bit here but that's worth discussing

joelcapitao:
We can close this PR c.f https://github.com/coreos/fedora-coreos-tracker/issues/2038#issuecomment-3498258143

PR #3869, opened 2025-11-05
build-rootfs: run dracut ourselves, pass through recommends

dustymabe:
This commit leverages recent work upstream to:

1. pass through recommends: true if needed (used in rhel-coreos-config)
2. disable initramfs generation in the initial rpm-ostree compose and run it ourselves after our postprocess scripts have run.

It also adds some compat for python3.9, to support running on EL9.

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -141,22 +141,23 @@ def build_rootfs(target_rootfs, manifest_path, packages, locked_nevras, overlays
     if nodocs and tmpd is not None:
         del tmpd
 
+def get_bootc_base_imagectl_help():
+    return subprocess.check_output(['/usr/libexec/bootc-base-imagectl', 'build-rootfs', '-h'], encoding='utf-8')
+
 
 def repo_arg_supported():
     # Detect if we have https://gitlab.com/fedora/bootc/base-images/-/merge_requests/248.
     # If not, then we can't use `--repo`. That's OK because that should only
     # happen on RHEL, where we don't have any default repos anyway and only rely on
     # the mounted secret repo file.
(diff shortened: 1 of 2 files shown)


dustymabe:
Note that there is one difference here in the images generated using this. The initramfs has fewer users and groups in `/etc/passwd` and `/etc/group`. I've opened https://gitlab.com/fedora/bootc/tracker/-/issues/78 to discuss the implications of this.

dustymabe:
> Note that there is one difference here in the images generated using this. The initramfs has fewer users and groups in `/etc/passwd` and `/etc/group`. I've opened https://gitlab.com/fedora/bootc/tracker/-/issues/78 to discuss the implications of this.

I think the realization in https://gitlab.com/fedora/bootc/tracker/-/issues/78#note_2873023424 means that we can dismiss this as a non-issue and continue here.

Marked this PR as ready for review.

build-rootfs
```diff
@@ -481,7 +481,9 @@ def calculate_inputhash(rootfs, overlays, manifest):
         all_files = sorted(all_files)
         for file in all_files:
             with open(file, 'rb') as f:
-                h.update(hashlib.file_digest(f, 'sha256').digest())
+                # When python3.11+ is the minimal version we can use hashlib.file_digest
+                # h.update(hashlib.file_digest(f, 'sha256').digest())
+                h.update(hashlib.sha256(f.read()).digest())
```
aaradhak, line 473: On trying to understand the difference between usage of file_digest and f.read() , I learnt that file_digest would be reading data in chunks whereas f.read would load the entire file into memory, wanted to know if that is a concern here?

dustymabe: The files that we read here are mostly just small text files so I'm not too concerned.

travier (approved):
Not tested but LGTM

PR #3870, opened 2025-11-07
Konflux: update on-cel to pick up trigger file changes

jbtrystram:
Write the initial stamp file that will be checked by Konflux to trigger
builds. This file is to be bumped by either bump-lockfile or the
coreos config bot when syncing the config changes from testing-devel.
It will be bumped at most once a day, ensuring we don't trigger builds
too often, needlessly increasing the load on the pipeline.

See coreos/fedora-coreos-tracker#2038 (comment)


.tekton/base/on-pull-request/fedora-coreos-on-pull-request.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && ".tekton/trigger".pathChanged()
```
joelcapitao: We still want to build an ostree container image at each PR, so I wouldn't change the _on-cel-expression_ of PR. 
BTW `on-pull-request` _on-cel-expression_ are specific for each stream e.g: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.tekton/rawhide/on-pull-request/kustomization.yaml#L27-L29

.tekton/base/on-push/fedora-coreos-on-push.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
     build.appstudio.redhat.com/commit_sha: '{{revision}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && ".tekton/trigger".pathChanged()
```
joelcapitao: We then need to remove the specific `on-push` _on-cel-expression_ e.g: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.tekton/rawhide/on-push/kustomization.yaml#L27-L29 otherwise this will be overwritten by them i.e: that's why the script `ci/generate-tekton-pipelinerun` didn't change the real pipelines when you run it

joelcapitao (changes requested):
I checked the `pathChanged()` method, it only checks if the content of the file changed, and not the metadata. It suits well our need.

.tekton/trigger
dustymabe, line 1: https://github.com/coreos/fedora-coreos-releng-automation/pull/220/files#r2503507682

joelcapitao: With https://github.com/coreos/fedora-coreos-releng-automation/compare/fbcf8565ebe375d3190870e8decb7891e26e2b00..3faf8c87c2780ad08c9aa2966c14e467ef3846d2 we don't need to initialize the file in `testing-devel` anymore

joelcapitao (dismissed):
LGTM
This should be merged before https://github.com/coreos/fedora-coreos-pipeline/pull/1257

.tekton/trigger-file-comment
```diff
@@ -0,0 +1,8 @@
+# Do not edit this file manually. This file is bumped by
+# either bump-lockfile or the coreos config bot when syncing
+# the branch. It will be bumped at most once a day, and watched
+# by konflux to trigger builds. This ensure we don't trigger
```
dustymabe: ensures

dustymabe:
LGTM - one suggestion.

I don't know much about konflux so I'll have to rely on @joelcapitao's expertise for the review of that change.

joelcapitao (approved):

PR #3871, opened 2025-11-07
[rawhide] Konflux: initialize the trigger file

jbtrystram:
Write the initial stamp file that will be checked by Konflux to trigger builds. This file is to be bumped by either bump-lockfile or the coreos config bot when syncing the config changes from testing-devel. It will be bumped at most once a day, ensuring we don't trigger builds too often, needlessly increasing the load on the pipeline.

See https://github.com/coreos/fedora-coreos-tracker/issues/2038#issuecomment-3498258143


joelcapitao:
This is not needed as it will be updated and propagated by `config-bot`.

PR #3874, opened 2025-11-10
WIP: Changes for booting composefs native systems

Johan-Liebert1:
I don't believe what we have here would be the final product. But this might be a good starting point


overlay.d/05core/usr/lib/dracut/modules.d/40bootc-coreos/bootc-disable.preset
```diff
@@ -0,0 +1,4 @@
+# Disable services that don't work with bootc/composefs
```
travier: We should probably add a condition to those unit that makes them not run on composefs native systems.

overlay.d/05core/usr/lib/dracut/modules.d/40bootc-coreos/ignition-ostree-check-rootfs-size.conf
```diff
@@ -0,0 +1,9 @@
+[Unit]
+After=
+After=bootc-initramfs-setup.service
```
travier: An option here could be to create a wrapping unit/script that runs the right command depending on the kernel parameter, to avoid duplicating everything.

overlay.d/05core/usr/lib/dracut/modules.d/40bootc-coreos/ignition-ostree-mount-var.conf
```diff
@@ -0,0 +1,10 @@
+[Unit]
+ConditionKernelCommandLine=
```
travier: Let's see if we can figure out a way to include both condition in the unit:
> If multiple conditions are specified, the unit will be executed if all of them apply (i.e. a logical AND is applied). Condition checks can use a pipe symbol ("|") after the equals sign ("Condition…=|…"), which causes the condition to become a triggering condition. If at least one triggering condition is defined for a unit, then the unit will be started if at least one of the triggering conditions of the unit applies and all of the regular (i.e. non-triggering) conditions apply. If you prefix an argument with the pipe symbol and an exclamation mark, the pipe symbol must be passed first, the exclamation second. If any of these options is assigned the empty string, the list of conditions is reset completely, all previous condition settings (of any kind) will have no effect.

https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Conditions%20and%20Asserts

overlay.d/05core/usr/lib/dracut/modules.d/40bootc-coreos/bootc-disable.preset
Johan-Liebert1: Yes, that would be better

overlay.d/05core/usr/lib/dracut/modules.d/40bootc-coreos/ignition-ostree-mount-var.conf
Johan-Liebert1: Yeah that would the way to do it. I have it in separate drop-ins right now just to keep track of what all I'm changing

overlay.d/05core/usr/lib/dracut/modules.d/37bootc/bootc-root-setup.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=bootc setup root
```
cgwalters: We definitely have the dracut module now this shouldn't be necessary

cgwalters:
Chatting about confidential computing and stuff so this doesn't get lost: I think we are also going to need to support setting up dm-integrity

cgwalters:
Also I just want to say here: I have come to the conclusion that I made a mistake in trying to create a stronger conceptual separation between "image mode/bootc" vs "coreos" - especially with respect to Ignition. It's of course *very* complicated because there was a rationale for that - integration with Anaconda unlocked a lot of use cases.

But still..I think we should prioritize the combined teams working on this PR and making the whole thing feel seamless. I know parts of it have been happening, but there's so much more to do (especially things like zincati -> bootc etc).

And a big goal should be "sealed FCOS" w/UKI.

travier:
> Chatting about confidential computing and stuff so this doesn't get lost: I think we are also going to need to support setting up dm-integrity

I think that should be set up by the Ignition config?

overlay.d/05core/usr/lib/dracut/modules.d/37bootc/bootc-root-setup.service
cgwalters: https://gitlab.com/fedora/bootc/base-images/-/merge_requests/426

overlay.d/05core/usr/lib/dracut/modules.d/37coreos-root-setup/coreos-root-setup.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=CoreOS Root Setup
+Documentation=man:bootc(1)
+DefaultDependencies=no
+ConditionKernelCommandLine=|composefs
```
cgwalters, line 5: Ideally...we don't hardcode this because I'd like the ability to change it in the future. That already came up because e.g. not having a `.` makes the Linux kernel emit a warning i.e. I'd like to switch to `composefs.digest` or so.

But also, maybe we actually switch to key based mounting per https://github.com/composefs/composefs-rs/pull/226

So...how about `ConditionPathExists=|/sysroot/composefs` ?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/boot-part-discovery.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+XBOOTLDR_UUID="BC13C2FF-59E6-4262-A352-B275FD6F7172"
```
cgwalters, line 3: Hmm it'd be nice to not write new load-bearing bash code run as root. In this case...can't the GPT auto generator mount this?

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -306,9 +306,32 @@ case "${1:-}" in
...
+                    hash_alg="sha256"
+                elif [[ $cfs_digest =~ ^[0-9a-fA-F]{128}$ ]]; then
+                    hash_alg="sha512"
+                else
+                    echo "Bad verity in cmdline '$cfs_digest'"
+                    exit 1
+                fi
+
+                echo "Enabling fs-verity again..."
+                find /sysroot/composefs/objects -type f -exec fsverity enable {} --hash-alg $hash_alg \;
```
cgwalters, line 331: Eeek. This is another thing exactly like https://github.com/ostreedev/ostree/pull/3094

I think at a minimum we could put this in rust in the bootc initramfs setup like `bootc-initramfs-setup post-copy /sysroot/composefs` ?

Also note that once we have https://github.com/composefs/composefs-rs/pull/268 then the tooling itself will immediately know the expected digest size just looking at the repo.

overlay.d/05core/usr/lib/dracut/modules.d/37coreos-root-setup/coreos-root-setup.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=CoreOS Root Setup
```
cgwalters, line 2: Hmm this is effectively a generic synchronization point between "ostree or composefs" right? I think we could just add this to bootc upstream as say `bootc-root-setup.target` - which then we ensure via a generator depends on the right service.

The bash code executed here is just reimplementing the upstream services which I think we don't want.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=Ignition OSTree: Grow Root Filesystem
 DefaultDependencies=false
-ConditionKernelCommandLine=ostree
+ConditionKernelCommandLine=|ostree
```
cgwalters, line 4: I said something like this before in a different PR on this repo but I think this would be a good time to change our logic so that we don't depend on the specific kernel command line at all.

One way we could handle this is in the initramfs we introduce `/run/bootc-requested` which we synthesize from a generator (like `/run/ostree-booted`). I think this should only go into the initramfs (unlike the ostree one) because right now we have another way to detect composefs (look at the mount source for `/`) and anyways the post-boot code should largely be talking to bootc anyways.

This `/run/bootc-requested` is something we could again do in bootc upstream right now.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/boot-part-discovery.sh
Johan-Liebert1: > can't the GPT auto generator mount this?

Yes, it could, and would mount /boot if the partuuid follows DPS, which it does after my other PR. This was mainly supposed to be a sync target for a lot of other services that depend on a boot partition being present.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
Johan-Liebert1: > I think at a minimum we could put this in rust in the bootc initramfs setup like bootc-initramfs-setup post-copy /sysroot/composefs ?

agreed. That would be much cleaner

And yes, repo metdata would help here

overlay.d/05core/usr/lib/dracut/modules.d/37coreos-root-setup/coreos-root-setup.service
Johan-Liebert1: > So...how about ConditionPathExists=|/sysroot/composefs ?

This sounds good to me, but I don't know if systemd will OR this with the kernel cmdline condition. I'll need to check

cgwalters:
Also xref https://github.com/systemd/systemd/pull/29731 - i'd like to opitmize the LUKS-without-filesyostem change case.

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-growfs.service
Johan-Liebert1: Discussed this in a 1-1 call. This is a good to have, but not a blocker for now

overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-transposefs.sh
```diff
@@ -306,9 +306,32 @@ case "${1:-}" in
         if [ -d "${saved_root}" ]; then
             echo "Restoring rootfs from RAM..."
             mount_and_restore_filesystem_by_label root /sysroot "${saved_root}"
```
cgwalters, line 308: Discussing this live:

For the sealed UKI case let's say we want to do the DPS. Then probably we should detect whether the disk has the DPS root UUID and if the bootloader set the BLI variable.

I think we could make this a clean distinct prep commit right? Basically "support DPS booting".

Questions:

- Will we need to suppress the generator and redo what it's doing for the Ignition transposefs case? Though this relates to a systemd PR around optimizing the cryptsetup case when we don't need to reinitialize the partition table for the root at all.

PR #3879, opened 2025-11-11
build-rootfs: Don't follow symlinks in the overlay

cgwalters:
Add the link target to the hash instead.

This broke the ostree builds with
`[2025-11-11T03:36:28.572Z] FileNotFoundError: [Errno 2] No such file or directory: '/src/overrides/rootfs/etc/grub.d/15_ostree'`

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -467,12 +467,17 @@ def calculate_inputhash(rootfs, overlays, manifest):
                 all_files.append(os.path.join(root, file))
         all_files = sorted(all_files)
         for file in all_files:
-            with open(file, 'rb') as f:
-                # When python3.11+ is the minimal version we can use hashlib.file_digest
-                # h.update(hashlib.file_digest(f, 'sha256').digest())
-                h.update(hashlib.sha256(f.read()).digest())
-                has_x_bit = os.stat(f.fileno()).st_mode & 0o111 != 0
-                h.update(bytes([has_x_bit]))
+            if os.path.islink(file):
+                # For symlinks, hash the link target instead of following it
+                link_target = os.readlink(file)
+                h.update(link_target.encode('utf-8'))
+            else:
+                with open(file, 'rb') as f:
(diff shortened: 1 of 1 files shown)


dustymabe:
This allows broken symlinks, right? Out of curiosity what was the symlink? and the target was `/etc/grub.d/15_ostree` ?

build-rootfs
```diff
@@ -467,12 +467,19 @@ def calculate_inputhash(rootfs, overlays, manifest):
...
-                # h.update(hashlib.file_digest(f, 'sha256').digest())
-                h.update(hashlib.sha256(f.read()).digest())
-                has_x_bit = os.stat(f.fileno()).st_mode & 0o111 != 0
-                h.update(bytes([has_x_bit]))
+            if os.path.islink(file):
+                # For symlinks, hash the link target instead of following it
+                link_target = os.readlink(file)
+                h.update(link_target.encode('utf-8'))
+                # Symlinks don't have meaningful execute bits
+                h.update(bytes([False]))
```
dustymabe: I would just drop this `h.update` since IIUC it won't really do anything (we're hashing a constant value).

alternatively just leave the `has_x_bit` and `h.update()` below outside the if/else and let it always run even though "Symlinks don't have meaningful execute bits".

dustymabe: ahh. can't really leave `has_x_bit` outside the if/else because it's under the `with open()`, so nvm on that suggestion.

cgwalters: Thanks, FTR that was a little bit of AI slop I missed myself in prereview.

cgwalters:
> This allows broken symlinks, right? Out of curiosity what was the symlink? and the target was /etc/etc/grub.d/15_ostree ?

Yes, not just broken but in this case absolute links which we'd need to containerize or use openat2(RESOLVE_IN_ROOT) to handle.

(There's a further problem here in that ostree has been installing the legacy grub logic even when doing CI on systems that don't need it like FCOS)

dustymabe (approved):
LGTM

PR #3880, opened 2025-11-12
image-base: use -Efragments for erofs creation

### LINKED ISSUE #1991: Switch EROFS live generation from `-Eall-fragments` to `-Efragments`
### Describe the enhancement

Since the latest erofs-utils 1.8.9~10 already supports multi-threaded `-Efragments` option (IOWs, the build time is much improved), it's preferable to use `-Efragments` instead of `-Eall-fragments,fragdedupe=inode` to maintain better runtime performance.

If it's possible, please consider switching to `-Efragments` instead.

### System details

_No response_

### Additional information

_No response_

dustymabe:
As suggested by the maintainer we'll change this option from `-Eall-fragments,fragdedupe=inode` to -Efragments.

Closes https://github.com/coreos/fedora-coreos-tracker/issues/1991

### DIFF
--- a/image-base.yaml
+++ b/image-base.yaml
@@ -39,6 +39,6 @@ container-imgref: "ostree-image-signed:docker://quay.io/fedora/fedora-coreos:{st
 
 # Enable 'erofs' by default for the rootfs in the Live ISO/PXE artifacts
 live-rootfs-fstype: "erofs"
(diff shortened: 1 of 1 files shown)


dustymabe:
The difference here in size is minimal but is an improvement:


From:

```
-rw-r--r--. 1 builder builder 791M Nov 12 13:56 root.erofs
```

To:

```
-rw-r--r--. 1 builder builder 789M Nov 12 13:56 root.erofs
```

marmijo (approved):
LGTM

PR #3881, opened 2025-11-12
build-args: pin on a specific bootc builder image

joelcapitao:
We are about to enable bump-lockfile job for `rawhide` branch which bumps bootc builder image as well.
The job expects the image format to be defined by digest.


jbtrystram (approved):

dustymabe (approved):
LGTM

PR #3884, opened 2025-11-13
versionary: pull Y from the arch-matching lockfile

jbtrystram:
Versionary pull the Y component of the version ID from the lockfile timestamp (if it exists).
Instead of always pulling from the x86 manifest, open the one lockfile matching the machine architecture.
This will fix a corner case of the `bump-lockfile` job failing when creating the lockfile for the first time: on the remote multi-arch builders, the x86 lockfile does not exist yet
(it will be there in the next commit.)

### DIFF
--- a/versionary
+++ b/versionary
@@ -13,6 +13,7 @@
 import argparse
 import json
 import os
+import platform
 import re
 import subprocess
 import sys
@@ -84,9 +85,10 @@ def get_y(manifest):
     # XXX: should sanity check that the lockfiles for all the basearches have
     # matching timestamps
     exts = ['json', 'yaml']
+    basearch = platform.machine()
     for ext in exts:
         try:
-            with open(f"src/config/manifest-lock.x86_64.{ext}") as f:
+            with open(f"src/config/manifest-lock.{basearch}.{ext}") as f:
                 lockfile = yaml.safe_load(f)
                 generated = lockfile.get('metadata', {}).get('generated')
                 if not generated:


joelcapitao:
Does it mean that currently we are taking the x86 manifest of the previous commit ?

jbtrystram:
> Does it mean that currently we are taking the x86 manifest of the previous commit ?

Yes, in the multi-arch builders. 

The pipeline runs `cosa fetch --update-lockfile` then do a build. The build process call `versionary` to craft the build ID, which pull the date from the lockfile.
The same happen in the multi-arch builders, but they are reusing the x86 lockfile that was already there. 
In the end the version ID from the "main" x86 build is used so it does not matter, which is why this have not been an issue. 

In the [last pipeline run](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/bump-lockfile/detail/bump-lockfile/239) if you look at the logs : 

for x86: 
```
[2025-11-13T05:17:14.265Z] x: 43 (from manifest)
[2025-11-13T05:17:14.265Z] y: 20251113 (from lockfile)
[2025-11-13T05:17:14.265Z] z: dev (overridden)
[2025-11-13T05:17:14.265Z] n: 0 (previous version 43.20251112.20.0 x.y.z does not match)
[2025-11-13T05:17:14.265Z] Building with container runtime (buildah) with VERSION=43.20251113.dev.0...
```

for s390x: 
```
[2025-11-13T05:17:20.196Z] x: 43 (from manifest)
[2025-11-13T05:17:20.196Z] y: 20251112 (from lockfile)
[2025-11-13T05:17:20.196Z] z: dev (overridden)
[2025-11-13T05:17:20.196Z] n: 0 (previous version 43.20251112.20.0 x.y.z does not match)
[2025-11-13T05:17:20.196Z] Building with container runtime (buildah) with VERSION=43.20251112.dev.0...
```

joelcapitao:
Alright, then we can ninja merge this one as it won't be tested in our use case.
If it breaks something, we'll revert it and initialize a dummy manifest with minimal metadata to unblock.

PR #3885, opened 2025-11-13
manifests: Replace ssh-key-dir by openssh config

travier:
We can now use the extended support for globs in `AuthorizedKeysFile` to tell sshd to look at the files written by Ignition and Afterburn for ssh key to accept.

See: https://github.com/coreos/ssh-key-dir/issues/188
See: https://man.openbsd.org/sshd_config#AuthorizedKeysFile

### DIFF
--- a/overlay.d/18sshd-authorized-keys-file/etc/ssh/sshd_config.d/40-authorized-keys-file.conf
+++ b/overlay.d/18sshd-authorized-keys-file/etc/ssh/sshd_config.d/40-authorized-keys-file.conf
@@ -1,0 +1,2 @@
+# Also accept keys configured by Ignition and Afterburn
+AuthorizedKeysFile=.ssh/authorized_keys .ssh/authorized_keys.d/*
--- a/manifest-lock.aarch64.json
+++ b/manifest-lock.aarch64.json
@@ -1203,9 +1203,6 @@
     "squashfs-tools": {
       "evra": "4.6.1-7.fc43.aarch64"
     },
-    "ssh-key-dir": {
-      "evra": "0.1.5-3.fc43.aarch64"
-    },
     "sssd-ad": {
       "evra": "2.11.1-4.fc43.aarch64"
     },
--- a/manifest-lock.ppc64le.json
+++ b/manifest-lock.ppc64le.json
@@ -1194,9 +1194,6 @@
     "squashfs-tools": {
       "evra": "4.6.1-7.fc43.ppc64le"
     },
-    "ssh-key-dir": {
-      "evra": "0.1.5-3.fc43.ppc64le"
-    },
     "sssd-ad": {
       "evra": "2.11.1-4.fc43.ppc64le"
     },
--- a/manifest-lock.s390x.json
+++ b/manifest-lock.s390x.json
@@ -1137,9 +1137,6 @@
     "squashfs-tools": {
       "evra": "4.6.1-7.fc43.s390x"
     },
-    "ssh-key-dir": {
-      "evra": "0.1.5-3.fc43.s390x"
-    },
     "sssd-ad": {
       "evra": "2.11.1-4.fc43.s390x"
     },
--- a/manifest-lock.x86_64.json
+++ b/manifest-lock.x86_64.json
@@ -1209,9 +1209,6 @@
     "squashfs-tools": {
       "evra": "4.6.1-7.fc43.x86_64"
     },
-    "ssh-key-dir": {
-      "evra": "0.1.5-3.fc43.x86_64"
-    },
     "sssd-ad": {
       "evra": "2.11.1-4.fc43.x86_64"
     },
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -43,6 +43,7 @@ conditional-include:
 ostree-layers:
   - overlay/15fcos
   - overlay/17fcos-container-signing
+  - overlay/18sshd-authorized-keys-file
 
 packages:
   - fedora-release-coreos
--- a/manifests/system-configuration.yaml
+++ b/manifests/system-configuration.yaml
@@ -33,8 +33,6 @@ packages:
   # Boost starving threads
   # https://github.com/coreos/fedora-coreos-tracker/issues/753
   - stalld
-  # Ignition aware SSH key management
-  - ssh-key-dir
 
 postprocess:
   # Mask systemd-repart. Ignition is responsible for partition setup on first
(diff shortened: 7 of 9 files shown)


travier:
Konlux builds are failing on repo fetch:
```
[build] error: Installing packages: Updating rpm-md repo 'fedora-coreos-pool': cannot update repo 'fedora-coreos-pool': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Status code: 503 for https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/latest/x86_64/repodata/repomd.xml (IP: 38.145.32.20)
[build] error: Executing compose install: ExitStatus(unix_wait_status(256))
[build] Error executing command: Command '['rpm-ostree', 'compose', 'rootfs', '--ostree-repo=/var/tmp/tmpii7fj2em', '--lockfile=/tmp/tmp071ntcw5.json', '--cachedir=/cache', '--source-root=/', '/tmp/tmphwqe1u60.json', '/target-rootfs']' returned non-zero exit status 1.
```

travier:
Ah, maybe because I did not remove the entries for the package?

joelcapitao (dismissed):
LGTM

dustymabe:
I'm wondering if we should create that `rhel-9.6` branch now. Changes like this definitely make me hesistant.

dustymabe:
> I'm wondering if we should create that `rhel-9.6` branch now. Changes like this definitely make me hesistant.

I guess, though. That if this was breaking things pretty much all tests would fail?

overlay.d/18sshd-authorized-keys-file/etc/ssh/sshd_config.d/40-authorized-keys-file.conf
dustymabe, line 1: This is in the 15fcos overlay. Don't we need ssh keys set via Ignition to work in RHCOS too?

I'd kind of prefer to have its own overlay for this. Or maybe even dropin files delivered by the RPMs (actually that would be better as we try to go for https://github.com/coreos/fedora-coreos-tracker/issues/1582, right?)

travier:
This indeed needs work to not break RHCOS and I agree that it might be best to move the config drop-in to Ignition (and Afterburn as well?).

travier:
Depends on https://github.com/coreos/rhel-coreos-config/pull/123 where we'll include `ssh-key-dir` in order to not break RHCOS.

travier:
> I'd kind of prefer to have its own overlay for this. Or maybe even dropin files delivered by the RPMs (actually that would be better as we try to go for [coreos/fedora-coreos-tracker#1582](https://github.com/coreos/fedora-coreos-tracker/issues/1582), right?)

So in the end I kept things FCOS only and here for now in order to be able to move forward and deprecate the package sooner rather than later. We can drop the config snippet or move it to an RPM in a followup PR.

dustymabe:
Sounds good to keep it FCOS only for now. 

Optional, but recommended: I'd still prefer if we put it in its own overlay (or manifest with a postprocess `cat <<EOF > file`) to raise awareness of it. A comment in the readme for the overlays or a comment in the manifest (if going the postprocess route) to explain the ultimate goal is to move this config to the RPMs.

overlay.d/18sshd-authorized-keys-file/etc/ssh/sshd_config.d/40-authorized-keys-file.conf
```diff
@@ -0,0 +1,2 @@
+# Also accept keys configured by Ignition and Afterburn
+AuthorizedKeysFile=.ssh/authorized_keys .ssh/authorized_keys.d/*
```
dustymabe, line 2: I'm wondering if we should be explicit:

```suggestion
AuthorizedKeysFile=.ssh/authorized_keys .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/afterburn
```

dustymabe (dismissed):

overlay.d/18sshd-authorized-keys-file/etc/ssh/sshd_config.d/40-authorized-keys-file.conf
dustymabe: though I guess the glob support is what we were waiting on :) 

ignore me

travier: yep :)

dustymabe: tell me if you agree with the statement in https://github.com/coreos/fedora-coreos-tracker/issues/1582#issuecomment-3641893141

travier:
Moved to a distinct overlay. Note that we'll have to figure out how to include this configuration in both Afterburn & Ignition.

dustymabe (dismissed):
LGTM

travier:
Ready now that we merged https://github.com/coreos/rhel-coreos-config/pull/124

travier:
I have not yet tested that however so this relies on our test coverage being good enough.

dustymabe (approved):
/lgtm

travier:
A simple test I did for this one was to do `mv .ssh/authorized_keys.d/ignition .ssh/authorized_keys.d/ignitionfoo` and then make sure I could still log in. not sure that it's worth a test given we should exercise the ssh path a lot already.

dustymabe:
> This indeed needs work to not break RHCOS and I agree that it might be best to move the config drop-in to Ignition (and Afterburn as well?).

This happened in:

- https://src.fedoraproject.org/rpms/ignition/c/3ccce9fe2c0462f4a17923937933ab9c26db0aa5?branch=rawhide       
- https://src.fedoraproject.org/rpms/rust-afterburn/c/e61c6e5f776410ea72e8f731cec88567577f6bb8?branch=rawhide 

and will roll out to F44+. F43 CoreOS will retain `/etc/ssh/sshd_config.d/40-authorized-keys-file.conf` but F44+ will have:

- [91-ignition-authorized-keys-file.conf](https://src.fedoraproject.org/rpms/ignition/blob/3ccce9fe2c0462f4a17923937933ab9c26db0aa5/f/91-ignition-authorized-keys-file.conf)
- [90-afterburn-authorized-keys-file.conf](https://src.fedoraproject.org/rpms/rust-afterburn/blob/e61c6e5f776410ea72e8f731cec88567577f6bb8/f/90-afterburn-authorized-keys-file.conf)

delivered by the RPMs for all consumers in Fedora.

PR #3888, opened 2025-11-13
build-rootfs: handle epoch corner case

dustymabe:
When verifying that the RPMs we locked on are what is in the built container we ran into a corner case where if a RPM has explicitly set their EPOCH to 0 then the NEVR from `rpm -q --qf` will include that `0:`.

`rpm-ostree` just assumes any `0` EPOCH shouldn't be included in the string [1] so it won't get written to the lockfile.

We need to account for this here.

We stumbled upon this by accident because perl was getting included [2] when it shouldn't and it have subpackages that set an EPOCH of `0`.

[1] https://github.com/coreos/rpm-ostree/blob/0ad2ee53f3c2d77bc4adfd618d865e919c943865/src/libpriv/rpmostree-rpm-util.cxx#L76-L84
[2] https://github.com/coreos/fedora-coreos-tracker/issues/2059

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -432,11 +432,21 @@ def inject_content_manifest(target_rootfs, manifest):
 
 
 def verify_strict_mode(rootfs, locked_nevras):
-    rpms = bwrap(rootfs, ['rpm', '-qa', '--qf', '%{NEVRA}\t%{NEVR}\n'], capture=True)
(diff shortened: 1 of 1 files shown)


jbtrystram (approved):
Nice

joelcapitao:
Thank you !

PR #3895, opened 2025-11-14
tests/kola: switch container helpers to Fedora 43

marmijo:
This should have been part of https://github.com/coreos/fedora-coreos-config/commit/51a217a1a22276091a3dfc27a3d4f45b293d7e74 but there was a failure observed in the `ext.config.podman.rootless-systemd` test that wasn't seen using the Fedora 42 container images. There must have been some update with either systemd, selinux-policy, coreos-assembler, or elsewhere that allows httpd to start in a rootless-systemd environment again becasue the test is now passing.


dustymabe (approved):
LGTM

PR #3897, opened 2025-11-18
build-rootfs: fix typo in rpm query format

dustymabe:
This should have been a % instead of a $. I think I somehow copy pasted it wrong.

Fixes 67cc34a

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -432,7 +432,7 @@ def inject_content_manifest(target_rootfs, manifest):
 
 
 def verify_strict_mode(rootfs, locked_nevras):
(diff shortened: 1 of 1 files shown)


dustymabe:
Didn't catch this before because the rpm` command doesn't actually return an error exit-code, just prints a log message to the screen:

```
bash-5.3# rpm    -q    --qf    '${EPOCH}\t%{NVRA}\t%{NVR}\t%{NEVRA}\t%{NEVR}\n' bash
error: incorrect format: unexpected }
bash-5.3# 
bash-5.3# echo $?
0
```

jbtrystram (approved):

PR #3898, opened 2025-11-18
overlay/ignition: delete pre-existing bootuuid.cfg

jbtrystram:
In our disk images, bootupd inject a grub script that looks for a `/boot/grub2/bootuuid.cfg` to find the boot partition. If the file is not present, the label `boot` is used to find the partition. See [1] for more.

During the first boot, the hard-coded uuid of the boot partition is replaced with a random one [2].
However this happen in two phases:
First `coreos-gpt-setup.service` assigns a new UUID to the partition, Then `coreos-boot-edit.service` writes that new uuid to `/boot/grub2/bootuuid.cfg`.
However, when ignition needs to reboot (e.g. for applying a karg), the reboot happens between the two.

In `bootc install to-disk` the UUID of the boot partition is written to `/boot/grub2/bootuuid.cfg`, so it will be present at first boot. But if we restamp the parition then reboot, the original `bootuuid.cfg` is still there, preventing the reboot.

So let's delete this file just before the restamp, so we don't end up in this inconsistent state.
This allows us to use `bootc install to-disk` to build our disk images. See [3][4].

[1] https://github.com/coreos/bootupd/blob/1768f31c85b5d8cd72f4339ab669b703d7071b51/src/grub2/grub-static-pre.cfg
[2] https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.service
[3] https://issues.redhat.com/browse/COS-3374
[4] https://github.com/coreos/coreos-assembler/pull/4224

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.service
@@ -28,3 +28,6 @@ OnFailureJobMode=isolate
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/sbin/coreos-gpt-setup /dev/disk/by-label/boot
+# MountFlags=slave is so the umount of /boot is guaranteed to happen.
+# /boot will only be mounted for the lifetime of the unit.
+MountFlags=slave
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh
@@ -26,6 +26,19 @@ if [ "${PTUUID:-}" != "$UNINITIALIZED_GUID" ]; then
     exit 0
 fi
 
+
+# Mount /boot. Note that we mount /boot but we don't unmount it because we
+# are run in a systemd unit with MountFlags=slave so it is unmounted for us.
+bootmnt=/mnt/boot_partition/
+bootdev=/dev/disk/by-label/boot
+mount -o rw ${bootdev} ${bootmnt}
+# Make sure we delete any pre-existing bootuuid.cfg
+# bootc install stamps it during the disk image creation
+# since we are changing the uuid here we want to remove this file
+# otherwise it will prevent a reboot
+rm -f ${bootmnt}/grub2/bootuuid.cfg
+
+
 echo "Randomizing disk GUID"
 sgdisk --disk-guid=R --move-second-header "$PKNAME"
 udevadm settle || :


travier:
> However, when ignition needs to reboot (e.g. for applying a karg), the reboot happens between the two.

That feels like what we should fix here.

travier:
This also feels related to https://github.com/coreos/fedora-coreos-tracker/issues/2026

jbtrystram:
> > However, when ignition needs to reboot (e.g. for applying a karg), the reboot happens between the two.
> 
> That feels like what we should fix here.

Do you mean writing the boot grub stub at the time of generating the uuid , before the reboot ?
I am afraid that would break if you re-partition the disks, which happen after the eventual reboot. Then you'd have a new UUID, and would need to re-run `rdcore` to re-do the bind boot phase[1].

[1] https://github.com/coreos/coreos-installer/blob/main/src/bin/rdcore/rootmap.rs#L249

Another fix is to ask bootc not to write that file in the first place or write a new `rm` osbuild stage to delete it

jbtrystram:
This whole thing happen because bootc hard-code writing the stamp file here : 

https://github.com/bootc-dev/bootc/blob/49d753f996747a9b1f531abf35ba4e207cf4f020/crates/lib/src/bootloader.rs#L103C37-L103C52

One approach would be to make this flag conditionnal in bootc. That would be another knob, though. 
Some details here https://github.com/coreos/coreos-assembler/pull/4224#discussion_r2247146403

jbtrystram:
Ok, diving deeper into this with the help of @joelcapitao and Opus 4.5 we came up with a chart for the boot sequence: 

```mermaid
flowchart TD
    subgraph "Phase 1: Early Boot - Device Setup"
        A[dev-disk-by-label-boot.device] --> B[coreos-unique-boot.service]
        B --> C[coreos-gpt-setup.service<br/><b>Randomize GPT disk UUID</b><br/>★ PR adds: delete bootuuid.cfg]
        C --> D[ignition-ostree-uuid-boot.service<br/><b>Restamp boot fs UUID</b>]
        C --> E[transposefs-save.service<br/><i>Save partitions to zram</i>]
    end

    subgraph "Phase 2: Ignition Kargs"
        D --> F[ignition-kargs.service<br/><i>Apply kargs from Ignition config</i>]
        E --> F
    end

    subgraph "Reboot Decision"
        F --> G{kargs changed?}
        G -->|Yes| H[coreos-kargs-reboot.service<br/><b>REBOOT</b>]
        H -.->|Second boot| A
        G -->|No| I[ignition-disks.service]
    end

    subgraph "Phase 3: Ignition Disk/File Stages"
        I --> J[transposefs-restore.service<br/><i>Restore from zram</i>]
        J --> K[ignition-files.service]
    end

    subgraph "Phase 4: Late Boot - Finalization"
        K --> L[coreos-boot-edit.service<br/><b>rdcore bind-boot</b><br/><i>Write final bootuuid.cfg</i>]
        L --> M[initrd.target]
    end

    style C fill:#ff9,stroke:#333,stroke-width:2px
    style H fill:#f99,stroke:#333,stroke-width:2px
    style L fill:#9f9,stroke:#333,stroke-width:2px
```

So the tension here that prevents writing the `/boot/.bootuuid.cfg` file at the same time of the re-stamping is `ignition-disks` : writing the stamp is useless if the partition will be re-formated anyway. So we skip writing the stamp altogether, that avoids mounting the partition. 

So I propose the following : 

1. Only delete the stamp if it's content equals the `UNINITIALIZED_GUID` value before the restamping.
2. Write the stamp after the restamping, even if it gets re-done afterwards.
 
Boths approaches requires mounting the boot partition anyway. But 2 is slightly more complicated.

jlebon:
> One approach would be to make this flag conditionnal in bootc. That would be another knob, though. Some details here [coreos/coreos-assembler#4224 (comment)](https://github.com/coreos/coreos-assembler/pull/4224#discussion_r2247146403)

Hmm yeah, that's weird. bootc shouldn't write that file since it's writing out the sentinel UUID. bootfs binding is purely something that should happen at runtime with the final UUIDs you'll run with.

dustymabe:
> > One approach would be to make this flag conditionnal in bootc. That would be another knob, though. Some details here [coreos/coreos-assembler#4224 (comment)](https://github.com/coreos/coreos-assembler/pull/4224#discussion_r2247146403)
> 
> Hmm yeah, that's weird. bootc shouldn't write that file since it's writing out the sentinel UUID. bootfs binding is purely something that should happen at runtime with the final UUIDs you'll run with.

i.e. The code should have some sort of "I'm building a golden image" option and not add `--write-uuid` to the `bootupctl backend install` command in [this code](https://github.com/bootc-dev/bootc/blob/1b898758c94e25b50d08283841ad188999b26aaa/crates/lib/src/bootloader.rs#L107-L111)

Is that what you are saying @jlebon ?

jbtrystram:
> Is that what you are saying @jlebon ?

That's what I understand as well

jlebon:
> > > One approach would be to make this flag conditionnal in bootc. That would be another knob, though. Some details here [coreos/coreos-assembler#4224 (comment)](https://github.com/coreos/coreos-assembler/pull/4224#discussion_r2247146403)
> > 
> > 
> > Hmm yeah, that's weird. bootc shouldn't write that file since it's writing out the sentinel UUID. bootfs binding is purely something that should happen at runtime with the final UUIDs you'll run with.
> 
> i.e. The code should have some sort of "I'm building a golden image" option and not add `--write-uuid` to the `bootupctl backend install` command in [this code](https://github.com/bootc-dev/bootc/blob/1b898758c94e25b50d08283841ad188999b26aaa/crates/lib/src/bootloader.rs#L107-L111)
> 
> Is that what you are saying @jlebon ?

Yes, exactly. Or... perhaps this should be tied to the existing `--boot-mount-spec=""` logic since that's also a signal that says "don't worry about the bootfs, we own that logic".

jbtrystram:
Closing this as i fixed it in bootc

PR #3899, opened 2025-11-18
Fix multipath.custom-partition test fcos version

tlbueno:
Set the fcos version to 1.5.0 to be able to run the test on a older version of FCOS/RHCOS like rhcos-4.18.


dustymabe (approved):
LGTM

PR #3905, opened 2025-11-21
ci/overrides.py: update rpms.lock.yaml during fast-track operations

Roshan-R:
Extend the fast-track workflow to also update rpms.lock.yaml, which is used for hermetic builds in Konflux. Previously, only manifest-lock.overrides.json was updated, causing rpms.lock.yaml to fall out of sync with fast-tracked package versions.

This change ensures that fast-tracked overrides are consistently applied across both lockfile types.

### DIFF
--- a/ci/overrides.py
+++ b/ci/overrides.py
@@ -115,6 +115,42 @@ def do_fast_track(args):
     for lockfile_path in get_lockfiles():
         merge_overrides(lockfile_path, overrides)
 
+    konflux_rpm_lock(overrides)
+
+
+def konflux_rpm_lock(overrides: dict):
+    # Modify the rpms.lock.yaml for override
+    lockfile_path = get_rpm_lockfile()
+    with open(lockfile_path) as f:
+        lockfile = yaml.safe_load(f)
+    for pkg_name, v in overrides.items():
+        is_noarch = v.get("evra") is not None
+        ver = v.get("evr") or v.get("evra")
+        nvr = f'{pkg_name}-{ver}'
+        update_rpm_lock_for_nvr(nvr, pkg_name, lockfile, is_noarch)
+    with open(lockfile_path, 'w') as f:
+        yaml.dump(lockfile, f)
+
+
+def update_rpm_lock_for_nvr(nvr, pkg_name, manifest, is_noarch):
+    nevra_form = libdnf5.rpm.VectorNevraForm(1, libdnf5.rpm.Nevra.Form_NEVRA)
+    updated = False
+    # Iterate over all the arches
+    for arch_block in manifest["arches"]:
+        arch = arch_block.get("arch")
+        packages = arch_block.get("packages")
+        # Check if the nvr already exists in the konflux rpm lock or not
+        for package_entry in packages:
+            package_nvr = package_entry["url"].split("/")[-1]
+            parsed_vector = libdnf5.rpm.Nevra.parse(package_nvr, nevra_form)
+            parsed = parsed_vector.pop()
+            # Update the nvr with the new url if a match is found
+            if parsed.get_name() == pkg_name:
+                package_entry["url"] = build_coreos_pool_url(nvr, arch, is_noarch)
+                updated = True
+    if not updated:
+        raise Exception(f"Cannot find specified package {nvr} in the konflux rpm lock")
+
 
 def do_pin(args):
     overrides = {}
@@ -307,6 +343,22 @@ def get_lockfiles():
     # lockfiles += [f'manifest-lock.overrides.{arch}.yaml' for arch in ARCHES]
     return [os.path.join(basedir, f) for f in lockfiles]
 
+
+def get_rpm_lockfile():
+    lockfile = 'rpms.lock.yaml'
+    return os.path.join(basedir, lockfile)
+
+
+def build_coreos_pool_url(nvr, arch, is_noarch):
+    """
+    nvr: e.g. 'audit-4.1.2-2.fc43'
+    """
+    first_letter = nvr[0].lower()
+    base = "https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/latest"
+    if is_noarch:
+        return f"{base}/{arch}/Packages/{first_letter}/{nvr}.rpm"
+    else:
+        return f"{base}/{arch}/Packages/{first_letter}/{nvr}.{arch}.rpm"
 
 def graduate_lockfile(base, fn):
     if not os.path.exists(fn):


ci/overrides.py
```diff
@@ -115,6 +115,35 @@ def do_fast_track(args):
...
+    write_lockfile(lockfile_path, lockfile)
+
+
+def update_rpm_lock_for_nvr(nvr, pkg_name, manifest):
+    nevra_form = libdnf5.rpm.VectorNevraForm(1, libdnf5.rpm.Nevra.Form_NEVRA)
+    # Iterate over all the arches
+    for arch_block in manifest["arches"]:
+        arch = arch_block.get("arch")
+        packages = arch_block.get("packages")
+        # Check if the nvr already exists in the manifest or not
```
jbtrystram: it's the  konflux rpm lock, not the manifest. I know i'm nitpicking but this is already hard to follow so let's make the comment crystal clear

ci/overrides.py
```diff
@@ -115,6 +115,35 @@ def do_fast_track(args):
...
+        for package_block in packages:
```
jbtrystram:
```suggestion
        for package_entry in packages:
        # each entry have 2 fields : `url` and `repoid`. We leave `repoid` untouched.
```

ci/overrides.py
```diff
@@ -115,6 +115,35 @@ def do_fast_track(args):
...
+            package_nvr = package_block["url"].split("/")[-1]
+            parsed = libdnf5.rpm.Nevra.parse(package_nvr, nevra_form)[0]
+            # Update the nvr with the new url if a match is found
+            if parsed.get_name() == pkg_name:
+                package_block["url"] = build_coreos_pool_url(nvr, arch)
```
jbtrystram: At the end of the loop we should yield an error if the lock entry was not found. I know it's very unlikely but if that were to happen let's give a clue rather than silently not updating the lockfile.

ci/overrides.py
```diff
@@ -115,6 +115,35 @@ def do_fast_track(args):
...
+    for pkg_name, v in overrides.items():
+        ver = v.get("evr") or v.get("evra")
+        nvr = f'{pkg_name}-{ver}'
+        update_rpm_lock_for_nvr(nvr, pkg_name, lockfile)
+    write_lockfile(lockfile_path, lockfile)
+
+
+def update_rpm_lock_for_nvr(nvr, pkg_name, manifest):
+    nevra_form = libdnf5.rpm.VectorNevraForm(1, libdnf5.rpm.Nevra.Form_NEVRA)
+    # Iterate over all the arches
```
jbtrystram, line 138: aren't the overrides already sorted by arches ?

ci/overrides.py
```diff
@@ -115,6 +115,35 @@ def do_fast_track(args):
     for lockfile_path in get_lockfiles():
         merge_overrides(lockfile_path, overrides)
 
+    konflux_rpm_lock(overrides)
+
+
+def konflux_rpm_lock(overrides: dict):
+    # Modify the rpms.lock.yaml for override
+    lockfile_path = get_rpm_lockfile()
+    with open(lockfile_path) as fh:
```
jbtrystram: `fh` ?

jbtrystram (changes requested):
Looks generally sane to me, not tested. A few suggestions

ci/overrides.py
Roshan-R: used to denote a filehandle

Roshan-R: I guess we could change it to the standard `f` variable

ci/overrides.py
Roshan-R: I have updated the logic to raise an exception

ci/overrides.py
```diff
@@ -115,6 +115,40 @@ def do_fast_track(args):
...
+def konflux_rpm_lock(overrides: dict):
+    # Modify the rpms.lock.yaml for override
+    lockfile_path = get_rpm_lockfile()
+    with open(lockfile_path) as f:
+        lockfile = yaml.safe_load(f)
+    for pkg_name, v in overrides.items():
+        ver = v.get("evr") or v.get("evra")
+        nvr = f'{pkg_name}-{ver}'
+        update_rpm_lock_for_nvr(nvr, pkg_name, lockfile)
+    write_lockfile(lockfile_path, lockfile)
```
joelcapitao: Using `write_lockfile` adds the OVERRIDE_HEADER at the beginning of the `rpms.lock.yaml` file but the content is misleading for that file. So I think we should dump the YAML content as we loaded (plus the modification ofc)

ci/overrides.py
joelcapitao: Yes, but it does not change the processing. I might miss your point

ci/overrides.py
```diff
@@ -115,6 +115,40 @@ def do_fast_track(args):
...
+    konflux_rpm_lock(overrides)
+
+
+def konflux_rpm_lock(overrides: dict):
+    # Modify the rpms.lock.yaml for override
+    lockfile_path = get_rpm_lockfile()
+    with open(lockfile_path) as f:
+        lockfile = yaml.safe_load(f)
+    for pkg_name, v in overrides.items():
+        ver = v.get("evr") or v.get("evra")
```
joelcapitao, line 128: Note that `evr` is for archful packages, whereas `evra` is for noarch ones.
Actually, I tested locally by specifying a `evra` and it's overriding the `noarch` by the `arch` value.

See:

```
- url: https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/latest/ppc64le/Packages/b/bash-completion-2.16-2.fc43.noarch.rpm
+ url: https://kojipkgs.fedoraproject.org/repos-dist/coreos-pool/latest/ppc64le/Packages/b/bash-completion-2.16-3.fc43.noarch.ppc64le.rpm
```

I tested with 
```python
overrides = {'bash-completion': {'evra': '2.16-3.fc43.noarch'}}
konflux_rpm_lock(overrides)
```

c.f a overrides commit with an `evra` https://github.com/coreos/fedora-coreos-config/commit/616deaa7aa2bba96b5dd9c1656802f1baa7ca49e

joelcapitao (changes requested):
Thank you for working on this feature and looks good overall 👍 
I left some comments but the main one is that we are missing the support of noarch packages (i.e the `evra` in this context). Could you add it please ?

I think we would need to change `build_coreos_pool_url` function with something like:
```python
if is_noarch:
    return f"{base}/{arch}/Packages/{first_letter}/{nvr}.rpm"
else:
    return f"{base}/{arch}/Packages/{first_letter}/{nvr}.{arch}.rpm"
```
but I let you find better implementation

ci/overrides.py
joelcapitao: Fixed in last commit

joelcapitao (approved):
LGTM thank you !

ci/overrides.py
jbtrystram: That was a misunderstanding from me. Sorry

jbtrystram (approved):
Thanks!

PR #3908, opened 2025-11-24
overrides: unpin less rpm

aaradhak:
We can now unpin this less rpm as there is a bodhi update available
- https://bodhi.fedoraproject.org/updates/FEDORA-2025-412e51ae6a,

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/2059


dustymabe (approved):
LGTM

PR #3912, opened 2025-11-25
ci/promotion-diff: ignore STREAM build args diff

joelcapitao:
The `STREAM` line in build-args.conf is stream-specific and expected to differ between branches.


dustymabe (approved):
LGTM

tlbueno (approved):
LGTM

PR #3913, opened 2025-11-25
build-rootfs: handle syntax warning

dustymabe:
/src/build-rootfs:446: SyntaxWarning: "is" with 'str' literal. Did you mean "=="?


tlbueno (approved):
LGTM

PR #3916, opened 2025-12-03
Add new rhel-9.8 variant

ravanelli:
- Reflect the new variant created adding it here as well.

### DIFF
--- a/manifests/shared-el.yaml
+++ b/manifests/shared-el.yaml
@@ -2,7 +2,7 @@ conditional-include:
   # Include shared-el9 workarounds on Fedora and EL9
   - if: id == "fedora"
     include: shared-el9.yaml
-  - if: osversion == "rhel-9.6"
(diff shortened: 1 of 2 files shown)


travier (dismissed):

jbtrystram (dismissed):

manifests/shared-el.yaml
```diff
@@ -2,6 +2,8 @@ conditional-include:
   # Include shared-el9 workarounds on Fedora and EL9
   - if: id == "fedora"
     include: shared-el9.yaml
+  - if: osversion == "rhel-9.8"
+    include: shared-el9.yaml
```
dustymabe, line 6: If we branch `rhel-9.6` before we merge this then we'd just replace `9.6` with `9.8` here.

dustymabe (approved):
LGTM

travier (approved):

PR #3919, opened 2025-12-09
versionary: add flag to compute iteration value from git history

Roshan-R:
The new method obtains the commit hash that last modified rpms.lock.yaml and counts the number of commits since that hash. This allows the iteration value to be self-contained, eliminating the need to fetch external resources like builds.json for historical data, and removes any network dependency during the build process.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2015

### DIFF
--- a/versionary
+++ b/versionary
@@ -50,8 +50,13 @@ def main():
     assert os.path.isdir('builds'), 'Missing builds/ dir'
 
     config = dotenv.dotenv_values('src/config/build-args.conf')
-    x, y, z = (get_x(config), get_y(config), get_z(config, args.dev))
-    n = get_next_iteration(x, y, z)
+    x, z = (get_x(config), get_z(config, args.dev))
+    y, y_as_timestamp = get_y(config)
+    # Default behavior generates the iteration from Git (dev mode uses builds)
+    if args.dev:
+        n = get_next_iteration_from_builds(x, y, z)
+    else:
+        n = get_next_iteration_from_git(y_as_timestamp)
     new_version = f'{x}.{y}.{z}.{n}'
 
     # sanity check the new version by trying to re-parse it
@@ -62,7 +67,9 @@ def main():
 def parse_args():
     parser = argparse.ArgumentParser()
     parser.add_argument('--workdir', help="path to cosa workdir")
-    parser.add_argument('--dev', action='store_true', help="generate a developer version")
+    parser.add_argument(
+        "--dev", action="store_true", help="generate a developer version"
+    )
     return parser.parse_args()
 
 
@@ -109,7 +116,7 @@ def get_y(config):
 
     ymd = dt.strftime('%Y%m%d')
     eprint(f"y: {ymd} ({msg_src})")
-    return int(ymd)
+    return int(ymd), str(dt)
 
 
 def get_z(config, dev):
@@ -126,7 +133,7 @@ def get_z(config, dev):
     return mapped
 
 
-def get_next_iteration(x, y, z):
+def get_next_iteration_from_builds(x, y, z):
     try:
         with open('builds/builds.json') as f:
             builds = json.load(f)
@@ -150,6 +157,36 @@ def get_next_iteration(x, y, z):
     n = last_version[3] + 1
     eprint(f"n: {n} (incremented from previous version {last_buildid})")
     return n
+
+
+def get_next_iteration_from_git(y_as_timestamp):
+    """
+    Compute the next iteration number based on git commit history.
+
+    Given the Y component of the version (YYYY-MM-DD HH:MM:SS date),
+    this counts all commits from the start of that date up to HEAD.
+    This guarantees that multiple builds on the same day each receive a
+    unique `.n` value, even if several changes occur.
+
+    See: https://github.com/coreos/fedora-coreos-tracker/issues/2015
+    """
+    try:
+        # Count commits after that point
+        commit_count_since_change = subprocess.check_output(
+            ['git', 'rev-list', '--count', '--after', y_as_timestamp, 'HEAD'],
+            cwd="src/config", text=True
+        ).strip()
+        print(
+            f"n: {commit_count_since_change} "
+            "(calculated using git commit history)"
+        )
+        return commit_count_since_change
+    except subprocess.CalledProcessError as err:
+        msg = (
+            "Git command failed: unable to determine the next "
+            f"iteration value ({err})"
+        )
+        raise RuntimeError(msg) from err
 
 
 def parse_version(version):


joelcapitao (changes requested):
We still need to handle the case where `rpms.lock.yaml` does not exist.
I see two options:

1. explicit: we add a flag e.g: `--konflux` or `--commit-based` to compute based on git-history, otherwise we stick to fetching builds.json
2. implicit: we look for a specific file e.g: `rpms.lock.yaml` to compute based on git-history

My personal preference is 1. as it's more flexible and explicit (better for maintenance/debugging)

Roshan-R:
> We still need to handle the case where `rpms.lock.yaml` does not exist. I see two options:
> 
>     1. explicit: we add a flag e.g: `--konflux` or `--commit-based` to compute based on git-history, otherwise we stick to fetching builds.json
> 
>     2. implicit: we look for a specific file e.g: `rpms.lock.yaml` to compute based on git-history
> 
> 
> My personal preference is 1. as it's more flexible and explicit (better for maintenance/debugging)

I agree, option 1 seems like the better approach. It’s more flexible and explicit, which will make the behavior clearer and easier to maintain or debug in the long run.

Roshan-R:
@joelcapitao  I’ve implemented the first option using a command-line flag named `--commit-based`, but I’m open to suggestions if there’s a better name for it.

joelcapitao:
> @joelcapitao I’ve implemented the first option using a command-line flag named `--commit-based`, but I’m open to suggestions if there’s a better name for it.

I don't have strong opinion on the option name, and wording is hard as we all know.
I tested it locally it looks good :+1:

joelcapitao (dismissed):

versionary
```diff
@@ -11,8 +11,8 @@
 '''
 
 import argparse
-import json
 import os
+import json
```
jbtrystram, line 15: why changing the import order ? It's no longer alphabetic and will probably fail flake8 linting

versionary
```diff
@@ -151,6 +164,36 @@ def get_next_iteration(x, y, z):
     return n
 
 
+def get_next_iteration_from_git(x, y, z):
+    """
+    Compute the next iteration number from git commit history.
+
+    Determines `n` by counting commits since the last modification of the
+    lockfile rpms.lock.yaml.
+    """
```
jbtrystram, line 172: let's add a link to the issue and explain why we want to do that

jbtrystram (changes requested):

versionary
Roshan-R: added

dustymabe:
/hold

reviewing now

dustymabe:
I'm trying to think of a case where we could end up attempting to re-use a build ID.

For example if `rpms.lock.yaml` got updated twice in one day (for whatever reason). we could end up with two `.0` builds for that day couldn't we?


I think we need to determine `.n` based on the number of commits since the date corresponding to the Y version.

psuedo code:

```
- given Y
- Find the first commit on day specified by Y (`FIRST`)
- `n` is number of commits between `FIRST` and `HEAD`
```

since x and z are unused in this determination they wouldn't be arguments to the function. Our signature would be something like:

```
def get_next_iteration_from_git(y):
```

Or we could rename it:

```
def get_number_of_commits_since_beginning_of_date(date):
```

but that's probably a bit too verbose.

dustymabe:
additionally. I'm not 100% sure we need to add a new option here. We could just default to the new `get_next_iteration_from_git` and then use `get_next_iteration_from_builds` for local builds (i.e. where users aren't likely to have made new commits between iterations). This could just be triggered based on the already existing `--dev`.

```
if args.dev:
    n = get_next_iteration_from_builds
else
    n = get_next_iteration_from_git
```

dustymabe:
note i'm also touching `versionary` over in https://github.com/coreos/fedora-coreos-config/pull/3921

depending on if this lands first or that lands first I can rebase the changes here on top so you don't have to.

Roshan-R:
I’ve updated the function using the approach @dustymabe suggested. I had initially made it the default, then switched to a command-line flag. Should we make it the default again, or keep it toggleable?

versionary
```diff
@@ -49,8 +49,12 @@ def main():
     assert os.path.isdir('builds'), 'Missing builds/ dir'
 
     manifest = get_flattened_manifest()
-    x, y, z = (get_x(manifest), get_y(manifest), get_z(manifest, args.dev))
-    n = get_next_iteration(x, y, z)
+    x, z = (get_x(manifest), get_z(manifest, args.dev))
+    y, y_as_timestamp = get_y(manifest)
```
joelcapitao: I'd keep returning only 1 i.e the `int` and then in l.55:
`get_next_iteration_from_git(str(y_as_timestamp))`.
It's easier to parse.
It would make sense to return 2 values if more complex operation were done, but it's just converting an integer to a string.

dustymabe:
> I’ve updated the function using the approach @dustymabe suggested.

Code LGTM

> I had initially made it the default, then switched to a command-line flag. Should we make it the default again, or keep it toggleable?

@jbtrystram @joelcapitao - WDYT? Should we make it the default and then the old behavior can get triggered for `--dev`?

joelcapitao:
> > I’ve updated the function using the approach @dustymabe suggested.
> 
> Code LGTM
> 
> > I had initially made it the default, then switched to a command-line flag. Should we make it the default again, or keep it toggleable?
> 
> @jbtrystram @joelcapitao - WDYT? Should we make it the default and then the old behavior can get triggered for `--dev`?

Yeah, I'm ok with that proposal and avoid adding new option :+1: 
I'm testing the code locally.

joelcapitao:
Posting the logs here because there is an ongoing issue with log retention in Konflux Fedora cluster.
```
STEP-PREPARE-BUILD-CONTEXT
  x: 43 (from manifest)
  z: 20 (mapped from stream testing-devel)
  y: 20251208 (from lockfile)
  fatal: detected dubious ownership in repository at '/var/workdir/source'
  To add an exception for this directory, call:
  
  	git config --global --add safe.directory /var/workdir/source
    File "/var/workdir/source/versionary", line 174, in get_next_iteration_from_git
      commit_count_since_change = subprocess.check_output(
                                  ~~~~~~~~~~~~~~~~~~~~~~~^
          ['git', 'rev-list', '--count', '--after', y_as_timestamp, 'HEAD'],
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          cwd="src/config", text=True
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      ).strip()
      ^
    File "/usr/lib64/python3.14/subprocess.py", line 472, in check_output
      return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
             ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 **kwargs).stdout
                 ^^^^^^^^^
    File "/usr/lib64/python3.14/subprocess.py", line 577, in run
      raise CalledProcessError(retcode, process.args,
                               output=stdout, stderr=stderr)
  subprocess.CalledProcessError: Command '['git', 'rev-list', '--count', '--after', '2025-12-08 00:00:00', 'HEAD']' returned non-zero exit status 128.
  
  The above exception was the direct cause of the following exception:
  
    File "/var/workdir/source/versionary", line 214, in <module>
      sys.exit(main())
               ~~~~^^
    File "/var/workdir/source/versionary", line 57, in main
      else get_next_iteration_from_git(y_as_timestamp)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
    File "/var/workdir/source/versionary", line 188, in get_next_iteration_from_git
      raise RuntimeError(msg) from err
  RuntimeError: Git command failed: unable to determine the next iteration value (Command '['git', 'rev-list', '--count', '--after', '2025-12-08 00:00:00', 'HEAD']' returned non-zero exit status 128.)
  Yielding label version=RuntimeError: Git command failed: unable to determine the next iteration value (Command '['git', 'rev-list', '--count', '--after', '2025-12-08 00:00:00', 'HEAD']' returned non-zero exit status 128.)
  Yielding label fedora-coreos.stream=testing-devel
  Created the following labels:
```
I guess it's because of `fatal: detected dubious ownership in repository at '/var/workdir/source'`

We'd need to add `git config --global --add safe.directory /var/workdir/source` in https://gitlab.com/fedora/bootc/tekton-catalog/-/blob/main/tasks/prepare-build-context/prepare-build-context.yaml, build the task and pipeline bundles, and rerun the job with the new digest pipeline.

versionary
```diff
@@ -49,8 +49,13 @@ def main():
...
+    # Default behavior generates the iteration from Git (dev mode uses builds)
+    n = (
+        get_next_iteration_from_builds(x, y, z) if args.dev
+        else get_next_iteration_from_git(y_as_timestamp)
+    )
```
dustymabe: This is purely a preference thing, but after reading code for many years

```
if condition:
   x = y
else:
   x = z
```

parses easier than what we have here and in this case it's actually the same number of lines.

Completely optional

dustymabe (dismissed):
LGTM

dustymabe:
> We'd need to add git config --global --add safe.directory /var/workdir/source

We're still blocked on ^^ here?

versionary
Roshan-R: I agree, a normal if else block is much more readable. This style is something that I picked up while working with python exclusively for some time and it stuck with me, happy to update it

Roshan-R:
The previous error for `prepare-build-context` have been resolved but there's some error in the `build-images` step. Seems like the log retention error is still going on and I can't get any logs to see what's wrong

joelcapitao:
Nice !

> The previous error for `prepare-build-context` have been resolved but there's some error in the `build-images` step. Seems like the log retention error is still going on and I can't get any logs to see what's wrong

I reported the error in #konflux:fedoraproject.org Matrix channel this morning.

Can you rebase https://gitlab.com/fedora/bootc/tekton-catalog/-/merge_requests/15 on latest change ? and also reset bundle change in this PR please ?

Roshan-R:
@joelcapitao I have made the requested changes

joelcapitao:
Thank you Roshan, I merged your MR https://gitlab.com/fedora/bootc/tekton-catalog/-/merge_requests/15.
I plan to update the bootc tasks and pipelines to include your change in registry, then we'll have to bump our pipeline definition to the new digest.

joelcapitao (approved):
LGTM

jbtrystram (approved):

PR #3922, opened 2025-12-12
Specify the image yaml file name in build-args.conf

dustymabe:
This will make it easier to support downstream rhel-coreos-config where we support building more than one variant from the main branch and the image.yaml files are named per variant.


aaradhak (approved):
Nice, LGTM.

dustymabe:
Not sure why the konflux jobs are failing because I can't see any logs when I click through.

There is a [JIRA issue](https://issues.redhat.com/browse/KONFLUX-11323) for not being able to see the logs.

dustymabe:
merging this since `testing-devel` isn't built with konflux anyway.

PR #3926, opened 2025-12-12
[rawhide] overrides: pin on adcli-0.9.2-10.fc43

dustymabe:
The new adcli-0.9.3.1-1.fc44 pulls in python [1].

[1] https://github.com/coreos/fedora-coreos-tracker/issues/2082

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -8,4 +8,9 @@
 # in the `metadata.reason` key, though it's acceptable to omit a `reason`
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
-packages: {}
+packages:
+  adcli:
+    evr: 0.9.2-10.fc43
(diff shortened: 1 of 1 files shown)


dustymabe:
The konflux image builds succeeded but the overall task failed because of:

```
Pipeline coreos-tenant/ can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "resolver type bundles\nname = deprecated-image-check\n": retryable error validating referenced object deprecated-image-check: Internal error occurred: failed calling webhook "webhook.pipeline.tekton.dev": failed to call webhook: Post "https://tekton-pipelines-webhook.openshift-pipelines.svc:443/defaulting?timeout=10s": EOF
```

I guess we need to fix something with our definitions. Unrelated to this PR.

HuijingHei (approved):
LGTM

PR #3928, opened 2025-12-15
[testing] tree: promote changes from testing-devel at d3f9ee4aa4b5e7a8b12e9a38840f46941f2ea377

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/1172.

PR #3929, opened 2025-12-15
[next] tree: promote changes from testing-devel at d3f9ee4

dustymabe:


dustymabe:
Promotion diffs is failing to `git checkout` and I have no idea why.

This is pretty much the same content as #3928, though. So I'm happy to ignore it for now.

jbtrystram (approved):

PR #3933, opened 2025-12-16
rename coreos-bootc-minimal-plus -> coreos-bootc-delta

dustymabe:
This file now just carries specific deltas (changes) that we have from bootc so let's rename it to something more appropriate.


aaradhak (dismissed):
LGTM

aaradhak (approved):

PR #3945, opened 2026-01-05
overrides: unpin adcli rpm

aaradhak:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/2082

There is a new pkg `adcli-0.9.3.1-3.fc44` available that would fix this issue.


dustymabe (approved):
LGTM

PR #3947, opened 2026-01-06
tests/platforms: add test for Azure SR-IOV udev rules

aaradhak:
Introduce a new test which verifies that udev rules for Azure SR-IOV network interfaces correctly mark them as unmanaged by NetworkManager. It only runs on Azure and uses Standard_D2s_v3 or larger instance type with Accelerated Networking enabled.

The test checks that SR-IOV interfaces (PCI devices with vendor drivers like mlx5_core) have the AZURE_UNMANAGED_SRIOV property set, and that NetworkManager respects this property by leaving them unmanaged.

### DIFF
--- a/tests/kola/platforms/azure/sriov/test.sh
+++ b/tests/kola/platforms/azure/sriov/test.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+## kola:
+##   # This test is targeted at Azure
+##   platforms: azure
+##   description: Verify that udev rules for Azure SRIOV network interfaces
+##                correctly mark them as unmanaged by NetworkManager.
+
+set -xeuo pipefail
+
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+# Find SRIOV network interfaces
+# Azure SR-IOV interfaces use Mellanox drivers (mlx4_core or mlx5_core)
+sriov_interfaces=()
+for iface in /sys/class/net/*; do
+    iface_name=$(basename "$iface")
+    # Skip loopback
+    if [ "$iface_name" = "lo" ]; then
+        continue
+    fi
+
+    if [ -e "$iface/device/driver" ]; then
+        driver=$(basename "$(readlink "$iface/device/driver")")
+        # SR-IOV interfaces on Azure use Mellanox drivers (mlx4_core or mlx5_core)
+        if [ "$driver" = "mlx4_core" ] || [ "$driver" = "mlx5_core" ]; then
+            sriov_interfaces+=("$iface_name")
+        fi
+    fi
+done
+
+# If no SRIOV interfaces found then this might be a VM size without Accelerated Networking
+# or the feature might not be enabled. We should have at least one SRIOV interface.
+if [ ${#sriov_interfaces[@]} -eq 0 ]; then
+    fatal "No SRIOV interfaces found, expected at least one Mellanox (mlx4_core/mlx5_core) network interface."
+fi
+
+# Check that each SRIOV interface has the AZURE_UNMANAGED_SRIOV property set
+# This property is set by the azure-vm-utils udev rules
+for iface in "${sriov_interfaces[@]}"; do
+    # Check the interface properties
+    if ! udevadm info --query=property --path="/sys/class/net/$iface" | grep -q "AZURE_UNMANAGED_SRIOV=1"; then
+        fatal "SRIOV interface $iface does not have AZURE_UNMANAGED_SRIOV=1 property."
+    fi
+done
+
+nm_devices=$(nmcli -t -f DEVICE,STATE device)
+
+for iface in "${sriov_interfaces[@]}"; do
+    if ! echo "$nm_devices" | grep -q "^$iface:"; then
+        fatal "SRIOV interface $iface not found in NetworkManager output."
+    fi
+    state=$(echo "$nm_devices" | grep "^$iface:" | cut -d: -f2)
+    if [ "$state" != "unmanaged" ]; then
+        fatal "NetworkManager is managing SRIOV interface $iface (state: $state). It should be unmanaged."
+    fi
+done
--- a/tests/kola/platforms/azure/sriov/data/commonlib.sh
+++ b/tests/kola/platforms/azure/sriov/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../../../data/commonlib.sh


src/config
```diff
@@ -0,0 +1 @@
+/home/aaradhak/coderepo/fedora-coreos-config
```
travier: You probably did not want to include this one

tests/kola/platforms/azure/sriov/test.sh
```diff
@@ -0,0 +1,64 @@
...
+# Check that each SRIOV interface has the AZURE_UNMANAGED_SRIOV property set
+# This property is set by the azure-vm-utils udev rules
+for iface in "${sriov_interfaces[@]}"; do
+    # Check the interface properties
+    if ! udevadm info --query=property --path="/sys/class/net/$iface" | grep -q "AZURE_UNMANAGED_SRIOV=1"; then
+        fatal "SRIOV interface $iface does not have AZURE_UNMANAGED_SRIOV=1 property."
+    fi
+done
+
+nm_devices=$(nmcli -t -f DEVICE,STATE device)
```
travier, line 46: Looks OK as is but can we get this in JSON format? That would be a bit nicer.

tests/kola/platforms/azure/sriov/test.sh
```diff
@@ -0,0 +1,64 @@
...
+    # Check NetworkManager's device list from cached output
+    if echo "$nm_devices" | grep -q "^$iface:"; then
+        # If the interface appears in nmcli output, check its state
+        state=$(echo "$nm_devices" | grep "^$iface:" | cut -d: -f2)
+        if [ "$state" != "unmanaged" ]; then
+            fatal "NetworkManager is managing SRIOV interface $iface (state: $state). It should be unmanaged."
+        fi
+    fi
+done
+
```
travier:
```suggestion
```

travier (commented):
LGTM. Mostly nits

travier:
You also have an override file that you probably don't want there: overrides/rootfs/usr/libexec/ignition-apply

src/config
aaradhak: my bad, I ll remove this

dustymabe:
Test itself LGTM. Obviously removing files that we don't want in this PR should happen before merge.

@marmijo mind a look as well since you are familiar with the Azure pieces?

aaradhak:
Yes, I have removed the unwanted files (it got added here accidentally) and also made the suggested change

tests/kola/platforms/azure/sriov/test.sh
aaradhak: I ll provide a JSON output to this and then parse the JSON.

dustymabe (dismissed):
LGTM

tests/kola/platforms/azure/sriov/test.sh
```diff
@@ -0,0 +1,60 @@
...
+    iface_name=$(basename "$iface")
+    # Skip loopback
+    if [ "$iface_name" = "lo" ]; then
+        continue
+    fi
+
+    if [ -e "$iface/device/driver" ]; then
+        driver=$(basename "$(readlink "$iface/device/driver")")
+        # SR-IOV interfaces are on PCI bus and not vmbus (hv_netvsc is the synthetic interface)
+        if [ "$driver" != "hv_netvsc" ] && [ -e "$iface/device/subsystem" ]; then
```
marmijo: In testing, I think were using `mlx5_core` here to determine which device to work on. Does this driver also work?

tests/kola/platforms/azure/sriov/test.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+## kola:
+##   # This test is targeted at Azure
+##   platforms: azure
+##   # This test requires an instance type that supports Accelerated Networking (SRIOV)
+##   # Standard_D2s_v3 and larger sizes support Accelerated Networking
+##   instanceType: "Standard_D2s_v3"
```
marmijo: We shouldn't need to declare an instance type since [`Standard_D2s_v3` is the default](http://github.com/coreos/coreos-assembler/blob/5c52e7983e8ccd7fd752987ce804f874dec162fa/mantle/cmd/kola/options.go#L242) for `x86_64`.
I didn't think about this when we were testing manually, but is SR-IOV networking supported on `aarch64`?

marmijo (commented):
LGTM, with a few comments.

@aaradhak and I tested a slightly different version of this manually together in an Azure VM. I'll try to run `cosa kola -p azure` with the test locally today.

tests/kola/platforms/azure/sriov/test.sh
aaradhak: No i dont think we would need the driver `hv_netvsc` for this test.

tests/kola/platforms/azure/sriov/test.sh
aaradhak: sounds good, I ll remove this reference.

dustymabe: > but is SR-IOV networking supported on aarch64

I would think it would be. Should be easy enough to check and make sure we run the test on both architectures.

tests/kola/platforms/azure/sriov/test.sh
aaradhak: I have explicitly mentioned about using mlx_core here now.

tests/kola/platforms/azure/sriov/test.sh
aaradhak: Based on this link - https://access.redhat.com/solutions/6965565 , I think SR-IOV networking is supported on aarch64

aaradhak:
I have tested this locally and it seems to be working 

```
[coreos-assembler]$ kola run ext.config.platforms.azure.sriov --arch=x86_64 -p=azure --azure-credentials /srv/azure-prod-secrets.json --azure-location westus2 --azure-disk-uri /subscriptions/8ddde9ba-f5e7-4a4f-a9a1-64181b250697/resourceGroups/ar-group/providers/Microsoft.Compute/galleries/argallery/images/argallery-def/versions/1.0.0
Using Standard_D2s_v3 size
2026-03-27T18:32:22Z kola: Unable to determine stream from '--denylist-stream' or meta.json. Won't consider denials based on stream.
=== RUN   ext.config.platforms.azure.sriov
--- PASS: ext.config.platforms.azure.sriov (300.16s)
PASS, output in tmp/kola/azure-2026-03-27-1832-50
```

tests/kola/platforms/azure/sriov/test.sh
aaradhak: @travier  Looks like the `-m json` option is not supported by nmcli, it only supports tabular or multiline modes. The kola test failed with Error: 'json' is not a valid argument for 'json' option. Using (-t) with grep seems to be a portable approach in this case.

tests/kola/platforms/azure/sriov/test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+for iface in "${sriov_interfaces[@]}"; do
+    # Check NetworkManager's device list from cached output
+    if echo "$nm_devices" | grep -q "^$iface:"; then
+        state=$(echo "$nm_devices" | grep "^$iface:" | cut -d: -f2)
+        if [ "$state" != "unmanaged" ]; then
+            fatal "NetworkManager is managing SRIOV interface $iface (state: $state). It should be unmanaged."
+        fi
+    fi
+done
```
marmijo, line 56:
```suggestion
for iface in "${sriov_interfaces[@]}"; do
    if ! echo "$nm_devices" | grep -q "^$iface:"; then
        fatal "SRIOV interface $iface not found in NetworkManager output."
    fi

    state=$(echo "$nm_devices" | grep "^$iface:" | cut -d: -f2)
    if [ "$state" != "unmanaged" ]; then
        fatal "NetworkManager is managing SRIOV interface $iface (state: $state). It should be unmanaged."
    fi
done
```

Since SR-IOV is expected to always be enabled for Azure kola runs, maybe this should also fail if the interface is absent from the `nmcli` output. This will help us catch regressions where the interface is present in the system but not surfaced to NetworkManager at all.

The earlier check on `[ ${#sriov_interfaces[@]} -eq 0 ]` verifies the interface exists, but not that NetworkManager sees it.

aaradhak: I have added the check for interface now.

marmijo (approved):
LGTM

PR #3948, opened 2026-01-07
tests/file-directory-permissions: Drop entry fixed in EL9.6+

travier:
tests/file-directory-permissions: Drop entry fixed in EL9.6+

---

tests/file-directory-permissions: Use function from shared lib


dustymabe (approved):
LGTM - thanks for the cleanups

PR #3949, opened 2026-01-08
test/kdump.crash: add sleep before crash

HuijingHei:
See https://github.com/coreos/rhel-coreos-config/issues/132


tests/kola/kdump/crash/test.sh
```diff
@@ -51,6 +51,8 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
       /tmp/autopkgtest-reboot-prepare aftercrash
       echo "Triggering sysrq"
       sync
+      # Give I/O a moment to settle after sync before crashing.
```
dustymabe, line 54:
```suggestion
    # Give I/O a moment to settle after sync before crashing.
    # We've seen this test hang forever in prow without this.
    # https://github.com/coreos/rhel-coreos-config/issues/132
```

PR #3959, opened 2026-01-14
[rawhide] overrides: fast-track grub2-2.12-50.fc44

dustymabe:
This update made it to stable rawhide repos but then was untagged (see https://forge.fedoraproject.org/releng/tickets/issues/13157) so we must fast-track it again until there is a new build for us to use. This will allow our ppc64le builds to work again see b35ea0e.


jbtrystram (approved):

marmijo (approved):
LGTM

PR #3964, opened 2026-01-15
tests/kola/commonlib: Fix get_fcos_stream for container case

travier:
Get the stream from the `com.coreos.stream` annotation from the container image.

### DIFF
--- a/tests/kola/data/commonlib.sh
+++ b/tests/kola/data/commonlib.sh
@@ -34,8 +34,10 @@ get_fedora_minimal_container_ref() {
     echo "${repo}:${tag}"
 }
 
+# Look for the "com.coreos.stream" annotation from the container image
 get_fcos_stream() {
-    rpm-ostree status -b --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]'
+    rpm-ostree status -b --json \
+        | jq --raw-output '.deployments[0]."base-commit-meta"."ostree.container.image-config" | fromjson | .config.Labels."com.coreos.stream"'
 }
 
 is_fcos() {


tests/kola/data/commonlib.sh
```diff
@@ -34,8 +34,11 @@ get_fedora_minimal_container_ref() {
     echo "${repo}:${tag}"
 }
 
+# Look for the "fedora-coreos.stream" annotation from the container image
 get_fcos_stream() {
-    rpm-ostree status -b --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]'
+    rpm-ostree status -b --json \
+        | jq -r '.deployments[0]."base-commit-meta"."ostree.manifest"' \
+        | jq -r '.annotations."fedora-coreos.stream"'
```
jbtrystram: we are shipping the stream in the `com.coreos.stream` label now as well, and the plan is to make this one canonical

https://github.com/coreos/fedora-coreos-config/blob/5d5b3190c2de8a6b8e639f71938e5bf278542a16/Containerfile#L53

jbtrystram (changes requested):

travier:
Thanks, updated.

tests/kola/data/commonlib.sh
```diff
@@ -34,8 +34,10 @@ get_fedora_minimal_container_ref() {
     echo "${repo}:${tag}"
 }
 
+# Look for the "com.coreos.stream" annotation from the container image
 get_fcos_stream() {
-    rpm-ostree status -b --json | jq -r '.deployments[0]["base-commit-meta"]["fedora-coreos.stream"]'
+    rpm-ostree status -b --json \
+        | jq -r '.deployments[0]."base-commit-meta"."ostree.manifest" | fromjson | .annotations."com.coreos.stream"'
```
dustymabe: I think what JB was saying is that we are moving towards the label being the source of truth (not the annotation) so:

```
[core@cosa-devsh ~]$ rpm-ostree status -b --json | jq --raw-output '.deployments[0]."base-commit-meta"."ostree.container.image-config" | fromjson | .config.Labels."com.coreos.stream"'
testing-devel
```

travier: Hum indeed. I don't know why it worked for me before.

dustymabe: we still have the annotation for now, so that's why it works either way

travier:
Updated

dustymabe (approved):
LGTM

PR #3965, opened 2026-01-15
.tekton: increase clone depth

joelcapitao:
Since 5c83258, we need to have a git tree containing more commits than just 1 (the default), because the python script versionary calculates the `n` value based on the number of commits. This patch increases the number of commits to clone up to 50. We expect not having more than 50 commits since the last manifest lockfile update.

### DIFF
--- a/.tekton/base/base/fedora-coreos.yaml
+++ b/.tekton/base/base/fedora-coreos.yaml
@@ -39,10 +39,12 @@ spec:
     - linux/arm64
     - linux/s390x
     - linux/ppc64le
+  - name: clone-depth
+    value: 50
   pipelineRef:
     params:
     - name: bundle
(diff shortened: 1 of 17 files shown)


joelcapitao:
FTR the Konflux log of this PR:
```
[prepare-build-context : prepare-build-context] + tee /tekton/results/labels
[prepare-build-context : prepare-build-context] [
[prepare-build-context : prepare-build-context]   "version=43.20260114.20.4",
[prepare-build-context : prepare-build-context]   "fedora-coreos.stream=testing-devel"
[prepare-build-context : prepare-build-context] ]
```

jbtrystram (approved):

dustymabe (approved):

dustymabe:
LGTM

PR #3973, opened 2026-01-21
build-rootfs: move buildroot-prep execution into build-rootfs

joelcapitao:
Move the buildroot-prep invocation from Containerfile into build-rootfs as a new run_buildroot_prep() function. This runs after inject_yumrepos() so that repos are available when buildroot-prep executes.

Also remove the cachi2 repo check from buildroot-prep since repo injection is now handled earlier in the build process by inject_yumrepos().

As a follow-up of #3723

Assisted-by: Claude (Opus 4)

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -112,6 +112,13 @@ def inject_yumrepos(srcdir):
             shutil.copy(repo, "/etc/yum.repos.d")
 
 
+def run_buildroot_prep():
+    # This allows FCOS/SCOS/RHCOS to do specific things before the main
+    # build process. Runs after inject_yumrepos() so repos are available.
+    buildroot_prep = os.path.join(SRCDIR, 'buildroot-prep')
+    if os.path.isfile(buildroot_prep):
+        subprocess.check_call([buildroot_prep])
+
 def build_rootfs(target_rootfs, srcdir, manifest_name,
                  image_config, osid, stream, version,
                  strict_mode, passwd_group_dir, mutate_os_release):
--- a/buildroot-prep
+++ b/buildroot-prep
@@ -1,22 +1,16 @@
 #!/bin/bash
 set -euo pipefail
 
-# This script is called in the buildah path to modify the build environment
-# before entering the shared build-rootfs script. Example use cases are to
-# fast-track fixes or work around bugs.
+# This script is called very early in the buildah path from build-rootfs
+# after inject_yumrepos() to modify the build environment. Example use cases
+# are to fast-track fixes or work around bugs.
+# Note that any dowloaded packages here needs to be included in lockfiles (the main
+# or overrides one).
 
 arch=$(uname -m)
 . /etc/os-release
 
-# cachi2 is the repo Konflux injects when hermetic build is enabled and
-# is self-sufficient to pull all the required RPMs.
-if [ ! -f "/etc/yum.repos.d/cachi2.repo" ]; then
-    cp /src/fedora-coreos-continuous.repo /etc/yum.repos.d
-fi
-# NOTE: try to remove anything that queries repos here once it's no longer
-# needed so that we don't unnecessarily pay for repo metadata.
-
 # make sure we have https://github.com/coreos/rpm-ostree/pull/5454
 if ! rpm-ostree compose build-chunked-oci -h | grep -q -- --label; then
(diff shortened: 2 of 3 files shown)


jbtrystram (dismissed):

buildroot-prep
```diff
@@ -1,21 +1,13 @@
...
-# is self-sufficient to pull all the required RPMs.
-if [ ! -f "/etc/yum.repos.d/cachi2.repo" ]; then
-    cp /src/fedora-coreos-continuous.repo /etc/yum.repos.d
-fi
-# NOTE: try to remove anything that queries repos here once it's no longer
-# needed so that we don't unnecessarily pay for repo metadata.
-
 # make sure we have https://github.com/coreos/rpm-ostree/pull/5454
 if ! rpm-ostree compose build-chunked-oci -h | grep -q -- --label; then
     sudo dnf update rpm-ostree -y --repo fedora-coreos-continuous --releasever "$VERSION_ID"
```
dustymabe: I think we need to rework this slightly. In the very least `--repo fedora-coreos-continuous` needs to be removed.

dustymabe:
So.. I suggested this change and I still think we should merge it.

But.. I'd like to challenge us to think about the problems we are trying to solve here and understand the limitations of the new scheme (which is required by the constraints we are trying to fit into with hermetic builds).

Given the new flow we can't really (I think) override the buildroot (bootc base image conainer we are using for the build) with any RPMs that we don't also ship in the actual CoreOS payload. Is my understanding correct?

Also, we can't override any RPM unless it's in the lockfiles that are then translated by some of the initial konflux steps. So would require us to update the lockfiles at the same time we update buildroot-prep, which I think is fine. I just want to make sure I understand correctly.

jbtrystram:
> Given the new flow we can't really (I think) override the buildroot (bootc base image conainer we are using for the build) with any RPMs that we don't also ship in the actual CoreOS payload. Is my understanding correct?

Yes. Well, maybe you could workaround this by an extra RPM in the overrides, so it's available during the container build, but if it's not specified in the manifests it's not going to be installed in the final rootfs. This would require testing.

> Also, we can't override any RPM unless it's in the lockfiles that are then translated by some of the initial konflux steps. So would require us to update the lockfiles at the same time we update buildroot-prep, which I think is fine. I just want to make sure I understand correctly.

Yes, correct. That said, with #3974 any changes in the override files will disable hermetic for the pipeline runs on the PR, so you can test that in the same PR :)

joelcapitao:
@dustymabe yes, you're right for both statement. Enabling hermetic constraints usage of `buildroot-prep`. I'll add a note in the script explaining that the packages install in the buildroot needs to be available in the lockfile (the main or override one).

> Yes, correct. That said, with https://github.com/coreos/fedora-coreos-config/pull/3974 any changes in the override files will disable hermetic for the pipeline runs on the PR, so you can test that in the same PR :)

I managed to make it work even when hermetic is enabled.

buildroot-prep
joelcapitao: I amended my commit to include your suggestion

dustymabe (approved):
LGTM

build-rootfs
```diff
@@ -112,6 +112,13 @@ def inject_yumrepos(srcdir):
             shutil.copy(repo, "/etc/yum.repos.d")
 
 
+def run_buildroot_prep():
```
jlebon, line 115: Hmm... this is not actually called anywhere?

jlebon (commented):
> Given the new flow we can't really (I think) override the buildroot (bootc base image conainer we are using for the build) with any RPMs that we don't also ship in the actual CoreOS payload. Is my understanding correct?

FWIW, this was exactly the use case I was trying out today as part of https://github.com/coreos/fedora-coreos-config/pull/3993 to use chunkah to split FCOS without shipping chunkah itself in FCOS. Not a blocker in this case because it's still a POC/WIP (and I think I'll end up proposing chunkah in the bootc base images), but regardless I think it'd be good to support this properly.

See also equivalent functionality in Hummingbird: https://hummingbird-project.io/docs/contributing/image-configuration-reference/#rpm_packagesbuild-deps.

build-rootfs
dustymabe: nope.. fixed in https://github.com/coreos/fedora-coreos-config/commit/7f4fdc769291c79f7f50a52aa8e8ba2016837fa8

PR #3985, opened 2026-01-28
[next-devel] Move more things into build-args.conf

dustymabe:
A few sources of information were moved around/deleted in https://github.com/coreos/fedora-coreos-config/pull/3978 so let's adjust our files here that need updating.

### DIFF
--- a/build-args.conf
+++ b/build-args.conf
@@ -1,18 +1,14 @@
 # Stream-specific build arguments.
 # Pass to buildah/podman using `--build-arg-file`.
-#
-# This is the base version. This value will get used as a starting point for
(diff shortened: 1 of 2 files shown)


joelcapitao (approved):
LGTM

joelcapitao:
Looks like 43 is not in https://dl.fedoraproject.org/pub/fedora/linux/development/ anymore ? but was it there before ? :thinking:

dustymabe:
This branch isn't active right now so I don't expect tests to pass. We'll fix any issues with it when we re-enable `next-devel` closer to F44 beta.

PR #3986, opened 2026-01-28
versionary: handle corner case with SCOS versioning

dustymabe:
Since the base version of SCOS is 9.0 then the y component will be 0 and the `if not y` will evaluate to True when we don't want it to. Let's switch the initial values to be None and explicitly check against None in the cases where we want to determine if the y or z have been set yet or not.


jbtrystram (approved):
Lgtm

PR #3988, opened 2026-01-29
denylist: deny kdump.crash.ssh && ext.config.kdump.crash on s390x

prestist:
https://github.com/coreos/fedora-coreos-tracker/issues/2100

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -7,3 +7,19 @@
 ##  warn: true
 ##  arches:
 ##    - aarch64
+
+- pattern: kdump.crash.ssh
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2100
+  snooze: 2026-02-28
+  arches:
+    - s390x
+  streams:
+    - rawhide
+
+- pattern: ext.config.kdump.crash
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2100
+  snooze: 2026-02-28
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM.

extra credit (only if you want to) you could test if using a pattern of `*kdump.crash*` would match both of these so you only needed one entry in the denylist. You can test it on your laptop by removing the arches: definition for the test.

prestist:
@dustymabe sure I will give that a quick try, im curious now.

dustymabe (changes requested):
actually let's limit this to the rawhide stream too so if it sneaks in to F43 we'll know

prestist:
Replaced by https://github.com/coreos/fedora-coreos-config/pull/3989

dustymabe:
> Replaced by #3989

you had it right. The denylist is shared across all branches here. just add streams: entry to limit it to rawhide.

prestist:
Ah sorry, thank you, ill update Also I did just update my local cosa, by specifying the glob as kdump.crash.* it pulls in another test, so I think its best to keep specific in this case. 

The test it pulled in is "kdump.crash.nfs"

dustymabe (approved):
LGTM

dustymabe:
CI here doesn't test s390x or rawhide so I'll admin merge it so we can kick off a new bump-lockfile for rawhide.

PR #3989, opened 2026-01-29
denylist: deny kdump.crash.ssh && ext.config.kdump.crash on s390x

prestist:
see : https://github.com/coreos/fedora-coreos-tracker/issues/2100


prestist:
Did not get around to checking glob yet @dustymabe dealing with storage constraints on laptop

PR #3990, opened 2026-01-30
denylist: extend kdump.crash denials to include kdump.crash.nfs

dustymabe:
This test is also failing on s390x along with the others.

https://github.com/coreos/fedora-coreos-tracker/issues/2100


HuijingHei (approved):
LGTM

dustymabe:
Merging this since CI on this PR won't test rawhide or s390x

PR #3996, opened 2026-02-02
[stable] tree: promote changes from testing at f365331791dcac8b47a540f529cf96e64fcd964e

coreosbot-releng:
Triggered by @tlbueno in https://github.com/coreos/fedora-coreos-streams/issues/1188.


dustymabe:
Added a commit here since COSA was updated in https://github.com/coreos/coreos-assembler/pull/4424 we need to forward-port c0a3ccca

dustymabe (approved):

dustymabe:
The promotion diff warnings are just for the changes I made in the extra commit I pushed here. We should be good.

PR #3999, opened 2026-02-04
build-rootfs: Gracefully handle short version strings

joelcapitao:
get_treefile() crashes when parsing versions with fewer than 3
components (e.g., "44"). Now we pad the version parts list with
"0" to ensure we always have at least 3 elements.
As a follow-up of e6d3ed2.


dustymabe:
I feel like this needs a heavy comment.

Here is something similar, but might be a little more obvious why the change is needed:

```
(x, y, _) = version.split('.', 2) if '.' in version else (version, '0', '0')
```

dustymabe (approved):

PR #4000, opened 2026-02-06
Drop fedora-repos-ostree in F44

travier:
This may have other implications. Let's see what CI says.


dustymabe:
Seems sane to me. Since we shouldn't be using them anyway it would probably be OK to just stop shipping them now, but phasing it in with the move to F44 would mean any side effects should get ironed out in `next` way before anything else.

dustymabe (approved):
LGTM

PR #4001, opened 2026-02-06
Containerfile: support injecting openshift version labels

dustymabe:
Downstream we've been injecting labels like this for a while via the `ostree-container-inject-openshift-cvo-labels` image.yaml setting. This ends up with a labels on an image like:

```
  "io.openshift.build.version-display-names": "machine-os=Red Hat Enterprise Linux CoreOS",
  "io.openshift.build.versions": "machine-os=9.6.20260128-1",
```

Since we now do a container build by default and we support a Containerfile let's just add the labels in the Container build directly and trigger it via a build arg.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -20,6 +20,7 @@ FROM ${BUILDER_IMG} as builder
 
 ARG ID=overridden
 ARG VERSION=overridden
+ARG DESCRIPTION=overridden
 ARG STREAM=overridden
 ARG MUTATE_OS_RELEASE=overridden
 ARG MANIFEST=overridden
@@ -27,6 +28,7 @@ ARG IMAGE_CONFIG=overridden
 # XXX: see inject_passwd_group() in build-rootfs
 ARG PASSWD_GROUP_DIR
 ARG STRICT_MODE=0
+ARG INJECT_OPENSHIFT_VERSION_LABELS=""
 
 COPY . /src
 # canonicalize permission bits, see also https://gitlab.com/fedora/bootc/base-images/-/merge_requests/274
@@ -51,7 +53,9 @@ RUN --mount=type=bind,target=/run/src,rw \
         --bootc --format-version=1 --rootfs /target-rootfs \
         --output oci-archive:/run/src/out.ociarchive \
         --label com.coreos.inputhash=$(cat /run/inputhash) \
-        --label com.coreos.stream=$STREAM
+        --label com.coreos.stream=$STREAM \
+        ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.versions=machine-os=${VERSION}} \
+        ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.version-display-names=machine-os="${DESCRIPTION}"}
 
 FROM oci-archive:./out.ociarchive
 ARG VERSION


Containerfile
```diff
@@ -51,7 +53,9 @@ RUN --mount=type=bind,target=/run/src,rw \
         --bootc --format-version=1 --rootfs /target-rootfs \
         --output oci-archive:/run/src/out.ociarchive \
         --label com.coreos.inputhash=$(cat /run/inputhash) \
-        --label com.coreos.stream=$STREAM
+        --label com.coreos.stream=$STREAM \
+        ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.versions=${VERSION}} \
+        ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.version-display-names=${DESCRIPTION}}
```
dustymabe: Note. Using the DESCRIPTION here means we'll most likely need to change the [DESCRIPTION downstream](https://github.com/coreos/rhel-coreos-config/blob/33c985d7636144df355608d2f3399d6ac955d850/build-args-rhel-9.8.conf#L7) slightly if we want the `io.openshift.build.version-display-names` label to stay similar to what it was in the past..

i.e. we need to make this change:

- `DESCRIPTION=RHEL CoreOS 9.8` -> `DESCRIPTION=Red Hat Enterprise Linux CoreOS 9.8`

So the label will change in this way:

- from `io.openshift.build.version-display-names=machine-os=Red Hat Enterprise Linux CoreOS`
- to `io.openshift.build.version-display-names=machine-os=Red Hat Enterprise Linux CoreOS 9.8`

The difference here being the added RHEL version at the end of the string.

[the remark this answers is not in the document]
dustymabe: > 1.

Good catch. Will fix.

> 2.

I broke it up on two lines because it was a really long line already.

dustymabe: Note the difference here in making this change is the lines below:

```
      org.opencontainers.image.title=$DESCRIPTION \                                                                                                                                          
      org.opencontainers.image.description=$DESCRIPTION
```

will cause the title/description to be changed. I think that Should be OK.

aaradhak (dismissed):
LGTM

Containerfile
```diff
@@ -20,13 +20,15 @@ FROM ${BUILDER_IMG} as builder
...
 ARG VERSION=overridden
+ARG DESCRIPTION=overridden
 ARG STREAM=overridden
 ARG MUTATE_OS_RELEASE=overridden
 ARG MANIFEST=overridden
 ARG IMAGE_CONFIG=overridden
 # XXX: see inject_passwd_group() in build-rootfs
 ARG PASSWD_GROUP_DIR
 ARG STRICT_MODE=0
+ARG INJECT_OPENSHIFT_VERSION_LABELS
```
jlebon:
```suggestion
ARG INJECT_OPENSHIFT_VERSION_LABELS=""
```

Otherwise this will cause a buildah warning when unset.

jlebon (commented):
Seems OK to me. Though part of me thinks perhaps we should do the opposite and move these labels _up_ to the node image build. That seems more appropriate given it's OpenShift-specific goop. It's very surprising to see these labels in f-c-c.

We'd then be able to have the OpenShift version in the display name. But it also give us flexibility we might need for diverging these values from the base CoreOS image. (Obviously possible as is too to override a base label, so we wouldn't be blocked on this or anything.)

E.g. right now the version string is used as the tag name in the DTK imagestream for looking up the associated DTK image. This will not be correct for NVidia if we're doing a kernel replacement from the node image build.

dustymabe:
> Seems OK to me. Though part of me thinks perhaps we should do the opposite and move these labels _up_ to the node image build. That seems more appropriate given it's OpenShift-specific goop. It's very surprising to see these labels in f-c-c.

I don't disagree here.

I'd argue the [way](https://github.com/coreos/fedora-coreos-pipeline/commit/b5669ddd1e6e24eacc01b2bf07522deb78bbc76c) we implemented `io.openshift.os.streamclass` would be more appropriate if it was somewhere in https://github.com/openshift/os but without a chunking step is there any way to add labels dynamically at that level?

jlebon:
> > Seems OK to me. Though part of me thinks perhaps we should do the opposite and move these labels _up_ to the node image build. That seems more appropriate given it's OpenShift-specific goop. It's very surprising to see these labels in f-c-c.
> 
> I don't disagree here.
> 
> I'd argue the [way](https://github.com/coreos/fedora-coreos-pipeline/commit/b5669ddd1e6e24eacc01b2bf07522deb78bbc76c) we implemented `io.openshift.os.streamclass` would be more appropriate if it was somewhere in [openshift/os](https://github.com/openshift/os?rgh-link-date=2026-02-06T19%3A53%3A26Z) but without a chunking step is there any way to add labels dynamically at that level?

The value would be dynamic but not whether or not we add a label at all, so a build arg should suffice.

dustymabe:
Talked with @jlebon about this and we agreed to move forward with this for now and when we move the openshift/os building to konflux we may adjust the setting of the labels.

jlebon (approved):

PR #4004, opened 2026-02-06
[branched] Update manifest/build-args on branched stream

dustymabe:
This is similar to 14f1100 for rawhide. It just gets us in line with where things are defined in latest Fedora CoreOS.


dustymabe:
I don't expect CI to pass on this one.

aaradhak (approved):
LGTM

PR #4007, opened 2026-02-10
workflows: add python3-dotenv deps to some actions

dustymabe:
In b54d03b ci/overrides.py gained a requirement on dotenv so let's add it to the install list for actions that use overrides.py.


aaradhak (approved):
LGTM

PR #4017, opened 2026-02-18
tests/toolbox: adjust to be runnable on RHCOS

### LINKED ISSUE #1035: EL9: Fix toolbox test
The toolbox test was not running effectively running and is not working for EL 9.

See https://github.com/openshift/os/pull/1034

Rolv-Apneseth:
Closes openshift/os#1035

This will should enable running the basic toolbox test for RHCOS.

To explain a couple of the reasons it would not work in its current state in RHCOS, and the changes this PR makes to get around them:

1. No `machinectl` command on RHCOS. I've instead used `su - core` to run commands as the core user. Also has the minor benefit of propagating error codes, though I opted to keep the explicit checks anyway.
2. Grep and awk calls were looking only for `fedora-toolbox-`. I simply expanded the search to also accept `rhel-toolbox-`.
3. The rhel-9.8 variant is using [this toolbox](https://github.com/coreos/toolbox) - this is the one that adds the most complexity to this script. Commands for this one are completely different, the script uses root podman and requires extra configuration. I have a butane config to setup the tool as required, and I've defined and overridden functions to test the same functionality across this script and the modern toolbox binary.

I would love feedback, as I'm not sure all the decisions I made were the best ones. For example, I could see a case for keeping the `machinectl` path for systems which have it available, as it actually simulates an interactive login. Also, do we even need to test that older toolbox script? Either way, I'm happy to make any changes.

### DIFF
--- a/tests/kola/toolbox/test.sh
+++ b/tests/kola/toolbox/test.sh
@@ -1,52 +1,114 @@
 #!/bin/bash
 ## kola:
 ##   tags: "platform-independent needs-internet"
-##   # This test only runs on FCOS because RHCOS is missing the `machinectl` command.
-##   # Additionally, there are some distro specific choices made for this test that
-##   # should/could be adapted for RHCOS.
-##   # TODO-RHCOS: adapt test for RHCOS specifics or create separate RHCOS toolbox test
-##   distros: fcos
 ##   # Toolbox container is currently available only for x86_64 and aarch64 in Fedora
 ##   architectures: x86_64 aarch64
 ##   description: Make sure that basic toolbox functionality works (creating,
 ##     running commands, and removing).
-
-# Important note: Commands are run indirectly via calls to `machinectl shell`
-# to re-create the user environment needed for unprivileged podman
-# functionality. However, machinectl shell does not propagate the exit
-# code/status of the invoked shell process thus we need additionnal checks to
-# ensure that previous commands were successful.
+##   creationDate: 2026-05-05
 
 set -xeuo pipefail
 
 # shellcheck disable=SC1091
 . "$KOLA_EXT_DATA/commonlib.sh"
 
+# IMPORTANT: Commands are run indirectly via `su - core` to re-create the
+# user environment needed for unprivileged podman functionality.
+run_as_core() {
+    su - core -c "$(printf '%q ' "$@")"
+}
+
+# Functions for testing basic functionality - overridden depending on toolbox being used
+toolbox_create() {
+    run_as_core /bin/toolbox create --assumeyes 1> /dev/null
+}
+
+toolbox_run_basic() {
+    run_as_core /bin/toolbox run touch ok_toolbox
+}
+
+toolbox_list() {
+    run_as_core /bin/toolbox list --containers
+}
+
+toolbox_count() {
+    toolbox_list | grep --count -E "(fedora|rhel)-toolbox-" || true
+}
+
+toolbox_rm() {
+    toolbox="$(toolbox_list | awk '/(fedora|rhel)-toolbox-/ {print $2}')"
+    run_as_core /bin/toolbox rm -f "${toolbox}"
+}
+
+# Older variants (e.g. RHEL-9.8) use https://github.com/coreos/toolbox
+# NOTE: script uses privileged podman
+if file /bin/toolbox | grep -q "shell script"; then
+    echo "Using toolbox script <https://github.com/coreos/toolbox>"
+
+    # Create configuration for coreos/toolbox
+    rhel_major_version=$(get_rhel_maj_ver)
+    cat << EOF > /var/home/core/.toolboxrc
+REGISTRY=registry.access.redhat.com
+IMAGE=ubi${rhel_major_version}/toolbox:latest
+TOOLBOX_NAME=rhel-toolbox-latest
+AUTHFILE=/var/home/core/config.json
+EOF
+    # Toolbox fails if the AUTHFILE does not exist, but we don't actually need credentials
+    echo '{}' > /var/home/core/config.json
+
+    toolbox_create() {
+        # Container created on first run of any command
+        run_as_core /bin/toolbox true
+    }
+
+    toolbox_run_basic() {
+        run_as_core /bin/toolbox touch /host/home/core/ok_toolbox
+    }
+
+    toolbox_list() {
+        /bin/podman ps -a
+    }
+
+    toolbox_rm() {
+        toolbox="$(toolbox_list | awk '/(fedora|rhel)-toolbox-/ {print $1}')"
+        /bin/podman rm -f "${toolbox}"
+    }
+elif is_rhcos; then
+    # When using an image that's not yet released (e.g.
+    # 10.2 as of writing this) the image won't exist in the
+    # registry. Let's just use the `:latest` tag so we don't
+    # hit that problem.
+    rhel_major_version=$(get_rhel_maj_ver)
+    cat << EOF > /etc/containers/toolbox.conf
+[general]
+image = "registry.access.redhat.com/ubi${rhel_major_version}/toolbox:latest"
+EOF
+fi
+
 # Try five times to create the toolbox to avoid container registry infra flakes
 for i in {1..5}; do
-    machinectl -q shell core@ /bin/toolbox create --assumeyes 1>/dev/null
-    if [[ $(machinectl -qE TERM=dumb shell core@ /bin/toolbox list --containers | grep --count fedora-toolbox-) -ne 1 ]]; then
-        echo "Could not create toolbox on try: $i"
-        sleep 10
-    else
+    toolbox_create || true
+    if [[ $(toolbox_count) -eq 1 ]]; then
         break
     fi
+
+    if [[ $i -eq 5 ]]; then
+        fatal "Could not create toolbox after 5 attempts"
+    fi
+
+    echo "Could not create toolbox on try: $i"
+    sleep 10
 done
-if [[ $(machinectl -qE TERM=dumb shell core@ /bin/toolbox list --containers | grep --count fedora-toolbox-) -ne 1 ]]; then
-    fatal "Could not create toolbox"
-fi
 ok toolbox create
 
-machinectl -q shell core@ /bin/toolbox run touch ok_toolbox
+toolbox_run_basic
 if [[ ! -f '/home/core/ok_toolbox' ]]; then
     fatal "Could not run a simple command inside a toolbox"
 fi
 ok toolbox run
 
-toolbox="$(machinectl -qE TERM=dumb shell core@ /bin/toolbox list --containers | awk '/fedora-toolbox-/ {print $2}')"
-machinectl -q shell core@ /bin/podman stop "${toolbox}"
-machinectl -q shell core@ /bin/toolbox rm "${toolbox}"
-if [[ -n "$(machinectl -qE TERM=dumb shell core@ /bin/toolbox list --containers)" ]]; then
+toolbox_rm
+if [[ $(toolbox_count) -ne 0 ]]; then
     fatal "Could not remove the toolbox container"
 fi
 ok toolbox rm


Rolv-Apneseth:
Anybody know why that CI is failing?

```
./tmt-test-wrapper-inner.sh-script-00-1-default-0-default-0: line 1: tmt: command not found
```

Also, should these be pointing at a fork? I assume this was for testing / debugging:

```
git clone --recurse-submodules https://github.com/Roshan-R/fedora-coreos-config testcode
...
git -C testcode fetch https://github.com/Roshan-R/fedora-coreos-config kola-testing-farm:gluetool/kola-testing-farm
```

tests/kola/toolbox/config.bu
```diff
@@ -0,0 +1,16 @@
+variant: fcos
+version: 1.6.0
+storage:
+  files:
+    # Toolbox configuration for older streams using https://github.com/coreos/toolbox
```
dustymabe: What do you mean by this exactly?

tests/kola/toolbox/config.bu
```diff
@@ -0,0 +1,16 @@
...
+    - path: /var/home/core/.toolboxrc
+      contents:
+        inline: |
+          REGISTRY=quay.io
+          IMAGE=fedora/fedora-toolbox:latest
+          TOOLBOX_NAME=fedora-toolbox-latest
+          AUTHFILE=/var/home/core/config.json
+    - path: /var/home/core/config.json
+      contents:
+        inline: "{}"
```
dustymabe: Not sure I understand the point of this

tests/kola/toolbox/test.sh
```diff
@@ -1,52 +1,104 @@
...
-# ensure that previous commands were successful.
-
 set -xeuo pipefail
 
 # shellcheck disable=SC1091
 . "$KOLA_EXT_DATA/commonlib.sh"
 
+# IMPORTANT: Commands are run indirectly via `su - core` to re-create the
+# user environment needed for unprivileged podman functionality.
+run_as_core()
```
dustymabe: we use a similar function to this in a few other tests (search for `runascoreuser`). might be a good idea to put a function like this in commonlib.sh and then use that everywhere.

migt be good to point AI at this and it could do it in like 10s.

dustymabe: Note though.. that `machinectl shell` is a more proper way to get an environment that more closely models what a user session would have. Maybe we should continue to use `machinectl shell` here.

cc @travier

dustymabe: oh - hmm. I now see the comment above that mentions RHCOS doesn't have `machinectl`. I wonder if we should just add it.

tests/kola/toolbox/test.sh
```diff
@@ -1,52 +1,104 @@
...
+toolbox_rm()
+{
+    toolbox="$(toolbox_list | awk '/(fedora|rhel)-toolbox-/ {print $2}')"
+    run_as_core /bin/podman stop "${toolbox}"
+    run_as_core /bin/toolbox rm "${toolbox}"
+}
+
+# Older variants (e.g. RHEL-9.8) use https://github.com/coreos/toolbox
+# NOTE: script uses privileged podman
+if run_as_core file /bin/toolbox | grep -q "shell script"; then
```
dustymabe:
```suggestion
if file /bin/toolbox | grep -q "shell script"; then
```

probably don't need to run through the user switching here?

tests/kola/toolbox/test.sh
```diff
@@ -1,52 +1,104 @@
...
+# IMPORTANT: Commands are run indirectly via `su - core` to re-create the
+# user environment needed for unprivileged podman functionality.
+run_as_core()
+{
+    su - core -c "$(printf '%q ' "$@")"
+}
+
+# Functions for testing basic functionality - overridden depending on toolbox being used
+toolbox_create()
+{
```
dustymabe: This is strictly a preference of mine and you don't have to take the suggestion..

I think most of our scripts have the opening brace on the same line:


```suggestion
toolbox_create() {
```

If you make this change I'd make it everywhere else in this file.

tests/kola/toolbox/test.sh
```diff
@@ -1,52 +1,104 @@
...
+toolbox_count()
+{
+    toolbox_list | grep --count -E "(fedora|rhel)-toolbox-" || true
+}
+
+toolbox_rm()
+{
+    toolbox="$(toolbox_list | awk '/(fedora|rhel)-toolbox-/ {print $2}')"
+    run_as_core /bin/podman stop "${toolbox}"
+    run_as_core /bin/toolbox rm "${toolbox}"
```
dustymabe: this is the same as just calling toolbox with a `-f` to stop the running containers first?

```suggestion
    run_as_core /bin/toolbox rm -f "${toolbox}"
```

tests/kola/toolbox/config.bu
Rolv-Apneseth: I will double check but I believe when I was running it, the coreos toolbox would fail if the file didn't exist

tests/kola/toolbox/config.bu
Rolv-Apneseth: Yeah sorry, wasn't sure how to word it. I suppose it's only RHCOS 9.x and older that use the coreos toolbox

tests/kola/toolbox/test.sh
Rolv-Apneseth: Yes that was a big part of the complication with getting it to run on rhcos. I did consider adding it to the common lib but I wanted to see if we were gonna keep the `machinectl` approach for systems that support it first

tests/kola/toolbox/test.sh
Rolv-Apneseth: Sure, I don't have a strong preference. It's just what my formatter did

tests/kola/toolbox/test.sh
Rolv-Apneseth: Didn't know about this, thanks

tests/kola/toolbox/test.sh
Rolv-Apneseth: Yeah not sure why I put that there

tests/kola/toolbox/test.sh
Rolv-Apneseth: I just copied from what was there before:

```
machinectl -q shell core@ /bin/podman stop "${toolbox}"
machinectl -q shell core@ /bin/toolbox rm "${toolbox}"
```

But I'll change them both to use `-f`

tests/kola/toolbox/test.sh
dustymabe: yeah, let's see what @travier thinks on the `machinectl` in RHCOS thing.

tests/kola/toolbox/config.bu
Rolv-Apneseth: Yeah, it fails without that. Either:

```
Error: credential file is not accessible: faccessat /var/lib/kubelet/config.json: no such file or directory
```

Or, if specifying AUTHFILE:

```
Error: credential file is not accessible: faccessat /var/home/core/config.json: no such file or directory
```

tests/kola/toolbox/config.bu
Rolv-Apneseth: How about

```
Toolbox configuration for RHCOS variants < 10.x using https://github.com/coreos/toolbox
```

tests/kola/toolbox/config.bu
Rolv-Apneseth: Added a comment

dustymabe: ok so you mean `coreos/toolbox` fails if a credentials file doesn't exist? Does it work if you use `AUTHFILE=/dev/null` ?

tests/kola/toolbox/config.bu
dustymabe: ok I see. So this is wholly for the legacy toolbox?

TBH in that case I would just put it in the test itself for the legacy case:

```
cat <<EOF > /var/home/core/.toolboxrc
REGISTRY=quay.io
IMAGE=fedora/fedora-toolbox:latest
TOOLBOX_NAME=fedora-toolbox-latest
AUTHFILE=/dev/null
EOF
```

travier: Unfortunately, RHCOS 10 also use the legacy coreos/toolbox implementation. So it's a RHCOS vs FCOS thing right now.

tests/kola/toolbox/config.bu
```diff
@@ -0,0 +1,17 @@
...
+  files:
+    # Toolbox configuration for older streams using https://github.com/coreos/toolbox
+    # RHEL registry requires credentials - use quay.io registry and a Fedora image for testing
+    - path: /var/home/core/.toolboxrc
+      contents:
+        inline: |
+          REGISTRY=quay.io
+          IMAGE=fedora/fedora-toolbox:latest
+          TOOLBOX_NAME=fedora-toolbox-latest
+          AUTHFILE=/var/home/core/config.json
```
travier: Unfortunately that will not test the same thing as the legacy coreos/toolbox behaves differently when there are special labels on the image and I think I remember those being there in the RHEL one but not the upstream containers/toolbox one.

tests/kola/toolbox/test.sh
travier: I would prefer we keep using machinectl on FCOS. I think I had tried with `su`, I don't remember why it did not work.

travier:
As this will never be used in FCOS, I think you should keep the current FCOS test as is and duplicate it in RHCOS instead. That way we will not have to undo the change here when we move away from the legacy one in RHCOS and we'll only have to delete it there.

tests/kola/toolbox/config.bu
Rolv-Apneseth: I'll try

tests/kola/toolbox/test.sh
dustymabe: I think my question was more: should we add `machinectl` to RHCOS?

travier: I don't think so if we don't need it

travier: if we don't need it for something else. I don't remember why it's in FCOS

tests/kola/toolbox/config.bu
Rolv-Apneseth: When I was testing it was only 9.8 that had it, not even c9s.

Here's the toolbox package metadata from the release browser for each (not sure where else to get a list of packages):

- 9.8: `[ "toolbox", "0", "0.1.2", "1.rhaos4.22.el9", "noarch" ]`
- 10.0: `[ "toolbox", "0", "0.3", "1.el10", "x86_64" ]`
- c9s: `[ "toolbox", "0", "0.3", "1.el9", "x86_64" ]`
- c10s: `[ "toolbox", "0", "0.3", "1.el10", "x86_64" ]`

tests/kola/toolbox/test.sh
Rolv-Apneseth: I had confirmed this works on both, which is why I didn't choose to split it here, but I don't mind keeping `machinectl` for FCOS if we think it's worth doing it the more "proper" way when we can

Rolv-Apneseth:
> As this will never be used in FCOS, I think you should keep the current FCOS test as is and duplicate it in RHCOS instead. That way we will not have to undo the change here when we move away from the legacy one in RHCOS and we'll only have to delete it there.

But then any changes/additions will need to be made in 2 places even though they'll eventually both be using containers/toolbox right?

For example, I was thinking we should add at least a line in this test checking toolbox works when `$SHELL` is unset or not exported, like that issue we had on openshift nodes running rhcos 10.2.

tests/kola/toolbox/config.bu
Rolv-Apneseth: Would you mind explaining, I don't really understand. Would I need to get it to run the RHEL images here then? How do I get the required authentication?

I assume [this](https://github.com/coreos/toolbox/blob/56dbd2ccdd15cd0efe9eb70394effd09fb264eb0/rhcos-toolbox#L48-L63) is what you're referring to with the labels

tests/kola/toolbox/config.bu
Rolv-Apneseth: Unfortunately not:

```
Error: unable to copy from source docker://quay.io/fedora/fedora-toolbox:latest: initializing source docker://quay.io/fedora/fedora-toolbox:latest: getting username and password: reading JSON file "/dev/null": unmarshaling JSON at "/dev/null": unexpected end of JSON input
```

Rolv-Apneseth: And yes, this whole butane file is only for coreos/toolbox

tests/kola/toolbox/config.bu
travier: > Would you mind explaining, I don't really understand. Would I need to get it to run the RHEL images here then? How do I get the required authentication?

Yes unfortunately.

> I assume [this](https://github.com/coreos/toolbox/blob/56dbd2ccdd15cd0efe9eb70394effd09fb264eb0/rhcos-toolbox#L48-L63) is what you're referring to with the labels

Yes

travier: But let's not block on that ,we can start with what you have here.

travier:
Answering https://github.com/coreos/fedora-coreos-config/pull/4017#discussion_r2966529430 here to not have it lost.

What's confusing is that it looks like we haven't properly completed RHCOS 10.2 branching work and thus we don't have the legacy coreos/toolbox in there yet. SCOS & FCOS use the new containers/tooblox. RHCOS must keep using the legacy one until we do https://redhat.atlassian.net/browse/COS-865. We can not unilaterally take the decision to switch, we need to help with docs, etc.

> > As this will never be used in FCOS, I think you should keep the current FCOS test as is and duplicate it in RHCOS instead. That way we will not have to undo the change here when we move away from the legacy one in RHCOS and we'll only have to delete it there.
> 
> But then any changes/additions will need to be made in 2 places even though they'll eventually both be using containers/toolbox right?
> 
> For example, I was thinking we should add at least a line in this test checking toolbox works when `$SHELL` is unset or not exported, like that issue we had on openshift nodes running rhcos 10.2.

Indeed, that's a good point, so let's keep everything here indeed.

tests/kola/toolbox/config.bu
Rolv-Apneseth: Alright, I can work on that after. Do you have any idea as to how I can get that kind of authentication from within the test?

Rolv-Apneseth:
I see, so 10.2 is meant to still be on coreos/toolbox. But alright then, so documentation and getting `sos-report` to work with `/run/host` are the 2 main things left after this test is merged?

travier:
Options are in https://redhat.atlassian.net/browse/COS-865?focusedCommentId=12594505. We could directly add sos-report to RHCOS and thus not require it to be in the toolbox.

tests/kola/toolbox/config.bu
Rolv-Apneseth: @dustymabe I moved both the config file and the auth file discussed below into the test itself

tests/kola/toolbox/test.sh
Rolv-Apneseth: @travier Shall I keep it using `machinectl` for FCOS? I'll just override the `run_as_core` function

tests/kola/toolbox/test.sh
```diff
@@ -1,52 +1,105 @@
...
+    cat << EOF > /var/home/core/.toolboxrc
+REGISTRY=quay.io
+IMAGE=fedora/fedora-toolbox:latest
+TOOLBOX_NAME=fedora-toolbox-latest
+AUTHFILE=/var/home/core/config.json
+EOF
+    # Toolbox fails if the AUTHFILE does not exist, but we don't actually need credentials
+    cat << EOF > /var/home/core/config.json
+{}
+EOF
```
dustymabe:
```suggestion
    echo '{}' > /var/home/core/config.json
```

tests/kola/toolbox/test.sh
dustymabe: my vote would be to use the same thing in both places. If we can add machinectl to RHCOS then use that. else let's not use it.

Rolv-Apneseth: I would tend to agree, but don't have a strong opinion

tests/kola/toolbox/test.sh
Rolv-Apneseth: Sure yeah

tests/kola/toolbox/test.sh
Rolv-Apneseth: @travier any opinion on this?

If we do keep it the same for both, should I move `run_as_core` into the common lib in this PR?

travier (dismissed):
LGTM. You can merge that if you've already checked that this works on RHCOS as we don't have that in CI here.

travier:
Maybe we should consider removing machinectl from FCOS.

travier:
> Maybe we should consider removing machinectl from FCOS.

OK it's part of systemd-container and that brings nspawn as well:
- https://github.com/coreos/fedora-coreos-tracker/issues/230
- https://github.com/coreos/fedora-coreos-config/pull/126

So that will be difficult to remove.

Rolv-Apneseth:
> LGTM. You can merge that if you've already checked that this works on RHCOS as we don't have that in CI here.

Just tried again and I'm getting an auth error now:

```
Error: failed to pull image registry.access.redhat.com/ubi10/toolbox:10.2
```

Also remembered we now have `creationDate` so I'll add that.

Rolv-Apneseth:
> Just tried again and I'm getting an auth error now:
> 
> ```
> Error: failed to pull image registry.access.redhat.com/ubi10/toolbox:10.2
> ```

Anybody know how to avoid this? Otherwise, just using the fedora image allows the test to pass:

```
cat << EOF > /etc/containers/toolbox.conf
[general]
image = "quay.io/fedora/fedora-toolbox:latest"
EOF
```

Rolv-Apneseth:
Confirmed that this works on RHCOS/SCOS by the way. I think it's good to go with the latest changes.

I can split out the `run_as_core` function into the common lib in a separate PR, or do you want it done here @dustymabe

dustymabe:
> > LGTM. You can merge that if you've already checked that this works on RHCOS as we don't have that in CI here.
> 
> Just tried again and I'm getting an auth error now:
> 
> ```
> Error: failed to pull image registry.access.redhat.com/ubi10/toolbox:10.2
> ```

so that repo shouldn't require auth IIUC.

ahh. it's because `10.2` isn't GA yet. 

Just use the `:latest` tag

`registry.access.redhat.com/ubi10/toolbox:latest`

dustymabe:
> I can split out the `run_as_core` function into the common lib in a separate PR, or do you want it done here @dustymabe

followup is fine.

Rolv-Apneseth:
> ahh. it's because `10.2` isn't GA yet.
> 
> Just use the `:latest` tag
> 
> `registry.access.redhat.com/ubi10/toolbox:latest`

Oh I see. OK I can update both paths to use that then probably. Will test it out

tests/kola/toolbox/test.sh
```diff
@@ -1,52 +1,110 @@
...
+        /bin/podman ps -a
+    }
+
+    toolbox_rm() {
+        toolbox="$(toolbox_list | awk '/(fedora|rhel)-toolbox-/ {print $1}')"
+        /bin/podman rm -f "${toolbox}"
+    }
+elif is_rhcos; then
+    # When using an image that's not yet released (e.g. 10.2 as of writing this),
+    # pulling from the registry  requires auth
```
dustymabe:
```suggestion
    # When using an image that's not yet released (e.g.
    # 10.2 as of writing this) the image won't exist in the
    # registry. Let's just use the `:latest` tag so we don't
    # hit that problem.
```

tests/kola/toolbox/test.sh
```diff
@@ -1,52 +1,110 @@
...
+    cat << EOF > /etc/containers/toolbox.conf
+[general]
+image = "registry.access.redhat.com/ubi$(get_rhel_maj_ver)/toolbox:latest"
```
dustymabe: preferably above you set a variable `rhel_major_version=$(get_rhel_maj_ver)` and then here:


```suggestion
image = "registry.access.redhat.com/ubi${rhel_major_version}/toolbox:latest"
```

the same would probably go for the if block above.

Rolv-Apneseth: Sure, will do

dustymabe (approved):

PR #4021, opened 2026-02-23
Skip acquiring the repo lock file

joelcapitao:
Since [1], libdnf stores a repo lock file in
`/usr/lib/sysimage/libdnf5/system-repo.lock` but as /usr is read-only, it fails.
We skip this operation in all the impacted dnf calls.

I'm editing the general dnf config instead of directly specifying the
`--skip-file-locks` option in order to keep backward compatibilty
with streams and downstream that do not have the new flag option.

Closes [2].

[1] https://github.com/rpm-software-management/dnf5/pull/2519
[2] https://github.com/coreos/fedora-coreos-tracker/issues/2113


jbtrystram (approved):
LGTM

PR #4022, opened 2026-02-23
extend denial of kdump.crash tests on s390x rawhide

nikita-dubrovskii:
RCs of kernel 6.20.0 have an issue where, after the kernel dump, qemu-system-s390x either hangs or terminates.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/2100


dustymabe (approved):
LGTM

PR #4024, opened 2026-02-23
.tekton: disable hermetic for branched stream

joelcapitao:
branched stream does not use manifest lockfiles, so building it hermetically is not possible as we derived from the manifest lockfiles to generate the Konflux ones.


dustymabe:
we are opending the PR against `testing-devel` because these files get synced over to the `branched` branch?

joelcapitao:
> we are opending the PR against testing-devel because these files get synced over to the branched branch?

yes, we centralize all the pipelines definitions in `testing-devel` and let the bot propagate to the other branches. We do so because we use a [shell script ](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/ci/generate-tekton-pipelinerun) to generate all the pipelines from the base one + the `kustomization.yaml` files for specific configuration for each branch.

dustymabe (approved):

PR #4026, opened 2026-02-24
overlay/container-signing: add F42, drop F45 key

dustymabe:
Rawhide is now F45 so we need to add that key to the policy. We can drop the F42 key since we are no longer building for F42.


marmijo (approved):
LGTM

PR #4030, opened 2026-02-26
konflux: integrate Testing Farm for multi-arch kola testing

Roshan-R:
Enable execution of kola tests in Konflux through Testing Farm using the upstream tmt-via-testing-farm Tekton pipeline.

This introduces two IntegrationTestScenario resources for x86_64 and aarch64, enabling Testing Farm to provision infrastructure and run kola via tmt. Other architectures are not yet enabled due to current Testing Farm provisioning limitations


tmt/tests/kola.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo $IMAGE_URL
+cosa import docker://$IMAGE_URL
+CONFIG_COMMIT=$(jq -r ".\"coreos-assembler.oci-imported-labels\".vcs-ref" ${COSA_DIR}/builds/latest/$(arch)/meta.json)
+pushd ${COSA_DIR}/src/config
+git checkout $CONFIG_COMMIT
+popd
+cosa buildextend-qemu
+cosa kola run --rerun --allow-rerun-success=tags=needs-internet --build=latest --on-warn-failure-exit-77 --arch=$(arch) '--tag=!reprovision' --parallel=5
+mv $COSA_DIR/tmp/kola $TMT_TEST_DATA
```
jbtrystram: we probably want to also test the ISO and build all the artifacts

.tekton/testing-devel/testing-farm/testing-farm.yaml
```diff
@@ -0,0 +1,65 @@
...
+    resolver: git
+  params:
+  - name: COMPOSE
+    value: Fedora-43
+  - name: ARCH
+    value: x86_64
+  - name: IMAGE_TAG
+    value: v3.16
+  - name: GIT_URL
+    value: https://github.com/Roshan-R/fedora-coreos-config
```
jbtrystram, line 21: Update to coreos org

.tekton/testing-devel/testing-farm/testing-farm.yaml
```diff
@@ -0,0 +1,65 @@
...
+    resolver: git
+  params:
+  - name: COMPOSE
+    value: Fedora-43
+  - name: ARCH
+    value: aarch64
+  - name: IMAGE_TAG
+    value: v3.16
+  - name: GIT_URL
+    value: https://github.com/Roshan-R/fedora-coreos-config
```
jbtrystram, line 56: update to coreOS org

tmt/plans/kola.fmf
```diff
@@ -0,0 +1,35 @@
...
+  - how: shell
+    script: podman pull quay.io/coreos-assembler/coreos-assembler:latest
+execute:
+  how: tmt
+
+/qemu:
+  summary: QEMU tests
+  tag: qemu
+  environment+:
+    PLATFORM: qemu
```
jbtrystram: Let's split that in `/build-qemu` and `/test-qemu` at least

tmt/tests/kola.sh
```diff
@@ -0,0 +1,31 @@
...
+set -euo pipefail
+
+cat <<EOF > $HOME/utils.sh
+export COSA_DIR=\$HOME/workspace/build;
+cosa ()
+{
+    env | grep --color=auto COREOS_ASSEMBLER;
+    local -r COREOS_ASSEMBLER_CONTAINER_LATEST="quay.io/coreos-assembler/coreos-assembler:latest";
+    set -x;
+    podman run --rm -ti --security-opt=label=disable --privileged --uidmap=1000:0:1 --uidmap=0:1:1000 --uidmap=1001:1001:64536 -v=\${COSA_DIR}:/srv/ --device=/dev/kvm --device=/dev/fuse --tmpfs=/tmp -v=/var/tmp:/var/tmp --name=cosa \${COREOS_ASSEMBLER_CONFIG_GIT:+-v=\$COREOS_ASSEMBLER_CONFIG_GIT:/srv/src/config/:ro} \${COREOS_ASSEMBLER_GIT:+-v=\$COREOS_ASSEMBLER_GIT/src/:/usr/lib/coreos-assembler/:ro} \${COREOS_ASSEMBLER_ADD_CERTS:+-v=/etc/pki/ca-trust:/etc/pki/ca-trust:ro} \${COREOS_ASSEMBLER_CONTAINER_RUNTIME_ARGS} \${COREOS_ASSEMBLER_CONTAINER:-\$COREOS_ASSEMBLER_CONTAINER_LATEST} "\$@";
```
jbtrystram: we can drop the customization env vars here

tmt/tests/kola.sh
```diff
@@ -0,0 +1,31 @@
...
+echo $TMT_TEST_DATA
+cosa init --force https://github.com/coreos/fedora-coreos-config
+
+echo $IMAGE_URL
+cosa import docker://$IMAGE_URL
+CONFIG_COMMIT=$(jq -r ".\"coreos-assembler.oci-imported-labels\".vcs-ref" ${COSA_DIR}/builds/latest/$(arch)/meta.json)
+pushd ${COSA_DIR}/src/config
+git checkout $CONFIG_COMMIT
+popd
+cosa buildextend-qemu
```
jbtrystram:
```suggestion
cosa osbuild qemu
```

jbtrystram (changes requested):
Initial review

tmt/tests/kola.fmf
```diff
@@ -0,0 +1,31 @@
...
+  environment:
+    TEST_CASE: build-qemu
+
+/test-qemu:
+  duration: 2h
+  order: 80
+  environment:
+    TEST_CASE: test-qemu
+
+/build-iso:
```
HuijingHei, line 27: Maybe we can put build-iso and test-iso in another VM to make the running time shorter?

Roshan-R: yes, that seems to be a good idea to implement. Let me look at how to do this, we probably would have to refactor a bit

Roshan-R: I have added a commit to enable this, let's see if it works

Roshan-R: looks like everything passed: https://artifacts.dev.testing-farm.io/072a5f10-68de-4bf3-97eb-ce926acb72d5/

tmt/plans/main.fmf
```diff
@@ -0,0 +1,32 @@
...
+      - libvirt
+      - qemu-kvm
+      - qemu-img
+      - skopeo
+      - virt-install
+      - wget
+      - jq
+      - git
+  - how: shell
+    script: podman pull quay.io/coreos-assembler/coreos-assembler:latest
```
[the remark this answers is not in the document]
HuijingHei: Agree with this.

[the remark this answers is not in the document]
Roshan-R: I have made use of plan inheritance and merged it to a single main.fmf file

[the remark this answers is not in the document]
Roshan-R: the new single file plan works as intended

tmt/tests/test-iso.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -euo pipefail
+set -x
+
+source $HOME/utils.sh
+trap collect_kola_artifacts EXIT
+
+KOLA_ACTION="testiso"
+KOLA_ID="testiso"
```
joelcapitao: Let's reuse the same name as we have currently in Jenkins: `kola-testiso`

tmt/tests/test-qemu.sh
joelcapitao, line 1: For QEMU, there are three tests ran sequentially:

1. kola
2. kola-reprovision
3. kola-upgrade

The first one is covered here, but not the others. I'm not against adding the remaining ones in a follow-up patch, but I think we need to structure this now to support multiple runs e.g: `trap collect_kola_artifacts EXIT` only works for one `KOLA_ID`.

.tekton/testing-devel/testing-farm/testing-farm.yaml
```diff
@@ -0,0 +1,65 @@
...
+metadata:
+  labels:
+    test.appstudio.openshift.io/optional: "true"
+  name: fedora-coreos-kola-tests-testing-devel-aarch64
+spec:
+  application: fedora-coreos-testing-devel
+  contexts:
+  - name: component_fedora-coreos-testing-devel
+  params:
+  - name: COMPOSE
```
joelcapitao: Why do we need to specify a compose value ? is it for the underlying VM ?

.tekton/testing-devel/testing-farm/testing-farm.yaml
```diff
@@ -0,0 +1,65 @@
...
+    value: Fedora-43
```
joelcapitao: I' surprised the default value is `Rawhide` c.f https://gitlab.com/testing-farm/integrations-konflux/-/blob/v3.16/task/testing-farm-konflux-scheduler/testing-farm-konflux-scheduler.yaml?ref_type=tags#L50.
Setting the value to `43` will bring more maintenance work, I'd go with the default value and see how it goes, wdyt ?

.tekton/testing-devel/testing-farm/testing-farm.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - name: GIT_REF
+    value: testing-devel
+  - name: TMT_PLAN
+    value: tmt
+  resolverRef:
+    params:
+    - name: url
+      value: https://gitlab.com/testing-farm/integrations-konflux
+    - name: revision
+      value: v3.16
```
joelcapitao, line 31: As a follow-up patch, we would need to configure MintMaker to update this tag on a weekly/monthly basis

tmt/plans/main.fmf
```diff
@@ -0,0 +1,44 @@
...
+        cpu:
+          processors: ">= 4"
+        memory: ">= 8 GB"
+        disk:
+          - size: ">= 50 GB"
+        virtualization:
+          is-supported: true
+prepare:
+  - how: install
+    package:
```
joelcapitao, line 14: Looks like `wget` and `skopeo` are not required.
Additionally, `git-core` is a better fit in CI/CD than the full `git` RPM, as we only need the `git` binary here. This avoids pulling all the Perl libs.

tmt/tests/test.sh
joelcapitao, line 1: The name of this file i.e: `test.sh` is confusing because there is `build.sh` which handles both `qemu` and `iso`, so we expect `test.sh` to achieve the same for the testing part. 
But that file acts as a bash entrypoint that routes to different scripts based on `TEST_CASE` envvar.
I would rename it to something like `dispatcher.sh` or  `entrypoint.sh`, but I'm open to other suggestions.

That way we could merge `test-qemu.sh` and `test-iso.sh` into one `test.sh` file and keep consistency with `build.sh`.
Overall, this would be temporary because we will remove all the building part once Konflux takes over the disks generation in the build pipeline.

.tekton/base/testing-farm/testing-farm.yaml
```diff
@@ -0,0 +1,66 @@
...
+    resolver: git
+  params:
+  - name: COMPOSE
+    value: Fedora-43
+  - name: ARCH
+    value: x86_64
+  - name: IMAGE_TAG
+    value: v3.16
+  - name: GIT_URL
+    value: https://github.com/coreos/fedora-coreos-config
```
joelcapitao: We should define `https://github.com/coreos/fedora-coreos-pipeline` here and propose in a PR the whole `tmt/` directory there.

joelcapitao (changes requested):
I left some comments but this PR should be splitted into 2 parts and proposed to other repos:

1. the whole `.tekton/` content should be hosted in the tenant repository alongside the other resources in https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/blob/main/cluster/kfluxfedorap01/coreos-tenant/fedora-coreos/template.yaml?ref_type=heads. All Konflux resources other than Tekton pipelines must be defined in the tenant repository.
2. the whole `tmt/` and `.fmf/` content should be hosted in the fedora-coreos-pipeline, the same way we do with Jenkins.

Also, I think we can already add all the supported archs.

tmt/tests/test-qemu.sh
Roshan-R: If I understand it correctly, the tests should be run sequentially. that would be kola -> kola-reprovision -> kola-upgrade. how would be handle let's say if kola-reprovision fails? should we stop the execution chain and thereby skip kola-upgrade? if so the new implementation  I worked on should be enough to do

joelcapitao: Actually, the kola:upgrade (~3 min to run) can be run in parallel. It might be good to run in that one apart.

Roshan-R: @joelcapitao If we want to parallelize this, we’ll likely need to introduce a new VM, similar to how we separated kola-qemu and kola-testiso. I haven’t found another viable way to achieve this.

If we go down that route, we’d need to define a new plan, something along these lines:

```
/kola-test-upgrade:
  summary: run kola qemu tests
  discover+:
    how: fmf
    test:
      - init
      - build-qemu
      - test-kola-upgrade
```
One thing to keep in mind is that this will likely increase resource usage quite a bit, so we’d want to plan for that as well.

joelcapitao: I think it's ok to provision another small VM for that purpose.
The idea is to have feedback as soon as we can for tests that are not dependent on each others.

tmt/tests/test.sh
Roshan-R: I have merged all the tests to a single test.sh file

joelcapitao: Nice, thank you Roshan. Could you now split this PR into 2 part and send them to https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/tree/main/cluster/kfluxfedorap01/coreos-tenant/fedora-coreos and https://github.com/coreos/fedora-coreos-pipeline accordingly please ? I'll then make the final review

.tekton/base/testing-farm/testing-farm.yaml
HuijingHei: I am not sure why we need `fedora-coreos-pipeline` here, could you help to clarify?

HuijingHei:
Sorry that I merge https://github.com/coreos/fedora-coreos-config/pull/4087 before this PR, maybe will resolve the conflicts firstly, will look at that.

.tekton/base/testing-farm/testing-farm.yaml
joelcapitao: The test definitions located in `tmt/` are both used for CI _and_ CD pipelines. As we'll share the same code with rhel-coreos-config repo, we better need to commit it to another repo than f-c-c. fedora-coreos-pipeline repo was meant for this purpose i.e: sharing same pipeline definitions across Jenkins instance.
Technically, we could host the testing part in f-c-c and r-c-c would inherit it, but that would make the maintenance harder IMHO.
As an example, the Fedora bootc community is using https://gitlab.com/fedora/bootc/tests/bootc-workflow-test/-/tree/main?ref_type=heads for both Fedora and CentOS Stream distro

joelcapitao:
> Sorry that I merge #4087 before this PR, maybe will resolve the conflicts firstly, will look at that.

It's ok as the tekton code should be moved to [gitlab.com/fedora/infrastructure/konflux/tenants-config](https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/blob/main/cluster/kfluxfedorap01/coreos-tenant/fedora-coreos/template.yaml?ref_type=heads) c.f: https://github.com/coreos/fedora-coreos-config/pull/4030#pullrequestreview-4049783600

.tekton/base/testing-farm/testing-farm.yaml
HuijingHei: Got it, thank you!

tmt/tests/test-qemu.sh
Roshan-R: added a new parallel test for kola:upgrade

.tekton/testing-devel/testing-farm/testing-farm.yaml
Roshan-R: I remember explicitly changing it to `Fedora-43` due to something, I forgot why. let me try changing it to `Fedora-Rawhide` and see if the tests pass.

Roshan-R: all tests have passed, I have changed it to rawhide :)

Roshan-R:
> Also, I think we can already add all the supported archs.

@joelcapitao I am currently using the public testing-farm ranch right now, that's why the other arches did not work. According to the docs, the redhat ranch should have the other arches too but I can't test that without merging this code to a coreos owned repo as it blocks external repos like my fork

joelcapitao:
> > Also, I think we can already add all the supported archs.
> 
> @joelcapitao I am currently using the public testing-farm ranch right now, that's why the other arches did not work. According to the docs, the redhat ranch should have the other arches too but I can't test that without merging this code to a coreos owned repo as it blocks external repos like my fork

Ok, then let's add them in a follow-up patch

thrix:
@Roshan-R yes, RH ranch acccess was locked down and this PR came up in our monitoring:

https://docs.testing-farm.io/Testing%20Farm/0.1/releases/2026-03.3.html#_breaking_changes

Anyway, we can allow your fork as you a RH employee so you can test on this PR, just send an MR pls

Roshan-R:
> @Roshan-R yes, RH ranch acccess was locked down and this PR came up in our monitoring:
> 
> https://docs.testing-farm.io/Testing%20Farm/0.1/releases/2026-03.3.html#_breaking_changes
> 
> Anyway, we can allow your fork as you a RH employee so you can test on this PR, just send an MR pls

thanks for the info, I have created a MR to the repo: https://gitlab.com/testing-farm/config/-/merge_requests/22
Since we are going to change this PR to the fedora-coreos-config repo, I have added that personal fork for the allowlist

Roshan-R:
PR moved to https://github.com/coreos/fedora-coreos-pipeline/pull/1334

PR #4038, opened 2026-03-02
tests/alternatives: enable using fedora-coreos-pool for overlaying RPM

dustymabe:
We occasionally hit an issue where rawhide becomes out of date for a few days (CI failures, etc) and the alternatives test will start failing if there is a new iptables package that shipped in the meantime.

Let's convert this test to be able to grab the right version of iptables from the coreos-pool, since that repo definitely has the all the subpackages of the iptables that was baked into the last `rawhide` build.


yasminvalim (approved):
LGTM :)

PR #4041, opened 2026-03-04
overrides: drop podman pin

dustymabe:
The experiment in `next` hasn't had any reports of failures so I'd say we can drop the pin here and when the package reaches bodhi stable it will flow into CoreOS.


marmijo (approved):
LGTM

PR #4042, opened 2026-03-04
kola: bump minimum memory for alternatives test

dustymabe:
The alternatives test is now failing because rpm-ostree gets OOM killed because we're now including the coreos-pool repo when layering iptables-legacy [1]. Let's just bump the minimum required memory for now.

[1] https://github.com/coreos/fedora-coreos-config/pull/4038


marmijo (approved):
LGTM

PR #4043, opened 2026-03-05
next-devel is now Fedora Linux 44

marmijo:
Fedora 44 beta is going to ship soon, so let's move next-devel over to Fedora 44 content so we can ship it in `next` alongside the beta.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2055


dustymabe:
for the CI failure.. can you just copy in the lockfiles from the  `branched` branch?

dustymabe:
or actually - we might as well just merge this and then kick off a bump-lockfile job immediately.

dustymabe (approved):
LGTM

PR #4046, opened 2026-03-06
overlay/core: write ignition issues to /run

jbtrystram:
Since util-linux 2.41[1] agetty will source issues from multiples places by default (` /etc`, `/run/` and  `/usr/lib`).

Systemd also worked around [2] this since v258 which has been in fedora 43 for a while now, so let's always write our issues to `/run`.

[1]
util-linux/util-linux@508fb0e
[2] systemd/systemd@864a5e9

Ref coreos/fedora-coreos-tracker#2108
See https://github.com/coreos/console-login-helper-messages/pull/130


jbtrystram:
So this will break on all the RHCOS versions we currently build because `util-linux` and `systemd` are both too old there. 
I'll add back the symlink under a rhel condition.

overlay.d/05core/usr/libexec/coreos-ignition-write-issues
```diff
@@ -66,6 +66,8 @@ else
...
 fi
 
-# Our makeshift way of getting /run/issue.d semantics. See:
-# https://github.com/coreos/console-login-helper-messages/blob/e06fc88ae8fbcc3a422bc8c686f70c15aebb9d9a/usr/lib/console-login-helper-messages/issue.defs#L8-L17
-ln -sf /run/issue.d/30_coreos_ignition_provisioning.issue /etc/issue.d/
+# Until we have `util-linux-2.41` in RHCOS, we need to have the issues in /etc
+source /etc/os-release
+if [ "${ID}" == "rhel" ]; then
+    ln -sf /run/issue.d/30_coreos_ignition_*.issue /etc/issue.d/
+fi
```
dustymabe: I wonder if this would be better served as a conditional include in the rhel-coreos-config. I'd add it (with a nice comment explanation) and make it conditional on el9/el10 so that when we move to el11 it won't be applied.

Leaving it here would work too, but I'd really like to make it stop applying when we move to 11 so that it's easy for us to spot and remove this in the future.

jbtrystram: Ah yes, that would be better indeed!

jbtrystram: Alright so I bolied it down to a simple drop-in: 
```
cat overlay.d/05core/usr/lib/systemd/system/coreos-ignition-write-issues.service.d/rhel-issue-compat-symlinks.conf
# Until RHEL 11 util-linux won't source  issue files from /run/issue.d/.
# Create symlinks from /etc pointing to /run/issue.d/ so agetty picks them up.
# This can be removed when we move to EL11
# See https://github.com/coreos/fedora-coreos-tracker/issues/2108
[Service]
ExecStartPost=/usr/bin/bash -c "mkdir -p /etc/issue.d && ln -sf /run/issue.d/30_coreos_ignition_*.issue /etc/issue.d/"
```

Now the question I have is how do we ship that in RHCOS. A separate overlay or a postprocess ?

dustymabe: There are so many different ways to do this. I've thought of 2 or 3 just sitting here.

I'm thinking maybe a postprocess that doesn't actually check el9/10/11 but rather checks the util-linux rpm version and then creates the symlink based on that. i.e. the symlink gets created at compose time (in a postprocess) rather than as part of the systemd service run - or maybe that won't work for some reason?

jbtrystram: > i.e. the symlink gets created at compose time (in a postprocess) rather than as part of the systemd service run - or maybe that won't work for some reason?

This would create a symlink for the whole `/etc/issue.d` directory to `/run`, removing the ability to create persistent issues. That would be a big regression.
What I am suggesting is to wait for `coreos-ignition-write-issues.service` to have run and symlink the individually created issues after.

jbtrystram: https://github.com/coreos/rhel-coreos-config/pull/215

dustymabe (approved):
LGTM

PR #4048, opened 2026-03-06
[next-devel] overrides: fast-tracks for next-devel

dustymabe:
- drop f43 overrides
- fast track any packages that would be downgraded when going from testing-devel to next-devel because of our beta freeze period for F44
- add in ignition/afterburn fast-tracks for https://github.com/coreos/fedora-coreos-config/pull/3885#issuecomment-4013155688


marmijo (approved):
LGTM

PR #4050, opened 2026-03-08
Revert "overrides: drop podman pin"

dustymabe:
This reverts commit 31a82c04fc64bc510308e9c0079bca059f9bdf07.

There's a report of some bugs so we'll re-add the pin for now.

https://github.com/coreos/fedora-coreos-tracker/issues/2110#issuecomment-4017324689


Luap99 (commented):
Thanks, LGTM

PR #4051, opened 2026-03-10
COS-4032: Add numad and numactl packages

angelcerveraroldan:
Add `numad` and `numactl` to `x86_64`, `aarch64`, and `ppc64le`. `numad` is not available for `s390x`, so we will not install the packages in that architecture.

### DIFF
--- a/manifests/system-configuration.yaml
+++ b/manifests/system-configuration.yaml
@@ -34,6 +34,17 @@ packages:
   # https://github.com/coreos/fedora-coreos-tracker/issues/753
   - stalld
 
+conditional-include:
+  # numad is not available for s390x
+  - if: basearch != "s390x"
+    include:
+      packages:
+        # https://github.com/coreos/fedora-coreos-tracker/issues/2096
+        # numad: Daemon that provides placement advice for efficient use of CPUs and memory on systems with NUMA topology.
+        # numactl: Control NUMA policy for processes or shared memory
+        - numad
+        - numactl
+
 postprocess:
   # Mask systemd-repart. Ignition is responsible for partition setup on first
   # boot and does not use systemd-repart currently. See also
--- a/tests/kola/numad/config.bu
+++ b/tests/kola/numad/config.bu
@@ -1,0 +1,6 @@
+variant: fcos
+version: 1.6.0
+systemd:
+  units:
+    - name: numad.service
+      enabled: true
--- a/tests/kola/numad/test.sh
+++ b/tests/kola/numad/test.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+## kola:
+##   platforms: qemu
+##   numaNodes: true
+##   minMemory: 2048
+##   architectures: "!s390x"
+##   description: Verify that numad detects nodes and tracks set -euo pipefail
+##   bindMountHostRO: ["/,/var/cosaroot"]
+##   creationDate: 2026-04-09
+
+set -euo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+if [[ $(systemctl show numad -p ActiveState) != "ActiveState=active" ]]; then
+    fatal "numad did not activate"
+fi
+
+# Call numad to change the already running daemon's settings for -l and -i
+# `-l 7` adds extra information to the logs
+# `-i 5` changes interval from every 15s to every 5s
+numad -l7 -i5
+
+if ! lscpu | grep -Eq "NUMA node\(s\):\s*2"; then
+    fatal "expected to find exactly 2 numa nodes"
+fi
+
+# As part of the test we want to run a somewhat intensive process, so that
+# we can verify that numad is successfully tracking processes. Here we
+# use the same pattern of using a mounted in COSA as the container root as:
+# https://github.com/coreos/coreos-assembler/blob/8dbfe3ea8b8f571e732e8cc0ab307e983a0be1f3/mantle/cmd/kola/resources/iscsi_butane_setup.yaml#L102-L113
+podman run --privileged --name stress-ng --pid=host                      \
+    --volume=root:/root/:nocopy --volume=vartmp:/var/tmp/:nocopy         \
+    --workdir /root --rootfs /var/cosaroot                               \
+    stress-ng --temp-path /var/tmp --vm 1 --vm-bytes 1024M --timeout 25s
+
+logfile="/var/log/numad.log"
+for node in 0 1; do
+	# Different versions of numad have a slightly different format for the log file,
+	# e.g. MBs_total vs MBs_tot. This pattern should match both versions.
+	if ! grep -Eq "Node.${node}.*MBs_tot(al)?.*CPUs_tot(al)?" "$logfile"; then
+		fatal "Numad didn't detect Node ${node}"
+	fi
+done
+
+# Check that the stress test was being monitored by numad
+if ! grep -q "stress-ng-vm" "$logfile"; then
+    fatal "Numad is not monitoring the stress test"
+fi
+
+ok "Numad working as expected"
--- a/manifest-lock.aarch64.json
+++ b/manifest-lock.aarch64.json
@@ -987,8 +987,14 @@
     "nss-altfiles": {
       "evra": "2.23.0-7.fc43.aarch64"
     },
+    "numactl": {
+      "evra": "2.0.19-3.fc43.aarch64"
+    },
     "numactl-libs": {
       "evra": "2.0.19-3.fc43.aarch64"
+    },
+    "numad": {
+      "evra": "0.5-47.20150602git.fc43.aarch64"
     },
     "nvidia-gpu-firmware": {
       "evra": "20260221-1.fc43.noarch"
--- a/manifest-lock.ppc64le.json
+++ b/manifest-lock.ppc64le.json
@@ -969,8 +969,14 @@
     "nss-altfiles": {
       "evra": "2.23.0-7.fc43.ppc64le"
     },
+    "numactl": {
+      "evra": "2.0.19-3.fc43.ppc64le"
+    },
     "numactl-libs": {
       "evra": "2.0.19-3.fc43.ppc64le"
+    },
+    "numad": {
+      "evra": "0.5-47.20150602git.fc43.ppc64le"
     },
     "nvidia-gpu-firmware": {
       "evra": "20260221-1.fc43.noarch"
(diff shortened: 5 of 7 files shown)


angelcerveraroldan:
Tests seems to be failing due to a kernel warning:

```
arch/x86/kernel/smpboot.c:332 topology_sane.isra.0+0xc5/0x200
```

The error may or may not show up depending on the host CPU. 

Depends on https://github.com/coreos/coreos-assembler/pull/4479 (which has now been merged)

tests/kola/numad/config.bu
```diff
@@ -0,0 +1,14 @@
+variant: fcos
+version: 1.6.0
+systemd:
+  units:
+    - name: numad.service
```
dustymabe, line 5: hmm. is there not a numad.service provided by the RPMs ?

tests/kola/numad/test.sh
```diff
@@ -0,0 +1,44 @@
...
+fi
+
+if ! lscpu | grep -Eq "NUMA node\(s\):\s*2"; then
+    fatal "expected to find exactly 2 numa nodes"
+fi
+
+# run a process and check that the daemon is actively monitoring it
+podman run --privileged \
+       --name pod-stress-ng --pid=host \
+       ghcr.io/colinianking/stress-ng \
```
dustymabe: Pulling from a namespace/registry we don't control is not something we typically do. 

`stress-ng` is [in fedora](https://packages.fedoraproject.org/pkgs/stress-ng/stress-ng/) so we can just build [our own container](https://github.com/coreos/coreos-assembler/tree/main/tests/containers) for it, but TBH it would be better to get rid of most of those containers and do something like https://github.com/coreos/coreos-assembler/commit/8dbfe3ea8b8f571e732e8cc0ab307e983a0be1f3 in more places (at least for tests that can only run on `qemu` platform).

tests/kola/numad/config.bu
angelcerveraroldan: It does come with a `numad.service`, but I wrote a custom one so that it launches with some extra flags that add more logs, which are then grepped in the test.

dustymabe: I see. The `-l 7` is important here for the logging, but what about the others?

I guess what I'd like to see is us use the default unit as much as possible.

It looks like the [default unit](https://pagure.io/numad/blob/master/f/numad.service) has:

```
[Unit]
Description=numad - The NUMA daemon that manages application locality.
Documentation=man:numad

[Service]
Type=forking
ExecStart=/usr/bin/numad -i 15

[Install]
WantedBy=multi-user.target
```

Are we required to change the `-i` and `-m` values for our test to work?

As for the `-l 7` value I wish numad supported setting an evironment variable for this, but right now [no luck](https://pagure.io/numad/blob/master/f/numad.8#_221). But the man page [does say](https://pagure.io/numad/blob/master/f/numad.8#_231):

```
Subsequent numad invocations while the daemon is running can be used to dynamically change most run-time options.
```

So I'm wondering if we just add a `numad -l 7` call to our `test.sh` here will give us the extra debug logging we want.

angelcerveraroldan: The `-m 100` could likely be removed without any issues.

I think `-i 5` may be a good idea to avoid flakiness. That is, if we run the `stress-ng` command for 20 seconds, with the default `-i 15`, numad may only do a single system scan during those 20 seconds.

But if we are trying to avoid using a custom .system file, I am very happy to move set the parameters in the `test.sh` file.

dustymabe: > That is, if we run the stress-ng command for 20 seconds, with the default -i 15, numad may only do a single system scan during those 20 seconds.

Could we run the stress-ng command for longer than 20 seconds and reduce the flakiness that way?

angelcerveraroldan: Yes, if running it for longer is acceptable, that would work just as well. It may even be the case that 20 second with `-i 15` would be fine, I'll check that too, I was just trying to make numad as "sensitive" as possible.

dustymabe: :+1:

tests/kola/numad/test.sh
angelcerveraroldan: I have modidfied the test to use `rpm-ostree install --apply-live stress-ng` instead of pulling the container, would this be better?

I have tested it on x86 (locally, CI in this PR) and ARM (debug-pod) with this change and it seems to work. I still need to test ppc64le.

dustymabe: > I have modidfied the test to use `rpm-ostree install --apply-live stress-ng` instead of pulling the container, would this be better?

not exactly. `rpm-ostree install` won't work on rhel-coreos where there aren't any yum repos configured by default.

I have an idea for how to implement https://github.com/coreos/coreos-assembler/commit/8dbfe3ea8b8f571e732e8cc0ab307e983a0be1f3 more generically and use it here, but I think I'm obligated to review https://github.com/coreos/coreos-assembler/pull/4377 and help nikita get that merged first before I make any more changes to the code base there. Let's sync early next week to see if that has merged or not.

manifests/system-configuration.yaml
```diff
@@ -34,6 +34,16 @@ packages:
   # https://github.com/coreos/fedora-coreos-tracker/issues/753
   - stalld
 
+conditional-include:
+  # numad is not available for s390x after el8
```
dustymabe: probably OK without the `after el9` part 

```suggestion
  # numad is not available for s390x
```

manifests/system-configuration.yaml
```diff
@@ -34,6 +34,16 @@ packages:
...
+  - if: basearch != "s390x"
+    include:
+      packages:
+        # Daemon that provides placement advice for efficient use of CPUs and memory on systems with NUMA topology.
+        - numad
+        # Control NUMA policy for processes or shared memory
+        - numactl
```
dustymabe: link to the tracker issue ticket and also group the comments together so it's clear both RPMs are related to that tracker issue ticket.
```suggestion
        # https://github.com/coreos/fedora-coreos-tracker/issues/2096
        #  - numad: Daemon that provides placement advice for efficient use of CPUs and memory on systems with NUMA topology.
        #  - numactl: Control NUMA policy for processes or shared memory
        - numad numactl
```

tests/kola/numad/test.sh
```diff
@@ -0,0 +1,50 @@
...
+set -euo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+if [[ $(systemctl show numad -p ActiveState) != "ActiveState=active" ]]; then
+    fatal "numad did not activate"
+fi
+
+# Add extra flags to (already running) numad
```
dustymabe: Because this behavior isn't typical where you can call a binary and it will change the settings of another already running process I think we should call it out in the comment:

```suggestion
# Call numad to change the already running daemon's settings for -l and -i.
```

tests/kola/numad/test.sh
angelcerveraroldan: Sounds good, thanks.

dustymabe:
@angelcerveraroldan with https://github.com/coreos/coreos-assembler/pull/4513 I was able to test this with:

```patch
diff --git a/tests/kola/numad/test.sh b/tests/kola/numad/test.sh
index 5b858a80..d8e008f9 100755
--- a/tests/kola/numad/test.sh
+++ b/tests/kola/numad/test.sh
@@ -7,6 +7,8 @@
 ##   minMemory: 2048
 ##   architectures: "!s390x"
 ##   description: Verify that numad detects nodes and tracks set -euo pipefail
+##   # We use the COSA filesystem to start a container to run stress-ng
+##   bindMountHostRO: ["/,/var/cosaroot"]
 
 set -euo pipefail
 
@@ -27,11 +29,13 @@ if ! lscpu | grep -Eq "NUMA node\(s\):\s*2"; then
 fi
 
 # As part of the test we want to run a somewhat intensive process, so that
-# we can verify that numad is successfully tracking processes.
-rpm-ostree install --apply-live stress-ng
-stress-ng --temp-path /var/tmp         \
-                 --vm 1 --vm-bytes 1024M       \
-                 --timeout 25s
+# we can verify that numad is successfully tracking processes. Here we
+# use the same pattern of using a mounted in COSA as the container root as:
+# https://github.com/coreos/coreos-assembler/blob/8dbfe3ea8b8f571e732e8cc0ab307e983a0be1f3/mantle/cmd/kola/resources/iscsi_butane_setup.yaml#L102-L113
+podman run --privileged --name stress-ng --pid=host              \
+    --volume=root:/root/:nocopy --volume=vartmp:/var/tmp/:nocopy \
+    --workdir /root --rootfs /var/cosaroot                       \
+        stress-ng --temp-path /var/tmp --vm 1 --vm-bytes 1024M --timeout 25s
 
 logfile="/var/log/numad.log"
 for node in 0 1; do
```

but I will note it failed a few times. I'm not sure if it's because the test is flaky or if running the container this way is causing it to be flaky.

angelcerveraroldan:
> but I will note it failed a few times. I'm not sure if it's because the test is flaky or if running the container this way is causing it to be flaky.

@dustymabe, I dont remeber it being flaky. I have ran the test (just using the old method, where I download the image from within the test) another 80 times in the aarch64 pod, and none of them failed.

manifests/system-configuration.yaml
```diff
@@ -34,6 +34,17 @@ packages:
...
   - stalld
 
+conditional-include:
+  # numad is not available for s390x
+  - if: basearch != "s390x"
+    include:
+      packages:
+        # https://github.com/coreos/fedora-coreos-tracker/issues/2096
+        # numad: Daemon that provides placement advice for efficient use of CPUs and memory on systems with NUMA topology.
+        # numadct: Control NUMA policy for processes or shared memory
```
dustymabe:
```suggestion
        # numactl: Control NUMA policy for processes or shared memory
```

dustymabe (dismissed):

dustymabe:
I disabled automerge. @angelcerveraroldan feel free to merge when it passes tests if you want.

I wasn't sure if you wanted to wait for and use the potential change in https://github.com/coreos/coreos-assembler/pull/4518 for this new test.

angelcerveraroldan:
> I wasn't sure if you wanted to wait for and use the potential change in https://github.com/coreos/coreos-assembler/pull/4518 for this new test.

I did want to wait for that change to merge first, that way if the flakiness turns out to be an issue, it won't be as big an issue.

dustymabe (approved):
LGTM

PR #4055, opened 2026-03-14
denylist: snooze *kdump.crash* in next-devel

marmijo:
These tests were snoozed in rawhide but somehow they were not running in next-devel either. The tests have been failing for, probably, the same reason as https://github.com/coreos/fedora-coreos-tracker/issues/2100. Let's re-snooze them in next-devel while we investigate.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -8,6 +8,14 @@
 ##  arches:
 ##    - aarch64
 
+- pattern: '*kdump.crash*'
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2100
+  snooze: 2026-03-31
+  arches:
+    - s390x
+  streams:
+    - next-devel
+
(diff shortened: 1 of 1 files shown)


dustymabe:
> probably, the same reason as https://github.com/coreos/rhel-coreos-config/issues/132

you mean https://github.com/coreos/fedora-coreos-tracker/issues/2100 ?

dustymabe (approved):
LGTM

marmijo:
> you mean https://github.com/coreos/fedora-coreos-tracker/issues/2100 ?

Yes, thanks! Updated.

nikita-dubrovskii:
Yep, kernel `6.19.6` has no `1708e7627881 s390/kexec: Disable stack protector in s390_reset_system()` . But `6.19.7` , which is now available in repos, comes with the fix.

PR #4060, opened 2026-03-16
[next-devel] overrides: fast-track podman-5.8.1-1.fc44

marmijo:
We're including podman v5.8.1 in testing-devel/testing`[1]`, so this package would show as downgraded in next-devel/next due to the beta freeze in F44. Fast-tracking this version to next-devel to avoid any downgrades.

`[1]`: https://github.com/coreos/fedora-coreos-config/pull/4059


dustymabe (approved):
LGTM

PR #4064, opened 2026-03-17
[next-devel] overrides: fast track libnetfilter_conntrack to avoid downgrade

dustymabe:
For libnetfilter_conntrack we are using the F43 version of the package because it was never built for F44 yet. Probably an oversight by the maintainer.


joelcapitao (approved):
LGTM

AdamWill:
[Building for F44 now](https://koji.fedoraproject.org/koji/taskinfo?taskID=143455824) at Dusty's request. Will submit an update.

PR #4078, opened 2026-03-25
tests/validate-symlinks: add exception for libmtcr_ul.a

nikita-dubrovskii:
Recently mstflint.rpm was added for el9 and el10, so this test now fails on broken /usr/lib64/libmtcr_ul.a:
```
Error: /usr/lib64/libmtcr_ul.a symlink to /usr/lib64/mstflint/libmtcr_ul.a which does not exist
```


nikita-dubrovskii:
https://redhat.atlassian.net/browse/RHEL-160059

nikita-dubrovskii:
Workaround - https://github.com/coreos/rhel-coreos-config/pull/220

PR #4093, opened 2026-04-02
overlays: create 10bootc

jbtrystram:
Introduce a new overlay to ship configuration files for bootc and image-builder. These file are sourced from the container during `bootc install to-filesystem`.
We can also use this later to ship other bits as we make the container more and more the source of truth, e.g. the partition table definition.

This is prep work for [1]

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1827 

See also https://github.com/coreos/coreos-assembler/pull/4224

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -15,6 +15,7 @@ ostree-layers:
   - overlay/05core
   - overlay/08nouveau
   - overlay/09misc
+  - overlay/10bootc
   - overlay/20platform-chrony
   - overlay/30lvmdevices
 
--- a/overlay.d/10bootc/usr/lib/bootc/install/10-grub-bls-append.toml
+++ b/overlay.d/10bootc/usr/lib/bootc/install/10-grub-bls-append.toml
@@ -1,0 +1,2 @@
+[install.ostree]
+bls-append-except-default = 'grub_users=""'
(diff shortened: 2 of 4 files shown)


joelcapitao (dismissed):
I guess we will move those config files to `05core` at some point ?

jbtrystram:
Just to note : this have no effect until we move to image builder, which have an integrated step to copy out config files from the container when building from a different buildroot.
In our case, COSA is the buildroot, so as long as we use `bootc install` manually rather than image builder we have to carry those files in COSA.

Still, it's good practice to ship them now, as this reflects the "coreOS defaults", if someone were to derive FCOS then use image-builder, they would inherit them.

dustymabe:
With https://github.com/coreos/coreos-assembler/pull/4534 (i.e. now we're using FCOS as the buildroot) this PR becomes load bearing in the bootc install to filesystem path. 

Any reason not to merge it?

dustymabe (approved):

jbtrystram:
> With https://github.com/coreos/coreos-assembler/pull/4534 (i.e. now we're using FCOS as the buildroot) this PR becomes load bearing in the bootc install to filesystem path. 
> 
> Any reason not to merge it? 

Bootc does not support the sigpolicy install config file yet. So if you go ahead with the python layer that will break the bootc install path.

PR #4096, opened 2026-04-07
tests: add test to verify updating multiple EFIs using RAID 1 works

HuijingHei:
Copy from https://github.com/coreos/bootupd/tree/main/tests/kola/raid1 to test on both RHCOS / FCOS.


travier (approved):

yasminvalim (approved):
LGTM

PR #4097, opened 2026-04-08
tests: add `raid1-boot` config with `layout`

HuijingHei:
The test failed on aarch64 / ppc64le because of missing `layout`.

Requires https://github.com/coreos/coreos-assembler/pull/4525

### DIFF
--- a/tests/kola/boot/raid1-boot/config.bu
+++ b/tests/kola/boot/raid1-boot/config.bu
@@ -1,6 +1,8 @@
 variant: fcos
 version: 1.5.0
 boot_device:
+  # This will be replaced by https://github.com/coreos/coreos-assembler/pull/4525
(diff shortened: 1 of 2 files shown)


tests/kola/boot/raid1-boot/config.bu
```diff
@@ -1,6 +1,7 @@
 variant: fcos
 version: 1.5.0
 boot_device:
+  layout: {{LAYOUT}}
```
dustymabe, line 5: I think we should comment here where this gets substituted (i.e. the path to the file in COSA git repo) so we know it's specific to `kola`.

dustymabe (approved):
LGTM

HuijingHei:
Hold on, need to merge https://github.com/coreos/coreos-assembler/pull/4525 firstly.

PR #4099, opened 2026-04-09
tests: verify the deployed container imgref

jbtrystram:
Add a test to verify that the deployed container imgref in our disk images start with the correct prefix.

Having an invalid value here can lead to Zincati not being able to update and such a change can slip through because image-build tools don't actually parse the field but simply check for a list of known acceptable values.

Here we use `ostree container info` to parse the whole imgref, ensuring it's valid.

### DIFF
--- a/tests/kola/disks/container-origin-imgref
+++ b/tests/kola/disks/container-origin-imgref
@@ -1,0 +1,29 @@
+#!/bin/bash
+## kola:
+##   # We need internet to check the container imgref
+##   tags: needs-internet
+##   exclusive: false
+##   description: |
+##      Verify that the deployed container image on the disk image has
+##      a correct ostree sigverify prefix.
+
+set -euo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+EXPECTED_IMGREF_PREFIX="ostree-image-signed:docker://"
+
+# get the running container imgref
+booted_imgref=$(rpm-ostree status --booted --json | jq -r '.deployments[0]."container-image-reference"')
+
+# Verify the booted imgref starts with the expected prefix
+if [[ ! $booted_imgref =~ ^$EXPECTED_IMGREF_PREFIX ]]; then
+  echo "booted deployment imgref:" "$booted_imgref"
+  echo "Expected prefix:" "$EXPECTED_IMGREF_PREFIX"
+  fatal "The booted deployment imgref does start with the expected prefix"
+fi
+
+# Verify ostree does accept the imgref as valid
+ostree container info "$booted_imgref"
+ok


tests/kola/disks/container-origin-imgref
```diff
@@ -0,0 +1,28 @@
...
+
+# get the running container imgref
+booted_imgref=$(rpm-ostree status --booted --json | jq -r '.deployments[0]."container-image-reference"')
+
+# Verify the booted imgref start with
+if [[ $booted_imgref =~ ^$EXPECTED_IMGREF_PREFIX ]]; then
+  echo "booted deployment imgref:" "$booted_imgref"
+  echo "Expected prefix:" "$EXPECTED_IMGREF_PREFIX"
+  fatal "The booted deployement imgref does start with the expected prefix"
+fi
```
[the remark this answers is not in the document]
jbtrystram: wooops :) fixed

tests/kola/disks/container-origin-imgref
```diff
@@ -0,0 +1,29 @@
...
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+EXPECTED_IMGREF_PREFIX="ostree-image-signed:docker://"
+
+# get the running container imgref
+booted_imgref=$(rpm-ostree status --booted --json | jq -r '.deployments[0]."container-image-reference"')
+
+# Verify the booted imgref start with
```
Rolv-Apneseth:
```suggestion
# Verify the booted imgref starts with the expected prefix
```

Rolv-Apneseth (dismissed):
LGTM, just a nit on the comment

Rolv-Apneseth (dismissed):

tests/kola/disks/container-origin-imgref
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+## kola:
+##   # We need internet to check the container imgref
+##   tags: needs-internet
+##   exclusive: false
+##   description: |
+##      Verify that the deployed container image on the disk image has
+##      a correct ostree sigverify prefix.
```
dustymabe, line 8: Maybe we should add a creationdate here just in case there's any fallout downstream.

[This](https://github.com/coreos/coreos-assembler/pull/4518) PR will probably merge today.

tests/kola/disks/container-origin-imgref
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+## kola:
+##   # We need internet to check the container imgref
+##   tags: needs-internet
```
dustymabe: Not sure I understand this requirement. Can you explain more? What remote resource is being accessed/fetched here?

dustymabe (commented):
Will this test pass downstream?

tests/kola/disks/container-origin-imgref
dustymabe: > [This](https://github.com/coreos/coreos-assembler/pull/4518) PR will probably merge today.

merged!

tests/kola/disks/container-origin-imgref
jbtrystram: Ostree container info reach out to the repo to get the hash of the image. That's a bit useless but that's the easy way I found to have ostree parse the reference

jbtrystram:
> Will this test pass downstream?

Good point, i guess not. Let's add a `fcos` kola distro filter ?

tests/kola/disks/container-origin-imgref
dustymabe: ok. Now I see what you are trying to do. I guess what you are verifying here is that because the image can be made with `bootc` (i.e. not `ostree`) that we can still use `ostree` to parse things.

Maybe add this to a comment below?

tests/kola/disks/container-origin-imgref
jbtrystram: done.

dustymabe: still no `creationDate` here that I see.

We don't need it, but having it could make rolling this out easier if the test fails for some reason that we need to investigate.

tests/kola/disks/container-origin-imgref
```diff
@@ -0,0 +1,35 @@
...
+
+# Verify the booted imgref starts with the expected prefix
+if [[ ! $booted_imgref =~ ^$EXPECTED_IMGREF_PREFIX ]]; then
+  echo "booted deployment imgref:" "$booted_imgref"
+  echo "Expected prefix:" "$EXPECTED_IMGREF_PREFIX"
+  fatal "The booted deployment imgref does start with the expected prefix"
+fi
+
+# Verify that ostree does accept the imgref as valid.
+# We don't really care about the result, but in order to
```
dustymabe: > We don't really care about the result

Saying "we don't really care about the result" might be confusing to some people.

i.e. what is the result?

To be more clear here we could say:

We don't really care about the output of this command (the information pulled from the remote), but just that the command doesn't fail given our `"$booted_imgref"` string as input.

dustymabe (dismissed):
this LGTM, dropped a few comments, but non-blocking.

tests/kola/disks/container-origin-imgref
```diff
@@ -0,0 +1,35 @@
...
+# do this operation ostree will parse the full imgref.
+# Since the image was built with bootc, we want to make
+# sure (rpm)-ostree does not trip on the created imgref.
+ostree container info "$booted_imgref"
+ok
```
Rolv-Apneseth, line 29: More a question for me, but is it preferred to end all tests with `ok` even if there's no message after it?

jbtrystram: Mhmm i haven't written a lot of these so I wouldn't know for sure.
I felt this was simply more explicit, i.e. "if we reach this. The test is passing"

Rolv-Apneseth: Sure, makes sense

tests/kola/disks/container-origin-imgref
jbtrystram: oops, forgot to push the change. Fixed now

travier (approved):

travier:
@jbtrystram This test is failing on RHCOS: https://github.com/coreos/rhel-coreos-config/pull/245

tests/kola/disks/container-origin-imgref
travier: I think you still did not add the date here :)

travier:
> > Will this test pass downstream?
> 
> Good point, i guess not. Let's add a `fcos` kola distro filter ?

I guess we missed that :)

PR #4101, opened 2026-04-10
tests: Remove the ext.config.boot.raid1-boot test

Rolv-Apneseth:
In relation to https://github.com/coreos/fedora-coreos-tracker/issues/2134

Remove the `ext.config.boot.raid1-boot` test for now, as it is failing on ppc64le and s390x. The failure on s390x means that we can't just denylist the test as the failure is introduced by the very presence of the test's butane config file.

Once we fix the issues causing the failures, we can add the test back in (I would assume cherry-pick rather than reverting the revert commits). At the very least, once the s390x issue is resolved the test can be added back in with a denylist entry or maybe just the new `creationDate` metadata field while we work on the ppc64le issue.


Rolv-Apneseth:
From the bot:

> Removing these helper functions from the shared `commonlib.sh` library is a breaking change for any other tests that rely on them. These assertions (`assert_file_has_content` and `assert_file_has_content_literal`) are generic and likely used by other test suites in the repository.

But these aren't currently used anywhere else.

> Furthermore, the current implementation of these functions is buggy as they call an undefined `_fatal_print_file` function.

This does seem like an issue but can be resolved after we bring these commits back.

dustymabe (approved):
LGTM

travier:
This reverted https://github.com/coreos/fedora-coreos-config/pull/4097

PR #4104, opened 2026-04-13
[next-devel] overrides: fast-track downgraded packages 2026-04-13

marmijo:
F44 is now in final freeze, which means that some packages in F43 will sort as newer than F44. We prevent this by fast-tracking any packages that show as downgraded. Today they are:

```
bootc
kernel
util-linux
selinux-policy
```


dustymabe (approved):
LGTM

PR #4107, opened 2026-04-14
[rawhide] overrides: fast-track glibc-2.43.9000-10.fc45

marmijo:
This newer version of glibc seems to resolve the recent failures seen in rawhide.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2133


jbtrystram (approved):
Thanks

PR #4109, opened 2026-04-15
overlay.d/05core: Remove obsolete 'bootupd.socket' preset

travier:
bootup is now using a transient systemd service and no longer use the socket unit.

The actual service unit trigerring the updates is enabled in the Fedora bootc image.

See: https://github.com/coreos/bootupd/issues/551


dustymabe (approved):
LGTM

PR #4114, opened 2026-04-20
numad/test.sh: Increase timeout from 25s to 45s

angelcerveraroldan:
This test can be flaky, potentially due to numad not having enough time to log information about the stress-ng process.


dustymabe (approved):
LGTM

PR #4116, opened 2026-04-20
overlay.d/05core: drop obsolete or redundant dracut configuration

dustymabe:
commit 287dabab9d4bd043179085a7b3dfd454328598d8
Date:   Mon Apr 20 17:03:38 2026 -0400

    overlay.d/05core: drop hostonly=no dracut config
    
    This is defined in the bootc base images [1] so we don't need to
    additionally define it here.
    
    [1] https://gitlab.com/fedora/bootc/base-images/-/blob/9f9824c9a4ec87062509fe2f9f56382088aebaba/minimal/initramfs.yaml#L9

commit 4a89ee6c0e9baee06d84524eef26fcf4c89924d2
Date:   Mon Apr 20 17:00:45 2026 -0400

    overlay.d/05core: drop dracut conf disabling strip
    
    Since we don't ship it anyway this will have no effect other than
    preventing a harmess warning message:
    
    ```
    dracut[I]: Could not find 'strip'. Not stripping the initramfs.
    ```
    
    I think keeping this config and spending cycles in the future to figure
    out if it is still useful is more harmful than it's worth.
    
    As a side note the bootc base images don't have this config. If we did
    decide it was important enough in the future to re-add this I would argue
    the config should get added there, rather than re-added here.

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/dracut.conf.d/60-coreos-nohostonly.conf
+++ b/overlay.d/05core/usr/lib/dracut/dracut.conf.d/60-coreos-nohostonly.conf
@@ -1,2 +1,0 @@
-# Default rpm-ostree model is server-side generated initramfs
-hostonly=no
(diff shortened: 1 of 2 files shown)


jbtrystram (approved):
LGTM

Small typo in the first commit message : `harmless`

dustymabe:
> LGTM

Thanks!

> 
> Small typo in the first commit message : `harmless`

Do you mean:

> I think keeping this config and spending cycles in the future to figure
out if it is still useful is more harmful than it's worth.

I meant to say "harmful" but maybe I could have been more clear. Basically what I was saying is that keeping the configuration here isn't worth the minor gain provided by keeping it. i.e. just having it is causing harm.

PR #4122, opened 2026-04-22
[testing] overlay.d/05core: disable systemd-oomd.socket too

dustymabe:
At some point the socket started getting enabled by default [1], which means systemd-oomd.service was running. Let's disable the socket too.

[1] https://src.fedoraproject.org/rpms/redhat-systemd-presets/blob/b056f04a013590fa58d765807c7d1e210608690c/f/90-default.preset#_15

(cherry picked from commit 34d14ce20c75b487ab4e94e46300c8882da8287c)


dustymabe:
context in https://github.com/coreos/fedora-coreos-tracker/issues/840#issuecomment-4298458387

This commit goes into the `testing` branch since we anticipate promoting the content in this branch to `stable`.

tlbueno:
lgtm

joelcapitao (approved):

PR #4128, opened 2026-04-24
[testing-devel] Move to Fedora 44

marmijo:
CI won't pass on this until F44 is officially released because the repos aren't populated yet. So this won't be able to merge until Tuesday.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2055

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -29,34 +29,11 @@ conditional-include:
     include:
       ostree-layers:
         - overlay/10aarch64
-  # In F43 and older pull in the nfs-utils-coreos package
-  - if: releasever <= 43
-    include:
-      packages:
-        - nfs-utils-coreos
-  # In F44 we can use the nfs-client-utils package, which obsoletes nfs-utils-coreos
-  - if: releasever >= 44
-    include:
-      packages:
-        - nfs-client-utils
-  # Do the alternatives migration from 43
-  - if: releasever >= 43
+  # Do the alternatives migration on F44 and older
+  - if: releasever <= 44
     include:
       ostree-layers:
         - overlay/50alternatives
-  # Drop fedora-repos-ostree in F44
-  - if: releasever < 44
-    include:
-      packages:
-        - fedora-repos-ostree
-  # We carry our own ssh config for authorized keys in <44. In 44+
-  # The config is delivered via the ignition + afterburn RPMs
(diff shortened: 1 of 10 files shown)


manifests/fedora-coreos.yaml
```diff
@@ -29,38 +29,11 @@ conditional-include:
...
-      packages:
-        - fedora-repos-ostree
-  # We carry our own ssh config for authorized keys in <44. In 44+
-  # The config is delivered via the ignition + afterburn RPMs
-  # https://src.fedoraproject.org/rpms/ignition/c/3ccce9fe2c0462f4a17923937933ab9c26db0aa5?branch=rawhide
-  # https://src.fedoraproject.org/rpms/rust-afterburn/c/e61c6e5f776410ea72e8f731cec88567577f6bb8?branch=rawhide
-  - if: releasever <= 43
-    include:
-      ostree-layers:
-        - overlay/18sshd-authorized-keys-file
```
dustymabe, line 59: should also delete the files associated with this overlay.

manifests/fedora-coreos.yaml
```diff
@@ -29,38 +29,11 @@ conditional-include:
...
-  # In F44 we can use the nfs-client-utils package, which obsoletes nfs-utils-coreos
-  - if: releasever >= 44
-    include:
-      packages:
-        - nfs-client-utils
-  # Do the alternatives migration from 43
-  - if: releasever >= 43
-    include:
-      ostree-layers:
-        - overlay/50alternatives
```
dustymabe, line 46: so you are properly retaining this, but I'm not sure we'll need it forever. Maybe let's keep it up here as conditional, but now make the conditional `if: releasever <= 44` so it will drop out when we reach F45?

cc @travier

manifests/fedora-coreos.yaml
marmijo: done!

manifests/fedora-coreos.yaml
marmijo: I updated the conditional to `if: releasever <= 44`. Let me know if any other changes are needed!

dustymabe:
Can we delete the overlay.d/README.md entry for 18sshd-authorized-keys-file

otherwise LGTM

dustymabe (dismissed):

dustymabe (approved):

PR #4143, opened 2026-04-30
[stable] overrides: fast-track kernel-6.19.13-200.fc43

coreosbot-releng:

Fixes #2140


travier:
So we have 6.19.12 in testing. Why not move to that if it has the fix? 6.19.13 has only been in next so far.

jbtrystram:
> So we have 6.19.12 in testing. Why not move to that if it has the fix? 6.19.13 has only been in next so far.

Sorry yeah, error on my end. Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/4144

PR #4147, opened 2026-05-04
.cci.jenkinsfile: drop testiso and run kola after all artifacts

nikita-dubrovskii:
Drop kolaTestIso() and instead run kola() after building all artifacts. The iso.* tests are now folded into the main kola test suite, so we build all artifacts first (metal, metal4k, live) before running tests.

Issue: https://github.com/coreos/coreos-assembler/issues/3989


dustymabe (approved):
LGTM

PR #4148, opened 2026-05-04
Drop alternatives migration and tests

travier:
Drop the migration code and corresponding test that added for [1] & [2] as we have kept this logic for the entire F43 cycle (and thus at least one barrier release) so all systems should be migrated now.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1818
[2] https://github.com/coreos/fedora-coreos-tracker/issues/677


dustymabe (approved):
LGTM

travier:
I wrote up my investigation for RHCOS in https://github.com/coreos/rhel-coreos-config/issues/250

PR #4150, opened 2026-05-05
tests: re-introduce raid1-boot test on x86_64 and aarch64 only

Rolv-Apneseth:
See https://github.com/coreos/fedora-coreos-tracker/issues/2134

This should allow the test to run on `aarch64` and `x86_64`, and not cause a failure on `s390x`.


travier (dismissed):
LGTM, let's start with that.

Rolv-Apneseth:
Whoops, thanks for fixing the title.

Rebased and made sure to include `creationDate` for each test.

travier (dismissed):

travier:
Konflux CI error is "expected" right now. See https://github.com/coreos/fedora-coreos-config/pull/4149#issuecomment-4386855702

Rolv-Apneseth:
> Konflux CI error is "expected" right now. See [#4149 (comment)](https://github.com/coreos/fedora-coreos-config/pull/4149#issuecomment-4386855702)

Ah OK, thanks

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
```diff
@@ -0,0 +1,62 @@
...
+
+check_bootupctl_components() {
+    local components=()
+    local output_file="out.txt"
+
+    # Define components based on the passed architecture
+    case "$(arch)" in
+        x86_64) components=("BIOS" "EFI") ;;
+        aarch64) components=("EFI") ;;
+        ppc64le) components=("BIOS") ;;
```
marmijo: Can we drop this `ppc64le` case since the test doesn't target that arch anymore?

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
```diff
@@ -0,0 +1,62 @@
...
+
+check_basic() {
+    srcdev=$(findmnt -nvr /sysroot -o SOURCE)
+    [[ "${srcdev}" == "/dev/md126" ]] && echo "sysroot has correct source"
+
+    blktype=$(lsblk -o TYPE "${srcdev}" --noheadings)
+    [[ "${blktype}" == "raid1" ]] && echo "block type is raid1"
+
+    fstype=$(findmnt -nvr /sysroot -o FSTYPE)
+    [[ "${fstype}" == "xfs" ]] && echo "source is XFS on RAID1 device"
```
marmijo: These checks appear to be informational, but non-fatal. If the first condition is `false`, the script just doesn't print and moves on. `fatal` or another assertion helper could be used here instead.


```bash
if [[ "${srcdev}" != "/dev/md126" ]]; then 
    fatal "sysroot does not have the correct source"
fi
```
or to keep the original style
```bash
[[ "${srcdev}" == "/dev/md126" ]] || fatal "sysroot does not have the correct source"
```

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
```diff
@@ -0,0 +1,62 @@
...
+# Common checks for raid1-boot tests across different architectures.
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+check_basic() {
+    srcdev=$(findmnt -nvr /sysroot -o SOURCE)
+    [[ "${srcdev}" == "/dev/md126" ]] && echo "sysroot has correct source"
```
marmijo: The check on `/dev/md126` looks fragile. The kernel assigned `md127` during my test run, so this failed (see my comment about `fatal`). Since the block type check below already confirms it's a RAID1 device, is this check actually needed?

marmijo (commented):
Mostly LGTM! I added a few comments.

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
Rolv-Apneseth: It's supposed to eventually I guess is why I kept it in, just there's currently issues with grub and our specific ppc64le raid setup. You think I should remove it for now and just add the case back later?

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
Rolv-Apneseth: Whoops, you're right. I'll change them to the second style, thanks

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
Rolv-Apneseth: I didn't know those were dynamically assigned, I thought `/sysroot` always gets md126 and `/boot` always gets md127. Bootupd also has [this check](https://github.com/coreos/bootupd/blob/214ca7fb40f0e7e311d1cf9fdf90fc3502716325/tests/kola/raid1/test.sh#L22-L23) - is that also fragile?

Rolv-Apneseth: Checks for `/boot` being on `md-boot` and the same for `/sysroot` on `md-root` might be worth having though

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
Rolv-Apneseth: With `set -xeuo pipefail` shouldn't these checks still fail actually? Maybe I don't fully understand how it works. I prefer the style of explicit `fatal` calls anyway though so I'll change those tomorrow

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
Rolv-Apneseth: Removed the ppc64le case, will add it back when/if we re-enable this test for that platform

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
marmijo: I don't think `set -e` helps here. The issue is that `[[ condition ]] && echo "..."` puts the `[[` in a conditional context, so a false result has exit code `1` but `set -e` never sees it as an unhandled error, it just skips the `&&` right-hand side and moves on. 


I ran this test with explicitly false conditions, and the test still passed (before the recent force pushes):

```bash
[[ "${fstype}" == "xfsasdfasdf" ]] && echo "source is XFS on RAID1 device"
```

With the recent changes, `|| fatal` ensures the test will fail.

marmijo (approved):
LGTM, tested locally on `x86_64`. Thanks for working on this!

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
marmijo: > I didn't know those were dynamically assigned, I thought /sysroot always gets md126 and /boot always gets md127. Bootupd also has [this check](https://github.com/coreos/bootupd/blob/214ca7fb40f0e7e311d1cf9fdf90fc3502716325/tests/kola/raid1/test.sh#L22-L23) - is that also fragile?

Not sure, in my local testing I saw both `md126` and `md127` on the same hardware, so it may depend on the environment. If the bootupd check is consistently passing in CI it's probably fine to leave as-is there. Either way, the name-based checks in the current version are more robust than relying on device numbers, so I wouldn't worry about it.

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
Rolv-Apneseth: Ah, thanks for explaining, that makes sense

tests/kola/boot/raid1-boot/data/commonlib_raid1.sh
Rolv-Apneseth: I may need to make some changes over there to remove the `{{LAYOUT}}` added to cosa so we can revert that commit. Thought maybe we could change that check too while I do that.

PR #4151, opened 2026-05-06
overrides: fast-track bootc-1.15.2-1.fc44

coreosbot-releng:
Requested by @joelcapitao via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


joelcapitao:
This fast-track does not solve our issue because we are using the `bootc` RPM of the fedora-bootc base image.

dustymabe (approved):
LGTM

PR #4152, opened 2026-05-06
tests/container-origin-imgref : add fcos tag

jbtrystram:
This test will only work on FCOS so let's add a filter for that to avoid it failing downstream.

This was supposed to go in the initial PR but I messed up the review follow-up and forgot to add this.

See https://github.com/coreos/fedora-coreos-config/pull/4099#issuecomment-4386934297


travier (approved):

travier:
Hum, CI looks stuck here?

PR #4158, opened 2026-05-08
test/bootupd: fix pipefail interaction with systemctl status

prestist:
The bootloader-update.service skip checks added in 82ab7dd always fail because `systemctl status` returns exit code 3 for an inactive/skipped service. With `set -o pipefail`, the pipeline `systemctl status ... | grep ...` inherits this non-zero exit code even when grep matches, causing the `if !` to always enter the failure branch.

Fix by capturing the status output into a variable first (suppressing the exit code), then grepping the variable.

Relates to : https://github.com/coreos/fedora-coreos-config/pull/4149

### DIFF
--- a/tests/kola/boot/bootupd
+++ b/tests/kola/boot/bootupd
@@ -29,7 +29,8 @@ case "$(arch)" in
         fi
         ;;
     ppc64le)
-        if ! systemctl status bootloader-update.service | grep -iq "ConditionArchitecture=!ppc64-le was not met"; then
+        bl_status=$(systemctl status bootloader-update.service 2>&1 || true)
+        if ! echo "$bl_status" | grep -iq "ConditionArchitecture=!ppc64-le was not met"; then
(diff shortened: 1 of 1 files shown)


tests/kola/boot/bootupd
```diff
@@ -29,7 +29,8 @@ case "$(arch)" in
         fi
         ;;
     ppc64le)
-        if ! systemctl status bootloader-update.service | grep -iq "ConditionArchitecture=!ppc64-le was not met"; then
+        bl_status=$(systemctl status bootloader-update.service 2>&1 || true)
+        if ! echo "$bl_status" | grep -iq "ConditionArchitecture=!ppc64-le was not met"; then
```
dustymabe, line 33: Maybe for the future use something like:

```suggestion
        if grep -iq "ConditionArchitecture=!ppc64-le was not met" <<< "$bl_status"; then
```

dustymabe (approved):
LGTM

PR #4160, opened 2026-05-08
denylist: skip ext.config.docker.swarm-overlay-network on s390x/rawhide

nikita-dubrovskii:
Test fails with:
  + docker run -t --network=testnetwork quay.io/fedora/fedora-minimal:44 curl https://icanhazip.com/ curl: (6) Could not resolve host: icanhazip.com
  + fatal 'Docker swarm overlay working not functional. Test fails.'

Tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2143

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -82,3 +82,12 @@
     - s390x
   streams:
     - rawhide
+
+- pattern: ext.config.docker.swarm-overlay-network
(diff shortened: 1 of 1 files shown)


dustymabe (approved):

c4rt0 (commented):
If I can't see the the bot tagging no new line at the end of the file, it probably means it's there already...

/lgtm

PR #4167, opened 2026-05-10
[stable] tree: promote changes from testing at cb88ba4717a6e8fecc711847f4efea868450df74

coreosbot-releng:
Triggered by @dustymabe in https://github.com/coreos/fedora-coreos-streams/issues/1237.


dustymabe:
check promotion diffs failures are expected because of my extra commits here.

dustymabe:
See https://github.com/coreos/fedora-coreos-tracker/issues/2142#issuecomment-4416057431 for logic on using the 6.19.14 kernel here.

joelcapitao (approved):
LGTM
The `Check promotion diffs` GH job is failing as expected because of the additional commit that overrides the kernel

dustymabe:
ci didn't run testiso tests because we dropped `kola testiso` in https://github.com/coreos/coreos-assembler/commit/84a7bf5ac0749e105d13081f4cc0a37b0c0a358b.

The Jenkinsfile here will get updated in the next `stable` promotion in two weeks and we won't have that problem anymore.

I verified everything else other than iso tests passed.

PR #4170, opened 2026-05-12
overlays/bootc: Add missing bootc install options

jbtrystram:
There are a few options that we are hard-coding in COSA right now, but with [1] we can pull them from the OCI container, so let's do that instead and make the container the canonical source of truth.

Also rework this overlay a bit in two files : the firstboot related options and the one related to the ostree repository.

[1] https://github.com/coreos/coreos-assembler/commit/364ccaebeb006c1653771158432541524954b604

### DIFF
--- a/overlay.d/10bootc/usr/lib/bootc/install/15-ostree.toml
+++ b/overlay.d/10bootc/usr/lib/bootc/install/15-ostree.toml
@@ -1,0 +1,7 @@
+[install]
+# Name of the ostree stateroot
+stateroot = "fedora-coreos"
+# Make sure bootc enforces that
+# `/etc/containers/policy.json` includes
+# a default policy that verifies our images signature.
+enforce-container-sigpolicy = true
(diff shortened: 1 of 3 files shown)


angelcerveraroldan (approved):
lgtm

overlay.d/10bootc/usr/lib/bootc/install/15-ostree.toml
```diff
@@ -0,0 +1,7 @@
+[install]
+# Name of the ostree stateroot
+stateroot = "fedora-coreos"
```
dustymabe, line 3: If we ever intend to share this layer with RHCOS I think we'd need to not hardcode this value?

jbtrystram: Opened https://github.com/coreos/fedora-coreos-config/pull/4173

PR #4154, opened 2026-05-07
[rhel-9.6] Disable automatic bootloader update for ppc64le, s390x

Johan-Liebert1:
Disabling this due to grub update issues with raid setups

See: https://github.com/coreos/fedora-coreos-tracker/issues/2134

Cherry picked commits from (#4149)

PR #4172, opened 2026-05-12
test: Update bootloader-update status check

Johan-Liebert1:
Instead of putting test for s390x in the default branch of the switch, put it in its own branch and fallthrough to to the default

See: https://github.com/coreos/fedora-coreos-config/pull/4154#discussion_r3209514868


prestist (approved):
Ah nice, lgtm, only nit would be commit message.

Maybe "tests/kola/boot/bootupd: update bootloader staus check" wdyt?

Johan-Liebert1:
> Maybe "tests/kola/boot/bootupd: update bootloader staus check" wdyt?

Done

PR #4173, opened 2026-05-13
overlays/bootc: Move stateroot to FCOS-specific overlay

jbtrystram:
The stateroot name is specific to FCOS, so move it from 10bootc overlay to 15fcos overlay. This way if the 10bootc layer is ever shared with RHCOS, the stateroot won't be hardcoded to the wrong value.

See https://github.com/coreos/fedora-coreos-config/pull/4170#discussion_r3227154358

Assisted-by: Opencode.ai <Opus 4.6>"

### DIFF
--- a/overlay.d/10bootc/usr/lib/bootc/install/15-ostree.toml
+++ b/overlay.d/10bootc/usr/lib/bootc/install/15-ostree.toml
@@ -1,6 +1,4 @@
 [install]
-# Name of the ostree stateroot
-stateroot = "fedora-coreos"
 # Make sure bootc enforces that
 # `/etc/containers/policy.json` includes
 # a default policy that verifies our images signature.
--- a/overlay.d/15fcos/usr/lib/bootc/install/15-ostree.toml
+++ b/overlay.d/15fcos/usr/lib/bootc/install/15-ostree.toml
@@ -1,0 +1,3 @@
+[install]
+# Name of the ostree stateroot
+stateroot = "fedora-coreos"


dustymabe:
The fact that the value is hardcoded makes things somewhat brittle and also means we need to maintain separate copies in RHCOS and FCOS (and I guess SCOS).

WDYT about instead of this we add a postprocess that creates the file (maybe `/usr/lib/bootc/install/15-ostree-stateroot.toml` would be a better name?) based on the [`NAME` build-arg](https://github.com/coreos/fedora-coreos-config/blob/1dadf63dd5f4810eb1dd7c81b64013360112b1d7/build-args.conf#L5) and then we can use the exact same code in all places.


Hmm... OK.. looking at this further it's complicated by the fact that passing the variable to a postprocess doesn't really happen. I think the best thing to do then is to add another function to do this in build-rootfs:

```diff
diff --git a/build-rootfs b/build-rootfs
index 2c7d0e8d..dbe9dca8 100755
--- a/build-rootfs
+++ b/build-rootfs
@@ -55,6 +55,7 @@ def main():
         'version': os.getenv('VERSION'),
         'stream': os.getenv('STREAM'),
         'osid': os.getenv('ID'),
+        'osname': os.getenv('NAME'),
         'mutate_os_release': os.getenv('MUTATE_OS_RELEASE'),
         'strict_mode': os.getenv('STRICT_MODE'),
         'passwd_group_dir': os.getenv('PASSWD_GROUP_DIR')
@@ -217,6 +218,7 @@ def build_rootfs(target_rootfs, srcdir, manifest_name,
     if strict_mode == '1':
         verify_strict_mode(target_rootfs, locked_nevras)
     run_postprocess_scripts(target_rootfs, manifest)
+    run_add_bootc_install_ostree_stateroot(target_rootfs, osname)
     run_dracut(target_rootfs)
     cleanup_extraneous_files(target_rootfs)
 
```

alternatively figure out some way to pass the variables into postprocess scripts (templating?) so they can be used there.

That gets us back to a place where this information is sourced from build-args (today it's just translated from build-args into the [osname argument](https://github.com/coreos/coreos-assembler/blob/287c1905d9446f8b5c0fd874e5973797900c3aa0/src/runvm-osbuild#L136) we pass to osbuild.

jbtrystram:
> The fact that the value is hardcoded makes things somewhat brittle and also means we need to maintain separate copies in RHCOS and FCOS (and I guess SCOS).

Isn't the point of overlays to ship hardcoded config values ? I get the desire to have as much shared code as possible, but writing python code in `build-rootfs` to avoid duplicating a single `key=value` pair in a TOML file feels like over-engineering.

> alternatively figure out some way to pass the variables into postprocess scripts (templating?) so they can be used there.

That's where I would go, going that extra mile so we could get a real benefit that would bring values to some other places probably. Maybe we could just make the `build-args` values available as environnement variables in the postprocess scripts.

dustymabe:
Let me be clear, I won't block this change (I'll even LGTM it). It's a fine change, but I think the cost of it will be greater (for example we'll have to go make changes downstream in `rhel-coreos-config` and manage those changes over time) than if we did it a different way. This is just advice.

> > The fact that the value is hardcoded makes things somewhat brittle and also means we need to maintain separate copies in RHCOS and FCOS (and I guess SCOS).
> 
> Isn't the point of overlays to ship hardcoded config values ?

The way we have architected things (where you can optionally ship some overlays in FCOS and others downstream) does offer that flexibility.

>  I get the desire to have as much shared code as possible, but writing python code in `build-rootfs` to avoid duplicating a single `key=value` pair in a TOML file feels like over-engineering.

I think it's less about shared code and more about the source of truth. The prior source of truth was build-args (and before that the manifest.yaml). i.e. there's only one place where the value is defined and any changes to it will flow down everywhere it's used.

Now, it just so happens that going back to having the source of truth be the NAME value in build-args means we can share the code.

> 
> > alternatively figure out some way to pass the variables into postprocess scripts (templating?) so they can be used there.
> 
> That's where I would go, going that extra mile so we could get a real benefit that would bring values to some other places probably. Maybe we could just make the `build-args` values available as environnement variables in the postprocess scripts.

+1 from my side.

dustymabe:
> Let me be clear, I won't block this change (I'll even LGTM it).

OK I lied :).. One potential problem:

The current way this PR is set up means that in the two overlays one of them will override the other. You'll need to choose a different filename for one of them than `/usr/lib/bootc/install/15-ostree.toml` right?

That is unless we don't want to set `enforce-container-sigpolicy = true` for both upstream and downstream?

jbtrystram:
closing this in favor of https://github.com/coreos/fedora-coreos-config/pull/4176

PR #4175, opened 2026-05-13
tests/kola/disks: handle missing remote image in container-origin-imgref

aaradhak:
The `ostree container info` command reaches out to the remote registry to fetch and validate the manifest. This fails when the remote tag does not have a manifest entry for the current architecture which can happen if a previous `bump-lockfile` run failed for that arch before the image was pushed.

Wrap the call so the test skips remote validation when the image is not available even when enforcing the prefix check.

### DIFF
--- a/tests/kola/disks/container-origin-imgref
+++ b/tests/kola/disks/container-origin-imgref
@@ -1,7 +1,5 @@
 #!/bin/bash
 ## kola:
-##   # We need internet to check the container imgref
-##   tags: needs-internet
 ##   exclusive: false
 ##   distros: fcos
 ##   description: |
@@ -22,9 +20,19 @@ booted_imgref=$(rpm-ostree status --booted --json | jq -r '.deployments[0]."cont
 if [[ ! $booted_imgref =~ ^$EXPECTED_IMGREF_PREFIX ]]; then
   echo "booted deployment imgref:" "$booted_imgref"
   echo "Expected prefix:" "$EXPECTED_IMGREF_PREFIX"
-  fatal "The booted deployment imgref does start with the expected prefix"
+  fatal "The booted deployment imgref does not start with the expected prefix"
 fi
 
-# Verify ostree does accept the imgref as valid
-ostree container info "$booted_imgref"
+# Verify ostree does accept the imgref as valid. Using --help
+# avoids reaching out to the remote registry while still
+# validating the imgref format locally.
+
+# First, verify the --help trick rejects an invalid imgref so we
+# know it is actually validating.
+if ostree container info "invalid-imgref:docker://example.com/repo:tag" --help 2>/dev/null; then
+  fatal "ostree container info did not reject an invalid imgref"
+fi
+
+# Now validate the real booted imgref
+ostree container info "$booted_imgref" --help
 ok


tests/kola/disks/container-origin-imgref
```diff
@@ -25,6 +25,11 @@ if [[ ! $booted_imgref =~ ^$EXPECTED_IMGREF_PREFIX ]]; then
...
 
-# Verify ostree does accept the imgref as valid
-ostree container info "$booted_imgref"
+# Verify ostree does accept the imgref as valid. The remote image
+# may not exist for this architecture yet (if a previous
+# bump-lockfile run failed for this arch), so skip in that case.
+if ! ostree container info "$booted_imgref" 2>&1; then
+  echo "ostree container info failed; remote image may not exist for this arch yet"
+  echo "Skipping validation of remote image"
+fi
```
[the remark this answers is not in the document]
aaradhak: Made the changes based on this review

tests/kola/disks/container-origin-imgref
```diff
@@ -25,6 +25,16 @@ if [[ ! $booted_imgref =~ ^$EXPECTED_IMGREF_PREFIX ]]; then
   fatal "The booted deployment imgref does start with the expected prefix"
 fi
 
-# Verify ostree does accept the imgref as valid
-ostree container info "$booted_imgref"
+# Verify ostree does accept the imgref as valid. The remote image
+# may not exist for this architecture yet (if a previous
+# bump-lockfile run failed for this arch), so skip in that case.
```
dustymabe:
```suggestion
# may not exist for this architecture (if the previous build
# failed for this architecture it won't be in the manifest list),
# so skip in that case.
```

tests/kola/disks/container-origin-imgref
```diff
@@ -25,6 +25,16 @@ if [[ ! $booted_imgref =~ ^$EXPECTED_IMGREF_PREFIX ]]; then
...
+if ! output=$(ostree container info "$booted_imgref" 2>&1); then
+  if grep -qi "no image found in image index for architecture" <<< "$output"; then
+    echo "ostree container info failed; remote image may not exist for this arch yet"
```
dustymabe:
```suggestion
    echo "ostree container info failed; remote image may not exist for this architecture"
```

tests/kola/disks/container-origin-imgref
jbtrystram: Ah, nice catch. When I wrote this test I was reasoning based on existing streams (where we always have previously published images with moving tags), and did not tought of the bump-lockfile case.

Maybe we can rework this test to avoid reaching out to the repo entirely, I found a way to trick ostree to do the validation without reaching out to the registry : 

```
ostree container info ostree-image-signed:registry:quay.io/fedora/fedora-coreos:stable --help
Print information about an exported ostree-container image

Usage: ostree-ext container info <IMGREF>

Arguments:
  <IMGREF>  Image reference, e.g. registry:quay.io/exampleos/exampleos:latest

Options:
  -h, --help  Print help
jbtrystram@Host-002:~$ 
jbtrystram@Host-002:~$ ostree container info invalid-imgref:registry:quay.io/fedora/fedora-coreos:stable --help
error: invalid value 'invalid-imgref:registry:quay.io/fedora/fedora-coreos:stable' for '<IMGREF>': Invalid ostree image reference scheme: invalid-imgref

For more information, try '--help'.
```

jbtrystram: The exit codes are consistent so we don't even need to grep for the error string

dustymabe: Nice. That means we could drop the needs-internet tag too.

The only concern I have here is if somehow this "trick" changed behavior in the code.

To alleviate that concern maybe we could add a negative test in here first where we call `ostree container info invalid-imgref:registry:quay.io/fedora/fedora-coreos:stable --help` first to make sure it errors with the expected error message first before we do the actual `"$booted_imgref"` check.

dustymabe: @aaradhak do you think yo ucould make those changes?

aaradhak: yes, I ll address this.

aaradhak: I have made the changes

dustymabe (approved):
LGTM

dustymabe:
Thanks for working on this!

PR #4176, opened 2026-05-14
build-rootfs: inject build-args to postprocess env

jbtrystram:
Injects all the build-args as environnement variables into the
postprocess bwrap. This will allow postprocess scripts to reuse these
values and make arbitrary changes based on it.

See discussion on #4173
for context.

Assisted-by: Opencode.ai <Opus 4.6>

### DIFF
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -44,6 +44,7 @@ documentation: false
 # ⚠⚠⚠ ONLY TEMPORARY HACKS ALLOWED HERE; ALL ENTRIES NEED TRACKER LINKS ⚠⚠⚠
 # See also the version of this in fedora-coreos.yaml
 postprocess:
+
   # Mask dnsmasq. We include dnsmasq for host services that use the dnsmasq
   # binary but intentionally mask the systemd service so users can't easily
   # use it as an external dns server. We prefer they use a container for that.
@@ -82,6 +83,17 @@ postprocess:
     #!/usr/bin/env bash
     set -eux -o pipefail
     rm -vrf /usr/lib/modules/*aarch64/dtb/qcom/
+
+  # Inject the ostree stateroot name from the NAME build-arg.
+  # This is done in postprocess rather than an overlay so that the
+  # value is sourced from build-args and shared across FCOS/RHCOS/SCOS.
+  - |
+    #!/usr/bin/bash
+    set -eux -o pipefail
+    cat > /usr/lib/bootc/install/20-ostree-stateroot.toml << EOF
+    [install]
+    stateroot = "${NAME}"
+    EOF
 
 # Packages listed here should be specific to Fedore CoreOS (as in not yet
 # available in RHCOS or not desired in RHCOS). All other packages should go
--- a/Containerfile
+++ b/Containerfile
@@ -19,6 +19,7 @@ ARG BUILDER_IMG=overridden
 FROM ${BUILDER_IMG} as builder
 
 ARG ID=overridden
+ARG NAME=overridden
 ARG VERSION=overridden
 ARG DESCRIPTION=overridden
 ARG STREAM=overridden
--- a/build-rootfs
+++ b/build-rootfs
@@ -22,6 +22,11 @@ import yaml
 ARCH = os.uname().machine
 INPUTHASH = '/run/inputhash'
 HERMETIC_REPO = '/etc/yum.repos.d/cachi2.repo'
+
+# Build-args to propagate as env vars into the postprocess bwrap sandbox.
+# Add entries here as needed; they must also be declared as ARG in the
+# Containerfile builder stage so they are present in os.environ.
+POSTPROCESS_ENV_VARS = ['NAME']
 IS_HERMETIC = os.path.exists(HERMETIC_REPO)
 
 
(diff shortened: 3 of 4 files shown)


dustymabe:
in the past I've avoided parsing build-args.conf in `build-rootfs`. In [other places](https://github.com/coreos/fedora-coreos-config/blob/a205a16d97af6e14bbfdec47ee4c8b34b25ce688/versionary#L52) we use a python library to parse it, but `build-rootfs` runs inside the bootc builder image and doens't have that library and also writing our own code to parse it (as you are doing here) isn't great.

Since all the values from build-args are already in the environment we can just use them directly. `$NAME` wasn't available because it wasn't specified in the containerfile.

Here's an example of how to do it, which also uses the builtin `--setenv` of `bwrap` instead of `env`:


```diff
diff --git a/Containerfile b/Containerfile
index 57ff8ea0..fc20f0a4 100644
--- a/Containerfile
+++ b/Containerfile
@@ -19,6 +19,7 @@ ARG BUILDER_IMG=overridden
 FROM ${BUILDER_IMG} as builder
 
 ARG ID=overridden
+ARG NAME=overridden
 ARG VERSION=overridden
 ARG DESCRIPTION=overridden
 ARG STREAM=overridden
diff --git a/build-rootfs b/build-rootfs
index 2c7d0e8d..ad7a7213 100755
--- a/build-rootfs
+++ b/build-rootfs
@@ -319,7 +319,7 @@ def run_postprocess_scripts(rootfs, manifest):
             os.fchmod(f.fileno(), 0o755)
             f.write(script)
         print(f"Running CoreOS postprocess script {i}", flush=True)
-        bwrap(rootfs, [f'/{name}'])
+        bwrap(rootfs, [f'/{name}'], env=os.environ)
         os.unlink(os.path.join(rootfs, name))
 
 
@@ -327,9 +327,9 @@ def run_dracut(rootfs):
     print(f"Running dracut to generate the initramfs", flush=True)
     # https://docs.fedoraproject.org/en-US/bootc/initramfs/#_modifying_and_regenerating_the_initrd
     kver = bwrap(rootfs, ['ls', '/usr/lib/modules'], capture=True).strip()
-    bwrap(rootfs, ['env', 'DRACUT_NO_XATTR=1',
-                   'dracut', '--verbose', '--force',  '--reproducible',
-                   '--no-hostonly', f"/usr/lib/modules/{kver}/initramfs.img", kver])
+    bwrap(rootfs, env={'DRACUT_NO_XATTR': '1'},
+          args=['dracut', '--verbose', '--force',  '--reproducible',
+                '--no-hostonly', f"/usr/lib/modules/{kver}/initramfs.img", kver])
 
 
 def prepare_local_rpm_overrides(rootfs, srcdir):
@@ -366,11 +366,15 @@ priority=1
 
 # Could upstream this as e.g. `bootc-base-imagectl runroot /rootfs <cmd>` maybe?
 # But we'd need to carry it anyway at least for RHCOS 9.6.
-def bwrap(rootfs, args, capture=False):
+def bwrap(rootfs, args, env=dict(), capture=False):
+    env_args = []
+    for k, v in env.items():
+        env_args.extend(['--setenv', k, v])
+
     args = ['bwrap', '--bind', f'{rootfs}', '/', '--dev', '/dev',
             '--proc', '/proc', '--tmpfs', '/tmp', '--tmpfs', '/var',
             '--tmpfs', '/var/tmp', '--tmpfs', '/run',
-            '--bind', '/run/.containerenv', '/run/.containerenv', '--'] + args
+            '--bind', '/run/.containerenv', '/run/.containerenv'] + env_args + ['--'] + args
     if capture:
         return subprocess.check_output(args, encoding='utf-8')
     subprocess.check_call(args)
```


This particular implementation isn't great (but it does work) because it passes the entire environment which could have some side effects. Likely we should just pass in the things we know we need (i.e. `$NAME`) to start out with and then expand it as more needs come up. 


WDYT?

jbtrystram:
> This particular implementation isn't great (but it does work) because it passes the entire environment which could have some side effects. Likely we should just pass in the things we know we need (i.e. `$NAME`) to start out with and then expand it as more needs come up.
> 

So we'd have to maintain an explicit list of the variables we want injected in the postprocesses in `build-rootfs`.  The LLM initially went there but I changed it because I felt it would be cumbersome to have to update both several places to get a variable piped through. 

I get your point of minimizing the potential side effects, and don't have a strong opinion. I'll make that change.

manifests/fedora-coreos-base.yaml
```diff
@@ -83,6 +84,17 @@ postprocess:
     set -eux -o pipefail
     rm -vrf /usr/lib/modules/*aarch64/dtb/qcom/
 
+  # Inject the ostree stateroot name from the NAME build-arg.
+  # This is done in postprocess rather than an overlay so that the
+  # value is sourced from build-args and shared across FCOS/RHCOS/SCOS.
+  - |
+    #!/usr/bin/bash
+    set -eux -o pipefail
+    cat > /usr/lib/bootc/install/20-stateroot.toml << EOF
```
dustymabe:
```suggestion
    cat > /usr/lib/bootc/install/20-ostree-stateroot.toml << EOF
```

might be good to be more explicit about this being an "ostree" related setting.

dustymabe (dismissed):
LGTM

manifests/fedora-coreos-base.yaml
jbtrystram: Agree ! Applied

dustymabe (approved):

PR #4182, opened 2026-05-15
[testing] overrides: pin kernel-7.0.8-200.fc44

coreosbot-releng:
Requested by @marmijo via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


dustymabe (approved):
LGTM

PR #4185, opened 2026-05-17
tests/upgrade: detect finish_pipe rpm-ostree stall

dustymabe:
RPM-OSTree can hang forever if the remote CDN (or maybe network
environment is acting up) [1]. Let's detect this case and get our
upgrade back on track.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/2149


joelcapitao (approved):

c4rt0 (approved):
That's cool, from what I gathered instead of the test failing like we saw earlier, this would detect the stall and self-heal.

/lgtm

PR #4194, opened 2026-05-27
overlay/bootc: add a root filesystem install config

jbtrystram:
Right now the filesystems are defined in COSA and then `bootc install
to-filesystem` is invoked on the prepared disk.
In the image-builder case, that configuration should come from the
container.
    
This is preliminary work for https://github.com/coreos/fedora-coreos-tracker/issues/1906

### DIFF
--- a/overlay.d/10bootc/usr/lib/bootc/install/12-root-filesystem.toml
+++ b/overlay.d/10bootc/usr/lib/bootc/install/12-root-filesystem.toml
@@ -1,0 +1,6 @@
+# This is actually overwritten by image-builder's
+# disk.yaml partition table definition but is still required
+# Drop this when https://github.com/osbuild/images/issues/2380
+# is fixed.
+[install.filesystem.root]
+type = "xfs"


joelcapitao:
This variable will be overridden by the value provided by `disk.yaml` file that we plan to bake in the container images e.g : https://github.com/coreos/fedora-coreos-config/pull/4195/changes#diff-54e4a8bc62aa1e7f8a8c64199ad6209465a6d97688397d4c831e219aca5f7db0R46
but `image-builder-cli` fails early if that root FS type is not defined in `/usr/lib/bootc/install/` location.
```
Manifest generation step
Building manifest for bootc-based-qcow2
error: failed to initialize bootc distro: missing required info: DefaultRootFs
```
Maybe we should file a bug and make image-builder-cli to also check that information in `/usr/lib/image-builder/bootc/disk.yaml` if available ?

In the meantime we could add a comment explaining this.

jbtrystram:
> This variable will be overridden by the value provided by `disk.yaml` file that we plan to bake in the container images 

Ah yes, good point, I had forgotten that. Let's file a bug

jbtrystram:
Talked today with image-builder folks, we agreed that this is a bug and they will add a way to not error if `disk.yaml` is present with a rootfs filesystem. 

So let's merge it as is to unblock the progress on disk-image building, and we'll revert later when image-builder is fixed.

angelcerveraroldan (approved):
LGTM!

joelcapitao (approved):
LGTM, putting the revert message in my TODO :)

travier (approved):
LGTM. Unfortunately that may confuse some people into thinking that `bootc install to-disk` could work but that likely won't be the case for a while if ever as we focus on image-builder.

jbtrystram:
> LGTM. Unfortunately that may confuse some people into thinking that `bootc install to-disk` could work but that likely won't be the case for a while if ever as we focus on image-builder.

yeah. Hopefully we will revert this sooner than later :)

PR #4195, opened 2026-05-27
image-builder: ship partitions tables in the container

jbtrystram:
Right now the filesystems are defined in COSA and then `bootc install
to-filesystem` is invoked on the prepared disk.
In the image-builder case, that configuration should come from the
container.

This is preliminary work for coreos/fedora-coreos-tracker#1906

Opening this as a draft for now as I'd like some input. Longer time we will probably do away with `image.yaml` almost entirely in favor of files inside the container and image-builder blueprints, and it would be nice for this repo to be the source of truth. Open to feedback here on how to get this started


joelcapitao:
> Opening this as a draft for now as I'd like some input. Longer time we will probably do away with image.yaml almost entirely in favor of files inside the container and image-builder blueprints, and it would be nice for this repo to be the source of truth. Open to feedback here on how to get this started

I guess this paragraph is related to https://github.com/joelcapitao/bib-fcos-experimentation/blob/6775346e51694e842afc5efb69df4ef2eaa61a69/Containerfile#L8-L9 that we want to get rid of, right ?
And this PR is baking that config into the container image already. I like the way how it's done with `build-rootfs`. Initially, I thought implementing this with ostree.

jbtrystram:
> I guess this paragraph is related to [joelcapitao/bib-fcos-experimentation@`6775346`/Containerfile#L8-L9](https://github.com/joelcapitao/bib-fcos-experimentation/blob/6775346e51694e842afc5efb69df4ef2eaa61a69/Containerfile#L8-L9) that we want to get rid of, right ? 

Yes. 

> Initially, I thought implementing this with ostree.

I evaluated doing that with arch-specific overlays but it would quickly become quite verbose and a bit awkward. But both works. We recently [added an arch-specific overlay](https://github.com/coreos/fedora-coreos-config/commit/2f779f9e3291653d53a7e5404682ae2b8ad3f9db), so maybe we can generalize this pattern.

joelcapitao:
Also, with regard to f-c-c being the source of truth containing the image builder blueprints, this intersects with the Konflux workflow topic.

Currently, I see two options:

1. f-c-c containing all the definitions to build the container images _and_ the disk images e.g: blueprints [1], the image-builder config files [2] and the Konflux pipelines (`.tekton/`) that consume those two files to build the disk images.
The automation (MintMaker or custom bot) would push commit to update the image-builder config files [2] containing the source image
The drawback is that the automation would push a lot of commits in f-c-c every day, at least one for every stream per day. If we're ok with that noise, I think that's a good solution

```bash
[1]:
~ cat 4knative-bp.yaml
[customizations.disk]
sector_size = 4096

[2]:
~ cat image-builder-rawhide.yaml
image-builder-cli: "ghcr.io/osbuild/image-builder-cli:latest@sha256:5ee033fafbcbe02834c2f14c552035fa15628eef25025081e2a2971a4934d967"
source-image: "quay.io/coreos-devel/fedora-coreos:rawhide@sha256:80abd9622231bcd5da0d3607dc3a1577810c87e72f6a30c5d75543590ed1c6a1"
```

2. We put all the disk images definitions into a separated repository where all the automation will run. So all the automation noise would be scoped to that repo.
The drawback is that f-c-c won't be only source of truth.

joelcapitao:
> I evaluated doing that with arch-specific overlays but it would quickly become quite verbose and a bit awkward. But both works. We recently [added an arch-specific overlay](https://github.com/coreos/fedora-coreos-config/commit/2f779f9e3291653d53a7e5404682ae2b8ad3f9db), so maybe we can generalize this pattern.

Right, so it depends if we want to keep consistency with the rest of the configuration. Maybe, doing it in `build-rootfs` here might be perceived as a hack. But, as it's metadata used only for the disk generation, I'd say it's ok to not be part of the ostree commit.

jbtrystram:
>     2. We put all the disk images definitions into a separated repository where all the automation will run. So all the automation noise would be scoped to that repo.
>        The drawback is that f-c-c won't be only source of truth.

I  would be fine with that, except that we have to bake in the partition table in the container, image-builder don't read it from elsewhere. And if we do that extra step before creating the disk image we end up with a different container being depoloyed on disk. This may break zincati because the update graph rely on the image digests.
See https://github.com/coreos/fedora-coreos-tracker/issues/2066 from a previous incident from this.

joelcapitao:
> except that we have to bake in the partition table in the container

yes, I'm aligned with that. The list of definitions I put above is exhaustive (`blueprints [1], the image-builder config files [2] and the Konflux pipelines (.tekton/)`), basically everything containing data specific to the tuple of `stream,platform,arch`.
The shared data e.g: the partition table defined in `disk.yaml` file should live in the container, hence defined in f-c-c.

jbtrystram:
Spoke with the image-builder folks today and they suggested to ship a symlink at container build time  if we want to have all the files in the container.

travier:
I think it's OK to only have the config for the architecture of the image in it and not include all arches in all images. I guess both options work. What would be the upside of either? Simpler layout if there is only one?

jbtrystram:
> I think it's OK to only have the config for the architecture of the image in it and not include all arches in all images. I guess both options work. What would be the upside of either? Simpler layout if there is only one?

Shipping everything + a symlink make things slightly easier and allow keeping the files in an overlay, which is nicer from a maintenance perspective IMO, but I don't have strong opinions.

build-rootfs
```diff
@@ -649,5 +650,11 @@ def inject_platforms_json(rootfs, srcdir):
             f.write('\n')
 
 
+def inject_disk_yaml(rootfs, srcdir):
+    src = os.path.join(rootfs, 'usr/lib/image-builder/bootc/', f'{ARCH}.yaml')
+    dst = os.path.join(rootfs, 'usr/lib/image-builder/bootc/disk.yaml')
+    os.symlink(src, dst)
+
+
```
joelcapitao, line 659: I checked the content with:
```bash
$ podman run --rm -it quay.io/konflux-fedora/coreos-tenant/fedora-coreos-testing-devel:on-pr-5cc5a0b825c70d9eab261fd76cb0a5360ce16702 ls -la /usr/lib/image-builder/bootc
total 20
drwxr-xr-x. 1 root root  108 Jun  4 12:21 .
drwxr-xr-x. 1 root root   10 Jun  4 12:21 ..
-rw-r--r--. 2 root root 1861 Jan  1  1970 aarch64.yaml
lrwxrwxrwx. 2 root root   54 Jan  1  1970 disk.yaml -> /target-rootfs/usr/lib/image-builder/bootc/x86_64.yaml
-rw-r--r--. 2 root root 1580 Jan  1  1970 ppc64le.yaml
-rw-r--r--. 2 root root 1483 Jan  1  1970 s390x.yaml
-rw-r--r--. 2 root root 1752 Jan  1  1970 x86_64.yaml

$ podman run --rm -it quay.io/konflux-fedora/coreos-tenant/fedora-coreos-testing-devel:on-pr-5cc5a0b825c70d9eab261fd76cb0a5360ce16702 cat /usr/lib/image-builder/bootc/disk.yaml
cat: /usr/lib/image-builder/bootc/disk.yaml: No such file or directory

$ podman run --rm -it quay.io/konflux-fedora/coreos-tenant/fedora-coreos-testing-devel:on-pr-5cc5a0b825c70d9eab261fd76cb0a5360ce16702 cat /target-rootfs/usr/lib/image-builder/bootc/disk.yaml
cat: /target-rootfs/usr/lib/image-builder/bootc/disk.yaml: No such file or directory
```

Generating the symlink from the built-rootfs does not work as expected. Maybe we should switch to relative symlink instead of absolute. Or directly copy/paste the right file instead of symlinks.

joelcapitao (changes requested):

build-rootfs
jbtrystram: relative symlink will do the trick I guess :)

jbtrystram: Uploaded a fix. let's see what's there when konflux is done

joelcapitao (dismissed):
It's working fine now 👍 

```bash
$ podman run --rm quay.io/konflux-fedora/coreos-tenant/fedora-coreos-testing-devel:on-pr-0d5bc83ea7a813ee811806abf4ae5bf5bfeb17ce-linux-amd64@sha256:13800fe9ef2261db5c2f8a1cd3bb4db88238d4728c6176e2d431322d9bcec82f ls -la /usr/lib/image-builder/bootc
total 20
drwxr-xr-x. 1 root root  108 Jun  4 14:26 .
drwxr-xr-x. 1 root root   10 Jun  4 14:26 ..
-rw-r--r--. 2 root root 1861 Jan  1  1970 aarch64.yaml
lrwxrwxrwx. 2 root root   11 Jan  1  1970 disk.yaml -> x86_64.yaml
-rw-r--r--. 2 root root 1580 Jan  1  1970 ppc64le.yaml
-rw-r--r--. 2 root root 1483 Jan  1  1970 s390x.yaml
-rw-r--r--. 2 root root 1752 Jan  1  1970 x86_64.yaml

$ podman run --rm quay.io/konflux-fedora/coreos-tenant/fedora-coreos-testing-devel:on-pr-0d5bc83ea7a813ee811806abf4ae5bf5bfeb17ce-linux-amd64@sha256:13800fe9ef2261db5c2f8a1cd3bb4db88238d4728c6176e2d431322d9bcec82f cat /usr/lib/image-builder/bootc/disk.yaml
# We don't want osbuild to inject mount units for the
# disks as we will change the disks uuids anyway
mount_configuration: none
partition_table:
  type: "gpt"
  uuid: "00000000-0000-4000-a000-000000000001"
  # Reserve the last 2048 512-byte blocks for CoreOS which moves the
  # secondary GPT header to the end of the disk on first boot.
  # See https://github.com/coreos/fedora-coreos-config/blob/7fbf510367d495529d2a915707b547d4c663f609/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-gpt-setup.sh#L30
  extra_padding: 1031168
[...]
```

jbtrystram:
rebased and pushed again to fix the merge conflict

joelcapitao (approved):

travier:
Does not work with RHCOS: https://github.com/coreos/rhel-coreos-config/pull/245

build-rootfs
```diff
@@ -649,5 +650,12 @@ def inject_platforms_json(rootfs, srcdir):
             f.write('\n')
 
 
+def inject_disk_yaml(rootfs, srcdir):
+    dst = os.path.join(rootfs, 'usr/lib/image-builder/bootc/disk.yaml')
```
travier, line 654: This should warn and ignore this step if the folder does not exists (i.e. RHCOS)

travier:
Untested fix in https://github.com/coreos/fedora-coreos-config/pull/4217

PR #4197, opened 2026-05-29
[rawhide] overrides: pin systemd-260.1-2.fc45

coreosbot-releng:
Requested by @angelcerveraroldan via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


dustymabe (approved):
LGTM

PR #4199, opened 2026-06-01
tests/kola: Drop lockdown with Secure Boot disabled test

### LINKED ISSUE #1966: lockdown kola tests can't co-exist
Right now if you try to run the lockdown tests they can't co-exist:

```
$ cosa kola run *lockdown*
kola -p qemu run *lockdown* --output-dir tmp/kola
⏭️  Skipping kola test pattern "ext.config.multipath.resilient":
  👉 https://github.com/coreos/fedora-coreos-tracker/issues/1937
=== RUN   non-exclusive-test-bucket-0/ext.config.security.lockdown_no_sb
=== RUN   non-exclusive-test-bucket-0/ext.config.security.lockdown
systemctl status kola-runext-64.service:
× kola-runext-64.service
     Loaded: loaded (/etc/systemd/system/kola-runext-64.service; static)
             └─10-timeout-abort.conf
     Active: failed (Result: exit-code) since Wed 2025-06-04 20:42:21 UTC; 130ms ago
   Duration: 26ms
 Invocation: 537ec46c2f32465c9ae3e20a5bec5d0b
    Process: 2291 ExecStart=/usr/local/bin/kola-runext-lockdown (code=exited, status=1/FAILURE)
   Main PID: 2291 (code=exited, status=1/FAILURE)
   Mem peak: 1.4M
        CPU: 9ms

Jun 04 20:42:21 qemu0 systemd[1]: Started kola-runext-64.service.
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: + . /var/opt/kola/extdata-64/commonlib.sh
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: ++ IFS=' '
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: ++ read -r -a cmdline
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2292]: ++ cat /sys/kernel/security/lockdown
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: + lockdown_state='[none] integrity confidentiality'
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2293]: ++ mokutil --sb
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2293]: EFI variables are not supported on this system
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: + [[ '' == \S\e\c\u\r\e\B\o\o\t\ \e\n\a\b\l\e\d ]]
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: + fatal 'system is not running with Secure Boot enabled'
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: + echo 'system is not running with Secure Boot enabled'
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: system is not running with Secure Boot enabled
Jun 04 20:42:21 qemu0 kola-runext-lockdown[2291]: + exit 1
Jun 04 20:42:21 qemu0 systemd[1]: kola-runext-64.service: Main process exited, code=exited, status=1/FAILURE
Jun 04 20:42:21 qemu0 systemd[1]: kola-runext-64.service: Failed with result 'exit-code'.
--- FAIL: non-exclusive-test-bucket-0 (26.96s)
    --- PASS: non-exclusive-test-bucket-0/ext.config.security.lockdown_no_sb (1.48s)
    --- FAIL: non-exclusive-test-bucket-0/ext.config.security.lockdown (0.66s)
            cluster.go:151: Error: Unit kola-runext-64.service exited with code 1
            cluster.go:151: 2025-06-04T20:42:21Z cli: Unit kola-runext-64.service exited with code 1
            harness.go:1269: kolet failed: : kolet run-test-unit failed: Process exited with status 1
FAIL, output in tmp/kola
2025-06-04T20:42:25Z cli: harness: test suite failed

```

The small fix to this problem is to just mark the one that requires secureboot as `exclusive: true` so it can be started with its own VM (so the thing that applies the firmware setting from https://github.com/coreos/coreos-assembler/pull/4114 will get applied).

But we also have the problem for the `ext.config.security.lockdown_no_sb` one. If we changed the default kola test to run UEFI and not BIOS then this would start failing. We should:

- mark it as `exclusive: true` too; since we don't want to rely on the default of `kola` for this test to pass or fail
- augment COSA to add a similar tag for making sure secureboot is off
    - better would probably be adding a new test metadata field of `firmware: <string>`, so we could choose all the different values that `--qemu-firmware` supports.

travier:
Testing the default value when Secure Boot is not enabled does not
really add value thus drop this test.

Fixes: coreos/fedora-coreos-tracker#1966


dustymabe (approved):
LGTM

PR #4200, opened 2026-06-01
manifests: Move systemd-gpt-auto-generator removal to a manifest shared with EL

travier:
We have our own generator for /boot and we don't mount /boot/efi by default thus we don't strictly need systemd-gpt-auto-generator on Fedora CoreOS.

With https://github.com/systemd/systemd/issues/13099 fixed, we can consider including it again after more testing.

Current known issues are:

- systemd-gpt-auto-generator conflicts with Ignition LUKS provisioning on EFI systems. See: https://github.com/coreos/fedora-coreos-tracker/issues/2137

Move this workaround to a manifest file that is shared with EL as we need it for SCOS/RHCOS as well.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2137

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -133,12 +133,3 @@ postprocess:
         echo "${actuals}" 1>&2
         exit 1
     fi
-
-  # We don't want auto-generated mount units. See also
-  # https://github.com/systemd/systemd/issues/13099
-  - |
-    #!/usr/bin/env bash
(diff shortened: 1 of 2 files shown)


jbtrystram (approved):
LGTM ! 
This one needs to be included into rhel-9.6 as well

travier:
RHEL 9.6 backport in https://github.com/coreos/fedora-coreos-config/pull/4214

PR #4201, opened 2026-06-02
tests/kola: add soft-reboot test for LUKS root encryption

prestist:
Validate that soft-reboot works correctly when the root filesystem is on a LUKS device. This exercises the x-initrd.attach crypttab option added in coreos/ignition#2219.


prestist:
I did end up verifying this test works locally as expected. i.e fails without ignition change and passes when ignition is provided with the latest change found in[ ignition/2219](https://github.com/coreos/ignition/pull/2219)

travier:
Hum, I'm not sure I follow. Where do we set that ostree should do a soft reboot here? Don't we need to create something that trigger an update?

travier:
I think you should be able to do the workaround manually in this test for now (i.e. updating the `/etc/fstab` entry), thus proving that the fix is correct.

prestist:
> Hum, I'm not sure I follow. Where do we set that ostree should do a soft reboot here? Don't we need to create something that trigger an update?

From what I could tell the bug is systemd-cryptsetup-generator adding Conflicts=umount.target — that conflict fires during any soft-reboot, regardless of what triggered it, so I figured we could avoid the heavy'er path of using ostree. We can see the same error repo'd with these changes in the ci.

prestist:
@travier with the workaround the test passes, what should be the next steps?

tests/kola/root-reprovision/luks/soft-reboot/test.sh
```diff
@@ -0,0 +1,49 @@
...
+  "")
+      srcdev=$(findmnt -nvr /sysroot -o SOURCE)
+      [[ ${srcdev} == /dev/mapper/myluksdev ]]
+
+      blktype=$(lsblk -o TYPE "${srcdev}" --noheadings)
+      [[ ${blktype} == crypt ]]
+      ok "root is on LUKS device"
+
+      # Workaround: manually add x-initrd.attach to crypttab if not
+      # already present. This can be removed once coreos/ignition#2219
```
marmijo, line 29: nit: since this is intentionally temporary and depends on the ignition PR, could we use the full PR URL here? It'll make it easier for someone cleaning this up later to find the dependency directly.

marmijo (approved):
LGTM

tests/kola/root-reprovision/luks/soft-reboot/test.sh
Rolv-Apneseth: Could we also have be a `TODO:`? Idk about others but for me these are a lot more visible, and I even have a plugin to display those in a list.

With it being so long since this was opened though, can that just be dropped outright?

PR #4202, opened 2026-06-05
tests: remove reprovision tag from kola tests

dustymabe:
With resource based scheduling [1] we should now be able to run all the tests together and not split them up into separate !reprovision and reprovision runs.

[1] https://github.com/coreos/coreos-assembler/pull/4470

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -8,11 +8,7 @@
 ##   # We do a bunch of copies/rebuilds of the whole OS.
 ##   minDisk: 20
 ##   # Needs internet access as we fetch files from koji
-##   # We add the "reprovision" tag here even though we aren't
-##   # reprovisioning as a hack so that in our pipeline the test
-##   # will execute with no other tests. We were seeing a lot of
-##   # timeouts on ppc64le.
-##   tags: "needs-internet platform-independent reprovision"
+##   tags: "needs-internet platform-independent"
 ##   description: Verify that build of a container image with a new kernel
 ##     and reboot into it succeeds.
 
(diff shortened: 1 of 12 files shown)


tests/kola/rpm-ostree/kernel-replace
```diff
@@ -8,11 +8,7 @@
 ##   # We do a bunch of copies/rebuilds of the whole OS.
 ##   minDisk: 20
 ##   # Needs internet access as we fetch files from koji
-##   # We add the "reprovision" tag here even though we aren't
-##   # reprovisioning as a hack so that in our pipeline the test
-##   # will execute with no other tests. We were seeing a lot of
-##   # timeouts on ppc64le.
-##   tags: "needs-internet platform-independent reprovision"
+##   tags: "needs-internet platform-independent"
```
c4rt0, line 11: Just because I only looked at the `coreos-ci-lib` [PR](https://github.com/coreos/coreos-ci-lib/pull/172#pullrequestreview-4451041808) I wonder if removing `reprovision` from kernel replace is something we want? Seems to me like a resource intensive exercise.

dustymabe: with https://github.com/coreos/coreos-assembler/pull/4470 at least for the memory side of things `kola` should only schedule when enough memory is available.

"resource" is a generic term, though. and there are a few different types:

- network
- memory
- cpu
- disk I/O

and you are right that kernel-replace is resource intensive on at least the latter 3.

I think cpu and memory should be covered by kola scheduling (i.e. --parallel is only set to the number of CPUs so each job should get its own CPU and now memory should be guaranteed too), so that leaves disk I/O. 

I'd say let's try to not special case kernel-replace and then if we have trouble (i.e. the test starts to become flaky) we can adjust the strategy.

c4rt0 (approved):

tests/kola/rpm-ostree/kernel-replace
c4rt0: Makes sense, thanks for the deep-dive.

PR #4207, opened 2026-06-11
tests: add raid1-boot test for ppc64le

Rolv-Apneseth:
Final piece in relation to https://github.com/coreos/fedora-coreos-tracker/issues/2134.

With f69730095caa645627bef97a1c83a5aa7d40c5f6, we now have the `grub2` version which includes https://github.com/rhboot/grub2/pull/212 (`2.12-60.fc44`). This test should now work as intended on our ppc64le builds.

Confirmed the test passes with a ppc64le debug pod.


dustymabe:
Nice. I imagine we'll need to denylist this test downstream OR does the grub there have the fix too?

Rolv-Apneseth:
Likely we will need to denylist downstream until the fix shows up there too since AFAICT it's not there yet. Do you want me to open that before we merge this?

Rolv-Apneseth:
Seems I'll need to do some more porting for rawhide and rhel. I'll get those done first so updates don't take too long.

tests/kola/boot/raid1-boot/ppc64le/test.sh
```diff
@@ -0,0 +1,21 @@
...
+##   # Root reprovisioning requires at least 4GiB of memory.
+##   minMemory: 4096
+##   # RAID 1 is setup on these disks.
+##   additionalDisks: ["10G"]
+##   # This test includes a lot of disk I/O and needs a higher
+##   # timeout value than the default.
+##   timeoutMin: 15
+##   architectures: "ppc64le"
+##   # This test reprovisions the boot and rootfs.
+##   tags: reprovision
```
c4rt0: Hmm, in [4202](https://github.com/coreos/fedora-coreos-config/pull/4202) we removed this tag from all tests.

```suggestion

c4rt0 (commented):
Other than the comment, this LGTM

tests/kola/boot/raid1-boot/ppc64le/test.sh
Rolv-Apneseth: Oops, I'll remove that

c4rt0 (approved):
/lgtm

dustymabe:
> Likely we will need to denylist downstream until the fix shows up there too since AFAICT it's not there yet. Do you want me to open that before we merge this?

it would be useful to be pro-active about it, yes.

this LGTM. Feel free to merge when you are ready.

Rolv-Apneseth:
I'm thinking to wait until at least the rawhide package is updated so we don't need to denylist here. There's no rush for this as far as I know, but I'll be trying to get those patches ported as soon as I get the chance

Rolv-Apneseth:
The rawhide update is available but we've been unable to pick it up in our builds due to https://github.com/coreos/fedora-coreos-tracker/issues/2171. When that's resolved and rawhide passes this test, I'll open a PR to denylist  downstream and merge this PR.

Rolv-Apneseth:
Rawhide now passes this test. Opened https://github.com/coreos/rhel-coreos-config/issues/281 for downstream, where I'll create a denylist entry.

Rolv-Apneseth:
https://github.com/coreos/rhel-coreos-config/pull/282

PR #4209, opened 2026-06-15
kola/test/pqc: Check that PQC algorithms are used by default for TLS

angelcerveraroldan:
Check that PQC algorithms are prioritised when doing TLS key exchange negotiation.

### DIFF
--- a/tests/kola/pqc/data/commonlib.sh
+++ b/tests/kola/pqc/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../data/commonlib.sh
--- a/tests/kola/pqc/test.sh
+++ b/tests/kola/pqc/test.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+## kola:
+##   tags: "needs-internet"
+##   description: Check that PQC algorithms are prioritised by default (TLS)
+##   creationDate: 2026-06-11
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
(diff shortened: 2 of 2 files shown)


travier:
Looks good. Unfortunately that will require some config changes to let the RHCOS pipeline open connections to those hosts/IPs. Not sure if it's worth doing now that you've confirmed that PQC algorithms are enabled by default in the policy (https://redhat.atlassian.net/browse/COS-4367).

angelcerveraroldan:
> Looks good. Unfortunately that will require some config changes to let the RHCOS pipeline open connections to those hosts/IPs. Not sure if it's worth doing now that you've confirmed that PQC algorithms are enabled by default in the policy (https://redhat.atlassian.net/browse/COS-4367).

I used the same links this test [1], would that not mean that they are already allowed ? I left some more context in the Jira ticket, but I was also working on a test using the `FUTURE` cypto policy which as stricter rules (i.e. TLS1.2 is not allowed).

I am not sure if this policy is used often or not by us - maybe its not worth adding the test.

[1] https://github.com/coreos/fedora-coreos-config/blob/941a8881366fbdece3fe3af92240917f62aea6a7/tests/kola/networking/tls#L13

travier:
Ah, good point, then this should be just fine

travier (dismissed):

travier:
For `FUTURE`, I don't know. I'm not sure if it's worth testing it.

travier (approved):

PR #4211, opened 2026-06-17
overlay/05core: Drop coreos-liveiso-success.service

dustymabe:
Originally added in e8abbe2, it was used to confirm autologin worked but the heuristc (`loginctl list-sessions --no-legend`) stopped working correctly [1] so we updated the test in COSA to just inspect the console.txt instead [2].

We can drop this now.

[1] https://github.com/coreos/coreos-assembler/issues/4581
[2] https://github.com/coreos/coreos-assembler/pull/4582


nikita-dubrovskii (approved):
LGTM

PR #4213, opened 2026-06-18
overlays/bootc: Move stateroot to FCOS-specific overlay

jbtrystram:
Move this postprocess to a manifest shared with rhel-9.6 as this service cause FIPS self tests to fail.

See https://redhat.atlassian.net/browse/OCPBUGS-86442


travier:
I had done https://github.com/coreos/fedora-coreos-config/pull/4200

PR #4214, opened 2026-06-18
[rhel-9.6] manifests: Move systemd-gpt-auto-generator removal to a manifest shared with EL

travier:
We have our own generator for /boot and we don't mount /boot/efi by default thus we don't strictly need systemd-gpt-auto-generator on Fedora CoreOS.

With https://github.com/systemd/systemd/issues/13099 fixed, we can consider including it again after more testing.

Current known issues are:

- systemd-gpt-auto-generator conflicts with Ignition LUKS provisioning on EFI systems. See: https://github.com/coreos/fedora-coreos-tracker/issues/2137

Move this workaround to a manifest file that is shared with EL as we need it for SCOS/RHCOS as well.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2137

### DIFF
--- a/manifests/ignition-and-ostree.yaml
+++ b/manifests/ignition-and-ostree.yaml
@@ -15,3 +15,18 @@ packages:
 packages-s390x:
   # for Secure Execution
   - veritysetup
+
+postprocess:
+  # We have our own generator for /boot and we don't mount /boot/efi by default
+  # thus we don't strictly need systemd-gpt-auto-generator on Fedora CoreOS.
+  # With https://github.com/systemd/systemd/issues/13099 fixed, we can consider
+  # including it again after more testing. Current known issues are:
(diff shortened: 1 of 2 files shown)


manifests/ignition-and-ostree.yaml
```diff
@@ -15,3 +15,18 @@ packages:
 packages-s390x:
   # for Secure Execution
   - veritysetup
+
+postprocess:
+  # We have our own generator for /boot and we don't mount /boot/efi by default
+  # thus we don't strictly need systemd-gpt-auto-generator on Fedora CoreOS.
```
c4rt0, line 21: Just a minor nit, since we're merging this to rhel-9.6, we should probably change `Fedora CoreOS` here in the future to something like `on CoreOS derived systems`.

c4rt0 (approved):
Backport of #4200. 

/lgtm

jbtrystram (approved):
Lgtm

PR #4218, opened 2026-06-18
manifests: Move openssl to a manifest shared with EL

travier:
Starting with EL 10, openssl (the binary) is not longer pulled into the image via dependencies. Move it to a manifest that is shared with EL variants to keep it explicitely included.

See: https://redhat.atlassian.net/browse/OCPBUGS-90400


ravanelli (approved):
/lgtm

travier:
Let's put this one on hold until the requirements for this binary are cleared.

travier:
We decided to not include openssl back in EL for now.

PR #4220, opened 2026-06-19
fcos-container-signing: Ensure /etc/containers/policy.json exists

### LINKED ISSUE #2171: [rawhide] Multiple tests failing when running the bump-lockfile job
Running the `bump-lockfile` job leads to dozens of test failures. One recurring log in the failing tests is:

```
Jun 18 09:58:50 qemu0 kola-runext-kernel-replace[2006]: error: Skopeo copy: skopeo failed: time="2026-06-18T09:58:50Z" level=fatal msg="Error loading trust policy: invalid policy in \"/etc/containers/policy.json\": unexpected end of JSON input"
```

`containers-common` used to install `policy.json` into `/etc/containers`, but now it will install it at `/usr/share/containers`

We are creating a bind mount over `/etc/containers/policy.json` that is only visible in `rpm-ostreed` mountspace. When `/etc/containers/policy.json` does not exists, systemd will create an empty file before making the mount.

https://github.com/coreos/fedora-coreos-config/blob/33622f8b441eaee5a4c31ddeed8dab1dcaa70cb9/overlay.d/17fcos-container-signing/usr/lib/systemd/system/rpm-ostreed.service.d/coreos-containers-policy.conf#L7

So after `rpm-ostreed` runs, we will have an empty file `/etc/containers/policy.json`. Tools like `podman` will try to parse this empty json file and error out.

# New image with packages bumped

```
[core@cosa-devsh ~]$ ls -lahZ /etc/containers/policy.json
-rw-r--r--. 1 root root system_u:object_r:etc_runtime_t:s0 0 Jun 18 13:33 /etc/containers/policy.json
[core@cosa-devsh ~]$ cat /etc/containers/policy.json
[core@cosa-devsh ~]$ cat /usr/share/containers/policy.json
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports":
        {
            "docker-daemon":
                {
                    "": [{"type":"insecureAcceptAnything"}]
                }
        }
}
```

# Older Image

```
[core@cosa-devsh ~]$ ls -lahZ /etc/containers/policy.json 
-rw-r--r--. 1 root root system_u:object_r:etc_t:s0 256 Aug  1  2022 /etc/containers/policy.json
[core@cosa-devsh ~]$ cat /etc/containers/policy.json 
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports":
        {
            "docker-daemon":
                {
                    "": [{"type":"insecureAcceptAnything"}]
                }
        }
}
[core@cosa-devsh ~]$ cat /usr/share/containers/policy.json
cat: /usr/share/containers/policy.json: No such file or directory
```

angelcerveraroldan:
New versions of containers-common no longer install policy.json in /etc/containers. Because of this, when coreos-containers-policy.conf creates a bind mount, systemd will first create an empty file /etc/containers/policy.json. After that, tools like podman and skopeo will error out when they try to parse that empty json file.

Closes: coreos/fedora-coreos-tracker#2171

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -29,6 +29,24 @@ conditional-include:
     include:
       ostree-layers:
         - overlay/10aarch64
+  # Temporary workaround. containers-common moved policy.json from
+  # /etc/containers to /usr/share/containers. The rpm-ostreed bind mount
+  # could use CONTAINERS_POLICY_JSON env var instead, but bootc has a
+  # hardcoded path that prevents this approach.
+  # Copy the file until bootc supports the new config location.
+  #
+  # See:
+  # https://github.com/bootc-dev/bootc/issues/2258
+  # https://github.com/coreos/fedora-coreos-config/issues/4215
+  - if: releasever == 45
+    include:
+      postprocess:
+        - |
+          #!/usr/bin/bash
+          set -eux -o pipefail
+          src=/usr/share/containers/policy.json
+          dest=/etc/containers/policy.json
+          cp "${src}" "${dest}"
 
 ostree-layers:
   - overlay/15fcos


joelcapitao:
I tested this PR locally and works fine: 

```
core@cosa-devsh:~$ rpm -qa | grep -e podman -e skopeo -e containers
podman-sequoia-0.3.2-2.fc45.x86_64
containers-common-0.68.0-1.fc45.noarch
containers-common-extra-0.68.0-1.fc45.noarch
skopeo-1.23.0-3.fc45.x86_64
podman-6.0.0~rc1-2.fc45.x86_64
core@cosa-devsh:~$ cat /etc/containers/policy.json
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports":
        {
            "docker-daemon":
                {
                    "": [{"type":"insecureAcceptAnything"}]
                }
        }
}
core@cosa-devsh:~$ ll /etc/containers/
total 4
drwxr-xr-x. 2 root root   6 Jun 19 13:21 certs.d
drwxr-xr-x. 2 root root   6 Jun 19 13:21 networks
drwxr-xr-x. 3 root root  21 Jun 19 13:21 oci
lrwxrwxrwx. 1 root root  33 Jun 19 13:21 policy.json -> /usr/share/containers/policy.json
drwxr-xr-x. 2 root root   6 Jun 19 13:21 registries.conf.d
drwxr-xr-x. 2 root root  32 Jun 19 13:21 registries.d
drwxr-xr-x. 2 root root   6 Jun 19 13:21 systemd
-rw-r--r--. 1 root root 699 Jun 19 13:21 toolbox.conf
core@cosa-devsh:~$ QEMU: Terminated
[coreos-assembler]$ kola run ext.config.toolbox coreos.misc.disk.varlibcontainers podman.workflow ext.config.podman.rootless-systemd ext.config.ntp.timesyncd.dhcp-propagation podman.network-single ext.config.ntp.chrony.dhcp-propagation podman.base coreos.misc.disk.varlibcontainers ext.config.podman.rootless-pasta-networking ext.config.selinux.podman-tmpfs-context ext.config.selinux.file-context-policy-match ext.config.podman.dns
=== RUN   ext.config.podman.rootless-systemd
=== RUN   podman.base
=== RUN   ext.config.toolbox
=== RUN   coreos.misc.disk.varlibcontainers
=== RUN   podman.network-single
=== RUN   podman.workflow
=== RUN   ext.config.ntp.chrony.dhcp-propagation
=== RUN   ext.config.ntp.timesyncd.dhcp-propagation
--- PASS: ext.config.podman.rootless-systemd (50.37s)
--- PASS: coreos.misc.disk.varlibcontainers (41.97s)
        cluster.go:142: Created symlink '/etc/systemd/system/var-lib-containers.mount.wants/systemd-mkfs@dev-disk-by-id-virtio-disk1.service' → '/etc/systemd/system/systemd-mkfs@dev-disk-by-id-virtio-disk1.service'.
        cluster.go:142: Created symlink '/etc/systemd/system/local-fs.target.wants/var-lib-containers.mount' → '/etc/systemd/system/var-lib-containers.mount'.
=== RUN   podman.workflow/run
=== RUN   podman.workflow/exec
=== RUN   podman.workflow/stop
=== RUN   podman.workflow/remove
=== RUN   podman.workflow/delete
--- PASS: podman.workflow (340.60s)
    --- PASS: podman.workflow/run (305.95s)
            cluster.go:142: Trying to pull quay.io/fedora/fedora:latest...
            cluster.go:142: Getting image source signatures
            cluster.go:142: Copying blob sha256:907016b89e86a1e461b5de341114d1a9f611282134c09c33f683f235884904d8
            cluster.go:142: Copying config sha256:34dfcd46f32e236f1a8260899f7faf6a4dd2eac76a831f8639b173b8bce46de4
            cluster.go:142: Writing manifest to image destination
    --- PASS: podman.workflow/exec (0.25s)
    --- PASS: podman.workflow/stop (10.36s)
    --- PASS: podman.workflow/remove (0.26s)
    --- PASS: podman.workflow/delete (0.25s)
--- PASS: podman.network-single (143.77s)
--- PASS: ext.config.ntp.timesyncd.dhcp-propagation (53.48s)
=== RUN   non-exclusive-test-bucket-0/ext.config.podman.rootless-pasta-networking
=== RUN   non-exclusive-test-bucket-0/ext.config.selinux.podman-tmpfs-context
--- PASS: non-exclusive-test-bucket-0 (33.78s)
    --- PASS: non-exclusive-test-bucket-0/ext.config.podman.rootless-pasta-networking (5.46s)
    --- PASS: non-exclusive-test-bucket-0/ext.config.selinux.podman-tmpfs-context (4.44s)
--- PASS: ext.config.toolbox (57.97s)
--- PASS: ext.config.ntp.chrony.dhcp-propagation (53.34s)
=== RUN   podman.base/info
=== RUN   podman.base/resources
--- PASS: podman.base (26.05s)
    --- PASS: podman.base/info (0.21s)
    --- PASS: podman.base/resources (2.62s)
PASS, output in tmp/kola/qemu-2026-06-19-1339-10960
```

joelcapitao (dismissed):

travier:
That's likely not what we want. What's the change behind the file being moved?

travier:
I think this comes from:
- https://github.com/podman-container-tools/container-libs/pull/795
- https://github.com/podman-container-tools/container-libs/pull/711/changes#diff-e15df7e1bf93bd0730afca2458c9f40818ba500cc1f29701a2b940bcab4b621f

travier:
I think we should use `CONTAINERS_POLICY_JSON` instead now that it exists.

travier:
I've filed https://github.com/bootc-dev/bootc/issues/2258 for bootc

angelcerveraroldan:
> I think we should use `CONTAINERS_POLICY_JSON` instead now that it exists.

I had not seen this, thanks. I assume that we should use `CONTAINERS_POLICY_JSON` instead of doing the bind mount in the `rpm-ostreed` service ? This way we also allow the user to set their own `/etc/containers/policy.json` files and not have them be ignored.

travier:
Yes. The config location / load order change is not a complete fix for our issue/workaround for rpm-ostree but at least it's an improvement over the previous status.

angelcerveraroldan:
Replacing  `BindReadOnlyPaths` with `Environment=CONTAINERS_POLICY_JSON=/usr/lib/coreos/coreos-containers-policy.json` in the rpm-ostreed drop-in resolved the ghost file problem.

Also, podman and skopeo work fine, since there is no file at `/etc/containers/policy.json`, they fall back to `/usr/share/containers/policy.json`, which is where the file is now located.

However, because the path is hardcoded to only `/etc/containers/policy.json` in bootc, we will start hitting different issues:

```
[core@cosa-devsh ~]$ sudo rpm-ostree upgrade
Pulling manifest: ostree-image-signed:docker://quay.io/fedora/fedora-coreos:rawhide
error: Preparing import: Fetching manifest: No such file or directory (os error 2)
```

I can think of two ways to unblock:
1. temporarily use the symlink and keep the bind mount as is
2. pin packages to avoid issues

travier:
I'm not a fan of the symlink approach as this changes the type of the "file" here and that could trigger other issues. I don't know how that case is handled by ostree on updates. Likely correctly but I've never tried. Maybe we should copy the config to `/etc` temporarily in our images instead as that would preserve the existing semantic and we'll remove it once we're ready.

angelcerveraroldan:
Sounds good. I'll look into copying the file into our image directly.

I assume that the concern is to do with the 3 way merge, where if we are adding the symlink/copying using `tmpfile.d`, then a later update may not be able to change these files since they're created during boot.

travier:
> Sounds good. I'll look into copying the file into our image directly.

👍🏻 

> I assume that the concern is to do with the 3 way merge, where if we are adding the symlink/copying using `tmpfile.d`, then a later update may not be able to change these files since they're created during boot.

Yes, we can not do a tmpfiles.d config for either a copy or symlink as it would make the file a local change from the ostree perspective and it would then never be removed on update.

Similarly, if we turned this file into a symlink then we would have to verify that it would not break users that currently write their own policy there. So I don't think we should do that.

cgwalters:
> However, because the path is hardcoded to only /etc/containers/policy.json in bootc, we will start hitting different issues:

This seems like an easy fix to make

manifests/fedora-coreos.yaml
dustymabe, line 53: need to also make `17fcos-container-signing` conditional (now on <= 45)

manifests/fedora-coreos.yaml
```diff
@@ -29,6 +29,28 @@ conditional-include:
...
+  # https://github.com/bootc-dev/bootc/issues/2258
+  # https://github.com/coreos/fedora-coreos-config/issues/4215
+  - if: releasever == 45
+    include:
+      postprocess:
+        - |
+          #!/usr/bin/bash
+          set -eux -o pipefail
+          src=/usr/share/containers/policy.json
+          if [ -f /run/.containerenv ]; then
```
dustymabe: this is always true now. We don't need to conditionial logic for it.

manifests/fedora-coreos.yaml
angelcerveraroldan: Would we not still need most of `17fcos-container-signing` in f46 ?

i.e. would we not still need the signing keys and the the repositories.d entry ?

dustymabe: ahh yes. you are right!

dustymabe (approved):
LGTM

PR #4217, opened 2026-06-18
build-rootfs: Only symlink existing image builder configs

travier:
The image-builer disk configuration have only been added to FCOS for now. Skip this step on RHCOS.

Fixes: https://github.com/coreos/fedora-coreos-config/pull/4195
Fixes: af0d6638 build-rootfs: inject disk partition layouts for image-builder

PR #4222, opened 2026-06-19
build-rootfs: Fix image builder config detection logic

travier:
Use the directory existing as the signal that the configuration will exist instead of a specific file. Fixes previous check that incorrectly looked for the symlink name instead of its target.

Fixes: https://github.com/coreos/fedora-coreos-config/pull/4217
Fixes: https://github.com/coreos/fedora-coreos-config/pull/4195
Fixes: af0d6638 build-rootfs: inject disk partition layouts for image-builder

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -651,13 +651,13 @@ def inject_platforms_json(rootfs, srcdir):
 
 
 def inject_disk_yaml(rootfs, srcdir):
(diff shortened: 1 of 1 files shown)


jbtrystram (approved):
Nice catch, thanks. I forgot this was synced to rhcos as-is.
LGTM

travier:
I can see the right line in the logs this time:
```
Creating symlink to arch specific image-builder config: 'disk.yaml' -> 'x86_64.yaml'
```

PR #4226, opened 2026-06-22
tests/upgrade: try for more console messages on shutdown

dustymabe:
Let's disable coreos-printk-quiet.service so hopefully we'll
get more messages from the shut down of an instance on each
upgrade transition.


nikita-dubrovskii (dismissed):
LGTM

nikita-dubrovskii (approved):

PR #4227, opened 2026-06-23
rework manifests/overlays

dustymabe:
Some rework of the manfifests and overlays here. See individual commit messages.

This should enable us to define more downstream shared definitions here in this repo without having to symlink or redefine them downstream.


dustymabe:
I still need to do a local run with and without this change to make sure there are no significant differences.

travier:
Approach sounds good. This should simplify the manifests in rhel-coreos-config if I understand correctly. Can you make a companion PR for that?

I have not checked if the move was exhaustive so we'll have to make sure that there are no or next to no change from this change.

dustymabe:
> Approach sounds good. This should simplify the manifests in rhel-coreos-config if I understand correctly. Can you make a companion PR for that?

yep. would look something like https://github.com/coreos/rhel-coreos-config/pull/276


> 
> I have not checked if the move was exhaustive so we'll have to make sure that there are no or next to no change from this change.

yeah. I mentioned in https://github.com/coreos/fedora-coreos-config/pull/4227#issuecomment-4775352112 basically will need to `cosa diff` thoroughly before merging this.

build-rootfs
jbtrystram, line 1: I find `overlays` to be more intuitive here given that the actual directories are prefixed `overlay` and how the contents are merged into the tree is like overlays are.
I understand it's not technically correct but `dirs` is generic and does not really give much information, while `overlay` does. 
Maybe the bootc option should be `merge-tree` to be more explicit :) 

/end of bikeshedding

dustymabe: yeah. maybe the bootc-base-imagectl `--add-dir` should be changed to `--overlay-dir` ? If you want to PR that change we can then update it here?

dustymabe:
> yeah. I mentioned in [#4227 (comment)](https://github.com/coreos/fedora-coreos-config/pull/4227#issuecomment-4775352112) basically will need to `cosa diff` thoroughly before merging this.

I ran the diffs for this repo and everything looked good. Marking as ready for review now.

Will also run the diffs for some `rhel-coreos-config` builds.

dustymabe:
> Will also run the diffs for some `rhel-coreos-config` builds.

These look good too

build-rootfs
```diff
@@ -453,19 +456,17 @@ def inject_version_info(rootfs, replace_version, version):
             f.write(f'{k}={v}\n')
 
 
-# This re-implements cosa's overlay logic.
-def gather_overlays(manifest, srcdir):
-    overlays = []
+def gather_add_dirs(manifest, srcdir):
+    add_dirs = []
     for layer in manifest.get('ostree-layers', []):
```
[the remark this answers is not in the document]
dustymabe: we'll fix it when that happens. For now I don't want to spin the CI cycle on it.

manifests/shared-el.yaml
```diff
@@ -28,3 +36,20 @@ conditional-include:
         - azure-vm-utils
         # Include udev rules azure storage and productid
         - WALinuxAgent-udev
+
+  # Include the layers that are shared amongst Fedora and EL
+  # (though at different import paths).
+  - if: id == "fedora"
+    include:
+      ostree-layers:
```
yasminvalim, line 44: I noticed that with this split files updated change the merge order: 09misc and 10disk-images move from between 08nouveau and 20platform-chrony to after 30lvmdevices. Is not a big thing and since you did the cosa diffs I guess will not cause any problems, but I am just asking to be sure if its intentional our could it cause problems if later layers override earlier ones?

yasminvalim (approved):
Left a question in the review, but otherwise looks good to me!  :+1:

manifests/shared-el.yaml
dustymabe: I had set this to automerge so it merged :).

That's a good question. Right now I don't think any of our "layers" conflict with each other, so it doesn't matter.

I could foresee us wanting to enforce the ordering in the future though if we wanted to override something from an earlier layer and I imagine the numbering would hint at that being needed.

dustymabe: added a commit over in https://github.com/coreos/fedora-coreos-config/pull/4230#issuecomment-4795605349 to address the problem.

jbtrystram (approved):

manifests/fedora-coreos.yaml
```diff
@@ -2,7 +2,19 @@
...
+recommends: false
+documentation: false
+
+include:
+  - system-configuration.yaml
+  - ignition-and-ostree.yaml
+  - file-transfer.yaml
+  - networking-tools.yaml
+  - user-experience.yaml
+  - shared-el.yaml
```
travier, line 15: While we're doing sharing, should we do it for https://github.com/coreos/rhel-coreos-config/blob/main/common.yaml#L3..L9 as well?

travier: and it's done in the next commit 😅

travier (commented):
LGTM

PR #4230, opened 2026-06-23
add virtiofs systemd credential handling

dustymabe:
As part of an effort to test more on non-CoreOS bootable
    container images we'd like a cross architecture way to
    expose systemd credentials into a VM. This commit adds a
    new import-virtiofs-systemd-credentials.service that runs
    in the initramfs and will copy credentials exposed via the
    io.systemd.credentials virtiofs tag into the /run/credentials/@initrd
    directory, which will then get imported by systemd itself.

    Related to https://github.com/systemd/systemd/issues/29175 upstream.
    
    Assisted-by: <anthropic/claude-opus-4.6>

### DIFF
--- a/overlay.d/40import-virtiofs-systemd-credentials/usr/lib/dracut/modules.d/40import-virtiofs-systemd-credentials/import-virtiofs-systemd-credentials.sh
+++ b/overlay.d/40import-virtiofs-systemd-credentials/usr/lib/dracut/modules.d/40import-virtiofs-systemd-credentials/import-virtiofs-systemd-credentials.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+# import-virtiofs-systemd-credentials.sh - Import systemd credentials from a virtiofs share
+#
+# This script scans /sys/fs/virtiofs/*/tag for a virtiofs filesystem tagged
+# "io.systemd.credentials". If found, it mounts the filesystem and copies
+# credential files into /run/credentials/@system/ where systemd will pick
+# them up as system credentials.
+#
+# virtiofs tag discovery requires kernel >= 6.9 (commit a8f62f50b4e4).
+# See https://systemd.io/CREDENTIALS/ for the systemd credentials specification.
+# See https://github.com/systemd/systemd/issues/29175 for upstream discussion.
+
+set -euo pipefail
+
+VIRTIOFS_TAG="io.systemd.credentials"
+SYSFS_DIR="/sys/fs/virtiofs"
+SYSTEMD_CREDS_DIR="" # defined below
+MOUNTPOINT="/run/credentials/@virtiofs"
+
+# For EL 9 we need to copy to a different directory because importing
+# from /run/credentials/@initrd was only added in systemd v254.
+# https://github.com/systemd/systemd/pull/28207
+source /etc/os-release
+if [ "${PLATFORM_ID:-}" == "platform:el9" ]; then
+    SYSTEMD_CREDS_DIR="/run/credstore" # Works with LoadCredential
+else
+    SYSTEMD_CREDS_DIR="/run/credentials/@initrd"   # Works with ImportCredential
+fi
+
+log() {
+    echo "import-virtiofs-systemd-credentials: $*" >&2
+}
+
+# Check if the virtiofs sysfs directory exists at all. If the virtiofs
+# kernel module isn't loaded or no virtiofs devices are present, there's
+# nothing to do.
+if [ ! -d "${SYSFS_DIR}" ]; then
+    log "no virtiofs sysfs directory found, nothing to do"
+    exit 0
+fi
+
+log "importing into ${SYSTEMD_CREDS_DIR}"
+
+# Scan all virtiofs devices for our well-known tag. The kernel exposes
+# virtiofs tags at /sys/fs/virtiofs/<N>/tag (since Linux 6.9).
+found=0
+for tagfile in "${SYSFS_DIR}"/*/tag; do
+    [ -e "${tagfile}" ] || continue
+    tag=$(cat "${tagfile}" 2>/dev/null) || continue
+    if [ "${tag}" = "${VIRTIOFS_TAG}" ]; then
+        found=1
+        break
+    fi
+done
+
+if [ "${found}" -eq 0 ]; then
+    log "no virtiofs share with tag '${VIRTIOFS_TAG}' found, nothing to do"
+    exit 0
+fi
+
+log "found virtiofs share with tag '${VIRTIOFS_TAG}'"
+
+# Create the mountpoint and mount the virtiofs share
+mkdir -p "${MOUNTPOINT}"
+if ! mount -t virtiofs "${VIRTIOFS_TAG}" "${MOUNTPOINT}"; then
+    log "failed to mount virtiofs tag '${VIRTIOFS_TAG}'"
+    rmdir "${MOUNTPOINT}" 2>/dev/null || true
+    exit 1
+fi
+
+# Create the system credentials directory if it doesn't exist.
+# This is where systemd looks for system-level credentials.
+mkdir -p "${SYSTEMD_CREDS_DIR}"
+
+# Copy each systemd credential file from the virtiofs mount into the systemd
+# credentials import initrd directory. Credential names are the filenames.
+count=0
+for credfile in "${MOUNTPOINT}"/*; do
+    [ -f "${credfile}" ] || continue
+    credname=$(basename "${credfile}")
+    cp "${credfile}" "${SYSTEMD_CREDS_DIR}/${credname}"
+    # Credentials should be readable only by root
+    chmod 0600 "${SYSTEMD_CREDS_DIR}/${credname}"
+    log "imported systemd credential: ${credname}"
+    count=$((count + 1))
+done
+
+# Clean up the mount
+umount "${MOUNTPOINT}"
+rmdir "${MOUNTPOINT}" 2>/dev/null || true
+
+log "imported ${count} systemd credential(s)"
--- a/overlay.d/40import-virtiofs-systemd-credentials/usr/lib/dracut/modules.d/40import-virtiofs-systemd-credentials/module-setup.sh
+++ b/overlay.d/40import-virtiofs-systemd-credentials/usr/lib/dracut/modules.d/40import-virtiofs-systemd-credentials/module-setup.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+#
+# Dracut module: import-virtiofs-systemd-credentials
+#
+# Imports systemd credentials from a virtiofs share tagged
+# "io.systemd.credentials" into /run/credentials/@system/.
+# This runs in the initramfs so that systemd credentials are available
+# to generators in the real root (e.g. systemd-debug-generator for
+# systemd.extra-unit.* and systemd.unit-dropin.*) and early services
+# (e.g. systemd-tmpfiles-setup for tmpfiles.extra).
+#
+# This provides a cross-architecture alternative to SMBIOS OEM strings
+# and fw_cfg for passing systemd credentials into VMs. Unlike those
+# mechanisms, virtiofs works on all architectures including s390x and
+# ppc64le.
+#
+# See https://systemd.io/CREDENTIALS/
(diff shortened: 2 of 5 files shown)


dustymabe:
This is built on top of https://github.com/coreos/fedora-coreos-config/pull/4227

Please just consider the last commit here.

dustymabe:
Looking at https://github.com/coreos/coreos-assembler/pull/4469 it looks like a desire here is to be able to run tests against other bootable container images (bootc) so adding this functionality maybe should go in https://gitlab.com/fedora/bootc/base-images instead of here.

yasminvalim:
@dustymabe Thanks for working on this! One question I have is: why use virtiofs instead of SMBIOS? Is there a specific reason to choose virtiofs? I'd just like to understand better and get a bit more context.

dustymabe:
> @dustymabe Thanks for working on this! One question I have is: why use virtiofs instead of SMBIOS? Is there a specific reason to choose virtiofs? I'd just like to understand better and get a bit more context.

Good question. Basically https://github.com/systemd/systemd/issues/29175#issuecomment-4722812253

We need something that will work across every architecture we want to run tests on if at all possible.

yasminvalim:
Got it, makes sense! Thank you for explain!  I will make changes in my PR to see if is possible to support virtiofs instead of SMBIOS so we can be in the same page :)

dustymabe:
> Got it, makes sense! Thank you for explain! I will make changes in my PR to see if is possible to support virtiofs instead of SMBIOS so we can be in the same page :)

Not quite ready to push this up yet but you can see the COSA side changes in the `mantle/platform: add virtiofs systemd credential passing` commit on this branch: https://github.com/dustymabe/coreos-assembler/commits/dusty-running/

dustymabe:
@travier interested in your thoughts on this from a security perspective.. and also the question of:

> so adding this functionality maybe should go in https://gitlab.com/fedora/bootc/base-images instead of here.

If we don't put it in bootc then we could add a `ConditionKernelCommandLine=ignition.firstboot` to limit the credential injection only to the first boot of the machine.

dustymabe:
Pushed a second commit here to fix the problem mentioned in https://github.com/coreos/fedora-coreos-config/pull/4227#discussion_r3470148076

```
build-rootfs: sort overlay dirs to allow overriding
```

travier:
> Pushed a second commit here to fix the problem mentioned in [#4227 (comment)](https://github.com/coreos/fedora-coreos-config/pull/4227#discussion_r3470148076)
> 
> ```
> build-rootfs: sort overlay dirs to allow overriding
> ```

This feels unrelated to this PR. Let's split into another one so that we can get it in directly?

dustymabe:
> > Pushed a second commit here to fix the problem mentioned in [#4227 (comment)](https://github.com/coreos/fedora-coreos-config/pull/4227#discussion_r3470148076)
> > ```
> > build-rootfs: sort overlay dirs to allow overriding
> > ```
> 
> This feels unrelated to this PR. Let's split into another one so that we can get it in directly?

https://github.com/coreos/fedora-coreos-config/pull/4236

travier:
Let's drop the second commit from this one.

This looks very interesting but I need to think about it a bit more and wrap my head around it. It's interesting that there is discussion upstream and it feels like something reasonable that we could tweak once upstream approves it.

Also related: https://systemd.io/VM_INTERFACE/

dustymabe:
> Let's drop the second commit from this one.

Done.

rebased this PR. on top of latest `testing-devel`. 

> 

> This looks very interesting but I need to think about it a bit more and wrap my head around it.

:+1: 

> Also related: https://systemd.io/VM_INTERFACE/

Is there anything there that would cause us to change this?

travier:
Small note, the `@initrd` gets interpreted in your commit message by GitHub so the poor user with that name will get a notification every time this commit is pushed.

travier:
So if I understand correctly, this is only for the QEMU virtualization case as this requires virtiofs. But in this case, shouldn't we use the existing [smbios-type-11](https://www.freedesktop.org/software/systemd/man/latest/smbios-type-11.html) support in systemd as that would work as well and should work in limit virtualization context that we have here?

travier:
> > Also related: [systemd.io/VM_INTERFACE](https://systemd.io/VM_INTERFACE/)
> Is there anything there that would cause us to change this?

Not directly in this PR but we could consider using `AF_VSOCK` for the connection to the system in the QEMU case (which I think is the idea here?) as that would remove the need for networking to be up. Most likely for later but that came to mind while reading this one.

dustymabe:
> So if I understand correctly, this is only for the QEMU virtualization case as this requires virtiofs. But in this case, shouldn't we use the existing [smbios-type-11](https://www.freedesktop.org/software/systemd/man/latest/smbios-type-11.html) support in systemd as that would work as well and should work in limit virtualization context that we have here?

I'm looking for something cross architecture. smbios doesn't work everywhere we need it to (kernel args would, but I think cracking open images to inject kernel args for every test would negate some of the test) and maintaining different solutions for different architectures is fraught with problems IMO.

dustymabe:
> Not directly in this PR but we could consider using `AF_VSOCK` for the connection to the system in the QEMU case (which I think is the idea here?) as that would remove the need for networking to be up. Most likely for later but that came to mind while reading this one.

I actually have AF_SOCK working (among some of the commits in https://github.com/dustymabe/coreos-assembler/commits/dusty-running/), but I haven't opened a PR for it yet.

Part of the problem there is https://github.com/coreos/fedora-coreos-tracker/issues/2098#issuecomment-4328820381

Basically I was relying on a Ignition provided SSH key, but using VSOCK to connect, but that doesn't work.

But in general, there are two problems:

1. connecting to the machine through a back channel (vsock)
2. getting a credential onto a machine through !ignition

The implementation for them is independent (one doesn't depend on the other).

travier:
> > So if I understand correctly, this is only for the QEMU virtualization case as this requires virtiofs. But in this case, shouldn't we use the existing [smbios-type-11](https://www.freedesktop.org/software/systemd/man/latest/smbios-type-11.html) support in systemd as that would work as well and should work in limit virtualization context that we have here?
> 
> I'm looking for something cross architecture. smbios doesn't work everywhere we need it to (kernel args would, but I think cracking open images to inject kernel args for every test would negate some of the test) and maintaining different solutions for different architectures is fraught with problems IMO.

Ah yes indeed. I remembered that this was not supported on all architectures (thus why this can not replace Ignition) but I thought maybe QEMU could still do it. Apparently that's unlikely and it's only documented under the i686 target: https://www.qemu.org/docs/master/system/invocation.html#hxtool-4

Also from https://www.dmtf.org/standards/smbios:
> Originally designed for Intel® processor architecture systems, SMBIOS now includes support for IA-32 (x86), x64 (x86-64, Intel64, AMD64, EM64T), Intel® Itanium® architecture, 32-bit ARM (Aarch32) and 64-bit ARM (Aarch64).

travier:
So overall this looks good, but this now becomes another access point that we would have to measure for confidential computing. So if that works for you, I would prefer we either restrict that to testing releases/builds (likely not ideal) or we add a condition to disable it when booted from a measured UKI (systemd has this option now).

dustymabe:
> So overall this looks good, but this now becomes another access point that we would have to measure for confidential computing.

hmm. how does systemd handle this today for say smbios based credential injection? Do they measure that?

> So if that works for you,

I really only care about testing and not trying to enable this generically for users.

> I would prefer we either restrict that to testing releases/builds (likely not ideal) or we add a condition to disable it when booted from a measured UKI (systemd has this option now).

Yeah not sure how we would do the "restrict to testing releases/builds" if we also wanted to use this to test our production builds.

I'm happy to add a `Condition` if you give me the string I'll add it to the unit file.


Also, this question still remains:

> Looking at https://github.com/coreos/coreos-assembler/pull/4469 it looks like a desire here is to be able to run tests against other bootable container images (bootc) so adding this functionality maybe should go in https://gitlab.com/fedora/bootc/base-images instead of here.

If we keep it here in CoreOS we can do a `ConditionKernelCommandLine=ignition.firstboot` to further limit potential problems, but having it in bootc starts to open us up to being able to run our kola tests against bootc images.

travier:
> > So overall this looks good, but this now becomes another access point that we would have to measure for confidential computing.
> 
> hmm. how does systemd handle this today for say smbios based credential injection? Do they measure that?

Yes, they started measuring SMBIOS Type 11 entries read from system-stub into a PCR (12 I think).

> > So if that works for you,
> 
> I really only care about testing and not trying to enable this generically for users.

👍🏻 but once it's in, someone will use it 🙂.

> > I would prefer we either restrict that to testing releases/builds (likely not ideal) or we add a condition to disable it when booted from a measured UKI (systemd has this option now).
> 
> Yeah not sure how we would do the "restrict to testing releases/builds" if we also wanted to use this to test our production builds.
> 
> I'm happy to add a `Condition` if you give me the string I'll add it to the unit file.

Will do.

> Also, this question still remains:
> 
> > Looking at [coreos/coreos-assembler#4469](https://github.com/coreos/coreos-assembler/pull/4469) it looks like a desire here is to be able to run tests against other bootable container images (bootc) so adding this functionality maybe should go in [gitlab.com/fedora/bootc/base-images](https://gitlab.com/fedora/bootc/base-images) instead of here.
> 
> If we keep it here in CoreOS we can do a `ConditionKernelCommandLine=ignition.firstboot` to further limit potential problems, but having it in bootc starts to open us up to being able to run our kola tests against bootc images.

Agree that it would be best to have that in the bootc base image for wider testing. Or we could use that for FCOS testing on PRs to the bootc repo / pipeline.

travier:
From https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#ConditionSecurity=:
```
ConditionSecurity=!cvm
ConditionSecurity=!measured-uki
ConditionSecurity=!measured-os
```

dustymabe:
> From https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#ConditionSecurity=:

added


> Agree that it would be best to have that in the bootc base image for wider testing. Or we could use that for FCOS testing on PRs to the bootc repo / pipeline.

Do you think I should open this against bootc then? Or maybe we do it in parallel and don't block on it being there just in case?

I updated this PR to have `ConditionKernelCommandLine=ignition.firstboot`

travier:
I think we can do it in parallel and we can start testing this here.

dustymabe:
@travier anything blocking this at this point?

overlay.d/40import-virtiofs-systemd-credentials/usr/lib/dracut/modules.d/40import-virtiofs-systemd-credentials/import-virtiofs-systemd-credentials.sh
```diff
@@ -0,0 +1,92 @@
...
+    credname=$(basename "${credfile}")
+    cp "${credfile}" "${SYSTEMD_CREDS_DIR}/${credname}"
+    # Credentials should be readable only by root
+    chmod 0600 "${SYSTEMD_CREDS_DIR}/${credname}"
+    log "imported systemd credential: ${credname}"
+    count=$((count + 1))
+done
+
+# Clean up the mount
+umount "${MOUNTPOINT}"
```
travier, line 89: We should be able to skip that by using a dedicated mount namespace like we do in other units that mount filesystems. https://github.com/coreos/bootupd/blob/main/systemd/bootloader-update.service#L16

overlay.d/40import-virtiofs-systemd-credentials/usr/lib/dracut/modules.d/40import-virtiofs-systemd-credentials/import-virtiofs-systemd-credentials.sh
```diff
@@ -0,0 +1,92 @@
...
+rmdir "${MOUNTPOINT}" 2>/dev/null || true
```
travier, line 90: Ah, maybe not if we need to cleanup the mountpoint

overlay.d/40import-virtiofs-systemd-credentials/usr/lib/dracut/modules.d/40import-virtiofs-systemd-credentials/module-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+    fi
+    return 0
+}
+
+depends() {
+    echo systemd
+}
+
+installkernel() {
+    instmods -c virtiofs
```
travier, line 35: This will make our initrd 35K (xz compressed) bigger so should be fine.

travier (approved):
One potential nit but could be a followup if actually possible.
LGTM

PR #4234, opened 2026-06-25
tests/bootupd: fix test and reduce code duplication

nikita-dubrovskii:
The test was failing on ppc64le and s390x with:
  + grep -iq 'ConditionArchitecture=!s390x was not met'
  + fatal 'bootloader-update.service was not skipped for s390x'

This is because bootupd now uses a preset file (10-ppcle-s390x-disable.conf) to disable the service on ppc64le/s390x, which makes the service 'inactive' instead of showing a ConditionArchitecture message.

Fix by using 'systemctl is-active' and checking for 'inactive' status. Also refactor to reduce code duplication with helper functions.

### DIFF
--- a/tests/kola/boot/bootupd
+++ b/tests/kola/boot/bootupd
@@ -17,44 +17,58 @@ set -xeuo pipefail
 # s390x, let's just conditionally check that *if* bootupd is installed, then
 # it's functioning as expected. We can harden it more once we've hard cut over
 # to 9.4.
+
+current_arch="$(arch)"
+
+check_bootloader_service_inactive() {
+    local bl_active
+    bl_active=$(systemctl is-active bootloader-update.service 2>&1 || true)
+    if [ "${bl_active}" != "inactive" ]; then
+        fatal "bootloader-update.service is ${bl_active} on ${current_arch}"
+    fi
+}
+
+check_grub_version() {
+    local package="$1"
+    local format="$2"
+    local version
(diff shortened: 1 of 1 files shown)


tests/kola/boot/bootupd
```diff
@@ -17,42 +17,55 @@ set -xeuo pipefail
...
-            fatal "bootupctl status output should include grub2-tools package version"
-        fi
-        ;;& # fallthrough to default
-    s390x)
-        bl_status=$(systemctl status bootloader-update.service 2>&1 || true)
-        if ! echo "$bl_status" | grep -iq "ConditionArchitecture=!s390x was not met"; then
-            fatal "bootloader-update.service was not skipped for s390x"
-        fi
-        ;;& # fallthrough to default
-    *)
```
Rolv-Apneseth, line 53: WDYT about having a default case like [like the one here](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/boot/raid1-boot/data/commonlib_raid1.sh#L30) in case there are new architectures that need to be handled in this test?

Rolv-Apneseth (commented):
One minor suggestion, otherwise LGTM

tests/kola/boot/bootupd
nikita-dubrovskii: Good catch! Done.

Rolv-Apneseth (approved):

PR #4240, opened 2026-07-02
tests: add a test to check if fwupd-refresh.service can run without failing

Rolv-Apneseth:
> Recently, an issue [1] with fwupd meant that this service fails when triggered by the timer. Only testing the timer means we didn't catch this issue, so let's add a test to hopefully prevent this in the future.
> 
> [1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1139344

Related to https://github.com/coreos/fedora-coreos-tracker/issues/2176.

This test passes on `44.20260607.1.1` and fails on `44.20260621.1.1`. Also seems to pass downstream (tested locally).


Rolv-Apneseth:
I didn't test across every single combination of arch and stream/variant but this seems to be passing across the board for {F,RH,S}COS. Happy to merge and revert if there are issues.

tests/kola/fwupd/fwupd-refresh-service
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+##   description: Verify fwupd-refresh.service can run.
+##   tags: "platform-independent needs-internet"
+##   creationDate: 2026-07-02
```
dustymabe: limit to just running on fcos here?

tests/kola/fwupd/fwupd-refresh-units
dustymabe, line 1: I don't see a lot of value in having two separate files for tests for the `.timer` and the `.service`. WDYT about just combining them into `tests/kola/fwupd/fwupd-refresh-units` ?

Rolv-Apneseth: Yeah I did think about it but chose to leave the pre-existing test alone. I can combine them.

tests/kola/fwupd/fwupd-refresh-service
Rolv-Apneseth: It seemed to pass just fine on rhcos/scos - but actually what's the reason `fwupd` is even on those? Would it ever be used?

dustymabe: hmm. ok. I didn't think it was there either, but if it is I guess we don't need to limit the tests any longer?

Rolv-Apneseth: I haven't tested whether the timer is on downstream but the comment in that file would have me believe it isn't. Either way, if `fwupd` isn't really meant to be used there, I'd say we should only run this test on FCOS. ~Maybe we should even try to remove `fwupd` from downstream?~

Rolv-Apneseth: Actually nevermind, I'm sure there's a good reason for it to be there. I remembered one of the cards I'll be working on is specifically verifying it can run on RHCOS.

It seems the timer is slightly trickier though. Here's what enables it in FCOS: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/15fcos/usr/lib/systemd/system-preset/45-fcos.preset#L3

A quick check on local builds shows:

RHCOS 10.2:

```bash
$ systemctl status fwupd-refresh.timer
● fwupd-refresh.timer - Refresh fwupd metadata regularly
     Loaded: loaded (/usr/lib/systemd/system/fwupd-refresh.timer>
     Active: active (waiting) since Sat 2026-07-18 15:41:44 UTC;>
 Invocation: fcedff9197d346b1a2f949ff083c8437
    Trigger: Sat 2026-07-18 16:09:11 UTC; 27min left
   Triggers: ● fwupd-refresh.service 
```

RHCOS 9.8:

```bash
$ systemctl status fwupd-refresh.timer
○ fwupd-refresh.timer - Refresh fwupd metadata regularly
     Loaded: loaded (/usr/lib/systemd/system/fwupd-refresh.timer>
     Active: inactive (dead)
    Trigger: n/a
   Triggers: ● fwupd-refresh.service 
```

SCOS 9 has it disabled, and SCOS 10 is enabled.

Not sure where these are inherited from but should I still merge into one test but skip the timer check on rhel <= 9?

Rolv-Apneseth:
Went ahead and merged the 2 tests, but skipping the timer check for EL9 (https://github.com/coreos/fedora-coreos-config/pull/4240#discussion_r3608674377)

dustymabe (approved):
LGTM

PR #4245, opened 2026-07-03
denylist: drop ext.config.docker.swarm-overlay-network entry

Rolv-Apneseth:
This test is no longer failing on any stream due to updated docker versions (F44 is now on v29.6.0, F45 is on v29.6.1).

cc @nikita-dubrovskii

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -38,13 +38,6 @@
   platforms:
     - gcp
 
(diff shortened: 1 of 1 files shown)


Rolv-Apneseth:
Tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2143

nikita-dubrovskii (approved):
LGTM!

PR #4253, opened 2026-07-08
overlay/disk-image: add missing partition GPT names

jbtrystram:
Add the partition names to the `disk.yaml` image-builder configuration file. The partitions were correctly labelled at the filesystem level but were missing names from the GPT parition table.
This was causing some issues with reprovisionning. Let's make sure the resulting partition table is identical to what we have when building disk-images with COSA.

See how partitions names are set in COSA at the GPT table : https://github.com/coreos/coreos-assembler/blob/a2b4dd6c901c074fb36e6d61df114b4386c4bc74/src/osbuild-manifests/coreos.osbuild.x86_64.mpp.yaml#L65


joelcapitao (approved):
LGTM

Roshan-R (approved):
LGTM

joelcapitao (approved):

jbtrystram:
The rpm-diff CI issue is unrelated

PR #4255, opened 2026-07-09
overlays/10disk-images: Create compat symlink for coreos-aleph

jbtrystram:
Add a compatibility symlink for coreos-aleph pointing to the aleph
created by bootc install.

A tmpfiles.d entry cannot be used because /sysroot is mounted
read-only in the running system. Instead, use a oneshot systemd
service with MountFlags=slave to transiently remount /sysroot
read-write in its own mount namespace and create the symlink.

The service is conditioned on the symlink not already existing
(ConditionPathExists=!) so it only runs once and is a no-op on
streams where the file is already present.

See https://github.com/coreos/fedora-coreos-tracker/issues/2187

### DIFF
--- a/overlay.d/10disk-images/usr/lib/systemd/system-preset/45-coreos-aleph-compat-symlink.preset
+++ b/overlay.d/10disk-images/usr/lib/systemd/system-preset/45-coreos-aleph-compat-symlink.preset
@@ -1,0 +1,1 @@
+enable coreos-aleph-compat-symlink.service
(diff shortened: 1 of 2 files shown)


travier:
https://github.com/coreos/fedora-coreos-tracker/issues/2187#issuecomment-4928146833

jbtrystram:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/4261

PR #4261, opened 2026-07-15
tests: adapt to bootc-created aleph

jbtrystram:
Disk images created by `bootc install` are shipping an aleph file at a different path than what existed historically.
Update our tests to look for both locations.

See https://github.com/coreos/fedora-coreos-tracker/issues/2187 And https://github.com/coreos/bootupd/pull/1124


tests/kola/boot/bootupd
```diff
@@ -78,8 +82,10 @@ fi
 
 # Verify `bootupctl status` output contains related content.
 # https://github.com/coreos/bootupd/issues/694
-if ! bootupctl status | grep "CoreOS aleph version"; then
-    fatal "bootupctl status output should include 'CoreOS aleph version'"
+# https://github.com/coreos/bootupd/pull/1124 renamed the output from
+# "CoreOS aleph version" to "Aleph version", match both for compatibility.
+if ! bootupctl status | grep -i "aleph version"; then
```
Johan-Liebert1: This would be cleaner with json output so that we don't end up matching some random string

jbtrystram: looked quickly into it, but `--json` does not give the aleph version, only the rpm version of available and installed components.

Johan-Liebert1: That feels like a disconnect between json printing and human readable printing in bootupd

Johan-Liebert1: I can see it's not in the `Status` struct in bootupd. Does it make sense to put it in the struct? If not we can just go with the current approach

travier: I think it might be good to update bootupd JSON output to include it as well if possible. Ideally, the JSON output should be the reference and the non-JSON output a potentially trimmed and pretty printed version of the JSON / struct.

travier: Can also happen as a follow up change.

tests/kola/files/aleph-version
```diff
@@ -1,14 +1,28 @@
...
 set -xeuo pipefail
 
 # shellcheck disable=SC1091
 . "$KOLA_EXT_DATA/commonlib.sh"
 
-jq . < /sysroot/.coreos-aleph-version.json >/dev/null
+aleph_file=""
+if [ -f /sysroot/.bootc-aleph.json ]; then
+    aleph_file=/sysroot/.bootc-aleph.json
+elif [ -f /sysroot/.coreos-aleph-version.json ]; then
```
travier, line 22: Hum, other tests use `.aleph-version.json`. Do we have 3 different names now?

jbtrystram: yep we have 3 ! The stage we have in osbuild symlinks it : 
https://github.com/osbuild/osbuild/blob/8685b7f295fd6ff073cdd67d105a49a9e1a0f578/stages/org.osbuild.ostree.aleph#L131-L133

tests/kola/boot/bootupd
jbtrystram: Let's update bootupd now and have something clean

jbtrystram: https://github.com/coreos/bootupd/pull/1131

Rolv-Apneseth (dismissed):
LGTM

jbtrystram:
@Rolv-Apneseth we should update this one to use https://github.com/coreos/bootupd/pull/1131

Rolv-Apneseth:
> @Rolv-Apneseth we should update this one to use [coreos/bootupd#1131](https://github.com/coreos/bootupd/pull/1131)

Alright. If you don't have the time to adjust this let me know

edit: ah, are we waiting on a release? We want v0.2.36 of `bootupd` but `testing-devel` is still on `0.2.35`

Rolv-Apneseth:
Updated to use the `--json` output as requested by @jbtrystram, and rebased while I was at it. Once we have `bootupd 0.2.36`, this should pass.

Rolv-Apneseth:
Forgot `-e` flag for `jq` - `null` would still succeed

Rolv-Apneseth:
Conflicts are because of https://github.com/coreos/fedora-coreos-config/pull/4322. Should we revert 15e178e4491fcb9df8ba5965b7733f87decaf0a9 in this PR?

On the bright side, that commit unblocked `bootupd` CI.

jbtrystram:
> Conflicts are because of #4322. Should we revert [15e178e](https://github.com/coreos/fedora-coreos-config/commit/15e178e4491fcb9df8ba5965b7733f87decaf0a9) in this PR?

Parsing the JSON output as we do in this PR makes it more robust so we should fix the conflict and merge this IMHO. I guess the commit should be reworded to highlight that improvement

jbtrystram:
The bootupd is released. Rebased and updated the commit message to mention the JSON parsing

Rolv-Apneseth (approved):

PR #4263, opened 2026-07-16
Containerfile: support chunking with chunkah

dustymabe:
This will implement chunking with chunkah based on the
BUILDER_IMG_CHUNKER build arg.

Also the BUILDER_IMG_CHUNKER_MAX_LAYERS build arg can be
used to set the maximum number of layers for a chunked image.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/2145

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -51,13 +51,13 @@ RUN --mount=type=bind,target=/run/src,rw \
         --bootc --format-version=1 --rootfs /target-rootfs \
         --output oci-archive:/run/src/out.ociarchive \
         --label com.coreos.inputhash=$(cat /run/inputhash) \
-        --label com.coreos.stream=$STREAM \
         ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.versions=machine-os=${VERSION}} \
         ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.version-display-names=machine-os="${DESCRIPTION}"}
 
 FROM oci-archive:./out.ociarchive
 ARG VERSION
 ARG NAME=overridden
+ARG STREAM=overridden
 ARG DESCRIPTION=overridden
 # Need to reference builder here to force ordering. But since we have to run
 # something anyway, we might as well cleanup after ourselves.
@@ -69,6 +69,7 @@ LABEL containers.bootc=1 \
       ostree.bootable=1 \
       org.opencontainers.image.version=$VERSION \
       com.coreos.osname=$NAME \
+      com.coreos.stream=$STREAM \
       org.opencontainers.image.title=$DESCRIPTION \
       org.opencontainers.image.description=$DESCRIPTION
 STOPSIGNAL SIGRTMIN+3
--- a/Containerfile
+++ b/Containerfile
@@ -30,6 +30,8 @@ ARG IMAGE_CONFIG=overridden
 ARG PASSWD_GROUP_DIR
 ARG STRICT_MODE=0
 ARG INJECT_OPENSHIFT_VERSION_LABELS=""
+ARG BUILDER_IMG_CHUNKER=""
+ARG BUILDER_IMG_CHUNKER_MAX_LAYERS=""
 
 COPY . /src
 # canonicalize permission bits, see also https://gitlab.com/fedora/bootc/base-images/-/merge_requests/274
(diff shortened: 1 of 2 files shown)


dustymabe:
Initially submitting as draft to text chunking with `chunkah` in CI.

Once CI passes I'll remove the build-args that configure using `chunkah` and open a PR for `rawhide` to use `chunkah` first.

build-args.conf
```diff
@@ -12,3 +12,11 @@ MANIFEST=manifest.yaml
 # XXX: see inject_passwd_group() in build-rootfs
 PASSWD_GROUP_DIR=manifests
 IMAGE_CONFIG=image.yaml
+# Use buildah for chunking and also set the number of
```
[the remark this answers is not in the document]
dustymabe: I'm dropping this commit so this will go away.

dustymabe:
ok. CI passed. Dropping the commit that enables it because we don't want it for `testing-devel` yet and I'll mark this as ready for review.

Containerfile
```diff
@@ -46,18 +48,32 @@ RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
...
+        # will strip OSTree data and labels for bootc compatibility; see
+        # https://github.com/coreos/chunkah#compatibility-with-bootable-bootc-images
+        chunkah build --rootfs /target-rootfs                                                             \
+            --prune /sysroot/ --label ostree.commit- --label ostree.final-diffid-                         \
+            --output oci-archive:/run/src/out.ociarchive                                                  \
+            --label com.coreos.inputhash=$(cat /run/inputhash)                                            \
+            ${BUILDER_IMG_CHUNKER_MAX_LAYERS:+--max-layers=${BUILDER_IMG_CHUNKER_MAX_LAYERS}}             \
+            ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.versions=machine-os=${VERSION}} \
+            ${INJECT_OPENSHIFT_VERSION_LABELS:+--label io.openshift.build.version-display-names=machine-os="${DESCRIPTION}"}
+    else
```
Rolv-Apneseth: Thoughts on trying to catch misconfigurations with an extra branch here?

```bash
elif [ -n "${BUILDER_IMG_CHUNKER:-}" ]; then
    echo "error: unknown BUILDER_IMG_CHUNKER value: ${BUILDER_IMG_CHUNKER}" >&2
    exit 1
```

Or a case statement I guess:

```bash
case "${BUILDER_IMG_CHUNKER:-}" in
    chunkah)
        chunkah build --rootfs /target-rootfs \
         ...
        ;;
    "")
        rpm-ostree experimental compose build-chunked-oci \
        ...
        ;;
    *)
        echo "error: unknown BUILDER_IMG_CHUNKER value: ${BUILDER_IMG_CHUNKER}" >&2
        exit 1
        ;;
esac
```

Rolv-Apneseth:
Anything to worry about with the Konflux failures?

Containerfile
dustymabe: ehh. I'm not sure it's worth the CI cycle. Supporting both workflows is hopefully a short term thing. 

How about this - if I do end up having to push again here (and thus have CI run again) I'll make the change? 

Also happy to accept a later PR to do the same thing if you feel strongly and have tested it.

Rolv-Apneseth: > How about this - if I do end up having to push again here (and thus have CI run again) I'll make the change?

Sounds good

aaradhak (dismissed):
LGTM

dustymabe:
> Anything to worry about with the Konflux failures?

```
[1/2] STEP 19/19: RUN --mount=type=bind,target=/run/src,rw . /cachi2/cachi2.env &&     <<EOF (if [ "${BUILDER_IMG_CHUNKER:-}" = 'chunkah' ]; then...)
/bin/sh: line 1: --prune: command not found
```

hmm. weird. I'm not sure why this would fail in konflux but not in the `podman build` we are doing in our Jenkins.

dustymabe:
did a new push here to see if changing the backticks to a `'` in a comment fixes the issue.

Containerfile
dustymabe: since I did a new push I'll implement this change.

dustymabe:
hmm. now we don't get the error/log message but the oci-archive file isn't found:

```
[1/2] STEP 19/19: RUN --mount=type=bind,target=/run/src,rw . /cachi2/cachi2.env &&     <<EOF (if [ "${BUILDER_IMG_CHUNKER:-}" = 'chunkah' ]; then...)
[2/2] STEP 1/12: FROM oci-archive:./out.ociarchive
Error: creating build container: creating temp directory: archive file not found: "/var/workdir/source/out.ociarchive"
```

dustymabe:
> hmm. now we don't get the error/log message but the oci-archive file isn't found:

added `set -x` so we can see where it's failing.

dustymabe:
ok I think this is some issue with how we wrap konflux:

```
[1/2] STEP 19/19: RUN --mount=type=bind,target=/run/src,rw . /cachi2/cachi2.env &&     <<EOF (if [ "${BUILDER_IMG_CHUNKER:-}" = 'chunkah' ]; then...)
```

Note the `. /cachi2/cachi2.env &&` in the command that's not actually in the Containerfile and I suspect this substitution is breaking heredoc `<<EOF` processing.

cc @joelcapitao

jbtrystram:
> hmm. weird. I'm not sure why this would fail in konflux but not in the `podman build` we are doing in our Jenkins.

Note that konflux builders are running rhel so in some cases we've hit weird bugs due to the host kernel being too old.

Containerfile
dustymabe: ok I implemented this now

dustymabe:
Added a workaround that should help I think:

`RUN --mount=type=bind,target=/run/src,rw <<EOF` -> `RUN --mount=type=bind,target=/run/src,rw bash <<EOF`

dustymabe:
at least one konflux test passing

cgwalters:
> Note the . /cachi2/cachi2.env && in the command that's not actually in the Containerfile and I suspect this substitution is breaking heredoc <<EOF processing.

https://github.com/konflux-ci/build-definitions/issues/1200

joelcapitao:
> > Note the . /cachi2/cachi2.env && in the command that's not actually in the Containerfile and I suspect this substitution is breaking heredoc <<EOF processing.
> 
> [konflux-ci/build-definitions#1200](https://github.com/konflux-ci/build-definitions/issues/1200)

Based on the last comment https://github.com/konflux-ci/build-definitions/issues/1200#issuecomment-4845087405 we need buildah task >= 0.10.3 and  prefetch-dependencies >= 0.3.1
Currently, we have buildah 0.9.3 c.f https://gitlab.com/fedora/bootc/tekton-catalog/-/blob/22f1030bb088dd392a5b46f09b913642f93e3559/tasks/buildah-remote-oci-ta/buildah-remote-oci-ta.yaml#L6, I'll bump the pipelines and tasks definition next week to get rid of this workaround (it's been a while since the last bump)

aaradhak (approved):

PR #4266, opened 2026-07-20
[stable] tree: promote changes from testing at 107c7b6bf2e5244507eb350928b788beb3337614

coreosbot-releng:
Triggered by @c4rt0 in https://github.com/coreos/fedora-coreos-streams/issues/1271.


c4rt0 (approved):
/lgtm

c4rt0:
Regarding the above failed CI test - Jenkins recognised downgrade of `fwupd` which is intentional, it can be safely ignored (see: https://github.com/coreos/fedora-coreos-config/pull/4260).

PR #4268, opened 2026-07-20
[next] tree: promote changes from testing-devel at 4febed40668d15b7e4413eb09b65970ea9d6c502

coreosbot-releng:
Triggered by @c4rt0 in https://github.com/coreos/fedora-coreos-streams/issues/1269.


c4rt0 (approved):
/lgtm

c4rt0:
Regarding the above failed CI test - Jenkins recognised downgrade of `fwupd` which is intentional and therefore it can be ignored (see: https://github.com/coreos/fedora-coreos-config/pull/4260).

PR #4270, opened 2026-07-20
overlay/16fcos: add a systemd generator to fix socket-activated sshd auth

Rolv-Apneseth:
> tests/ssh: add test to confirm socket-activated sshd works
> 
> Add a test to confirm that Ignition-provided SSH keys work with
> socket-activated sshd.
> 
> overlay/15fcos: add openssh drop-in to fix SSH login on socket-activated sshd
> 
> The upstream sshd@.service overrides AuthorizedKeysFile in the ExecStart
> command, which excludes .ssh/authorized_keys.d/* where Ignition and
> Afterburn place SSH keys. This breaks SSH logins for systems using
> socket-activated sshd.
>
> Add this drop-in suggested by @travier as a workaround to override ExecStart
> while a better solution is discussed upstream.

See https://github.com/coreos/fedora-coreos-tracker/issues/2098

**EDIT:** for context, we ended up with a dynamic approach using a systemd generator instead

### DIFF
--- a/overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system-generators/coreos-sshd-generator
+++ b/overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system-generators/coreos-sshd-generator
@@ -1,0 +1,32 @@
+#!/bin/bash
+# The upstream sshd@.service overrides AuthorizedKeysFile in the
+# ExecStart command, which excludes .ssh/authorized_keys.d/* where ignition
+# and afterburn place SSH keys. This generator reads the configured
+# AuthorizedKeysFile via sshd -G and appends the systemd credentials
+# path so that both user/distro config and ephemeral keys are honoured.
+#
+# See https://github.com/coreos/fedora-coreos-tracker/issues/2098
+
+NORMAL_DIR="$1"
+
+# First, ensure that the ExecStart line hasn't changed
+upstream_service="/usr/lib/systemd/system/sshd@.service"
+if ! grep -q -- '-o "AuthorizedKeysFile ${CREDENTIALS_DIRECTORY}/ssh.ephemeral-authorized_keys-all .ssh/authorized_keys"' "$upstream_service" 2>/dev/null; then
+    echo "coreos-sshd-generator: upstream sshd@.service ExecStart has changed" >&2
+    exit 1
+fi
+
+# Parse AuthorizedKeysFile value from sshd -G
+auth_keys=$(sshd -G | grep -i "^authorizedkeysfile " | cut -d" " -f2-)
+if [ -z "$auth_keys" ]; then
+    echo "coreos-sshd-generator: failed to get AuthorizedKeysFile from sshd -G" >&2
+    exit 1
+fi
+
+# Write drop-in to override ExecStart
+mkdir -p "$NORMAL_DIR/sshd@.service.d"
+cat > "$NORMAL_DIR/sshd@.service.d/50-authorized-keys-file.conf" <<EOF
+[Service]
+ExecStart=
+ExecStart=-/usr/sbin/sshd -i \$OPTIONS -o "AuthorizedKeysFile \${CREDENTIALS_DIRECTORY}/ssh.ephemeral-authorized_keys-all $auth_keys"
+EOF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -54,6 +54,24 @@ conditional-include:
           dest=/etc/containers/policy.json
           cp "${src}" "${dest}"
 
+  # WARN: Temporary workaround
+  #
+  # The Fedora sshd@.service overrides AuthorizedKeysFile on the
+  # ExecStart command line, excluding .ssh/authorized_keys.d/* where
+  # Ignition and Afterburn place SSH keys. This breaks auth for
+  # socket-activated sshd.
+  #
+  # See:
+  # - https://github.com/coreos/fedora-coreos-tracker/issues/2098
+  # - https://bugzilla.mindrot.org/show_bug.cgi?id=3957
+  # - https://bugzilla.redhat.com/show_bug.cgi?id=2503149
+  #
+  # TODO: delete if openssh ever supports AuthorizedKeysFile+= (https://bugzilla.mindrot.org/show_bug.cgi?id=3957)
+  - if: releasever <= 45
+    include:
+      ostree-layers:
+        - overlay.d/16fcos-sshd-workaround
+
 ostree-layers:
   - overlay.d/09misc
   - overlay.d/10disk-images
--- a/tests/kola/ssh/socket-auth/config.bu
+++ b/tests/kola/ssh/socket-auth/config.bu
@@ -1,0 +1,8 @@
+variant: fcos
+version: 1.6.0
+systemd:
+  units:
+    - name: sshd.service
+      mask: true
+    - name: sshd.socket
+      enabled: true
--- a/tests/kola/ssh/socket-auth/data/commonlib.sh
+++ b/tests/kola/ssh/socket-auth/data/commonlib.sh
@@ -1,0 +1,1 @@
+../../../data/commonlib.sh
--- a/tests/kola/ssh/socket-auth/test.sh
+++ b/tests/kola/ssh/socket-auth/test.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+## kola:
+##   distros: fcos
+##   tags: platform-independent
+##   creationDate: 2026-07-20
+##   description: Verify that Ignition-provided SSH keys work
+##     with socket-activated sshd.
+##   timeoutMin: 5
+
+# This test masks sshd.service and enables sshd.socket using butane,
+# which forces kola to connect through the socket-activated path.
+#
+# See https://github.com/coreos/fedora-coreos-tracker/issues/2098
+#
+# The primary check here is implicit on whether the SSH connection to
+# the kola VM succeeds at all, but some additional sanity checks are
+# also included below.
+
+set -xeuo pipefail
+
+# shellcheck disable=SC1091
+. "$KOLA_EXT_DATA/commonlib.sh"
+
+if systemctl is-active sshd.service 2> /dev/null; then
+    fatal "sshd.service is active; expected only sshd.socket"
+fi
+ok "sshd.service is not active"
+
+if ! systemctl is-active sshd.socket 2> /dev/null; then
+    systemctl status sshd.socket
+    fatal "sshd.socket is not active"
+fi
+ok "sshd.socket is active"
+
+ok "SSH authentication works via socket-activated sshd"


overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system/sshd@.service.d/50-authorized-keys-file.conf
```diff
@@ -0,0 +1,11 @@
...
+# ExecStart command, which excludes .ssh/authorized_keys.d/* where ignition
+# and afterburn place SSH keys. This drop-in is a workaround that restores
+# authorized_keys.d support for socket-activated sshd.
+#
+# See https://github.com/coreos/fedora-coreos-tracker/issues/2098
+#
+# TODO: delete if openssh ever supports AuthorizedKeysFile+= (https://bugzilla.mindrot.org/show_bug.cgi?id=3957)
+[Service]
+ExecStart=
+ExecStart=-/usr/sbin/sshd -i $OPTIONS -o "AuthorizedKeysFile ${CREDENTIALS_DIRECTORY}/ssh.ephemeral-authorized_keys-all .ssh/authorized_keys .ssh/authorized_keys.d/*"
```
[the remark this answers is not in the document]
Rolv-Apneseth: Just copying what's in the upstream service: https://src.fedoraproject.org/rpms/openssh/blob/rawhide/f/sshd%40.service

Rolv-Apneseth:
Confirmed passing on testing-devel and rawhide across our supported arches

dustymabe:
I really want to see this problem fixed, but I'm not sure if the right solution is for us to fix it only in CoreOS.

Over the years my opinion has become that if the fix upstream isn't evident or imminent then we shouldn't workaround things here.

We snowflake ourselves into oblivion that way.

What I am saying here:

- I'm happy to carry a workaround like this, but....
- I'd like there to be some acknowledgement upstream about the problem and a fix prior to us adding in a workaround. When we do add the workaround I'd like it to be version limited, so by default the workaround will drop out and our tests will either start failing or continue to pass.

Rolv-Apneseth:
> I'd like there to be some acknowledgement upstream about the problem and a fix prior to us adding in a workaround. When we do add the workaround I'd like it to be version limited, so by default the workaround will drop out and our tests will either start failing or continue to pass.

Has there been any movement in https://bugzilla.mindrot.org/show_bug.cgi?id=3957? I think I need to create an account to view it.

But you're saying you'd want an acknowledgement and potential fix from whoever owns the upstream service file?

What could we do for your second point about version limiting the workaround?

dustymabe:
> > I'd like there to be some acknowledgement upstream about the problem and a fix prior to us adding in a workaround. When we do add the workaround I'd like it to be version limited, so by default the workaround will drop out and our tests will either start failing or continue to pass.
> 
> Has there been any movement in https://bugzilla.mindrot.org/show_bug.cgi?id=3957? I think I need to create an account to view it.

No :(

> 
> But you're saying you'd want an acknowledgement and potential fix from whoever owns the upstream service file?

Ideally. "Upstream" here could be sshd OR Fedora's packaging of sshd.

> 
> What could we do for your second point about version limiting the workaround?

Something like https://github.com/coreos/fedora-coreos-config/commit/9b14cac40de19089d4362b960cb7d098fd523301 where the workaround will automatically stop being applied at some point in time.

dustymabe:
> > Has there been any movement in https://bugzilla.mindrot.org/show_bug.cgi?id=3957? I think I need to create an account to view it.
> 
> No :(

https://github.com/coreos/fedora-coreos-tracker/issues/2098#issuecomment-5027137102

Rolv-Apneseth:
I've changed to a version-limited approach, let me know if that's what you had in mind. And as mentioned in the tracker, I've opened https://bugzilla.redhat.com/show_bug.cgi?id=2503149

dustymabe:
> I've changed to a version-limited approach, let me know if that's what you had in mind.

Yes. This is what I had in mind.

My preferred prereq of "I'd like there to be some acknowledgement upstream about the problem and a fix prior to us adding in a workaround" still applies here, though.

Basically can we get systemd or sshd or Fedora sshd maintainer to agree and take some action? if not maybe we shouldn't fix because then we're agreeing to carry that fix forever and should we be doing that here?

Maybe the underlying problem is that we're using Ignition/afterburn to write separate auth keys files other than the main one OR we should go back to the old ssh-key-dir approach.

If we "fix" it only here (via a workaround) we don't ever either solve the problem correctly upstream, OR change the way we use things so that the problem doesn't exist to begin with.

Rolv-Apneseth:
> My preferred prereq of "I'd like there to be some acknowledgement upstream about the problem and a fix prior to us adding in a workaround" still applies here, though.
>
> Basically can we get systemd or sshd or Fedora sshd maintainer to agree and take some action? if not maybe we shouldn't fix because then we're agreeing to carry that fix forever and should we be doing that here?

Of course, yes. I'd personally lean towards getting the workaround in for now and continuing upstream discussions, especially with us version-limiting it like this, but I understand.

> Maybe the underlying problem is that we're using Ignition/afterburn to write separate auth keys files other than the main one OR we should go back to the old ssh-key-dir approach.
> 
> If we "fix" it only here (via a workaround) we don't ever either solve the problem correctly upstream, OR change the way we use things so that the problem doesn't exist to begin with.

 To me, https://bugzilla.mindrot.org/show_bug.cgi?id=3957 is what makes the most sense as an actual solution here, but no reply there yet. And we'll see how the discussion goes in https://bugzilla.redhat.com/show_bug.cgi?id=2503149. But I do think the fix belongs upstream in this scenario. Just not entirely sure how long we should wait for that with an issue like this since it's actively causing problems for users.

Rolv-Apneseth:
That was a premature push, still working through how to take a more dynamic approach using `sshd -G`

Rolv-Apneseth:
So, I've tried:

1. `ExecStartPre` + `EnvironmentFile` - timing issue with env file being read before `ExecStartPre` (whoops)
2. Just running the bash directly in `ExecStart` - seems that `sshd -G` interferes with stdin which breaks `sshd -i`
3. A wrapper script called by `ExecStart`, saving + restoring the socket stdin - runs in the wrong SELinux domain and re-labeling doesn't seem feasible

Any ideas on how else to approach this @dustymabe

dustymabe:
only other thing I can think of is a separate unit that runs that populates the environmentfile in `/run/` that this unit then uses (via your override). It's ugly :(

Rolv-Apneseth:
> only other thing I can think of is a separate unit that runs that populates the environmentfile in `/run/` that this unit then uses (via your override).

That was a good suggestion, thanks. Took a bit to get it working but it passed locally (see latest changes)

> It's ugly :(

Yeah :(

dustymabe:
> Took a bit to get it working but it passed locally (see latest changes)

Nice!

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system/sshd-authorized-keys-config.service
```diff
@@ -0,0 +1,16 @@
...
+Description=Generate sshd AuthorizedKeysFile config for socket activation (workaround)
+DefaultDependencies=no
+After=local-fs.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/bash -c '\
+    auth_keys_file=$(sshd -G | grep -i "^authorizedkeysfile " | cut -d" " -f2-); \
+    echo "SSHD_AUTHORIZED_KEYS_FILE=$auth_keys_file" > /run/sshd-authorized-keys.env'
```
dustymabe: Since it's not the actual file but a string representing the config for `AuthorizedKeysFile`:

```suggestion
    echo "SSHD_AUTHORIZED_KEYS_FILE_CONFIG=$auth_keys_file" > /run/sshd-authorized-keys.env'
```

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system/sshd@.service.d/50-authorized-keys-file.conf
```diff
@@ -0,0 +1,12 @@
+# The upstream sshd@.service overrides AuthorizedKeysFile in the
+# ExecStart command, which excludes .ssh/authorized_keys.d/* where ignition
+# and afterburn place SSH keys. This drop-in reads the configured
+# AuthorizedKeysFile from sshd-authorized-keys-config.service and
+# appends the systemd credentials path so that both user/distro
+# config and ephemeral keys are honoured.
+#
+# See https://github.com/coreos/fedora-coreos-tracker/issues/2098
+[Service]
+EnvironmentFile=/run/sshd-authorized-keys.env
```
dustymabe: > EnvironmentFile=/run/sshd-authorized-keys.env


I think for this we need someone with a security mind to consider the implications here.

i.e. does the existence of this file (and the time between when it's created and when `sshd@.service` is run) allow for any nefarious actions. Certainly if someone were to be able to access it they could point sshd at a different config which defines a different set of files.

I'm going to lean on @travier here if possible.

dustymabe (commented):
This LGTM from my side, but want a security review from @travier if we can get that before we merge if you don't mind @Rolv-Apneseth

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system/sshd@.service.d/50-authorized-keys-file.conf
Rolv-Apneseth: Of course, we can wait to see if he has concerns

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system/sshd-authorized-keys-config.service
Rolv-Apneseth: Fair enough. I'll change after seeing if CI passes

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system/sshd@.service.d/50-authorized-keys-file.conf
Rolv-Apneseth: My understanding is that only `root` can write to that file, but should I add a `umask 077` in case it matters if others can read the file?

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system/sshd-authorized-keys-config.service
Rolv-Apneseth: CI passed, making the change

travier:
I would prefer we add a [generator instead](https://github.com/coreos/fedora-coreos-tracker/issues/2098#issuecomment-4418554794) as those run early and don't modify the graph of dependency / don't add custom units that would be FCOS only. It would mostly run the same command and then would write a config override for the `ExecStart` line in `/run/systemd/system/...` thus there would not be any questions about reading files and permissions.

travier:
We could have the generator fail if the `sshd@.service` `ExecStart` line changes so that we can catch changes in CI.

Or even have a script in the overlay fail.

Rolv-Apneseth:
OK, TIL about [systemd.generator](https://manpages.debian.org/testing/systemd/systemd.generator.7.en.html). Pushed some changes, that does seem better

travier (approved):
Not tested but that looks much cleaner to me. Thanks!

Rolv-Apneseth:
> Not tested but that looks much cleaner to me. Thanks!

It passed locally but we'll see how the CI goes.

I'll let @dustymabe merge if/when he's happy with it. I don't see any movement in the upstream issues unfortunately. **Edit:** actually, https://github.com/systemd/systemd/pull/43157 is relevant. The sshd -g in the generator seems to work fine for FCOS though.

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system-generators/coreos-sshd-generator
```diff
@@ -0,0 +1,32 @@
...
+# AuthorizedKeysFile via sshd -G and appends the systemd credentials
+# path so that both user/distro config and ephemeral keys are honoured.
+#
+# See https://github.com/coreos/fedora-coreos-tracker/issues/2098
+
+NORMAL_DIR="$1"
+
+# First, ensure that the ExecStart line hasn't changed
+upstream_service="/usr/lib/systemd/system/sshd@.service"
+if ! grep -q -- '-o "AuthorizedKeysFile ${CREDENTIALS_DIRECTORY}/ssh.ephemeral-authorized_keys-all .ssh/authorized_keys"' "$upstream_service" 2>/dev/null; then
```
dustymabe, line 14: can you change the grep here slightly (i.e. `XXXAutherizedKeysFile`) to verify this will break our CI when the generator fails so we'll catch it.

dustymabe (approved):
This LGTM.

Should we consider proposing this at the bootc baseimg level instead of here in CoreOS? Doesn't need to block merging this.

overlay.d/16fcos-sshd-workaround/usr/lib/systemd/system-generators/coreos-sshd-generator
Rolv-Apneseth: Confirmed that changing that does make the test fail yes, and the journal has:

```
(generato[966]: /usr/lib/systemd/system-generators/coreos-sshd-generator failed with exit status 1.
```

dustymabe: perfect

dustymabe: Merge when you are ready. We can deal with if we should put in bootc base image or not later I guess.

Rolv-Apneseth:
> Should we consider proposing this at the bootc baseimg level instead of here in CoreOS?

Yeah that's probably a good idea as a follow-up

PR #4272, opened 2026-07-20
[next] switch to create disk images with bootc

jbtrystram:
Using `bootc install to-filesystem` in osbuild.

See https://github.com/coreos/fedora-coreos-tracker/issues/1827


dustymabe:
can we also re-enable `next-devel` and make a similar change there?

dustymabe (approved):

PR #4273, opened 2026-07-20
overlay.d/05core: skip /run cleanup when using iSCSI

dustymabe:
In the case that we don't take down networking we can get into
a weird situation where NetworkManager will detect that the NICs
are already configured and "regnerate" NM connection files in
/run/, but sometimes they miss some information (like DNS) [1].

Let's just skip the cleanup of /run/ files here if we didn't
actually take down networking because of iSCSI.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/2196


dustymabe:
creating as draft for now while I do more testing

dustymabe:
This is ready for review now.

nikita-dubrovskii (approved):
LGTM

PR #4274, opened 2026-07-21
Containerfile: fix quoting for OCP label

dustymabe:
In 60131984 we switched to a heredoc and the downstream builds don't work with it:

```
 + rpm-ostree experimental compose build-chunked-oci
   --bootc --format-version=1 --rootfs /target-rootfs
   --output oci-archive:/run/src/out.ociarchive
   --label com.coreos.inputhash=21c232c31c93fea53caef132eb6633df0f683a560cc0843cdf63017839edb3d4
   --label io.openshift.build.versions=machine-os=9.8.20260720-dev0
   --label io.openshift.build.version-display-names=machine-os=Red Hat Enterprise Linux CoreOS 9.8

error: unexpected argument 'Hat' found
```

Fix the quoting so we don't have this problem.


nikita-dubrovskii (approved):
LGTM

PR #4276, opened 2026-07-23
[next-devel] switch to create disk images with bootc

jbtrystram:

PR #4277, opened 2026-07-23
[next-devel] manifest.yaml: align with testing-devel

c4rt0:
Drop build_with_buildah, switch from fedora-next to fedora (same as testing-devel).

See: https://github.com/coreos/fedora-coreos-config/pull/4276#issuecomment-5062307183

### DIFF
--- a/build-args.conf
+++ b/build-args.conf
@@ -7,7 +7,7 @@ STREAM=next-devel
 DESCRIPTION=Fedora CoreOS next-devel
 VERSION=44
 MUTATE_OS_RELEASE=44
-BUILDER_IMG=quay.io/bootc-devel/fedora-bootc-44-standard@sha256:1093a468e82abb1f410e362ee5e8cb3d938ebe6e92bdc6e87f876fcce44b810a
+BUILDER_IMG=quay.io/bootc-devel/fedora-bootc-44-standard@sha256:3ffaedbf5b28717e6c3ff5f06767c231a0516689acf81c92abab63954ddea72e
 MANIFEST=manifest.yaml
 # XXX: see inject_passwd_group() in build-rootfs
 PASSWD_GROUP_DIR=manifests
--- a/manifest-lock.aarch64.json
+++ b/manifest-lock.aarch64.json
@@ -1,19 +1,19 @@
 {
   "packages": {
     "NetworkManager": {
-      "evra": "1:1.56.0-1.fc44.aarch64"
+      "evra": "1:1.56.1-2.fc44.aarch64"
     },
     "NetworkManager-cloud-setup": {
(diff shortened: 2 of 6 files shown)


c4rt0:
cc. @dustymabe - I hope this is what was requested. I assumed a full alignment was needed since both branches are building F44 now and this was the prerequisit for bootc install. Please let me know if this was supposed to be just the repo swap (fedora-next -> fedora) instead.

dustymabe:
this riscv stuff is noise for now.. let's just drop the `build_with_buildah` and then comment out fedora-next* and uncomment fedora and fedora-updates.

jbtrystram:
`bootc container lint` is failing : 
```
Found unhandled xattrs in files: 148
  "security.selinux\0"
error: reading /proc/1/ns/ipc: Permission denied (os error 13)
Error executing command: Command '['bootc', 'container', 'lint', '--rootfs=/target-rootfs']' returned non-zero exit status 1.
``` 

Not sure why. we are not seeing this on testing-devel

c4rt0:
> Not sure why. we are not seeing this on testing-devel

@jbtrystram: I noticed the [BUILDER_IMG](https://github.com/coreos/fedora-coreos-config/blob/next-devel/build-args.conf#L10) was last edited 3 months ago. 

Would bumping it in next-devel to match testing-devel's fix it?

jbtrystram:
That's an easy thing to try

dustymabe (dismissed):
/lgtm

overlay.d/10disk-images/usr/lib/bootc/install/15-ostree.toml
```diff
@@ -2,4 +2,7 @@
 # Make sure bootc enforces that
 # `/etc/containers/policy.json` includes
 # a default policy that verifies our images signature.
-enforce-container-sigpolicy = true
+# NOTE: enforce-container-sigpolicy requires bootc >= 1.15.2
+# (containers/bootc#2116). Re-enable once the lockfile is
+# bumped past bootc-1.15.1.
+# enforce-container-sigpolicy = true
```
jbtrystram: I'd rather add the lockfile bump manually in this PR as another commit so we don't accidentally forget this.

jbtrystram (changes requested):

overlay.d/10disk-images/usr/lib/bootc/install/15-ostree.toml
c4rt0: Thank you - added the lockfile bump as suggested, reverted the above to it's original state.

jbtrystram (approved):
/lgtm

PR #4278, opened 2026-07-24
[next-devel] build-args.conf: bump BUILDER_IMG to match testing-devel

c4rt0:
Bump next-devel image to match testing-devel. This may possibly fix the bootc container lint failure in #4277.

See: https://github.com/coreos/fedora-coreos-config/pull/4277

### DIFF
--- a/build-args.conf
+++ b/build-args.conf
@@ -7,7 +7,7 @@ STREAM=next-devel
 DESCRIPTION=Fedora CoreOS next-devel
 VERSION=44
 MUTATE_OS_RELEASE=44
-BUILDER_IMG=quay.io/bootc-devel/fedora-bootc-44-standard@sha256:1093a468e82abb1f410e362ee5e8cb3d938ebe6e92bdc6e87f876fcce44b810a
+BUILDER_IMG=quay.io/bootc-devel/fedora-bootc-44-standard@sha256:3ffaedbf5b28717e6c3ff5f06767c231a0516689acf81c92abab63954ddea72e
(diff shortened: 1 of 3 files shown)


jbtrystram:
@c4rt0 look like we should also bump the lockfiles. Let's use what's in `next`.

c4rt0:
I decided to keep all the changes in #4277 instead (manifest.yaml, BUILDER_IMG bump  and the ostree.toml fix). I will be closing this PR soon. 

> @c4rt0 look like we should also bump the lockfiles. Let's use what's in `next`.

For the lockfile bump I commented out enforce-container-sigpolicy in the ostree.toml so the build passes with the current lockfile. Once it merges, the lockfile bot should resume and pick up bootc >= 1.15.2, right?

jbtrystram:
> For the lockfile bump I commented out enforce-container-sigpolicy in the ostree.toml so the build passes with the current lockfile. Once it merges, the lockfile bot should resume and pick up bootc >= 1.15.2, right?

yeah. But the produced image in the meantime is going to be a bit borked. Why wait for the bot ? It's just a commit to cherry-pick since lockfile content should be identical between `next` and `next-devel`

c4rt0:
> yeah. But the produced image in the meantime is going to be a bit borked. Why wait for the bot ? It's just a commit to cherry-pick since lockfile content should be identical between `next` and `next-devel`

Thanks for the suggestion - working on it. Further implementation in the discussed 4277

PR #4280, opened 2026-07-28
[rawhide] overrides: pin console-login-helper-messages-0.21.3

c4rt0:
clhm 0.22.0 breaks gensnippet-os-release and gensnippet-ssh-keys systemd units. Fix is merged upstream (coreos/console-login-helper-messages#137, v0.23.1) but blocked on selinux-policy (fedora-selinux/selinux-policy#3169). Pin to last working version until the selinux fix lands.


nikita-dubrovskii (approved):
/lgtm

PR #4289, opened 2026-08-12
[branched] Branched is now Fedora Linux 45

marmijo:
Fedora 45 has been branched off of rawhide so we can update our branched stream definitions here.

---

The `fedora-bootc-45-standard` image doesn't exist yet, so this PR will remain in draft state until the image is available.


dustymabe:
Can we apply 10f346e here too?

dustymabe:
ok I think this is ready now (crossing fingers on CI passing)

dustymabe:
konflux is failing with some issue cloning the git repo. Maybe branched isn't set up properly in konflux? either way let's get this in if jenkins CI passes and we'll figure out konflux soon.

ravanelli (approved):
/lgmt

PR #4288, opened 2026-08-11
[rawhide] overrides: pin dracut-109-7.fc45

coreosbot-releng:
Requested by @angelcerveraroldan via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).

PR #4287, opened 2026-08-10
denylist: skip ext.config.rpm-ostree.kernel-replace on rawhide

nikita-dubrovskii:
bump-lockfile job fails on this test possibly due to crun 1.29 changes (new mount API, OPEN_TREE_NAMESPACE, trusted /proc fd).

Tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2205

PR #4286, opened 2026-08-10
[rawhide] overrides: unpin fwupd and libjcat

yasminvalim:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/2176

PR #4291, opened 2026-08-12
build-rootfs: always add lockfile repos

joelcapitao:
Previously, we didn't add the `fedora-coreos-pool` repo when building hermetically but it is required since the merge of [1] which privileges this repo when crafting the RPM URLs.
Note that we don't need to modify the pool repo in hermetic mode as it is handled by Hermeto i.e. in cachi2.repo.

[1] https://github.com/coreos/coreos-assembler/pull/4600


joelcapitao:
Tested with https://github.com/coreos/fedora-coreos-config/pull/4290 in rawhide to unblock https://github.com/coreos/fedora-coreos-config/pull/4288

angelcerveraroldan (approved):
LGTM

It should also unblock #4287 and #4286

PR #4295, opened 2026-08-13
tests: generalise matching of bootupctl aleph version output

Rolv-Apneseth:
With https://github.com/coreos/bootupd/pull/1124 generalising the aleph handling to support bootc installs, the output status line changed from 'CoreOS aleph version' to just 'Aleph version'.

Adjust the matches present in our tests to match both cases to handle future bootupd versions and unblock bootupd CI (which is currently failing due to these tests). Also update the error messages to reflect the change.

See https://github.com/coreos/bootupd/pull/1136 for the bootupd companion PR which inspired this.

### DIFF
--- a/tests/kola/boot/bootupd
+++ b/tests/kola/boot/bootupd
@@ -78,8 +78,8 @@ fi
 
 # Verify `bootupctl status` output contains related content.
 # https://github.com/coreos/bootupd/issues/694
-if ! bootupctl status | grep "CoreOS aleph version"; then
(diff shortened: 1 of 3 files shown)


angelcerveraroldan (approved):
LGTM

jbtrystram:
See https://github.com/coreos/fedora-coreos-config/pull/4261

Rolv-Apneseth:
> See #4261

Whoops, thanks for sharing. Approved that one

PR #4296, opened 2026-08-13
manifest: copy RPM configurations to old location

angelcerveraroldan:
RPM config has moved:
- /usr/share/dnf5/repos.d -> /etc/yum.repos.d
- /usr/share/pki/rpm-gpg -> /etc/pki/rpm-gpg

Right now rpm-ostree is not ready for this change, so for the time being we will copy the files from the new location to the old location

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -35,6 +35,19 @@ conditional-include:
     include:
       ostree-layers:
         - overlay.d/10aarch64
+  # Temporary workaround. RPM repo configuration has moved from /etc/yum.repos.d
+  # to /usr/share/dnf5/repos.d, leading to rpm-ostree to no longer work. Until
+  # we are ready to support the new path, we can copy the config over.
+  #
+  # See: https://github.com/coreos/fedora-coreos-tracker/issues/2172
+  - if: releasever >= 45
+    include:
+      postprocess:
+        - |
+          #!/usr/bin/bash
+          set -eux -o pipefail
+          cp /usr/share/dnf5/repos.d/* /etc/yum.repos.d/
+          cp /usr/share/pki/rpm-gpg/* /etc/pki/rpm-gpg/
   # Temporary workaround. containers-common moved policy.json from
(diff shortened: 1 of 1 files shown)


manifests/fedora-coreos.yaml
```diff
@@ -35,6 +35,19 @@ conditional-include:
     include:
       ostree-layers:
         - overlay.d/10aarch64
+  # Temporary workaround. RPM repo configuration has moved from /etc/yum.repos.d
+  # to /usr/share/dnf5/repos.d, leading to rpm-ostree to no longer work. Until
+  # we are ready to support the new path, we can copy the config over.
+  #
+  # See: https://github.com/coreos/fedora-coreos-tracker/issues/2172
+  - if: stream == "rawhide"
```
dustymabe: let's make this something like `if: releasever >= 45` instead since this is going to be in f45 too (and will immediately affect our `branched` stream too).

manifests/fedora-coreos.yaml
```diff
@@ -35,6 +35,19 @@ conditional-include:
...
+    include:
+      postprocess:
+        - |
+          #!/usr/bin/bash
+          set -eux -o pipefail
+          cp /usr/share/dnf5/repos.d/* /etc/yum.repos.d
+          cp /usr/share/pki/rpm-gpg/* /etc/pki/rpm-gpg
```
dustymabe:
```suggestion
          cp /usr/share/dnf5/repos.d/* /etc/yum.repos.d/
          cp /usr/share/pki/rpm-gpg/* /etc/pki/rpm-gpg/
```

my mind reads this easier

manifests/fedora-coreos.yaml
dustymabe, line 68: I wonder if we can drop this workaround now

angelcerveraroldan: We can, I will make a follow up PR for this

angelcerveraroldan:
I tested these changes locally (rawhide image), and it helped with most of the errors. There are some that are due to another issue that I am looking into.

dustymabe (approved):
LGTM

PR #4299, opened 2026-08-13
[next] overrides: unpin fwupd and libjcat

yasminvalim:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/2176


dustymabe:
For `testing` and `next` branches these pins will get cleaned up on the next promotion. We don't need a separate PR for it.

PR #4300, opened 2026-08-14
denylist: kernel-replace on rawhide is failing

angelcerveraroldan:
Due to the transition from f45 to f46 we are seeing a failure on the kernel-replace test on all arches.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2209

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -22,6 +22,13 @@
   arches:
     - aarch64
 
+- pattern: ext.config.rpm-ostree.kernel-replace
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/2209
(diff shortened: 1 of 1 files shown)


angelcerveraroldan:
For the time being, lets denylist the test so that the CI is unblocked. I am going to look into modifying the way we download the packages so that this test works consistently, as we will likely see this issue again when we transition from `f46` to `f47`.

angelcerveraroldan:
I see that there is another PR #4287 that is also trying to denylist this test on all arches for rawhide

Rolv-Apneseth (approved):
LGTM

angelcerveraroldan:
Closing this PR; rather than deny-list we will modify the test.  #4301

PR #4301, opened 2026-08-14
tests/kernel-replace: skip on rawhide kernel collision

angelcerveraroldan:
Right after a fedora branch such as f45 -> f46, the newly branched
release can have the same kernel as the previous release. This leads
to errors as the expected files will not download as they already
exist.

### DIFF
--- a/tests/kola/rpm-ostree/kernel-replace
+++ b/tests/kola/rpm-ostree/kernel-replace
@@ -70,6 +70,7 @@ build_derived_image() {
     contextdir=$(mktemp -d)
     pushd ${contextdir}
 
+	running_kver=$(rpm -q kernel --qf '%{EVR}')
     case "$OS_ID" in
       rhcos|scos|rhel|centos)
         # Use the matching CentOS Stream version based on RHEL/SCOS major version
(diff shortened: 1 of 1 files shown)


tests/kola/rpm-ostree/kernel-replace
```diff
@@ -84,25 +85,30 @@ build_derived_image() {
...
+                  grep -v "${running_kver}" | tail -n1)
         ;;
       fedora)
         VERSION_ID=$(. /etc/os-release; echo $VERSION_ID)
         previous_version_id=$((VERSION_ID - 1))
         # Workaround https://github.com/coreos/fedora-coreos-tracker/issues/2113
         # with the skip_system_repo_lock=true setting in dnf configuration file.
         printf "[main]\nskip_system_repo_lock=true" > dnf_config
-        dnf download --config=dnf_config --releasever "${previous_version_id}" \
-          --resolve --disablerepo=* --enablerepo=updates --enablerepo=fedora kernel
```
dustymabe, line 97: my suggestion here is to leave this logic the same but just add `kernel-core` to the above line at the end and then compare `$kver` to `$running_kver` below and do the `ok "Skipping` if they match and on rawhide.

dustymabe: basically the reason we are hitting an error today is because `dnf download --resolve` will only download the kernel RPM if the version in the repos is the same as what's installed because it's the only one you told it to download and all of the other deps were installed, which is why the `rpm -qp --qf '%{EVR}' ./kernel-core*rpm` fails below.

dustymabe (approved):
LGTM

PR #4304, opened 2026-08-14
[next-devel] overrides: unpin fwupd and libjcat

yasminvalim:
Fixes: coreos/fedora-coreos-tracker#2176


manifest-lock.overrides.yaml
```diff
@@ -9,13 +9,3 @@
 # for FCOS-specific packages (ignition, afterburn, etc.).
 
 packages:
```
dustymabe:
```suggestion
packages: {}
```

dustymabe (changes requested):

dustymabe (approved):

PR #4306, opened 2026-08-17
fasttrack: Fasttrack bootupd 0.2.36-1.fc46

Johan-Liebert1:


Johan-Liebert1:
Didn't realize I had built for f46

Johan-Liebert1:
I'll update this PR once we have releases for earlier fedora version

PR #4311, opened 2026-08-20
Rawhide now maps to Fedora 46

AdamWill:
Fedora 45 has branched.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/2210

### DIFF
--- a/build-args.conf
+++ b/build-args.conf
@@ -5,8 +5,8 @@ ID=fedora
 NAME=fedora-coreos
 STREAM=rawhide
 DESCRIPTION=Fedora CoreOS rawhide
-VERSION=45
-MUTATE_OS_RELEASE=45
+VERSION=46
+MUTATE_OS_RELEASE=46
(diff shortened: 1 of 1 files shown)


AdamWill:
Note this does not update `BUILDER_IMG` as the corresponding commit for F44 branching did. It probably should, but fedora-bootc-rawhide-standard has [not been built since before branching](https://gitlab.com/fedora/bootc/base-images/-/work_items/83) so that needs to be fixed first.

dustymabe:
> Note this does not update `BUILDER_IMG` as the corresponding commit for F44 branching did. It probably should, but fedora-bootc-rawhide-standard has [not been built since before branching](https://gitlab.com/fedora/bootc/base-images/-/work_items/83) so that needs to be fixed first.

correct. We were waiting on f46 based images for bootc to exist first before opening a PR like this.

dustymabe:
Thanks for trying to push things forward here @AdamWill.

This PR was obsoleted by https://github.com/coreos/fedora-coreos-config/pull/4320 (needed a few other changes in the same PR).

PR #4318, opened 2026-08-24
Switch our gpg key path locations to /usr/

dustymabe:
Part of:

- https://github.com/coreos/fedora-coreos-tracker/issues/2172
- https://fedoraproject.org/wiki/Changes/RelocateRpmRepoConfigsToUsr

Because of [1] we now have the gpgkeys also in /usr on F44 so we can
go ahead and start using those paths everywhere.

[1] https://src.fedoraproject.org/rpms/fedora-repos/c/107895f6f8dc26d329d3e40dbd5ccbabe1e8f301?branch=f44

### DIFF
--- a/build-rootfs
+++ b/build-rootfs
@@ -117,10 +117,10 @@ def inject_yumrepos(srcdir):
             shutil.copy(repo, "/etc/yum.repos.d")
 
 
-def run_buildroot_prep():
+def run_buildroot_prep(srcdir):
     # This allows FCOS/SCOS/RHCOS to do specific things before the main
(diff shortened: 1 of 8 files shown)


AdamWill:
I don't know the flows here in detail, but as a general note, remember to consider the "updating from an older 43/44-based image to a newer one" path and make sure things will work for that?

dustymabe:
> remember to consider the "updating from an older 43/44-based image to a newer one" path and make sure things will work for that?

We should be covered since newer F44 based FCOS (current production streams) will have both gpgkeys in /etc/ and /usr.

ravanelli (approved):
lgtm

PR #4325, opened 2026-08-27
DRAFT: Use image-builder generic-iso to build coreos live iso

Roshan-R:
Build CoreOS ISOs using image-builder's generic-iso 

The image-builder used to build the ISO comes from an unmerged image-builder fork containing the changes from [osbuild/image-builder#2414.](https://github.com/osbuild/image-builder/pull/2414)

This PR builds on @jbtrystram 's work to get CoreOS ISOs working with image-builder's generic-iso pipeline. It adds the Fedora CoreOS-specific configuration needed to build a bootable live ISO using `image-builder` instead of coreos-assembler.

The generated ISO currently boots successfully when SELinux is disabled with selinux=0. There are still a number of workarounds in place, and the resulting ISO has several limitations that need to be addressed before this can provide a complete replacement for the existing CoreOS ISO build.

This PR adds the FCOS-specific pieces needed on top of image-builder's generic ISO infrastructure:

- Add the GRUB CD bootloader, EROFS, `dracut-live`, and `shim` packages required by the generic ISO.
- Provide the EFI layout expected by image-builder's GRUB ISO stage.
- Add an FCOS-specific iso.yaml with the required live ISO kernel arguments and serial console configuration.
- Disable composefs for the live image, since the EROFS/overlayfs root setup is incompatible with composefs.
- Adapt the FCOS live dracut modules to image-builder's ISO layout, including avoiding the boot partition and rootfs.img paths used by coreos-assembler media.
- Dynamically size the root partition based on the generated rootfs instead of using a fixed disk size.

### Offline installation is not supported.

The ISO does not currently contain the CoreOS .osmet files required for offline installation with coreos-installer.
See: https://github.com/joelcapitao/bib-fcos-experimentation/issues/9

### coreos-installer iso * customization is not supported.

The ISO is missing the CoreOS-specific embed areas required by coreos-installer for ISO customization.
See: https://github.com/joelcapitao/bib-fcos-experimentation/issues/102

### Only x86_64 has been tested.
The configuration includes support for other architectures, but aarch64 and other architectures have not yet been tested.

### ~SELinux must currently be disabled.~
~The generated ISO only boots successfully with selinux=0. Without this, lot of systemd services fail due to selinux permission issues.~

### The live-image stamp file is created in the dracut module.

The stamp file checked by is-live-image is currently created from within the dracut module itself. Ideally, this file should be created as part of the ISO build process rather than being added by the initramfs module.


Roshan-R:
I was able to get SELinux working. The issue was caused by systemd-etc.mount failing because dmsquash-live mounts the full overlayfs at /sysroot, which means /sysroot/etc is already covered by the overlayfs mount.

When sysroot-etc.mount subsequently tries to bind mount /run/ephemeral/etc over /sysroot/etc, systemd detects that the path is already covered by the existing overlayfs mount. As a result, the mount condition evaluates to false and systemd skips the bind mount entirely.

To work around this, I replaced systemd-etc.mount with a systemd-etc-setup.service that explicitly unmounts the existing /sysroot/etc mount before setting up the bind mount. This allows the /run/ephemeral/etc bind mount to be applied correctly and makes SELinux work as expected.

manifests/system-configuration.yaml
```diff
@@ -36,6 +36,17 @@ packages:
   # we need dmsquash-live dracut module to create the live iso
   # in image-builder.
   - dracut-live
+  # erofs-utils provides mkfs.erofs, needed by osbuild when building live ISOs
+  - erofs-utils
+  # python is needed by osbuild runners
+  - python3
```
jbtrystram, line 42: why do we need this ? Aren´t we using another container as the build environment ?

manifests/system-configuration.yaml
```diff
@@ -36,6 +36,17 @@ packages:
...
+  # grub2-efi-x64-cdboot provides gcdx64.efi needed by osbuild's grub2.iso
+  # stage when building live ISOs
+  - shim
+  - grub2-efi-*-cdboot
+  - grub2-efi-x64-cdboot
```
jbtrystram, line 47: One of those two is redundant

manifests/system-configuration.yaml
```diff
@@ -56,3 +67,21 @@ postprocess:
...
+    mkdir -p /boot/efi/EFI/fedora
+    for f in /usr/lib/efi/shim/*/EFI/fedora/* /usr/lib/efi/grub2/*/EFI/fedora/*; do
+      ln -sfn "${f}" /boot/efi/EFI/fedora/"$(basename "${f}")"
+    done
+  # Disable composefs for ostree as it is incompatible with the erofs rootfs
+  # used by the bootc-generic-iso pipeline.
+  - |
+    #!/usr/bin/bash
+    set -eux -o pipefail
+    printf '[composefs]\nenabled = no\n[sysroot]\nreadonly = false\n' > /usr/lib/ostree/prepare-root.conf
```
jbtrystram, line 87: These are prime examples of "why we can't do this in the base FCOS image"

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/module-setup.sh
```diff
@@ -58,4 +58,11 @@ install() {
 
     install_and_enable_unit "coreos-livepxe-persist-osmet.service" \
         "initrd.target"
+
+    # HACK: Create the stamp file that is-live-image checks to detect a live boot.
+    # In cosa-built ISOs this is created by buildextend-live;
+    # This should be created inside while creating the ISO
+    # by the image-builder
+    mkdir -p "${initdir}/etc"
+    : > "${initdir}/etc/coreos-live-initramfs"
```
jbtrystram, line 67: We already have 
https://github.com/coreos/fedora-coreos-config/blob/6afed2b7af0a08fbe9bafda089cc1844992718ec/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator#L27-L28

What does having this stamp burned at build-time brings ?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/ostree-cmdline.sh
```diff
@@ -22,10 +22,12 @@ case "${1:-unset}" in
         # anyway but ostree-prepare-root still wants to e.g. remount `/etc`.
         # Could tweak the logic there, but for now just mask the file entirely.
         # We have our own /var and /etc transient mounts.
-        mount --bind /dev/null /usr/lib/ostree/prepare-root.conf
+
+        # Commenting out the old bind mount to /dev/null
+        # mount --bind /dev/null /usr/lib/ostree/prepare-root.conf
```
jbtrystram, line 27: why is this necessary ?

overlay.d/10disk-images/usr/lib/dracut/dracut.conf.d/10-dmsquash-live.conf
```diff
@@ -1,2 +1,4 @@
 # image-builder requires the dracut dmsquash-live to build the live iso
-#add_dracutmodules+=" dmsquash-live livenet ostree"
+add_dracutmodules+=" qemu qemu-net dmsquash-live livenet ostree"
+early_microcode="no"
```
jbtrystram, line 3: why do we need this ?

overlay.d/10disk-images/usr/lib/image-builder/bootc/iso.yaml
```diff
@@ -0,0 +1,10 @@
+label: fedora-coreos
+kernel_args:
+  - console=tty0
+  - console=ttyS0
```
jbtrystram, line 4: these should be set via blueprint

jbtrystram:
I am closing this since this should live in bib-fcos-experimentation for now as an extra container build.

overlay.d/10disk-images/usr/lib/dracut/dracut.conf.d/10-dmsquash-live.conf
Roshan-R: added this as it was documented in https://github.com/brlane-rht/images/blob/dff612d79cdc7ee082f6857daa0822da70c76984/doc/20-advanced/20-bootc/10-isos.md

Roshan-R:
Closed in favor of https://github.com/joelcapitao/bib-fcos-experimentation/pull/112

PR #4326, opened 2026-08-27
overrides: fast-track ignition-2.27.0-1.fc44

coreosbot-releng:


dustymabe:
hmm. let's bring this into `branched` and `rawhide` first if we can (since it's an F45 change).

Technically this isn't backwards incompatible, so maybe we can also go ahead and introduce in `next-devel` (and thus `next`).

but maybe let's give it a cycle of soak time before introducing in `testing`?

prestist:
> hmm. let's bring this into `branched` and `rawhide` first if we can (since it's an F45 change).
> 
> Technically this isn't backwards incompatible, so maybe we can also go ahead and introduce in `next-devel` (and thus `next`).
> 
> but maybe let's give it a cycle of soak time before introducing in `testing`?

Sure! thats fine, I will make the fast-track for branched.

PR #4327, opened 2026-08-27
[next-devel] overrides: fast-track ignition-2.27.0-1.fc44

coreosbot-releng:


dustymabe (approved):
LGTM - note the comment I made over in https://github.com/coreos/fedora-coreos-config/pull/4326#issuecomment-5441168100

PR #4333, opened 2026-08-31
Enable systemd-oomd and zram swap

Nemric:
Going to https://forge.fedoraproject.org/fesco/tickets/issues/3661
Add zram-generator.conf file to enable zram swap
Remove systemd-oomd preset overrides to enable systemd-oomd service and socket

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -45,8 +45,8 @@ RUN chmod -R a=rX,u+w /src
 RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
         rm -rf /cache/cache/*lock*
 RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
-    --mount=type=secret,id=yumrepos,target=/etc/yum.repos.d/secret.repo \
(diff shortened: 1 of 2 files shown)


dustymabe:
probably the more appropriate thing to do here is install the `zram-generator-defaults` package which will provide `/etc/systemd/zram-generator.conf`.

Also we'll need to update the tests in `tests/kola/swap/zram-generator/`.

dustymabe:
we also kind of need this change to only apply to F45+ - I'll have to brainstorm a bit on the best way to achieve that.

jbtrystram:
> we also kind of need this change to only apply to F45+ - I'll have to brainstorm a bit on the best way to achieve that.

can't we move the preset disabling it to a separate overlay that we only ship in <45 ?

Like for the [sshd workaround](https://github.com/coreos/fedora-coreos-config/blob/ac61d9f8a992b7d997b8052868b677cdb7187f9b/manifests/fedora-coreos.yaml#L51-L55)

PR #4335, opened 2026-09-02
overrides: pin libselinux and related packages to 3.10-*.fc44

Rolv-Apneseth:
See https://github.com/coreos/fedora-coreos-tracker/issues/2216. Turns out just pinning `policycoreutils` isn't enough since the affected code is actually in `libselinux`.

Also add a test to catch this kind of regression going forward.


Rolv-Apneseth:
Expected failure in the rpm diff:

```
15:48:04  Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.
```

tlbueno (approved):
lgtm

PR #4336, opened 2026-09-02
copy-network: stop reading firstboot network config from /boot

ravanelli:
- coreos-installer now writes network config files directly to /etc/coreos-firstboot-network instead of /boot (see coreos/coreos-installer#1765).

 - Remove the /boot mount and the live-vs-diskful from coreos-copy-firstboot-network, and drop the boot-device dependency from the systemd unit.

 - https://github.com/coreos/coreos-installer/issues/1758
 
 
 ### Handle legacy firstboot network config in /boot
 - coreos-installer now embeds network config in the initramfs
   instead of writing to /boot/coreos-firstboot-network. When a
   system is provisioned with an older coreos-installer but boots
   a newer image, the network config is stranded in /boot and
   never applied — breaking network-dependent boot paths like iSCSI.

 - Add a three-layer backward compatibility mechanism:

   1. Initramfs rescue service (coreos-rescue-misplaced-network):
      if no config exists in /etc/coreos-firstboot-network, mount
      /boot and copy any legacy config found there so that the
      existing coreos-copy-firstboot-network.service can apply it.
      This is the active fix — it recovers the stranded config and
      allows boot to succeed.

   2. Dracut emergency hook (coreos-warn-misplaced-network.sh):
      on iSCSI systems, /boot lives on a network device — but the
      network config needed to reach it is stranded inside /boot
      itself, so the rescue service cannot help. When this
      chicken-and-egg problem causes the system to drop to an
      emergency shell, this hook explains the situation and advises
      re-installing with a newer coreos-installer.

   3. Real-root warning service (coreos-warn-misplaced-network):
      runs after boot has already succeeded (thanks to the rescue
      service above). It does not fix anything — it warns via
      kmsg/console that legacy config was found in /boot, prompting
      the user to re-install with a newer coreos-installer so
      future boots do not depend on the rescue path. If /boot is
      not mounted at all, it warns that this may itself be a
      symptom of missing network config.
      
    4. We may drop it in the feature, since there is a cost of keeping checking it

### DIFF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
@@ -103,36 +103,30 @@ if ! cmdline_bool 'ignition.firstboot'; then
     exit 0
 fi
 
-# coreos-ignition-setup-user.service and `coreos-copy-firstboot-network.service`
+# coreos-ignition-setup-user.service
 # should depend on the boot device node only on diskful boots
 mkdir -p "${UNIT_DIR}/coreos-ignition-setup-user.service.d"
-mkdir -p "${UNIT_DIR}/coreos-copy-firstboot-network.service.d"
 cat > "${UNIT_DIR}/coreos-ignition-setup-user.service.d/diskful.conf" <<EOF
 [Unit]
 Requires=dev-disk-by\x2dlabel-boot.device
 After=dev-disk-by\x2dlabel-boot.device
 EOF
-cp "${UNIT_DIR}/coreos-ignition-setup-user.service.d/diskful.conf" \
-    "${UNIT_DIR}/coreos-copy-firstboot-network.service.d/diskful.conf"
 
-# If booting from iSCSI, then we need networking first before we
-# can see the bootfs. This has some implications.
+# If booting from iSCSI, the boot device (where the Ignition config lives)
+# is only reachable after networking is up. Note: this is about the Ignition
+# config, not the network config — network config is already in the initramfs
+# (placed there by coreos-installer) and handled by
+# coreos-copy-firstboot-network.service without needing the boot device.
+#
+# Since `ignition-fetch-offline.service` tries to read the Ignition config
+# from the boot device immediately (before networking), it would fail on
+# iSCSI. We neuter it and let `ignition-fetch.service` (which runs after
+# networking is up) fetch the Ignition config instead.
 if test -n "$(cmdline_arg rd.iscsi.firmware)" || test -n "$(cmdline_arg netroot)"; then
     mkdir -p "${EARLY_DIR}/ignition-complete.target.requires"
-
-    # In the injected Ignition config case, `coreos-ignition-setup-user.service`
-    # can't "pass" the config to `ignition-fetch-offline.service`. So we neuter
-    # the latter so that `ignition-fetch.service` (which runs after networking
-    # comes up) can pick it up instead.
     ln -sf /dev/null "${EARLY_DIR}/ignition-complete.target.requires/ignition-fetch-offline.service"
-    # activate ignition-fetch.service
     mkdir -p /run/ignition
     touch /run/ignition/neednet
-
-    # With iSCSI, configuring networking via kargs is a hard requirement
-    # since it can't be picked up from the bootfs (i.e. injected via
-    # `--copy-network`). So neuter `coreos-copy-firstboot-network.service`.
-    ln -sf /dev/null "${EARLY_DIR}/ignition-complete.target.requires/coreos-copy-firstboot-network.service"
 fi
 
 # create symlink for udev rule
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
@@ -1,17 +1,11 @@
 # This unit will run early in boot and detect if:
-# - In the diskful case, the user copied in firstboot networking config files
-#   into `/boot` (most likely by using `coreos-installer install
+# - user embedded firstboot networking config files
+#   into the initramfs `/etc` (by using `coreos-installer install
 #   --copy-network`).
-# - In the live case, the user provided firstboot networking config files in
-#   `/etc` (most likely by using `coreos-installer iso network embed`).
 #
 # Since this unit is modifying network configuration there are some
 # dependencies that we have:
 #
-# - In the diskful case, we need to look for networking configuration on the
-#   /boot partition
-#     - i.e. after /dev/disk/by-label/boot is available
-#     - which is injected via coreos-diskful-generator
 # - Need to run before networking is brought up.
 #     - This is done in nm-initrd.service [1]
 #     - i.e. Before=nm-initrd.service
@@ -23,8 +17,7 @@
 #         - Another is to run *after* nm-config.sh [2] in dracut-cmdline [3]
 #           and just delete all the files created by nm-initrd-generator.
 #             - i.e. After=dracut-cmdline.service, but Before=nm-initrd.service
-#     - We'll go with the second option here because the need for the /boot
-#       device (mentioned above) means we can't start before dracut-cmdline.service
+#     - We'll go with the second option here.
 #
 # [1] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/nm-initrd.service
 # [2] https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/nm-config.sh
@@ -33,6 +26,7 @@
 [Unit]
 Description=Copy CoreOS Firstboot Networking Config
 ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=/etc/coreos-firstboot-network
 DefaultDependencies=false
 # We're pulled in by ignition-complete.target; as good practice, add a matching
 # Before to be explicit about it gating on this unit passing.
@@ -41,22 +35,12 @@ Before=nm-initrd.service
 # compat: remove when everyone is on dracut 054+
 Before=dracut-initqueue.service
 After=dracut-cmdline.service
-# And since the boot device may be on multipath; optionally wait for it to
-# appear via the dynamic target.
-After=coreos-multipath-wait.target
 # Need to run after coreos-enable-network since it may re-run the NM cmdline
 # hook which will generate NM configs from the network kargs, but we want to
 # have precedence.
 After=coreos-enable-network.service
-# We've seen races with ignition-kargs.service, which accesses /boot rw.
-# Let's introduce some ordering here. Need to use `Before` because otherwise
-# we get a systemd ordering cycle. https://github.com/coreos/fedora-coreos-tracker/issues/883
-Before=ignition-kargs.service
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-# The MountFlags=slave is so the umount of /boot is guaranteed to happen
-# /boot will only be mounted for the lifetime of the unit.
-MountFlags=slave
 ExecStart=/usr/sbin/coreos-copy-firstboot-network
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
@@ -32,8 +32,6 @@ DefaultDependencies=false
 # Before to be explicit about it gating on this unit passing.
 Before=ignition-complete.target
 Before=nm-initrd.service
-# compat: remove when everyone is on dracut 054+
-Before=dracut-initqueue.service
 After=dracut-cmdline.service
 # Need to run after coreos-enable-network since it may re-run the NM cmdline
 # hook which will generate NM configs from the network kargs, but we want to
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-rescue-misplaced-network.service
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-rescue-misplaced-network.service
@@ -1,0 +1,21 @@
+# If an older coreos-installer wrote firstboot network config to
+# /boot/coreos-firstboot-network, copy it to /etc so that
+# coreos-copy-firstboot-network.service can pick it up.
+# Only runs when /etc/coreos-firstboot-network does not already exist
+# (i.e. no config was embedded in the initramfs by a newer installer).
+[Unit]
+Description=Rescue Misplaced Firstboot Network Config From /boot
+ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=!/etc/coreos-firstboot-network
+DefaultDependencies=false
+Before=coreos-copy-firstboot-network.service
+After=dracut-cmdline.service
+After=coreos-enable-network.service
+# Avoid races with ignition-kargs.service which accesses /boot rw.
+Before=ignition-kargs.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+MountFlags=slave
+ExecStart=/usr/sbin/coreos-rescue-misplaced-network
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-rescue-misplaced-network.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-rescue-misplaced-network.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -euo pipefail
+
+# Rescue firstboot network config that an older coreos-installer placed
+# in /boot/coreos-firstboot-network.  Copy it to /etc so that
+# coreos-copy-firstboot-network can apply it normally.
+
+bootmnt=/mnt/boot_partition
+bootdev=/dev/disk/by-label/boot
+legacy_dir_name="coreos-firstboot-network"
+boot_firstboot_network_dir="${bootmnt}/${legacy_dir_name}"
+etc_firstboot_network_dir="/etc/${legacy_dir_name}"
+
+if [ ! -b "${bootdev}" ]; then
+    echo "info: boot device ${bootdev} not found; skipping legacy network rescue"
+    exit 0
+fi
+
+mkdir -p "${bootmnt}"
+if ! mount -o ro "${bootdev}" "${bootmnt}"; then
+    echo "warning: failed to mount ${bootdev}; skipping legacy network rescue"
+    exit 0
+fi
+
+if [ -n "$(ls -A "${boot_firstboot_network_dir}" 2>/dev/null)" ]; then
+    echo "warning: found legacy firstboot network config in ${boot_firstboot_network_dir}" | tee /dev/kmsg || true
+    echo "warning: an older coreos-installer placed network config in /boot instead of the initramfs" | tee /dev/kmsg || true
+    echo "warning: rescuing config — copying to ${etc_firstboot_network_dir}" | tee /dev/kmsg || true
+    echo "warning: re-install using an updated coreos-installer to avoid this in the future" | tee /dev/kmsg || true
+    mkdir -p "${etc_firstboot_network_dir}"
+    cp -v "${boot_firstboot_network_dir}"/* "${etc_firstboot_network_dir}/"
+else
+    echo "info: no legacy network config in ${boot_firstboot_network_dir}; skipping"
+fi
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-warn-misplaced-network.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-warn-misplaced-network.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Emergency hook: warn that an older coreos-installer may have written
+# firstboot network config to /boot, which this image no longer reads.
+# In the iSCSI case that missing config prevents the network from
+# coming up, which in turn prevents mounting root.
+
+if [ -d /etc/coreos-firstboot-network ] && \
+   [ -n "$(ls -A /etc/coreos-firstboot-network 2>/dev/null)" ]; then
+    # Config is in the initramfs — this emergency is not caused by
+    # misplaced network config.
+    return 0
+fi
+
+cat <<EOF
+
+------
+If you used 'coreos-installer install --copy-network' with an older
+version of coreos-installer, the network configuration may have been
+written to /boot/coreos-firstboot-network.
+
+This CoreOS image no longer reads firstboot network config from /boot.
+The configuration must be embedded in the initramfs.
+
+Re-install using an updated version of coreos-installer.
+------
+
+EOF
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/module-setup.sh
@@ -23,6 +23,11 @@ install() {
     install_and_enable_unit "coreos-enable-network.service" \
         "ignition-complete.target"
 
+    inst_simple "$moddir/coreos-rescue-misplaced-network.sh" \
+        "/usr/sbin/coreos-rescue-misplaced-network"
+    install_and_enable_unit "coreos-rescue-misplaced-network.service" \
+        "ignition-complete.target"
+
     inst_simple "$moddir/coreos-copy-firstboot-network.sh" \
         "/usr/sbin/coreos-copy-firstboot-network"
     install_and_enable_unit "coreos-copy-firstboot-network.service" \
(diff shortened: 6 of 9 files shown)


ravanelli:
Do not merge yet, it still need the feature check part

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
```diff
@@ -1,17 +1,11 @@
 # This unit will run early in boot and detect if:
-# - In the diskful case, the user copied in firstboot networking config files
-#   into `/boot` (most likely by using `coreos-installer install
+# - user copied in firstboot networking config files
+#   into `/etc` (by using `coreos-installer install
```
dustymabe:
```suggestion
# - user embedded firstboot networking config files
#   into  the initramfs `/etc` (by using `coreos-installer install
```

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
```diff
@@ -33,13 +26,14 @@
...
 ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=/etc/coreos-firstboot-network
 DefaultDependencies=false
 # We're pulled in by ignition-complete.target; as good practice, add a matching
 # Before to be explicit about it gating on this unit passing.
+Before=network-pre.target ignition-diskful.target
 Before=ignition-complete.target
 Before=nm-initrd.service
-# compat: remove when everyone is on dracut 054+
-Before=dracut-initqueue.service
```
dustymabe, line 42: if we are removing this because everyone (downstreams) is on 054+ let's do it in a separate commit.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
```diff
@@ -33,13 +26,14 @@
...
+Wants=network-pre.target
```
dustymabe: why the new dep here?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
```diff
@@ -33,13 +26,14 @@
 [Unit]
 Description=Copy CoreOS Firstboot Networking Config
 ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=/etc/coreos-firstboot-network
 DefaultDependencies=false
 # We're pulled in by ignition-complete.target; as good practice, add a matching
 # Before to be explicit about it gating on this unit passing.
+Before=network-pre.target ignition-diskful.target
```
dustymabe: why are these new deps needed?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service
dustymabe: we no longer depend on `/boot` so.. drop this?

dustymabe: we no longer access `/boot` so might be able to drop this.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator
```diff
@@ -103,36 +103,30 @@ if ! cmdline_bool 'ignition.firstboot'; then
...
 [Unit]
 Requires=dev-disk-by\x2dlabel-boot.device
 After=dev-disk-by\x2dlabel-boot.device
 EOF
-cp "${UNIT_DIR}/coreos-ignition-setup-user.service.d/diskful.conf" \
-    "${UNIT_DIR}/coreos-copy-firstboot-network.service.d/diskful.conf"
 
-# If booting from iSCSI, then we need networking first before we
-# can see the bootfs. This has some implications.
+# If booting from iSCSI, the boot device (where the Ignition config lives)
```
dustymabe, line 115: hmm. part of me feels like we should also just place the ignition config into the initrd as well. This would be a separate feature request.

The idea here being we would still support `/boot/ignition/config.ign` but also `/etc/ignition/config.ign` (i.e. initramfs) and then we'd just drop support for the special handling in the diskful-generator here. i.e. for iscsi we'd only support `/etc/ignition/config.ign` and not `/boot`

overlay.d/05core/usr/libexec/coreos-warn-misplaced-network
```diff
@@ -0,0 +1,39 @@
...
+        warn "Network config placed there by an older coreos-installer was NOT applied."
+        warn ""
+        warn "This is especially critical for systems that require networking in early"
+        warn "initramfs (e.g. iSCSI boot) since the config was not available when needed."
+        warn ""
+        warn "ACTION REQUIRED: Re-install using an updated 'coreos-installer' that"
+        warn "embeds network config in the initramfs."
+        warn "========================================================================"
+    fi
+else
```
dustymabe, line 27: I wonder if we just skip this effort if `/boot` isn't mounted for some reason?

overlay.d/05core/usr/libexec/coreos-warn-misplaced-network
```diff
@@ -0,0 +1,39 @@
...
+        warn ""
+        warn "This CoreOS image no longer reads firstboot network config from /boot."
+        warn "Network config placed there by an older coreos-installer was NOT applied."
+        warn ""
+        warn "This is especially critical for systems that require networking in early"
+        warn "initramfs (e.g. iSCSI boot) since the config was not available when needed."
+        warn ""
+        warn "ACTION REQUIRED: Re-install using an updated 'coreos-installer' that"
+        warn "embeds network config in the initramfs."
+        warn "========================================================================"
```
dustymabe, line 25: This feels kind of like something we should drop down as part of a CLHM.

Maybe the service only runs on first boot and writes out a persistent CLHM into `/etc/` and then instructs the user to delete the file if they want to acknowledge the warning and not see the warning message again?

overlay.d/05core/usr/lib/systemd/system/coreos-warn-misplaced-network.service
```diff
@@ -0,0 +1,12 @@
...
+After=boot.mount
+Before=emergency.service rescue.service multi-user.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/coreos-warn-misplaced-network
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target emergency.target rescue.target
```
dustymabe, line 12: With this service, what happens on upgrading systems. i.e. systems that were installed a year ago and used `--copy-network`? do they get the warning too?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-rescue-misplaced-network.service
```diff
@@ -0,0 +1,21 @@
...
+# /boot/coreos-firstboot-network, copy it to /etc so that
+# coreos-copy-firstboot-network.service can pick it up.
+# Only runs when /etc/coreos-firstboot-network does not already exist
+# (i.e. no config was embedded in the initramfs by a newer installer).
+[Unit]
+Description=Rescue Misplaced Firstboot Network Config From /boot
+ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=!/etc/coreos-firstboot-network
+DefaultDependencies=false
+Before=coreos-copy-firstboot-network.service
```
dustymabe, line 11: delete `Before=coreos-copy-firstboot-network.service` 

this one has `ConditionPathExists=!/etc/coreos-firstboot-network` and `coreos-copy-firstboot-network.service` has `ConditionPathExists=/etc/coreos-firstboot-network` so they'll never run at the same time?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-rescue-misplaced-network.service
```diff
@@ -0,0 +1,21 @@
...
+After=dracut-cmdline.service
+After=coreos-enable-network.service
```
dustymabe, line 13: looks like the original had:

```
# And since the boot device may be on multipath; optionally wait for it to
# appear via the dynamic target.                                          
After=coreos-multipath-wait.target                                        
```

should we add that back here?

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-rescue-misplaced-network.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+set -euo pipefail
+
+# Rescue firstboot network config that an older coreos-installer placed
+# in /boot/coreos-firstboot-network.  Copy it to /etc so that
+# coreos-copy-firstboot-network can apply it normally.
```
dustymabe, line 6: Let's not do this. 

I think what we should do here is let the two services be exclusive (either one or the other runs, but not both).

So here instead of copying `/boot/coreos-firstboot-network` into `/etc/` we just copy it into the target destination: `/run/NetworkManager/system-connections/`

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-warn-misplaced-network.sh
dustymabe, line 1: I don't really think this emergency hook warning makes sense.

Since we have `coreos-rescue-misplaced-network.service`, which will still copy the files in /boot then we should basically have the same functionality as before.

To respond to the comment below:

> ```
> # Emergency hook: warn that an older coreos-installer may have written
> # firstboot network config to /boot, which this image no longer reads.
> # In the iSCSI case that missing config prevents the network from
> # coming up, which in turn prevents mounting root.
> ```

In the ISCSI case we never read network config from /boot - so we're not really dropping any functionality that we had previously.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/module-setup.sh
```diff
@@ -23,6 +23,11 @@ install() {
     install_and_enable_unit "coreos-enable-network.service" \
         "ignition-complete.target"
 
+    inst_simple "$moddir/coreos-rescue-misplaced-network.sh" \
+        "/usr/sbin/coreos-rescue-misplaced-network"
+    install_and_enable_unit "coreos-rescue-misplaced-network.service" \
+        "ignition-complete.target"
```
dustymabe, line 29: if we're part of `ignition-complete.target` then we'd need to add back logic in `coreos-diskful-generator` to neuter `coreos-rescue-misplaced-network.service` right? 

Another option may be to use a `ConditionPathExists=/dev/disk/by-label/boot` and `After=dev-disk-by\x2dlabel-boot.device`. Not sure if that would do it or not.

dustymabe (commented):
Thank you for working on this!

I think it would be good for us to put the backwards compatibility pieces into a separate overlay if we can and conditionally include them (i.e. include them in <= Fedora 45 and <= RHEL 10 so they natually drop out when Fedora 46 and RHEL 11 are here.

overlay.d/05core/usr/libexec/coreos-warn-misplaced-network
ravanelli: I think that's the case I tried to add here for ISCSI

```
# Emergency hook: warn that an older coreos-installer may have written
# firstboot network config to /boot, which this image no longer reads.
# In the iSCSI case that missing config prevents the network from
# coming up, which in turn prevents mounting root.
```

But indeed, it is a new feature, so users should know that before this version it won't work

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/module-setup.sh
ravanelli: ` After=dev-disk-by\x2dlabel-boot.device`  combined with Before=nm-initrd.service created a deadlock on iSCSI in my tests. 
 So even though the condition would fail on iSCSI (no /dev/disk/by-label/boot), systemd never gets to check it,  it's stuck waiting for the boot device to satisfy the After= ordering. Only `ConditionPathExists=/dev/disk/by-label/boot` worked

overlay.d/05core/usr/lib/systemd/system/coreos-warn-misplaced-network.service
ravanelli: No, if I understood the upgrade doesn't have the `ConditionFirstBoot=true`, so no warn is present. However, `coreos-rescue-misplaced-network` will make the copy, is the something particular with the upgrade that we may use to run it?

overlay.d/11legacy-copy-network/usr/lib/dracut/modules.d/36coreos-network-legacy/coreos-rescue-misplaced-network.sh
```diff
@@ -0,0 +1,48 @@
...
+}
+
+if [ -n "$(ls -A "${boot_firstboot_network_dir}" 2>/dev/null)" ]; then
+    warn "warning: found legacy firstboot network config in ${boot_firstboot_network_dir}"
+    warn "warning: an older coreos-installer placed network config in /boot instead of the initramfs"
+    warn "warning: rescuing config — copying to ${initramfs_network_dir}"
+    warn "warning: re-install using an updated coreos-installer to avoid this in the future"
+    # Clear out any files that may have already been generated from
+    # kargs by nm-initrd-generator
+    rm -f ${initramfs_network_dir}/*
```
dustymabe, line 39: so we can see what files were deleted:

```suggestion
    rm -vf ${initramfs_network_dir}/*
```

overlay.d/11legacy-copy-network/usr/libexec/coreos-warn-misplaced-network
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -euo pipefail
+
+MOTD_FILE="/etc/motd.d/coreos-warn-misplaced-network.motd"
+
+mkdir -p /etc/motd.d
+cat > "${MOTD_FILE}" <<'EOF'
```
dustymabe, line 7: could we potentially add some color here like:

https://github.com/coreos/fedora-coreos-config/blob/1300e3547ba43d86723c7d3afdfc333021119346/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/coreos-check-rootfs-size#L22-L32

manifests/fedora-coreos.yaml
```diff
@@ -91,6 +54,13 @@ conditional-include:
       ostree-layers:
         - overlay.d/16fcos-sshd-workaround
 
+  # Backward compatibility: rescue/warn firstboot network config
+  # coreos-installer placed in /boot. Drop when Fedora 46 lands.
+  - if: releasever <= 45
+    include:
+      ostree-layers:
+        - overlay.d/11legacy-copy-network
+
```
dustymabe, line 63: Let's also add a similar entry to `shared-el10.yaml` and `shared-el9.yaml`. Above those entries we should mention we wnat this to drop out in el11 so that's why we are putting them in a el9 and el10 specific manifest.

overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/module-setup.sh
dustymabe: hmm. my understanding is that `After=dev-disk-by\x2dlabel-boot.device` without `Requires=dev-disk-by\x2dlabel-boot.device` wouldn't cause things to block.. ahh, but coreos-diskful-generator pulls it into the transaction:

https://github.com/coreos/fedora-coreos-config/blob/1300e3547ba43d86723c7d3afdfc333021119346/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator#L110-L113

so yes we've created a deadlock and I think we need to now do the same thing we were doing before for `coreos-copy-firstboot-network.service` but now for `coreos-rescue-misplaced-network.service`

https://github.com/coreos/fedora-coreos-config/blob/1300e3547ba43d86723c7d3afdfc333021119346/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-diskful-generator#L135

don't get me wrong. what you have here is likely working, but I don't think the guarantees are strong enough that if a /boot device exists it will be waited on before this systemd service runs.

overlay.d/11legacy-copy-network/usr/lib/dracut/modules.d/36coreos-network-legacy/coreos-rescue-misplaced-network.service
```diff
@@ -0,0 +1,24 @@
+# If an older coreos-installer wrote firstboot network config to
+# /boot/coreos-firstboot-network, copy it directly to the NetworkManager
+# connections directory. Mutually exclusive with
+# coreos-copy-firstboot-network.service via ConditionPathExists.
+[Unit]
+Description=Rescue Misplaced Firstboot Network Config From /boot
+ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=!/etc/coreos-firstboot-network
+ConditionPathExists=/dev/disk/by-label/boot
+DefaultDependencies=false
```
dustymabe, line 10:
```suggestion
# We're pulled in by ignition-complete.target; as good practice, add a matching
# Before to be explicit about it gating on this unit passing.
Before=ignition-complete.target
```

similar to https://github.com/coreos/fedora-coreos-config/blob/1300e3547ba43d86723c7d3afdfc333021119346/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service#L37-L39

overlay.d/11legacy-copy-network/usr/lib/dracut/modules.d/36coreos-network-legacy/coreos-warn-misplaced-network.sh
dustymabe: I'm still not sure we need this.

i.e. I think you are trying to cover a case where the user read about the new feature (i.e. "`coreos-installer install --copy-network` now works with ISCSI"), but they used the old `coreos-installer` and it's not working.

Honestly I'm not too worried about outputing a warning there. Especially if we don't know that the emergency.target failure is related to that or not. It would likely be more confusing than anything.

overlay.d/11legacy-copy-network/usr/lib/dracut/modules.d/36coreos-network-legacy/coreos-warn-misplaced-network.sh
```diff
@@ -0,0 +1,27 @@
...
+------
+If you used 'coreos-installer install --copy-network' with an older
+version of coreos-installer, the network configuration may have been
+written to /boot/coreos-firstboot-network.
+
+This CoreOS image no longer reads firstboot network config from /boot.
+The configuration must be embedded in the initramfs.
+
+Re-install using an updated version of coreos-installer.
+------
```
dustymabe, line 25: I'm wondering if we should link the user to somewhere with more information on this.

overlay.d/11legacy-copy-network/usr/lib/dracut/modules.d/36coreos-network-legacy/coreos-rescue-misplaced-network.service
dustymabe, line 1: can we rename this dracut module to be closer to the overlay name:

```
36coreos-network-legacy/ -> 36legacy-copy-network/
```

manifests/shared-el10.yaml
```diff
@@ -1,4 +1,13 @@
 # Shared with EL 10 (CentOS Stream 10 and RHCOS 10) only
+
+# Backward compatibility: rescue/warn firstboot network config
+# coreos-installer placed in /boot. Drop when no longer needed.
```
dustymabe: Slightly more specific:

```suggestion
# Backward compatibility: rescue/warn firstboot network config
# coreos-installer placed in /boot. This will drop out in EL11+.
```

manifests/shared-el9.yaml
```diff
@@ -1,4 +1,13 @@
 # Shared with EL 9 (CentOS Stream 9 and RHCOS 9) only
+
+# Backward compatibility: rescue/warn firstboot network config
+# coreos-installer placed in /boot. Drop when no longer needed.
```
dustymabe: Slightly more specific:

```suggestion
# Backward compatibility: rescue/warn firstboot network config
# coreos-installer placed in /boot. This will drop out in EL11+.
```

manifests/shared-el10.yaml
```diff
@@ -1,4 +1,13 @@
...
+conditional-include:
+  - if: id != "fedora"
+    include:
+      ostree-layers:
+        - fedora-coreos-config/overlay.d/11legacy-copy-network
+
```
dustymabe: This whole file `shared-el10.yaml` is conditional so this include doesn't need to be conditional:

```suggestion

ostree-layers:
  - fedora-coreos-config/overlay.d/11legacy-copy-network

```

same goes for this same inclusion in `shared-el9.yaml`

overlay.d/11legacy-copy-network/usr/lib/dracut/modules.d/36legacy-copy-network/coreos-rescue-misplaced-network.service
```diff
@@ -0,0 +1,30 @@
...
+# /boot/coreos-firstboot-network, copy it directly to the NetworkManager
+# connections directory. Mutually exclusive with
+# coreos-copy-firstboot-network.service via ConditionPathExists.
+[Unit]
+Description=Rescue Misplaced Firstboot Network Config From /boot
+ConditionPathExists=/usr/lib/initrd-release
+ConditionPathExists=!/etc/coreos-firstboot-network
+ConditionPathExists=/dev/disk/by-label/boot
+# We're pulled in by ignition-complete.target; as good practice, add a matching
+# # Before to be explicit about it gating on this unit passing.
```
dustymabe:
```suggestion
# Before to be explicit about it gating on this unit passing.
```

overlay.d/11legacy-copy-network/usr/lib/dracut/modules.d/36legacy-copy-network/coreos-warn-misplaced-network.sh
dustymabe: to be more explicit, I think we should delete this emergency hook. Do you not?

overlay.d/11legacy-copy-network/usr/libexec/coreos-warn-misplaced-network
```diff
@@ -0,0 +1,27 @@
...
+During first boot, network configuration was found in
+/boot/coreos-firstboot-network. This was placed there by an older
+version of coreos-installer. The configuration was automatically
+rescued and applied, but this compatibility workaround will be
+removed in a future release.
+
+ACTION REQUIRED: Re-install using an updated coreos-installer that
+embeds network config in the initramfs instead of /boot.
+
+See https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-network-configuration/
```
dustymabe: not sure if this is the best URL - maybe we should make an entry in our FAQ for it?

https://docs.fedoraproject.org/en-US/fedora-coreos/faq/

dustymabe (commented):
This is getting close!

manifests/shared-el10.yaml
ravanelli: Actually Fedora does call it. And since the path differs for fcos and rhcos for the overlay, we need it

https://github.com/coreos/fedora-coreos-config/blob/2c30fa1be719fe86d5250cc082dc3ed61de8d11c/manifests/fedora-coreos.yaml#L9-L10

https://github.com/coreos/fedora-coreos-config/blob/2c30fa1be719fe86d5250cc082dc3ed61de8d11c/manifests/shared-el.yaml#L18-L20

https://github.com/coreos/fedora-coreos-config/blob/2c30fa1be719fe86d5250cc082dc3ed61de8d11c/manifests/shared-el.yaml#L9-L12

dustymabe: ok. thanks for tracking that down.

overlay.d/11legacy-copy-network/usr/libexec/coreos-warn-misplaced-network
```diff
@@ -0,0 +1,27 @@
...
+During first boot, network configuration was found in
+/boot/coreos-firstboot-network. This was placed there by an older
+version of coreos-installer. The configuration was automatically
+rescued and applied, but this compatibility workaround will be
+removed in a future release.
+
+ACTION REQUIRED: Re-install using an updated coreos-installer that
+embeds network config in the initramfs instead of /boot.
+
+See https://docs.fedoraproject.org/en-US/fedora-coreos/faq/
```
dustymabe: I was thinking a specific faq entry (link) that we added to explain this problem.

dustymabe (dismissed):
LGTM - a better link in the warning would be nice, but I'll let you handle that.

don't block on another review from me.

overlay.d/11legacy-copy-network/usr/libexec/coreos-warn-misplaced-network
```diff
@@ -0,0 +1,27 @@
...
+During first boot, network configuration was found in
+/boot/coreos-firstboot-network. This was placed there by an older
+version of coreos-installer. The configuration was automatically
+rescued and applied, but this compatibility workaround will be
+removed in a future release.
+
+ACTION REQUIRED: Re-install using an updated coreos-installer that
+embeds network config in the initramfs instead of /boot.
+
+See https://docs.fedoraproject.org/en-US/fedora-coreos/faq/#_why_am_i_getting_a_warning_about_misplaced_firstboot_network_configuration
```
dustymabe: Are we supposed to merge this before that link being live? Is there a PR to make it exist?

ravanelli: PR opened https://github.com/coreos/fedora-coreos-docs/pull/830

PR #4340, opened 2026-09-02
overrides: pin selinux-policy to 44.6-1.fc44

Rolv-Apneseth:
tests/kola: add test to verify podman-remote can connect to the podman socket over SSH

    This test uses podman-remote to connect to the root podman socket over
    SSH to localhost, catching SELinux regressions where sshd_session_t is
    denied connectto on container_runtime_t:unix_stream_socket.

    overrides: pin selinux-policy to 44.6-1.fc44

    This pin is to avoid an SELinux regression in 44.7-1+ where
    sshd_session_t is denied connectto on
    container_runtime_t:unix_stream_socket for podman, breaking SSH
    connections for podman-remote.

See https://github.com/coreos/fedora-coreos-tracker/issues/2217.

Test is a modified version of the suggestion from https://github.com/coreos/fedora-coreos-tracker/issues/2217#issuecomment-5512997176, thanks @Luap99. I chose to have the socket activation and key setup handled directly in the butane config though.

Verified that the test fails without the pin.

### DIFF
--- a/manifest-lock.overrides.yaml
+++ b/manifest-lock.overrides.yaml
@@ -34,3 +34,14 @@ packages:
     metadata:
       type: pin
       reason: https://github.com/coreos/fedora-coreos-tracker/issues/2216
+
+  selinux-policy:
+    evra: 44.6-1.fc44.noarch
+    metadata:
+      type: pin
+      reason: https://github.com/coreos/fedora-coreos-tracker/issues/2217
+  selinux-policy-targeted:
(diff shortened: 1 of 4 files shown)


tlbueno (approved):
lgtm

dustymabe (approved):
LGTM

dustymabe:
> echo 'Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.

This is intentional.

PR #4342, opened 2026-09-03
[testing] overrides: pin selinux-policy-44.6-1.fc44

coreosbot-releng:


dustymabe (approved):

dustymabe:
> Downgrade detected. This is likely unintentional. If not, you may safely ignore this error.

This is intentional. Merging.

PR #4346, opened 2026-09-06
[rawhide] overrides: pin bootc-1.16.9-1.fc46

coreosbot-releng:


dustymabe (approved):
LGTM

dustymabe:
The test failures in CI are `warn: true` tests. Merging.

PR #4349, opened 2026-09-08
tests/upgrade: select latest reachable upgrade release

dustymabe:
The previous code here had some logic for if the latest release in the index wasn't actually in the update graph (i.e. we had stopped the rollout), but it didn't handle the case where there were multiple releases where we had stopped the rollout, which is the case for the last few testing and next releases we've done.

Assisted-by: <openai/gpt-5.6-luna>


jbtrystram (approved):
LGTM

PR #4350, opened 2026-09-08
Containerfile: prefix Konflux build secret IDs

Roshan-R:
Konflux exposes ADDITIONAL_SECRET files using IDs in the form `<secret-name>/<filename>`. Prefix the yumrepos and contentsets secret IDs with container-image-build and update build-rootfs to use the corresponding contentsets path.

Depends on https://github.com/coreos/coreos-assembler/pull/4613

ref: https://github.com/coreos/rhel-coreos-config/issues/307

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -45,8 +45,8 @@ RUN chmod -R a=rX,u+w /src
 RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
         rm -rf /cache/cache/*lock*
 RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
-    --mount=type=secret,id=yumrepos,target=/etc/yum.repos.d/secret.repo \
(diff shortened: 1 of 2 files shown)


Rolv-Apneseth (approved):

joelcapitao (approved):
LGTM
As it's not used in f-c-c it will pass CI, but we need to make sure that https://github.com/coreos/fedora-coreos-config/pull/4350 is merged just before the `Bump fedora-coreos-config` PR that will sync it in r-c-c. If we merge 4350 now r-c-c build wil be broken.
Let's first try to merge the pending one https://github.com/coreos/rhel-coreos-config/pull/292

c4rt0 (approved):
I would probably modify description and say:

Depends on [coreos-assembler#4613](https://github.com/coreos/coreos-assembler/pull/4613)

/lgtm

PR #4353, opened 2026-09-08
Containerfile: quote labels to fix heredoc parsing without bash

joelcapitao:
The recent removal of the `bash` prefix from the heredoc (commit 6e2c0881) caused PROW to incorrectly parse labels containing spaces in ${DESCRIPTION} (e.g. "Red Hat Enterprise Linux CoreOS 9.8").

Without bash, buildah splits the value at spaces, resulting in:
  --label 'io...=machine-os="Red' Hat Enterprise Linux CoreOS '9.8"'

Wrapping the entire label argument in single quotes prevents this incorrect word splitting while preserving the inner double quotes for the label value.

Assisted-by: OpenCode (Claude Opus 4.5)"


joelcapitao:
Tested it with https://github.com/coreos/rhel-coreos-config/pull/310. The tests are still not finished but it passed the failure point.

joelcapitao:
This shoud unblock https://github.com/coreos/rhel-coreos-config/pull/292

Johan-Liebert1 (approved):

jbtrystram (approved):
Thanks!

PR #4354, opened 2026-09-09
tests(fwupd-refresh-units): skip timer check for EL9

joelcapitao:
The timer is not enabled on EL9 but the current condition does not skip SCOS9, but only RHCOS9. With this change, we skip the check for all EL9 distros.

The failure was caught with https://github.com/coreos/rhel-coreos-config/pull/292


Rolv-Apneseth (approved):
Ah, my bad, thanks @joelcapitao

Rolv-Apneseth:
Fix for test added/adjusted in https://github.com/coreos/fedora-coreos-config/pull/4240

PR #4356, opened 2026-09-09
Enable systemd-oomd and zram swap

Nemric:
Going to https://forge.fedoraproject.org/fesco/tickets/issues/3661
Add zram-generator.conf file to enable zram swap
Remove systemd-oomd preset overrides to enable systemd-oomd service and socket


manifests/fedora-coreos.yaml
```diff
@@ -54,6 +54,19 @@ conditional-include:
...
+  # See :
+  #   - https://fedoraproject.org/wiki/Changes/CoreOS_default_to_enable_systemd-oomd_and_swap_on_Zram
+  - if: releasever >= 45
+    include:
+      packages:
+        - zram-generator-defaults
+  - if: releasever < 45
+    include:
+      ostree-layers:
+        - overlay.d/05core/usr/lib/systemd/system-preset/40-coreos-systemd.preset
```
Rolv-Apneseth: I think we need to move that preset to it's own directory under `overlay.d/` so we can point this at e.g. `overalay.d/19_coreos_systemd_preset`?

Rolv-Apneseth: Or maybe the naming should reflect it's specifically for disabling systemd-oomd

Rolv-Apneseth (changes requested):

manifests/fedora-coreos.yaml
Nemric: Sorry, I didn't see your message, I did create the same 05-core with conditional suffix to differentiate regular vs conditional ones
Let me know if it's ok

angelcerveraroldan (commented):
Looks good to me! Thanks for working on this. I wonder should we add a small test for this too.

dustymabe:
F45 beta got approved for GO on Tuesday 09/15

I think we need to defer this to F46 since we now need to shift focus to building `next` based on F45.

Can you update the conditionals to be F46 ?

manifests/fedora-coreos.yaml
```diff
@@ -54,6 +54,20 @@ conditional-include:
...
+  #   - https://fedoraproject.org/wiki/Changes/CoreOS_default_to_enable_systemd-oomd_and_swap_on_Zram
+  - if: releasever >= 45
+    include:
+      packages:
+        - zram-generator-defaults
+  # For releases <45 we explicitely disable oomd service and socket
+  - if: releasever < 45
+    include:
+      ostree-layers:
+        - overlay.d/05core-conditional/usr/lib/systemd/system-preset
```
jbtrystram: Let's go with something more explicit, e.g `10-disable-systemd-oomd/`.
Also, you can inject the whole directory

jbtrystram (changes requested):

manifests/fedora-coreos.yaml
Nemric: Well, I think we don't need to add the `10-` prefix, but ok to rename the directory with its "feature" name only, WDYT ?

manifests/fedora-coreos.yaml
```diff
@@ -54,6 +54,20 @@ conditional-include:
       ostree-layers:
         - overlay.d/16fcos-sshd-workaround
 
+  # add zram-generator-defaults for releases >= 45
```
Rolv-Apneseth: Looks like the comments need to updated to 46

manifests/fedora-coreos.yaml
```diff
@@ -54,6 +54,20 @@ conditional-include:
...
+  #   - https://fedoraproject.org/wiki/Changes/CoreOS_default_to_enable_systemd-oomd_and_swap_on_Zram
+  - if: releasever >= 46
+    include:
+      packages:
+        - zram-generator-defaults
+  # For releases <45 we explicitely disable oomd service and socket
+  - if: releasever < 46
+    include:
+      ostree-layers:
+        - overlay.d/05core-conditional/usr/lib/systemd/disable-oomd
```
Rolv-Apneseth: As @jbtrystram suggested:

```suggestion
        - overlay.d/10-disable-systemd-oomd
````

File: overlay.d/05core/usr/lib/systemd/system-preset/40-coreos-systemd.preset [hunk 5]
```diff
@@ -1,6 +1,3 @@
 # This file contains overrides for systemd services that are
```
Rolv-Apneseth: Do we still need this file if it was only used to disable oomd?

Nemric: I don't know, I was thinking about a template file for "new comers" ^^
It's useless now, you're right, let me know if it's a kind of template or not

Nemric: :/ no need to overload the repo, I'll remove it

jbtrystram: let's keep it as scaffolding for the next time we need to disable something.

jbtrystram (approved):
Thanks @Nemric !

angelcerveraroldan (changes requested):
Looks good to me too! As discussed in element, I  think that the only change needed now is [this test](https://github.com/coreos/fedora-coreos-config/blob/5bc80391aa27ce7ed729b3329e12f733ae8da836/tests/kola/swap/zram-default), which will currently fail in `rawhide`.

PR #4359, opened 2026-09-10
Containerfile: use bash array for labels with spaces in heredoc

joelcapitao:
Bash parameter expansion with alternate value syntax (`${VAR:+value}`) does not handle word-splitting correctly inside heredocs when the substituted value contains spaces. The `DESCRIPTION` variable contains "Red Hat Enterprise Linux CoreOS 9.8" which caused the build to fail with truncated arguments.

Using a bash array and expanding it with `"${EXTRA_ARGS[@]}"` preserves proper quoting and prevents word-splitting.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -51,6 +51,11 @@ RUN --mount=type=cache,rw,id=coreos-build-cache,target=/cache \
 # Take the rootfs and created a chunked container out of it
 RUN --mount=type=bind,target=/run/src,rw <<EOF
     set -eux -o pipefail
+    EXTRA_ARGS=()
+    if [ -n "${INJECT_OPENSHIFT_VERSION_LABELS}" ]; then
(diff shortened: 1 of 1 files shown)


joelcapitao:
This was tested with https://github.com/coreos/rhel-coreos-config/pull/316

```
skopeo inspect docker://quay.io/redhat-user-workloads/coreos-tenant/rhel-coreos-9-8:on-pr-b89048d38447c05dbf21fa4fc8a695c2fdf683c3 | jq .Labels

  "io.openshift.build.version-display-names": "machine-os=Red Hat Enterprise Linux CoreOS 9.8",
  "io.openshift.build.versions": "machine-os=9.8.20260910-2",

```

jbtrystram (approved):

yasminvalim (approved):
/LGTM

PR #4361, opened 2026-09-10
[next-devel] Start tracking Fedora 45

marmijo:
See: https://github.com/coreos/fedora-coreos-tracker/issues/2127

PR #4362, opened 2026-09-11
manifest/fcos: add workaround back

angelcerveraroldan:
When reverting to bind mounting [1], we also need to add the workaround that ensures that the path we mount to exists.

This brings back the workaround [2] that I removed too early. It also extends it to f46 as well as f46. We should be able to delete this relatively soon, when this lands: coreos/rpm-ostree#5630 (without any errors this time :D).

[1] cff712a2aad99511a6c2ca01adfeb48c9892a5ff
[2] 9736538a8f7f6fe305accbd2269c65d2a37385ed

### DIFF
--- a/manifests/fedora-coreos.yaml
+++ b/manifests/fedora-coreos.yaml
@@ -35,6 +35,24 @@ conditional-include:
     include:
       ostree-layers:
         - overlay.d/10aarch64
+  # Temporary workaround. containers-common moved policy.json from
+  # /etc/containers to /usr/share/containers. The rpm-ostreed bind mount
+  # could use CONTAINERS_POLICY_JSON env var instead, but bootc has a
+  # hardcoded path that prevents this approach.
+  # Copy the file until bootc supports the new config location.
+  #
+  # See:
+  # https://github.com/bootc-dev/bootc/issues/2258
+  # https://github.com/coreos/fedora-coreos-config/issues/4215
+  - if: releasever >= 45
+    include:
+      postprocess:
+        - |
+          #!/usr/bin/bash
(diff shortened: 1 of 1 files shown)


dustymabe:
@angelcerveraroldan I suggested in https://github.com/coreos/fedora-coreos-config/pull/4361#issuecomment-5626189122 that we just fast-track bootc. WDYT of that instead of this?

angelcerveraroldan:
@dustymabe that sounds good to me, but we would need to fast track bootc as well as rpm-ostree right ? Since having the new patched `bootc` version, but the old `rpm-ostree` one will lead to the same issue we saw.

angelcerveraroldan:
I think that we already have a new enough version of bootc without needed to fastrack, since the fix landed in 1.16.3: https://github.com/bootc-dev/bootc/releases#release-v1.16.3

Rolv-Apneseth:
@dustymabe IIUC a new `bootc` doesn't do anything here, the issue was that `rpm-ostree` had an old version of `ostree-ext` from `bootc` pinned (it consumes it as a git source [here](https://github.com/coreos/rpm-ostree/blob/bd9ad6bce96bf2d1b78967ccda4433833fe02924/Cargo.toml#L76)). So all we need is https://github.com/coreos/rpm-ostree/releases/tag/v2026.3

dustymabe:
ok makes sense to me now. Took some time to follow all the problems but I understand.

LGTM

dustymabe (approved):
LGTM

Rolv-Apneseth:
> Took some time to follow all the problems but I understand.

Yep a lot going on at the moment it seems

PR #4373, opened 2026-09-14
overrides: fast-track rust-bootupd-0.3.2-1.fc44

coreosbot-releng:
Requested by @joelcapitao via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


jbtrystram (approved):

dustymabe:
we need to target both `testing-devel` and `next-devel` for fast-tracks for now while `next` is decoupled from `testing`. 

Can you open a fast-track there too?

PR #4375, opened 2026-09-15
Update Konflux pipeline to latest digest

joelcapitao:


jbtrystram (approved):
LGTM. Note that we should update the `fcos-buildroot` references as well

Rolv-Apneseth:
Would this be the fix for https://github.com/coreos/fedora-coreos-config/pull/4381#issuecomment-5700887179?

jbtrystram:
> Would this be the fix for https://github.com/coreos/fedora-coreos-config/pull/4381#issuecomment-5700887179?

Nope. This seems to be caused by stale lockfiles

PR #4376, opened 2026-09-16
[rawhide] overrides: fast-track ignition-2.27.0-3.fc46

coreosbot-releng:
Requested by @Rolv-Apneseth via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


Rolv-Apneseth:
CI failure was `ext.config.podman.remote-ssh` (https://github.com/coreos/fedora-coreos-tracker/issues/2217). Might need a denylist entry for now.

Rolv-Apneseth:
See https://github.com/coreos/fedora-coreos-tracker/issues/2216

```
10:43:50  --- PASS: ext.config.ignition.home-dir-symlink (37.15s)
```

dustymabe:
Ignition is already `stable` in bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2026-a0298b69a7

rather than merge this bump-lockfile needs to run successfully and then this will be unblocked?

Rolv-Apneseth:
Ah, is that how it works? Works for me if that's the case

dustymabe:
> Ah, is that how it works? Works for me if that's the case

yeah we just need to work to get bumplockfile green and then we're good.

Rolv-Apneseth:
@dustymabe same for https://github.com/coreos/fedora-coreos-config/pull/4379 then I assume?  https://bodhi.fedoraproject.org/updates/FEDORA-2026-93d15da3c4 is stable too

PR #4378, opened 2026-09-16
overrides: fast-track ignition-2.27.0-1.fc44

coreosbot-releng:
Requested by @Rolv-Apneseth via [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/add-override.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/add-override.yml)).


Rolv-Apneseth:
Wrong version, ignore

PR #4386, opened 2026-09-16
Generate unique NVMe hostnqn

aaradhak:
[OCPBUGS-100442](https://redhat.atlassian.net/browse/OCPBUGS-100442) - fix duplicate or missing NVMe identities on fcos nodes.

The identities were generated during image creation, so nodes booted from the same image could share the same `hostnqn` and `hostid` preventing reliable NVMe-oF storage access.

The first commit adds the first-boot service and systemd preset to generate unique values. 
The second commit adds kola test for the identity files and reboot behavior.


dustymabe:
Based on the discussion in: 

- https://redhat.atlassian.net/browse/RHEL-8041
- https://redhat.atlassian.net/browse/OCPBUGS-34629 (older version of OCPBUGS-100442 it seems).

Maybe we just do this on el9 only (i.e. not in this repo but back in `rhel-coreos-config`). It seems like el10 runs stateless (no `/etc/nvme/{hostid,hostnqn}`) well enough.

This allows us to have a timeboxed (i.e. only el9) workaround of the bug they wouldn't fix in RHEL 9.8 (RHEL-8041) that is now fixed in RHEL10+ (https://redhat.atlassian.net/browse/RHEL-68495).

aaradhak:
> Based on the discussion in:
> 
> * https://redhat.atlassian.net/browse/RHEL-8041
> * https://redhat.atlassian.net/browse/OCPBUGS-34629 (older version of OCPBUGS-100442 it seems).
> 
> Maybe we just do this on el9 only (i.e. not in this repo but back in `rhel-coreos-config`). It seems like el10 runs stateless (no `/etc/nvme/{hostid,hostnqn}`) well enough.
> 
> This allows us to have a timeboxed (i.e. only el9) workaround of the bug they wouldn't fix in RHEL 9.8 (RHEL-8041) that is now fixed in RHEL10+ (https://redhat.atlassian.net/browse/RHEL-68495).

I had opened a [PR](https://github.com/coreos/rhel-coreos-config/pull/325) earlier in rhel-coreos-config. There were some [suggestions](https://github.com/coreos/rhel-coreos-config/pull/325#discussion_r4023775107) on adding this to the upstream because the NVMe identity handling appeared generally useful, so I opened this PR here.

dustymabe:
> There were some [suggestions](https://github.com/coreos/rhel-coreos-config/pull/325#discussion_r4023775107) on adding this to the upstream

I'm not sure the `upstream` he was referring to there was f-c-c, but could be wrong.

Either way, I don't think this is a problem in Fedora or RHEL 10 (someone correct me if I'm wrong). If that's true we should put this in rhel-coreos-config *and* limit it to el9 only.

tlbueno (dismissed):

aaradhak:
Closing this PR as the change needs to be added only to el9

PR #4387, opened 2026-09-17
manifest/fedora-coreos: remove all DTBs on aarch64 for F45+

dustymabe:
It's suspected these are never used so let's delete them because they are causing extra pressure on our precious amount of space in the /boot/ partition.

See https://github.com/coreos/fedora-coreos-tracker/issues/2222


marmijo (approved):
LGTM

angelcerveraroldan:
LGTM, thanks.

Do we need to make this change to next-devel too? Or will it sync?

Rolv-Apneseth (approved):

dustymabe:
> Or will it sync?

it will sync and then the conditional will mean all DTBs get removed because next-devel is on 45+

PR #4389, opened 2026-09-21
Do not mount RO the secret.repo in `yum.repos.d`

joelcapitao:
We need those changes as a prep work to enable RHCOS hermetic build in Konflux
See the commit messages for more details


Johan-Liebert1 (approved):

build-rootfs
```diff
@@ -103,16 +103,16 @@ def get_treefile(manifest_path, osid, stream, version):
...
         if repo == HERMETIC_REPO:
             # this is the repo Konflux injects when hermetic build is enabled
             continue
         os.unlink(repo)
 
     # and now inject our repos
     if not IS_HERMETIC:
+        secret_repo = '/run/secrets/yumrepos'
+        if os.path.exists(secret_repo):
+            shutil.copy(secret_repo, '/etc/yum.repos.d/secret.repo')
```
dustymabe, line 115: I think if we use `--mount=type=secret,id=container-image-build/yumrepos,target=/etc/yum.repos.d/secret.repo` podman/buildah will take care to make sure `/etc/yum.repos.d/secret.repo` isn't in the resulting container image.

If we copy it I'm not sure if that still happens.

joelcapitao: I doubled check and there is nothing in `/etc/yum.repos.d/secret.repo`

```
$ podman run --rm -it quay.io/redhat-user-workloads/coreos-tenant/rhel-coreos-10-2:on-pr-308b4410f7e6fbaf85fa6d9374e49f395a8390ab-linux-amd64 ls -la /etc/yum.repos.d/
total 0
drwxr-xr-x. 1 root root  0 Jan  1  1970 .
drwxr-xr-x. 1 root root 38 Sep 22 16:14 ..
```

My understanding is that `/etc/yum.repos.d/secret.repo` is only available in the builder image before rpm-ostree composing the rootfs

joelcapitao: The container image comes from https://github.com/coreos/rhel-coreos-config/pull/328

PR #4391, opened 2026-09-21
overrides: drop console-login-helper-messages pin

aaradhak:
Drop the temporary pins for console-login-helper-messages and its helper packages now that the selinux-policy change is available.

Resolves the temporary override tracked by https://github.com/fedora-selinux/selinux-policy/pull/3169.


jbtrystram:
This can't go in until we have dropped the selinux-policy pin : https://github.com/coreos/fedora-coreos-config/commit/f5bd99f74801e518632bec91a975e2cfe4504151

PR #4392, opened 2026-09-21
kola-denylist: snooze ext.config.upgrade.extended

prestist:
## Summary

- Snooze `ext.config.upgrade.extended` until 2026-10-21 while the upstream rpm-ostree fix lands
- kola-upgrade #6659 (next-devel/x86_64) fails consistently on both BIOS and UEFI-SECURE firmware types, initial run and retry, with `Unexpected mislabeled file found: /var/lib/selinux/targeted`

## Root cause

rpm-ostree PR [#5620](https://github.com/coreos/rpm-ostree/pull/5620) (included in `45.20260917.1.2`) added a `file_contexts.subs_dist` equivalence rule mapping `/var/lib/selinux` → `/etc/selinux`. This changes the expected SELinux label from `semanage_store_t` to `selinux_config_t` for systems that still have the legacy real directory (pre-`44.20260705.1.1` layout). The upgrade test starts from Fedora 37, so it reliably hits this.

Tracked in: https://github.com/coreos/fedora-coreos-tracker/issues/2226

The fix direction per @cgwalters is a `tmpfiles.d` relabel rule in rpm-ostree for `/var/lib/selinux`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


c4rt0 (approved):
LGTM

dustymabe:
hmm. I'm actually not sure if we want to do this.

The upgrade test itself is non-blocking. The alerts are noisy, which is a problem, but there's a [fix in flight](https://github.com/coreos/rpm-ostree/pull/5633) and we'll try to fast-track it ASAP.

In the very least we should re-target this snooze to be for `next-devel` (and maybe `rawhide` if it's failing there), but we'd immediately want to un-snooze it when the fix lands.

Maybe the right approach is just make it warn:true with no snooze specifically on those branches.

dustymabe:
> but there's a https://github.com/coreos/rpm-ostree/pull/5633 and we'll try to fast-track it ASAP.

- https://github.com/coreos/fedora-coreos-config/pull/4399
- https://github.com/coreos/fedora-coreos-config/pull/4400

prestist:
Closing this in favour of the upstream fix being fast-tracked via #4399 and #4400. Once those land and builds run, `ext.config.upgrade.extended` should start passing again without needing a snooze. Thanks for the reviews!

